PR #24, opened 2018-12-18
Issue 16 exptree

### LINKED ISSUE #16: use an expression tree to encode a model
After discussions, we've decided to switch to using an expression tree to encode each model and submodel (see #14). The idea is that this would decouple the different aspects of PyBaMM (e.g. model/set_parmeters/discretisation/solver) and allow for a pipeline approach, where the expression tree is mutated by each aspect in turn.

This issue is to provide an initial implementation of the expression tree so that it can start being used by the rest of the code.

## Possible class hierarchy:

1. `class Symbol`: base class, holds list of child nodes, has a (string) name
     - operators used to build up tree defined here (e.g. `symbol + symbol` returns `Addition(symbol,symbol))`
     - note: could we use anytree to provide basic tree functionality?: https://anytree.readthedocs.io/en/2.4.3/index.html
2. `class UnaryOperator(Symbol)`: provides a base class for all the unary operators (e.g. Negate)
3. `class BinaryOperator(Symbol)`: same but for binary ops
4. `class Addition(BinaryOperator)`: implementation of the addition op, 
5. `class Laplacian(UnaryOperator)`: implementation of the laplacian op.
6. many other ops....
7. `class Parameter(Symbol)`: a parameter (symbolic, no value). perhaps with units?
8. `class Variable(Symbol)`: a variable that will (eventually) be discretised over the mesh. At the moment it is purely symbolic
9. `class Vector(Symbol)`: a node representing a vector (the spatial discretisation class will add these in)
10. `class Matrix(Symbol)`: a node representing a matrix (the spatial discretisation class will add these in)

## iteration through tree

provide functionality for iterating through the tree, e.g.

```python
[node.name for node in PreOrderIter(f)] # pre-order traversal
[node.name for node in PostOrderIter(f)] # post-order traversal
[node.name for node in LevelOrderIter(f)] # level-order (i.e. breadth-first)
```
(note: these examples come from the anytree library)

martinjrobins:
Fixes #16 

## Type of change

# Key checklist:

## Further checks: 


tinosulzer:
Can we change to `evaluate(self, t, y)` for full generality? Some objects will need to see `t` (e.g. current function might be time-dependent)

PR #44, opened 2018-12-21
#36 add power binary operator, and unary negate and abs operators

### LINKED ISSUE #36: Non-spatial unary operators
Symbol types might need `__neg__`, `__pow__`, `__abs__`, others?

martinjrobins:
# Description

Fixes #36 

## Type of change

# Key checklist:

## Further checks: 

### DIFF
--- a/docs/source/expression_tree/binary_operator.rst
+++ b/docs/source/expression_tree/binary_operator.rst
@@ -7,6 +7,9 @@ Binary Operators
 .. autoclass:: pybamm.Addition
   :members:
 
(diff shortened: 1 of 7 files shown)


tinosulzer:
Could `pow` also take a float as `other`?

martinjrobins:
it could, but I thought it would be more generally useful to have `pow` as a binary operator.

What might be useful is to overload the `__pow__` function on a `Symbol` to take either an `int`, `float` or another symbol. If it is an `int` or `float` just convert it into a `Scalar`

PR #46, opened 2018-12-21
Issue 19 refactor base model

### LINKED ISSUE #19: refactor BaseModel to hold an expression tree
See #16. 

The base model will internally hold an expression tree representing the rhs of an ode model. e.g.

```python
class BaseModel:
   def __init__(self):
         self._rhs = ... # lhs is a dictionary mapping variables to expressions
         self._initial_conditions = ... # a dictionary mapping expression to expressions representing the initial conditions
         self._boundary_conditions = ... # a dictionary mapping expression to expressions representing the boundary conditions
   @property
   def rhs(self):
          return self._rhs
   ...
```

Note: rather than having `rhs` return a dict, it might be nicer to have `BaseModel` emulate a dictionary, so you could write `model[variable]` and get the rhs expression associated with that variable (see https://docs.python.org/3/reference/datamodel.html?emulating-container-types#emulating-container-types)

Scottmar93:
# Description

Refactored BaseModel class to have:  
- empty dictionaries (as these will be set later) for _rhs, _initial_conditions, _boundary_conditions
- setters and getters for each dictionary using @property and @rhs.setter etc
- assigned dictionary behaviour so that rhs expressions can be accessed by model[" "] 

Note these changes break the reaction_diffusion tests etc but don't think we can get around that given the changes we are making (I have skipped these tests for now) 

Fixes #19 

## Type of change
- [x] Refactor

# Key checklist:

## Further checks: 

(I don't understand this final checkpoint)


Scottmar93:
okay merge. is that what the last checkpoint refers to?

pybamm/models/core.py
```diff
@@ -4,131 +4,52 @@
...
-        self.name = "Base Model"
-        # Assign tests as an attribute
-        if tests:
-            assert set(tests.keys()) == {
-                "inits",
-                "bcs",
-                "sources",
-            }, "tests.keys() must include, 'inits', 'bcs' and 'sources'"
-        self.tests = tests
+    _rhs: dict
```
martinjrobins:
```suggestion
    rhs: dict
```

pybamm/models/core.py
```diff
@@ -4,131 +4,52 @@
...
+        A dictionary that maps expressions (variables) to expressions that represent
+        the rhs
+    _initial_conditions: dict
```
martinjrobins:
```suggestion
    initial_conditions: dict
```

pybamm/models/core.py
```diff
@@ -4,131 +4,52 @@
...
+        A dictionary that maps expressions (variables) to expressions that represent
+        the initial conditions
+    _boundary_conditions: dict
```
martinjrobins:
```suggestion
    boundary_conditions: dict
```

martinjrobins:
> okay merged. is that what the last checkpoint refers to?

"last checkpoint": I'm not sure, what context was this in?

Scottmar93:
> "last checkpoint": I'm not sure, what context was this in?

I was referring to the last checkpoint on the pull request template

>Cheers @Scottmar93 , looks good. Could you just merge the new master branch and resolve the conflicts?

> okay merged. is that what the last checkpoint refers to?

> ## Further checks:
>  [ ]  Any dependent changes have been merged and published in downstream modules

martinjrobins:
ahhh. No, that says that if the changes you make in this PR break any other modules/classes, then you need to fix those too.

PR #47, opened 2018-12-21
Issue 37 standard parameters

### LINKED ISSUE #37: ParameterValues class for a specific model
Building on issue #18 / PR #35 

For a particular, model, parameter values may be needed that are combinations of "raw" parameter values. For example, the total electrode length `L = Ln + Ls + Lp`, or any dimensionless parameter. 

It doesn't make sense to write these all in the csv file that gets imported, so there needs to be a class that calculates the "derived" parameters, maybe split by domain ("negative_electrode", "separator", "positive_electrode" or "whole_cell"). There may be overlap between the domains.
This class should also contain scales for non/re-dimensionalisation

For example,
```python3
class DiffusionParameterValues(BaseParameters):
    def __init__(self, base_parameters={}, optional_parameters={}):
       super().__init__(base_parameters, optional_parameters)
       
       self.scales = {"length": self.raw["Ln"] + self.raw["Ls"] + self.raw["Lp"],
                      "time": ...,
                      etc}
       self.whole_cell = {"L": self.raw["Ln"] + self.raw["Ls"] + self.raw["Lp"],
                          "Cd": self.scales["length"]**2/self.raw["D"]/self.scales["time"],
                          etc}
       self.negative_electrode = {"L": self.raw["Ln"], etc}
       self.separator = {"L": self.raw["Ls"], etc}
       self.positive_electrode = {"L": self.raw["Ln"], etc}
```
Maybe make `DiffusionParameterValues` emulate a dictionary, similarly to #19 , with keys being domain names.

Now if we want to include new parameters, say for a temperature model, we will need more scales and derived parameters, but these should be in a separate class as they might call upon `raw` values that would not necessarily be given to a `DiffusionParameterValues` class.

```python3
class TemperatureParameterValues(BaseParameters):
    def __init__(self, base_parameters={}, optional_parameters={}):
       super().__init__(base_parameters, optional_parameters)
       
       self.scales = {"Temp": self.raw["T_init"]}
       self.whole_cell = {"k": self.raw["D_temp"]/ bla bla}
       self.negative_electrode = {}
       self.separator = {}
       self.positive_electrode = {}
```
Then a `BatteryParameterValues` could multiply-inherit from `DiffusionParameterValues` and `TemperatureParameterValues`, combining the dictionary attributes from each of these (not sure how to do this)

@martinjrobins @Scottmar93 do you think this is a good approach? See [previous implementation](https://github.com/tinosulzer/PyBaMM/blob/structure/pybamm/parameters/parameter_values.py) (quite different) for an idea of which derived parameters need to be implemented.

tinosulzer:
# Description

Add a few geometric parameters to a module `standard_parameters`. More parameters should be added to this as they appear in models, if they are standard battery parameters. This file could later be split into different files for different chemistries.
- Not sure whether/how to document a module that doesn't contain any classes?
- Could now rename `BaseParameterValues` to `ParameterValues` as that function won't need to be overloaded?

Fixes #37 

## Type of change

# Key checklist:

## Further checks: 


tinosulzer:
This should now be ready to merge

PR #53, opened 2018-12-21
Issue 21 solver

### LINKED ISSUE #21: refactor solver class
see #16.

This class will take an instance of a model that has been processed by the Parameter (#18) and Spatial Discretisation (#20) classes. That is, the solver can assume that all Parameter, Variable and Operator nodes representing gradients (e.g. Laplace) nodes in the expression tree have been replaced with concrete Vector and Matrix nodes. That is, the expression tree now represents a concrete linear algebra expression

For now we only consider Models representing ODE equations and solvers implementing of methods of lines (MOL). But the implementation should be flexible enough to be extended to DAE equations, and time discretistion with a newton iteration solver rather than MOL.

The solver class needs to:
   1. verify that the model rhs expressions are concrete linear algebra expressions
   2. loop over time:
           2.1. evaluate RHS as a function of current state variable `y` and `t`
           2.2. step forward in time using RHS
 
Solver class should store the current state vector internally, and have an `integrate` function that can move the simulation forward in time, to allow for output/plotting over time, e.g.

```python
model = ReactionDiffusionModel()
param_file = Parameters("file.dat")
param_file.process(model)
disc = FiniteVolumeDiscretisation(mesh_dx=0.1, something_else=2)
y0 = disc.process(model)
solver = RungeKuttaSolver(y0, tolerance=1e-5)
Tf = 1.0
nout = 100
nout_dt = Tf / nout
for i in range(nout):
   solver.integrate(nout_dt)
   plot(solver.y, disc)
```

tinosulzer:
# Description

Base solver class and scipy solver class (using `scipy.integrate.solve_ivp`).
Also fixes a bug in `discretisation.process_symbol` where children of binary operators were being swapped halfway (so the left child was being copied)

Fixes #21 

## Type of change

# Key checklist:

## Further checks: 


pybamm/solvers/base_solver.py
```diff
@@ -0,0 +1,84 @@
...
+    @property
+    def y(self):
+        return self._y
+
+    @y.setter
+    def y(self, value):
+        self._y = value
+
+    def solve(self, model, t_eval):
+        """Calculate the solution of the model at specified timesself.
```
martinjrobins:
```suggestion
        """Calculate the solution of the model at specified times.
```

pybamm/solvers/base_solver.py
```diff
@@ -0,0 +1,84 @@
...
+
+        Parameters
+        ----------
+        derivs : method
+            A function that takes in t and y and returns the time-derivative dydt
+        y0 : numeric type
+            The initial conditions
+        t_eval : numeric type
+            The times at which to compute the solution
+
```
martinjrobins: we should have a common return type(s) for all the derived classes that implement this function. Either just `y` and `t` as 2D and 1D numpy arrays, or a custom class that we create that will hold `y` and `t`, as well as diagnostic stuff (solver errors etc)

tinosulzer: Good point - see update. I think this is ready to merge

PR #61, opened 2018-12-22
Issue 49 refactor stefan maxwell diffusion

### LINKED ISSUE #49: Refactor StefanMaxwellDiffusion class
This issue must be completed prior to refactoring the reaction diffusion model #17

StefanMaxwellDiffusion should inherit from BaseModel so that it is a complete model in its own right. It will contain a dictionary with the set of equations in the electrolyte implemented in symbolic form employing the expression tree. 

Will looks something like:

```
class StefanMaxwellDiffusion(BaseModel): 
        def __init__(self): 
            self._rhs = {c: "epression"}
            self._initial_conditions = .... etc 

           self._variable = {"c": c, ...} 
```

EDIT. 
The keys should be of class Symbol expect for the _variable dictionary.

Scottmar93:
# Description

Refactored StefanMaxwellDiffusion class to create an expression tree. Only a simple test to see if the tree is created without errors is implemented. This will need to be revisited when: 
- power operator is implemented for class Parameter
- can write `rhs = {c: - grad(c)}` instead of `rhs = {c: Symbol(0) - grad(c)}` 
- dimensionless parameters are created and set by the parameter class  (I was unsure how to set a parameter that is different in different regions). 

Fixes #49 

## Type of change

# Key checklist:

## Further checks: 


pybamm/models/submodels/electrolyte.py
```diff
@@ -3,180 +3,61 @@
...
+    *Extends:* :class:`BaseModel`
+    """
+
+    def __init__(self, G):
+        super().__init__()
+
+        electrolyte_domain = [
+            "negative electrode",
+            "separator",
+            "positive " "electrode",
```
tinosulzer:
```suggestion
            "positive electrode",
```

pybamm/models/submodels/electrolyte.py
```diff
@@ -3,180 +3,61 @@
...
+        ]
+
+        c_e = pybamm.Variable("c_e", domain=electrolyte_domain)
```
tinosulzer: The way discretisation is currently implemented `electrolyte_domain` needs to be `["whole cell"]` - otherwise, I don't know how to implement boundary conditions between subdomains. It might be that defining `elecrolyte_domain` as you've done here is clearer, and we need to change `discretisation`.

pybamm/models/submodels/electrolyte.py
```diff
@@ -3,180 +3,61 @@
...
+
+        delta = pybamm.Parameter("delta")
```
tinosulzer: I'm not sure what these parameters are, but if they don't appear as "raw" parameters in a parameter csv file they need to be defined as a combination of raw parameters here. e.g. 
```python3
a = Parameter("something in the csv file")
b = Parameter("something else in the csv file")
delta = a / b
```
If such "combined" parameters are standard battery parameters, they can be defined in and imported from the module `standard_parameters` to avoid having to repeatedly define them in submodels. See issue #37

pybamm/models/submodels/electrolyte.py
```diff
@@ -3,180 +3,61 @@
...
+        epsilon = pybamm.Parameter("epsilon", domain=electrolyte_domain)
+        b = pybamm.Parameter("b")
+        D_e = pybamm.Parameter("D_e")
+        nu = pybamm.Parameter("nu")
+        t_plus = pybamm.Parameter("t_plus")
+
+        c_e0 = pybamm.Parameter("c_e0")  # Should this be a parameter?
```
tinosulzer: Yep parameter is correct here - will be converted to Scalar when setting parameters, then to Vector when discretising

pybamm/models/submodels/electrolyte.py
```diff
@@ -3,180 +3,61 @@
...
+
+        # Change expression once Binary operations can cope with None input
+        N_e = pybamm.Scalar(0) - D_e * (epsilon * b) * pybamm.Gradient(c_e)
```
tinosulzer: Negate and Power operators implemented in the branch `issue-45-overloaded-binary-ops`

pybamm/models/submodels/electrolyte.py
```diff
@@ -3,180 +3,61 @@
...
+        # power operator is apparently not defined for class parameters
+
+        self.rhs = {
+            c_e: pybamm.Scalar(0)
+            - pybamm.Divergence(N_e) / delta / epsilon
+            + nu * (pybamm.Scalar(1) - t_plus) * G
```
tinosulzer: Negate operator implemented in the branch `issue-45-overloaded-binary-ops`; also, `1-t_plus` should work with the fix in that branch (`1` will automatically be converted to `Scalar(1)`)

pybamm/models/submodels/electrolyte.py
```diff
@@ -3,180 +3,61 @@
...
+        }
+        self.initial_conditions = {c_e: c_e0}
+        self.boundary_conditions = {N_e: {"left": 0, "right": 0}}
```
tinosulzer: Discretisation currently assumes the boundary conditions are a tuple (left, right) - though I do think this way is clearer, will need to change discretisation for this form to work. `0` should be `Scalar(0)`.

pybamm/models/submodels/electrolyte.py
martinjrobins: note that this is now merged with master

pybamm/models/submodels/electrolyte.py
martinjrobins: agreed that this is clearer, I made an issue on this: #62

tinosulzer: now updated and merged with master

Scottmar93: I think the param.process function also assumes a tuple. I was getting an error when I tried to process the boundary conditions in a test I was creating

tinosulzer: Good spot, have opened #65 for this

PR #77, opened 2019-01-15
Issue 67 variable domain params

### LINKED ISSUE #67: discretisation of different parameter values in different domains
Different parameters can have different values in different domains. This can be represented in a model by a concatentation of Parameters, which will change to a concat of Scalars. The discretisation class need to change this to a concatenation of `Vectors`.

martinjrobins:
# Description

Fixes #67 

@tinosulzer: I used your old `scalar_to_vector` function in the discretisation class to implement this, then merged master and found out that you had deleted it. I've re-added `scalar_to_vector`, can you have a look and make sure it makes sense with your changes.

## Type of change

# Key checklist:

## Further checks: 


tinosulzer:
@martinjrobins feel free to merge and make a separate issue for the scalar + variable case if you feel that is more appropriate than editing this one

martinjrobins:
> P.S. note that we have moved to max line length of 88 as per recommendation here https://black.readthedocs.io/en/stable/the_black_code_style.html#line-length.
> We should agree on 79/80 vs 88, otherwise there will be a lot of unnecessary automatic edits happening

Good to know, whatever is in the `.flake8` in the repo my autoformatter should respect. Note sure how this works with vims default line width though... I'll keep an eye on it.

PR #83, opened 2019-01-15
Issue 82 domains

### LINKED ISSUE #82: Give domain to all expressions
Currently, only `Variable`, `Parameter` and `Scalar` nodes have domains. It would be useful to be able to evaluate the domain of an arbitrary node, e.g. `2*var` which is a Multiplication.
Proposed edit is:
- For a `UnaryOperator`, inherit domain from child (can be `None`)
- For a `BinaryOperator`, check that children have the same domain (or either or both `None`) and inherit that domain
- Additionally, check that key and value have the same domain in `model.rhs` and `model.initial_conditions` (via the respective setters in `BaseModel`)

This will help with #78 , and allow us to not have to pass `domain` as an argument in the discretisation functions.
@martinjrobins thoughts on this?

tinosulzer:
# Description

Add domain to all nodes, with inheritance from children nodes for operators and checks that domain combinations make sense
- Merge `Domain` class into `Symbol` class (default domain is `[]`)
- Unary operator inherits from child
- Binary operator checks that children have same (or empty) domain and inherits 
- Concatenation checks that children have disjoint domains and inherits
- BaseModel checks that variable/equation pairs have the same domain in `rhs`. also in `initial_conditions`.
- Separate discretised rhs/initial_conditions from concatenated rhs/initial_conditions (so that the above checks still work after discretisation)
- Remove `domain` from `disc.process_symbol` (now obsolete)

Fixes #82 

## Type of change

# Key checklist:

## Further checks: 


tinosulzer:
Merging and creating new issue #85

Issue #55, opened 2018-12-21
optimise adjacent matricies and vectors after discretisation


martinjrobins:
After the discretisation step, there is the possibility to optimise adjacent `Matrix`, `Vector` and `Scalar` nodes, like in #54. 

This might be best to do just once after the discretisation step, which would also solve #54.

martinjrobins:
fixed by #233

Issue #54, opened 2018-12-21
optimise adjacent scalars in tree after set parameters


martinjrobins:
After `ParameterValues` swaps out all the `Parameter` for `Scalar` nodes, there is the possibility that you get things like `Scalar(4) * Scalar(5)` in the tree, which can be combined into `Scalar(20)`

PR #94, opened 2019-01-18
Issue 80 concat var scalar

### LINKED ISSUE #80: allow concatenation of variables and scalars
From @tinosulzer  in #77:

It would useful to allow concatenation of variables and scalars, to deal with a case like the following:
Interfacial current G is some function in the positive and negative electrodes, and zero in the separator. In the model, this would probably be coded as

Gn = cn ** 2
Gs = Scalar(0)
Gp = cp ** 3
G = Concatenation(Gn, Gs, Gp)
Then G should get discretised to a concatenation of an expression tree with a StateVector, representing cn, for the negative electrode, a Vector of zeros with the right shape, and an expression tree StateVector, representing cp, for the positive electrode.

martinjrobins:
# Description

To allow concatenation of more than just Scalars, made a new concatenation class `DomainConcatenation` that concatenates vectors (evaluated by child symbols) according to their domains, respecting both the ordering in KNOWN_DOMAINS, and the number of points within each domain. Just like the base `Concatenation`, it is assumed that the domains of the children are all distinct

As part of this work, added a `is_constant` function to `Symbol` that tests if evaluating that symbol depends on either y or t (this will also help with #54 and #55)

Fixes #80 

## Type of change

# Key checklist:

## Further checks: 

### DIFF
--- a/pybamm/expression_tree/concatenations.py
+++ b/pybamm/expression_tree/concatenations.py
@@ -6,6 +6,7 @@ from __future__ import print_function, unicode_literals
 import pybamm
 
 import numpy as np
+import numbers
 
 
 class Concatenation(pybamm.Symbol):
@@ -23,7 +24,6 @@ class Concatenation(pybamm.Symbol):
     def __init__(self, *children, name=None):
         if name is None:
             name = "concatenation"
-
         domain = self.get_children_domains(children)
         super().__init__(name, children, domain=domain)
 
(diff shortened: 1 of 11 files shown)


pybamm/expression_tree/concatenations.py
```diff
@@ -73,6 +73,114 @@ def __init__(self, *children):
...
+        slices = {}
+        start = 0
+        end = 0
+        for dom in node.domain:
+            end += mesh[dom].npts
+            slices[dom] = slice(start, end)
+            start = end
+        return slices
+
+    def process_node_for_concantate(self, node, mesh):
```
tinosulzer:
```suggestion
    def process_node_for_concatenate(self, node, mesh):
```

pybamm/expression_tree/concatenations.py
```diff
@@ -73,6 +73,114 @@ def __init__(self, *children):
...
+        if self.domain == ["whole cell"]:
+            self.domain = ["negative electrode", "separator", "positive electrode"]
+
+        # create dict of domain => slice of final vector
+        self._slices = self.create_slices(self, mesh)
+
+        # store size of final vector
+        self._size = self._slices[self.domain[-1]].stop
+
+        # deal with "whole cell" special case
```
tinosulzer: duplicate?

martinjrobins: good catch, that was a copy and paste I forgot to change! should be fixed now

PR #98, opened 2019-01-25
Issue 91 average diffusivity

### LINKED ISSUE #91: Discretising "diffusivity times gradient" in Finite Volumes
Flux is of the form `N = -D(c) * grad(c)`

With the current implementation, after discretisation, `grad(c)` will evaluate to a vector of size (n-1,) (at cell edges) while `D(c)` (an arbitrary function of `c`, e.g. `c**2`) will evaluate to a vector of size (n,) (at cell centres), where (n,) is the size of c. So trying to multiply these two will give an error.

The Finite Volume discretisation needs to figure out that we have to take the average of `D(c)` to give `D_edges` which is a vector of size (n-1,) at cell edges. This could probably be done at the `BinaryOperator` stage of `disc.process_symbol`, i.e. if `right` is a `Gradient` object then take the average of `left`.

tinosulzer:
# Description

Discretise diffusivity appropriately so that shape matches with discretised gradient
Fixes #91 

## Type of change

# Key checklist:

## Further checks: 


pybamm/expression_tree/unary_operators.py
```diff
@@ -112,6 +112,48 @@ def __init__(self, child):
         super().__init__("div", child)
 
 
+class SpatialAverage(SpatialOperator):
+    """A node in the expression tree representing a spatially averaging operator.
```
martinjrobins: I think this needs more documentation. I would normally assume that a spatially averaging would take place over the entire vector, but here the averaging is taking place over adjacent elements only.

tinosulzer:
@martinjrobins tried a new naming/documentation for the spatial averaging of diffusivity, still not convinced this is very clear so am open to suggestions

martinjrobins:
I'm happy with this as is. Its a difficult one, since it is an operator that really only makes sense for Finite Volume. There might be a better way to represent this, but I can't see it for now.

PR #106, opened 2019-01-30
Issue 91 average diffusivity

### LINKED ISSUE #91: Discretising "diffusivity times gradient" in Finite Volumes

tinosulzer:
# Description

Fixes #91 
Have re-opened #91 to fix the case where there is both a `divergence` and a `gradient`:
```
Vector * Gradient(Variable) # should do node_to_edge
Vector * Divergence(Gradient(Variable)) # should not do node_to_edge
```
In doing so, realised there was an issue in `BaseDiscretisation.process_symbol` where, in the final `else`, detaching then reattaching `symbol.parent` switches around the order of `parent.children`:
```
# before disc:
symbol = Scalar * Vector 
symbol.children -> (Scalar, Vector)
# after disc 
disc.process_symbol(symbol)
symbol.children -> (Vector, Scalar)
```
This is a problem for recognising `symbol.id` in the boundary conditions

## Type of change

# Key checklist:

## Further checks: 

### DIFF
--- a/pybamm/discretisations/base_discretisation.py
+++ b/pybamm/discretisations/base_discretisation.py
@@ -200,6 +200,12 @@ class BaseDiscretisation(object):
 
             return new_symbol
 
+        elif isinstance(symbol, pybamm.Scalar):
+            return pybamm.Scalar(symbol.value, domain=symbol.domain)
+
+        elif isinstance(symbol, pybamm.Array):
+            return symbol.__class__(symbol.entries, domain=symbol.domain)
+
         else:
             # hack to copy the symbol but without a parent
             # (building tree from bottom up)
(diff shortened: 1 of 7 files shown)


pybamm/discretisations/base_discretisation.py
```diff
@@ -200,15 +199,22 @@ def process_symbol(self, symbol, y_slices=None, boundary_conditions={}):
 
             return new_symbol
 
+        elif isinstance(symbol, pybamm.Scalar):
+            return pybamm.Scalar(symbol.value, domain=symbol.domain)
+
+        elif isinstance(symbol, pybamm.Array):
+            return symbol.__class__(symbol.entries, domain=symbol.domain)
```
martinjrobins:
```suggestion
            return symbol.Array(symbol.entries, domain=symbol.domain)
```

martinjrobins: not sure what is better, but should be consistent with Scalar elif above

tinosulzer: `Array` can be either `Vector` or `Matrix`, would be better to return the same type I think

martinjrobins: ok, forgot about that. Ignore that suggestion then!

PR #126, opened 2019-02-10
Issue 125 model well-posedness

### LINKED ISSUE #125: Model well-posedness
When checking that a PDE model is well-posed, make sure that all variables that appear in the rhs dict values expression trees (i.e. the model equations) also appear in the rhs keys. 
For DAE models, give some "free passes" corresponding to the number of algebraic equations (for which we don't explicitly define a variable)

Most submodels won't be well-posed, but they don't need to be

tinosulzer:
# Description

Fixes #125 

## Type of change

# Key checklist:

## Further checks: 


tinosulzer:
This will be affected by the changes in #128

tests/test_models/test_base_model.py
```diff
@@ -156,7 +156,47 @@ def test_update(self):
...
+        model.rhs = {c: 5 * pybamm.div(pybamm.grad(d)) - 1}
+        model.algebraic = [e - c - d]
+        with self.assertRaises(pybamm.ModelError) as error:
+            model.check_well_posedness()
+            self.assertIn("underdetermined", error.exception.args[0])
+
+        # Underdetermined model - not enough algebraic equations
+        model.rhs = {c: 5 * pybamm.div(pybamm.grad(d)) - 1, d: -c - e}
+        model.algebraic = []
+        with self.assertRaises(pybamm.ModelError) as error:
```
martinjrobins: could also use assertRaisesRegex here, eg
```python
with self.assertRaisesRegex(pybamm.ModelError, 'underdetermined'):
   model.check_well_posedness())
```

tests/test_models/test_base_model.py
```diff
@@ -156,7 +156,47 @@ def test_update(self):
...
+            model.check_well_posedness()
+            self.assertIn("underdetermined", error.exception.args[0])
+
+        # Overdetermined model - too many algebraic equations
+        # Model cannot be overdetermined by too many differential equations, as rhs
+        # keys are all unique variables
+        model.algebraic = [c - d, e + d]
+        with self.assertRaises(pybamm.ModelError) as error:
+            model.check_well_posedness()
+        self.assertIn("overdetermined", error.exception.args[0])
```
martinjrobins: should be indented?, or see previous comment

tests/test_models/test_base_model.py
tinosulzer: cool, cheers, didn't know about assertRaisesRegex

PR #134, opened 2019-02-11
Issue 123 scikit instruct

### LINKED ISSUE #123: Add installation instructions for sundials and scikits.odes
- Running scripts/install_scikits_odes.sh
- Adding the relevant lines to `.bashrc` or `bin/activate` (if using a virtualenv) so that the sundials path is always set

martinjrobins:
# Description

Fixes #123 

## Type of change

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ which will save you needing to set your `LD_LIBRARY_PATH` every time you log in.
 example, to add this line to your `.bashrc` you can type:
 
 ```bash 
-$ echo 'export LD_LIBRARY_PATH=$INSTALL_DIR/lib:$LD_LIBRARY_PATH' >> ~/.bashrc
+$ echo "export LD_LIBRARY_PATH=$INSTALL_DIR/lib:\$LD_LIBRARY_PATH" >> ~/.bashrc
 ```
 
 Please see the [scikits.odes
--- a/README.md
+++ b/README.md
@@ -43,6 +43,47 @@ To uninstall again, type
 $ pip uninstall pybamm
 ```
 
+## Optional third-party libraries
+
+### [scikits.odes](https://github.com/bmcage/odes)
+
+Users can install [scikits.odes](https://github.com/bmcage/odes) in order to use the
+wrapped SUNDIALS ODE and DAE
+[solvers](https://pybamm.readthedocs.io/en/latest/source/solvers/scikits_solvers.html).
+To install scikits.odes, either follow the instructions below, or use the provided
+install script in `scripts/install_scikits_odes.sh`.
+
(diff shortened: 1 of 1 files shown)


martinjrobins:
Hi @fcooper8472. I don't think you have installed scikits.odes yet, so I'll let you see if these instructions are sufficient :)

README.md
```diff
@@ -43,6 +43,47 @@ To uninstall again, type
...
 
+## Optional dependencies
+
+### [scikits.odes](https://github.com/bmcage/odes)
+
+Users can install [scikits.odes](https://github.com/bmcage/odes) in order to use the
+wrapped SUNDIALS ODE and DAE
+[solvers](https://pybamm.readthedocs.io/en/latest/source/solvers/scikits_solvers.html).
+To install scikits.odes, either follow the instructions below, or use the provided
+install script in `scripts/install_scikits_odes.sh`.
```
Scottmar93: Not sure how detailed these instructions should be (or if these are intended as the instructions for others in Faraday) but I think it would be easier for others in Faraday if there was exact commands to enter e.g 
``` 
cd PyBaMM
source scripts/install_scikits_odes.sh
```
and then have the instruction to put the sundials path in venv/bin/activate or .bashrc here aswell. 

It wasn't obvious to me reading this whether adding the sundials path is part of the alternative installation or part of both.

martinjrobins: Ok, I'll make it clear that users should always add the sundials path LD_LIBRARY_PATH, whether they run the script or not.

I didn't really want users blindly running that script, as it does a pip install, the result of which is going to be rather different if they are in/out of a virtualenv. They might even prefer conda, depending on their setup. At least if they are copying one line at a time, they will see the pip install and adjust accordingly

martinjrobins: how is this @Scottmar93?  I removed the pointer to `scrips/install_scikits_odes.sh` as well, thinking about it more I think it would be a bad idea to get users to run this

Scottmar93: I think this looks great and easy to follow. I hadn't really thought about different setups etc so seems sensible to remove a pip specific script.

fcooper8472:
Looks good to me.  Two comments:

* The `$` at the start of each line makes it more difficult to copy and paste to the terminal.  It's marked up as a code block anyway - could they be removed?
* `echo 'export LD_LIBRARY_PATH=$INSTALL_DIR/lib:$LD_LIBRARY_PATH' >> ~/.bashrc` will literally put `$INSTALL_DIR` as a string and will not expand it to what the variable was set as.

martinjrobins:
fixed the bashrc line. All the bash blocks in pybamm have the `$`, as you say we probably don't need them, and would be better for copying/pasteing. I'll make another issue for this

PR #137, opened 2019-02-12
Issue 136 copy symbol

### LINKED ISSUE #136: Unpack a concatenation of variables (overwrite Symbol.__copy__ to remove parent)
## Problem

In some cases, we want to unpack a concatenation (usually of variables). The natural way would be 
```python3
# make concatenation
var_n = Variable("var", ["negative electrode"])
var_s = Variable("var", ["separator"])
var_p = Variable("var", ["positive electrode"])
var = Concatenation(var_n, var_s, var_p)
# unpack
newvar_n, newvar_s, newvar_p = var.children
```
However, manipulating the `newvar` objects then causes issues with the expression tree (tree internal data is corrupt)

## Suggested change

Write an `unpack_concatenation` function that unpacks a Concatenation object by creating new `Variable` objects with the same `name` and `domain` but no `parent` (and no `children` - `Variable` never has children anyway)

tinosulzer:
# Description

- Use `deepcopy` instead of `copy` when processing parameters, discretising and unpacking a Concatenation. This should help to avoid some TreeErrors to do with corrupt internal tree data. Still use copy when creating the expression tree (in `Symbol.__init__`) for speed/memory
- Simplify `ParameterValues.process_symbol` and `BaseDiscretisation.process_symbol` accordingly

Fixes #136 

## Type of change

# Key checklist:

## Further checks: 

### DIFF
--- a/pybamm/expression_tree/concatenations.py
+++ b/pybamm/expression_tree/concatenations.py
@@ -47,6 +47,15 @@ class Concatenation(pybamm.Symbol):
 
         return domain
 
+    @property
+    def children(self):
+        children_list = []
+        for child in tuple(self._NodeMixin__children_):
+            if isinstance(child, pybamm.Variable):
+                child = pybamm.Variable(child.name, child.domain)
+            children_list.append(child)
+        return children_list
+
 
 class NumpyModelConcatenation(pybamm.Symbol):
     """A node in the expression tree representing a concatenation of equations.
--- a/pybamm/expression_tree/concatenations.py
+++ b/pybamm/expression_tree/concatenations.py
@@ -5,7 +5,6 @@ from __future__ import absolute_import, division
 from __future__ import print_function, unicode_literals
 import pybamm
 
-import copy
 import numpy as np
 import numbers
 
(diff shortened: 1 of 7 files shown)


tinosulzer:
Tests passing locally but not in travis

pybamm/expression_tree/concatenations.py
```diff
@@ -47,6 +49,19 @@ def get_children_domains(self, children):
 
         return domain
 
+    @property
+    def children(self):
```
martinjrobins: I'm unsure about overloading this. Presumably there are, or will be methods in Symbol which use `children` assuming that it gives the children *with* parents. Perhaps it can be called `children_without_parents`, or `orphans` even.

tinosulzer: ok, makes sense - `orphans` is a good extension of syntax! In that case, it makes sense to put this function in `Symbol` itself rather than just `Concatenation`?

tinosulzer:
@martinjrobins , have put the `orphans` function in `Symbol`. Do you think the deepcopying is ok? Doesn't seem to be slowing things down for the current tests

martinjrobins:
It should be ok, in `process_symbol` it should only be called for leaf nodes, where there is not much to copy. Just have to remember that using `orphans` could be expensive. 

We should profile everything once we have some reasonable large models in pybamm to check.

PR #139, opened 2019-02-12
Issue 128 algebraic list to dict

### LINKED ISSUE #128: for algebraic equations: change from list to dict
dict key: `pybamm.Variable`
dict value: algebraic equation associated with that variable

tinosulzer:
# Description

Change algebraic list to dict and update model well-posedness tests
Fixes #128 

## Type of change

# Key checklist:

## Further checks: 

### DIFF
--- a/pybamm/models/base_model.py
+++ b/pybamm/models/base_model.py
@@ -20,7 +20,7 @@ class BaseModel(object):
     algebraic: dict
         A dictionary that maps expressions (variables) to expressions that represent
         the algebraic equations. The algebraic expressions are assumed to equate
-        to zero
+        to zero. Note that all the variables in the model must exist in the keys of `rhs` or `algebraic`.
     initial_conditions: dict
         A dictionary that maps expressions (variables) to expressions that represent
         the initial conditions for the state variables y
--- a/pybamm/models/base_model.py
+++ b/pybamm/models/base_model.py
@@ -37,7 +37,7 @@ class BaseModel(object):
     def __init__(self):
         # Initialise empty model
         self._rhs = {}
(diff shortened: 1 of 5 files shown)


pybamm/models/base_model.py
```diff
@@ -195,33 +197,46 @@ def update(self, *submodels):
...
-        - All the variables that appear in the rhs and algebraic equations appear in
-        the rhs keys, with leeway for exactly n variables where n is the number of
-        algebraic equations. Overdetermined if more equations than variables,
-        underdetermined if more variables than equations.
+        - Model is not over- or underdetermined, by comparing keys and equations in rhs
+        and algebraic. Overdetermined if more equations than variables, underdetermined
+        if more variables than equations.
         - There is an initial condition in self.initial_conditions for each
         variable/equation pair in self.rhs
         - There are appropriate boundary conditions in self.boundary_conditions for each
```
Scottmar93: We should also check that there are appropriate boundary conditions for the variable/equation pair in self.rhs

pybamm/models/base_model.py
```diff
@@ -18,7 +18,9 @@ class BaseModel(object):
         A dictionary that maps expressions (variables) to expressions that represent
         the rhs
     algebraic: dict
-        A list of algebraic expressions that are assumed to equate to zero
+        A dictionary that maps expressions (variables) to expressions that represent
+        the algebraic equations. The algebraic expressions are assumed to equate
+        to zero
```
martinjrobins:
```suggestion
        to zero. Note that all the variables in the model must exist in the keys of `rhs` or `algebraic`.
```

PR #140, opened 2019-02-12
Changed geometry keys to indepedent variables and added tests for each

### LINKED ISSUE #132: Convert Geometry Keys to Variables
Convert the keys in the Geometry class from strings (e.g. "x") to pybamm independent variables

Scottmar93:
The keys of the geometry dictionary are now independent variables. A set of tests have also been added to check that each key is an instance of `pybamm.IndependentVariable`

Fixes #132 

## Type of change

# Key checklist:

## Further checks: 

### DIFF
--- a/pybamm/discretisations/meshes.py
+++ b/pybamm/discretisations/meshes.py
@@ -7,7 +7,14 @@ from __future__ import print_function, unicode_literals
 import numpy as np
 import pybamm
 
-KNOWN_DOMAINS = ["negative electrode", "separator", "positive electrode", "test"]
+KNOWN_DOMAINS = [
+    "negative electrode",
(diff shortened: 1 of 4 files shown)


Scottmar93:
don't merge this yet as I didn't add tests correctly

Scottmar93:
> Cool, would just use `Space` instead of `IndependentVariable` (this would also require changing the `Space` class to take inputs `x`, `y` or `z`). Maybe also clearer to rename `Space` to `SpatialVariable`?

Can you add this as a separate issue? We discussed that this would have some awkward extra details. I agree that SpatialVariable is better than Space

tinosulzer:
Opened #145 , merging

PR #147, opened 2019-02-14
Issue 68 function node

### LINKED ISSUE #68: add a `Function` node to expression tree
for `Parameters` that vary with time, user puts in the name of the function (functions defined in <name>.py in the same directory as the parameter file). Then the parameter class replaces `Parameter` with `Function` of that name

fcooper8472:
# Description

#68 Implements `Function` and `FuntionParamter` as well as machinery for dynamically loading a function given string representing the filename.

Fixes #68

## Type of change

# Key checklist:

## Further checks: 


input/parameters/lithium-ion/example_function.py
```diff
@@ -0,0 +1,11 @@
...
+# Example function, for testing
+#
+from __future__ import absolute_import, division
+from __future__ import print_function, unicode_literals
+
+import numpy as np
+
+
+def example_function(x):
+    return x * np.exp(x - 1)
```
tinosulzer: I guess we don't need this file any more?

pybamm/expression_tree/unary_operators.py
```diff
@@ -67,6 +67,22 @@ def evaluate(self, t=None, y=None):
         return np.abs(self.children[0].evaluate(t, y))
 
 
+class Function(UnaryOperator):
+    """A node in the expression tree representing an arbitrary function
+
+    **Extends:** :class:`UnaryOperator`
+    """
+
+    def __init__(self, child, func):
```
tinosulzer: I see your point in the issue about the order of arguments - would switch this one around then, so that we call `FunctionParameter` and `Function` the same way

input/parameters/lithium-ion/example_function.py
fcooper8472: Good point!

pybamm/expression_tree/unary_operators.py
fcooper8472: OK - changed that in the new commit

Issue #40, opened 2018-12-20
visualise simulation results


tinosulzer:
Add class(es)/function(s) to visualise the result of a simulation. Options:
- time vs variable plot for variables that are only functions of time (current, voltage, temperature, etc)
- Slider plot (space vs variable at a time definable by a slider) for variables that depend on space and time (concentration, etc)
- Export for plotting using external software (e.g. pgfplots, Paraview)
- others?

martinjrobins:
should take in a `y` and `t`, plus an expression to plot (or `model.variables`) and do required plot

PR #148, opened 2019-02-14
Issue 39 simulation class

### LINKED ISSUE #39: refactor Simulation class
Refactor `Simulation` class to run the whole process (create model, set parameters, discretise, solve, visualise). Needs to take optional inputs for the parameter class, parameter file, mesh/discretisation, solver, and set defaults if these are `None`. Can later add capability to save solution, load a saved solution, etc.
```python3
class Simulation(object):
   def __init__(
        self,
        model,
        parameter_values=None,
        discretisation=None,
        solver=None,
        name="unnamed",
    ):
        # Defaults
        if parameter_values is None:
           # default parameter_values
        if discretisation is None:
           # default discretisation
        if solver is None:
            # default solver

        self.model = model
        self.parameter_values = parameter_values
        self.discretisation = discretisation
        self.solver = solver
        self.name = name

    def set_parameters(self):
        self.parameter_values.process(self.model)

    def discretise(self):
        y0 = self.discretisation.process(self.model)
        return y0

    def solve(self, y0):
        self.solver.solve(self.model, y0)

    def run(self):
        self.set_parameters()
        y0 = self.discretise()
        self.solve(y0)

    def load(self):

    def save(self):

    def plot(self):

```

tinosulzer:
# Description

Not convinced I've done this the best way but it might help with #40 
Fixes #39 

## Type of change

# Key checklist:

## Further checks: 


README.md
```diff
@@ -10,6 +10,14 @@ Python Battery Mathematical Modelling solves continuum models for batteries, usi
 
 ## How do I use PyBaMM?
 
+The easiest way to use PyBaMM is to run a constant-current discharge with a model of your choice:
+```
+$ python3 pybamm/simulation.py model_name
+```
+where model_name is the name of one of the [model classes](pybamm/models/). 
+
```
martinjrobins: might want to also say something about how users can customise simulations using the base mode/parameter/discretisation/solver classes here

pybamm/simulation.py
```diff
@@ -16,84 +16,109 @@ class Simulation(object):
...
-        self.vars.average()
-
-    @property
-    def filename_io(self):
-        path = os.path.join("out", "simulations")
-        filename = "model_{}_solver_{}".format(
-            pickle.dumps(self.model), pickle.dumps(self.solver)
-        )
-        return os.path.join(path, filename)
+    def set_parameters(self):
```
martinjrobins: maybe mark these as private functions i.e. `_set_parameters(self)`? I would expect that only people new to pybamm would use this class, and you don't want them calling:
```python
simulation.discretise()
simulation.set_parameters()
simulation.solve()
```

pybamm/simulation.py
```diff
@@ -16,84 +16,109 @@ class Simulation(object):
...
 
     def save(self):
-        """Save simulation."""
         raise NotImplementedError
 
     def plot(self):
         raise NotImplementedError
+
+
+# QUESTION: is there a way to test this?
```
martinjrobins: you could use `result = os.system("python simulation.py")` and check the result? Not sure that would work with the coverage, but it would at least run it

pybamm/simulation.py
```diff
@@ -16,84 +16,109 @@ class Simulation(object):
...
+    # parser.add_argument("-s", "--save", action="store_true", help="save the output")
+    # parser.add_argument(
+    #     "-f",
+    #     "--force",
+    #     action="store_true",
+    #     help="overwrite saved output even if it is available",
+    # )
+    args = parser.parse_args()
+
+    # QUESTION: there must be a cleaner way to do this?
```
martinjrobins: You could also use the `dir` function to see the attributes of a model, but this looks fine.

tinosulzer:
Ok, I will have a go at doing this differently. I think you're right that as an "introduction to PyBaMM" it would work better as a notebook, but it would still be useful for everyday usage to be able to
- Run and plot a model from command line
- Save and load simulation outputs (e.g. when it comes to making plots and having to run several times just to get the axis labels etc to look right, although maybe that is a nooby way of doing it)

martinjrobins:
You might be right. I don't tend to run simulations/plotting directly from the command line, but if you think its helpful I'm happy to go with it. I usually have a script or notebook per project that does everything, interspersed with pickle load/stores so that everything is cached and I don't have to, for example, run the simulation every time I change the plot labels.

Scottmar93:
This class is great for quickly comparing different models but I think Martins point about it hiding alot of stuff away is valid. Maybe we could mostly interact through entering dicts in a script and change this `Simulation` object to `QuickSimulation`. We could set it up so that users can create their own subclasses of QuickSim with sets of defaults specific to their problem. This then allow quick comparisons and plotting etc of models beyond the standard set.

Issue #110, opened 2019-02-04
smarter domain concatenation and broadcast


martinjrobins:
At the moment, the DomainConcatentation class will convert any Scalar children to vectors. This should be generalised, so that if the `evaluate` of its children return a scalar, this is appropriately mapped to a vector at run-time (i.e. for time dependent functions)

tinosulzer:
This issue should:
- unskip the test from #149 
- remove the warning from #124 
- implement #155 
- test that concatenation of 2D matrices works as expected, not just 1D vectors (e.g. the matrix representing c as a function of space and time)

martinjrobins:
discussed this in meeting, see result here: https://github.com/tinosulzer/PyBaMM/pull/163#issuecomment-468256714

PR #150, opened 2019-02-14
Issue 149 interfacial reactions

### LINKED ISSUE #149: Interfacial reactions
Sub-issue from #120 for the Butler-Volmer interfacial reaction functions

tinosulzer:
# Description

Add functions for the interfacial reactions. Skipping one test that is reliant on concatenating `StateVector` objects (#110), but I will unskip this test when doing #110 
Fixes #149 

## Type of change

# Key checklist:

## Further checks: 


pybamm/parameters/standard_parameters_lead_acid.py
```diff
@@ -91,17 +91,35 @@
...
-# Functions
-D_dim = pybamm.Parameter("epsn_max")
+# --------------------------------------------------------------------------------------
+"""Functions"""
+
 rho_dim = pybamm.Parameter("epsn_max")
 mu_dim = pybamm.Parameter("epsn_max")
 
+
+def D_dim(c):
```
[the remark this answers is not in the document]
Scottmar93: I can understand the meaning here but I think we should have a comment for every function and parameter describing what they are for convenience.

The models should all also be written up in the documentation for even more detail.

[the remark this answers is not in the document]
Scottmar93: We could maybe focus on documentation and commenting of models once we get the set we want to implement up and running

pybamm/parameters/standard_parameters_lead_acid.py
```diff
@@ -191,3 +210,16 @@
 epsn_init = epsn_max - epsDelta_n * (1 - q_init)  # Initial pororsity (neg) [-]
 epss_init = epss_max  # Initial pororsity (sep) [-]
 epsp_init = epsp_max - epsDelta_p * (1 - q_init)  # Initial pororsity (pos) [-]
+
+
+# Concatenated symbols
+# artificially add domains - doesn't seem like the best way of doing this
+# TODO: look into how to do this more elegantly
+neg = pybamm.Scalar(1, domain=["negative electrode"])
```
Scottmar93: could just add a piecewise constant function that takes in a list of values and a list of domains and outputs the concatenated scalar object

pybamm/models/submodels/interface.py
```diff
@@ -19,3 +19,120 @@ def homogeneous_reaction():
...
+
+    Returns
+    -------
+    :class:`pybamm.Symbol`
+        The interfacial current density in the appropriate domain
+    """
+    if domain is None:
+        # default is for domain to be across the whole cell
+        # (if not specified by c.domain or phi.domain)
+        if c.domain == [] and phi.domain == []:
```
Scottmar93: I'm thinking in terms of the SPM here so may not be applicable. If the domain=[] then that means that c should be independent of space. But doesn't this returns an interfacial current that will eventually be converted into a vector discretised on the whole_cell?

tests/test_models/test_interface.py
```diff
@@ -66,6 +69,218 @@ def test_discretisation(self):
...
+            "potential difference", domain=["negative electrode"]
+        )
+        self.phis = pybamm.Variable("potential difference", domain=["separator"])
+        self.phip = pybamm.Variable(
+            "potential difference", domain=["positive electrode"]
+        )
+        self.phi = pybamm.Concatenation(self.phin, self.phis, self.phip)
+
+    def tearDown(self):
+        del self.cn
```
Scottmar93: what is the reason for doing the teardown test? should I also add this when making tests?

tinosulzer: I actually have no idea as it seems to work fine without - think this is a question for the C++ experts @martinjrobins @fcooper8472

pybamm/models/submodels/interface.py
tinosulzer: If `c.domain == [] and `phi.domain==[]`, and `domain is None` then we have to pick whether to calculate Butler Volmer in the negative or positive electrode. I think this just makes things confusing though so will raise an error instead if no domain is given by either the string `domain` or `c` or `phi`

PR #152, opened 2019-02-15
Issue 129 enable per equation discretisation

### LINKED ISSUE #129: enable per-equation discretiation
depends on #128 

dict of {variable: discretisation} passed as input to base discretisation class. 

The discretiation class passed in the dict handles things like gradient/divergence discretisation + broadcasting

Scottmar93:
# Description
A per **domain** implementation of the spatial methods. This is instead of per equation because of some issues described below. 

This can be made sufficient for different finite elements by adding a polynomial type property to the variables we are solving for. It won't allow for finite volumes and finite elements both on the macroscale but we don't foresee that being something which is needed. 

Fixes #129 

# Details
I have implemented a per domain spatial method assignment. The main reason to do this instead of per equation is that with the per equation method it is not clear how which spatial method to use to discretise the model.variables dictionary which is a dict of useful variables for plotting etc. 

If we figure out a way to implement the model.variables dictionary differently then changing the per domain setup to per equation is quite a small change which only requires a few keys in the discretisation.py file to be edited along with the corresponding tests. 

## Type of change

# Key checklist:

## Further checks: 


pybamm/models/base_model.py
```diff
@@ -67,8 +68,15 @@ def __init__(self):
...
+        self.default_spatial_methods = {
+            "negative electrode": pybamm.FiniteVolume,
+            "separator": pybamm.FiniteVolume,
+            "positive electrode": pybamm.FiniteVolume,
+        }
         self.mesh = pybamm.Mesh(self.default_geometry, submesh_types, submesh_pts)
-        self.default_discretisation = pybamm.FiniteVolumeDiscretisation(self.mesh)
+        self.default_discretisation = pybamm.Discretisation(
+            self.mesh, self.default_spatial_methods
+        )
```
tinosulzer: It could be left to external classes, such as `StandardModelTest`, to call `Mesh(self.default_geometry, self.default_submesh_types, self.default_submesh_pts)` and `Discretisation(mesh, self.default_spatial_methods)` (given that there is now only one `Mesh` class and only one `Discretisation` class).
Otherwise, in every model, if you want to change the default submesh types then you also have to go through and redefine `self.mesh` and then redefine `self.default_discretisation`. 
Similar comment also applies to the line `self.default_parameter_values.process_geometry(self.default_geometry)`

pybamm/spatial_methods/finite_volume.py
```diff
@@ -155,33 +120,33 @@ def gradient_matrix(self, domain):
...
         # Add Neumann boundary conditions if defined
         if symbol.id in boundary_conditions:
             # for the particles there will be a "negative particle" "left" and "right"
             # and also a "positive particle" left and right.
             lbc = boundary_conditions[symbol.id]["left"]
             rbc = boundary_conditions[symbol.id]["right"]
-            discretised_symbol = self.concatenate(lbc, discretised_symbol, rbc)
+            discretised_symbol = pybamm.NumpyModelConcatenation(
+                lbc, discretised_symbol, rbc
+            )
```
tinosulzer: This should be a `DomainConcatenation` (sorry, realise it's a mistake brought forward from previous implementation)

pybamm/spatial_methods/finite_volume.py
```diff
@@ -155,33 +120,33 @@ def gradient_matrix(self, domain):
...
         domain = symbol.domain
-        # check for
+        # check for spherical domains
         if ("negative particle" or "positive particle") in domain:
 
             # implement spherical operator
             divergence_matrix = self.divergence_matrix(domain)
 
-            submesh = self.mesh.combine_submeshes(*domain)
+            submesh = self._mesh.combine_submeshes(*domain)
```
tinosulzer: Would you ever have subdomains within the particles? I don't think you'd ever need to combine submeshes, unlike on the macroscale

tests/shared.py
```diff
@@ -52,25 +58,48 @@ def __init__(self, n):
 
         self.mesh = pybamm.Mesh(self.geometry, self.submesh_types, self.submesh_pts)
 
+        self.spatial_methods = {"negative particle": pybamm.FiniteVolume}
```
tinosulzer:
```suggestion
        self.spatial_methods = {"negative particle": pybamm.SpatialMethodForTesting}
```

pybamm/expression_tree/unary_operators.py
```diff
@@ -172,19 +172,19 @@ class NumpyBroadcast(Broadcast):
...
 
-    def __init__(self, child, domain, mesh):
+    def __init__(self, child, domain, npts):
         super().__init__(child, domain, name="numpy broadcast")
         # determine broadcasting vector size (size 1 if the domain is empty)
         if domain == []:
             self.broadcasting_vector_size = 1
         else:
-            self.broadcasting_vector_size = sum([mesh[dom].npts for dom in domain])
+            self.broadcasting_vector_size = sum([npts[dom] for dom in domain])
```
tinosulzer: This seems like an unnecessary change that obscures what the class is doing, and requires you to artificially create the dictionary `npts` before calling `NumpyBroadcast`. I think that the `npts` attribute of a submesh should always be defined in such a way that `mesh[dom].npts` is what you want to use for broadcasting

pybamm/models/base_model.py
```diff
@@ -54,6 +54,7 @@ def __init__(self):
         )
 
         self.default_geometry = pybamm.Geometry1DMacro()
+
         self.default_parameter_values.process_geometry(self.default_geometry)
```
tinosulzer: See comment below

pybamm/models/base_model.py
Scottmar93: Yeah probably should reformulate all this so that just call `tests.shared.testing_disc` and tests.shared.finite_volume_1D_macro_disc` etc. I'll add a new issue to do this as it's a bit of a pain.

tests/shared.py
Scottmar93: Making this small change will break the spherical tests later. This will be dealt with properly in #153 so leaving for now.

pybamm/expression_tree/unary_operators.py
Scottmar93: The choice is to:
1) change `SubMesh` so that we add  .npts to each SubMesh within `SpatialMethod` and then pass the new mesh
or 
2) create a dict of npts in each domain and change broadcast to take in npts dict 

I didn't want to change the `SubMesh` within a spatial method so that the same property names refer to slightly different things for different spatial methods. It seems clearer to me to have a submesh.cells and a submesh.edges property rather than renaming one of them submesh.nodes within the a spatial method. Likewise, having submesh.n_cells and a submesh.n_edges instead of submesh.npts.

tinosulzer: ok so rename `nodes` to `cells` and `npts` to `n_cells` and then use `mesh[dom].n_cells`?

pybamm/expression_tree/symbol.py
```diff
@@ -37,6 +37,7 @@ def __init__(self, name, children=[], domain=[]):
             # this also adds copy.copy(child) to self.children
             copy.copy(child).parent = self
         self.domain = domain
+        self.spatial_method = None
```
tinosulzer:
```suggestion
```

PR #156, opened 2019-02-15
Issue 124 current input

### LINKED ISSUE #124: Allow custom current input into models
## Summary

Most battery models are "current-controlled", i.e. you feed in a current (function of time) and read the resulting voltage.

We should have functionality to allow the user to input their own current function into the model. 

Requires #68 

## Suggested change

1. Pass `current_scale` and `current_function` as parameters to the model (`current_scale` is needed as it is used in the non-dimensionalisation)
2. In `standard_parameters.py`, define `current` as a function that takes in `current_scale` and `current_function` and returns a `pybamm.Function` object:
```python3
def current(current_scale, current_function):
    t = pybamm.t
    return pybamm.Function(t, current_scale*current_function)
```
3. In the model class, call `current` as:
```python3
def __init__(self, current_scale, current_function):
    ...
    current = pybamm.standard_parameters(current_scale, current_function)
    ...
```
We could also have a `parameters.standard_current_functions.py` so that the user can choose from some pre-defined current functions, including interpolation of time series (e.g. from experimental data)

tinosulzer:
# Description

Include a current input in the model. Fixes #124
Had to make slight changes to `DomainConcatenation` to allow processing of the `Time` variable (which isn't constant but we want to broadcast anyway), again will need to do this more robustly in #110 
Also added functionality to process a `Function` object (just return a few `Function` with processed child). 

## Type of change

# Key checklist:

## Further checks: 


pybamm/expression_tree/concatenations.py
```diff
@@ -144,25 +144,22 @@ def process_node_for_concatenate(self, node, mesh):
...
-
-        # convert to a Vector node
-        return pybamm.Vector(value, domain=node.domain)
+        try:
+            node_size = node.size
+        except AttributeError:
+            node_size = 0
+
+        if node_size > 1:
+            if node_size == sum([mesh[dom].npts for dom in node.domain]):
```
Scottmar93: I guess this follows on from how we name the components of the mesh but should we be passing the mesh to this function if it doesn't belong to a specific SpatialMethod Class?

Scottmar93:
I tried a git merge master for this branch and it still passes tests

pybamm/expression_tree/concatenations.py
```diff
@@ -179,23 +176,29 @@ def evaluate(self, t=None, y=None):
         return vector
 
 
-def piecewise_constant(neg_value, sep_value, pos_value):
-    """Concatenate three values that don't have a domain into a piecewise constant
-    concatenation. This is useful when we don't want to assign a domain to the inputs
+class PiecewiseConstant(Concatenation):
```
martinjrobins: Why do we need this class? Why wouldn't the children have domains?

pybamm/parameters/standard_parameters_lead_acid.py
```diff
@@ -91,9 +91,37 @@
...
+interfacial_current_scale_pos = ibar / (Ap * Lx)
+velocity_scale = ibar / (cmax * F)  # Reaction velocity scale
+
 # --------------------------------------------------------------------------------------
 """Functions"""
 
+
+def dimensional_current(t):
+    """Returns the dimensional current as a function of time """
+    t = pybamm.t
```
martinjrobins: Should this function take no arguments, or be a variable (since you are simply overwriting this argument anyway)?

pybamm/parameters/standard_parameters_lead_acid.py
```diff
@@ -91,9 +91,37 @@
...
+    return ibar * pybamm.FunctionParameter("current function", t)
+
+
+def dimensionless_current(t):
+    """Returns the dimensionless current as a function of time """
+    return pybamm.FunctionParameter("current function", t)
```
martinjrobins: will the input for this function always by `pybamm.t`?

pybamm/expression_tree/concatenations.py
tinosulzer: One example is the porosities `epsn`, `epss`, `epsp` and reaction rates `sn` and `sp` - we don't want them to have domains so that we can [add them together in the LOQS](https://github.com/tinosulzer/PyBaMM/blob/9dd06cc356d2f495b188ef9a81c1b3ecb5e7f830/pybamm/models/lead_acid/loqs.py#L65) without throwing domain errors, but we do want to plot the porosity as a function of x (in [this line](https://github.com/tinosulzer/PyBaMM/blob/9dd06cc356d2f495b188ef9a81c1b3ecb5e7f830/pybamm/models/lead_acid/loqs.py#L85) we would replace`Concatenation` by `PiecewiseConstant`) or have a piecewise constant reaction rate in the submodels (e.g. replace [this line](https://github.com/tinosulzer/PyBaMM/blob/7a1e87f3e5c04b91e016509b1839d86cbd6ef12c/pybamm/models/submodels/electrolyte_concentration.py#L76) by `PiecewiseConstant(sn, 0, sp)`. 

There is definitely some work to do to make Broadcasts and Concatenations clearer though so I am open to superior alternatives :)

pybamm/parameters/standard_parameters_lead_acid.py
tinosulzer: Oops, yeah, input will always be `pybamm.t` as far as I can think of so don't know whether it's clearer to make that explicit in the model (so this function takes a parameter) or leave it implicit

pybamm/parameters/standard_parameters_lead_acid.py
tinosulzer: Oops, yeah, input will always be `pybamm.t` as far as I can think of so don't know whether it's clearer to make that explicit in the model (so this function takes a parameter) or leave it implicit

pybamm/expression_tree/concatenations.py
martinjrobins: ok, lets discuss at the Thurs meeting then. Perhaps things like `epsn`, `epss`, `epsp` can be domain-free, and if a model wants to create a piecewise constant vector over a given set of domains it could write something like:

```python
eps = pybamm.Concatenation(pybamm.Broadcast(epsn,"domain 1"), 
                           pybamm.Broadcast(epss,"domain 2"), 
                           pybamm.Broadcast(epsp,"domain 3"))
```

pybamm/parameters/standard_parameters_lead_acid.py
martinjrobins: being a function implies that `dimensionless_current` can be a function of *not* t, which contradicts the comment. 

I would use a variable and make it clear in the name:

```python
current_with_t = pybamm.FunctionParameter("current function", pybamm.t)
dim_current_with_t = return ibar * current_with_time
```

pybamm/expression_tree/concatenations.py
tinosulzer: Could we merge this issue and discuss concatenations in the context of #163 ? Just so that we can start using the function-processing part of this issue

PR #157, opened 2019-02-16
Issue 145 spatial variables

### LINKED ISSUE #145: Make Spatial Variables class
Class should replace the current `Space` class and allow input of `x`, `y`, `z` or `r` for different spatial variables. Will the need to update 
- Geometry classes to take `SpatialVariable` instead of `IndependentVariable`
- Discretisation of `Space` to discretise the specific variable given (return Vector on the appropriate mesh for `x` and `r`, raise `NotImplementedError` for `y` and `z` to be returned to later when implementing 2D/3D)

tinosulzer:
# Description

Creates and discretises SpatialVariable class
Fixes #145 

## Type of change

# Key checklist:

## Further checks: 


tinosulzer:
Good spot, fixed

Issue #155, opened 2019-02-15
Change NumpyModelConcatenation to DomainConcatenation in FiniteVolume


Scottmar93:
As title says. This will cause errors though so need to figure out why.

tinosulzer:
Having dug into this, maybe `DomainConcatenation` isn't the most appropriate function to use here. 
`DomainConcatenation` assumes that each symbol in the list of children being passed to it has a non-empty domain, which isn't the case in `FiniteVolume.divergence` as the boundary conditions have empty domain (they are just scalars)

Scottmar93:
This issue is solved if we remove DomainConcatenation as you suggest in the discussion in #163. DomainConcatenation is very specific to the macroscale domains which I think we should avoid.

martinjrobins:
discussed this in meeting, see result here: https://github.com/tinosulzer/PyBaMM/pull/163#issuecomment-468256714

tinosulzer:
Closed by #163

PR #163, opened 2019-02-20
Issue 110 domain concatenation

### LINKED ISSUE #110: smarter domain concatenation and broadcast

tinosulzer:
# Description

**To be reviewed after #156 **
Simplifies the `DomainConcatenation` and allows it to work with a range of classes. 
- [x] Allow concatenation of StateVector objects 
- [x] Allow concatenation of 2D objects (e.g. solutions of the model in different domains) - a little bit hacky as currently written, with lots of if statements

Fixes #110 and #155 

## Type of change

# Key checklist:

## Further checks: 


pybamm/discretisations/discretisation.py
```diff
@@ -155,18 +155,13 @@ def process_dict(self, var_eqn_dict):
...
-                    eqn = pybamm.NumpyBroadcast(eqn, eqn_key.domain, {})
-                else:
-                    eqn = self._spatial_methods[eqn_key.domain[0]].broadcast(
-                        eqn, eqn_key.domain
-                    )
+                eqn = pybamm.Broadcast(eqn, var.domain)
 
             # Process symbol (original or broadcasted)
-            var_eqn_dict[eqn_key] = self.process_symbol(eqn)
+            var_eqn_dict[var] = self.process_symbol(eqn)
```
Scottmar93: might be just because pull request was made before previous merges but should this still read eqn.id?

pybamm/expression_tree/concatenations.py
```diff
@@ -48,12 +48,13 @@ def get_children_domains(self, children):
         return domain
 
 
-class NumpyModelConcatenation(pybamm.Symbol):
-    """A node in the expression tree representing a concatenation of equations.
+class NumpyConcatenation(pybamm.Symbol):
+    """A node in the expression tree representing a concatenation of equations, when we
+    *don't* care about domains. The class :class:`pybamm.DomainConcatenation`, which
```
Scottmar93: For me it isn't clear exactly what we mean when we say *don't* care about domains. I'm probably missing some of the other functionality here but I can't think of when we wouldn't pass variables, or parameters without domains into a concatenation

Scottmar93: Or is this only used to create one big long rhs vector?

pybamm/expression_tree/concatenations.py
```diff
@@ -80,7 +81,8 @@ def evaluate(self, t=None, y=None):
 
 
 class DomainConcatenation(Concatenation):
-    """A node in the expression tree representing a concatenation of symbols.
+    """A node in the expression tree representing a concatenation of symbols, being
+    careful about domains.
```
Scottmar93: DomainConcatenation is only for combining variables on subdomains to make a new variable on a combined subdomain? (e.g. the electrolyte)

pybamm/expression_tree/concatenations.py
Scottmar93: Reading through more, NumpyConcatenation seems like quite a different thing to the other concatenations. Maybe we should name it differently to avoid confusion?

tinosulzer:
I think it might be clearest to get rid of `DomainConcatenation` and always use `NumpyConcatenation` instead. We would need always either broadcast explicitly or use `PiecewiseConstant` if we want to broadcast in all three domains, which will be the main use case. Disadvantage is that we lose the re-ordering of domains from `self._slices` and `self._children_slices`, but I think it is worth it to avoid having lots of classes that do similar things.

Scottmar93:
> I think it might be clearest to get rid of `DomainConcatenation` and always use `NumpyConcatenation` instead.

I agree.

tinosulzer:
- Get rid of PiecewiseConstant and do Broadcast explicitly
- Don't expect DomainConcatenation to do any broadcasting

pybamm/expression_tree/concatenations.py
```diff
@@ -133,72 +131,32 @@ def create_slices(self, node, mesh):
...
-            npts = {dom: submesh.npts for dom, submesh in mesh.items()}
-            return pybamm.NumpyBroadcast(node, node.domain, npts)
-
     def evaluate(self, t=None, y=None):
         """ See :meth:`pybamm.Symbol.evaluate()`. """
-
+        # return np.concatenate([child.evaluate(t, y) for child in self.children])
         # preallocate vector
-        vector = np.empty(self._size)
+        if t is None or isinstance(t, numbers.Number):
```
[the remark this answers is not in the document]
tinosulzer: Ah ok, you're right, there is actually never any need to work with `y` as a 2D matrix. This has been a headache to implement and would probably be a headache to maintain so very happy to get rid of it! 
I was worried that using a for loop might be slow, but now realise that this will never need to happen during the time stepping (by definition) so shouldn't be an issue

PR #166, opened 2019-02-21
Updated the update model function so that it also works for algebraic equations

Scottmar93:

# Key checklist:

## Further checks: 

### DIFF
--- a/pybamm/models/base_model.py
+++ b/pybamm/models/base_model.py
@@ -202,8 +202,8 @@ class BaseModel(object):
 
     def check_and_combine_dict(self, dict1, dict2):
         # check that the key ids are distinct
-        ids1 = set(map(lambda x: x.id, list(dict1.keys())))
-        ids2 = set(map(lambda x: x.id, list(dict2.keys())))
+        ids1 = set(x.id for x in dict1.keys())
+        ids2 = set(x.id for x in dict2.keys())
         assert len(ids1.intersection(ids2)) == 0, pybamm.ModelError(
             "Submodel incompatible: duplicate variables"
         )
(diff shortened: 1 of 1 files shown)


pybamm/models/base_model.py
```diff
@@ -188,17 +188,26 @@ def update(self, *submodels):
...
+                self._initial_conditions, submodel.initial_conditions
+            )
+            self.check_and_combine_dict(
+                self._boundary_conditions, submodel.boundary_conditions
+            )
+            self._variables.update(submodel.variables)  # keys are strings so no check
+
+    def check_and_combine_dict(self, dict1, dict2):
+        # check that the key ids are distinct
+        ids1 = set(map(lambda x: x.id, list(dict1.keys())))
```
tinosulzer: I think the below would be slightly more legible
```python3
ids1 = set([x.id for x in dict1.keys()])
```

pybamm/models/base_model.py
```diff
@@ -188,17 +188,26 @@ def update(self, *submodels):
...
+            # check and then update dicts
+            self.check_and_combine_dict(self._rhs, submodel.rhs)
+            self.check_and_combine_dict(self._algebraic, submodel.algebraic)
+            self.check_and_combine_dict(
+                self._initial_conditions, submodel.initial_conditions
+            )
+            self.check_and_combine_dict(
+                self._boundary_conditions, submodel.boundary_conditions
+            )
+            self._variables.update(submodel.variables)  # keys are strings so no check
```
tinosulzer: Could check that strings aren't being repeated (and hence overwritten) but not crucial

Issue #153, opened 2019-02-15
Move all test statments of the discretisation in to tests.shared


Scottmar93:
Reformulate tests in test_discretisation and test_finite_volume to call `tests.shared.testing_disc` and tests.shared.finite_volume_1D_macro_disc` etc. instead of defining the disc from the defaults

tinosulzer:
It would be easier to be able to call both micro and micro discretisations from the same function in `tests.shared`

PR #167, opened 2019-02-21
Issue 72 implement spm

### LINKED ISSUE #72: Implement Single Particle Model (SPM)
**Summary**
Implement the SPM. We require: 
-  a `StandardParticle` class which implements spherical diffusion on a spherically symmetric particle taking a boundary condition `G` for the flux on the particle surface. 
- an `SPM` class which calls `homogeneous_reactions` to create `G` and then calls `StandardParticle` twice to create a negative electrode particle and a positive electrode particle. 
- The SPM class should also contain an expression for the voltage in terms of the other fundamental variables (which are just c_n and c_p in this case). This will be contained within the `variables` dictionary of the model

rtimms:
# Description

Implemented a simple SPM for constant current discharge. Will probably need to be updated later to be built from two Particle submodels. For now the Butler-Volmer equation is linearised.

Fixes #72 

Also changed `SurfaceValue` so that it had no domain. This allows you to make expressions which contain surface values of variables from different domains.

# Key checklist:

## Further checks: 


pybamm/models/li_ion/single_particle_model.py
```diff
@@ -0,0 +1,116 @@
...
+        A dictionary that maps strings to expressions that represent
+        the useful variables
+
+    """
+
+    def __init__(self):
+        super().__init__()
+
+        # Overwrite default geometry
+        # NOTE: Is this the best way/place to do this?
```
tinosulzer: no but currently only way to do it - will deal with it in #153

Issue #164, opened 2019-02-20
Spatial methods macroscale


tinosulzer:
User should only be able to input a single spatial method for the macroscale, rather than mix and match. The discretisation class should then read `spatial_method["macroscale"]` and assign that to each of the domains:
```python3
# in Discretisation.__init__
if "macroscale" in spatial_method.keys():
    method = spatial_method["macroscale"]
    spatial_method["negative electrode"] = method
    spatial_method["separator"] = method
    spatial_method["positive electrode"] = method
self._spatial_method = spatial_method
```

This can probably be done at the same time as #153

tinosulzer:
Closed by #172

PR #172, opened 2019-02-22
Issue 153 test statements

### LINKED ISSUE #153: Move all test statments of the discretisation in to tests.shared

tinosulzer:
# Description

- Merge macroscale and microscale in `tests.shared`.
- Decouple the defaults in the models (means we have to do more work outside the model, but I think this is worth it)
- Change spatial_methods dict to take a "macroscale" input

Fixes #153 and #164

## Type of change

# Key checklist:

## Further checks: 

### DIFF
--- a/pybamm/geometry/geometry.py
+++ b/pybamm/geometry/geometry.py
@@ -18,8 +18,20 @@ class Geometry(dict):
      custom_geometry : dict containing any extra user defined geometry
      """
 
-    # left empty for now but I think we should have methods to
-    # add Geometries together (i.e. choose 3D macro with 1D micro etc)
+    def __init__(self, *geometries, custom_geometry={}):
+        for geometry in geometries:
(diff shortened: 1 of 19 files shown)


pybamm/geometry/geometry.py
```diff
@@ -18,8 +18,20 @@ class Geometry(dict):
      custom_geometry : dict containing any extra user defined geometry
      """
 
-    # left empty for now but I think we should have methods to
-    # add Geometries together (i.e. choose 3D macro with 1D micro etc)
+    def __init__(self, *geometries, custom_geometry={}):
```
Scottmar93: I really like this as an approach for creating geometries.

PR #187, opened 2019-03-06
Issue 185 cutoff event

### LINKED ISSUE #185: Implement cut-off events in solvers
Solver needs to stop when certain events occur such as concentration going below zero or voltage goes below a threshold.
These events could be implemented in the model, e.g. 
```python3
self.events = [c_e, V-V_threshold]
```
and then the solver knows to stop if any of the values in `self.events` go negative.

**Motivation**
Solutions with concentration below zero are unphysical. In particular, for lead-acid, the open-circuit voltage is a logarithmic function of concentration, so we get a run-time warning if the concentration goes below zero, e.g. [here](https://travis-ci.org/tinosulzer/PyBaMM/jobs/500904077#L2387)

tinosulzer:
# Description

Implement cut-off events in solvers, add cut-off events to models.
Fixes #185 

## Type of change

# Key checklist:

## Further checks: 

### DIFF
--- a/pybamm/discretisations/discretisation.py
+++ b/pybamm/discretisations/discretisation.py
@@ -70,6 +70,10 @@ class Discretisation(object):
         # model.initial_conditions and model.boundary_conditions
         model.variables = self.process_dict(model.variables)
 
+        # Process events
+        model.events = self.process_list(model.events)
+        model.concatenated_events = self.concatenate(*model.events)
+
         # Check that resulting model makes sense
         self.check_model(model)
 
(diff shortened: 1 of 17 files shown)


pybamm/solvers/scipy_solver.py
```diff
@@ -60,6 +68,7 @@ def integrate(self, derivs, y0, t_eval):
             method=self.method,
             rtol=self.tol,
             atol=self.tol,
+            events=events,
         )
         # TODO: implement concentration cut-off event
```
rtimms: I guess this is done now?

tinosulzer: yep!

pybamm/discretisations/discretisation.py
```diff
@@ -181,6 +184,35 @@ def process_dict(self, var_eqn_dict):
             # keep calling .id
         return var_eqn_dict
 
+    def process_list(self, eqn_list):
+        """Discretise a list of equations, broadcasting if necessary (for model.events).
```
martinjrobins: why do the events need to be broadcasted? Presumably you only want a single number (ie. the minimum concentration)?

PR #188, opened 2019-03-06
Issue 177 clean up parameters

### LINKED ISSUE #177: Clean up parameters
After discussions with @Scottmar93 and @rtimms we have come up with a set of parameters that is common to both lead-acid and lithium-ion models.

We now need to clean up the files `standard_parameters` and `standard_parameters_lead_acid` to use this common parameter framework. We should also use more informative names in the parameter strings as there is no reason to use an abbreviation there, e.g. 
```python3
c_max = Parameter("maximum electrolyte concentration")
```

tinosulzer:
# Description

- Change parameter notation to be consistent between different chemistries
- Change parameter names to be more informative
-  Add some functions (diffusivity, etc)
- Make parameters take up less space in model statement

Fixes #177 

## Type of change

# Key checklist:

## Further checks: 


pybamm/models/li_ion/single_particle_model.py
```diff
@@ -9,91 +9,60 @@
...
     """
 
     def __init__(self):
         super().__init__()
 
         # Variables
-        cn = pybamm.Variable("cn", domain="negative particle")
-        cp = pybamm.Variable("cp", domain="positive particle")
+        c_n = pybamm.Variable("Particle concentration", domain="negative particle")
+        c_p = pybamm.Variable("Particle concentration", domain="positive particle")
```
rtimms: Does it matter that both `c_n` and `c_p` are called `"Particle concentration"`, or should they be called `"Negative particle concentration"` and `"Positive particle concentration"`? Does having the same variable name cause an issue, even if they have different domains?

pybamm/models/li_ion/single_particle_model.py
```diff
@@ -9,91 +9,60 @@
...
-        dcpdt = -pybamm.div(Np)
-        self.rhs = {cn: dcndt, cp: dcpdt}
+        N_n = -spli.C_n * spli.D_n(c_n) * pybamm.grad(c_n)
+        dc_n_dt = -pybamm.div(N_n)
+        N_p = -spli.C_p * spli.D_p(c_p) * pybamm.grad(c_p)
+        dc_p_dt = -pybamm.div(N_p)
+        self.rhs = {c_n: dc_n_dt, c_p: dc_p_dt}
 
         # Boundary conditions
         # Note: this is for constant current discharge only
```
rtimms: note can be removed as `i_cell` is now a function of time

pybamm/parameters/standard_parameters.py
```diff
@@ -16,11 +16,11 @@
 Macroscale Geometry
 -------------------
 
-Ln, Ls, Lp
+L_n, Ls, Lp
     The widths of the negative electrode, separator and positive electrode, respectively
```
rtimms: Also change `Ls` and `Lp` to `L_s` and `L_p`, similar below on lines 23 and 57

PR #194, opened 2019-03-07
Issue 121 lead acid composite

### LINKED ISSUE #121: Implement lead-acid "composite" model

tinosulzer:
# Description

- Add the composite model for lead-acid
- Fix interface to work for both lead-acid and lithium-ion

Fixes #121 

## Type of change

# Key checklist:

## Further checks: 


pybamm/models/submodels/electrolyte_current.py
```diff
@@ -0,0 +1,118 @@
...
+        j0bar_1p = pybamm.Integral(j0_1p, x_p) / param.l_p
+        V_1 = (
+            Phibar_1p
+            + dU_0p__dc0 * cbar_1p
+            - j0bar_1p * pybamm.Function(np.tanh, eta_0p) / j0_0p
+        )
+
+        # Variables
+        self.variables = {
+            "c": c_e,
```
Scottmar93: Should these outputs be "c_e" and "phi_e" so as not to name clash elsewhere?

Scottmar93: Also should we output the current and the flux aswell since other submodels do aswell?

Scottmar93: Also should we be only outputting the potentials here and not "c_e"

pybamm/models/submodels/electrolyte_current.py
```diff
@@ -0,0 +1,118 @@
...
+# Equation classes for the electrolyte current
+#
+from __future__ import absolute_import, division
+from __future__ import print_function, unicode_literals
+import pybamm
+
+import numpy as np
+
+
+class FirstOrderPotential(pybamm.BaseModel):
```
Scottmar93: Should the name indicate that this is specific to the StefanMaxwell electrolyte? We may have a NernstPlanck asymptotic solution aswell at some point

Issue #189, opened 2019-03-06
Symbolic differentiation


tinosulzer:
We need a function (e.g. `pybamm.diff`) that calculates the derivative of an expression tree with respect to a variable. For example, some lead-acid models require the user to provide a function for the open-circuit potential `U(c)` while others require its derivative `dU/dc` (and also for the exchange-current density). 

We could either ask the user to provide the appropriate derivatives manually at the same time as the functions themselves, or calculate the derivative using the expression tree and appropriate applications of the chain rule and power rule. I think the latter is safer

tinosulzer:
Autograd is working well for the arbitrary numpy functions, but the drawback is that we have to import `autograd.numpy` instead of `numpy`. Are we happy to force the user to do this?

rtimms:
It looks like `autograd` is probably the easiest tool out there for calculate the Jacobian for use in the solver, which would force us to use `autograd.numpy` anyway. Maybe there are other options, but from what I can see this is probably the simplest fix

martinjrobins:
does that mean that the user *wouldn't* be able to use normal numpy?

tinosulzer:
We could do something like 
```python3
try:
    import autograd.numpy as np
except ModuleNotFoundError:
    import numpy as np
```

Then throw errors/warnings that the functionality provided by autograd (symbolic differentiation, automatic jacobian in #197) can't be used if it's not installed.
Might be best to do this in `__init__.py` and then `pybamm.numpy as np` whenever needed so that we don't need to repeat the above 4 lines every time.

In any case, autograd is easy to install with pip and seems well maintained so it should be fine to just include in `setup.py`

martinjrobins:
seems sensible to start out using autograd for both this issue and #197 and we can see how the PR looks. if we use autograd this should be fairly straightforward to implement

tinosulzer:
I've submitted a PR for this issue

tinosulzer:
Re-opening: need to deal with `FunctionParameter` nodes better

PR #198, opened 2019-03-12
Issue 197 auto diff

### LINKED ISSUE #197: automatic differentiation
Automatically compute the jacobian for use in the solver. This speeds up sundials, and gives better convergence etc.  One option is to use `autograd`. This has the benefit of being able to handle `numpy` functions, but the cost is that you have to import `autograd.numpy` instead of just `numpy` (as mentioned in #189). Since we already have the expression tree, we could write something ourselves to take the derivative of the expression tree (defining derivatives of all the standard functions and use things like `pybamm.cos` instead of `np.cos` etc), but maybe this is a lot of effort if there are already tools out there which can do it for us?

rtimms:
Uses `autograd` to automatically compute the jacobian for use in `ScikitsDaeSolver`. At the moment it assumes autograd is installed (as in #189), but thought I'd make the PR now and either add a pybamm.numpy (see https://github.com/tinosulzer/PyBaMM/issues/189#issuecomment-471639600) after comments, or as a separate issue. The jacobian is calculated within `ScikitsDaeSolver` instead of in `DaeSolver`, since the required form of the jacobian may be different for different solvers. Happy to move it if it would fit better elsewhere. 

Fixes #197

## Type of change

# Key checklist:

## Further checks: 


pybamm/solvers/dae_solver.py
```diff
@@ -69,6 +69,8 @@ def integrate(self, residuals, y0, ydot0, t_eval, events=None):
             The initial conditions
         t_eval : numeric type
             The times at which to compute the solution
+        jacobian : method, optional
+        A function that takes in t, y and ydot and returns the Jacobian
```
tinosulzer:
```suggestion
            A function that takes in t, y and ydot and returns the Jacobian
```

pybamm/solvers/scikits_dae_solver.py
```diff
@@ -56,6 +57,8 @@ def integrate(self, residuals, y0, ydot0, t_eval, events=None):
             The initial conditions
         t_eval : numeric type
             The times at which to compute the solution
+        jacobian : method, optional
+        A function that takes in t, y and ydot and returns the Jacobian
```
tinosulzer:
```suggestion
           A function that takes in t, y and ydot and returns the Jacobian
```

pybamm/solvers/scikits_dae_solver.py
```diff
@@ -69,6 +72,23 @@ def rootfn(t, y, ydot, return_root):
             return_root[:] = [event(t, y) for event in events]
 
         extra_options = {"old_api": False, "rtol": self.tol, "atol": self.tol}
+
+        # If no Jacobian provided (default), use autograd to compute the
+        # Jacobian. If autograd not installed, the solver will approximate the
+        # Jacobian (see SUNDIALS documentation).
```
tinosulzer: if you put this in the docstring you can put in a hyperlink to SUNDIALS docs (see [here](http://docutils.sourceforge.net/docs/user/rst/quickref.html#external-hyperlink-targets))

pybamm/solvers/scikits_dae_solver.py
```diff
@@ -69,6 +72,23 @@ def rootfn(t, y, ydot, return_root):
...
+        # TO DO: check here if autograd installed
```
tinosulzer: should be possible using similar code to the `scikits_odes_spec` bloc at the top of this file

pybamm/solvers/scikits_dae_solver.py
```diff
@@ -77,3 +97,38 @@ def rootfn(t, y, ydot, return_root):
...
+        t : numeric type
+            The time at which to evaluate the Jacobian
+        y : numeric type
+            The values of the discretised variables used to evaluate the Jacobian
+        ydot : numeric type
+            The values of the discretised time derivatives used to evaluate the Jacobian
+        """
+        mass_matrix_eval = -self.jacobian_ydot(t, y, ydot)
+        jac_rhs_alg_eval = self.jacobian_rhs_alg(t, y, ydot)
+        return (mass_matrix_eval, jac_rhs_alg_eval)
```
tinosulzer: I'm not convinced that the functions `auto_jac` and `jacobian` need to be abstracted like this - as far as I can tell, `auto_jac` is only used once, and `jacobian` is only called by the tests?

pybamm/solvers/scikits_dae_solver.py
```diff
@@ -77,3 +97,38 @@ def rootfn(t, y, ydot, return_root):
...
+        Parameters
+        ----------
+        residuals : method
+            A function that takes in t, y and ydot and returns the residuals of the
+            equations
+
+        """
+        self.jacobian_ydot = autograd.jacobian(residuals, 2)
+        self.jacobian_rhs_alg = autograd.jacobian(residuals, 1)
+        return self.jacobian_ydot, self.jacobian_rhs_alg
```
tinosulzer: python functions should either assign a class attribute or return a result, doing both is potentially confusing

rtimms: Thanks for your help, changing to `set_auto_jac` and removing `jacobian` as discussed. I'll also add jacobian calculation to the other solvers

rtimms:
The outstanding issue with this is to do with `autograd` not liking sparse matrices. You get the error

`TypeError: no supported conversion for types: (dtype('float64'), dtype('O'))`

which comes about from trying to use `matmul`  (can be reproduce by running e.g. `test_li_ion`).  Chatted with @tinosulzer about making an `evaluate_full` method which replaces the sparse matrices with full ones, and is called when calculating the jacobian. Does anyone have any other thoughts on this @martinjrobins @Scottmar93 ? An alternative would be to wrap `scipy` in a similar way to how `numpy` is wrapped by `autograd`, but I think that might be beyond my capabilities!

rtimms:
I've made a separate issue for including a user-supplied jacobian in the solvers (see #223) so that the code for providing the solvers with a jacobian and testing the solution can be pulled in. left to do on here is actually compute the jacobian of a model (probably by differentiating the discretised tree wrt the state vector)

martinjrobins:
> The outstanding issue with this is to do with `autograd` not liking sparse matrices. 

Can we calculate the jacobian via the `Symbol.diff` function? and leave autograd to the user-defined function? This might be the easiest path, rather than trying to wrap scipy.sparse in a way that autograd likes.

rtimms:
I agree, I think doing `Symbol.diff` is the way to go. For the jacobian we need to take the derivative of the tree with respect to the entire state vector after the tree has been discretised, so I think we need to write something that differentiates a slice of the state vector with respect to the entire state vector, if that makes sense?

martinjrobins:
yup, so if you differentiate (wrt a statevector of size n) a slice of the statevector with size m, you would get a matrix of size m x n (or perhaps the transpose of that....), then that would presumably go through some sort of concatenation with other branches of the tree, to finally give you a jacobian of n x n

rtimms:
Just to update, computing the Jacobian of the tree using `jac` and then concatenating is now working for everything except Newman Tiedemann, where it is giving a value error. This is due to something slipping through the net, and ending up being a `MatrixMultiplication` instead of a regular `Multiplication`. It's a little tricky to debug as the tree is so complicated, but hoping taking the Jacobian of the simplified tree might fix this. I've put in a Try/Except for now which checks the Jacobian can actually evaluate before passing it to the solver (again, the only Jacobian giving an error is Newman Tiedemann). 

At present, taking the Jacobian of the simplified tree doesn't work due to shape mismatches from reshaping vectors when simplifying multiplication (see #264), but I thought I'd wait to see how simplify may change before trying to fix this here.

Other things to do are:

-  come up with some more difficult tests, maybe on a simple model. At the moment the tests which are explicitly for `jac` are on some relatively simple equations. 
-  get sparse Jacobian into scikits
-  when using autograd to compute the Jacobian of a function it returns a dense result. This then gets put into a sparse format when concatenated into the full Jacobian, but maybe it is better to change it to a sparse result as part of `evaluate` (or somewhere else more sensible?). At the moment if there ends up being a dense autograd jacobian in an equation, then doing any operations with it turns the other elements of the equations into dense matrices too. I suspect this is not ideal...

Scottmar93:
Is that a green tick I see next to the tests?

rtimms:
> Is that a green tick I see next to the tests?

Or your eyes deceive you??!!

It's passing the tests, apart from the jacobian in Newman Tiedemann doesn't evaluate due to a ValueError. At the moment this is being caught by a Try/Except (if the jacobian can't evaluate, then scikits just approximates it). I think this may be fixed by taking the jacobian of the simplified tree, and then simplifying before evaluate, but at the moment taking the jacobian of the simplified tree doesn't work due to matrix multiply reshaping things. Waiting on #264 before I attempt a fix.

tinosulzer:
Fixed matrix multiplications have been merged to master so jacobian might now work on the simplified tree

rtimms:
thanks @tinosulzer it looks like this is working well! still getting warnings about constant functions, but this should be fixed by #266 . It isn't particularly fast, but my guess is this will be helped by changing the way we use autograd (as we were chatting about yesterday) and by getting the sparse jacobian into scikits

tinosulzer:
Let's fix conflicts on this and merge if it works (assuming no more warnings about constant functions?), and open two new issues for:
- sparse autograd (elementwise grad for functions)
- sparse scikits

rtimms:
sounds good. have merged with master but still getting a warning in newman tiedemann about constant functions. this is the only one though. trying to figure out why now...

rtimms:
it is somewhere in the algebraic part if you have any suggestion where to look @tinosulzer ?

tinosulzer:
It looks like it's to do with `node_to_edge`:
```
ipdb> jac_algebraic.children[1].children[0].children[0].children[1].children[0].children[1].children[0]                                                                                                     
Function(0x3736d02b070754, function (jacobian_of_node_to_edge_wrt_argnum_0), children=['-1980753.5287086302 * 1.0 - y[9:12] ** 1.5'], domain=[])
ipdb> jac_algebraic.children[1].children[0].children[0].children[1].children[0].children[1].children[0].evaluate(0,y0)                                                                                      
/home/sulzer/Documents/Batteries/PyBaMM/venv/lib/python3.6/site-packages/autograd/tracer.py:14: UserWarning: Output seems independent of input.
  warnings.warn("Output seems independent of input.")
array([[0., 0., 0.],
       [0., 0., 0.],
       [0., 0., 0.]])
ipdb> jac_algebraic.children[1].children[0].children[0].children[1].children[0].children[1].children[0].children[0].evaluate(0,y0)                                                                          
array([-638230.01594354, -638230.01594354, -638230.01594354])
ipdb> jac_algebraic.children[1].children[0].children[0].children[1].children[0].children[1].children[0].func(1.)                                                                                            
array(1.)
ipdb> jac_algebraic.children[1].children[0].children[0].children[1].children[0].children[1].children[0].func(np.array([1.,2.]))                                                                             
/home/sulzer/Documents/Batteries/PyBaMM/venv/lib/python3.6/site-packages/autograd/tracer.py:14: UserWarning: Output seems independent of input.
  warnings.warn("Output seems independent of input.")
array([[0., 0.],
       [0., 0.]])
ipdb> jac_algebraic.children[1].children[0].children[0].children[1].children[0].children[1].children[0].func(np.array([1.,2.,3.]))                                                                          
/home/sulzer/Documents/Batteries/PyBaMM/venv/lib/python3.6/site-packages/autograd/tracer.py:14: UserWarning: Output seems independent of input.
  warnings.warn("Output seems independent of input.")
array([[0., 0., 0.],
       [0., 0., 0.],
       [0., 0., 0.]])
ipdb> jac_algebraic.children[1].children[0].children[0].children[1].children[0].children[1].children[0].func(np.array([1.3,2.5]))                                                                           
/home/sulzer/Documents/Batteries/PyBaMM/venv/lib/python3.6/site-packages/autograd/tracer.py:14: UserWarning: Output seems independent of input.
  warnings.warn("Output seems independent of input.")
array([[0., 0.],
       [0., 0.]])
```
FYI, if you do 
```
ipdb> import warnings                                                                                                                                                                                       
ipdb> warnings.simplefilter("always")
```
after starting ipdb then warnings will always be displayed (not just the first time), so you can step through the tree and find where it's failing

rtimms:
This was due to importing `numpy` instead of `autograd.numpy` in `finite_volume`. This is now fixed. There are some FutureWarnings, but these are in autograd. Output of the above `node_to_edge` part is now 
```
jac_algebraic.children[1].children[0].children[0].children[1].children[0].children[1].children[0].evaluate(0,y0)                                                                                      
/home/timms/Documents/PyBaMM/venv/lib/python3.6/site-packages/autograd/numpy/numpy_vjps.py:444: FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.
  return lambda g: g[idxs]
/home/timms/Documents/PyBaMM/venv/lib/python3.6/site-packages/autograd/numpy/numpy_vjps.py:444: FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.
  return lambda g: g[idxs]
/home/timms/Documents/PyBaMM/venv/lib/python3.6/site-packages/autograd/numpy/numpy_vjps.py:444: FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.
  return lambda g: g[idxs]
/home/timms/Documents/PyBaMM/venv/lib/python3.6/site-packages/autograd/numpy/numpy_vjps.py:444: FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.
  return lambda g: g[idxs]
/home/timms/Documents/PyBaMM/venv/lib/python3.6/site-packages/autograd/numpy/numpy_vjps.py:444: FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.
  return lambda g: g[idxs]
/home/timms/Documents/PyBaMM/venv/lib/python3.6/site-packages/autograd/numpy/numpy_vjps.py:444: FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.
  return lambda g: g[idxs]
array([[ 1.5, -0.5,  0. ],
       [ 0.5,  0.5,  0. ],
       [ 0. ,  0.5,  0.5]])
```

tests/shared.py
```diff
@@ -24,16 +24,28 @@ def gradient(self, symbol, discretised_symbol, boundary_conditions):
...
         for domain in symbol.domain:
             n += self.mesh[domain][0].npts
-        divergence_matrix = pybamm.Matrix(np.eye(n))
+        divergence_matrix = pybamm.Matrix(eye(n))
         return divergence_matrix @ discretised_symbol
 
+    def compute_diffusivity(
+        self, symbol, extrapolate_left=None, extrapolate_right=None
+    ):
+        return symbol
```
tinosulzer: can leave this out to take default behaviour from the base `SpatialMethod` class

pybamm/discretisations/discretisation.py
```diff
@@ -170,8 +167,8 @@ def create_mass_matrix(self, model):
 
         Parameters
         ----------
-        model : :class:`pybamm.BaseModel`
-            Model to dicretise. Must have attributes rhs, initial_conditions and
+        model : :class:`pybamm.BaseModel` (or subclass)
```
tinosulzer: Very minor point: I now think it might be a bit cleaner to leave out the `or subclass` in the doc strings

pybamm/expression_tree/concatenations.py
```diff
@@ -99,6 +100,14 @@ def _concatenation_evaluate(self, children_eval):
         else:
             return np.concatenate([child for child in children_eval])
 
+    def jac(self, variable):
+        """ See :meth:`pybamm.Symbol.jac()`. """
+        if len(self.children) == 0:
+            # NOTE: need to think about if this is the right thing to do here
+            return pybamm.Scalar(0)
```
tinosulzer: looks like this works fine? i.e. ok to remove the note?

pybamm/expression_tree/concatenations.py
```diff
@@ -208,6 +216,14 @@ def _concatenation_evaluate(self, children_eval):
 
         return vector
 
+    def jac(self, variable):
+        """ See :meth:`pybamm.Symbol.jac()`. """
+        if len(self.children) == 0:
+            # NOTE: need to think about if this is the right thing to do here
+            return pybamm.Scalar(0)
```
tinosulzer: looks like this works fine?

pybamm/expression_tree/concatenations.py
```diff
@@ -232,3 +248,57 @@ def simplify(self):
...
+
+    def evaluate_child(self, child, t, y):
+        """ Evaluates the child and, if the results is dense, converts the
+        result as a sparse matrix.
+        """
+        # NOTE: I think this probably a very hacky way of doing this, but need
+        # some way of concatenating sparse matrices with dense matrices that
+        # come from evaluate. The other way would be to make all the matrices
+        # dense and then do a NumpyConcatenation, but I fear this will be bad
+        # when the Jacobian is very large
```
tinosulzer: new treatment of functions should solve this?

pybamm/discretisations/discretisation.py
```diff
@@ -209,11 +206,32 @@ def create_jacobian(self, model):
...
+        jac_algebraic = dict.fromkeys(model.algebraic.keys())
+        for eqn_key, eqn in model.algebraic.items():
+            jac_algebraic[eqn_key] = eqn.jac(y)
+
+        jacobian = self.sparse_stack(*jac_rhs.values(), *jac_algebraic.values())
+
+        def jacfn(t, y):
+            return jacobian.simplify().evaluate(t, y)
+
+        model.jacobian = jacfn
```
tinosulzer: are we using this function at all now? if not, ok to remove it entirely

pybamm/expression_tree/concatenations.py
rtimms: Yep, this can be resolved once we replace `node_to_edge` etc. with matrix multiplication

pybamm/discretisations/discretisation.py
rtimms: yeah this isn't used an was a hangover from before. have removed.

rtimms:
Thanks for the review @tinosulzer! I agree the different treatment of mass matrix and jacobian could be confusing. I think it should still be possible to set either (or both) as a model attribute. There may be some user case where the user wants to attach a particular jacobian (e.g. an approximation to it for cases where evaluating jac is slow, or just to compare approximate vs exact jacobian) to a model and use this instead. Maybe a model attribute isn't the right way to this, though. We can discuss and fix this in the future.

PR #202, opened 2019-03-16
Issue 189 diff function parameter

### LINKED ISSUE #189: Symbolic differentiation

tinosulzer:
# Description

Implement `diff` for a `FunctionParameter` object. The `value` attribute of a `Scalar` has to be converted to a float to work with autograd but this doesn't seem to cause any problems
Fixes #189 

## Type of change

# Key checklist:

## Further checks: 


pybamm/expression_tree/function_parameter.py
```diff
@@ -23,5 +23,13 @@ class FunctionParameter(pybamm.UnaryOperator):
 
     """
 
-    def __init__(self, name, child):
+    def __init__(self, name, child, diff_variable=None):
```
martinjrobins: the docstring for this class doesn't match the `__init__`, and needs to cover `diff_variable`

PR #220, opened 2019-03-26
Issue 161 change parameters

### LINKED ISSUE #161: Changable parameters
**Summary**
Add ability to have parameter values that change without recreating the tree. 

- Add an extra subclass of symbol `ChangableParameter`. 
- Do not insert any values for `ChangableParameters` when we `params.process(model)` 
- During `params.process(model)` replace each `ChangableParameter` by `changable_parameters[ChangableParameters]` (a dict with keys that are the changable paramerters).
- add an extra optional input to `model.concatenated_rhs.evaluate(t, y, changable_params)` which is a dict of the changed parameters with their values which will update the `changable_parameters` dict. 
- create a model.set_param() which takes in a dict of parameter values and puts them into the correct location in the changable_params vector

 **Motivation**
 Allows parameters to be changed easily without rebuilding the tree. Efficiently changing parameter values is a requirement of the control community. 

Time dependent functions could also be implemented within this approach. 

**Questions** 
It would be possible to have this as the default approach for setting all parameters. We should consider this as it means someone doesn't need to rebuild the discretisation if they realise afterwards that they want to change a parameter value. I also don't think that this will slow thing down to any great extent. We would also only pass in the parameters values that are changed an update the parameters dict. 
What are peoples thought on this?

tinosulzer:
# Description

Basic functionality to update a parameter value without re-discretising. A Scalar object now inherits the name from Parameter
We lose out on a bit of functionality since a Scalar can no longer be assumed to be constant (and so doesn't get simplified to a Vector). So if the discretisation step isn't too slow, it might just be better to start from scratch each time. Alternatively, we could make the `simplify` function make a copy of the model so that we retain a copy of the original information for changing parameters
Fixes #161 

## Type of change

# Key checklist:

## Further checks: 


pybamm/expression_tree/symbol.py
```diff
@@ -354,13 +354,22 @@ def is_constant(self):
...
 
         """
-
         # if any of the nodes are instances of any of these types, then the whole
         # expression depends on either t or y
         search_types = (pybamm.Variable, pybamm.StateVector, pybamm.IndependentVariable)
 
         # do the search, return true if no relevent nodes are found
-        return all([not isinstance(n, search_types) for n in self.pre_order()])
+        # a Scalar with a name different to its is not constant, for parameter changing
```
martinjrobins:
```suggestion
        # a Scalar with a name different to its name is not constant, for parameter changing
```

martinjrobins: why wouldn't a Scalar with a name not be constant? I presume that the parameters wouldn't be changed *during* a simulation?

tinosulzer:
If PyBaMM thinks that a scalar is constant then it might replace it with something else (eg change a DomainConcatenation to a Vector) and then we would lose the information about what parameter that scalar came from. My impression was that we'd want to change the parameters on the "linear algebra" tree that gets fed to the simulation, rather than the tree that gets fed to the discretisation? Currently, once we have discretised a tree, we lose information about its form pre-discretisation (since discretisation is in-place)

martinjrobins:
yes, I agree that in general we can't simplify out a scalar with a name. The exception to this is within the solver, once the parameters are set the solver should be able to create a maximally simplified expression tree to actually execute during the time-stepping loop.

But the `is_constant` function should return true if the expression tree does not depend on `y` or `t`, that is the definition of "constant" for this function according to the docstring. So if you want to make this change I think you need to update the docstring.

Perhaps in the future we should avoid individual classes doing simplifications on the expression tree, except in their `simplify` function, just so everything is in the one place and is consistent. Then simplify could take a parameter that turned on/off the simplification of named scalars.

tinosulzer:
sure, that makes sense. I will revert to the old `is_constant` and remove the simplification to a `Vector` in the places it appears. I definitely agree that only doing simplifications explicitly is the way to go

PR #221, opened 2019-03-26
Issue 215 mass matrix

### LINKED ISSUE #215: Add mass matrix
Allow construction of a non-trivial mass matrix when discretising the tree to allow for different discretisations e.g. 1D piece-wise finite element, or control volumes. This then needs to be passed to the solver.

rtimms:
# Description

Adds `mass_matrix` attribute to model which is processed during discretisation to give the correct mass matrix for a given spatial method

Fixes #215 

## Type of change

# Key checklist:

## Further checks: 

### DIFF
--- a/pybamm/discretisations/discretisation.py
+++ b/pybamm/discretisations/discretisation.py
@@ -7,6 +7,7 @@ import pybamm
 
 import copy
 import numpy as np
+from scipy.sparse import block_diag, csr_matrix
 
 
 class Discretisation(object):
@@ -74,6 +75,9 @@ class Discretisation(object):
         for idx, event in enumerate(model.events):
             model.events[idx] = self.process_symbol(event)
         model.concatenated_events = self.concatenate(*model.events)
+
+        # Process mass matrix
+        self.process_mass_matrix(model)
 
         # Check that resulting model makes sense
         self.check_model(model)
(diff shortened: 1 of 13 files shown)


tests/test_spatial_methods/test_finite_volume.py
```diff
@@ -1028,6 +1029,92 @@ def test_discretise_spatial_variable(self):
...
+        disc = pybamm.Discretisation(mesh, spatial_methods)
+
+        combined_submesh = mesh.combine_submeshes(*whole_cell)
+        disc.process_model(model)
+
+        # mass matrix
+        mass = np.eye(combined_submesh[0].npts)
+        np.testing.assert_array_equal(mass, model.mass_matrix.entries.toarray())
+
+    def test_mass_matrix_Neumann_bcs(self):
```
tinosulzer: same function twice?

pybamm/discretisations/discretisation.py
```diff
@@ -152,6 +158,43 @@ def process_rhs_and_algebraic(self, model):
         model.algebraic = self.process_dict(model.algebraic)
         model.concatenated_algebraic = self.concatenate(*model.algebraic.values())
 
+    def process_mass_matrix(self, model):
```
tinosulzer: I think "create_mass_matrix" is a more appropriate name

tests/test_spatial_methods/test_finite_volume.py
rtimms: Good spot, thanks! had originally had tests for Dirichlet and Neumann BCs as I thought they were handled differently. have removed the duplicate now

pybamm/discretisations/discretisation.py
rtimms: I agree

pybamm/solvers/scikits_dae_solver.py
```diff
@@ -43,7 +43,7 @@ def method(self):
     def method(self, value):
         self._method = value
 
-    def integrate(self, residuals, y0, t_eval, events=None):
+    def integrate(self, residuals, y0, t_eval, events=None, mass_matrix=None):
```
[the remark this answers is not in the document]
rtimms: For finite volumes the mass matrix is just the identity, but for other schemes this may be different. Jamie has some specific implementations of finite elements and control volumes where the mass matrix is different, and this may be a way around needing a separate strong and weak form for each equation. From what I can tell you can't pass a mass matrix to scipy odeint -- if you want to use that for problems of the form M*y_dot = f(t,y) you have to invert M first, which I guess is not ideal. You can pass the mass matrix into scikits.odes if using IDA, but it is wrapped up with jacobian so can't be put in until that is working. In this case, maybe it is better to not pass this to all solvers?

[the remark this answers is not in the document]
martinjrobins: no, probably leave it as is. even if the solvers we currently have don't support it, they all could (in theory) use the mass matrix for time integration, so it should be part of the interface. I'm happy to merge it as is

PR #224, opened 2019-03-27
Issue 210 different bcs

### LINKED ISSUE #210: Different boundary condition types on left and right
For the lead-acid DAE a case has come up where we need Dirichlet bcs on one side but Neumann on the other, this isn't currently possible so the finite volume discretisation needs to be made more flexible

tinosulzer:
# Description

Allow different types of boundary condition (Dirichlet and Neumann) on each side of an equation, and update fluxes accordingly
Fixes #210 

## Type of change

# Key checklist:

## Further checks: 


pybamm/meshes/meshes.py
```diff
@@ -47,6 +47,9 @@ def __init__(self, geometry, submesh_types, submesh_pts):
                 submesh_types[domain](geometry[domain]["primary"], submesh_pts[domain])
             ] * repeats
 
+        # add ghost meshes
+        self.add_ghost_meshes()
```
Scottmar93: Does this replace the add_ghost_meshes() in discretisation.__init__() ?

pybamm/discretisations/discretisation.py
```diff
@@ -351,16 +351,30 @@ def process_binary_operators(self, bin_op):
...
+            # Extrapolate at either end depending on the ghost cells (from gradient)
+            extrapolate_left = any(
+                [x.has_left_ghost_cell for x in new_left.pre_order()]
+            )
+            extrapolate_right = any(
+                [x.has_right_ghost_cell for x in new_left.pre_order()]
+            )
             new_right = self._spatial_methods[bin_op.domain[0]].compute_diffusivity(
-                new_right
+                new_right, extrapolate_left, extrapolate_right
```
Scottmar93: Does this just save us from having to add ghost nodes to new_right?

pybamm/meshes/meshes.py
tinosulzer: it should do hopefully, I think this is a more sensible place for it

pybamm/spatial_methods/finite_volume.py
```diff
@@ -124,11 +135,7 @@ def gradient_matrix(self, domain):
         # Create 1D matrix using submesh
         n = submesh.npts
         e = 1 / submesh.d_nodes
-        data = np.vstack(
-            [np.concatenate([-e, np.array([0])]), np.concatenate([np.array([0]), e])]
-        )
-        diags = np.array([0, 1])
-        sub_matrix = spdiags(data, diags, n - 1, n)
+        sub_matrix = diags([-e, e], [0, 1], shape=(n - 1, n))
```
Scottmar93: nice

pybamm/spatial_methods/finite_volume.py
```diff
@@ -234,21 +271,17 @@ def divergence_matrix(self, domain, bc_type="dirichlet"):
...
-            # we don't have to act on bc fluxes which are now in
-            # the bc vector
-            data = np.vstack([-e[1:], e[:-1]])
-            diags = np.array([-1, 0])
-            sub_matrix = spdiags(data, diags, n - 1, n - 2)
+        if bc_type == "dirichlet_dirichlet":
+            sub_matrix = diags([-e, e], [0, 1], shape=(n - 1, n))
+        elif bc_type == "dirichlet_neumann":
+            # we don't have to act on right bc flux which is now in the bc vector
+            sub_matrix = diags([-e, e], [0, 1], shape=(n - 1, n - 1))
```
Scottmar93: You probably have already checked this but just want to say that we should be really careful to test each of these submatrix forms as results really depend on having the correct vectors e[1:] or e. If we don't the 1/dx terms are associated with the wrong variables. At the moment this will just appear as errors at the neg_sep and sep_pos interfaces.

pybamm/discretisations/discretisation.py
tinosulzer: this is for the `NodeToEdge` function to extrapolate appropriately when there are ghost cells

pybamm/meshes/meshes.py
Scottmar93: Yeah agree. We should remove the one in discretisation.__init__(), I think its still there. Otherwise, we will have "negative electrode_left_ghost_left_ghost" kicking around.

PR #225, opened 2019-03-27
Issue 213 create extract op

### LINKED ISSUE #213: Create Extract operator for x dependent particle BCs
**Summary** 
Create a unary operator `Extract` which contains the value of an index to be extracted from child. Upon evaluation, `Extract` will first evaluate the child, creating a vector, and then perform `child[i]` to obtain the correct entry. 

 **Motivation**
 The particle boundary conditions in the DFN are dependant on x. We therefore need a way to extract the correct x location for each particle from the boundary condition (which is a vector in x)

**Additional context**
Optimisations to the tree which involve evaluating the same pattern fit nicely with this approach as the children of different particles `Extract` operators will be the same. 
With current approach of having all secondary dimensions just create a list of independent points, this should also hold for the 2+1D case without any changes

Scottmar93:
# Description

Create an `Extract` node which which upon evaluation takes the ith entry of the evaluated child. An appropriate if state for when this Extract operator should be inserted into the tree is included when creating the boundary conditions. We can now have x dependent boundary conditions on the particles.

Fixes #213

## Type of change

# Key checklist:

## Further checks: 


tests/test_spatial_methods/test_finite_volume.py
```diff
@@ -1093,7 +1153,7 @@ def test_p2d_mass_matrix_shape(self):
         model.rhs = {c: pybamm.div(N)}
         model.initial_conditions = {c: pybamm.Scalar(0)}
         model.boundary_conditions = {
-            c: {"left": pybamm.Scalar(0), "right": pybamm.Scalar(0)}
+            N: {"left": pybamm.Scalar(0), "right": pybamm.Scalar(0)}
```
Scottmar93: Not sure why this has changed. Will check if this was an error

pybamm/discretisations/discretisation.py
```diff
@@ -53,7 +53,10 @@ def process_model(self, model):
 
         """
         # set boundary conditions (only need key ids for boundary_conditions)
-        self._bcs = {key.id: value for key, value in model.boundary_conditions.items()}
+        self._bcs = {
+            key.id: self.process_dict(value)
+            for key, value in model.boundary_conditions.items()
+        }
```
tinosulzer: for clarity the boundary conditions should be processed outside of this function (and maybe `self._bcs` should be changed to `self._processed_bcs`)

tests/test_spatial_methods/test_finite_volume.py
```diff
@@ -1055,6 +1055,66 @@ def test_discretise_spatial_variable(self):
...
+            error = np.linalg.norm(approx_internal - exact_internal) / np.linalg.norm(
+                exact_internal
+            )
+            return error
+
+        # Get errors
+        ns = 10 * (2 ** np.arange(2, 7))
+        errs = np.array([get_l2_error(int(n)) for n in ns])
+
+        # Get rates: expect h**1.5 convergence because of boundary conditions
```
tinosulzer:
```suggestion
        # Get rates: expect h**2 convergence
```

Scottmar93: fixed along with other comment errors

pybamm/__init__.py
```diff
@@ -72,6 +72,7 @@ def version(formatted=False):
     Negate,
     AbsoluteValue,
     Function,
+    Extract,
```
martinjrobins: I think a better term would be Index or IndexVector

martinjrobins:
looks great, merging...

PR #226, opened 2019-03-27
Issue 223 user jacobian

### LINKED ISSUE #223: User-supplied Jacobian
**Summary**
Allow users to supply a Jacobian to the solvers.

 **Motivation**
Users can either enter a Jacobian (or an approximation to it) to be used in the solver. In the future, we will compute this automatically from the expression tree (see #197) , but for now this allows the user to input their own Jacobian which has been computed in advance.

rtimms:
# Description
Adds `jacobian` attribute to model which defaults to `None`. The user can supply their own jacobian and either set this using `model.jacobian` and solve the model, or directly supply the jacobian to `solver.integrate`. In the future `model.jacobian` will be calculated automatically during discretisation by differentiating the tree (see #197) 

Fixes #223 

## Type of change

# Key checklist:

## Further checks: 


pybamm/solvers/ode_solver.py
```diff
@@ -63,7 +72,9 @@ def integrate(self, derivs, y0, t_eval, events=None, mass_matrix=None):
         events : method, optional
             A function that takes in t and y and returns conditions for the solver to
             stop
-        mass_matrix : :class:`pybamm.Matrix`
+        mass_matrix : array_like, optional
             The (sparse) mass matrix for the chosen spatial method.
+        jacobian : method, optional
+            A function that takes in t and y and returns the Jacobian
```
tinosulzer: are there any cases where `mass_matrix` depends on `t` and `y`?

rtimms: I can't think of any off the top of my head, but maybe there are some cases of this. Will probably be good to make `model.mass_matrix`a method sometime in the future as you suggest

pybamm/solvers/scikits_dae_solver.py
```diff
@@ -14,6 +14,11 @@
     if scikits_odes_spec is not None:
         scikits_odes = importlib.util.module_from_spec(scikits_odes_spec)
         scikits_odes_spec.loader.exec_module(scikits_odes)
+        # NOTE: Gives error module 'scikits.odes' has no attribute 'sundials'
+        # if you try to make the Jacobian class using
+        # scikits_odes.sundials.ida.IDA_JacRhsFunction
+        # but is OK if you import ida here?
+        from scikits.odes.sundials import ida
```
Scottmar93: Weird, do other things like:
```python
import scikits.odes.sundials as sundials
...
sundials.ida
```
also break?

Scottmar93: I have no idea though. Probably need @martinjrobins

pybamm/solvers/scikits_ode_solver.py
```diff
@@ -14,6 +14,11 @@
     if scikits_odes_spec is not None:
         scikits_odes = importlib.util.module_from_spec(scikits_odes_spec)
         scikits_odes_spec.loader.exec_module(scikits_odes)
+        # NOTE: Gives error module 'scikits.odes' has no attribute 'sundials'
+        # if you try to make the Jacobian class using
+        # scikits_odes.sundials.cvode.CV_JacRhsFunction
+        # but is OK if you import cvode here?
+        from scikits.odes.sundials import cvode
```
Scottmar93: again @martinjrobins

pybamm/__init__.py
```diff
@@ -163,6 +163,8 @@ def version(formatted=False):
 from .solvers.scipy_solver import ScipySolver
 from .solvers.scikits_dae_solver import ScikitsDaeSolver
 from .solvers.scikits_ode_solver import ScikitsOdeSolver
+from .solvers.scikits_ode_solver import JacobianFunctionCV
```
martinjrobins: Don't think these functions need to be used outside of scikits_ode_solver, so can probably get rid of this line

martinjrobins:
```suggestion
```

pybamm/__init__.py
```diff
@@ -163,6 +163,8 @@ def version(formatted=False):
...
+from .solvers.scikits_dae_solver import JacobianFunctionIDA
```
martinjrobins:
```suggestion
```

pybamm/solvers/scikits_dae_solver.py
martinjrobins: depends how the scikits.odes ppl have set it up. You might not normally need scikits.odes.sundials.ida, so it might not all be brought in when you import scikits.odes. It might be purposeful (so you only import what you need), or the authors just forgot to import it. I think some submodules of scipy do that as well (can't think offhand which ones)

pybamm/solvers/scikits_ode_solver.py
martinjrobins: I think this is fine (or at least out of our hands)

PR #228, opened 2019-03-28
#203 more binary and unary simplifications, including concatentaions

### LINKED ISSUE #203: implement simplifications for unary operators
similar to binary operators in #201

martinjrobins:
# Description

- expanded binary simplification
- added unary simplifications
- made sure DomainConcatenation simplification works
- for *all* nodes, if the expression evaluates to a constant scalar/vector/matrix, then it is simplified to an appropriate Scalar/Vector/Matrix node

Fixes #203 

## Type of change

# Key checklist:

## Further checks: 

### DIFF
--- a/pybamm/expression_tree/binary_operators.py
+++ b/pybamm/expression_tree/binary_operators.py
@@ -9,6 +9,17 @@ import numbers
 import autograd.numpy as np
 
 
+def is_zero(expr):
+    """
+    Utility function to test if an expression evaluates to a constant scalar zero
+    """
+    if expr.is_constant():
+        result = expr.evaluate_ignoring_errors()
+        return isinstance(result, numbers.Number) and result == 0
+    else:
+        return False
+
+
 class BinaryOperator(pybamm.Symbol):
     """A node in the expression tree representing a binary operator (e.g. `+`, `*`)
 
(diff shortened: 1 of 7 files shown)


pybamm/expression_tree/binary_operators.py
```diff
@@ -240,7 +248,7 @@ def _binary_simplify(self, left, right):
         """ See :meth:`pybamm.BinaryOperator.simplify()`. """
 
         # anything multiplied by a scalar zero returns a scalar zero
-        if left.evaluates_to_value(0) or right.evaluates_to_value(0):
+        if is_zero(left) or is_zero(right):
             return pybamm.Scalar(0)
```
tinosulzer: could also have anything times 1 returns itself

pybamm/expression_tree/binary_operators.py
```diff
@@ -275,15 +283,15 @@ def _binary_simplify(self, left, right):
...
 
         # zero divided by anything returns zero
-        if left.evaluates_to_value(0):
+        if is_zero(left):
             return pybamm.Scalar(0)
 
         # anything divided by zero returns inf
-        if right.evaluates_to_value(0):
+        if is_zero(right):
             return pybamm.Scalar(np.inf)
```
tinosulzer: could also have anything divided by 1 returns itself

pybamm/expression_tree/symbol.py
```diff
@@ -361,28 +380,32 @@ def is_constant(self):
...
         evaluate : evaluate the expression
 
         """
-        return (
-            self.is_constant()
-            and self.evaluates_to_number()
-            and self.evaluate() == value
-        )
+        try:
+            result = self.evaluate(t=0)
```
tinosulzer: I know I started this but it now makes me a bit uncomfortable (avoiding the `t` dependence). Works for now and I can't think of a better way but something to keep an eye out for

martinjrobins: yea, me too :) For the simplifications, I always use is_constant(), so I should be able to remove the `t=0`, but it would be needed for `evaluates_to_number`.

PR #232, opened 2019-03-28
Issue 120 lead acid dae

### LINKED ISSUE #120: Implement lead-acid "full 1D" model (as dae system)
Should give the same results as #119 
1902 01771 pdf 1

tinosulzer:
# Description

Standard one-dimensional model for a lead-acid battery. Tests are on a coarse grid, for the sake of speed.
Fixes #120 

## Type of change

# Key checklist:

## Further checks: 


pybamm/models/submodels/electrode.py
```diff
@@ -0,0 +1,73 @@
...
+            self.initial_conditions = {phi_s: 0}
+            self.variables = {
+                "Negative electrode solid potential": phi_s,
+                "Negative electrode solid current": i_s_n,
+            }
+        elif phi_s.domain == ["positive electrode"]:
+            i_s_p = -param.sigma_p * (1 - eps) ** param.b * pybamm.grad(phi_s)
+            self.algebraic = {phi_s: pybamm.div(i_s_p) + j}
+            self.boundary_conditions = {i_s_p: {"left": 0, "right": current}}
+            self.initial_conditions = {
```
Scottmar93: It probably fine for now, but I'm not sure whether we should put initial conditions for the space only ODEs since their initial state should really be found by their consistency with the initial conditions for c etc. I can't think of a better place to put this initial guess though.

pybamm/models/lead_acid/newman_tiedemann.py
```diff
@@ -0,0 +1,125 @@
...
+        #
+        # Variables
+        #
+        whole_cell = ["negative electrode", "separator", "positive electrode"]
+        #
+        # Variables
+        #
+        # Concentration variables
+        c_e_n = pybamm.Variable(
+            "Negative electrode concentration", domain=["negative electrode"]
```
Scottmar93: Might be slightly clearer to call this "electrolyte concentration in negative electrode". That way it is clear that its in the electrolyte and not the electrode. There isn't any confusion in lead-acid but there might be in li-ion.

PR #235, opened 2019-03-29
Issue 227 post process

### LINKED ISSUE #227: Post-process model variables
**Summary**
 After solving a model, post-process the variables to objects that can be evaluated as functions of `x` and `t`, rather than `t` and `y`, and without having to loop through indices.
For example, a new class `ProcessedVariable` that has an attribute `entries` which is the variables evaluated at all the discrete values of `t` and `y` from the solver, and a method `evaluate(x, t)` that interpolates appropriately.

 **Motivation**
Easier plotting and comparison of variables, e.g. as suggested by @martinjrobins in #209 . Will also help #69 for example.

tinosulzer:
# Description

Post process a variable for easier evaluation, with interpolation. Currently won't work for:
- Symbols that are like grad(u) that get evaluated on edges rather than nodes
- Microscale symbols

To do:
- [x] implement this in the model tests 

Fixes #227 

## Type of change

# Key checklist:

## Further checks: 


pybamm/processed_variable.py
```diff
@@ -0,0 +1,69 @@
...
+
+    """
+
+    def __init__(self, base_variable, t_sol, y_sol, mesh=None, interp_kind="linear"):
+        if mesh is not None and base_variable.domain != []:
+            # Process the discretisation to get x values
+            x_sol = np.concatenate([mesh[dom][0].nodes for dom in base_variable.domain])
+            len_x = len(x_sol)
+        else:
+            # No discretisation provided, or variable has no domain (function of t only)
```
martinjrobins: a user could forget to pass in a mesh for a variable discretised on a domain. Should raise an exception in this case.

PR #258, opened 2019-04-05
Issue 254 clean up spme outputs

### LINKED ISSUE #254: Clean up SPMe outputs
**Summary**
Clean up SPMe outputs so that large expressions are abstracted into submodels

 **Motivation**
Simplify model files and also allows for a more maintainable code if the large expressions are stored within a submodel instead of being in every model.

Scottmar93:
# Description
Cleaned up SPMe output by implementing appropriate equations with submodels. Also written a standard set of output variables which we can ensure is consistent across all model for a specific chemistry.

Fixes #254

## Type of change

# Key checklist:

## Further checks: 

### DIFF
--- a/pybamm/models/base_models.py
+++ b/pybamm/models/base_models.py
@@ -365,6 +365,16 @@ class BaseModel(object):
                         )
                     )
 
+        # Standard Output Variables
+        for output, expression in self._variables.items():
+            if expression is None:
+                raise pybamm.ModelError(
+                    """The standard output variable '{}' which is
+                    required for testing has not been supplied.""".format(
+                        output
+                    )
+                )
+
 
 class LeadAcidBaseModel(BaseModel):
     """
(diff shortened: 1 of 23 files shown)


martinjrobins:
yea, I think users are all going to be interested to calculate and plot different things. You could have a small number of standard output variables. in the `pybamm.LithiumIonBaseModel` or `pybamm.BaseModel` class. Just put the name in the dict equated to `None`, then have the derived classes replace that with the actual expression.

tinosulzer:
Refactor using {"variable": None} dictionary in the base model

Scottmar93:
Thanks for the feedback on this I think it is much better now. I have added standard variables to base_model as suggested. I have also put a check in check_model_wellposedness to checks if these variables have been added. I also went ahead and did the DFN outputs aswell which was pretty simple to do.

pybamm/models/base_models.py
```diff
@@ -365,6 +365,16 @@ def check_well_posedness(self, post_discretisation=False):
...
 
+        # Standard Output Variables
+        for output, expression in self._variables.items():
+            if expression is None:
+                raise pybamm.ModelError(
+                    """The standard output variable '{}' which is
+                    required for testing has not been supplied.""".format(
+                        output
+                    )
+                )
```
tinosulzer: can you add a test for this (to check it fails as expected?)

Issue #253, opened 2019-04-04
Simplify Concatenation (and subsequent bug)


tinosulzer:
**Describe the bug**
The model wasn't being simplified as no simplification was implemented for a `NumpyConcatenation`, when this is added there is a dimension error with matrices being multiplied. It seems that the original tree in the model was
```
Matrix (9,8) * (vector (8,) * Matrix(8,9)  ---> Matrix(9,8) * Matrix(8,9) ---> Matrix(9,9)
```
but this gets changed to 
```
(Matrix (9,8) * vector (8,)) * Matrix(8,9)  ---> Vector(9,) * Matrix(8,9) ---> Error
```

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to branch `issue-253-simplify-bug`
2. Run `python3 -m tests.test_models.test_lead_acid.test_lead_acid_newman_tiedemann`
3. ipdb starts in `MatrixMultiplication.evaluate()`, type `self.children[0].evaluate(t,y).shape` (returns (9,)), `self.children[0].evaluate(t,y).shape` (returns (8,9))

martinjrobins:
@tinosulzer, do you mean the original tree was

```
Matrix (9,8)   ( vector (8,) @ Matrix(8,9) )
```

But `vector (8,) @ Matrix(8,9)` gives `vector(9,)`?

tinosulzer:
Sorry that wasn't clear, there is a mix of matrix multiplications and elementwise multiplications:
original:
```
Matrix (9,8) @ (vector (8,) * Matrix(8,9))  ---> Matrix(9,8) @ Matrix(8,9) ---> Matrix(9,9)
```
after simplification:
```
(Matrix (9,8) @ vector (8,)) * Matrix(8,9)  ---> Vector(9,) * Matrix(8,9) ---> Error
```

martinjrobins:
but `vector (8,)` is a row vector, so  `vector (8,) * Matrix(8,9)`  shouldn't work? should be `vector (8,1)`?

```
In [1]: import numpy as np

In [2]: v = np.ones(8)

In [3]: m = np.ones((8,9))

In [4]: v*m
---------------------------------------------------------------------------
ValueError    Traceback (most recent call last) <ipython-input-4-c1d8335750dc> in <module>
----> 1 v*m

ValueError: operands could not be broadcast together with shapes (8,) (8,9)
```

martinjrobins:
[...]

PR #259, opened 2019-04-08
Issue 209 example model

### LINKED ISSUE #209: example notebook: Run a pre-defined model
see #199

martinjrobins:
# Description

added an example notebook for the SPM model

Fixes #209 

## Type of change

# Key checklist:


examples/notebooks/models/SPM.ipynb
```diff
@@ -43,33 +43,451 @@
...
+   "outputs": [],
+   "source": [
+    "model = pybamm.lithium_ion.SPM()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The model object is a subtype of [`pybamm.BaseModel`](https://pybamm.readthedocs.io/en/latest/source/models/base_model.html), and contains all the equations that define this particular model. For example, the `rhs` dict contained in `model` has a dictionary mapping variables such as $c_n$ to the equation representing its rate of change with time (i.e. $\\frac{c_n}{dt}$). We can see this explicitly be visualising this entry in the `rhs` dict:"
```
tinosulzer:
```suggestion
    "The model object is a subtype of [`pybamm.BaseModel`](https://pybamm.readthedocs.io/en/latest/source/models/base_model.html), and contains all the equations that define this particular model. For example, the `rhs` dict contained in `model` has a dictionary mapping variables such as $c_k$ to the equation representing its rate of change with time (i.e. $\partial c_k /\partial t$). We can see this explicitly be visualising this entry in the `rhs` dict:"
```

examples/notebooks/models/SPM.ipynb
```diff
@@ -43,33 +43,451 @@
...
+   "source": [
+    "disc = pybamm.Discretisation(mesh, model.default_spatial_methods)\n",
+    "disc.process_model(model)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "After this stage, all the equations in `model` have been discretised into purely linear algebra expressions that are ready to be evaluated within a time-stepping loop of a given solver. Once again we use the default ODE solver for this model"
```
tinosulzer: We could show the (simplified) discretised model here?

martinjrobins: yup, I've put that in now. The `simplify()` didn't work due to dimension problems, but after #253 is resolved I'll put in the simplify.

PR #265, opened 2019-04-09
Issue 175 indefinite integral

### LINKED ISSUE #175: Indefinite integral
Implementing something like F(x) = int_0^x f(x) dx (a bit tricky with finite volumes, dealing with all four cases of whether the integrand / resulting integral is to be evaluated on the nodes or on the edges. 

Might be easier to implement as solving an ODE in x: dF/dx = f(x).

Scottmar93:
# Description
Implements the indefinite integral for the cases which we require (i.e. integrate current to get potential). 

Fixes #175

## Type of change

# Key checklist:

## Further checks: 

### DIFF
--- a/pybamm/spatial_methods/finite_volume.py
+++ b/pybamm/spatial_methods/finite_volume.py
@@ -389,15 +389,6 @@ class FiniteVolume(pybamm.SpatialMethod):
         n = submesh.npts
         sec_pts = len(submesh_list)
 
-        # # case input i, output phi
-        # # note we have added a row of zeros at top for F(0) = 0
-        # du_e = submesh.d_edges
-        # du_entries = [du_e] * n
-        # offset = -np.arange(1, n + 1, 1)
-        # sub_matrix = diags(du_entries, offset, shape=(n + 1, n))
-        # matrix = kron(eye(sec_pts), sub_matrix)
-
-        # case input i, output phi
         # note we have added a row of zeros at top for F(0) = 0
         du_n = submesh.d_nodes
         du_entries = [du_n] * (n - 1)
--- a/pybamm/spatial_methods/finite_volume.py
+++ b/pybamm/spatial_methods/finite_volume.py
@@ -342,6 +342,71 @@ class FiniteVolume(pybamm.SpatialMethod):
 
         return pybamm.Vector(vector)
 
+    def indefinite_integral(self, domain, symbol, discretised_symbol):
+        """Implementation of the indefinite integral operator. The
+        input discretised symbol must be defined on the internal mesh edges.
+        See :meth:`pybamm.BaseDiscretisation.indefinite_integral`
+        """
+
+        if not symbol.has_gradient_and_not_divergence():
+            raise pybamm.ModelError(
+                "Symbol to be integrated must be valid on the mesh edges"
+            )
+
+        # Calculate integration matrix
+        integration_matrix = self.indefinite_integral_matrix(domain)
+
+        # Don't need to check for spherical domains as spherical polars
+        # only change the diveregence (symbols here have grad and no div)
+        out = integration_matrix @ discretised_symbol
+
+        out.domain = domain
+
+        return out
+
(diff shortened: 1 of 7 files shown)


pybamm/spatial_methods/finite_volume.py
```diff
@@ -342,6 +342,71 @@ def definite_integral_vector(self, domain):
 
         return pybamm.Vector(vector)
 
+    def indefinite_integral(self, domain, symbol, discretised_symbol):
+        """Implementation of the indefinite integral operator. The
+        input discretised symbol must be defined on the internal mesh edges.
+        See :meth:`pybamm.BaseDiscretisation.indefinite_integral`
+        """
+
+        if not symbol.has_gradient_and_not_divergence():
```
martinjrobins: we've got a lot of these little functions in `pybamm.Symbol` now. Might be worth consolidating them, for example into:
```python
def has_node_types(self, list_of_types):
   # implementation here
``` 

this is an aside, not for this issue...

pybamm/spatial_methods/finite_volume.py
```diff
@@ -342,6 +342,71 @@ def definite_integral_vector(self, domain):
...
+            The finite volume integral vector for the domain
+        """
+
+        # Create appropriate submesh by combining submeshes in domain
+        submesh_list = self.mesh.combine_submeshes(*domain)
+        submesh = submesh_list[0]
+        n = submesh.npts
+        sec_pts = len(submesh_list)
+
+        # # case input i, output phi
```
martinjrobins: this looks like it is no longer needed?

tests/test_spatial_methods/test_finite_volume.py
```diff
@@ -854,73 +854,133 @@ def test_indefinite_integral(self):
...
+            disc.process_symbol(phi_integral)
+
+        grad_and_div = pybamm.div(pybamm.grad(phi))
+        phi_integral = pybamm.IndefiniteIntegral(grad_and_div, x)
+        disc.set_variable_slices([phi])
+
+        with self.assertRaisesRegex(pybamm.ModelError, "integrated"):
+            disc.process_symbol(phi_integral)
+
+        # # microscale variable
```
martinjrobins: this looks like it is no longer needed?

pybamm/spatial_methods/finite_volume.py
Scottmar93: Yeah that's a good idea, would be very helpfully in quite a few places in the discretisation and spatial methods

tests/test_spatial_methods/test_finite_volume.py
```diff
@@ -854,73 +854,112 @@ def test_indefinite_integral(self):
             "positive particle": pybamm.FiniteVolume,
         }
         disc = pybamm.Discretisation(mesh, spatial_methods)
-        # lengths
-        ln = mesh["negative electrode"].edges[-1]
 
-        # macroscale variable
-        var = pybamm.Variable("var", domain=["negative electrode", "separator"])
+        # input a phi, take grad, then integrate to recover phi approximation
```
tinosulzer: could you add a comment here that this is the only way to test because the integrand has to have a grad?

tests/test_spatial_methods/test_finite_volume.py
```diff
@@ -854,73 +854,112 @@ def test_indefinite_integral(self):
...
+        phi = pybamm.Variable("phi", domain=["negative electrode", "separator"])
+        i = pybamm.grad(phi)  # create test current (variable on edges)
+
         x = pybamm.SpatialVariable("x", ["negative electrode", "separator"])
-        integral_eqn = pybamm.IndefiniteIntegral(var, x)
-        disc.set_variable_slices([var])
-        integral_eqn_disc = disc.process_symbol(integral_eqn)
+        phi_integral = pybamm.IndefiniteIntegral(i, x)
```
tinosulzer: maybe change `phi_integral` to `int_grad_phi` (here and elsewhere)?

Issue #264, opened 2019-04-08
more simplifications/optimisations


tinosulzer:
Dumping some thoughts on more things that could be slowing down the evaluation of the tree and that we should try to optimise:
- [x] reshapes in `Multiplication` (see test and timings [here](https://github.com/pybamm-team/PyBaMM/blob/3007362a19efc0263cab8df71e4fa28296562139/tests/test_expression_tree/test_matrix.py#L101)) - PR #268 
- [ ] concatenations (some ideas [here](https://stackoverflow.com/questions/38470264/numpy-concatenate-is-slow-any-alternative-approach/38470331) about creating the array of the right size first? though we are doing this already for domain concatenation)
- [x] broadcast - use vector multiplication instead #381 
- [x] functions that are independent of input (e.g. [constant current function](pybamm/parameters/standard_current_functions/constant_current.py)) - PR #266 
- [x] repeated operations - issue #267 
- [x] `BoundaryValue(Broadcast(child))` simplifies to just `child`
- [x] experiment with writing out the expression tree as a python function #332

tinosulzer:
Toy problem with simplifying (m1@(v1*(m2@v2)). 
Case 2: v1 is a state vector
Case 3: v1 is a constant vector
```
In [7]: expr2.render()                                                                                                                                                                                      
@
├── Matrix (300, 299)
└── *
    ├── y[0:299]
    └── @
        ├── Matrix (299, 300)
        └── y[0:300]

In [8]: expr2simp.render()                                                                                                                                                                                  
@
├── @
│   ├── Matrix (300, 299)
│   └── *
│       ├── y[0:299]
│       └── Matrix (299, 300)
└── y[0:300]

In [9]: %timeit expr2.evaluate(y=np.ones(300))                                                                                                                                                              
54.8 µs ± 4.82 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [10]: %timeit expr2simp.evaluate(y=np.ones(300))                                                                                                                                                         
1.7 ms ± 71.9 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [11]: expr3.render()                                                                                                                                                                                     
@
├── Matrix (300, 299)
└── *
    ├── Vector of length 299
    └── @
        ├── Matrix (299, 300)
        └── y[0:300]

In [12]: expr3simp.render()                                                                                                                                                                                 
@
├── Matrix (300, 300)
└── y[0:300]

[...]

PR #266, opened 2019-04-10
#264 simplify constant function

tinosulzer:
# Description

Add functionality to `parameter_values.process_symbol` so that `FunctionParameter` can get converted to a `Scalar` if a non-callable function is passed. This is useful for constant current as the simplified tree will have fewer broadcasts and concatenations. The name `FunctionParameter` could now be a bit misleading though.
Fixes part of #264 (not whole issue)

## Type of change

# Key checklist:

## Further checks: 

### DIFF
--- a/pybamm/parameters/parameter_values.py
+++ b/pybamm/parameters/parameter_values.py
@@ -181,13 +181,21 @@ class ParameterValues(dict):
         elif isinstance(symbol, pybamm.FunctionParameter):
             new_child = self.process_symbol(symbol.children[0])
             function_name = self.get_parameter_value(symbol)
-            function = pybamm.Function(pybamm.load_function(function_name), new_child)
+            function = pybamm.load_function(function_name)
+            # It's possible to define a FunctionParameter with a function name that
+            # just returns a scalar (i.e. isn't really a function). For example, the
+            # constant current function is just equal to 1.
+            if not callable(function):
+                # If not callable then 'function' is a number and we return a Scalar
+                return pybamm.Scalar(function, name=function_name)
+            # Otherwise proceed as expected
+            function_symbol = pybamm.Function(function, new_child)
             if symbol.diff_variable is None:
-                return function
+                return function_symbol
             else:
                 # return differentiated function
                 new_diff_variable = self.process_symbol(symbol.children[0])
-                return function.diff(new_diff_variable)
+                return function_symbol.diff(new_diff_variable)
 
         elif isinstance(symbol, pybamm.Scalar):
             # update any Scalar nodes if their name is in the parameter dict (no error)
--- a/pybamm/parameters/parameter_values.py
+++ b/pybamm/parameters/parameter_values.py
@@ -5,9 +5,8 @@ from __future__ import absolute_import, division
 from __future__ import print_function, unicode_literals
 import pybamm
 
-import copy
 import pandas as pd
-from inspect import signature
+import copy
 
 
 class ParameterValues(dict):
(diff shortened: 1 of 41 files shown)


pybamm/parameters/parameter_values.py
```diff
@@ -181,13 +181,21 @@ def process_symbol(self, symbol):
         elif isinstance(symbol, pybamm.FunctionParameter):
             new_child = self.process_symbol(symbol.children[0])
             function_name = self.get_parameter_value(symbol)
-            function = pybamm.Function(pybamm.load_function(function_name), new_child)
+            function = pybamm.load_function(function_name)
+            # It's possible to define a FunctionParameter with a function name that
+            # just returns a scalar (i.e. isn't really a function). For example, the
+            # constant current function is just equal to 1.
+            if not callable(function):
```
martinjrobins: not sure I understand this. The function
```python
def return_a_constant():
    return 1
```
is still callable, isn't it?

martinjrobins: couldn't you just call `pybamm.Symbol.is_constant()` on the FunctionParameter node?

tinosulzer: the issue is that the function (the function that we don't know whether it's callable or not) is only called by `Function.evaluate()`, so while we are building the tree we don't know whether the function depends on y or t. In the expression tree, `Function` is a node with a child (`t` in the case of current), regardless of what its function does. 

The function 
```python3
def return_a_constant(t):
    return 1
```
 is callable, but I don't know a way of checking that it always returns a constant without either using a flag or testing for an arbitrary number of values of t. Is there something in the function signature that could tell us?

martinjrobins: but we do know if the function depends on y or t. This FunctionParameter depends on t:

```python
f = pybamm.FunctionParameter('depends_on_t', pybamm.t)
```

This one depends on the parameter `p`:

```python
p = pybamm.Parameter('p')
f = pybamm.FunctionParameter('depends_on_p', p)
```

The only way the function can depend on either y or t is if `pybamm.Variable` or `pybamm.t` are in the expression tree below the `FunctionParameter`

martinjrobins: actually, looking at the `simplify` function for `pyamm.Function`, this simplification should already be taking place??

tinosulzer: But the Function
```python3
f = pybamm.Function("return_a_constant", pybamm.t)
```
depends on t in the "expression tree" sense, but when evaluated it will be equal to 1 regardless of the value of t. But `f.is_constant()` is false, since the expression tree thinks that f depends on t. In this case, we should have just defined `f=Scalar(1)`, but we don't know that when we are creating the model

martinjrobins: ahhh, ok. I get it now. That is a tricky one. 

you could use the inspect module (https://docs.python.org/3/library/inspect.html#introspecting-callables-with-the-signature-object).

The constant function could then look like this:
```
def constant():
    return 1
```

You can use the inspect module to count the number of arguments of the given function. If its 0, then you can simplify, if its 1, then you can't, if its > 1 then you raise an error. What do you think? 

The trouble with what you have done, is that say the user makes a mistake with their function and it not callable (when it should be), then pybamm silently simplifies it away.....

tinosulzer:
Updated using signature to inspect the function

tinosulzer:
ah so FunctionParameter should always be processed to a Function, which then gets simplified to a Scalar, rather than processing to a Scalar straight away?

PR #268, opened 2019-04-10
Issue 264 matrix multiplication

tinosulzer:
# Description

Slightly modify `simplify_multiplication_division` to ensure that simplified tree is faster than the original tree:
- reverse the order of `fold_multiply` so that operations are performed in optimal order (but still simplified if constant)
- Catch sparse matrices as being constant

Fixes part of #264 

## Type of change

# Key checklist:

## Further checks: 

### DIFF
--- a/pybamm/expression_tree/binary_operators.py
+++ b/pybamm/expression_tree/binary_operators.py
@@ -115,20 +115,20 @@ def simplify_addition_subtraction(myclass, left, right):
         ret = None
         if len(array) > 0:
             ret = array[0]
-            for child, t in zip(array[1:], types[1:]):
-                if t == pybamm.Addition:
+            for child, type in zip(array[1:], types[1:]):
+                if type == pybamm.Addition:
                     ret += child
                 else:
                     ret -= child
         return ret
 
     # can reorder the numerator
-    (constant, nonconstant, constant_t, nonconstant_t) = partition_by_constant(
+    (constant, nonconstant, constant_types, nonconstant_types) = partition_by_constant(
         numerator, numerator_types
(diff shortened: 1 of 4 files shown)


pybamm/expression_tree/binary_operators.py
```diff
@@ -115,20 +115,20 @@ def fold_add_subtract(array, types):
         ret = None
         if len(array) > 0:
             ret = array[0]
-            for child, t in zip(array[1:], types[1:]):
-                if t == pybamm.Addition:
+            for child, type in zip(array[1:], types[1:]):
```
martinjrobins: don't think we should use `type` as a variable name, as it is also a built-in function in python

pybamm/expression_tree/binary_operators.py
```diff
@@ -289,16 +306,19 @@ def fold_multiply(array, types=None):
...
-            ret = array[0]
             if types is None:
+                ret = array[0]
                 for child in array[1:]:
                     ret *= child
             else:
-                for child, t in zip(array[1:], types[1:]):
-                    if t == pybamm.MatrixMultiplication:
-                        ret = ret @ child
+                # work backwards through 'array' and 'types' so that multiplications
```
martinjrobins: nice :)

pybamm/expression_tree/binary_operators.py
tinosulzer: ah ok, will change to `typ` (was finding `t` confusing as it's also time)

tinosulzer:
Yep seeing good speed-up now!

PR #281, opened 2019-04-15
#280 faster update parameters

### LINKED ISSUE #280: better parameter update
Related to #161 

We don't need to rebuild the tree each time we want to change the parameters, just walk through the tree and change the `value` of relevant `Symbol` nodes.

tinosulzer:
# Description

Update parameters inplace rather than creating a new model
Fixes #280 

## Type of change

# Key checklist:

## Further checks: 


pybamm/parameters/parameter_values.py
```diff
@@ -78,7 +78,7 @@ def get_parameter_value(self, parameter):
         """
         return self[parameter.name]
 
-    def process_model(self, model):
+    def process_model(self, model, processing="process"):
```
martinjrobins: need to explain `processing` in docstring

PR #294, opened 2019-04-17
Issue 287 inf norm

### LINKED ISSUE #287: Use inf norm instead of l2 norm
For convergence testing

tinosulzer:
# Description

Use inf norm instead of l2 norm in tests, get proper convergence (separating bulk from boundaries)
Fixes #287 

## Type of change

# Key checklist:

## Further checks: 


tests/convergence/test_spatial_methods/test_finite_volume.py
```diff
@@ -0,0 +1,272 @@
...
+
+        # Get errors
+        ns = 10 * 2 ** np.arange(6)
+        errs = {n: get_error(int(n)) for n in ns}
+        # expect quadratic convergence everywhere
+        err_norm = np.array([np.linalg.norm(errs[n], np.inf) for n in ns])
+        rates = np.log2(err_norm[:-1] / err_norm[1:])
+        np.testing.assert_array_less(1.99 * np.ones_like(rates), rates)
+
+    def test_spherical_div_convergence_linear(self):
```
Scottmar93: Maybe @jamiemfoster knows if we lose quadratic convergence for finite volumes on spherical operators. If so maybe the change of variables: u= r * c in the particles might be worth it? (this only works when the diffusion coefficient in the particle is constant though)

jamiemfoster:
Is this in reference to the solution of the spherical diffusion equation
with a symmetry BC at the core and a state-dependent flux condition at the
surface? I.e., the solid stransport problem? If so, then an analogous
finite volume treatment to the one we were discussing earlier on the week
would yield second order convergence internally, but would be limited to a
global first order rate owing to the extrapolation needed to impose the
flux BC at the surface. Does that make sense?

-------------------------------------------------
Jamie M. Foster,
School of Mathematics and Physics,
University of Portsmouth,
Portsmouth, Hampshire, UK (PO1 2UP).
Phone: +44 02392 846 356 (internal: 6356)
Email: jamie.michael.foster@gmail.com
URL: http://ms.mcmaster.ca/~jmfoster/
-------------------------------------------------

tests/convergence/test_spatial_methods/test_finite_volume.py
jamiemfoster: Is this in reference to the solution of the spherical diffusion equation with a symmetry BC at the core and a state-dependent flux condition at the surface? I.e., the solid transport problem? If so, then an analogous finite volume treatment to the one we were discussing earlier on the week would (I believe) yield second order convergence internally, but would be limited to a global first order rate owing to the extrapolation needed to impose the flux BC at the surface.

I don't think that there are any nuances to do with the spherical operators that would cause you to drop to first order on the internal points. 

Another note about this is that you may well lose perfect conservation in the spherical case. For example, if I write at the internal points that
ci_dot = 1/ri^2 1/delta ( Fi+1/2 - Fi+1/2 )
Then it is not true that Sum Vi ci = constant, where Vi is the volume of the ith shell. To get round this, we can either do a more tailored scheme (e.g. control volumes) or make mapping to slab diffusion as @Scottmar93 suggests. The trouble with the latter is that it gets messy if we state dependent diffusion, which we probably do for most materials.

tinosulzer: It's a bit simpler than solving a spherical diffusion equation. I'm using a known function of x and seeing how its divergence converges to the real solution (flux is known exactly). But for some forms of the flux I get bad behaviour near `r=0`. If I test that 
```
div( r**2 * sin(r) ) == 4*r*sin(r) - r**2*cos(r)
```
then the inf norm [converges quadratically](https://github.com/pybamm-team/PyBaMM/blob/cd996f357ba0d31e2a9ad4ae3c059cedbd13ae27/tests/convergence/test_spatial_methods/test_finite_volume.py#L138) as I double the mesh size, but if I test 
```
div( r*sin(r) ) == 3*sin(r) + r*cos(r)
```
then the convergence is only [linear](https://github.com/pybamm-team/PyBaMM/blob/cd996f357ba0d31e2a9ad4ae3c059cedbd13ae27/tests/convergence/test_spatial_methods/test_finite_volume.py#L177), with the worst behaviour closest to r=0. Testing 
```
div( sin(r) ) == (2/r)*sin(r) + cos(r)
```
is even worse, the numerical solution tends to 4 as r-->0, instead of tending to 3 as you'd expect, independently of grid size

Good point on conservation, we'll have to keep an eye on that

pybamm/models/reaction_diffusion.py
```diff
@@ -27,10 +28,15 @@ def __init__(self):
         #
         # Submodels
         #
-        # Load reaction flux from submodels
-        j = pybamm.interface.homogeneous_reaction(whole_cell)
+        # Use cos(x) as the source term
+        x = pybamm.SpatialVariable("x", domain=whole_cell)
+        j = pybamm.Function(np.cos, x)
+        # Use uniform porosity
+        epsilon = pybamm.Scalar(1)
```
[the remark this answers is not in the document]
tinosulzer: Ok, will think about adding this test

PR #297, opened 2019-04-17
#295 allow mat-muls on the denominator

### LINKED ISSUE #295: allow mat-muls on denominator during simplification
see https://github.com/pybamm-team/PyBaMM/issues/271#issuecomment-484137582

martinjrobins:
# Description

Fixes #295 

removed timing test in test_matrix.py as was failing occasionally. changed to simply checking that the simplified expression was identical (according to the id)

## Type of change

# Key checklist:

## Further checks: 


martinjrobins:
> looks good but I'm a bit nervous about the behaviour of `__mul__` with `None` down the line - are we sure this won't allow the user to do weird things?

Yea, you are probably right that this could be dangerous, I've now removed the mult and div by None

PR #300, opened 2019-04-18
Issue 282 discrete

### LINKED ISSUE #282: Return discretised model (not in place)
For convergence tests, we have to make a deepcopy of the model each time as we cannot re-discretise a model that has already been discretised.

We could avoid this by having the option of returning a discretised model when discretising, leaving the original model unaffected

tinosulzer:
# Description

Discretise a model not in place, to avoid having to make copies in convergence testing
Fixes #282 

## Type of change

# Key checklist:

## Further checks: 


tinosulzer:
ok, makes sense, fixed

tinosulzer:
Zero extra time cost (I think), but some memory cost as we have to create new model object? I agree that returning a new model by default is more obvious, as well as being consistent with the behaviour of `process_symbol`.

Note that for `parameter_values`, I think the default should still be in place, because we have the `update_model` function, which is much quicker than `process_model`. We can't write a similar update function for the discretisation, because we can't tell from the discretised model what the original model was supposed to be (e.g. is a matrix-vector multiplication a gradient or a divergence?)

PR #316, opened 2019-04-26
Issue 271 node to edge

### LINKED ISSUE #271: node to edge and boundary value functions
Use matrix operations instead of functions to implement these. 

In particular, we should be able to define a single matrix that calculates a boundary value and concatenates it to the existing state vector

rtimms:
# Description

`compute_diffusivity`  and `boundary_value` are now matmuls in `finite_volume`

Fixes #271 

## Type of change

# Key checklist:

## Further checks: 


tests/unit/test_spatial_methods/test_finite_volume.py
```diff
@@ -76,7 +76,7 @@ def test_extrapolate_left_right(self):
         # check linear variable extrapolates correctly
         linear_y = micro_submesh[0].nodes
         y_surf = micro_submesh[0].nodes[-1] + micro_submesh[0].d_nodes[-1] / 2
-        self.assertEqual(surf_eqn_disc.evaluate(None, linear_y), y_surf)
+        self.assertAlmostEqual(surf_eqn_disc.evaluate(None, linear_y)[0], y_surf)
```
tinosulzer: sorry, I'm being slow - why is this now a different shape? I assume it's not an issue, since all the other tests pass

rtimms: for some reason this wasn’t exactly equal in this test (was out in like the 16th decimal or something). changing to an almost equal wasnt happy about it being an array. surf now returns an array as it comes from a matmul. maybe it should return just a number, but this doesn’t seem to be an issue elsewhere

tinosulzer: Ah ok, yeah I think I have seen `AlmostEqual` being stricter about shapes than `Equal`

PR #324, opened 2019-04-29
#319 cache children

### LINKED ISSUE #319: Cache children
Predefine children, such as `left` and `right` in `BinaryOperator`, to get around `self.children` being slow

tinosulzer:
# Description

As discussed this morning, pre-assign children to get around slow anytree lookup
Fixes #319 

## Type of change

# Key checklist:

## Further checks: 

### DIFF
--- a/pybamm/expression_tree/concatenations.py
+++ b/pybamm/expression_tree/concatenations.py
@@ -30,6 +30,7 @@ class Concatenation(pybamm.Symbol):
         else:
             domain = []
         super().__init__(name, children, domain=domain)
+        self.cached_children = self.children
 
     def get_children_domains(self, children):
         # combine domains from children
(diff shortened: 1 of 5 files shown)


pybamm/expression_tree/concatenations.py
```diff
@@ -30,6 +30,7 @@ def __init__(self, *children, name=None, check_domain=True):
         else:
             domain = []
         super().__init__(name, children, domain=domain)
+        self.cached_children = self.children
```
martinjrobins: is there a reason to only do this for binary operators and concatenations? Could simply do this for the base class `Symbol`, and then overload the `children` attribute to return `cached_children`?

tinosulzer: Ok, no reason, just hadn't thought of doing it that way. So if I understand correctly, we could just add a line in `Symbol.__init__` that does something like the following?
```
self.children = anytree.NodeMixin(self, children)
```

tinosulzer:
@martinjrobins , I have moved `cached_children` to `Symbol.__init__`, but couldn't figure out how to overload the `children` property while still being able to call `NodeMixin.children` initially. If you have a good idea how to do this, could you take over the issue?

tinosulzer:
great, thanks - good to know how to do that! Merging ...

Issue #315, opened 2019-04-24
Speed up jacobian


tinosulzer:
- [x] Simplify binary operations that involve matrices that are zero (by adapting `is_zero`) - #326 
- [x] Don't create new objects when evaluating the following - #325 
  - `Diagonal` (dense to sparse)
  - `SparseStack` (dense to sparse)
  - other binary operations (sparse to dense)
- [x] allow using `known_evals` dict - #318 
- [x] don't use jacobian in ODE models - the jacobian would be almost full anyway, and inverting the matrix (for implicit time-stepping) is easy, so the jacobian doesn't add anything

After #325 it looks like autograd might be the limiting step. Would there be a way to pre-compute the gradients?

tinosulzer:
`PendingDeprecationWarning` is coming from sparse - numpy operation (converts sparse matrix to numpy array)

PR #327, opened 2019-04-30
#315 add flag to skip using jacobian for ODE models

tinosulzer:
# Description

Add a flag to the model to allow skipping the use of the jacobian (e.g. for ODE models, where the jacobian is almost full and inverting the matrix during an implicit time step is cheap, so we don't need the jacobian)
Part of #315 

## Type of change

# Key checklist:

## Further checks: 

### DIFF
--- a/pybamm/solvers/ode_solver.py
+++ b/pybamm/solvers/ode_solver.py
@@ -51,10 +51,9 @@ class OdeSolver(pybamm.BaseSolver):
 
         y0 = model.concatenated_initial_conditions
 
-        # Create Jacobian from simplified rhs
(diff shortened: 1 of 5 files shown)


pybamm/solvers/ode_solver.py
```diff
@@ -53,10 +53,15 @@ def eval_event(t, y):
 
         # Create Jacobian from simplified rhs
         y = pybamm.StateVector(slice(0, np.size(y0)))
```
rtimms: should creating the state vector be moved inside the if use_jacobain loop?

Issue #322, opened 2019-04-29
currently converting our sparse jacobian to a dense


martinjrobins:
this is slow!, unclear how to use a sparse jacobian with scikits.odes (see https://github.com/bmcage/odes/issues/77)

martinjrobins:
Looks like we can use a sparse jacobian with any of sundials iterative solvers, out of the box as long as scikits.odes is concerned. 

So, this is implemented as:
if the jacobian evaluate returns a sparse scipy matrix, swap to using the SPGMR (Scaled, Preconditioned, Generalized Minimum Residual) solver. Otherwise assume it is a dense matrix and use the default dense sundials solver

Note that we use NO preconditioning, so this might not be the best approach, comments welcome..!

martinjrobins:
will keep this open for now as this is only implemented for CVODE, not IDA. Would need to get a PR accepted into scikits.odes to do the same for IDA, have asked the maintainer about this and we'll see.

martinjrobins:
yup, the new code is merged in with scikits.odes, so can now use the iterative solvers for both CVODE and IDA

PR #328, opened 2019-05-02
Issue 322 sparse

martinjrobins:
# Description

Fixes half (ode solver) of #322

Enables user to set `linsolver` for scikits_odes_solver. For the `dense` solver, the sparse jacobian changed to a dense array, for the iterative solvers ('spgmr', 'spbcgs', 'sptfqmr') the sparse matrix is used directly.

However, I found that using the iterative solvers generally made the solvers take significantly longer to run, so `dense` is currently left as the default

@rtimms, @tinosulzer: is an iterative solver the approach we want to take for the sparse jacobian? If so, do we need preconditioning? 

Note that I've only done this for the sundials ode solver, as scikits.odes only exposes this functionality for cvode. I would have to edit scikits.odes to get this working for ida

## Type of change

# Key checklist:

## Further checks: 

### DIFF
--- a/pybamm/solvers/scikits_ode_solver.py
+++ b/pybamm/solvers/scikits_ode_solver.py
@@ -27,14 +27,17 @@ class ScikitsOdeSolver(pybamm.OdeSolver):
     tolerance : float, optional
         The tolerance for the solver (default is 1e-8). Set as the both reltol and
         abstol in solve_ivp.
+    linsolver : str, optional
+            Can be 'dense' (= default), 'lapackdense', 'spgmr', 'spbcgs', 'sptfqmr'
     """
 
-    def __init__(self, method="cvode", tol=1e-8):
+    def __init__(self, method="cvode", tol=1e-8, linsolver="dense"):
         if scikits_odes_spec is None:
             raise ImportError("scikits.odes is not installed")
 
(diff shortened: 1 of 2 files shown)


pybamm/solvers/scikits_ode_solver.py
```diff
@@ -72,6 +68,8 @@ def integrate(
             A function that takes in t and y and returns the Jacobian. If
             None, the solver will approximate the Jacobian.
             (see `SUNDIALS docs. <https://computation.llnl.gov/projects/sundials>`).
+        linsolver : method, optional
```
tinosulzer: Type should be "string"

Should there be an option in the base ODE solver class to set `linsolver`? Or does it depend on `self.method`?

martinjrobins: Yes, should be a string, I'll change that.

Yes, at the moment a user can't set `linsolver`. I'll provide an interface for that.

martinjrobins:
> might be that direct solvers are quicker for small problems and iterative solvers scale better?

Yea, that was my understanding (that iterative solvers scale better). Ok, I'll leave the default solver to `dense` for the moment, and provide an option for users to swap to using an iterative method for larger systems.

Issue #334, opened 2019-05-02
Add averaging to processed variable


Scottmar93:
**Summary**
Add a function to average quantities within `ProcessedVariable`. Mesh is already stored in `ProcessedVariable` so this should be a simple addition and it gets rid of a whole bunch of "Average" outputs.  

 **Motivation**
It would be useful to have more averaged variables for testing but I don't want the output_variables list unnecessarily bloated.

tinosulzer:
Issue with this is that the `Integrate` operator needs to be discretised

tinosulzer:
I think the workflow for this should be that the variable to be averaged should be added to the list of variables before the model is built and solved. This can be done in a script, for example:
```python3
model = pybamm.lithium_ion.DFN()
variable = model.variables["example variable"]
model.variables["X-averaged example variable"] = pybamm.x_average(variable)
...
```
Closing but feel free to reopen if different functionality is needed

Issue #333, opened 2019-05-02
Issue exporting fluxes


Scottmar93:
**Describe the bug**
Shape size error when trying to output fluxes and currents.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to SPM.ipynb
2. Run all
3. Scroll to In[17] (below the list of output variables)
4. Add the line:
```python
pybamm.ProcessedVariable(
    model.variables['Positive particle flux'], solver.t, solver.y, mesh=mesh
)
```

Get the following error message:
```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-17-74cbf4a13b3d> in <module>
      7 )
      8 pybamm.ProcessedVariable(
----> 9     model.variables['Positive particle flux'], solver.t, solver.y, mesh=mesh
     10 )

~/PycharmProjects/PyBaMM/pybamm/processed_variable.py in __init__(self, base_variable, t_sol, y_sol, mesh, interp_kind)
     52         # Evaluate the base_variable index-by-index
     53         for idx in range(len(t_sol)):
---> 54             entries[:, idx] = base_variable.evaluate(t_sol[idx], y_sol[:, idx])
     55 
     56         # assign attributes for reference

ValueError: could not broadcast input array from shape (9) into shape (10)
```

tinosulzer:
Will have a look, think I can fix this

tinosulzer:
There is a fundamental issue with evaluating concatenations of fluxes (we don't know how big each part should be, because it depends on the boundary conditions). I think we should change how the boundary conditions are done to fix this, see #335

PR #337, opened 2019-05-04
Issue 248 test spm

### LINKED ISSUE #248: Test the SPM
**Summary**
Add a set of tests for the SPM. Do this in a way which will allow for the tests to be also easily used by other models

 **Motivation**
Testing is good

**Tests**
Appropriate tests should created within standard_model_lithium_ion_output_tests as these should be applicable to all lithium ion models. We should also **include opposite tests for a charge**

Voltage:
- [ ] neg and pos reaction overpotentials < 0 during discharge
- [ ] combined reaction overpotential < 0 during discharge 
- [ ] neg ocp decreases during discharge
- [ ] pos ocp decreases during discharge
- [ ] ocv = pos ocp - neg ocp
- [ ] ocv decreases during discharge
- [ ] v < ocv during discharge 
- [ ] v  = ocv + react over

Particle concentrations and Fluxes:
- [ ] all neg particle concentrations are decreasing during discharge
- [ ] all pos particle concentration are increasing during discharge
- [ ] neg and pos concentration remain between 0 and c_s_n_max/c_s_p_max
- [ ] conservation of lithium in the particles (i.e sum of neg concentrations = sum of pos concentrations) for all time
- [ ] neg surf is less than (or equal to) all other neg particle concentrations for all time during discharge
- [ ] pos surf is greater than (or equal to) all other pos particle concentrations for all time during discharge
- [ ] r average of neg = (1/R_n) int_0^R_n(c_s_n) dr
- [ ] r average of pos = (1/R_p) int_0^R_p(c_s_p) dr
- [ ] there exists concentrations greater than and less than the r-average in both particles
- [ ] neg surf is less then neg r-average during discharge 
- [ ] pos surf is greater than pos r-average during discharge
- [ ] flux in centre of the particle is zero 
- [ ] neg particle flux is negative at surface during discharge 
- [ ] pos particle flux is positive at surface during discharge
- [ ] neg particle flux * L_n * a_n = pos particle flux * L_p * a_p (check the correct relation)

Electrolyte concentration and fluxes:
- [ ] electrolyte concentration is greater than 0
- [ ] conservation of lithium in electrolyte 
- [ ] average lithium-ion concentration in electrolyte is c_e_typ
- [ ] average neg electrolyte concentration is either constant or decreasing during discharge (I'm not 100% this is true for all time, could impose a weaker test that it is less during discharge than the initial concentration)
- [ ] average pos electrolyte concentration is either constant or increasing during discharge (I'm not 100% this is true for all time, could impose a weaker test that it is greater during discharge than the initial concentration)
- [ ] internal boundary fluxes are equal for the neg flux, sep flux, and pos flux symbols (i.e. on L_n)
- [ ] left and right electrolyte fluxes are zero
- [ ] concatenating neg flux, sep flux, and pos flux gives the full electrolyte flux (account for internal boundary values)

Potentials and current:
- [ ] check neg electrode potential is zero on left boundary
- [ ] check electrolyte potential is less than neg electrode potential
- [ ] check pos electrode potential is greater than neg electrode potential
- [ ] check pos electrode potential = terminal voltage on the right boundary
- [ ] check x-av neg exchange current density * L_n * a_n = x-av pos exchange current density * L_p * a_p (check this formula is correct before implementing)
- [ ] check electrode current + electrolyte current = i_typ
- [ ] check the concatenations of currents are equal to the full cell currents i.e (j_n,j_s,j_p) = j

Scottmar93:
# Description

Added a set of general tests on the outputs of SPM, SPMe, LOQS, Newman-tiedeman. I have omitted tests on the fluxes for now as this requires #333. DFN cannot be tested due to an issue exporting the OCV as a processed variable (@tinosulzer I tried merging your latest changes but this didn't resolve the issue). Once these two issues are resolved it should be a simple extension to add the relevant tests. 

I have also commented out many tests involving averages for now instead of adding all of these quantities to the standard output list. Instead, I think it would be better as suggested in #334 to have the ability to average a processed variable. This condenses the output variables list by quite a lot.

Lead-acid composite does not pass the voltage consistency test so maybe I have missed something @tinosulzer? I have just skipped testing composite for now.

EDIT:
I should additionally say that I have fixed algebraic expressions that were wrong in the process of doing this testing. The main one was that I had previously used boundary ocps and overpotentials to determine the constant in the electrolyte potential and positive electrode potential. We actually need to use the averaged versions which required a bit of shifting around of the order things are defined in the models. 

Fixes #248 

## Type of change

# Key checklist:

## Further checks: 


tests/unit/test_models/standard_output_tests.py
```diff
@@ -0,0 +1,448 @@
...
+        )
+        tests.test_all()
+
+    def test_currents(self):
+        tests = CurrentTests(
+            self.model, self.disc, self.solver, self.operating_condition
+        )
+        tests.test_all()
+
+    def test_all(self):
```
tinosulzer: This is a very good way of splitting up the tests! You could remove some redundant code with something like 
```python3
def run_tests(self, class_name):
     tests = class_name(self.model, ...)
     tests.test_all 

def test_all(self):
     self.run_class(VoltageTests)
     etc
```
Also, the two `test_all` functions doing different things might be confusing. Maybe call this one `run_all_output_tests` or something

tests/unit/test_models/standard_output_tests.py
```diff
@@ -0,0 +1,448 @@
...
+        # self.c_e_p_av = self.get_var("Average positive electrolyte concentration")
+
+        # TODO: fix flux bug
+        # self.N_e = self.get_var("Electrolyte flux")
+
+        # self.N_e_n = self.get_var("Negative electrolyte flux")
+        # self.N_e_s = self.get_var("Separator electrolyte flux")
+        # self.N_e_p = self.get_var("Positive electrolyte flux")
+
+        # self.N_e_hat = self.get_var("Reduced cation flux")
```
tinosulzer: (just a comment) - not obvious how we get "Negative electrolyte flux" in the DFN, without hacking it (similarly for "Negative electrolyte potential")

examples/notebooks/models/SPM.ipynb
```diff
@@ -401,16 +481,32 @@
...
+      "\u001b[0;31mValueError\u001b[0m: could not broadcast input array from shape (9) into shape (10)"
```
tinosulzer: Notebook has broken

pybamm/models/submodels/electrolyte_current.py
```diff
@@ -126,7 +126,7 @@ def set_algebraic_system(self, phi_e, c_e, reactions, epsilon=None):
         self.rhs = {}
 
         # Variables
-        # eta_c_av and delta_phi_e_av not defined?
+        # eta_c_av and delta_phi_e_av are not defined`independently
```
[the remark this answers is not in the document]
Scottmar93: We can probably remove them from the list of required standard output parameter but leave them for the SPMe they are useful to have there and probably your composite model. They cannot in general be obtained for DFN or Newman tiedeman. For the constant current case its possible to approximate delta_phi_e using the initial electrolyte resistance (when the concentrations are uniform) and assume its constant but its probably not worthwhile checking for all of that.

tests/unit/test_models/standard_output_tests.py
Scottmar93: Yeah I agree, given that this really isn't an OO bit of code it's probably best to just replace the class by a function. That should cut out alot of unneeded extra code. It's a little out of style with the rest of the tests though which is why I made a class

tests/unit/test_models/standard_output_tests.py
Scottmar93: Ah thank you good to note

tests/unit/test_models/standard_output_tests.py
tinosulzer: I think it's fine to leave `VoltageTests` etc as classes, but you don't have to write a separate function like `test_voltage` for each one. Made a typo above, what I meant was:
```python3
def run_class(self, class_name):
     tests = class_name(self.model, ...) # <--  here class_name can be any class like `VoltageTests`, `ParticleConcentrationTests` etc
     tests.test_all()

def test_all(self):
     self.run_class(VoltageTests)
     etc
```

pybamm/models/submodels/electrolyte_current.py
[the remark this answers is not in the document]
tinosulzer: Yeah seems sensible to remove them from the list of required output parameters then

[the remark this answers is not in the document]
Scottmar93: fixed this within this issue.

tinosulzer:
Looks good, go ahead and merge when this is ready

PR #339, opened 2019-05-05
Remove tests

tinosulzer:
# Description

Remove redundant tests (now that we do the output tests as standard), and change `standard_output_tests` back to object-oriented
@Scottmar93  see if you like this

## Type of change

# Key checklist:

## Further checks: 


tinosulzer:
I think using a class makes it a bit cleaner, and easier to add new tests

PR #356, opened 2019-05-07
#355 add more detail to README

### LINKED ISSUE #355: add section on python virtual environments and cloning repo in README.md

martinjrobins:
# Description

Fixes #355 

## Type of change

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,24 +10,58 @@ Python Battery Mathematical Modelling solves continuum models for batteries, usi
 
 ## How do I use PyBaMM?
 
-PyBaMM comes with a number of [detailed examples](examples/README.md), hosted here on github. In addition, there is a [full API documentation](http://pybamm.readthedocs.io/), hosted on [Read The Docs](readthedocs.io).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,24 +10,58 @@ Python Battery Mathematical Modelling solves continuum models for batteries, usi
...
 pip install --upgrade pip
 ```
 
-Then navigate to the path where you downloaded PyBaMM to, and install both PyBaMM and its dependencies by typing:
+Then navigate to the path where you downloaded PyBaMM to (you will already be in the
+correct location if you followed the instructions above), and install both PyBaMM and
+its dependencies by typing:
 
 ```bash
 pip install .
```
Scottmar93: I know this is from previous commits but will `pip install .` install pybamm so that it in the system path and therefore visible from the notebooks?

martinjrobins: yea, they should. The `-e` just allows them to edit the source, and have the package update to reflect this. If you don't put it in, then any changes you make don't update the package you have installed

PR #357, opened 2019-05-07
Issue 354 quick plot

### LINKED ISSUE #354: create quick plotting routine for model outputs

Scottmar93:
# Description
Added a QuickPlot class which allows the user to quickly plot common output variables which are useful for getting a quick overview of a model. I have implemented this in such a way that you can use it in both a jupyter notebook with widgets or with a slider plot in matplotlib when you are within a script.

Fixes #354 

## Type of change

# Key checklist:

## Further checks: 

### DIFF
--- a/pybamm/quick_plot.py
+++ b/pybamm/quick_plot.py
@@ -1,0 +1,177 @@
+import numpy as np
+import pybamm
+import matplotlib.pyplot as plt
+from matplotlib.widgets import Slider
+
+
+def quick_plot(model, param, mesh, solver):
+    """
+    Generates a quick plot of a subset of key outputs of the model so that the model outputs can be easily assessed.
+
+    Parameters
+    ----------
+    model: :class: pybamm.BaseModel
+        The model to plot the outputs of.
+    mesh: :class: pybamm.Mesh
+        The mesh on which the model solved
+    solver: :class: pybamm.Solver
+        The numerical solver for the model which contained the solution to the model.
+    """
+
+    # obtain parameters values
+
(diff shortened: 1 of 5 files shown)


pybamm/quick_plot.py
```diff
@@ -0,0 +1,227 @@
...
+        self.x_p = np.linspace(self.l_n + self.l_s, 1, 40)
+        self.x = np.linspace(0, 1, 100)
+        self.t = solver.t
+
+        self.set_output_variables(solver.y, model, mesh)
+        self.reset_axis()
+
+    def set_output_variables(self, y, model, mesh):
+        self.c_s_n_surf = pybamm.ProcessedVariable(
+            model.variables["Negative particle surface concentration"], self.t, y, mesh
```
tinosulzer: should the concentration plots be dimensional?

pybamm/quick_plot.py
```diff
@@ -0,0 +1,227 @@
...
+            top=0.92, bottom=0.15, left=0.10, right=0.9, hspace=0.5, wspace=0.5
+        )
+
+        if not testing:
+            plt.show()
+
+    def update(self, val):
+        """
+        Update the plot in self.plot() with values at new time
+        """
```
tinosulzer: from the codecov it looks like this isn't being tested, even though you have a test for it? not sure why

pybamm/quick_plot.py
Scottmar93: I thought about it but decided the numbers are clearer if they aren't, particularly in the particles which will be just be like *20,000 mols/m^3. Whats your thoughts?

PR #358, opened 2019-05-08
Issue 335 explicit dirichlet neumann bcs

### LINKED ISSUE #335: Explicit Dirichlet and Newman boundary conditions
I've had a change of heart and now think we should explicitly state boundary conditions on `Variable` objects (or concatenations of variables). @jamiemfoster  and @ikorotkin  what do you guys think of this, would it help for control volumes?
e.g.
```
model.Dirichlet_bcs = {c: {"left": 0}}
model.Newman_bcs = {c: {"right": 0}}
```
to replace
```
N = pybamm.grad(c)
model.boundary_conditions = {c: {"left": 0}, N: {"right": 0}}
```

**Pros**
- Easier to implement spatial methods: e.g. for finite volumes, we always add a ghost node when discretising the gradient, either to satisfy dirichlet or newman bcs. 
- More in line with UFL (for Fenics/Firedrake)
- Flux-like variables (e.g. current density) always have the same shape (not different shapes depending on the boundary conditions, as is currently the case)

**Cons**
- Slightly more verbose 
- A bit less obvious what the boundary condition should be in some cases (but in most cases, it's no-flux so fine)

tinosulzer:
# Description

Change boundary conditions to always be applied on variables (not fluxes), with explicit "Dirichlet" or "Neumann" flags. @Scottmar93 - note this will affect #360 
Have added some extra functionality to do this: 
- A `BoundaryFlux` node which does the same thing as `BoundaryValue` but returns the flux on the boundary instead of the value. Needed this for `MacInnesCapacitance`.
- Extrapolation of 2D processed variables, we can now test that the current on a boundary is what we expect it to be

To do:
- [x] notebooks
- [x] coverage
- [x] conflicts

Fixes #335 

## Type of change

# Key checklist:

## Further checks: 


pybamm/expression_tree/symbol.py
```diff
@@ -44,10 +44,6 @@ def __init__(self, name, children=[], domain=[]):
         # Set domain (and hence id)
         self.domain = domain
 
-        # useful flags
-        self._has_left_ghost_cell = False
```
martinjrobins: very happy we can get rid of these now! :)

pybamm/models/base_models.py
```diff
@@ -112,10 +112,20 @@ def boundary_conditions(self):
...
         # Convert any numbers to a pybamm.Scalar
         for var, bcs in boundary_conditions.items():
-            for side, eqn in bcs.items():
-                if isinstance(eqn, numbers.Number):
-                    boundary_conditions[var][side] = pybamm.Scalar(eqn)
-
+            for side, bc in bcs.items():
+                if isinstance(bc[0], numbers.Number):
+                    # type is the type of the bc, e.g. "Dirichlet" or "Neumann"
+                    eqn, type = boundary_conditions[var][side]
```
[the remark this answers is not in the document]
tinosulzer: ah, yeah, keep forgetting about that - thanks!

pybamm/models/submodels/particle.py
```diff
@@ -42,24 +42,23 @@ def set_differential_system(self, c, j, broadcast=False):
             )
 
         if c.domain[0] == "negative particle":
-            N = -(1 / param.C_n) * pybamm.grad(c)
-            self.rhs = {c: -pybamm.div(N)}
+            N = -pybamm.grad(c)
+            self.rhs = {c: -(1 / param.C_n) * pybamm.div(N)}
```
Scottmar93: Thanks for correcting this here

pybamm/models/submodels/particle.py
```diff
@@ -42,24 +42,23 @@ def set_differential_system(self, c, j, broadcast=False):
...
             self.algebraic = {}
             self.initial_conditions = {c: param.c_n_init}
+            rbc = -param.C_n * j / param.a_n
```
Scottmar93: What would we do for a mixed boundary condition of form:
  c + a * dcdx = f
Would we rearrange for c or dcdx?

tests/unit/test_models/standard_output_tests.py
```diff
@@ -387,7 +408,7 @@ def test_potential_differences(self):
 
         # TODO: these tests with averages
 
-        np.testing.assert_array_less(-self.phi_s_p.entries, 0)
+        np.testing.assert_array_less(-self.phi_s_p(self.t, self.x_p), 0)
```
Scottmar93: Nice, makes sense to use the Processed Variable interface here.

Scottmar93:
Merged mid-review :(

Issue #191, opened 2019-03-07
Test failures on Python 3.5


fcooper8472:
**Describe the bug**
Some tests are failing on Python 3.5 (but not 3.6 or 3.7).  Examples are in this build:
https://travis-ci.org/tinosulzer/PyBaMM/jobs/503052350

**To Reproduce**
Steps to reproduce the behaviour:
python3.5 tests/test_discretisations/test_discretisation.py

fcooper8472:
This seems to be related to the order of items in a dictionary.  From Python 3.6 onward, the order appears to be the order of insertion, but this is not guaranteed before.

fcooper8472:
There's still one test failure:
```
======================================================================
FAIL: test_process_model_ode (__main__.TestDiscretise)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "tests/test_discretisations/test_discretisation.py", line 384, in test_process_model_ode
    8 * np.ones_like(mesh["negative electrode"].nodes),
  File "/home/fergus/GitRepos/PyBaMM/venv35/lib/python3.5/site-packages/numpy/testing/_private/utils.py", line 896, in assert_array_equal
    verbose=verbose, header='Arrays are not equal')
  File "/home/fergus/GitRepos/PyBaMM/venv35/lib/python3.5/site-packages/numpy/testing/_private/utils.py", line 819, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Arrays are not equal

Mismatch: 44.4%
Max absolute difference: 3.
Max relative difference: 1.5
 x: array([5., 5., 5., 5., 5., 5., 5., 5., 5., 5., 5., 5., 5., 5., 5., 5., 5.,
       5., 5., 5., 5., 5., 5., 5., 5., 5., 5., 5., 5., 5., 5., 5., 5., 5.,
       5., 5., 5., 5., 5., 5., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2.,...
 y: array([2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2.,
       2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2.,
       2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2.,...

----------------------------------------------------------------------
```

I don't understand what's going on here...

[...]

PR #359, opened 2019-05-08
#350 fix notebook failures

### LINKED ISSUE #350: notebook tests failing on travis (cron)

martinjrobins:
# Description

Fixes #350 

1. Added ability to create an empty discretisation (ie no mesh), to use in notebooks that need it
2. removed all magic notebook cells (this was failing when we tried to execute the notebooks from the command line) and replaced with standard python
3. wrap exceptions in "try-except" blocks so don't trip errors

## Type of change

# Key checklist:

## Further checks: 


martinjrobins:
> Maybe we should test notebooks as part of the normal travis routine, given that they don't seem too slow?

Yes, agreed. Confusingly, I've implemented this as part of getting python3.5 to work (#191)

PR #363, opened 2019-05-09
Issue 344 compare lead acid

### LINKED ISSUE #344: Compare lead-acid models
Notebook showing how to compare models, with results from lead-acid models

tinosulzer:
# Description

Edit `QuickPlot` to be more flexible, and add scripts and notebooks for comparing models.
Fixes #344 

## Type of change

# Key checklist:

## Further checks: 


tinosulzer:
Ok, will remove those. 
"Typical current density" was a bug, should have been "Typical current" from the start. (current is in [Amps] and current density is in [A m-2])
We might want to add explicit units to parameters (with the `[unit]` notation) to make these things clearer

pybamm/models/base_models.py
```diff
@@ -537,12 +569,12 @@ def __init__(self):
...
-        icell_dim = pybamm.electrical_parameters.dimensional_current_with_time
-        self.variables.update(
-            {"Total current density": icell, "Total current density [A m-2]": icell_dim}
-        )
+        # Overwrite geometry
+        self.default_geometry = pybamm.Geometry("1D macro")
+
+        # Standard time variable
+        time_scale = pybamm.standard_parameters_lead_acid.tau_discharge
+        self.variables.update({"Time [h]": pybamm.t * time_scale / 3600})
```
Scottmar93: Could we add time in seconds and time in minutes aswell? Also discharge capacity (if its simple to do here, otherwise will create another issue)

pybamm/quick_plot.py
```diff
@@ -11,182 +29,181 @@ class QuickPlot(object):
...
+                self.axis[name] = [x_scaled[0], x_scaled[-1], y_min, y_max]
 
     def plot(self, t):
         """Produces a quick plot with the internal states at time t.
 
         Parameters
         ----------
         t : float
-            Time at which to plot.
+            Dimensional time at which to plot.
```
Scottmar93: If you implement my previous comment, then we should also state that this must be in hours aswell. Unless we want to pass an optional flag to QuickPlot and set `self.time_dimensions`

Scottmar93:
> We might want to add explicit units to parameters (with the [unit] notation) to make these things clearer

I am for this as it makes inputs consistent with outputs

Scottmar93:
This is also failing for the 10^(-5) need to increase to 10^(-4)...

pybamm/models/base_models.py
tinosulzer: Should be easy enough. Discharge capacity in [Ah]?

PR #364, opened 2019-05-09
Issue 191 python3.5

### LINKED ISSUE #191: Test failures on Python 3.5

martinjrobins:
# Description

Fixes #191 

Key issue was that for python 3.5 (and lower) the `dict` class is unordered, so python is free to reorder things as it likes. We were assuming in a few places that the order of our `dicts` *was* fixed, so this was throwing errors. 

other issues were to do with scikits.odes:
1. there was a strange cython bug regarding multiple imports, which occured when we tried to import ida many times
2. need to make sure travis has the correct version of the python header files
3. scikits.odes actually depends on numpy v 1.16, so bumped this version in pybamm's requirements

## Type of change

# Key checklist:

## Further checks: 

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,10 +29,10 @@ matrix:
         #   env:
         #     - PYBAMM_UNIT=true
         #   if: type != cron
-        # - python: "3.5"
-        #   env:
-        #     - PYBAMM_UNIT=true
-        #   if: type != cron
+        - python: "3.5"
+          env:
+            - PYBAMM_UNIT=true
+          if: type != cron
         - python: "3.6"
           env:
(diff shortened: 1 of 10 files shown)


martinjrobins:
> Was there anything else that needed to be done from #191? (rewriting docs?)
I don't think so, the README already states that we support Python 3.5+, so nothing to change there

PR #370, opened 2019-05-10
Issue 360 add a model

Scottmar93:
# Description
Created an Add a Model rst file which details how to create a model class and perform tests etc. I have also added a notebook which goes through the full procedure of nondimensionalising, creating, adding, and solving a model within one notebook so that all the steps can be seen clearly. 

Have also added the ability to pass functions into the parameters values dictionary instead of just a path to a file which contains the function.

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks: 


examples/scripts/create-model.py
```diff
@@ -0,0 +1,125 @@
...
+# 3. State governing equations ---------------------------------------------------------
+R = -k * pybamm.BoundaryValue(c, "left")  # SEI reaction flux
+N = -(1 / L) * D(c) * pybamm.grad(c)  # solvent flux
+dcdt = -(1 / L) * pybamm.div(N)  # solvent concentration governing equation
+dLdt = -V_hat * R  # SEI thickness governing equation
+
+model.rhs = {c: dcdt, L: dLdt}  # add to model
+
+# 4. State boundary conditions ---------------------------------------------------------
+grad_c_left = L * R / D(pybamm.BoundaryValue(c, "left"))  # left bc
```
tinosulzer: Note that you need to do `pybamm.BoundaryValue(D(c), "left")` (because `D` is non-linear) to get the right accuracy for the boundary condition. Might be worth mentioning this in the tutorial.

examples/scripts/create-model.py
```diff
@@ -0,0 +1,125 @@
...
+
+# Extract output variables
+L_dim_out = pybamm.ProcessedVariable(
+    model.variables["SEI thickness"], solver.t, solver.y, mesh
+)
+pybamm.LithiumIonBaseModel().defa
+# plot
+plt.plot(solver.t, L_dim_out(solver.t))
+plt.show()
+
```
tinosulzer: not sure it will work but could try to use `QuickPlot` here as well

pybamm/parameters/parameter_values.py
```diff
@@ -181,14 +164,21 @@ def process_symbol(self, symbol):
...
             return pybamm.Scalar(value, name=symbol.name, domain=symbol.domain)
 
         elif isinstance(symbol, pybamm.FunctionParameter):
             new_child = self.process_symbol(symbol.children[0])
-            function_name = self.get_parameter_value(symbol)
-            function = pybamm.Function(pybamm.load_function(function_name), new_child)
+            function_name = self[symbol.name]
+
+            if callable(function_name):
+                function = pybamm.Function(function_name, new_child)
```
tinosulzer: can you add a test for this?

PR #371, opened 2019-05-10
#353 add notebook on changing parameter, discretisation or solvers

### LINKED ISSUE #353: example notebook: Change the settings (parameters, discretisation or solver)
#199

martinjrobins:
# Description

1. added a "changing parameters" notebook
1. moved the "expression_tree" notebook to a subdirectory

Fixes #353 

## Type of change

# Key checklist:

## Further checks: 


martinjrobins:
> Cheers @martinjrobins , this looks good. Changes from #344 should make the plots nicer

Ok, I'll wait until this has been merged then regenerate the plots

martinjrobins:
@tinosulzer, at the moment the change I'm making to typical current isn't making a huge change in the plots. Is there a better parameter to vary to see a more dramatic change in the output?

tinosulzer:
You might see a bigger difference with a different model like SPMe or DFN. The dimensionless SPM is qualitatively invariant to most parameters, so it would just be quantitative changes (e.g. discharge time is faster), which is hard to see on separate plots. 
The other models should get more spatially heterogeneous results with bigger currents (though this might also be hard to see, as the axis will change scale)

martinjrobins:
I've edited this a bit to more properly compare the different solutions with different parameters. @tinosulzer: I notice that you multiply the terminal voltage by 6 in the lead_acid notebook, I did the same here, but not sure why we need this?

examples/notebooks/change-settings.ipynb
```diff
@@ -1,11 +1,446 @@
...
+    "mesh = pybamm.Mesh(geometry, model.default_submesh_types, model.default_var_pts)\n",
+    "\n",
+    "# discretise the model equations\n",
+    "disc = pybamm.Discretisation(mesh, model.default_spatial_methods)\n",
+    "disc.process_model(model)\n",
+    "\n",
+    "# Solve the model at the given time points\n",
+    "solver = model.default_solver\n",
+    "n = 100\n",
+    "t_eval = np.linspace(0, 2, n)\n",
```
tinosulzer: using something like `t_eval = np.linspace(0, 0.2, n)` will give smoother curves (because we hit terminal voltage / concentration at t=0.2)

examples/notebooks/change-settings.ipynb
```diff
@@ -1,11 +1,446 @@
...
+   ],
+   "source": [
+    "t, y = solver.t, solver.y\n",
+    "new_voltage = pybamm.ProcessedVariable(model.variables[\"Terminal voltage [V]\"], t, y)\n",
+    "plt.plot(time, voltage(t) * 6, lw=2, label=\"Typical current = {}\".format(old_value))\n",
+    "plt.plot(time, new_voltage(t) * 6, lw=2, label=\"Typical current = {}\".format(new_value))\n",
+    "plt.xlabel(\"Time [h]\", fontsize=15)\n",
+    "plt.ylabel(\"Terminal voltage [V]\", fontsize=15)\n",
+    "plt.legend(fontsize=15)\n",
+    "plt.show()"
```
tinosulzer: This is great!

martinjrobins:
> For lead-acid it's because you usually connect six 2V cells in series to make a 12V cell, I don't think that's the case for lithium-ion though. I could remove it from the lead-acid notebook if it makes things confusing

Don't have to remove it, but perhaps add a comment to that effect to make it clear what you are doing.

PR #376, opened 2019-05-10
remove future imports

tinosulzer:
# Description

Remove all unused future imports


Scottmar93:
This is my favourite pull request

tinosulzer:
Was very satisfying

tinosulzer:
I don't know, it's also a bit slower using 3.5 than 3.6

tinosulzer:
Also some of the solvers fail to converge with 3.5 (but work fine on 3.6) - https://travis-ci.org/pybamm-team/PyBaMM/jobs/531078575#L3079
Can be fixed by using fewer grid points or stopping at an earlier time, but it's a bit annoying

PR #388, opened 2019-05-13
#386 rename convergence tests to integration, move model testing in t…

### LINKED ISSUE #386: rename convergence tests to "integration"
and put in standard model tests to integration

martinjrobins:
…here as well

# Description

1. changed dir structure `tests/convergence` to `tests/integration`. 
1. moved model and submodel tests to `tests/integration`
1. moved quick plot tests that involved solving entire models to `tests/integration`
1. updated `add-model.rst` to point to `tests/integration`

Fixes #386 

## Type of change

# Key checklist:

## Further checks: 


tests/unit/test_quick_plot.py
```diff
@@ -1,94 +1,12 @@
 import pybamm
 import unittest
-import numpy as np
 
 
 class TestQuickPlot(unittest.TestCase):
     """
     Tests that QuickPlot is created correctly
     """
```
tinosulzer: did you move this somewhere else?

martinjrobins: Yes, I'll fix this

martinjrobins:
> Thanks @martinjrobins , could you also add unit tests that just create the models and run `model.check_well_posedness`?
> This could be a single file for all the lead-acid models, and a single file for all the lithium-ion models. (I think best practice is to only check coverage from the unit tests?)
> Not sure whether we should keep the submodels as unit or integration, the tests only take ~1 second for each submodel.

Yup, I'll add the well_posedness check for each model. I originally kept submodels in unit tests, but there were still noticably slower than all the other tests to eventually decided to move them to integration

tinosulzer:
> I originally kept submodels in unit tests, but there were still noticably slower than all the other tests to eventually decided to move them to integration

Fair enough, happy to have them in integration then

martinjrobins:
looks like a fair bit of our coverage was due to the model tests. These should be covered with unit tests, will create a separate issue for this

PR #389, opened 2019-05-13
#383 missing output variable warning

### LINKED ISSUE #383: Warnings for missing required variables
Change the model so that we only give a warning if a user doesn't define any of the required variables. Warning something along the lines of letting users know they might not have defined all the variables required to compare with other models.

rtimms:
# Description

Gives a `pybamm.ModelWarning` if a user leaves a variable as `None`. The `None` entries are then removed from the dictionary as part of `model.check_well_posedness()`.

Fixes #383

## Type of change

# Key checklist:

## Further checks: 


rtimms:
This seems to work ok on my machine, but is failing on travis with 
```
  File "/home/travis/build/pybamm-team/PyBaMM/tests/unit/test_models/test_base_model.py", line 327, in test_check_well_posedness_output_variables
    with self.assertWarns(pybamm.ModelWarning):
  File "/opt/python/3.5.6/lib/python3.5/unittest/case.py", line 230, in __enter__
    for v in sys.modules.values():
RuntimeError: dictionary changed size during iteration
```

As far as I can tell this is an issue with using `assertWarns`, but I'm not too sure why `sys.modules` is changing size?

tinosulzer:
> This seems to work ok on my machine, but is failing on travis
Weird, can you see if something like 
```python3
with self.assertWarns(...):
    warnings.warn(...)
```
works?

martinjrobins:
This might be a bug in unittest: https://bugs.python.org/issue29620
 
Perhaps you can try the other version of assertWarns: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertWarns

(the one where the callable is one of the function arguments)

rtimms:
> This might be a bug in unittest: https://bugs.python.org/issue29620
> 
> Perhaps you can try the other version of assertWarns: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertWarns
> 
> (the one where the callable is one of the function arguments)

Thanks @martinjrobins, this works!

PR #393, opened 2019-05-13
Issue 340 fix output tests

### LINKED ISSUE #340: Fix output tests for DFN, Composite and Reaction-Diffusion
- DFN: needs #333 to fix particles 
- Composite: not sure what the problem is
- Reaction-diffusion: shouldn't be a `LeadAcidBaseModel`

tinosulzer:
# Description

Fix output tests for all models except `NewmanTiedemannCapacitance`.
Still a few things to fix for lead-acid, but opening a PR as there is already a fair bit in here
Fixes #340 

## Type of change

# Key checklist:

## Further checks: 


pybamm/models/base_models.py
```diff
@@ -183,6 +186,15 @@ def jacobian(self):
     def jacobian(self, jacobian):
         self._jacobian = jacobian
 
+    @property
+    def use_capacitance(self):
+        return self._use_capacitance
+
+    @use_capacitance.setter
+    def use_capacitance(self, use_capacitance):
```
martinjrobins: where is `use_capacitance` used? I'm guessing that it governs the equations that are put in the model (during __init__). So would calls to this setter after the model is created have any effect? Should this be a read-only option?

tests/integration/test_models/standard_output_comparison.py
```diff
@@ -0,0 +1,181 @@
...
+        self.compare("Average open circuit voltage")
+        # Currents
+        self.compare("Average negative electrode interfacial current density")
+        self.compare("Average positive electrode interfacial current density")
+        # Concentration
+        self.compare("Average electrolyte concentration")
+
+
+class VariablesComparison(BaseOutputComparison):
+    "Compare variables that vary across models"
```
martinjrobins: I don't get this.... doesn't `self.compare` test that the variables are similar across models? I feel I'm missing something obvious

pybamm/models/base_models.py
tinosulzer: Yes, will change this to read-only

tests/integration/test_models/standard_output_comparison.py
tinosulzer: Yeah re-reading now this is bad wording ... the distinction between `VariablesComparison` and `AveragesComparison` is that averages should be the same across models regardless of current, whereas variables will differ a bit and only be approximately equal for certain (low) values of the current. 
May this should just be "compare variables across models"?

tinosulzer: Yeah re-reading now this is bad wording ... the distinction between VariablesComparison and AveragesComparison is that averages should be the same across models regardless of current, whereas variables will differ a bit and only be approximately equal for certain (low) values of the current.
May this should just be "compare variables across models"?

tinosulzer: @martinjrobins - I have changed to "compare variables across models", let me know if this is ok and I will merge

tinosulzer:
Cool, will merge this. The problem was reading `phi_e` before updating it with the composite solution, i.e. using `phi_e` from the leading-order model to calculate the voltage

PR #403, opened 2019-05-14
Issue 329 deepcopy

### LINKED ISSUE #329: Reduce the number of deepcopies in the workflow
These are very slow (usually account for ~60% of total runtime)

- [x] Alternative for setting parameters - #330 
    - Create a new object instead of copying in the cases where no parameters are to be set ([this line](https://github.com/pybamm-team/PyBaMM/blob/cefbaeda1736de76c5275cc68aa2fb6e2fb8aa8a/pybamm/parameters/parameter_values.py#L247))? TBC whether this will be faster
- [x] Alternative for discretising - #330 
   - As above
- [x] Alternative for simplifying
- [x] Alternative for `orphans` - seems difficult but the above three are the priority - #403

tinosulzer:
# Description

Create function `make_new_copy` to replace `deepcopy`
Fixes #329 

## Type of change

# Key checklist:

## Further checks: 


pybamm/expression_tree/copy.py
```diff
@@ -0,0 +1,61 @@
+#
+# Make a new copy of a symbol
+#
+import pybamm
+
+
+def make_new_copy(symbol):
```
martinjrobins: make_new_copy seems very similar in structure to simplify or evaluate, perhaps we should do this in a similar way. That is, have a `new_copy` function on Symbol, which is overridden on derived classes. You are already sorta doing that for the unary operators and concatenate anyway. What do you think?

tinosulzer: Yeah, it is a bit inconsistent at the moment. I was trying to keep it in line with the `ParameterValues.process_symbol`, `Discretisation.process_symbol` and `simplify` functions. The advantage of those is that we could later add a dictionary of already-processed symbols so that we avoid re-processing the same symbol multiple times. But maybe copying is straightforward enough that it should just be a function on Symbol.

martinjrobins: don't think we could use a dictionary of already-processed symbols here. Even if we've already copied a symbol, we still need to make another copy even if we find an identical symbol in the tree

tinosulzer: I think it would work, because what's causing problems with anytree is the `parent` attribute. So if we have a dict of variables whose parent is `None`, we can insert them into different places in the tree (since they get shallow-copied in `Symbol.__init__`) without having to deepcopy them again. As an aside, if we can find a better way of making a copy of an object in anytree with parent removed, then we don't need this whole thing

tinosulzer:
@martinjrobins - have now updated to your suggestion above

PR #409, opened 2019-05-15
Issue 407 evaluate parameters

### LINKED ISSUE #407: Evaluate dimensionless parameters
Create helper function to easily evaluate a set of dimensionless parameters from a set of dimensional ones

tinosulzer:
# Description

Create a function to evaluate and print all parameters 
(Also added a few more lead-acid parameters)
Fixes #407 

## Type of change

# Key checklist:

## Further checks: 


pybamm/parameters/print_parameters.py
```diff
@@ -0,0 +1,109 @@
...
+    "pybamm",
+    "constants",
+]
+
+
+def print_parameters(parameters, parameter_values, output_file=None):
+    """
+    Evaluate and print parameters to an output file.
+    For dimensionless parameters that depend on the C-rate, the value is given as a
+    function of the C-rate (either x * Crate or x / Crate depending on the dependence)
```
martinjrobins: is `Crate` or `C-rate` a common battery term? Perhaps we should give the full name at least once?

tinosulzer: `C-rate` is the full name, it is very common so all the battery people should know what this means (A C-rate of 1 C is the current required to fully discharge the battery in 1 hour, 2 C is current to discharge to the battery in 0.5 hours, etc)

PR #420, opened 2019-05-21
Comsol voltage compare

### LINKED ISSUE #288: DFN Comsol benchmark
Find information on Comsol DFN model:
- [ ] how many grid points
- [ ] how fast does it run
- [ ] convergence?
- [ ] parameter values

rtimms:
# Description

Added scripts and a notebook for comparing results with comsol. Will make a separate issue for testing convergence against a comsol result on a very fine mesh.

Fixes #288 

## Type of change

# Key checklist:

## Further checks: 


rtimms:
so this is actually a fair amount of data to upload here... I can reduce the number of time points the comsol files, so they are a bit smaller. also, do we just want to upload the voltage data, or is it useful for people to be able to run the script which compares internal variables? thoughts @tinosulzer @Scottmar93 ?

tinosulzer:
I think it would be useful to compare internal variables with the slider.
You could make a single dictionary of comsol variables and then use [pickle](https://docs.python.org/3/library/pickle.html) to store it?

PR #422, opened 2019-05-21
Issue 421 shape test

tinosulzer:
# Description

Fixes #421 

## Type of change

Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)

# Key checklist:

- [x] No style issues: `$ flake8`
- [x] All tests pass: `$ python run-tests.py --unit`
- [x] The documentation builds: `$ cd docs` and then `$ make clean; make html`

You can run all three at once, using `$ python run-tests.py --quick`.

## Further checks: 

- [x] Code is commented, particularly in hard-to-understand areas
- [x] Tests added that prove fix is effective or that feature works
- [x] Any dependent changes have been merged and published in downstream modules

PR #423, opened 2019-05-21
Newman tiedemann capacitance

tinosulzer:
# Description

Fix Newman-Tiedemann model with capacitance to agree with the model without capacitance.
Also fix Indefinite Integral so that `BoundaryValue(IndefiniteIntegral(...), "left")` is always zero
Note: branches from #426 (which branches from #422)

## Type of change

# Key checklist:

## Further checks: 


pybamm/spatial_methods/finite_volume.py
```diff
@@ -273,11 +300,16 @@ def indefinite_integral_matrix(self, domain):
         n = submesh.npts
         sec_pts = len(submesh_list)
 
-        # note we have added a row of zeros at top for F(0) = 0
         du_n = submesh.d_nodes
         du_entries = [du_n] * (n - 1)
         offset = -np.arange(1, n, 1)
-        sub_matrix = diags(du_entries, offset, shape=(n, n - 1))
+        main_integral_matrix = diags(du_entries, offset, shape=(n, n - 1))
+        bc_offset_matrix = lil_matrix((n, n - 1))
```
Scottmar93: lil matrix ft. DJ diags

pybamm/spatial_methods/finite_volume.py
```diff
@@ -273,11 +300,16 @@ def indefinite_integral_matrix(self, domain):
...
+        bc_offset_matrix[:, 0] = du_n[0] / 2
+        sub_matrix = main_integral_matrix + bc_offset_matrix
+        # add a column of zeros at each end
+        zero_col = csr_matrix((n, 1))
+        sub_matrix = hstack([zero_col, sub_matrix, zero_col])
         matrix = kron(eye(sec_pts), sub_matrix)
```
Scottmar93: should we csr this too?

tests/unit/test_expression_tree/test_symbol.py
```diff
@@ -267,17 +267,17 @@ def test_has_spatial_derivatives(self):
         grad_div_eqn = pybamm.div(grad_eqn)
         algebraic_eqn = 2 * var + 3
         self.assertTrue(grad_eqn.has_spatial_derivatives())
-        self.assertTrue(grad_eqn.has_gradient())
-        self.assertFalse(grad_eqn.has_divergence())
+        self.assertTrue(grad_eqn.has_symbol_of_class(pybamm.Gradient))
```
Scottmar93: I think it would be okay to just rename the function `has` if there isn't any clashes. Then doing
```python
    grad_eqn.has(pybamm.Gradient)
```

This really minor though so I don't mind if you prefer to keep it the way it is. 
Is a little neater? Could also create a super class for Gradient and Divergence `pybamm.SpatialDerivative` (or something) and then can just use 
```
 grad_eqn.has(pybamm.SpatialDerivative)
```

tinosulzer: I prefer to err on the side of clarity on this one

PR #427, opened 2019-05-23
Issue 426 refactor ghost nodes

### LINKED ISSUE #426: Refactor `add_ghost_nodes` to work with different symbols
`add_ghost_nodes` currently needs to access `y_slices` and so only works with `StateVectors` or concatenations of.
Refactor (e.g. using matrix multiplication instead) so that it can work with any kind of symbol. Needed for some of the different models.

tinosulzer:
# Description

- Refactor `add_ghost_nodes` to work with any symbol
- Refactor `Index`, add jacobian

Fixes #426 
Note: branches from #422 

## Type of change

# Key checklist:

## Further checks: 


tests/unit/test_spatial_methods/test_finite_volume.py
```diff
@@ -959,7 +956,12 @@ def test_indefinite_integral(self):
         grad_and_div = pybamm.div(pybamm.grad(phi))
         int_grad_phi = pybamm.IndefiniteIntegral(grad_and_div, x)
         disc.set_variable_slices([phi])
-
+        disc._bcs = {
```
Scottmar93: why do we now have boundary conditions here?

tinosulzer: Needed so that the integrand has the right shape, will add a comment

PR #432, opened 2019-05-24
Issue 431 check solver

### LINKED ISSUE #431: Solver check
Add checks to solver to make sure we don't try to solve an ODE model with a DAE solver, and vice versa

tinosulzer:
# Description

Add a check to make sure that the right solver is being used for the model.
Also split up `solver.solve` into two shorter functions.
Fixes #431 

## Type of change

# Key checklist:

## Further checks: 


pybamm/solvers/dae_solver.py
```diff
@@ -125,6 +97,78 @@ def jacobian(t, y):
...
+            Jacobian matrix for the differential and algebraic equations
+
+        Raises
+        ------
+        :class:`pybamm.SolverError`
+            If the model contains any algebraic equations (in which case a DAE solver
+            should be used instead)
+        """
+        if len(model.algebraic) == 0:
+            raise pybamm.SolverError(
```
martinjrobins: Should this be an error? It is valid to solve an ODE system with a DAE solver (not sure if it is actually useful though!)

tinosulzer: Yeah I just got lazy and said this should be an error ... do you think it should give a warning or just pass?

rtimms: Worth noting that sundials cvode doesn't allow for ODEs with a mass matrix (i.e. of the form `M*y_dot = f(t,y)`), it requires that you invert to give `y_dot` explicitly. You can pass ODEs like `M*y_dot = f(t,y)` to ida, so I guess people might want to do this if they don't like the scipy solver for some reason...

PR #436, opened 2019-05-27
Issue 400 improve coverage

Scottmar93:
# Description

Coverage is up to 99% now. Still some things left to do so will leave the issue open. The main outstanding things left to cover or simplify and quick_plot. 

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks: 


pybamm/__init__.py
```diff
@@ -107,6 +107,8 @@ def version(formatted=False):
     SolverError,
     ShapeError,
     ModelWarning,
+    UndefinedOperation,
```
tinosulzer: just semantics, but I think `UndefinedOperationError` would be more consistent with the other errors

pybamm/quick_plot.py
```diff
@@ -73,7 +73,7 @@ def __init__(self, models, mesh, solvers, output_variables=None):
 
         # Default output variables for lead-acid and lithium-ion
         if output_variables is None:
-            if isinstance(models[0], pybamm.LithiumIonBaseModel):
+            if isinstance(models[0], pybamm.LithiumIonBaseModel):  # pragma: no cover
```
tinosulzer: it should be possible to test this?

pybamm/util.py
```diff
@@ -72,7 +72,7 @@ def time(self):
         return timeit.default_timer() - self._start
 
 
-def profile(code, sort="cumulative", num=30):
+def profile(code, sort="cumulative", num=30):  # pragma: no cover
```
tinosulzer: turns out we don't need this bit of code, so can just delete it

PR #443, opened 2019-05-30
Issue 441 shape test

### LINKED ISSUE #441: Shape tests for all expression tree objects
When creating a model, at the moment some errors are only raised when we try to discretise, which makes it a bit difficult to debug. 
The aim of this issue is to find a way to do shape tests at the node creation stage for nodes that don't have an obvious shape, i.e. can't be evaluated (e.g. `Variable`, `Concatenation`, etc).

The best I can come with so far is an `evaluate_for_shape` function that, by default, calls `evaluate`, but if this fails returns a numpy array of the appropriate shape. For example, always give variables in the negative electrode shape '17', variables in the separator shape '23', etc.
Whenever a new object is created, the function `evaluate_for_shape` is called to make sure that it's possible to create that object without getting shape errors further down the line.

tinosulzer:
# Description

Add a test on node initialisation to check that the node can be evaluated without any shape errors.
Works even for nodes that can't be evaluated directly (creates a vector of the appropriate shape instead).
Will hopefully help with catching errors earlier when creating new models, but isn't flawless (there can still be errors in the discretisation, for example). 
Checking for shape on every node initialisation makes the models pretty slow, so I have added a settings class with a `debug_mode` attribute, which is `False` by default but can be set to `True` from anywhere. Not sure whether this is a good idea ...
Fixes #441

## Type of change

# Key checklist:

## Further checks: 

### DIFF
--- a/pybamm/__init__.py
+++ b/pybamm/__init__.py
@@ -56,7 +56,7 @@ from .expression_tree.exceptions import (
     ShapeError,
     ModelWarning,
 )
-from .expression_tree.symbol import Symbol
+from .expression_tree.symbol import Symbol, DOMAIN_SIZES_FOR_TESTING
 from .expression_tree.binary_operators import (
     is_scalar_zero,
     is_matrix_zero,
--- a/pybamm/__init__.py
+++ b/pybamm/__init__.py
@@ -29,10 +29,13 @@ def version(formatted=False):
 
 
 #
-# Set debug mode to false by default
+# Utility classes and methods
 #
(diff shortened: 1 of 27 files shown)


pybamm/__init__.py
```diff
@@ -28,6 +28,15 @@ def version(formatted=False):
         return VERSION_INT
 
 
+#
+# Utility classes and methods
```
Scottmar93: I see you moving things around the `__init__.py` to increase line count

PR #448, opened 2019-06-02
Fix lead acid composite

tinosulzer:
# Description

Fix lead-acid composite model to agree better with old code and full model

## Type of change

# Key checklist:

## Further checks: 


pybamm/solvers/base_solver.py
```diff
@@ -46,3 +46,30 @@ def solve(self, model, t_eval):
...
+        that works for all the different solvers.
+
+        Parameters
+        ----------
+        solution : :class:`pybamm.Solution`
+            The solution object
+        events : dict
+            Dictionary of events
+        """
+        if solution.termination == "final time":
```
rtimms: nice! this is very useful

tests/integration/test_models/test_submodels/test_interface.py
```diff
@@ -162,6 +162,68 @@ def test_discretisation(self):
         whole_cell_mesh = disc.mesh.combine_submeshes(*whole_cell)
         self.assertEqual(j.evaluate(None, y).shape, (whole_cell_mesh[0].npts, 1))
 
+    def test_diff_c_e_lead_acid(self):
```
rtimms: these are nice tests

PR #450, opened 2019-06-03
Issue 415 solution object

### LINKED ISSUE #415: Solution object
Output solution to a separate object than the solver

tinosulzer:
# Description

Reformat solution to return separate solution object, and add timings for solve.
Should we also put the variables from the model and/or the mesh into the solution object, so that a single object contains the whole solution?
Fixes #415 

## Type of change

# Key checklist:

## Further checks: 


pybamm/solvers/solution.py
```diff
@@ -0,0 +1,30 @@
+#
+# Solution class
+#
+
+
+class Solution(object):
+    """
+    Class containing the solution of, and various attributes associated with, a PyBaMM
+    model.
+    """
```
martinjrobins: needs description of t and y parameters for __init__

pybamm/solvers/solution.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def __init__(self, t, y):
+        self.t = t
+        self.y = y
+
+    @property
+    def t(self):
```
martinjrobins: docstrings for t property, and others below

tinosulzer:
Ok, will leave as is for now

PR #464, opened 2019-06-06
Issue 130 add windows and mac testing

### LINKED ISSUE #130: Windows and MacOS support
Check whether we can get everything working on Windows and MacOS, and add these to travis

martinjrobins:
# Description

Fixes #130 

Adds windows testing via appveyor and mac os x testing via travis. Both only run the unit and integration tests, without scikits.odes installed (might need a separate issue for installing this on windows/mac)

Note: I have had to skip a *lot* of tests in the case where scikits.odes are not installed. Once we have Ivan's timestepper in pybamm, might want to make this the default solver when scikits.odes are not installed

## Type of change

# Key checklist:

## Further checks: 


.appveyor.yml
```diff
@@ -0,0 +1,20 @@
+branches:
+  only:
+    - issue-130-windows
```
tinosulzer: Should this be removed before merge?

setup.py
```diff
@@ -14,6 +14,7 @@
     author="Valentin Sulzer",
```
tinosulzer: we should definitely change this now!

.appveyor.yml
martinjrobins: Yes! good spot

setup.py
martinjrobins: Perhaps we can say "The PyBaMM development team", and have an AUTHORS file with all the names?

tinosulzer: yep sounds good, we could even just leave `author` out completely

martinjrobins:
> so we could just make that the default solver for all ODE models? 

I've gone through and done this for all the ode models I could find

setup.py
martinjrobins: ok, I've now removed the author and author email

PR #454, opened 2019-06-04
2plus1 d models

rtimms:
# Description

A first stab at adding the finite element method in the y-z plane which uses fenics for the mesh and to get the mass and stiffness matrices. At the moment it can only discretise the laplacian and source terms (applying either Dirichlet or Neumann BCs at the tabs, zero flux elsewhere). Also added are new geometries for (2+xD)Macro1DMicro, and a current collector submodel. There are some tests the fenics spatial method computes the laplacian correctly etc., but not yet tested in a battery model.

To do:
- test in SPM
- add into other models (e.g. SPMe, DFN)
- more testing
- some basic plotting for 2/3D results
- good instructions for installing dolfin into a virtual environment 

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Key checklist:

- [x] No style issues: `$ flake8`
- [x] All tests pass: `$ python run-tests.py --unit`
- [x] The documentation builds: `$ cd docs` and then `$ make clean; make html`

You can run all three at once, using `$ python run-tests.py --quick`.

## Further checks: 

- [x] Code is commented, particularly in hard-to-understand areas
- [x] Tests added that prove fix is effective or that feature works
- [ ] Any dependent changes have been merged and published in downstream modules

PR #466, opened 2019-06-07
2plus1D fenics code

rtimms:
# Description
Was originally PR #454. I made a new branch off of that one to remove the fenics code so that the other stuff could be merged in. When I did that, the fenics PR also merged (but without the fenics code?).  Anyway, this is the branch with the fenics finite element spatial method.

A first stab at adding the finite element method in the y-z plane which uses fenics for the mesh and to get the mass and stiffness matrices. At the moment it can only discretise the laplacian and source terms (applying either Dirichlet or Neumann BCs at the tabs, zero flux elsewhere). Also added are new geometries for (2+xD)Macro1DMicro, and a current collector submodel. There are some tests the fenics spatial method computes the laplacian correctly etc., but not yet tested in a battery model.

To do:
- fix dolin install for testing
- test in SPM
- add into other models (e.g. SPMe, DFN)
- more testing
- some basic plotting for 2/3D results
- good instructions for installing dolfin into a virtual environment

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks: 


rtimms:
@tinosulzer @martinjrobins  just to update: I've managed to get the required libraries (I think),  but it is failing building the wheel for fenics (see [here](https://travis-ci.org/pybamm-team/PyBaMM/jobs/543774456#L4107)) I think this is the same error you were getting @Scottmar93 ?

tinosulzer:
Looks like you might need to install `setuptools` as well (see [this issue](https://github.com/pypa/pip/issues/1064)).

tinosulzer:
This looks painful ... have asked Flo and apparently the recommended way of installing FEniCS these days is via docker image. They have [some information online](https://fenics.readthedocs.io/projects/containers/en/latest/work_flows.html#create-a-custom-image-for-my-project) on how to do this. The advantage would be we wouldn't have to compile everything from scratch every time we test
Though using docker opens up a whole new can of worms, @martinjrobins what do you think about this?I definitely think we should stick to virtualenvs for non-testing purposes

PR #469, opened 2019-06-09
#468 add check for variables in list of variables, and remove some bad tests

### LINKED ISSUE #468: check that all variables in the model appear in `disc._y_slices`
Otherwise can lead to slightly obscure `KeyError` when discretising

tinosulzer:
# Description

Fixes #468 
Also used this branch to remove some of the tests that sometimes fail (timer tests, tolerance test for consistent initial conditions)

## Type of change

# Key checklist:

## Further checks: 

### DIFF
--- a/tests/unit/test_models/test_base_model.py
+++ b/tests/unit/test_models/test_base_model.py
@@ -329,11 +329,11 @@ class TestBaseModel(unittest.TestCase):
         }
 
         # Check warning raised
-        # self.assertWarns(pybamm.ModelWarning, model.check_well_posedness())
-        #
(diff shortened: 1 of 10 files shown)


tests/unit/test_models/test_base_model.py
```diff
@@ -331,10 +331,30 @@ def test_check_well_posedness_output_variables(self):
         # Check warning raised
         self.assertWarns(pybamm.ModelWarning, model.check_well_posedness())
 
-        # Check None entries have been removed from the variables dictionary
+        # # Check None entries have been removed from the variables dictionary
```
rtimms: double # here

PR #476, opened 2019-06-12
#475 add scikit-fem mesh and spatial method

### LINKED ISSUE #475: scikit-fem mesh and spatial method
**Summary**
Implement finite element method in current collector domain using [scikit-fem](https://github.com/kinnala/scikit-fem)

 **Motivation**
More lightweight than fenics

rtimms:
# Description

Add scikit-fem mesh and spatial method for 2D current collector submodels

To do: 
- implement 2D submodel in li-ion models (as separate issue)

Fixes #475 

## Type of change

# Key checklist:

## Further checks: 

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,7 @@ generate meshes and assemble discrete operators for use in the finite element me
 the finite element method is only implemented for a small number of submodels in PyBaMM
 (for example, see the [current collector submodel](https://github.com/pybamm-team/PyBaMM/blob/master/pybamm/models/submodels/current_collector.py)).
 
-To install scikit-fem, on the command-line type:
+Note that scikit-fem requires Python 3.6+. To install scikit-fem, on the command-line type:
 
 ```bash
 pip install scikit-fem
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ You'll need the following requirements:
 
 The first step is to get the code by cloning this repository
 
-```bash 
+```bash
 git clone https://github.com/pybamm-team/PyBaMM.git
 cd PyBaMM
 ```
@@ -40,21 +40,21 @@ python3 -m venv env
 
 You can then "activate" the environment using:
 
-```bash 
-source env/bin/activate 
+```bash
+source env/bin/activate
 ```
 
 Now all the calls to pip described below will install PyBaMM and its dependencies into
 the environment `env`. When you are ready to exit the environment and go back to your
 original system, just type:
 
-```bash 
-deactivate 
+```bash
+deactivate
 ```
 
 PyBaMM has the following python libraries as dependencies: `numpy`, `scipy`, `pandas`,
 `matplotlib`. These, along with PyBaMM, can easily be installed using `pip`. First, make
-sure you have activated your virtual environment as above, and that you have the latest 
+sure you have activated your virtual environment as above, and that you have the latest
 version of pip installed:
 
 ```bash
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -135,6 +135,20 @@ Please see the [scikits.odes
 documentation](https://scikits-odes.readthedocs.io/en/latest/installation.html) for more
 detailed installation instructions.
 
+### [scikit.fem](https://github.com/kinnala/scikit-fem)
```
tinosulzer: since `scikit.fem` is just a pip package, we should just add it to `setup.py` and `.requirements-docs.txt`, then we don't need separate instructions for installation (it will automatically install with `pip install .`) 
Then we also don't need separate options in travis

pybamm/meshes/scikit_fem_submeshes.py
```diff
@@ -0,0 +1,139 @@
...
+import pybamm
+
+import numpy as np
+import importlib
+
+skfem_spec = importlib.util.find_spec("skfem")
+if skfem_spec is not None:
+    skfem = importlib.util.module_from_spec(skfem_spec)
+    skfem_spec.loader.exec_module(skfem)
+
```
tinosulzer: a bit neater to use something like `have_scikits_fem` similar to `have_scikits_odes` (see `scikits_ode_solver.py`). Though given that it's just a pip package, we can probably assume it will always be installed anyway

pybamm/spatial_methods/scikit_finite_element.py
```diff
@@ -0,0 +1,314 @@
...
+        if lbc_type == "Dirichlet":
+            self.bc_apply(stiffness, mesh.negative_tab)
+        if rbc_type == "Dirichlet":
+            self.bc_apply(stiffness, mesh.positive_tab)
+
+        return pybamm.Matrix(stiffness)
+
+    def integral(self, domain, symbol, discretised_symbol):
+        """Vector-vector dot product to implement the integral operator.
+        See :meth:`pybamm.BaseDiscretisation.integral`
```
tinosulzer: should be `SpatialMethod.integral`

pybamm/spatial_methods/scikit_finite_element.py
```diff
@@ -0,0 +1,314 @@
...
+            return v
+
+        # assemble
+        vector = skfem.asm(integral_form, mesh.basis)
+        return pybamm.Matrix(vector[np.newaxis, :])
+
+    def indefinite_integral(self, domain, symbol, discretised_symbol):
+        """Implementation of the indefinite integral operator. The
+        input discretised symbol must be defined on the internal mesh edges.
+        See :meth:`pybamm.BaseDiscretisation.indefinite_integral`
```
tinosulzer: `SpatialMethod`

tests/unit/test_spatial_methods/test_scikit_finite_element.py
```diff
@@ -0,0 +1,222 @@
...
+        with self.assertRaises(NotImplementedError):
+            spatial_method.gradient(None, None, None)
+        with self.assertRaises(NotImplementedError):
+            spatial_method.divergence(None, None, None)
+        with self.assertRaises(NotImplementedError):
+            spatial_method.indefinite_integral(None, None, None)
+        with self.assertRaises(NotImplementedError):
+            spatial_method.boundary_value_or_flux(None, None)
+
+    def test(self):
```
tinosulzer: maybe use a more informative name

README.md
rtimms: `scikit.fem` requires python 3.6+, so it looks like we'll need to keep it as optional unless we drop 3.5?

tinosulzer: Ok, fair enough

Issue #482, opened 2019-06-27
Add handling gradients of subdomain variables


Scottmar93:
Currently if we try to take the gradient of `phi_e_n` we get a symbol of the wrong size because `phi_e_n` is not in the boundary conditions (and shouldn't be for checking purposes) instead `phi_e` is in the boundary conditions.

Scottmar93:
solution: automatically add to boundary conditions dict as this doesn't affect model checks

Scottmar93:
add new method to `BaseSubModel` called `set_internal_boundary_conditions` which looks through the submodel boundary conditions and if the key is a concatenation, creates internal boundary conditions for the orphans. 
This function if then called within the set model equations loop in `BaseBatteryModel`

Issue #481, opened 2019-06-27
Add handling of derivatives of derived variables


Scottmar93:
In SPMe we have explicit expressions for `phi_s`. We want to take gradients of `phi_s` to obtain the ohmic heating. However `phi_s` is not in the boundary conditions (and shouldn't be) so when we take the gradient the resulting symbol is of the wrong size.

Scottmar93:
add to boundary conditions on a case by case basis. this will not affect the model checks

TomTranter:
Can you explain how you evaluate the gradient please

TomTranter:
Is there a way to access gradients of variables easily such as 
```
phis_neg = pybamm.ProcessedVariable(model.variables['Negative electrode potential [V]'], solution.t, solution.y, mesh=mesh)
grad_phis = pybamm.Gradient(model.variables['Negative electrode potential [V]'])
grad_phis.evaluate(solution.t, x)
```

Scottmar93:
Sorry for missing this @TomTranter. Currently, we can only take gradients before we discretise the model. This is because `pybamm.grad` creates a `pybamm.Gradient` symbol in the expression tree which is then converted to a matrix when the model is discretised (which is dependent upon the type of spatial method that is employed e.g. finite volume). 

Therefore, the easiest way to get direct access to the gradient of a symbol would be to add it to the model.variables dictionary before discretising and solving the model in a similar manner to other output variables. The following should work for the DFN: 

```
import pybamm
import numpy as np

model = pybamm.lithium_ion.DFN()

phi_s_n = model.variables["Negative electrode potential"]
model.variables.update(
    {"Gradient of negative electrode potential": pybamm.grad(phi_s_n)}
)

geometry = model.default_geometry

param = model.default_parameter_values
param.process_model(model)
param.process_geometry(geometry)

mesh = pybamm.Mesh(geometry, model.default_submesh_types, model.default_var_pts)

# discretise model
disc = pybamm.Discretisation(mesh, model.default_spatial_methods)
disc.process_model(model)

# solve model
t_eval = np.linspace(0, 2, 100)
solution = model.default_solver.solve(model, t_eval)

model.variables = pybamm.post_process_variables(
    model.variables, solution.t, solution.y, disc.mesh
)

grad_phi_s_n = model.variables["Gradient of negative electrode potential"]
``` 

[...]

Issue #480, opened 2019-06-27
Allow average correction for j


Scottmar93:
We want to ensure the average property of j is maintained. In the interface submodel, we therefore want to do: 
```
    j = j_true_av + (j - pybamm.average(j))
```
to enforce this. 

However, at the moment this causes shape error. I think it might be something related to Jacobian.

PR #484, opened 2019-06-28
Issue 457 add temperature equations

### LINKED ISSUE #457: Add temperature to lithium-ion models
**Summary**
Add temperature equations to pybamm with focus of getting working for lithium-ion for now 

 **Motivation**
need temperature dependence for paper

Scottmar93:
# Description
Sorry for the monster pull request. Main changes are that the submodels have been refactored so that it is easier to insert new submodels such as the thermal submodel which I have also added here. 

Current state: 
- all lithium-ion battery models work in the new format
- all standard form lead acid batteries work (although some convergence properties aren't working yet) I think that will be solved by #480 though. 
- some aspects of the thermal model don't work yet. these need #482 #481 first.  
- new folder structure for submodels has been created
- docs have been added
- basic standard tests for submodels have been added (but we should maybe add to these later)

Since lead acid model and all the additional functionality in them are not yet up and running, I have moved all of the old submodels and old models to folders with the prefix 'old' and given all classes the prefix 'Old' so that these can all still be used without any interference with the new models. We can then move these across slowly because there are a lot of features to do all at once. 

Fixes #457 

## Type of change

# Key checklist:

## Further checks: 


Scottmar93:
Thank you for the comment @tinosulzer. I have made the suggested changes. Addressing your general comments 
- I did initially try to do this because I wanted to preserve the original interface with full models as much possible. However, I felt that after stripping back the models to just combining submodels in some order that a surface-form model and its equivalent standard-form model were actually about as different as two standard-form models (e.g. SPM and SPMe). This is because surface-form models require a different set of independent variables and they also require submodels to be loaded in a different order. So almost the whole model just became one big if statement (I think the only common component is the electrolyte diffusion equation?) at which point I thought might aswell split it to make thing
- i tried to track down some of the empty test files.

pybamm/models/submodels/electrolyte/stefan_maxwell/conductivity/surface_potential_form/leading_capacitance_stefan_maxwell_conductivity.py
```diff
@@ -0,0 +1,47 @@
+#
+# Class for full surface form electrolyte conductivity employing stefan-maxwell
```
rtimms: leading order

Scottmar93: Thank you for flagging these inconsistensies in the doc strings. I'll check again through each model

PR #487, opened 2019-07-02
Issue 483 input currents

### LINKED ISSUE #483: More input currents
Add some more input currents which are a function of (dimensional) time. Allow to read a drive cycle from .csv file (e.g. C-rate vs time [seconds])

rtimms:
# Description

Added some new current profiles. In particular, data can be loaded from a csv file (see `get_current_data.py`). Interpolating at the top of the function worked (thanks @tinosulzer !) but I thought it would be easier for the user to just pass the filename to an object when creating a model, rather than explicitly changing `get_current_data.py`. This now means that some current functions are just functions, but that one is a class. I could change them all to being classes, or just put this one in a folder separate to standard_current_functions to make it clear it behaves differently. Thoughts @tinosulzer @Scottmar93  ? I think in general it is probably useful to be able to pass parameters to the user-defined input currents (e.g. see results/drive_cycles/user_sin_current_simulation.py, where you may want to pass some list of parameters to a current function and compare the results).

P.S. after you comments, I'll add more info so that it is very clear what data `get_current_data.py` expects (non-dimensional current amplitude vs time in seconds), and maybe a notebook about adding your own input current function.

Fixes #483 

## Type of change

# Key checklist:

## Further checks: 


Scottmar93:
Let me know when this is ready again for review :)

rtimms:
still need to merge master, fill out doc properly and (maybe) an example notebook about adding your own current function, but I've added the classes `GetConstantCurrent` which accepts one parameter (the value of the current, by default `I_typ`), `GetUserCurrent` which allows the user to enter their own function with however many parameters they want, and `GetCurrentData` which loads current data (dimensional or not) and optionally voltage data and creates an interpolant. See [here](https://github.com/pybamm-team/PyBaMM/tree/issue-483-input-currents/results/drive_cycles) for examples.

rtimms:
@tinosulzer @Scottmar93  I think this is ready for another look now

pybamm/models/full_battery_models/lithium_ion/base_lithium_ion_model.py
```diff
@@ -19,10 +19,22 @@ def __init__(self, options=None):
 
     def set_standard_output_variables(self):
         super().set_standard_output_variables()
-        # Additional standard output variables
+        # Current
+        icell = pybamm.standard_parameters_lithium_ion.current_with_time
```
tinosulzer: should this be `current_density_with_time`?

tests/integration/test_old_models/test_old_submodels/test_old_interface.py
```diff
@@ -66,9 +66,9 @@ def test_simplify_constant_current(self):
 
         j = disc.process_symbol(parameter_values.process_symbol(j))
 
-        # Simplifiy, since current is constant this should give a vector
+        # Simplifiy, since current is constant this should give an array when evaluated
         j_simp = j.simplify()
-        self.assertIsInstance(j_simp, pybamm.Vector)
+        self.assertIsInstance(j_simp.evaluate(t=0), np.ndarray)
```
tinosulzer: This test no longer really does anything, as j will always give an array when evaluated. Probably best to remove, or think about how the `GetConstantCurrent` function would simplify

rtimms: good spot!

pybamm/models/full_battery_models/lithium_ion/base_lithium_ion_model.py
rtimms: thanks, should this also be the case for lead acid? since it ends up in the variable "Current density" ?

tinosulzer: yeah, am expecting `icell` to be a current density for lead-acid as well

PR #497, opened 2019-07-04
#496 multiply ydot by mass

### LINKED ISSUE #496: residuals in dae solver not getting multiplied by mass
**Describe the bug**
When computing the residuals in dae_solver, ydot isn't getting multiplied by the mass matrix. hasn't shown up as a problem as matrix has always been identity for finite volumes

rtimms:
# Description

Fixes #496 
## Type of change

# Key checklist:

## Further checks: 

### DIFF
--- a/pybamm/solvers/dae_solver.py
+++ b/pybamm/solvers/dae_solver.py
@@ -73,13 +73,8 @@ class DaeSolver(pybamm.BaseSolver):
             # turn into 1D arrays
             rhs_eval = rhs_eval[:, 0]
             alg_eval = alg_eval[:, 0]
-            # get entries of the mass matrix corresponding to rhs
-            mass_matrix = model.mass_matrix.entries[
-                0 : rhs_eval.shape[0], 0 : rhs_eval.shape[0]
-            ]
-            return np.concatenate(
-                (rhs_eval - mass_matrix @ ydot[: rhs_eval.shape[0]], alg_eval)
-            )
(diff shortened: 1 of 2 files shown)


pybamm/solvers/dae_solver.py
```diff
@@ -73,7 +73,13 @@ def residuals(t, y, ydot):
...
             rhs_eval = rhs_eval[:, 0]
             alg_eval = alg_eval[:, 0]
-            return np.concatenate((rhs_eval - ydot[: rhs_eval.shape[0]], alg_eval))
+            # get entries of the mass matrix corresponding to rhs
+            mass_matrix = model.mass_matrix.entries[
+                0 : rhs_eval.shape[0], 0 : rhs_eval.shape[0]
+            ]
+            return np.concatenate(
+                (rhs_eval - mass_matrix @ ydot[: rhs_eval.shape[0]], alg_eval)
+            )
```
tinosulzer: `mass_matrix @ ydot` should have the same functionality (assuming `mass_matrix` has all zeros in the algebraic block).
Worth investigating which implementation is quicker, as this happens every time step

rtimms: yeah that is much faster thanks @tinosulzer !

tinosulzer: Np. I'd also put `model.mass_matrix.entries` outside of the `residuals` function

PR #523, opened 2019-07-09
Issue 477 2plus li ion

### LINKED ISSUE #477: 2plus1D li-ion models
**Summary**
Add 2+1D li-ion models using `scikit-fem` mesh and method in the 2D current collector domain

rtimms:
# Description
Adds 2plus1D SPM and SPMe. Broadcasts can now be over the full domain, or just the primary domain.

Fixes #477

## Type of change

# Key checklist:

## Further checks: 


pybamm/discretisations/discretisation.py
```diff
@@ -415,6 +415,8 @@ def process_dict(self, var_eqn_dict):
             pybamm.logger.debug("Discretise {!r}".format(eqn_key))
             new_var_eqn_dict[eqn_key] = self.process_symbol(eqn)
 
+            new_var_eqn_dict[eqn_key].shape
```
tinosulzer: `.test_shape()` would make it clearer what the functionality of this line is

pybamm/expression_tree/binary_operators.py
```diff
@@ -398,19 +401,33 @@ def __init__(self, left, right):
...
-        if isinstance(left, pybamm.Array):
+        if isinstance(left, pybamm.Array) or (
+            isinstance(left, pybamm.Negate) and isinstance(left.child, pybamm.Array)
+        ):
             left = pybamm.Matrix(csr_matrix(left.evaluate()))
-            return left @ right.jac(variable)
+            jac = left @ right.jac(variable)
+            # Remove domain as matrix multiplies which come from an integral
+            # end up getting the child domains back which throws an error
+            jac.domain = []
```
tinosulzer: ah nice, I was getting loads of errors like this but didn't think of doing this :)

pybamm/expression_tree/binary_operators.py
```diff
@@ -605,10 +622,11 @@ def __init__(self, left, right):
             left.domain != ["current collector"]
             and left.domain != ["negative particle"]
             and left.domain != ["positive particle"]
+            and left.domain != []
         ):
             raise pybamm.DomainError(
                 """left child domain must be 'current collector', 'negative particle',
-                or 'positive particle', not'{}""".format(
+                'positive particle', or '', not'{}""".format(
```
tinosulzer: I think (lower case) `outer` function should just return a multiply instead in the case where `left.domain = []`? or does this just not work?

pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -372,10 +372,35 @@ def set_voltage_variables(self):
...
+            phi_s_cp = self.variables["Positive current collector potential"]
+            phi_s_cn_dim = self.variables["Negative current collector potential [V]"]
+            phi_s_cp_dim = self.variables["Positive current collector potential [V]"]
+            # In 2D left corresponds to the negative tab and right the positive tab
+            V = pybamm.BoundaryValue(phi_s_cp, "right") - pybamm.BoundaryValue(
+                phi_s_cn, "left"
+            )
+            V_dim = pybamm.BoundaryValue(phi_s_cp_dim, "right") - pybamm.BoundaryValue(
+                phi_s_cn_dim, "left"
+            )
```
tinosulzer: ok for now but we'll need to think of less confusing nomenclature - maybe allow "negative tab" and "positive tab" as keys in the bc dicts, but this is an issue for later

pybamm/expression_tree/unary_operators.py
```diff
@@ -635,7 +637,8 @@ def average(symbol):
         return new_symbol
     # If symbol is a Broadcast, its average value is its child
     elif isinstance(symbol, pybamm.Broadcast):
-        return symbol.orphans[0]
+        new_symbol = symbol.orphans[0]
```
Scottmar93: `return symbol.orphans[0]`

pybamm/expression_tree/binary_operators.py
rtimms: new primary broadcast sometimes forces an upper case `Outer` for symbols with domain `[]`

pybamm/models/full_battery_models/base_battery_model.py
rtimms: I agree

pybamm/expression_tree/unary_operators.py
rtimms: thanks

rtimms:
didn't run the integration tests locally... will fix now

rtimms:
@tinosulzer @Scottmar93  I've changed `jac` so that it removes domains. Think this should fix a lot of errors when adding new models. I think the shape test error with particles in 2plus1D requires a bit more thought, so for now I've turned debug mode off for the 2plus1D tests. The models still build/run ok, as everything is the right shape, just the domain isn't being set correctly

tinosulzer:
Yeah, let's get this merged and I'll look into shape tests next week

PR #529, opened 2019-07-10
Issue 481 add handling derivatives of derived variables

### LINKED ISSUE #481: Add handling of derivatives of derived variables

Scottmar93:
# Description
Added derivatives of derived variables. Gradients can now be taken of phi_e_n etc for SPMe. Additionally, taking gradients of subdomain variables has been extended to the 2+1D models. Ohmic heating terms in all the thermal models have also been added. 

Still failing 1 debug mode test for 2+1 models. 

Fixes #481 

## Type of change

# Key checklist:

- [ ] Debug mode tests work. 

## Further checks: 


pybamm/models/submodels/electrolyte/stefan_maxwell/conductivity/surface_potential_form/leading_surface_form_stefan_maxwell_conductivity.py
```diff
@@ -62,6 +63,18 @@ def get_coupled_variables(self, variables):
         else:
             return variables
 
+    # def set_boundary_conditions(self, variables):
+    #     pass
```
rtimms: can this be removed now?

tests/integration/test_models/test_full_battery_models/test_lithium_ion/test_dfn.py
```diff
@@ -8,6 +8,9 @@
 import unittest
 
 
+pybamm.settings.debug_mode = True
```
tinosulzer: should this be removed?

tests/unit/test_models/test_submodels/test_electrolyte/test_stefan_maxwell/test_conductivity/test_base_stefan_maxwell_conductivity.py
```diff
@@ -7,11 +7,16 @@
 import unittest
 
 
+pybamm.settings.debug_mode = True
```
tinosulzer: should this be removed?

Scottmar93:
> looks great, thanks @Scottmar93 ! maybe not related to this issue, but can we take gradients etc. on variables after solving, and then process the result? or do you have to put them in as variables pre-solve?

Will need to take gradient before and store them in the variables dict at the moment. To simulate the functionality you describe, we could add an option in the `grad` function where if you pass in a processed variable, so that it looks in the varaibles dictionary for the gradient (or we could change the processed variables class so that the gradient is stored within the processed variable class and we can do therefore just do `processed_variable.grad`.

PR #550, opened 2019-07-24
introduce different methods of stepping solution

TomTranter:
# Description
Adds a step method for the solvers. At the moment this returns a new solution object each time step is called (this is the same behavior as the step method in scikits-odes), but maybe it would be good to automatically append the next step to the existing solution? 

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks: 


martinjrobins:
the error you are getting on travis for Python 3.5 might be fixed in #626. This is now in master so merge it from there

Issue #546, opened 2019-07-21
2+1D lead-acid


tinosulzer:
Start to create 1+1D and 2+1D lead-acid models

PR #552, opened 2019-07-27
Issue 546 2plus1 lead acid

### LINKED ISSUE #527: improve shape tests for 2plus1D
Sometimes shape test gives an error because the domain of a variable isn't set to current collector in 2D, when it should be. In practice the object ends up being the correct shape, just the domain it is labelled with is wrong. e.g. in `inverse_butler_volmer` the line `eta_r = (2 / ne) * pybamm.Function(np.arcsinh, j / (2 * j0))` gives a shape error. `j` has the expected shape, but `j0` doesn't as it contains a broadcast whose child had domain `[]` but should be `["current collector"]`

tinosulzer:
# Description

Some reformating to make shape tests less hacky. Should help generally with implementation of 2+1D models hopefully. Some notable changes:
- Changed "Average ..." to "X-averaged ..." and `pybamm.average` to `pybamm.x_average`
- Added `auxiliary_domains` to explicitly give the secondary (and tertiary) domain of an object. This helps with shape tests and might be useful for discretisations (use auxiliary domains instead of geometry) 

Part of #546 , fixes #527 

## Type of change

# Key checklist:

## Further checks: 


pybamm/models/submodels/electrolyte/stefan_maxwell/diffusion/first_order_stefan_maxwell_diffusion.py
```diff
@@ -95,26 +105,26 @@ def get_coupled_variables(self, variables):
...
-            pybamm.Broadcast(c_e_0, "negative electrode") + param.C_e * c_e_n_1,
-            pybamm.Broadcast(c_e_0, "separator") + param.C_e * c_e_s_1,
-            pybamm.Broadcast(c_e_0, "positive electrode") + param.C_e * c_e_p_1,
+            pybamm.PrimaryBroadcast(c_e_0, "negative electrode") + param.C_e * c_e_n_1,
+            pybamm.PrimaryBroadcast(c_e_0, "separator") + param.C_e * c_e_s_1,
+            pybamm.PrimaryBroadcast(c_e_0, "positive electrode") + param.C_e * c_e_p_1,
         )
         variables.update(self._get_standard_concentration_variables(c_e))
-        # Update with analytical expressions for first-order averages
+        # Update with analytical expressions for first-order x-averageds
```
Scottmar93: x-averages

pybamm/models/submodels/oxygen_diffusion/first_order_oxygen_diffusion.py
```diff
@@ -37,8 +37,8 @@ def get_coupled_variables(self, variables):
         x_p = pybamm.standard_spatial_vars.x_p
 
         # Unpack
-        eps_s_0 = variables["Leading-order average separator porosity"]
-        eps_p_0 = variables["Leading-order average positive electrode porosity"]
+        eps_s_0 = variables["Leading-order x-averaged separator porosity"]
```
Scottmar93: maybe call `eps_s_0_av` for clarity

PR #556, opened 2019-07-30
Typo in current collector eqns

### LINKED ISSUE #555: typo in 2d current collector eqns
The conductivity is multiplying the through-cell current density instead of the potential

rtimms:
# Description

Fixes typo in current collector equations

Fixes #555

## Type of change

# Key checklist:

## Further checks: 


pybamm/parameters/standard_parameters_lithium_ion.py
```diff
@@ -272,12 +271,17 @@ def U_p_dimensional(sto, T):
...
+sigma_p_prime = sigma_p * delta
+sigma_cp_prime = sigma_cp * delta ** 2
+sigma_cn_dbl_prime = sigma_cn_prime * delta
+sigma_cp_dbl_prime = sigma_cp_prime * delta
 # should rename this to avoid confusion with Butler-Volmer
-
-alpha = 1 / (sigma_cn * (L_x / L_z) ** 2 * l_cn) + 1 / (
-    sigma_cp * (L_x / L_z) ** 2 * l_cp
+alpha = 1 / (sigma_cn * delta ** 2 * l_cn) + 1 / (
+    sigma_cp * delta ** 2 * l_cp
```
tinosulzer: We should still find a better variable for this that doesn't clash with the Butler-Volmer symmetry factor

rtimms:
yeah it makes the potentials in the current collector as expected. didn’t spot it before as the ref potential was missing from the rescaling of phi_s_cp. now the potential looks like the ocv plus a small spatial correction as expected

rtimms:
also i agree about alpha. will try to think of a more meaningful name...

tinosulzer:
I'd been using \varsigma in latex but that's slightly outside the standard set of letters

rtimms:
fixed by #564

PR #557, opened 2019-07-30
Issue 554 2dbug

### LINKED ISSUE #554: discretising 2D bug
when objects that are concatenated are discretised in 2D, they come out the right shape but wrong order
to do:
- [x] state vector
- [x] domain concatenation
- [x] discretisation
- [x] jacobians
- [x] evaluators

tinosulzer:
# Description

Fix discretisation of `StateVector` and `DomainConcatenation` so that 2D objects get turned into column vectors in the right order when concatenated
Fixes #554 

## Type of change

# Key checklist:

## Further checks: 


pybamm/expression_tree/concatenations.py
```diff
@@ -262,17 +269,27 @@ def _concatenation_evaluate(self, children_eval):
...
         return vector
 
     def _jac(self, variable):
         """ See :meth:`pybamm.Symbol._jac()`. """
-        children = self.cached_children
-        if len(children) == 0:
-            return pybamm.Scalar(0)
-        else:
-            return SparseStack(*[child.jac(variable) for child in children])
+        # note that this assumes that the children are in the right order and only have
```
rtimms: what happens when a child has more than one domain? does this ever happen?

tinosulzer: currently this doesn't happen in any of the models and `NotImplementedError` would be raised

PR #559, opened 2019-07-31
#558 print out named scalar as name = value in Symbol.render()

### LINKED ISSUE #558: print value of named scalars for render
a parameter node, and a named scalar node is currently printed the same when you call `symbol.render()`. There is plenty of space, so think the latter should be printed as `name = value`

martinjrobins:
# Description

Fixes #558 

## Type of change

# Key checklist:

## Further checks: 


martinjrobins:
> * overload the `__repr__` operator for `Scalar`

tried that, but it doesn't play nicely when you `print` an equation (you get weird equals signs in there), so just defined it for `render`

> * use `.format()` instead of `%s`

Yes, I'll fix this

PR #564, opened 2019-08-01
Issue 546 models

### LINKED ISSUE #546: 2+1D lead-acid

tinosulzer:
# Description

Add some 1+1D models for lead-acid

Fixes #546 

## Type of change

# Key checklist:

## Further checks: 


pybamm/models/submodels/current_collector/quite_conductive_potential_pair.py
```diff
@@ -0,0 +1,97 @@
...
+        param = self.param
+        applied_current = param.current_with_time
+        cc_area = self._get_effective_current_collector_area()
+        z = pybamm.standard_spatial_vars.z
+
+        phi_s_cn = variables["Negative current collector potential"]
+        phi_s_cp = variables["Positive current collector potential"]
+        i_boundary_cc = variables["Current collector current density"]
+        i_boundary_cc_0 = variables["Leading-order current collector current density"]
+        c = variables["Lagrangian multiplier"]
```
rtimms: minor, but should be Lagrange multiplier rather than Lagrangian

pybamm/models/submodels/current_collector/quite_conductive_potential_pair.py
```diff
@@ -0,0 +1,97 @@
...
+
+        self.algebraic = {
+            phi_s_cn: (param.sigma_cn * param.delta ** 2 * param.l_cn)
+            * pybamm.laplacian(phi_s_cn)
+            - pybamm.source(i_boundary_cc_0, phi_s_cn),
+            i_boundary_cc: (param.sigma_cp * param.delta ** 2 * param.l_cp)
+            * pybamm.laplacian(phi_s_cp)
+            + pybamm.source(i_boundary_cc_0, phi_s_cp)
```
rtimms: might be worth adding a comment that the second argument in source is the variable which appears in the laplacian for that eqn (i.e. the var in that eqn to which BCs are applied) but maybe this is clear from the docs for source

pybamm/models/submodels/current_collector/homogeneous_current_collector.py
```diff
@@ -23,13 +23,21 @@ def __init__(self, param):
 
     def get_fundamental_variables(self):
 
-        whole_cell = ["negative electrode", "separator", "positive electrode"]
-        i_cc = pybamm.FullBroadcast(pybamm.Scalar(0), whole_cell, "current collector")
-
+        # TODO: grad not implemented for 2D yet
+        i_cc = pybamm.Scalar(0)
         i_boundary_cc = pybamm.PrimaryBroadcast(
```
rtimms: for 1D do we expect the `i_boundary_cc` to be `current_with_time` or `current_with_time/cc_area` ? Not for this issue, but I'm not sure if it would be more user friendly to provide cell capacity in the csv files and then set the typical current to be the 1C discharge current and C-rate to be 1 by default (obviously the user can adjust these later)

tinosulzer: In 1D I think it should be (dimensionless) `current_with_time`, which is the same as dimensionless `current_density_with_time` but not the same as `current_with_time/cc_area`.
Don't completely understand the point about 1C discharge currents?

rtimms: ah ok. for the 1C thing I was just thinking that for users it might seem strange to provide the typical current in a parameter file, and may be more natural to give the cell capacity as the parameter, which then sets the default typical current (to the 1C discharge current). maybe this is a bit of a non-issue though

tinosulzer: Ok - I guess ideal scenario is you can do either and it updates the other. Isn't most battery data current vs voltage rather than Crate vs voltage though?

rtimms: Even if you are reading in current vs voltage data you still only set one typical current used to non-dimensionlise. As you say, setting either would be best. In any case, I think the default current shouldn't be set to 1A and should instead be set based on the cell capacity (if provided). Then if you are going to be loading in current vs voltage data you might want to change this to a typical current based on your data and then the C-rate changes accordingly (I feel like I'm not explaining this very well haha)

PR #581, opened 2019-08-04
Issue 580 processed variable

### LINKED ISSUE #580: ProcessedVariable for 1+1D models
Allow ProcessedVariable to take an object with domain electrode + current collector and return an object that can be evaluated at any t, x, z

tinosulzer:
# Description

Change ProcessedVariable to be able to plot variables as functions of t, x and z
This is still pretty hacky but seems to work ok for now.
We will need to think of a better way to do this at some point, perhaps by assigning some more attributes to the variables during discretisation (e.g. whether they depend on x, r or z)

Getting a codacy error with `eval(...)` but can't get `ast.literal_eval(...)` to work. How bad is it to use `eval`?

Fixes #580 

## Type of change

# Key checklist:

## Further checks: 


pybamm/expression_tree/parameter.py
```diff
@@ -90,6 +93,27 @@ def get_children_domains(self, children_list):
 
         return domain
 
+    def get_children_auxiliary_domains(self, children):
```
martinjrobins: this looks the same as the function in functions.py. Should it be in the base class (`Symbol`?)

pybamm/processed_variable.py
```diff
@@ -200,37 +200,80 @@ def initialise_2D(self):
...
-        if self.domain in [["negative particle"], ["negative electrode"]]:
+        # Dealt with weird particle/electrode case
+        if self.domain in [
+            ["negative electrode"],
+            ["positive electrode"],
+        ] and self.auxiliary_domains["secondary"] in [
+            ["negative particle"],
+            ["positive particle"],
+        ]:
+            # Switch domain and auxiliary domains and set order to "F"
```
martinjrobins: what is "F"?

pybamm/processed_variable.py
```diff
@@ -200,37 +200,80 @@ def initialise_2D(self):
...
+            dom = self.domain
+            self.domain = self.auxiliary_domains["secondary"]
+            self.auxiliary_domains["secondary"] = dom
+            order = "F"
+        else:
+            order = "C"
```
martinjrobins: what is "C"?

PR #582, opened 2019-08-06
Issue 520 spmecc

### LINKED ISSUE #520: SPMeCC
Write a script which solves algebraic eqns in current collector to compute effective resistance `R_cc`, then solves SPMe, adding the extra current collector ohmic loss to the terminal voltage

rtimms:
# Description

- Adds model to compute effective resistance of current collector 
- Updates how BCs are handled in finite elements
- Add new solver for algebraic only systems (basically the root finding algo used to get consistent ICs in DAE solver
- Adds comparison of 2plus1D and 1D models
- Adds 2+1D DFN model

Fixes #520 

## Type of change

# Key checklist:

## Further checks: 


rtimms:
changing `html_static_path = ["_static"]` to `html_static_path = []` in `conf.py` seems to fix this (see [here](https://github.com/readthedocs/readthedocs.org/issues/1776)) @martinjrobins I don't really understand what this does or if changing it will cause another problem? also, I'm not sure why it broke in the first place?

pybamm/spatial_methods/finite_volume.py
```diff
@@ -225,6 +263,8 @@ def definite_integral_matrix(self, domain):
         ----------
         domain : list
             The domain(s) of integration
+        vector_type : str, optional
+            Whether to return a row or column vector (defualt is row)
```
tinosulzer: this isn't an input to `definite_integral_matrix`. I'm not entirely clear on the difference between `definite_integral_vector` and `definite_integral_matrix`?

results/2plus1D/compare_lead_acid_1plus1D.py
```diff
@@ -0,0 +1,74 @@
...
+    )
+
+# plot terminal voltage
+t = np.linspace(0, solution.t[-1], 100)
+for i, model in enumerate(models):
+    plt.plot(times[i](t), voltages[i](t), lw=2, label=model.name)
+plt.xlabel("Time [h]", fontsize=15)
+plt.ylabel("Terminal voltage [V]", fontsize=15)
+plt.legend(fontsize=15)
+plt.show()
```
tinosulzer: nice, thanks! did you try 2+1 lead acid?

pybamm/__init__.py
```diff
@@ -95,6 +95,8 @@ def version(formatted=False):
     BoundaryGradient,
     Integral,
     IndefiniteIntegral,
+    DefiniteIntegralVector,
```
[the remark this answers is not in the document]
rtimms: the matrix is the same, but (I think) Integral then takes the matrix vector product with something (e.g. 1) then that broadcasts the result, which is different

pybamm/spatial_methods/finite_volume.py
rtimms: yeah I think I can get rid of `definite_integral_vector` here and just pass the row/column input to `definite_integral_matrix`

results/2plus1D/compare_lead_acid_1plus1D.py
rtimms: haven't tried but I can!

PR #596, opened 2019-08-16
Issue 493 thermal curr coll

### LINKED ISSUE #493: Thermal equations in current collectors
Add the thermal equations in the current collectors so that the 3D effects can be captured.

rtimms:
# Description
Adds thermal equations in the current collectors. Merge after #582 

Fixes #493 

## Type of change

# Key checklist:

## Further checks: 


pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -398,11 +403,49 @@ def build_model(self):
     def set_thermal_submodel(self):
 
         if self.options["thermal"] is None:
-            thermal_submodel = pybamm.thermal.Isothermal(self.param)
+            if self.options["dimensionality"] == 0:
+                thermal_submodel = pybamm.thermal.Isothermal(self.param)
+            elif self.options["dimensionality"] == 1:
+                thermal_submodel = pybamm.thermal.current_collector.Isothermal2D(
```
Scottmar93: Should this be `Isothermal1plus1S` to be consistent with `Full1plus1D` and `Lumped1plus1D` ?

pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -398,11 +403,49 @@ def build_model(self):
...
+                    self.param
+                )
+            elif self.options["dimensionality"] == 2:
+                thermal_submodel = pybamm.thermal.current_collector.Isothermal3D(
```
Scottmar93: Similarly `2plus1D`

pybamm/models/submodels/thermal/current_collector/base_current_collector_thermal.py
```diff
@@ -0,0 +1,220 @@
...
+        T_s = pybamm.PrimaryBroadcast(T_av, ["separator"])
+        T_p = pybamm.PrimaryBroadcast(T_av, ["positive electrode"])
+        T_cp = T_av
+
+        # Note: T is the temperature defined only in the negative electrode,
+        # separator and positive electrode. This is the quanitity used in other
+        # submodels (e.g. electrolyte)
+        T = pybamm.Concatenation(T_n, T_s, T_p)
+
+        T_volume_av = self._yz_average(T_av)
```
Scottmar93: This will only be valid for the lumped thermal model then? (this is the most relevant one). But, do you think that if we wanted to do the full thermal model, we would need two separate current collector?

pybamm/models/submodels/thermal/current_collector/base_current_collector_thermal.py
```diff
@@ -0,0 +1,220 @@
...
+
+        q = self._flux_law(T)
+
+        variables = {
+            "Negative current collector temperature": T_cn,
```
Scottmar93: There is quite a bit of repeated code here from `base_thermal`. Its fine for now, but I think should be able to inherit from `base_thermal`

pybamm/models/submodels/thermal/current_collector/current_collector_full_thermal.py
```diff
@@ -0,0 +1,138 @@
...
+class Full1plus1D(BaseNplus1D):
+    """Class for 1+1D thermal submodel"""
+
+    def __init__(self, param):
+        super().__init__(param)
+
+    def set_rhs(self, variables):
+        T_av, _, Q_av = self._unpack(variables)
+        self.rhs = {
+            T_av: (
```
Scottmar93: I thought full thermal would require the boundary temperature from the "whole_cell" domain?

pybamm/models/full_battery_models/base_battery_model.py
rtimms: yeah probably would be more consistent that way, but the model is still valid in full 2 or 3D were we to ever implement that. I'll change to Nplus1D for now though so it fits more clearly with the model structure

pybamm/models/submodels/thermal/current_collector/base_current_collector_thermal.py
rtimms: all of the Nplus1D models are in the limit where T doesn't depend on x (this is partly why I did a different base class for the current collector stuff, since its fundamental variable is really the x-averaged temperature. I think to do the full model (i.e. T a function of x, y and z) we would need to put in a mesh for each current collector so you can apply BCs at the edges of the current collector

pybamm/models/submodels/thermal/current_collector/current_collector_full_thermal.py
rtimms: "full" here refers to T just being a function of y and z, so the BCs "left" and "right" correspond to the negative and positive tabs (here `T_av` lives on the current collector domain)

Scottmar93: Ah that makes sense.

pybamm/models/submodels/thermal/current_collector/base_current_collector_thermal.py
Scottmar93: Ah okay so all current collector thermal models are x-lumped models. Whereas "lumped-model" here means "x-y-z lumped"

Scottmar93: I think with regards to inheritance structure would be somthing like: 

thermal/
  base_thermal.py
  x-lumped/
    base_x_lumped.py
    1D_x_lumped.py
    (current collector models)
  x-full/
    1D_full.py 
    
?

rtimms: yeah that structure would work. yeah I was taking lumped to be equivalent to volume averaged I guess. I think the structure you've put here in nice though. happy to leave merging this for now, and merge after restructuring

Scottmar93: I am making these changes on this branch now. Basic structure is implemented but still need to change elsewhere to pass tests.

PR #601, opened 2019-08-20
#483 change parameter names

### LINKED ISSUE #583: Better dimension names
Change name  of parameters `L_x`, etc from "width" to "thickness"
Change name of parameters `L_y`, etc from "depth" to "width"

tinosulzer:
# Description

Change `L` variable names to thickness and `W` variable names to width
Fixes #583 

## Type of change

# Key checklist:

## Further checks: 


tinosulzer:
@rtimms @Scottmar93 @ferranbrosa this doesn't really need review but just checking you don't disagree with this renaming / are aware in case adding different parameter files

rtimms:
I’m happy with this

PR #608, opened 2019-09-05
Issue 607 fast particle

### LINKED ISSUE #607: Fast diffusion in particles submodel
Fill out the submodel for fast diffusion in particles (i.e. ODEs not PDEs for concentration)

rtimms:
# Description

Adds submodel for fast diffusion in the particles. 

As a separate issue, we should allow for Broadcasts onto cell edges. This is needed here, in `constant_stefan_maxwell_diffusion` and in `isothermal` models. At the moment these all broadcast zero flux to the appropriate domain, but this ends up being evaluated on cell centers, not edges.

Fixes #607 

## Type of change

# Key checklist:

## Further checks: 


tests/integration/test_models/test_full_battery_models/test_lithium_ion/test_dfn.py
```diff
@@ -86,6 +86,12 @@ def test_lumped_thermal(self):
         modeltest = tests.StandardModelTest(model, var_pts=var_pts)
         modeltest.test_all()
 
+    def test_particle_fast_diffusion(self):
+        options = {"particle": "fast diffusion"}
+        model = pybamm.lithium_ion.SPM(options)
```
martinjrobins: Don't think this should be in the testing for DFN

tests/integration/test_models/test_full_battery_models/test_lithium_ion/test_spme.py
```diff
@@ -78,6 +78,12 @@ def test_thermal(self):
         modeltest = tests.StandardModelTest(model)
         modeltest.test_all()
 
+    def test_particle_fast_diffusion(self):
+        options = {"particle": "fast diffusion"}
+        model = pybamm.lithium_ion.SPM(options)
```
martinjrobins: should be SPME here?

pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -146,8 +146,9 @@ def options(self, extra_options):
             "side reactions": [],
             "interfacial surface area": "constant",
             "current collector": "uniform",
+            "particle": "Fickian diffusion",
```
martinjrobins: all these options should be documented in the doctring

rtimms:
Thanks @martinjrobins . I've started to add the options to the docstring for `BaseBatteryModel`. I wasn't sure of the correct/best way to do this, or how much detail about each of the options should be added there. I guess one option is to briefly explain what each selection you make in an option does, and another would be to give what particular submodel that option loads and then direct to the docs there for detail.

rtimms:
Also I wasn't sure of the proper way to document things like `concatenated_rhs` in `BaseModel`. These are `None`, but then get replaced with various different thing during discretisation

martinjrobins:
> Also I wasn't sure of the proper way to document things like `concatenated_rhs` in `BaseModel`. These are `None`, but then get replaced with various different thing during discretisation

What you have done looks good. Even though they are set to `none` in the base class, you just explain the purpose and semantic meaning of each variable so that those implementing derived classes know how they should be using them

martinjrobins:
happy to merge when you are ready @rtimms

PR #610, opened 2019-09-05
Issue 576 windows readme

### LINKED ISSUE #576: step by step instructions for windows users
- based on WSL - describe how WSL works
- virtual-env, how to install. remember to re-activate
- how to install scikits.odes dependencies

based on the exact order

martinjrobins:
# Description

Fixes #576 

I've split the installation instructions to separate files for linux and windows. We should probably create another issue for mac install instructions....

Can someone who has not yet install WSL run through these and check that they work? Doesn't need everyone to do it, just the first person with some time and Windows installed :) I've already got everything installed so I'm not sure if I missed anything.

## Type of change

# Key checklist:

## Further checks: 


ferranbrosa:
There is still the issue of showing the plots in Windows which requires following [these instructions ](https://stackoverflow.com/questions/43397162/show-matplotlib-plots-in-ubuntu-windows-subsystem-for-linux)and changing the backend of matplotlib. I am now on holiday but when I am back I will try to get it up and running in another computer and write the instructions down.

martinjrobins:
Thanks @ferranbrosa! I tried running the example script `examples/scripts/create_model.py`, which generates a plot, in the python interactive window in vscode. This worked perfectly without having to install an x server in windows. However, it would be useful to be able to have separate windows for plotting, so I think these instructions would be great to include.

tinosulzer:
`cmake` should be added to the list of requirements for installation (doesn't come with WSL).

Also we might want to have a troubleshooting page where we say how to resolve common errors (installation and otherwise). Some common errors (that I have done) might be:
- forgetting to use `-e` in `pip install -e .`
- not installing with dev and docs
- forget to install blas/lapack

ferranbrosa:
I finally got a Windows computer into which make a fresh installation and I have a few comments about the current instructions. Some of them are very basic things but I am thinking of users with very little expertise in Python which maybe only want to run some scripts to reproduce plots from papers.

- We should put an example on how to **install packages with pip** (now we just say that numpy, scipy... need to be installed using pip).
- **Scikits.fem** is already installed with PyBaMM so the instructions in optional dependencies are redundant.
- **Optional dependencies**: we should at least explain where are they used so people can decide if they want to install them. Right now, and given the point above, the only optional dependency is scikits in order to solve the DFN model. In my opinion, given the importance of the model, that shouldn't be an optional feature. Also, I would write down the command to install all the prerequisites given that it is only one command line (and, especially for OpenBLAS, it required some googling). It should be:
```sudo apt-get install python3-dev gcc gfortran libopenblas-dev cmake```
- **Sundials**:  given that we give all the steps, I would not put it as a prerequisite. The steps worked fine, but we can remove the folder `build-sundials-3.1.1` and the tar.gz file when we remove  `sundials-3.1.1`. Also, I have noticed that the following warning appeared when installing Sundials:
```
WARNING: EXAMPLES_ENABLE is deprecated and will be removed in the future.
Copying value to EXAMPLES_ENABLE_C.
```
- I think it would be good to suggest the user to **run the tests**, just to make sure installation worked well.

I have also set VSCode to launch windows for plotting. It requires the following:
1. Install VcXsrv from [here](https://sourceforge.net/projects/vcxsrv/).
2. Set the display port: `echo "export DISPLAY=localhost:0.0" >> ~/.bashrc`
3. Install python3-tk: `sudo apt-get install python3-tk`
4. Set the matplotlib backend to TKAgg: `echo "backend : TKAgg" >> ~/.config/matplotlib/matplotlibrc`
5. Before running the code, just launch XLaunch (with the default settings). Then the code works as usual.

Given that I had to play around a little bit until I got it running, could someone check that the steps I described work properly?

tinosulzer:
Thanks, Ferran, those instructions at the end worked. I've made changes to everything except the comments on optional dependencies and the sundials install instructions

PR #612, opened 2019-09-06
604 allow BCs set on neg and pos tabs

### LINKED ISSUE #604: Allow for different tab locations in 1+1D
At present the 1plus1D current collector model assumes both tabs are at the top (regardless of the tab locations set in the parameters). Also, in 2plus1D we use "left" and "right" to correspond to the negative and positive tab, respectively. It would be cleaner to be able to set BCs on "negative tab" and "positive tab". It would also be useful to then have something like `pybamm.BoundaryValue(var, "negative tab")` which can be used to get e.g. terminal voltage.

rtimms:
# Description
Allows setting boundary conditions and taking boundary value on tabs. Currently this is implemented for 1D meshes by having `discretisation` change "negative tab" and "positive tab" to "left" or "right" depending on the tab location, which is now stored in the mesh. The benefit of doing it this way is that when implementing any new 1D spatial methods you just deal with left and right, and don't worry about catching tabs too. There may be benefits to doing this in a different way though? In 2D the spatial method deals directly with the strings  "negative tab" and "positive tab" since "left" and "right" don't really make sense in this context, and were always taken to mean the tabs. 

What are your thoughts on this @Scottmar93 @tinosulzer @martinjrobins ?

Fixes #604 

## Type of change

# Key checklist:

## Further checks: 


rtimms:
Just realised this doesn't actually work for the case where both tabs are on the same side... would also need an option along the lines of applying BC on a tab and *not* on a tab, if that makes sense? Will think of a better solution today

PR #627, opened 2019-09-19
Issue 623 soc enhancement

TomTranter:
# Description

Calculate SOC.
Introduce r_average operator

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks: 


examples/scripts/SPMe_SOC.py
```diff
@@ -0,0 +1,110 @@
+import pybamm
```
rtimms: could you add a very short description of what the script is doing here?

pybamm/models/submodels/particle/base_particle.py
```diff
@@ -26,12 +26,16 @@ def _get_standard_concentration_variables(self, c_s, c_s_xav):
...
 
         if self.domain == "Negative":
             c_scale = self.param.c_n_max
+            active_volume = geo_param.a_n_dim * geo_param.R_n / 3
         elif self.domain == "Positive":
             c_scale = self.param.c_p_max
-
+            active_volume = geo_param.a_p_dim * geo_param.R_p / 3
+        c_s_r_av = pybamm.r_average(c_s_xav)
+        c_s_r_av_vol = active_volume * c_s_r_av * c_scale
```
rtimms: maybe it would be better here to define `c_s_r_av_vol = active_volume * c_s_r_av`, then times by `c_scale` in the variable dict to get the dimensional version? just to stick with the convention of the variables being non-dimensional?

TomTranter:
@rtimms @tinosulzer I made the example script a bit longer. You may want to put this in results instead? It's good to go otherwise

PR #631, opened 2019-09-20
#630 fixes to thermal equations

### LINKED ISSUE #630: Discrepancy between 1D lumped and full thermal models
**Describe the bug**
The models `x_full_no_current_collector` and  `x_lumped_no_current_collector` do not agree. This is because the lumped model is taking the average over the whole cell (including current collectors) when calculating the source terms, resulting in an under prediction.

rtimms:
# Description
Changes the no current collector lumped model to use the correct average (not including current collectors). Also fixes some small typos/missing parameters in other equations.

Fixes #630

## Type of change

# Key checklist:

## Further checks: 


tinosulzer:
It won't let me add a review on line 183 in `base_thermal.py`, but could you create new domains called "negative current collector" and "positive current collector" and then do `Broadcast(var_cn, "negative current collector")`? Admittedly it might be confusing then what is the difference between negative current collector, positive current collector, and current collector

rtimms:
yeah that could work. would you also then change the potentials to have `"negative current collector"` and also set auxiliary domains in this way? Alternatively, do you think it would be likely to cause issues if we allowed concatenations of variables whose domain is `"current collector"`?

rtimms:
thanks @tinosulzer . I think adding separate domains for negative and positive current collector might be something worth looking at in the future. If people aren't familiar with how we've formulated the 2+1D stuff maybe it is confusing just having one current collector domain... something to think about anyway.

PR #639, opened 2019-09-30
Issue 617 nonuniform mesh

### LINKED ISSUE #617: Add non-uniform meshes
Add non-uniform meshes (e.g. more points at left or right boundary, or both). In particular, in 2+1D this may help to reduce the number of points needed in the current collector domain

rtimms:
# Description
Adds some extra meshes. Some of them require parameters to be passed, so I've made classes like `GetUserSupplied1DSubMesh` which takes in the required parameters, then `__call__` creates an instance of `UserSupplied1DSubMesh` during mesh creation. I'm not sure if this is the correct/best way to do this?

Fixes #617 

## Type of change

# Key checklist:

## Further checks: 


rtimms:
yeah I felt a bit uneasy about submesh_types having a mix of mesh classes and generators. happy to go with the suggestion of making everything a generator. at least that way it is very clear to the user whether you are passing parameters to a mesh or not. 

yeah I think passing everything to the User mesh is a bit odd, but I think it is a good way of checking the users supplied edges make sense (well at least passing lims prevents the case where the user doesnt put edges on the boundaries of the geometry) I couldnt figure out a sensible/non-hacky way to get npts automatically from the mesh rather than the other way around

pybamm/meshes/one_dimensional_submeshes.py
```diff
@@ -30,10 +31,62 @@ def near(x, point, tol=3e-16):
...
+    submesh_type: str, optional
+        The type of submeshes to use. Can be "Uniform", "Exponential", "Chebyshev"
+        or "User". Default is "Uniform".
+    submesh_params: dict, optional
+        Contains any parameters required by the submesh.
+
+    **Extends**: :class:`pybamm.MeshGenerator`
+    """
+
+    def __init__(self, submesh_type="Uniform", submesh_params=None):
```
Scottmar93: submesh_type could be the class itself e.g. Uniform1DSubMesh. And all Submesh types can have inputs lims, npts, tabs, submesh_params. Then we just have: 
```python
def __call__(self, lims, npts, tabs=None)
return self.submesh_type(lims, npts, tabs, self.submesh_params)
``` 
Just to have fewer add/maintain if new meshes are added in the future

pybamm/meshes/zero_dimensional_submesh.py
```diff
@@ -33,3 +35,32 @@ def __init__(self, position, npts=None):
...
+
+class MeshGenerator0D(MeshGenerator):
+    """
+    A class to generate a submesh on a 1D domain.
+
+    Parameters
+    ----------
+
+    submesh_type: str, optional
+        The type of submeshes to use. Can be "Position". Default is "Position".
```
Scottmar93: Maybe a little more documentation on what is meant by "Position" here. Do we just mean a spatial mesh? or something else

pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -140,18 +140,18 @@ def default_var_pts(self):
     @property
     def default_submesh_types(self):
         base_submeshes = {
-            "negative electrode": pybamm.Uniform1DSubMesh,
-            "separator": pybamm.Uniform1DSubMesh,
-            "positive electrode": pybamm.Uniform1DSubMesh,
-            "negative particle": pybamm.Uniform1DSubMesh,
-            "positive particle": pybamm.Uniform1DSubMesh,
+            "negative electrode": pybamm.MeshGenerator1D(),
```
Scottmar93: I think this bit makes it a little unclear as to what mesh is being used in each domain. The following might reverse what we are trying to do though.. Maybe we could leave the mesh types as `pybamm.UniformSubMesh` etc but in somewhere else (maybe in `pybamm.Mesh`) create a dict of MeshGenereators and then use that dict. e.g.
In model:
```python
submesh_types = {"negative electrode": pybamm.UniformMesh1D, etc} 
```
In Mesh: 
```python
for domain in submesh_types.keys():
    if isinstance(submesh_types[domain], Submesh1D):
       submesh_generator[domain] = MeshGenerator1D(submesh_types[domain]) etc
    elif isinstance(submesh_types[domain], Submesh2D):
       submesh_generator[domain] = MeshGenerator2D(submesh_types[domain]) etc
```

Scottmar93: Alternative would be to leave the argument to MeshGenerator undefined

tinosulzer:
Have just skimmed over this but it looks good to me if you address Scott's comments and overload `__repr__` to something like
```python3
def __repr__(self):
     return "Generator for {}".format(submesh.__repr__) # <-- or something else like this
```

tinosulzer:
also remember to update Changelog now :)

PR #641, opened 2019-10-01
Issue 638 process post solve

### LINKED ISSUE #638: Define and evaluate expressions post solve
It would be good to be able to define new expressions from existing model variables after the solve. Then process and discretise them. In the first instance this could probably be done by allowing parameters, geometry and discs to be passed to the method `process_parameters_and_discretise`. E.g. after the normal model set up an solve do

``` 
c = model.variables["Concentration"]
my_expr = pybamm.grad(c)
my_expr_disc = model.process_parameters_and_discretise(my_expr)
```

now `my_expr_disc` is just like any other entry in `model.variables`

In doing this we could probably reduce the number of variables calculated by default (e.g. maybe we only want to calculate actual independent variables and not their gradient)

rtimms:
# Description

Changes `model.process_parameters_and_discretise` to accept parameter values and a discretisation so that users can define new expressions in scripts to be processed and discretised. 

Fixes #638 

## Type of change

# Key checklist:

## Further checks: 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 # don't ignore important .txt files
 !requirements*
 !LICENSE.txt
+!CMakeLists.txt
 
 # running files
 *.pyc
@@ -64,3 +65,5 @@ pyvenv.cfg
 
 # sundials
 sundials
+sundials4
+SuiteSparse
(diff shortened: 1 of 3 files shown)


rtimms:
yeah exactly. think it would be a very tidy way of keeping track of things

PR #643, opened 2019-10-01
Issue 606 change submodel notebook

### LINKED ISSUE #606: Example notebook showing changing submodel
Add an example notebook showing how to load up a default model (e.g. SPM) and swap out a submodel, e.g. ODEs for particle concentration (fast diffusion) instead of Fickian. Also show how to build up your own model from submodels.

rtimms:
# Description
Adds a notebook which shows users how to swap out a submodel in an existing model, and how to build a new model using existing submodels. I've added an extra option `build=True` to the models so that you can load up a model without building (e.g. `pybamm.lithium_ion.SPM(build=False)`) then change a submodel before running `model.build_model()`. 

While doing this I though users might find it annoying that they have to be careful about which order they set the submodels in (for getting the coupled variables). I've changed the way `base_battery_model.build_model`  sets the coupled variables for the submodels . It now runs through them in order, but passes over KeyErrors. It then makes multiple passes, attempting to set the coupled variables for any that failed. If it gets to the stage where there are no submodels left to try and it still gets a KeyError then it throws an error. The way I've implemented this probably (definitely?) isnt the best. Does anyone envisage any problems in doing this? Or have suggestion for a cleaner implementation? 

Fixes #606 

## Type of change

# Key checklist:

## Further checks: 


examples/notebooks/README.md
```diff
@@ -52,6 +56,7 @@ The following models are implemented and can easily be used or [compared](./mode
 
 The following spatial method is implemented
```
tinosulzer:
```suggestion
The following spatial methods are implemented
```

pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -424,13 +432,37 @@ def build_model(self):
...
-                "Getting coupled variables for {} submodel ({})".format(
-                    submodel_name, self.name
-                )
-            )
-            self.variables.update(submodel.get_coupled_variables(self.variables))
+        # Note: pybamm will try to get the coupled variables for the submodels in the
+        # order they are set by the user. If this fails for a particular submodel,
+        # return to it later and try again. If setting coupled variables fails and
+        # there are no more submodels to try, raise an error.
+        submodels = list(self.submodels.keys())
```
tinosulzer: this is neat!

PR #645, opened 2019-10-03
#644 set atol and rtol

### LINKED ISSUE #644: set relative and absolute tolerances separately for solver
At the moment we use the same tol for both. Would be good to have additional control over these. Also, the defaults are pretty tight so we might relax them a bit to get a good balance of speed and accuracy.

rtimms:
# Description

Allows the user to set `rtol` and `atol` separately. For the defaults I've selected 1e-3 for rtol and 1e-6 for atol. This is the same as other solvers (e.g. scipy.integrate or MATLABs ODE15s) and provides a pretty significant speedup vs 1e-8 for both. I've also reduced the default number of points in x, as again this speed things up quite a bit. Both of these result in a RMS voltage error of around 1e-4 compared with the previous defaults, but the default DFN now runs in around 3 seconds vs 23 seconds (on my computer at least).

I think for most users, this kind of error is ok out of the box (the plots are indistinguishable), particularly if they are new users and just want to quickly run some examples. Users now have more control over both solver tolerances, so I think this is an ok compromise.

EDIT: I've decreased rtol to 1e-6 too. Seems like the thermal model and some other models didn't like rtol as low as 1e-3. Probably safer to have both set as 1e-6 so that the solvers don't immediately  fall over when people add new physics.

Fixes #644 

## Type of change

# Key checklist:

## Further checks: 

### DIFF
--- a/tests/integration/test_models/standard_model_tests.py
+++ b/tests/integration/test_models/standard_model_tests.py
@@ -71,6 +71,10 @@ class StandardModelTest(object):
         # Overwrite solver if given
         if solver is not None:
             self.solver = solver
+        else:
+            # use tighter default tolerances for testing
+            self.solver.rtol = 1e-8
+            self.solver.atol = 1e-8
         if t_eval is None:
             t_eval = np.linspace(0, 1, 100)
 
--- a/tests/integration/test_models/standard_model_tests.py
+++ b/tests/integration/test_models/standard_model_tests.py
@@ -71,10 +71,10 @@ class StandardModelTest(object):
         # Overwrite solver if given
         if solver is not None:
             self.solver = solver
-        else:
-            # use tighter default tolerances for testing
-            self.solver.rtol = 1e-8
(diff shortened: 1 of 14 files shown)


tests/integration/test_models/standard_model_tests.py
```diff
@@ -71,6 +71,10 @@ def test_solving(self, solver=None, t_eval=None):
         # Overwrite solver if given
         if solver is not None:
             self.solver = solver
+        else:
```
tinosulzer: should this be else or just default behaviour?

rtimms: yeah probably should just be default. 

I agree, I think a separate repository on pybamm-team with paper results and any undocumented (or poorly documented) examples would be good. Then make sure the examples that come as part of pybamm are actually really useful for learning how to use it/showcasing feature, rather than just a relatively random collection of scripts we make during development.

PR #647, opened 2019-10-07
Issue 563 parameters by material

### LINKED ISSUE #563: add parameters by material
add parameters by choosing materials e.g. `pybamm.ParameterValues(cathode=...,anode=...)` then have library which allows user to pick and choose different materials

tinosulzer:
# Description

- [x] Reorganised parameters by material.
- [x] Specified path to functions in csv files
- [x] Allowed setting either current or C-rate
- [x] Updated docs on how to add new parameters

Fixes #563 

## Type of change

# Key checklist:

## Further checks: 


rtimms:
some comments for discussion:

- I think there could be a "geometry" or "cell" file which contains the things like cell geometry and maybe other cell-scale parameters like voltage cut-off or number of cells to make a battery. Maybe nominal cell capacity as stated by manufacturer. From this we then define the typical current to be 1C based on the nominal cell capacity.

- For li-ion we will need a set of thermal parameters. The structure could be the same for lead-acid, with a boring file with the ref and max temp in

- It would be good to have the ability to just pass the name of a paper to quickly set all the parameters from that paper 

- I think a misc params file is probably ok, but should also be named with author_year 

- maybe not for this issue, but for effective properties e.g. conductivity it would be nice to select how they are calculated e.g. bruggeman or otherwise. At the moment we hard-code bruggeman into the models

- I think this has already been discussed, but where is the best place to set initial conditions?

ferranbrosa:
I like @Scottmar93 suggestion in #563 about defining an experiment. It is good to distinguish between the following: battery parameters (intrinsic of the battery materials or, from a maths point of view, of the equations), which could be split into further subcategories; geometry and those related to the experiment (how we cycle the battery). This poses some interesting questions about defining the SoC of the electrodes (or maybe we should use state of lithiation to avoid confusion) as they are related to the potentials. 

I agree also with @rtimms points above. In particular, I think that we should define a new parameter/function that accounts for the effectiveness of the medium, so then one can choose between different models. In particular, I think it would be interesting to use the values obtained via homogenisation using Rayleigh's multipole method (see Bruna & Chapman 2015).

docs/tutorials/add-parameter-values.rst
```diff
@@ -16,32 +16,61 @@ At the stage of creating a model, we use :class:`pybamm.Parameter` and :class:`p
...
+
+    param = pybamm.ParameterValues(
+        chemistry={
+            "chemistry": "lithium-ion",
+            "cell": "kokam_Marquis2019",
+            "anode": "new_anode_chemistry_Bloggs2019",
+            "separator": "separator_Marquis2019",
+            "cathode": "lico2_Marquis2019",
+            "electrolyte": "lipf6_Marquis2019",
+            "experiment": "1C_discharge_from_full_Marquis2019",
```
rtimms: Maybe there should be a very brief description of the parameters that go into the experiment folder? Though maybe it is clear from looking at the existing parameter sets.

pybamm/parameters/parameter_sets.py
```diff
@@ -0,0 +1,49 @@
+#
+# Parameter sets from papers
```
rtimms: this is very user-friendly!

docs/tutorials/add-parameter-values.rst
ferranbrosa: I also think that a list of the required parameters in the documentation would be helpful.

PR #657, opened 2019-10-10
Issue 633 add klu

### LINKED ISSUE #633: Add KLU solver interface
**Summary**
Add sundials with KLU linear solver interface using pybind11

Scottmar93:
# Description

Added pybind11 interface to sundials with the sparse linear solver KLU works for all models including 2p1D now. Still fairly simple requiring the jacobian and events to be passed in but we can add these as options later. Additional things added: 
- a unit test to ensure it works on one of the standard examples included in sundials (ida_roberts_klu)
- an integration test to ensure it works on the spme (in 1D). (I didn't add the 2p1D case because that takes too long)
- an install script in "scripts/install_sundials_4.1.0.sh" that downloads sparsesuite, sundials4.1.0, sets up pybind11, and then generates the klu solver shared library so that it can be imported in python
- some simple instructions in the README.md for how to install using the install script
- added KLU to travis.yml (I think I have done this right but please check) (Unfortunately this will add quite a bit of time to tests because needs to download and install SuiteSparse which takes a while). 

Fixes #633 

##  Type of Change

# Key checklist:

## Further checks: 


README.md
```diff
@@ -28,6 +28,33 @@ For instructions on installing PyBaMM on Debian-based distributions, please see
...
+./scripts/install_sundials_4.1.0.sh
+```
+Note that this script has only been tested on Ubuntu 18.04.3 LTS. If you run into issues on 
+another distribution, we recommend you first have a look through `install_sundials_4.1.0.sh` 
+as it may be relatively simple to modify this for your purposes. 
+
+In principle, the install should be possible in other operating systems by following the same
+process as performed in `install_sundials_4.1.0.sh`. Although there may be some issues on 
+Windows in building SuiteSparse as the only build option is via a Makefile. 
+
```
tinosulzer: add this to `INSTALL-LINUX.md` instead. Also not super clear what to do if it doesn't work

SuiteSparse/.gitignore
```diff
@@ -0,0 +1,220 @@
...
+ktruss/allktruss_graphblas_main
+ktruss/allktruss_main
+ktruss/ktruss_graphblas_main
+ktruss/ktruss_main
+
+ssget/files/ss_index_old.mat
+ssget/files/ssstats_old.csv
+
+# Do not ignore this file
+!.gitignore
```
tinosulzer: why can't you just ignore SuiteSparse (and then not need this file)?

tests/unit/test_solvers/test_klu_sparse_solver.py
```diff
@@ -0,0 +1,87 @@
+#
+# Tests for the KLU Solver class
+#
+import pybamm
+import numpy as np
+import scipy.sparse as sparse
+import unittest
+
+
+@unittest.skipIf(pybamm.have_klu(), "klu is not installed")
```
tinosulzer:
```suggestion
@unittest.skipIf(pybamm.have_klu(), "klu solver is not installed")
```

tests/integration/test_solvers/test_klu.py
```diff
@@ -0,0 +1,30 @@
+import pybamm
+import numpy as np
+import sys
+import unittest
+
+
+@unittest.skipIf(pybamm.have_klu(), "klu is not installed")
```
tinosulzer:
```suggestion
@unittest.skipIf(pybamm.have_klu(), "klu solver is not installed")
```

pybamm/solvers/klu_sparse_solver.py
```diff
@@ -0,0 +1,169 @@
...
+    def __init__(
+        self, rtol=1e-6, atol=1e-6, root_method="lm", root_tol=1e-6, max_steps=1000
+    ):
+
+        if klu_spec is None:
+            raise ImportError("KLU is not installed")
+
+        super().__init__("ida", rtol, atol, root_method, root_tol, max_steps)
+
+    def integrate(self, residuals, y0, t_eval, events, mass_matrix, jacobian):
```
rtimms: does this work if events and jacobian are None? If not, could you raise an exception when they arent passed

examples/scripts/compare-dae-solver.py
```diff
@@ -0,0 +1,38 @@
...
+mesh = pybamm.Mesh(geometry, model.default_submesh_types, var_pts)
+
+# discretise model
+disc = pybamm.Discretisation(mesh, model.default_spatial_methods)
+disc.process_model(model)
+
+# solve model
+t_eval = np.linspace(0, 0.2, 100)
+
+klu_sol = pybamm.KLU(tol=1e-8).solve(model, t_eval)
```
rtimms: think you need to pass atol and rtol separately here

rtimms:
>   The source directory
> 
>     /mnt/c/Users/vsulzer/Documents/Batteries/PyBaMM/third-party/pybind11
> 
>   does not contain a CMakeLists.txt file.
> 
> 
> CMake Error at CMakeLists.txt:9 (pybind11_add_module):
>   Unknown CMake command "pybind11_add_module".
> 
> 
> -- Configuring incomplete, errors occurred!
> See also "/mnt/c/Users/vsulzer/Documents/Batteries/PyBaMM/CMakeFiles/CMakeOutput.log".
> make: *** No targets specified and no makefile found.  Stop.
> rm: cannot remove 'cmake_install.cmake': No such file or directory
> ```

I also got this error

SuiteSparse/.gitignore
Scottmar93: Cheers for noticing this. I did add SuiteSparse to the gitignore so I am not sure why this gitignore has been added

README.md
Scottmar93: I'll copy across the various components of the install script into INSTALL-LINUX.md and talk people through what is happening. That should help at least break down what is going on so that it's easier to figure out what part is breaking.

Scottmar93:
Install error should be fixed now. I was doing: 
```
if cd pybind11; then git pull; else git clone "pybind11 repo name" 
```
but i think the pybind11 folder was already initially there so when you git pulled nothing happened. I have replaced with
```
rm -rf pybind11
git clone "pybind11 reppo"
```
which is slightly less efficient as could potentially remove repo that is already there but its so quick that it doesn't really matter.

rtimms:
> another thing is KLU should be the default solver where available:
> 
> ```
> if pybamm.have_klu():
>    return pybamm.KLUSolver()
> else:
>    # whatever it was before
> ```

do we want this to be the case for all models? e.g. for 1D DFN with default points the scikits solver is faster (I think?). so maybe if people have both installed they should be set as defaults according to model/options? but maybe this is over complicating things

tinosulzer:
> > another thing is KLU should be the default solver where available:
> > ```
> > if pybamm.have_klu():
> >    return pybamm.KLUSolver()
> > else:
> >    # whatever it was before
> > ```
> 
> do we want this to be the case for all models? e.g. for 1D DFN with default points the scikits solver is faster (I think?). so maybe if people have both installed they should be set as defaults according to model/options? but maybe this is over complicating things

good point, I think it's fine to change the default solver based on the options (in this case dimensionality). For example in lead-acid the default solver switches between ODE and DAE depending on whether capacitance is included. Is scikits really that much faster for 1d dfn?

tinosulzer:
Also I've just noticed that `klu.cpp` still says "currently just in a quick and dirty form to get something working" at the top, though it looks pretty clean now to my untrained eye

- We should impose a style for cpp code similar to flake8 for python
- Would be good to get @tlestang or @martinjrobins  to check over this file

Lots of comments but great work getting this working!

CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
+cmake_minimum_required(VERSION 3.10)
+project(klu)
+
+set (CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_FLAGS_RELEASE "-O3")
```
martinjrobins: RELEASE mode should already add this flag I believe.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "third-party/pybind11"]
```
martinjrobins: did you mean to add pybind11 as a submodule here? In the instructions below you ask users to clone it directly, so this is not needed

pybamm/solvers/c_solvers/klu.cpp
```diff
@@ -0,0 +1,429 @@
...
+namespace py = pybind11;
+
+using residual_type = std::function<py::array_t<double>(double, py::array_t<double>, py::array_t<double>)>;
+using jacobian_type = std::function<py::array_t<double>(double, py::array_t<double>, double)>;
+using event_type = std::function<py::array_t<double>(double, py::array_t<double>)>;
+using np_array = py::array_t<double>;
+
+using jac_get_type = std::function<np_array()>;
+
+#pragma GCC visibility push(hidden)
```
martinjrobins: does this work on compilers other than gcc?

pybamm/solvers/c_solvers/klu.cpp
```diff
@@ -0,0 +1,429 @@
...
+
+    int n = python_functions.number_of_states;
+    py::array_t<double> y_np = py::array_t<double>(n, yval);
+    py::array_t<double> yp_np = py::array_t<double>(n, ypval);
+
+    py::array_t<double> r_np;
+
+    r_np = python_functions.res(tres, y_np, yp_np);
+
+    double *r_np_ptr = (double *)r_np.request().ptr;
```
martinjrobins: Note: you can use:
```
    auto r = x.unchecked<1>(); // x must have ndim = 1; can be non-writeable
```
to get access to the array without bounds checking etc, which is faster, see here: https://pybind11.readthedocs.io/en/stable/advanced/pycpp/numpy.html#direct-access

The docs say that this directly references the data, so you might be able to get a raw pointer from this to pass to sundials without copying..?

tinosulzer:
@martinjrobins I assume clang-tidy just does auto-formatting? Is there a good tool for style-checking on travis?

martinjrobins:
clang-tidy is different to clang-format. clang-tidy does static analysis, a whole bunch of tests on your code that are over and above what the compiler does. clang-format autoformats your code according to a particular style. clang-format can probably style-check on travis as well, I've not done this before.

pybamm/solvers/c_solvers/klu.cpp
Scottmar93: This was the suggested thing to do in response to some warning I was getting when compiling. No longer getting it (not sure why) so will just remove this line.

.gitmodules
Scottmar93: Ah okay, should we change the instuctions to: `git submodule update --init --recursive` instead as discussed in https://git-scm.com/book/en/v2/Git-Tools-Submodules ? or should we remove the gitmodules thing?

PR #661, opened 2019-10-14
Issue 600 interpolate

### LINKED ISSUE #600: OCP curves from data
**Summary**
It would be useful to define a function within PyBaMM that, given a data file, produces the interpolating function and also its Jacobian within `autograd`. That would be useful for OCPs, but shouldn't limit to that as some other parameter function might be provided as a dataset of experimental measurements (e.g. the entropic change).

I have already the code that produces this in the main script, but I think it would be a good idea to have it as a PyBaMM function that deals with it behind the scenes. 

 **Motivation**
 In many cases, the parameter functions come from experimental data so interpolation is a straightforward way to include them in the simulations. 

**Additional context**
The code in my main script looks like
```
data_cathode = pd.read_csv("my_data.csv")

interpolated_OCV_cathode = interpolate.CubicSpline(
    data_cathode.to_numpy()[:,0], 
    data_cathode.to_numpy()[:,1], 
    extrapolate=True
)

dOCV_cathode = interpolated_OCV_cathode.derivative()

@primitive
def OCV_cathode(sto):
    out = interpolated_OCV_cathode(sto)
    if np.size(out) == 1:
        out = np.array([out])[0]
    return out

def OCV_cathode_vjp(ans,sto):
    sto_shape = sto.shape
    return lambda g: np.full(sto_shape, g) * dOCV_cathode(sto)

defvjp(OCV_cathode, OCV_cathode_vjp)
```

tinosulzer:
# Description

Add `Interpolant` class to interpolate experimental data (e.g. OCP curves).

Fixes #600 

## Type of change

# Key checklist:

## Further checks: 


input/parameters/lithium-ion/cathodes/lico2_Marquis2019/lico2_data_example.csv
```diff
@@ -0,0 +1,50 @@
+0.000000000000000000e+00 4.714135898019971016e+00
```
ferranbrosa: Maybe we should agree on a format for the data (like headers, where the data comes from...).

Issue #651, opened 2019-10-09
Jacobian by equation


Scottmar93:
**Summary**
Perform the Jacobian calculation on each individual equation and then concatenate the result. 

Also simplify by equation. 

**Reason**
Currently, we discretize equations and then combine into one large concatenated vector. This leads to very large expression tree which are lead to issues with RAM and are hard to interpret. We can instead simply hold off concatenating the discretized equations (i.e. move this step to the solver) and take the Jacobian of each equation individually. This helps with a few things: 
- it might help limit some of the RAM issues we were having on the larger expression trees
- it keeps the expression trees reasonably small and manageable
- we can add a "by equation" progress bar for calculating the Jacobian to help us understand what equations are slow
- we could calculate the Jacobian of different equations in parallel using pythons multiprocessing

rtimms:
something that might help is checking if any state vectors are in the tree in the base method for jac  [here](https://github.com/pybamm-team/PyBaMM/blob/41cb9dc7d5e3fe262067cd77eaa9c9233500ae03/pybamm/expression_tree/symbol.py#L429) and returning zeros of the right shape if not. not sure how often it happens but it might avoid calculating jac down a particular branch unnecessarily. should at least stop the tree for jac becoming so large

tinosulzer:
Can this now be closed @rtimms ?

rtimms:
yep. the functionality is there to do this if calculating the jacobian using the method in `discretisation` rather than during solver set up

PR #670, opened 2019-10-18
Issue 665 jac class

### LINKED ISSUE #665: Jacobian class
Create a class for `Jacobian` similar to `Simplification` that reuses known jacobians

rtimms:
# Description
Adds a class `Jacobian` to re-use know jacobians of any expressions. Also changes the behavior of some binary simplifies which were returning dense matrices of zeros in some cases.

I was going to do Jacobian by equation as part of this (#651) but (I think) that needs to be done in discretisation so you can use `_concatenate_in_order`. I've added a method `disc.create_jacobian` which lets you do compute the jacobian, but by default I've left calculation of jac until the solve. The main reason being that it is much faster to compute the jacobian on the simplified model, but you dont want to simplify as part of discretisation as this prevents you from updating parameters. I've kept the `create_jacobian` method as there may be use cases where you want to compute the jacobian (once) of the unsimplified model, and then go ahead and solve multiple times but with different params etc.

Fixes #665 

## Type of change

# Key checklist:

## Further checks:


pybamm/discretisations/discretisation.py
```diff
@@ -497,15 +501,66 @@ def create_mass_matrix(self, model):
 
         return pybamm.Matrix(mass_matrix)
 
+    def create_jacobian(self, model):
+        """Creates Jacobian of the discretised model.
+        Note that the model is assumed to be of the form M*y_dot = f(t,y), where
+        M is the (possibly singular) mass matrix. The Jacobian is df/dy.
+
```
tinosulzer: Add a comment here referring to this PR to explain why this function isn't used. I could see us forgetting and wondering why we don't use this in future

pybamm/discretisations/discretisation.py
```diff
@@ -696,8 +754,15 @@ def _concatenate_in_order(self, var_eqn_dict, check_complete=False):
         ----------
         var_eqn_dict : dict
             Equations ({variable: equation} dict) to dicretise
+        check_complete : bool, optional
+            Whether to check keys in var_eqn_dict against self.y_slices. Defualt
```
tinosulzer: got some defualt sneaking in again :eyes:

pybamm/expression_tree/binary_operators.py
```diff
@@ -573,17 +592,25 @@ def _binary_evaluate(self, left, right):
...
+                    return pybamm.Matrix(csr_matrix(shape))
+        if is_scalar_zero(right):
+            shape = left.shape_for_testing
+            if shape == ():
                 return pybamm.Scalar(0)
+            else:
+                if len(shape) == 1 or shape[1] == 1:
+                    return pybamm.Vector(np.zeros(shape))
+                else:
+                    return pybamm.Matrix(csr_matrix(shape))
```
tinosulzer: seeing them all together in review mode, these lines appear several times together and could be abstracted into a functon `pybamm.zeros_of_shape(shape)`

pybamm/expression_tree/jacobian.py
```diff
@@ -0,0 +1,46 @@
...
+            self._known_jacs[symbol.id] = jac
+            return jac
+
+    def _jac(self, symbol, variable):
+        """ See :meth:`Jacobian.jac()`. """
+        jac = symbol._jac(variable)
+        # jacobian removes the domain(s)
+        jac.domain = []
+        jac.auxiliary_domains = {}
+        return jac
```
tinosulzer: In order to reuse already-calculated jacobians, this function needs to keep calling its own `jac` method. For example, for a binary operator, call `Jacobian.jac` on left and right, and then call `symbol._jac` on processed_left and processed_right:
```python3
processed_left = self.jac(symbol.left)
processed_right = self.jac(symbol.right)
return symbol._jac(processed_left, processed_right)
```
Otherwise a new `Jacobian` object just gets created each time. See the `Simplify` class for example.
Apologies if you're doing this in a different way that I've missed.

pybamm/expression_tree/symbol.py
```diff
@@ -412,33 +412,19 @@ def _diff(self, variable):
...
         """
-        if variable.id == self.id:
-            return pybamm.Scalar(1)
-        else:
-            jac = self._jac(variable)
-            # jacobian removes the domain(s)
-            jac.domain = []
-            jac.auxiliary_domains = {}
-            return jac
+        return pybamm.Jacobian(known_jacs).jac(self, variable)
```
tinosulzer: have just seen this `known_jacs` thing. I still think it's more efficient to only create the `Jacobian` class once

pybamm/models/base_model.py
```diff
@@ -52,7 +52,12 @@ class BaseModel(object):
         automatically
     jacobian : :class:`pybamm.Concatenation`
         Contains the Jacobian for the model. If model.use_jacobian is True, the
-        Jacobian is computed automatically during the set up in solve
+        Jacobian is computed automatically during discrisation
```
tinosulzer:
```suggestion
        Jacobian is computed automatically during discretisation
```

pybamm/models/base_model.py
```diff
@@ -52,7 +52,12 @@ class BaseModel(object):
...
+    jacobian_algebraic : :class:`pybamm.Concatenation`
+        Contains the Jacobian for the algebraic part of the model. This may be used
+        by the solver when calculating consistent initial conditions. If
+        model.use_jacobian is True, the Jacobian is computed automatically during
+        discrisation
```
tinosulzer:
```suggestion
        discretisation
```

pybamm/models/base_model.py
```diff
@@ -226,6 +232,14 @@ def jacobian(self):
     def jacobian(self, jacobian):
         self._jacobian = jacobian
 
+    @property
+    def jacobian_algebraic(self):
```
tinosulzer: why is there a jacobian_algebraic but no jacobian_rhs?

pybamm/expression_tree/binary_operators.py
rtimms: good suggestion

pybamm/expression_tree/jacobian.py
rtimms: ah, thanks -- I had misunderstood how `known_jacs` was getting passed around.

pybamm/models/base_model.py
rtimms: only because we only ever use either the full jacobian (for the solver) or the algebraic part (for consistent ics). I'll add `jacobian_rhs` too for completeness.

pybamm/discretisations/discretisation.py
rtimms: also, if we make some kind of master simulation class we could use this method from the disretisation in the solver set up since we would be able to access the disc of the model when in the solver set up

pybamm/discretisations/discretisation.py
tinosulzer: haha I assumed it had got into your autocorrect

pybamm/models/base_model.py
tinosulzer: oh right, I had forgotten about that, probably fine without `jacobian_rhs` then. Need to make sure we aren't calculating the jacobian for the algebraic part twice then (can't remember if already doing this)

pybamm/expression_tree/jacobian.py
```diff
@@ -0,0 +1,92 @@
...
+        if isinstance(symbol, pybamm.Outer):
+            # Need to treat outer differently. If the left child of an Outer
+            # evaluates to number then we need to return a matrix of zeros
+            # of the correct size, which requires variable.evaluation_array
+            left, right = symbol.children
+            # process children
+            left_jac = self.jac(left, variable)
+            right_jac = self.jac(right, variable)
+            # _outer_jac defined in pybamm.Outer
+            jac = symbol._outer_jac(left_jac, right_jac, variable)
```
tinosulzer: not a big deal but you could have the isinstance(Outer) check on line 60 to avoid writing the process_children lines twice

tests/unit/test_expression_tree/test_jac.py
```diff
@@ -80,9 +80,9 @@ def test_linear(self):
 
         # test jac of outer if left evaluates to number
         func = pybamm.Outer(pybamm.Scalar(1), pybamm.Scalar(4))
-        jacobian = np.zeros((1, 4))
+        # jacobian = np.zeros((1, 4))
```
tinosulzer: can be removed

PR #681, opened 2019-10-25
Issue 570 processed var warning

### LINKED ISSUE #570: processed variable error when t,x,or r outside of range
make `ProcessedVariable` throw an error if you try and call outside of the solution time and/or space

tinosulzer:
# Description

Add warning if processed variable is called outside range
Fixes #570 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/logger.py
+++ b/pybamm/logger.py
@@ -3,6 +3,11 @@
 #
 import logging
 
+
+def set_logging_level(level):
+    logger.setLevel(level)
+
+
 format = (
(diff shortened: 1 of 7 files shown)


pybamm/logger.py
```diff
@@ -12,7 +17,5 @@
 
 # Create a custom logger
 logger = logging.getLogger(__name__)
+set_logging_level("WARNING")
```
rtimms: loks like the `test_logger` test needs updating as the default logger level has changed

Issue #686, opened 2019-10-26
Casadi C-code generation


tinosulzer:
Use CasADi to generate C-code; interface more efficiently with the KLU solver.

tinosulzer:
Closing as currently very low priority

Issue #685, opened 2019-10-26
Casadi interpolator


tinosulzer:
Convert a pybamm `Interpolant` to appropriate CasADi object

Issue #684, opened 2019-10-26
Casadi events


tinosulzer:
Hack events with the casadi solver by doing step-and-check (if possible). May slow down the solver too much.

Issue #646, opened 2019-10-04
Bug on Mac install (ft. sinister error message)


tinosulzer:
After running `pip install -e .[dev,docs]` and `python3 run-tests.py --unit`, `test_symobl_visualise` fails (hopefully just needs an extra dependency). 
NB - I don't know who thought it would be a good idea to call a function `_execute_child`
```
======================================================================
ERROR: test_symbol_visualise (test_expression_tree.test_symbol.TestSymbol)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/vsulzer/Documents/PyBaMM/tests/unit/test_expression_tree/test_symbol.py", line 312, in test_symbol_visualise
    rhs.visualise("StefanMaxwell_test.png")
  File "/Users/vsulzer/Documents/PyBaMM/pybamm/expression_tree/symbol.py", line 224, in visualise
    ).to_picture(filename)
  File "/Users/vsulzer/Documents/PyBaMM/PyBaMM-env/lib/python3.7/site-packages/anytree/exporter/dotexporter.py", line 268, in to_picture
    check_call(cmd)
  File "/usr/local/Cellar/python/3.7.4_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/subprocess.py", line 342, in check_call
    retcode = call(*popenargs, **kwargs)
  File "/usr/local/Cellar/python/3.7.4_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/subprocess.py", line 323, in call
    with Popen(*popenargs, **kwargs) as p:
  File "/usr/local/Cellar/python/3.7.4_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/subprocess.py", line 775, in __init__
    restore_signals, start_new_session)
  File "/usr/local/Cellar/python/3.7.4_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/subprocess.py", line 1522, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'dot': 'dot'
```

martinjrobins:
looks like `graphviz`  needs to be added as a dependency for PyBaMM. This provides the `dot` executable. Note this isn't pip installable, so user would have to install manually (https://graphviz.gitlab.io/download/)

ferranbrosa:
I noticed that when I ran the tests in my computer one failed (which was related to this `dot` thing).

PR #687, opened 2019-10-26
Issue 664 casadi

### LINKED ISSUE #664: Casadi solver
Add casadi solver. Requires writing the tree to a casadi format. Makes use of their Jacobian evaluation, and hopefully later also optimization toolbox

tinosulzer:
# Description

Use CasADi to provide big speed-up to both setup and solve:

- Convert PyBaMM expression tree to CasADi
- Wrap CasADi solver

Fixes #664 

Still to do:

- Events (#684 )
- Interpolants (#685 )
- C-code generation (#686 )

## Type of change

# Key checklist:

## Further checks:


pybamm/solvers/casadi_solver.py
```diff
@@ -0,0 +1,103 @@
...
+
+        solve_start_time = timer.time()
+        pybamm.logger.info("Calling DAE solver")
+        solution = self.integrate_casadi(
+            self.casadi_problem, self.y0, t_eval, mass_matrix=model.mass_matrix.entries
+        )
+        solve_time = timer.time() - solve_start_time
+
+        # Identify the event that caused termination
+        termination = self.get_termination_reason(solution, self.events)
```
Scottmar93: You managed to get events working?

pybamm/solvers/dae_solver.py
```diff
@@ -206,14 +207,113 @@ def eval_event(t, y):
...
+
+        def residuals(t, y, ydot):
+            pybamm.logger.debug(
+                "Evaluating residuals for {} at t={}".format(model.name, t)
+            )
+            states_eval = all_states_fn(t, y).full()[:, 0]
+            return states_eval - model.mass_matrix.entries @ ydot
+
+        # Create event-dependent function to evaluate events
+        def event_fun(event):
```
Scottmar93: If events aren't working should we remove this bit of code?

tests/unit/test_solvers/test_idaklu_solver.py
```diff
@@ -53,7 +53,7 @@ def rhs(t, y):
         def alg(t, y):
             return np.array([1 - y[1]])
 
-        solver = pybamm.IDAKLU()
+        solver = pybamm.IDAKLUSolver()
```
Scottmar93: Not a fan of my naming haha

pybamm/models/base_model.py
```diff
@@ -96,7 +103,7 @@ def __init__(self, name="Unnamed model"):
         # Default behaviour is to use the jacobian and simplify
         self.use_jacobian = True
         self.use_simplify = True
-        self.use_to_python = True
+        self.convert_to_format = "python"
```
rtimms: if casadi performs better should we just use that? though maybe nice to keep the pybamm expression tree by default. maybe it would be good to add a notebook or some kind of readme about using pybamm with casadi to highlight why/when you might want to use this option

pybamm/solvers/casadi_solver.py
tinosulzer: Not yet, will remove this

pybamm/solvers/dae_solver.py
tinosulzer: Still need this for setting up with casadi and passing to other solvers

pybamm/models/base_model.py
tinosulzer: Will update the install instructions separately in #646 as I've made a few changes there anyway. Casadi is faster than python but harder to debug (not sure how you step through the expression tree to see which bit of code is messing up) so not sure which one should be default? I lean towards casadi

tinosulzer: Actually, casadi doesn't handle interpolation yet (in pybamm, it has the functionality but needs to be wrapped), and other similar issues might appear, so I will leave python as default until more confident with casadi

rtimms: sounds sensible

PR #693, opened 2019-10-30
Issue 688 add simulation class

### LINKED ISSUE #688: Add simulation class
**Summary**
Create a simulation class which wraps model, discretization, parameters etc. 

**Motivation** 
Can be extended to allow linking of the experiment, model, discretization, and solver together and allow each to reference each other. Also simplifies the pybamm user interface so that users only need to see what the actually are setting

Scottmar93:
# Description

Convenience class for interacting with the front-end of PyBaMM. I have set it up to try to be Jupyter notebook friendly in that things won't break if you try to build twice or solve without first building. It'll also automatically reset everything if you decide to change a spatial_method / parameter etc. Would be good to add some functionality on saving an loading later as well as allowing for experiment specific things to be set such as the current. 

Fixes #688 

## Type of change

# Key checklist:

## Further checks:


pybamm/simulation.py
```diff
@@ -0,0 +1,236 @@
...
+class Simulation:
+    """A Simulation class for easy building and running of PyBaMM simulations.
+
+    Parameters
+    ----------
+    model : :class:`pybamm.BaseModel`
+        The model to be simulated
+    """
+
+    def __init__(self, model):
```
rtimms: would be good to (optionally) pass params, geometry etc here when you create the simulation

pybamm/simulation.py
```diff
@@ -0,0 +1,236 @@
...
+        A method to reset a simulation back to its unprocessed state.
+        """
+        self.model = self._model_class(self._model_options)
+        self.geometry = copy.deepcopy(self._unprocessed_geometry)
+        self._mesh = None
+        self._disc = None
+        self._solution = None
+        self._status = "Unprocessed"
+
+    def parameterize(self):
```
rtimms: minor point, but I dont think I like the name `parameterize` -- I think it suggests it is finding the parameters for the model. maybe `set_parameters`?

examples/scripts/run_simulation.py
```diff
@@ -0,0 +1,6 @@
+import pybamm
+
+model = pybamm.lithium_ion.SPM()
+sim = pybamm.Simulation(model)
+sim.solve()
+sim.plot()
```
tinosulzer: Could you add this code snippet to the readme as a super quick way of running a model? e.g. (DFN is more standard than SPM)

## How do I use PyBaMM?

The easiest way to use PyBaMM is to run a 1A constant-current discharge with a model of your choice with all the default settings:
```python3
import pybamm
model = pybamm.lithium_ion.DFN() # Doyle-Fuller-Newman model
sim.solve()
sim.plot()
```
However, much greater customisation is available. It is possible to change the physics, parameter values, geometry, submesh type,  number of submesh points, methods for spatial discretisation and solver for integration (see DFN [script](examples/scripts/DFN.py) or [notebook](examples/notebooks/models/dfn.ipynb)).

Further details on how to customise simulations and add your own models and settings are provided in the [examples and documentation](#examples-and-documentation).

pybamm/simulation.py
tinosulzer: Agreed

pybamm/simulation.py
tinosulzer: Agreed

pybamm/simulation.py
```diff
@@ -0,0 +1,236 @@
...
+        if self._status != "Unprocessed":
+            return None
+
+        self._parameter_values.process_model(self._model)
+        self._parameter_values.process_geometry(self._geometry)
+        self._status = "Parameterized"
+
+    def build(self):
+        """
+        A method to build the model as a pure linear algebra expression. If the model
```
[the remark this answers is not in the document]
tinosulzer: I think "Linear algebra" refers to matrices and vectors, rather than linearity/non-linearity of the problem. We could just say "matrices and vectors" instead of linear algebra though

pybamm/simulation.py
ferranbrosa: I agree

examples/scripts/run_simulation.py
rtimms: 1A or 1C here?

pybamm/simulation.py
ferranbrosa: Would it be reasonable to allow to optionally pass another parameter set?

Scottmar93: @ferranbrosa do you mean something different from @rtimms suggestion?

ferranbrosa: Sorry, no. I read it too fast so I only got geometry (missed params).

examples/scripts/run_simulation.py
Scottmar93: Changed to 1C

PR #723, opened 2019-11-12
Issue 699 remove autograd

### LINKED ISSUE #699: Problems pickling objects
**Describe the bug**
Some objects cannot be pickled due to the following error:
AttributeError: Can't pickle local object 'primitive.<locals>.f_wrapped'

**To Reproduce**
```
import pybamm
import numpy as np
import pickle

def save_obj(fname, obj):
    with open(fname, 'wb') as f:
        pickle.dump(obj, f)

# load model

# create geometry

# load parameter values and process model and geometry
param.process_model(model)

# set mesh

# discretise model
disc.process_model(model)

# solve model
t_eval = np.linspace(0, 0.2, 100)
solver = pybamm.KLU()

# plot
plot = pybamm.QuickPlot(model, mesh, solution)
plot.dynamic_plot()

save_obj('test_save_param', param)
```

**Expected behaviour**
Pickle saves the object

**Additional context**
This type of error can occur when classes are defined within other classes or functions i.e. not the global scope. There was one example of this in pybamm with the klu solver but I think the problem may also be with autograd as the f_wrapped function is from there

tinosulzer:
# Description

Remove autograd dependency. Also had to reformat the current functions and introduce a Heaviside symbol.

Fixes #699 

## Type of change

# Key checklist:

## Further checks:


pybamm/parameters/parameter_values.py
```diff
@@ -426,38 +431,35 @@ def _process_symbol(self, symbol):
...
-            # store the evaluated symbol in the parameters_eval dict
-            if isinstance(function_name, pybamm.BaseCurrent):
-                for param, sym in function_name.parameters.items():
-                    if isinstance(sym, pybamm.Symbol):
-                        new_sym = self.process_symbol(sym)
-                        function_name.parameters[param] = new_sym
-                        function_name.parameters_eval[param] = new_sym.evaluate()
-
-            # Create Function or Interpolant objec
+            # Create Function or Interpolant or Scalar objec
```
rtimms: typo: object

CHANGELOG.md
```diff
@@ -22,6 +24,7 @@
 
 ## Bug fixes
 
+-   Improved flexibility of parameter values so that parameters can be set as functions or scalars ([#723](https://github.com/pybamm-team/PyBaMM/pull/723))
```
rtimms: does this work for things like diffusivity which we assume is a function of some of the variables, or just functions of time/parameters?

tinosulzer: It should work with both. If something is assumed to be a function in the model then we can pass a scalar and it will always take that scalar value without changing the model. But the other way round (scalar to function) doesn't (can't) work, as you have to tell the model what the function depends on

PR #726, opened 2019-11-13
Issue 579 volume fractions

### LINKED ISSUE #579: Add active and inactive material volume fractions
**Summary**
In the current definition of the parameters the microstructure is defined by the porosity. I assume this means that the rest of the volume fraction is assumed to be active material. It would be interesting to split the volume fractions into electrolyte (i.e. porosity), active material and inactive material.

 **Motivation**
In some cases the conductive filler in the electrode can make up to 10% of the volume so that has an impact on the amount of active material to consider in the model.

### LINKED ISSUE #648: add submodels for effective parameters
**Summary**
Currently, the Bruggeman correlation is hard-coded into the model. We should make it more general by introducing a parameter that accounts for the effective properties of the material similar to the curly B in Richardson et al. (2012). As discussed in the meeting (07/10/19) we could introduce a submodel for it (called porosity, for example) which calculates this parameter from the other ones.

 **Motivation**
Bruggeman correlation is widely used in battery modelling, but it is only one empirical approach to determining the effective properties of a porous materials. There are other approaches such as asymptotic homogenisation or softwares like TauFactor so a more flexible definition of the porosity would allow to use these other techniques.

tinosulzer:
# Description

Add tortuosity submodels, and add active/inactive material volume fractions

Fixes #579 
Fixes #648 

## Type of change

# Key checklist:

## Further checks:


pybamm/models/submodels/tortuosity/bruggeman_tortuosity.py
```diff
@@ -0,0 +1,35 @@
...
+        param = self.param
+
+        if self.phase == "Electrolyte":
+            eps = variables["Porosity"]
+        elif self.phase == "Electrode":
+            eps = variables["Active material volume fraction"]
+
+        eps_n, eps_s, eps_p = eps.orphans
+        tor = pybamm.Concatenation(
+            eps_n ** param.b_n, eps_s ** param.b_s, eps_p ** param.b_p
```
ferranbrosa: If I understand it correctly, here all the parameters `param.b_j` are the same for electrode and electrolyte in a given part. I think that is not necessarily true because the geometry that defines the tortuosity is quite different between the electrode and electrolyte. Would it be possible to pass each `b_j` as an argument when you call a Bruggeman tortuosity model?

pybamm/parameters/standard_parameters_lithium_ion.py
```diff
@@ -277,6 +277,11 @@ def U_p_dimensional(sto, T):
     pybamm.FullBroadcast(epsilon_s, ["separator"], "current collector"),
     pybamm.FullBroadcast(epsilon_p, ["positive electrode"], "current collector"),
 )
+epsilon_s_n = pybamm.Parameter("Negative electrode active material volume fraction")
+epsilon_s_p = pybamm.Parameter("Positive electrode active material volume fraction")
+epsilon_binder_n = 1 - epsilon_n - epsilon_s_n
```
ferranbrosa: Minor point: it might be more accurate to call it `epsilon_inactive_n`.

TomTranter:
Tortuosity is a mess in terms of a consistent definition in the literature. This paper explains a common misinterpretation and distinguishes between tortuosity being the ratio of the diffusion path length and domain length and tortuosity factor being this ratio squared. http://www.sciencedirect.com/science/article/pii/0009250989850535
There are a bunch of alternatives to Bruggeman which was derived for sphere packs explored here 
https://www.sciencedirect.com/science/article/pii/S0009250907003144
And also I have published a few papers on tortuosity. One was extracted from experimental data using a transient diffusion method and you have to be careful as Fick's second law has a porosity terms on both sides of the equation so effective diffusivity is dependent on tortuosity alone. http://linkinghub.elsevier.com/retrieve/pii/S0017931016328174 and also I wrote a python package to calculate the tortuosity from a tomography image called pytrax that uses random walks. The definition of tortuosity used here is the ratio of the mean square displacement with time. It is 1 in open space and proportional to 1/tortuosity in porous media https://doi.org/10.1016/j.softx.2019.100277 The theory for that is well explained in this ref https://www.jstage.jst.go.jp/article/jagh1987/43/1/43_13/_article

pybamm/models/submodels/tortuosity/bruggeman_tortuosity.py
tinosulzer: Would it just be a case of having a different parameter for electrolyte and solid?

ferranbrosa: Yes. Also, they could be different for different transport phenomena in the same domain. Right now it is not a problem as we only need one number for electrode and one for electrolyte, but later on we might need more.

tinosulzer:
Thanks Tom for the info. It sounds like there are lots of different options and it would be interesting to implement a few and compare their impact on the DFN (at the low C-rates we have been considering so far electrolyte diffusion is a higher order effect, but it could play a big role if we start to consider higher C rates). 
For this particular issue I'm just going to focus on getting the framework in place to allow adding more submodels. What would you suggest calling these submodels (and corresponding variables) instead of "tortuosity"? Maybe "effective transport factor"?

TomTranter:
I think tortuosity is ok

PR #728, opened 2019-11-13
Issue 692 external temperature matrix

### LINKED ISSUE #692: Input external temperature matrix
**Summary**
Allow for an external temperature matrix T(x,y,z) (and a potetial boundary conditions phi(y,z)) and output an internal heat generation matrix.

Scottmar93:
# Description

Added a more general method for interfacing with external variables. Can choose any submodel to be external and then one needs to just pass the variable that would have been solved for by the submodel if it wasn't made external to the solver at each time step. This only works for stepping. 

Works for solvers except Casadi (tino knows how to get this working though).

I have also added a couple of helper methods to Simulation that make it easy to pass and extract arrays corresponding to variables to/from pybamm. 

Fixes #692 

## Type of change

# Key checklist:

## Further checks:


pybamm/discretisations/discretisation.py
```diff
@@ -147,7 +156,26 @@ def process_model(self, model, inplace=True):
 
         model_disc.bcs = self.bcs
 
-        # Process initial condtions
+        self.external_variables = model.external_variables
+        # find where external variables begin in state vector
```
rtimms: maybe add a comment here saying something about how the external vars are always appended to the end of the state vector, so you only care where they start?

PR #748, opened 2019-11-25
Issue 741 neumann bc

### LINKED ISSUE #741: Neumann BCs in FV
At the moment we are adding ghost nodes for Neumann conditions in Finite Volumes rather than using the actual value

rtimms:
# Description
Changes finite volumes to use the exact value of the gradient at the cell edges if given by a Neumann value, instead of adding a ghost node.

Fixes #741 

## Type of change

# Key checklist:

## Further checks:


pybamm/spatial_methods/finite_volume.py
```diff
@@ -543,6 +545,14 @@ def add_ghost_nodes(self, symbol, discretised_symbol, bcs):
...
         rbc_value, rbc_type = bcs["right"]
 
+        # Add ghost node(s) to domain
+        if lbc_type == "Dirichlet":
+            domain = [domain[0] + "_left ghost cell"] + domain
+        if rbc_type == "Dirichlet":
+            domain = domain + [domain[-1] + "_right ghost cell"]
+
+        # Calculate values for ghost nodes for any Dirichlet boundary conditions
+        # and adjust the domain name to account for the new ghost nodes
```
tinosulzer: where is domain name adjusted?

rtimms: ah thanks domain name is adjusted in the lines above, but i forgot to move the comment. maybe i misunderstood what was happening before. it looked like we added a ghost node assuming the function was linear there with slope given by the gradient? that was probably consistent with the scheme anyway, and was giving the correct result. it doesn't look like doing things the way i've implemented actually makes any different other than we no longer unnecessarily add extra ghost nodes

tinosulzer: Before it was adding a ghost node in such a way that (ghost_node - last_internal_node) / dx = flux, but good to avoid adding unnecessary ghost nodes

PR #753, opened 2019-12-02
Issue 619 make pip installable

### LINKED ISSUE #619: Make pybamm pip installable

tlestang:
# Description
This simplifies the installation of PyBaMM on Linux systems. 
The core of PyBaMM can now be installed via the single command
```bash
pip install -i https://test.pypi.org/simple/ -extra-index-url https://pypi.org/simple pybamm
```
`-i https://test.pypi.org/simple/` points to the the TestPyPI index and ` -extra-index-url https://pypi.org/simple` allows pip to fetch dependencies from the regular PyPI index.
This will change to `pip install pybamm` once this PR is merged.

Additionally, the installation of `scikits.odes` and the KLU module
are automated, along with the download and/or installation of their own requirements (SUNDIALS and SuiteSparse). For instance, the KLU module can be built invoking the following command
```bash
python setup.py install_klu
```
If neither of the SUNDIALS nor SuiteSparse are present on the system, the above command
will download and build the libraries and build the KLU python module using the `CMakeLists.txt` included in the PyBaMM root directory.

A good starting point for reviewing this PR is the modified installation instructions for linux.
Apart from that, the only modified file is `setup.py`.

## Expected use cases
- User only requires the core of PyBaMM
```bash
pip install pybamm
```
- User requires PyBaMM and `scikits.odes`
```bash
git clone https://github.com/pybamm-team/PyBaMM.git 
cd PyBaMM/
python setup.py install_odes
```
This will download the SUNDIALS library, compile it and install `scikits.odes` via pip.

- User requires PyBaMM and the KLU solver
```bash
...
python setup.py install_klu
```
- User requires the KLU solver. User used `install_odes` before and the source code for the SUNDIALS is located in `sundials-4.1.0`.
```bash
python setup.py install_klu --sundials-src=sundials-4.1.0
```
This will download an build SuiteSparse if needed. 
If the SUNDIALS KLU module is not detected (not installed when installing `scikits.odes`), then SUNDIALS will be built again with KLU enabled.

- User requires `scikits.odes`. 
User used `install_klu` before and the source code for the SUNDIALS is located in `sundials-4.1.0`.
```bash
python setup.py install_odes --sundials-src=sundials-4.1.0
```
This will skip download and build the SUNDIALS if needed.

## A few words about the implementation
The download/installation of `scikits.odes`, KLU module as well a SUNDIALS and SuiteSparse is implemented in the `setup.py` script.
The `setup.py` script is based on the `distutils` (or its enhancement `setuptools`) and gives access to commands controlling the packaging and installation of python packages.
The most used command is `install`, that builds the package and then install it.
By default, a lot of commands are available. The full list can be accessed via
```bash
python setup.py --help-commands
```
Another example is the `bdist_wheel` command, that builds a binary wheel distribution that can be then uploaded to PyPI for distribution of the package.
This is what I used to distribute PyBaMM, [among a few other things](https://packaging.python.org/tutorials/packaging-projects/) on TestPyPI.

Custom commands can be defined by subclassing the existing commands.
I implemented both the `install_klu` and `install_odes` commands by subclassing `distutils.cmd.Command` which is the top-level abstract class
for a command. 
All the other commands are derived from this abstract class.
When writing a `distutils.cmd.Command` class, it is required to define the three following methods
- `initialize_options()` (Initial values for options)
- `finalize_options` (process values of options)
- `run()` (actual functionality of the command)

To review the changes, I'd suggest to start with the class `InstallODES`, which corresponds to the command `install_odes`. 
See the `setup()` function for the correspondence between command classes and commands.

## Limitations and future plans
- Only the core of PyBaMM can be installed through pip.
  Users wishing to install `scikits.odes` or the KLU module are still required to download the PyBaMM source code.
  The reason for that is that they need access to the `setup.py` file, where the installation of these additional dependencies is implemented.
  A solution would be to distribute pre-compiled binaries and header files for SuiteSparse and SUNDIALS along with PyBaMM.
- The klu module is created in the PyBaMM root dir, and is therefore only accessible from this directory. A solution is to build the klu module as a C extension by subclassing the `build_ext` distutils command.
his PR only addresses the installation of PyBaMM on linux.
- This PR only addresses the installation of PyBaMM on Linux. A solution would be to use `scikit-build` and/or `conda` provide a cross-platform package. I'm looking into it.

Fixes #619 
 
## Type of change

# Key checklist:

## Further checks:


INSTALL-LINUX.md
```diff
@@ -1,66 +1,74 @@
...
 source env/bin/activate
 ```
-
 Now all the calls to pip described below will install PyBaMM and its dependencies into
 the environment `env`. When you are ready to exit the environment and go back to your
 original system, just type:
 
 ```bash
 deactivate
 ```
```
tinosulzer: I would put this before the "user install" section as otherwise someone might do `pip install pybamm` before setting up a venv (and it also applies to developer install)

setup.py
```diff
@@ -1,15 +1,420 @@
...
     from distutils.core import setup, find_packages
+from distutils.cmd import Command
+
+
+def download_extract_library(url):
+    # Download and extract archive at url
+    if NO_WGET:
+        # The NO_WGET is set to true if the wget could not be
+        # imported.
+        error_msg = "Could not find wget module. Please install wget module."
```
tinosulzer: can you provide instructions on how to install in this error message? edit: I got this error even after installing wget

tlestang: Oh yes, the error message should be
`Could not find wget module. Please install wget module (pip install wget).`
This uses the `wget` python package (https://pypi.org/project/wget/). I did not add it to the list of pybamm dependencies because... well it's not really a dependency.
I hope that in a near future, I'll find the time to replace `wget` by the `urllib` module which is in the standard library.
Also I should update the installation notes, because I realize now `wget` is not mentioned anywhere

tlestang:
Thanks for the review @tinosulzer . 
About the scikit-fem error, that's because pip is trying to fetch the dependencies from the test index as well.
Two solutions:
- Do not install dependencies (`pip install --no-deps`)
- Specifiy an extra index where to look for dependencies
```
pip install -i https://test.pypi.org/simple/ -extra-index-url https://pypi.org/simple pybamm
```
I updated the description of the PR

tinosulzer:
Ok I can now install without the sundials dependencies. Had to fix a couple of typos for the others to work, I can push these straight to this branch if you like?

After installing with `install_odes`, trying to install with `install_klu` gave
```
/usr/lib/python3.6/distutils/dist.py:261: UserWarning: Unknown distribution option: 'long_description_content_type'
  warnings.warn(msg)
running install_klu
GNU Make 4.1
Built for x86_64-pc-linux-gnu
Copyright (C) 1988-2014 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
About to download SuiteSparse, proceed? (y/n): y
-1 / unknown
```
but then worked fine after ~20 seconds (the `-1 / unknown` might confuse people). Btw, the lazy option is not to use `--sundials-src=` so I guess most people won't do this. Also, the error message "Make sure the pybind11 repository was cloned in ./third-party/" should be very strongly emphasized (bold/different color?)

Install could be sped up by using multiple cores for `make` if available.

Also it would be good to suggest the user checks their install has worked by running either
`python3 -c "import pybamm; print(pybamm.have_scikits_odes())"` or `python3 -c "import pybamm; print(pybamm.have_idaklu())"` (note that these are the wrong way round in your branch - return False where should be True and vice versa - but fixed in master)

tlestang:
- Yes, please push any typos-fixing commit you'd like :) Hope it wasnt too bad
- The `-1 / unknown` message comes from the `wget` module, I'm not sure there's anything we can do about it because it's probably related to the amount of information the module gets from the the server? Because it works fine when download the SUNDIALS... Anyway, this will probably go away when switching to `urllib`.

> Also it would be good to suggest the user checks their install
Good point, will do

PR #767, opened 2019-12-20
Issue 765 compute grad fem

### LINKED ISSUE #765: current collector ohmic heating in 2+1D
at the moment the ohmic heating in the current collector `sigma*abs(grad(phi))^2` is being computed incorrectly (it sometimes goes negative). Note: it is correct when using finite volumes, so is only a problem when using 2D current collectors, which are discretised with finite elements. a short term fix is to replace it with the approximation that the heating in the current collector is uniform. for most use cases the current collector conductivity is extremely large, so the contribution of the ohmic heating in the current collector is small.

rtimms:
# Description
Added the gradient operation in the finite element method. This has then been used to compute the ohmic heating in the current collector 

Fixes #765 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/tests/unit/test_spatial_methods/test_scikit_finite_element.py
+++ b/tests/unit/test_spatial_methods/test_scikit_finite_element.py
@@ -141,8 +141,20 @@ class TestScikitFiniteElement(unittest.TestCase):
         grad_disc = disc.process_symbol(gradient)
         grad_disc_y, grad_disc_z = grad_disc.children
 
-        np.testing.assert_array_almost_equal(grad_disc_y.evaluate(None, 5*y + 6*z), 5*np.ones_like(y)[:, np.newaxis])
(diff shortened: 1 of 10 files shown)


tests/unit/test_spatial_methods/test_scikit_finite_element.py
```diff
@@ -125,6 +122,40 @@ def test_discretise_equations(self):
...
+        np.testing.assert_array_almost_equal(
+            grad_disc_z.evaluate(None, 5 * y + 6 * z),
+            6 * np.ones_like(z)[:, np.newaxis],
+        )
+
+        # check grad_squared positive
+        eqn = pybamm.grad_squared(var)
+        eqn_disc = disc.process_symbol(eqn)
+        ans = eqn_disc.evaluate(None, 3 * y ** 2)
+        self.assertTrue(all(ans > 0))
```
tinosulzer: np.testing.assert_array_less(0, ans) would give more informative failure messages

PR #783, opened 2020-01-13
#782 add harmonic mean

### LINKED ISSUE #782: Harmonic mean for node to edge in fluxes
Use the harmonic mean when doing `node_to_edge` in binary operators so that fluxes have the correct behavior

rtimms:
# Description
Adds the harmonic mean to the finite element method. This is then used in `process_binary_operators` so that fluxes are computed using the harmonic mean of the diffusivity (or conductivity etc.) 

Fixes #782 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/spatial_methods/finite_volume.py
+++ b/pybamm/spatial_methods/finite_volume.py
@@ -967,7 +967,11 @@ class FiniteVolume(pybamm.SpatialMethod):
     def process_binary_operators(self, bin_op, left, right, disc_left, disc_right):
         """Discretise binary operators in model equations.  Performs appropriate
         averaging of diffusivities if one of the children is a gradient operator, so
-        that discretised sizes match up.
+        that discretised sizes match up. For this avergaing we use the harmonic
+        mean [1].
+
+        [1] Recktenwald, Gerald. "The control-volume finite-difference approximation to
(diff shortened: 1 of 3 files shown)


pybamm/spatial_methods/finite_volume.py
```diff
@@ -1121,10 +1140,138 @@ def arithmetic_mean(array):
...
+            .. math::
+                D_{eff} = \\frac{D_1  D_2}{\\beta D_2 + (1 - \\beta) D_1},
+
+            where
+
+            .. math::
+                \\beta = \\frac{\\Delta x_1}{\\Delta x_2 + \\Delta x_1}
+
+            accounts for the difference in the control volume widths. This is the
+            definiton from [1].
```
tinosulzer: is this the same definition as the Recktenwald paper?

rtimms: yeah, just in slightly different notation. Could remove this ref if you think it's cleaner to just provide one

tinosulzer: just saying it's the same but with different notation would be fine (you can add that commit with [ci skip] if you don't want to wait for the tests)

PR #795, opened 2020-01-23
Issue 791 complete models

### LINKED ISSUE #791: Create "complete" SPM and DFN classes
Create classes where the whole model can be seen in a single place

tinosulzer:
# Description

Create a basic SPM and basic DFN in order to help introduce the battery models in a more user-friendly way.

Fixes #791 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/examples/notebooks/README.md
+++ b/examples/notebooks/README.md
@@ -50,7 +50,7 @@ The following notebooks are specific to different stages of the PyBaMM pipeline,
 
 Several battery models are implemented and can easily be used or [compared](./models/lead-acid.ipynb). The notebooks below show the solution of each individual model. We always welcome [new models](https://pybamm.readthedocs.io/en/latest/tutorials/add-model.html)!
 
-Once you are comfortable with the expression tree structure, a good starting point to understand the models in PyBaMM is to take a look at the [basic SPM](https://github.com/pybamm-team/PyBaMM/blob/master/pybamm/models/full_battery_models/lithium_ion/basic_spm.py) and [basic DFN](https://github.com/pybamm-team/PyBaMM/blob/master/pybamm/models/full_battery_models/lithium_ion/basic_dfn.py), since these show the entire model in one place and so are easier to understand. However, we recommend that you subsequently use the full models as they offer much greater flexibility for coupling different physical effects and visualising a greater range of variables.
+Once you are comfortable with the expression tree structure, a good starting point to understand the models in PyBaMM is to take a look at the [basic SPM](https://github.com/pybamm-team/PyBaMM/blob/master/pybamm/models/full_battery_models/lithium_ion/basic_spm.py) and [basic DFN](https://github.com/pybamm-team/PyBaMM/blob/master/pybamm/models/full_battery_models/lithium_ion/basic_dfn.py), since these define the entire model (variables, equations, initial and boundary conditions, events) in a single class and so are easier to understand. However, we recommend that you subsequently use the full models as they offer much greater flexibility for coupling different physical effects and visualising a greater range of variables.
 
 #### Lithium-ion models
 
--- a/examples/notebooks/README.md
+++ b/examples/notebooks/README.md
@@ -48,7 +48,9 @@ The following notebooks are specific to different stages of the PyBaMM pipeline,
 
 ### Models
 
-The following models are implemented and can easily be used or [compared](./models/lead-acid.ipynb). We always welcome [new models](https://pybamm.readthedocs.io/en/latest/tutorials/add-model.html)!
+Several battery models are implemented and can easily be used or [compared](./models/lead-acid.ipynb). The notebooks below show the solution of each individual model. We always welcome [new models](https://pybamm.readthedocs.io/en/latest/tutorials/add-model.html)!
+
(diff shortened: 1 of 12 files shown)


pybamm/models/full_battery_models/lithium_ion/basic_dfn.py
```diff
@@ -0,0 +1,270 @@
...
+        The name of the model.
+
+    References
+    ----------
+    .. [2] SG Marquis, V Sulzer, R Timms, CP Please and SJ Chapman. “An asymptotic
+           derivation of a single particle model with electrolyte”. In: arXiv preprint
+           arXiv:1905.12553 (2019).
+
+
+    **Extends:** :class:`pybamm.lithium_ion.BaseModel`
```
tlestang: I got a bit confused until I realized that there are 2 different `BaseModel` classes, and that the second one overwrites the first!

pybamm/models/full_battery_models/lithium_ion/basic_spm.py
```diff
@@ -0,0 +1,171 @@
...
+    ----------
+    .. [2] SG Marquis, V Sulzer, R Timms, CP Please and SJ Chapman. “An asymptotic
+           derivation of a single particle model with electrolyte”. In: arXiv preprint
+           arXiv:1905.12553 (2019).
+
+
+    **Extends:** :class:`pybamm.lithium_ion.BaseModel`
+    """
+
+    def __init__(self, name="Doyle-Fuller-Newman model"):
```
tlestang: `name` should be changed here

tests/integration/test_models/test_full_battery_models/test_lithium_ion/test_compare_basic_models.py
```diff
@@ -0,0 +1,68 @@
...
+
+        # Solve main SPM
+        sim = pybamm.Simulation(spm)
+        t_eval = np.linspace(0, 1)
+        sim.solve(t_eval)
+        sol = sim.solution
+
+        # Compare solution data
+        np.testing.assert_array_almost_equal(basic_sol.y, sol.y)
+        np.testing.assert_array_almost_equal(basic_sol.t, sol.t)
```
tlestang: probably silly but why not testing with 4 decimal points like in the case of dfns in lines 28 and 29?

examples/notebooks/README.md
```diff
@@ -48,7 +48,9 @@ The following notebooks are specific to different stages of the PyBaMM pipeline,
 
 ### Models
 
-The following models are implemented and can easily be used or [compared](./models/lead-acid.ipynb). We always welcome [new models](https://pybamm.readthedocs.io/en/latest/tutorials/add-model.html)!
+Several battery models are implemented and can easily be used or [compared](./models/lead-acid.ipynb). The notebooks below show the solution of each individual model. We always welcome [new models](https://pybamm.readthedocs.io/en/latest/tutorials/add-model.html)!
+
+Once you are comfortable with the expression tree structure, a good starting point to understand the models in PyBaMM is to take a look at the [basic SPM](https://github.com/pybamm-team/PyBaMM/blob/master/pybamm/models/full_battery_models/lithium_ion/basic_spm.py) and [basic DFN](https://github.com/pybamm-team/PyBaMM/blob/master/pybamm/models/full_battery_models/lithium_ion/basic_dfn.py), since these show the entire model in one place and so are easier to understand. However, we recommend that you subsequently use the full models as they offer much greater flexibility for coupling different physical effects and visualising a greater range of variables.
```
tlestang: At first I didn't exactly understand what _showing the entire model in one place_ means.
After looking at the implementation of models, I think that means that the model parameters, variables, equations.. are set within the class definition explicitly, instead of happening hidden in the various sub models. Is that correct?

pybamm/models/submodels/thermal/isothermal/isothermal.py
```diff
@@ -23,7 +23,7 @@ def __init__(self, param):
 
     def get_fundamental_variables(self):
 
-        T_x_av = pybamm.PrimaryBroadcast(0, "current collector")
+        T_x_av = pybamm.PrimaryBroadcast(self.param.T_init, "current collector")
```
tlestang: Is this likely to affect other models other than the new basic ones?

tinosulzer:
Thanks for the review Thibault, yes the 2 basic models are purely pedagogical: people can see all the bits that go into those models in one place, and make their own copy to easily edit if they want to make changes to the model.

The basic SPM should be (almost) exactly the same as the "full" SPM with default options, with the only difference being that the full SPM variables have the option of being discretised to become 3D variables, but the basic SPM ones don't. So after discretisation they should be exactly the same.

The flexibility here means being able to combine different submodels (e.g. different thermal models, mechanics, degradation) in any arbitrary combination, which isn't possible with the basic SPM. We could do this with lots of if statements too but decided it would get too messy

pybamm/models/submodels/thermal/isothermal/isothermal.py
tinosulzer: yeah, this was a bug fix

examples/notebooks/README.md
tinosulzer: that's correct, I will update to say that

tests/integration/test_models/test_full_battery_models/test_lithium_ion/test_compare_basic_models.py
tinosulzer: this tests with the default of 6 decimal points (reduced to 4 for the DFN because the model is more complex so solution less accurate)

PR #796, opened 2020-01-24
Issue 790 param not recognized

### LINKED ISSUE #662: Use fuzzy string matching
Fuzzy string matching libraries [already exist](https://chairnerd.seatgeek.com/fuzzywuzzy-fuzzy-string-matching-in-python/).
We could help the problem of "lots of variables in the variables dict" by implementing `model.variables` as a "fuzzy dictionary" which, when a key is required, does something like:

- try finding the key
- if `KeyError`, look in dict for similar keys (e.g. at most 3 most similar keys above a certain threshold of similarity)
- allow the user to change the key they're asking for
- continue

### LINKED ISSUE #790: Raise error if parameter not recognised
After initial creation, raise an error if user tries to update a parameter value that hasn't already been set by the defaults

tinosulzer:
# Description

Add fuzzy dict and raise errors if a parameter is updated that wasn't previously defined (so users don't think they've updated them when they haven't)

Fixes #790 
Fixes #662 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/parameters/parameter_values.py
+++ b/pybamm/parameters/parameter_values.py
@@ -159,6 +159,26 @@ class ParameterValues:
         return {k: v for (k, v) in zip(df["Name [units]"], df["Value"])}
 
     def update(self, values, check_conflict=False, check_already_exists=True, path=""):
+        """
+        Update parameter dictionary, while also performing some basic checks.
+
+        Parameters
+        ----------
+        values : dict
+            Dictionary of parameter values to update parameter dictionary with
+        check_conflict : bool, optional
+            Whether to check that a parameter in `values` has not already been defined
+            in the parameter class when updating it, and if so that its value does not
+            change. This is set to True during initialisation, when parameters are
+            combined from different sources, and is False by default otherwise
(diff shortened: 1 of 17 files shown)


TomTranter:
Hey @tinosulzer what is the process for adding new parameters now?
Also... fuzzywuzzy is kinda racist slur in Britain... would be better to not use this package imho

tinosulzer:
Same process as before, but if you try to update a parameter that didn't exist before there will be an error that can be overwritten by setting an extra flag in `update`.
That is unfortunate for the name, will look for different package.

pybamm/parameters/parameter_values.py
```diff
@@ -111,7 +124,12 @@ def update_from_chemistry(self, chemistry):
                 os.path.join(component_path, "parameters.csv")
             )
             # Update parameters, making sure to check any conflicts
-            self.update(component_params, check_conflict=True, path=component_path)
```
TomTranter: Would be good to have some docstrings for these parameters

PR #800, opened 2020-01-28
Issue 784 simplify solver

### LINKED ISSUE #463: example notebooks and scripts assume scikits.odes is installed
need to figure out a good way to skip those that use it, or to put a check in the notebooks themselves

### LINKED ISSUE #784: Simplifying solver interface
It should be possible to merge 

- `base_solver`, `ode_solver` and `dae_solver` into a single base class
- `scikits_ode_solver` and `scikits_dae_solver` into a single class

Do you think this would simplify the interface @rtimms ?

### LINKED ISSUE #789: Events for Casadi solver when stepping
**Summary**
Hi, can anyone explain why events are not implemented in the casadi solver compute_solution method and is this something that will never work or something we haven't got round to doing? There is a line saying events not implemented in this method and termination is always set to "final time". However, in the solve method when mode is safe there does seem to be some event handling. I was wondering whether this could be refactored a little so that the simulation step method could also use it as it currently just calls compute_solution.

 **Motivation**
When stepping solution it's good to know why it finished

tinosulzer:
# Description

Simplify the various solver classes by merging the `set_up` methods and getting rid of `compute_solution`. Also combined the implementations for external variables and inputs

Fixes #784 
Fixes #789 
Fixes #463 

## Type of change

# Key checklist:

## Further checks:


pybamm/solvers/base_solver.py
```diff
@@ -49,6 +75,294 @@ def atol(self):
...
+            raise pybamm.SolverError(
+                """
+                Could not find consistent initial conditions: solver terminated
+                successfully, but maximum solution error ({}) above tolerance ({})
+                """.format(
+                    np.max(sol.fun), self.root_tol * len(sol.x)
+                )
+            )
+
     def solve(self, model, t_eval, external_variables=None, inputs=None):
```
Scottmar93: Does external variables actually work when just solving? I thought it was only implemented for "stepping"

tinosulzer: Yes, it should work identically with both now - though I guess the main use cases will be for stepping

PR #801, opened 2020-02-02
Issue 709 simplify on creation

### LINKED ISSUE #709: Simplify some objects on creation
One way to reduce the size of the tree and hence make simplify faster could be to have some simple rules for simplifying objects on creation. For example, `Scalar(5) * Scalar(3)` returns `Scalar(15)`; similarly for `Scalar * Array` and `Array @ Array`.

Since we are now planning on having specific `InputParameter` objects, this would not cause any problems for parameter estimation. We just lose functionality of being able to update parameter values in an already-created model - so you have to specify up front which parameters you will want to update. This is probably something to do after implementing some kind of input parameter node though, just to be safe.

Thoughts?

tinosulzer:
# Description

Simplify objects as soon as they are created, if they are constant.
Also, removed `update_model` functionality, as this can be more efficiently achieved by using `InputParameter`.
Fixes #709 

## Type of change

# Key checklist:

## Further checks:


pybamm/expression_tree/symbol.py
```diff
@@ -360,79 +360,109 @@ def __repr__(self):
...
         """return a :class:`Heaviside` object"""
-        return pybamm.Heaviside(self, other, equal=True)
+        return pybamm.simplify_if_constant(
+            pybamm.Heaviside(self, other, equal=True), keep_domains=True
+        )
 
     def __gt__(self, other):
         """return a :class:`Heaviside` object"""
-        return pybamm.Heaviside(other, self, equal=False)
+        return pybamm.simplify_if_constant(
```
Scottmar93: Instead of having `pybamm.simplify_if_constant( )` in every function can we have `self = pybamm.simplify_if_constant(self)` in the `__init__` of `Symbol`? Maybe this screws with the tree though?

tests/unit/test_expression_tree/test_matrix.py
```diff
@@ -24,18 +24,13 @@ def test_matrix_evaluate(self):
...
         np.testing.assert_array_equal((self.mat + self.mat).evaluate(), 2 * self.A)
-        np.testing.assert_array_equal((self.mat - self.mat).evaluate(), 0 * self.A)
+        np.testing.assert_array_equal(
+            (self.mat - self.mat).evaluate().toarray(), 0 * self.A
+        )
         np.testing.assert_array_equal(
             (self.mat @ self.vect).evaluate(), np.array([[5], [2], [3]])
         )
 
-    def test_matrix_modification(self):
```
Scottmar93: Should this be removed? I don't see how it is related to simplifying on creation.

tests/unit/test_expression_tree/test_symbol.py
```diff
@@ -339,10 +339,10 @@ def test_has_spatial_derivatives(self):
 
     def test_orphans(self):
         a = pybamm.Scalar(1)
-        b = pybamm.Scalar(2)
-        sum = a + b
+        b = pybamm.Parameter("b")
+        summ = a + b
```
Scottmar93: ?

pybamm/expression_tree/symbol.py
tinosulzer: I tried but python doesn't let you assign self in the `__init__`, only its arguments (or at least I couldn't figure out how to), so this seems a bit safer

tests/unit/test_expression_tree/test_matrix.py
tinosulzer: simplifying on creation broke the test, then I thought we were never using this functionality so could get rid of the test. We might actually want to prevent matrix modification outright, to avoid weird bugs?

tests/unit/test_expression_tree/test_symbol.py
tinosulzer: again simplify on creation broke the test

PR #807, opened 2020-02-06
Issue 658 experiment

### LINKED ISSUE #658: Add an Experiment class
**Summary**
Add an `Experiment` class where the user can define all thing related to the specific set up of an experiment, including:
- initial conditions
- cycling profile
- SOC to concentrations calculator
- cooling set up? 

Feel free to suggest other things that should be in this

 **Motivation**
Performing different experiments on models will be one of the main use cases of PyBaMM so we should have a clear and simple method for setting this up.

tinosulzer:
# Description

Added experiment class and integration with the simulation class
Fixes #658 

## Type of change

# Key checklist:

## Further checks:


pybamm/experiments/experiment.py
```diff
@@ -0,0 +1,220 @@
...
+        return converted_operating_conditions, events
+
+    def read_string(self, cond):
+        """
+        Convert a string to a tuple of the right format
+
+        Parameters
+        ----------
+        cond : str
+            String of appropriate form for example "Charge at x C for y hours". x and y
```
rtimms: It might be useful to have the structure of the strings explained in the docs for experiments (or somewhere else more prominent)

pybamm/simulation.py
```diff
@@ -20,13 +20,35 @@ def isnotebook():
...
+
+
 class Simulation:
     """A Simulation class for easy building and running of PyBaMM simulations.
 
     Parameters
     ----------
     model : :class:`pybamm.BaseModel`
         The model to be simulated
+    experiment : : class:`pybamm.Experiment`
```
rtimms: this is optional

pybamm/simulation.py
```diff
@@ -83,6 +110,126 @@ def __init__(
 
             warnings.filterwarnings("ignore")
 
+    def set_up_experiment(self, model, experiment):
```
rtimms: I found this pretty confusing to follow. It's really neat, but took a few goes through!

pybamm/simulation.py
```diff
@@ -181,28 +328,68 @@ def solve(
...
 
-        self.t_eval = t_eval
-        self._solution = solver.solve(
-            self.built_model,
-            t_eval,
-            external_variables=external_variables,
-            inputs=inputs,
-        )
+        if self.operating_mode == "without experiment":
+            # Solve the old way, with a single solve
```
rtimms: maybe solve the "normal" or "usual" way rather than old?

pybamm/solvers/solution.py
```diff
@@ -230,3 +202,93 @@ def save_data(self, filename):
...
+    """
+
+    def __init__(
+        self, t, y, t_event=None, y_event=None, termination="final time",
+    ):
+        super().__init__(t, y, t_event, y_event, termination)
+
+    @property
+    def sub_solutions(self):
+        "Model used for solution"
```
rtimms: not sure this docstring is very clear?

tests/unit/test_experiments/test_experiment.py
```diff
@@ -0,0 +1,139 @@
...
+if __name__ == "__main__":
+    print("Add -v for more debug output")
+    import sys
+
+    if "-v" in sys.argv:
+        debug = True
+    pybamm.settings.debug_mode = True
+    unittest.main()
+
+# {
```
rtimms: this can be removed?

pybamm/simulation.py
rtimms: it's probably fine, but it could be more explicit that we handle experiments by creating a submodel for each instruction and then appending all the solutions together. I think that is obvious when you look at the big picture (e.g. seeing the solution object), but might not be super clear if you just looked here

pybamm/experiments/experiment.py
tinosulzer: have expanded the docstrings. it prints out a load of examples if it doesn't recognize the string, so hopefully will be intuitive enough

pybamm/solvers/solution.py
tinosulzer: oops yeah that is totally a copy paste error

PR #808, opened 2020-02-06
Issue 759 non smooth dae

### LINKED ISSUE #759: DAE solver for non-smooth currents
Fix (at least one of the) DAE solvers to work when the input current is not smooth.

Hacky approach is to do step-by-step and match up discontinuities with steps, or try stepping the differential and algebraic equations separately

martinjrobins:
# Description

Fixes #759 

- adds `pybamm.Event` class for holding information about events
- added a new class of event that returns the time that a discontinuity in the solution occurs
- changed the BaseSolver to filter out the discontinuity events. Integration is split over multiple segments of the time series, new time points are added just before and just after the discontinuity and added to the solution object
- for models using heavyside functions in certain patterns, the discretisation class automatically adds in discontinuity events required

## Type of change

# Key checklist:

## Further checks:


pybamm/models/base_model.py
```diff
@@ -104,8 +105,9 @@ def __init__(self, name="Unnamed model"):
         self._algebraic = {}
         self._initial_conditions = {}
         self._boundary_conditions = {}
+        self._variables = {}
```
tinosulzer: this should be removed

martinjrobins: done

PR #817, opened 2020-02-11
#759 add changelog

martinjrobins:
# Description

Adds changelog for #759 

Sorry, forgot to add this!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,6 @@
 
 ## Features
 
-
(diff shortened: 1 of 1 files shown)


martinjrobins:
> We have been referring to pull requests rather than issues in the changelog

I'm happy to change this, but note that this particular breaking change was documented in the issue, not the PR.

tinosulzer:
oh right, good point, leave as is then. Thanks!

PR #818, opened 2020-02-11
Issue 613 citing

### LINKED ISSUE #613: Citing PyBaMM
Post paper release, add section to README.md (and docs?) on how to cite pybamm

tinosulzer:
# Description

Add citation for pybamm to readme, and add functionality to print citations for the specific parts (models/solvers/parameters) of pybamm that were used when running a simulation

Fixes #613 

## Type of change

# Key checklist:

## Further checks:


dalonsoa:
Hi all, 

Relevant to this PR, at Imperial we have created [R2T2](https://imperialcollegelondon.github.io/R2T2/) precisely to cover this functionality. We are in the process of polishing it and making sure it serves the community, but it might be worth have a look at it and - hopefully - implement it :) . It is in PyPI and can be installed as any other dependency. 

Please, have a look at it and let me know if you need some clarification or have any comment. 

Best wishes, 
Diego

tinosulzer:
Looks great! I will change to use that and let you know if I have any questions

dalonsoa:
Reading the references from an external file and including the citations using a Key, which is what you use, is not implemented yet, but that specific functionality was in the backlog, so it might be a good time to implement it if that is the way you want to use it.

tinosulzer:
it wouldn't be too bad, we could just keep the `Citations` class and then call them with `@add_reference("short purpose", pybamm.citations["authorYear"])`. But I'm wondering about testing that the right references are added for each functionality, and also being able to call this from a script (in case people don't like running from command line) - have messaged you on slack, might be quicker to sort out

tinosulzer:
@dalonsoa I am tempted to merge this so we have it for v0.2.0 and then update to use R2T2 when that is ready, would that be ok with you?

tinosulzer:
I assumed that this would be covered by convert_to_casadi but no harm adding it anyway

rtimms:
ah I missed that. yeah should be fine as is then

PR #819, opened 2020-02-11
#813 get t_eval from data

### LINKED ISSUE #813: fix t_eval to data for drive cycles
when simulating drive cycles the solvers work if `t_eval` is the same as the times provided in the data. when solving we should check if the current function has "[current data]" and then force the solve times to be those in the data (with a warning message that the solver is fixing t_eval). see below example

```
#
# Simulate drive cycle loaded from csv file
#
import pybamm
import numpy as np
import pandas as pd
import os

# load model

# create geometry

# get drive cycle data
cycle_name = "US06"
drive_cycle = pd.read_csv(
    os.path.join(pybamm.root_dir(), "input", "drive_cycles", cycle_name+".csv"),
    comment="#",
    skip_blank_lines=True
)
time_dim = drive_cycle.values[:, 0]
current_dim = drive_cycle.values[:, 1]

# load parameter values and process model and geometry
param["Current function [A]"] = "[current data]" + cycle_name
param.process_model(model)

# set mesh
var_pts = {var.x_n: 11, var.x_s: 11, var.x_p: 11, var.r_n: 5, var.r_p: 5}
mesh = pybamm.Mesh(geometry, model.default_submesh_types, var_pts)

# discretise model
disc.process_model(model)

# get dimensionless times for evaluation
tau = param.evaluate(pybamm.standard_parameters_lithium_ion.tau_discharge)
t_eval = time_dim / tau  # this works
# t_eval = np.linspace(0, 600 / tau, 600)  # this fails

# simulate US06 drive cycle
solver = pybamm.CasadiSolver(mode="fast")

# plot
plot = pybamm.QuickPlot(solution)
plot.dynamic_plot()
```

Note that this works with Casadi in "fast" mode!

rtimms:
# Description
When using the `Simulation` class with the current provided as data (i..e drive cycles) `solve` now sets `t_eval` (if it isn't provided) to be the times in the data (non-dimensionlised). If `t_eval` is provided, `solve` checks that it contains all of the points in the data, and raises a warning if it doesn't. Maybe this should raise an actual exception to be totally fool proof, but there may be cases where people only want the solution returned at a subset of points. Happy for input on this. I think if people are solving in the long-winded way (i.e. not using `Simulation`) then we should just trust them to pick `t_eval` appropriately. 

Fixes #813 

## Type of change

# Key checklist:

## Further checks:


pybamm/simulation.py
```diff
@@ -340,19 +342,51 @@ def solve(
...
+                    pybamm.logger.info(
+                        "Setting t_eval as specified by the data '{}'".format(filename)
+                    )
+                    t_eval = time_data
+                else:
+                    # If t_eval is provided we check that it at least contains all
+                    # of the data points. We only raise a warning here, as users
+                    # may genuinely only want the solution returned at some
+                    # specified points.
+                    if set(time_data).issubset(set(t_eval)) is False:
```
tinosulzer: I'm wondering if this will have problems with round-off errors? And should be also allow `t_eval` with a sampling time smaller than the smallest gap in the data, but that doesn't hit exactly the time points?

rtimms: yeah good point re round-off errors. yeah that would probably be fine. not sure of the best way of testing this for all cases and give an informative message (e.g. if t_eval or the data non-uniform, would you then just check the biggest gap in t_eval and smallest in the data and just warn that the user should check to ensure they have enough resolution in t_eval?). perhaps doing that is best, as doing something naive like just checking the number of values might pass if e.g. t_eval had loads of points clustered near the start and not many later

tinosulzer: ah yeah I hadn't thought of the non-uniform t_eval case. I think we can allow both options: checking that either the biggest gap in t_eval is smaller than the smallest in the data, or that each point in the data has a corresponding point in t_eval within some small error tolerance, should cover most cases

rtimms: I've updated to just check max dt_eval vs min dt_data as this also catches the case where t_eval = t_data (since dt is equal here). let me know your thoughts and i'll change or merge as is

tinosulzer: I don't think this would  always catch the case t_eval = t_data, for example if `t_eval=t_data = [0,1,5,6]` then `max(t_eval) = 4 > 1 = min(t_data)`

rtimms: oh yeah good point haha

PR #826, opened 2020-02-13
Parameter data distribution

tlestang:
# Description

This PR addresses the distribution of parameter files along with the python package.
First, it fixes the definition of the data files in `setup.py`, which was incorrect.
Second, it moves the `input` directory to the package directory.
Because PyBaMM can now be pip installed, having the `input` in the project root isn't great, since the `input` dir would be installed in the `site-packages` dir, i.e.:
```
my-virtualenv/lib/pythonX.Y/
    site-packages/
        input/
        pybamm/ 
```
The downside is that it makes it less easy for users to manually add parameters.
I added a small command line entry `pybamm_add_parameters` that, given a parameter directory, takes care of copying it to the correct location:
```
(venv) $ pybamm_add_param --help
usage: pybamm_add_param [-h]
                        parameter_dir {lithium-ion,lead-acid}
                        {anodes,cathodes,cells,electrolytes,experiments,separators}

Add a new parameter set

positional arguments:
  parameter_dir         a str containing the name of the parameter directory
  {lithium-ion,lead-acid}
  {anodes,cathodes,cells,electrolytes,experiments,separators}
```
Example:
```
(venv) $  pybamm_add_param my_new_cathode_Lestang_2020 lithium-ion cathodes
```

Finally, parameters are considered as a package resource and are therefore accessed using
the `pkg_resources` module, shipped with `setuptools`, see https://setuptools.readthedocs.io/en/latest/setuptools.html#accessing-data-files-at-runtime

## Type of change

# Key checklist:

## Further checks:


Scottmar93:
I think the `pybamm_add_param` method sounds like a good solution but one issue is that it doesn't show other parameter files that can be used as a template when entering parameters (without these I think it is really pretty hard). Can discuss in the meeting if this would work but might also be good to have a function to pull the whole input folder into the current working directory (e.g. `pybamm_edit_inputs`) and set the path for the inputs to the current working directory so that the files can be easily modified etc? This might also give an option to reset the input files to default values as well.

tlestang:
@Scottmar93 Good point. I'm going to add a command `pybamm_edit_inputs` that copies the input dir in the current working directory, and another `pybamm_update_inputs` that copies the local `input` dir in the package directory.
It would be nice to avoid this second copy step, having pybamm automatically looking for input files preferentially in the current working directory. But then means that users have to be careful from where they execute their script

tlestang:
A few changes:
- PyBaMM first looks in the current directory for a structure `input/parameters/<chemistry>/<component>/<parameter_dir>` when looking up for parameters. It falls back to the package directory in `pybamm/input/...` if nothing is found in current dir.
- I added a command `pybamm_edit_input` to pull a parameter for a given chemistry and component into the current directory for editing.
- It is possible to list all currently available parameters for a given chemistry and component using `pybamm_list_parameters`. The list list parameters available in the package directory, and parameters available from the current directory.

PR #845, opened 2020-02-20
Issue 757 rate capability

### LINKED ISSUE #757: Rate capability plots
**Summary**
Create a script to simulate rate capability experiments (how much capacity can be pulled out before we hit a given limit voltage at CC for different C-rates, see picture below).

image

 **Motivation**
This is a standard experiment on batteries and it would be pretty easy to reproduce in PyBaMM with what is available so far.

**Additional context**

ferranbrosa:
# Description

I have added an example script that generates rate capability plots. The implementation is not the best, so feedback on how to make it more efficient is welcome. In any case, I think it is a good first step to include in v0.2. 

Because it is an example, rather than a new feature, I haven't added to CHANGELOG.md.

Fixes #757 

## Type of change

# Key checklist:

## Further checks:


examples/scripts/DFN_high_Crate.py
```diff
@@ -0,0 +1,48 @@
...
+var = pybamm.standard_spatial_vars
+var_pts = {var.x_n: 30, var.x_s: 30, var.x_p: 30, var.r_n: 10, var.r_p: 10}
+mesh = pybamm.Mesh(geometry, model.default_submesh_types, var_pts)
+
+# discretise model
+disc = pybamm.Discretisation(mesh, model.default_spatial_methods)
+disc.process_model(model)
+
+# solve model
+t_eval = np.linspace(0, 0.2, 1E4)
```
tinosulzer: needs to be updated with t_eval in seconds

examples/scripts/rate_capability.py
```diff
@@ -0,0 +1,48 @@
...
+plt.figure(1)
+plt.scatter(C_rates, capacities)
+plt.xlabel('C-rate')
+plt.ylabel('Capacity [Ah]')
+
+plt.figure(2)
+plt.scatter(currents * voltage_av, capacities * voltage_av)
+plt.xlabel('Power [W]')
+plt.ylabel('Energy [Wh]')
+
```
tinosulzer: nice!

.gitignore
```diff
@@ -91,3 +91,6 @@ cmake_install.cmake
 
 
 third-party/pybind11
+
+# LGM50 results
+results/LGM50/
```
tinosulzer: this can be removed?

examples/scripts/DFN_high_Crate.py
ferranbrosa: I did this such a long time ago that I thought someone else had done it haha. I have removed it now.

.gitignore
ferranbrosa: Done

PR #855, opened 2020-02-24
Add GitHub Actions for CI

dalonsoa:
# Description

Implements GitHub Actions as the CI framework. The rationale for this suggestion is the following:

1. The current implementation uses Travis for CI in Linux and MacOS and Appveyor for Windows. GitHub Actions can handle the three platforms at once. 
2. When using external CI systems, someone forking the repo will need to configure their own CI system. If using GitHub Actions, this CI is built-in for any fork, simplifying future development by external contributors. 
3. The current extensive testing of unit tests and examples seems to be constrained by the time limit imposed by Travis. In GitHub Actions, the time limit per individual job (e.g. using Python 3.7 under Windows) is 6h, which should be more than enough for any sensible testing.
4. Eventually, this same system could be configure to upload packages - in source of wheel form - to PyPI whenever a new tagged commit to the master branch is pushed. 

I've open this PR as a draft because, at the moment, it does not fully cover the features of the existing testing platform - although it does a more extensive testing in other aspects. In particular:

- Tests in MacOS are failing because it cannot find Sundials solvers. I don't understand why this is happening because the installation of the libraries and the PATHs seem correct. Any help with this would be most welcome. 
- I have not included any scheduled testing, although this should not be a problem as [GitHub Actions is perfectly capable of doing it](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#onschedule).

Fixes # None

## Type of change

# Key checklist:

## Further checks:


tinosulzer:
Sounds like very compelling arguments! I'm happy to switch to this if you can get it set up. The only thing is I'd still like to keep test CI time as short as possible even if we can go over 50 minutes (but CRON jobs can be longer)

dalonsoa:
As far as I can tell, this version of the workflow should be able to execute the tests in the three platforms, including those related to Sundials in MacOS, which were failing before. 

I've tried to use the new installation scripts for `scikit` and `KLU` solvers, but I have to admit I'm a bit confused about whether they are two separate set of solvers using both Sundials, if one depend on the other or if they are completely intependent. I hope it makes sense. 

Many thanks to @tlestang for pointing me in the right direction!

.github/workflows/test_on_push.yml
```diff
@@ -0,0 +1,105 @@
...
+      run: |
+        python -m pip install --upgrade pip wheel setuptools
+        pip install -e .
+
+    - name: Install SUNDIALS and SuiteSparse
+      if: matrix.os != 'windows-latest'
+      run: |
+        pip install wget
+        python scripts/setup_KLU_module_build.py
+        pybamm_install_odes --install-sundials
```
tlestang, line 72: `pybamm_install_odes` is an entry point defined in the setup.py, so I'm surprised the command is found before pybamm is installed!

tlestang: `setup_KLU_module_build.py` already installs the sundials in `$HOME/.local/`, so the `pybamm_install_odes` command could be replaced by
```
export SUNDIALS_INST= ${{ HOME }}/.local
pip install scikits.odes
```

.github/workflows/test_on_push.yml
```diff
@@ -0,0 +1,105 @@
...
+      run: |
+        sudo apt install gfortran gcc libopenblas-dev liblapack-dev graphviz
+        sudo apt install python${{ matrix.python-version }}.dev
+    
+    - name: Install MacOS system dependencies
+      if: matrix.os == 'macos-latest'
+      run: |
+        brew update
+        brew install graphviz
+        brew install openblas suitesparse
```
tlestang, line 56: suitesparse isnt required since compiled and installed later on

tlestang:
@dalonsoa I agree that it can be confusing :)
The scikits.odes-based and `idaklu` solver are independent, but they both rely on the Sundials library.

The `idaklu` solver relies on several components of the Sundials library, including the KLU module. This means that, for the `idaklu` module to be compiled successfully, the Sundials must be linked against SuiteSparse. 
Now to use the scikits.odes based solver, you have to `pip install scikits.odes`, which requires some of Sundials components to be installed, but not the KLU component.

`scripts/setup_KLU_module_build.py` is a convenience script that downloads and installs SuiteSparse, then download and installs the Sundials library, including the KLU component.
For users that do not care about the `idaklu` solver but care about the scikits.odes solver, the command `pybamm_install_odes` downloads and install Sundials (without KLU and therefore without SuiteSparse). It also sets the correct envvars and `pip install scikits.odes`.

Does that make sense?

dalonsoa:
All good, now!

PR #856, opened 2020-02-25
Issue 820 release

### LINKED ISSUE #820: v0.2.0 release
Final things to do for release:

- update version number
- update version number in docs
- update authors in docs
- update changelog section title from "unreleased"

anything else? @tlestang is all the pip stuff automatic?

tinosulzer:
# Description

Admin changes for the v0.2.0 release. Default branch will be changed to `develop` after this release.

Fixes #820


tlestang:
Do you get why Travis fails on Mac? 
`integration.test_models.test_full_battery_models.test_lithium_ion.test_dfn.TestDFN` seems to cause the issue.. but no issue on other platforms.

tinosulzer:
I think (but not sure yet) this could be because I updated the scikit-fem version to 0.4.0, it still works but much slower for some reason - so the test times out. We only use scikit-fem for the 2+1D models at the moment

tinosulzer:
Didn't make any difference locally but trying to revert and see if that works better

PR #860, opened 2020-02-26
Add new electrolyte functions from Landesfeind 2019

TomTranter:
# Description

Add new electrolyte functions for conductivity and diffusivity from Landesfeind 2019

Fixes # 859

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/input/parameters/lithium-ion/electrolytes/lipf6_Landesfeind2019/electrolyte_conductivity_Landesfeind2019_EC_DMC_1_1.py
+++ b/pybamm/input/parameters/lithium-ion/electrolytes/lipf6_Landesfeind2019/electrolyte_conductivity_Landesfeind2019_EC_DMC_1_1.py
@@ -1,0 +1,31 @@
+import electrolyte_conductivity_Landesfeind2019_base as base
+import numpy as np
+
+
+def electrolyte_conductivity_Landesfeind2019_EC_DMC_1_1(c_e, T, T_inf, E_k_e, R_g):
+    """
+    Conductivity of LiPF6 in EC:DMC (1:1) as a function of ion concentration and
+    Temperature. The data comes from [1].
+    References
+    ----------
+    .. [1] Landesfeind, J. and Gasteiger, H.A., 2019. Temperature and Concentration
+    Dependence of the Ionic Transport Properties of Lithium-Ion Battery Electrolytes.
+    Journal of The Electrochemical Society, 166(14), pp.A3079-A3097.
+    ----------
+    c_e: :class: `numpy.Array`
+        Dimensional electrolyte concentration
+    T: :class: `numpy.Array`
+        Dimensional temperature
+    Returns
+    -------
+    :`numpy.Array`
+        Electrolyte diffusivity
+    """
+    coeffs = np.array([7.98e-1,
+                       2.28e2,
+                       -1.22,
+                       5.09e-1,
+                       -4e-3,
+                       3.79e-3])
+
+    return base.electrolyte_conductivity_Landesfeind2019_base(c_e, T, coeffs)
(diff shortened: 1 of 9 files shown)


pybamm/input/parameters/lithium-ion/electrolytes/lipf6_Landesfeind2019/electrolyte_conductivity_Landesfeind2019_EC_DMC_1_1.py
```diff
@@ -0,0 +1,31 @@
+import electrolyte_conductivity_Landesfeind2019_base as base
+import numpy as np
+
+
+def electrolyte_conductivity_Landesfeind2019_EC_DMC_1_1(c_e, T, T_inf, E_k_e, R_g):
```
tinosulzer: should we get rid of T_inf, E_k_E and R_g as inputs to these functions in general? When they are needed, they can easily be called from within the function as `pybamm.standard_parameters_lithium_ion.T_inf`, etc

TomTranter: For R_g I would say yes. For T_inf could there be a situation where the reference temperature for different functions is different? We could also use **kwargs potentially although I don't know if that would work with the expression tree. But yeah if you see fewer problems I'm happy to make the changes

pybamm/input/parameters/lithium-ion/electrolytes/lipf6_Landesfeind2019/electrolyte_diffusivity_Landesfeind2019_base.py
```diff
@@ -0,0 +1,32 @@
...
+    -------
+    :`numpy.Array`
+        Electrolyte diffusivity
+    """
+    c = c_e / 1000  # mol.m-3 -> mol.l
+    p1, p2, p3, p4 = coeffs
+    A = p1 * exp(p2 * c)
+    B = exp(p3 / T)
+    C = exp(p4 * c / T)
+    D_e = A * B * C * 1e-6  # m2/s
```
ferranbrosa: Isn't this result in cm^2/s? (eq [18] in the paper)

pybamm/input/parameters/lithium-ion/electrolytes/lipf6_Landesfeind2019/electrolyte_conductivity_Landesfeind2019_EC_DMC_1_1.py
rtimms: I think we should definitely be open to the idea of different functional dependencies of these things, so being less prescriptive in terms of inputs is probably good

pybamm/input/parameters/lithium-ion/electrolytes/lipf6_Landesfeind2019/electrolyte_diffusivity_Landesfeind2019_base.py
TomTranter: Yep good spot.

pybamm/input/parameters/lithium-ion/electrolytes/lipf6_Landesfeind2019/electrolyte_conductivity_Landesfeind2019_EC_DMC_1_1.py
tinosulzer: if it's specific to a function then maybe it should be hard-coded into the function itself (same with E_k_e)

TomTranter: Yeah maybe you are right, If there were a specific reference temperature it would probably be hardcoded into the function if there were other parameters dependent on that.

PR #868, opened 2020-03-01
Issue 830 algebraic

### LINKED ISSUE #830: Solving algebraic models
Fix a few small bugs that make it harder to solve algebraic models:
- casadi solver complains because mass_matrix_inv is empty
- solution complains because len(t) = 1

tinosulzer:
# Description

Reformat algebraic solver to be more in line with the other solvers, and add casadi algebraic solver.
We could call the algebraic solvers to initialize the dae models, would reduce code duplication, but might be a little bit confusing?

Fixes #830

## Type of change

# Key checklist:

## Further checks:


pybamm/solvers/base_solver.py
```diff
@@ -481,6 +507,11 @@ def solve(self, model, t_eval, external_variables=None, inputs=None):
         if len(model.rhs) == 0 and len(model.algebraic) == 0:
             raise pybamm.ModelError("Cannot solve empty model")
 
+        # t_eval can only be None if the solver is an algebraic solver. In that case
+        # set it to 0
+        if self.algebraic_solver is True and t_eval is None:
+            t_eval = np.array([0])
+
```
rtimms: should we need to raise an error here if t_eval is None and the solver isn't algebraic?

PR #873, opened 2020-03-05
#872 added ambient temperature

### LINKED ISSUE #872: Adding option to change ambient temperature profile v time
**Summary**
 We'd like to be able to pass in an ambient temperature profile v. time for it to run during a simulation.

viviantran27:
# Description

Added ambient temperature

Fixes #872 

## Type of change

# Key checklist:

## Further checks:


pybamm/models/submodels/thermal/isothermal/isothermal.py
```diff
@@ -33,6 +35,10 @@ def get_fundamental_variables(self):
         T_cp = T_x_av
 
         variables = self._get_standard_fundamental_variables(T, T_cn, T_cp)
+        variables.update(
```
rtimms: could this be moved to `base_thermal`to avoid repeating in each of the separate base classes?

pybamm/models/submodels/thermal/x_lumped/x_lumped_2D_current_collectors.py
```diff
@@ -23,11 +24,11 @@ def set_rhs(self, variables):
         # TODO: update to allow different cooling conditions at the tabs
         self.rhs = {
             T_av: (
-                pybamm.laplacian(T_av)
-                + self.param.B * pybamm.source(Q_av, T_av)
-                + cooling_coeff * pybamm.source(T_av, T_av)
+                pybamm.laplacian(T_av - T_amb)
```
rtimms: I don't think there should be a T_amb inside the laplacian

pybamm/models/submodels/thermal/x_lumped/x_lumped_2D_current_collectors.py
```diff
@@ -23,11 +24,11 @@ def set_rhs(self, variables):
...
+                + self.param.B * pybamm.source(Q_av, T_av - T_amb)
```
rtimms: the second argument in `pybamm.source` is just the variable for which this is the equation for and the first argument is the source term in the equation (so that boundary conditions are handled correctly in finite elements). sorry, I don't think that is particularly well documented. As a result this should just be `pybamm.source(Q_av, T_av)`

pybamm/models/submodels/thermal/x_lumped/x_lumped_2D_current_collectors.py
```diff
@@ -23,11 +24,11 @@ def set_rhs(self, variables):
...
+                + cooling_coeff * pybamm.source(T_av, T_av - T_amb)
```
rtimms: similar to above this should be `pybamm.source T_av-T_amb,T_av)`

pybamm/models/submodels/thermal/x_lumped/x_lumped_2D_current_collectors.py
rtimms: same for laplacian in other files

pybamm/models/submodels/thermal/x_lumped/x_lumped_2D_current_collectors.py
```diff
@@ -23,11 +24,11 @@ def set_rhs(self, variables):
...
                 - (self.param.h / self.param.delta)
-                * pybamm.source(T_av, T_av, boundary=True)
+                * pybamm.source(T_av - T_amb, T_av - T_amb, boundary=True)
```
rtimms: `pybamm.source T_av-T_amb,T_av, boundary=True)`

pybamm/models/submodels/thermal/x_lumped/x_lumped_1D_current_collectors.py
```diff
@@ -15,11 +15,16 @@ def __init__(self, param):
...
 
         self.rhs = {
-            T_av: (pybamm.laplacian(T_av) + self.param.B * Q_av + cooling_coeff * T_av)
+            T_av: (
+                pybamm.laplacian(T_av - T_amb)
+                + self.param.B * Q_av
+                + cooling_coeff * (T_av - T_amb)
+            )
             / self.param.C_th
         }
```
rtimms: just thought - for the models that are x-lumped with 1D and 2D current collectors I think you also need to change the boundary conditions to `T-T_amb` as well

rtimms:
is there an implicit assumption that the dimensionless ambient temp is zero at t=t0? i.e. you start at the reference temperature? if so, maybe there could be a check for this?

pybamm/models/submodels/thermal/x_lumped/x_lumped_2D_current_collectors.py
viviantran27: ohh i see, thanks for the clarification

viviantran27:
> is there an implicit assumption that the dimensionless ambient temp is zero at t=t0? i.e. you start at the reference temperature? if so, maybe there could be a check for this?

Should we keep T_init separate from T_amb(0)? For example, it could be used to simulate a case where an EV exits an air-conditioned garage to a different temperature outside or other rapid temperature change. Can you clarify what sort of check you're imagining?

rtimms:
> > is there an implicit assumption that the dimensionless ambient temp is zero at t=t0? i.e. you start at the reference temperature? if so, maybe there could be a check for this?
> 
> Should we keep T_init separate from T_amb(0)? For example, it could be used to simulate a case where an EV exits an air-conditioned garage to a different temperature outside or other rapid temperature change. Can you clarify what sort of check you're imagining?

ah yeah I hadn’t thought of that use case. Yeah I agree to keep T_init and T_amb(0) separate. I had in my head that the battery would start at the ambient temperature, but I see now that isn’t the case, so no need to check. Thanks!

examples/scripts/DFN_ambient_temperature.py
```diff
@@ -0,0 +1,52 @@
+#
+# Example showing how to load and solve the DFN
```
rtimms: the title could be updated to explain this is showing the ambient temperature feature

tinosulzer: I'll merge this PR and make that change before merging #868

PR #881, opened 2020-03-11
Get function args

### LINKED ISSUE #880: add function parameter helper function
as title

Scottmar93:
# Description

A helper function to get information on symbols in a model. Accessed by doing model.info("Negative electrode diffusivity [m2.s-1]"). This then displays class of the symbol associated with that name. If it is a functionparameter it lists the parameters it requires in this model. 

## Type of change

# Key checklist:

## Further checks:


rtimms:
this is very useful. do we need the string description of the inputs? or can we just get them from symbol.name? then FunctionParameter can just take the function name and a list of symbols?

Scottmar93:
@tinosulzer I am happy to change to a dict if we feel this is less messy. However, I opted for an optional list as it was the minimal change and definitely wouldn't interfere with any existing functionality (i.e. to_python or casadi and back etc which I didn't want to deal with but maybe this isn't a problem?). I think if we do go for the dict though, that we should enforce it consistently. Thoughts? 

@rtimms 
Unfortunately we can't use `symbol.name` as sometimes symbols will be a concatenation or combinations of operated on symbols :( 
The logic of using model.info() is that we could in theory define two models with the same function name but which require different inputs. So the form of the function is set by how it is defined in the model. 
Yeah, longer-term `model.info()` could serve to print out some nice summary information of the model. E.g. number of algebraic equations, number of non-algebraic equations, number of parameters, number of variables, etc. Then passing arguments could give you more specific information e.g.  `model.info(parameter="Negative electrode diffusivity [m2.s-1]")`. Would also be really nice to pass in a parameter set and determine whether it is compatible with the model, then print out parameters that the model needs that aren't in the set, but yeah longer term...

tinosulzer:
It should be fine because `to_python` and `casadi` both act on the already-built expression tree. Agree dict should be used consistently

tests/unit/test_models/test_submodels/test_external_circuit/test_function_control.py
```diff
@@ -9,7 +9,13 @@
 def external_circuit_function(variables):
     I = variables["Current [A]"]
     V = variables["Terminal voltage [V]"]
-    return V + I - pybamm.FunctionParameter("Current plus voltage function", pybamm.t)
+    return (
+        V
+        + I
+        - pybamm.FunctionParameter(
+            "Current plus voltage function", {"Time [s]": pybamm.t}
```
tinosulzer: nb: time isn't in seconds here, will open separate issue

pybamm/expression_tree/parameter.py
```diff
@@ -48,19 +48,21 @@ class FunctionParameter(pybamm.Symbol):
 
     name : str
         name of the node
-    child : :class:`Symbol`
-        child node
+    inputs : dict
+        A dictionary with string keys and :class:`pybamm.Symbol` values representing
```
rtimms: maybe say what the string should be i.e. useful/descriptive names for the inputs

pybamm/expression_tree/parameter.py
```diff
@@ -76,6 +78,28 @@ def __init__(self, name, *children, diff_variable=None):
...
+                print(inp)
+
+    @input_names.setter
+    def input_names(self, inp=None):
+        if inp:
+            if inp.__class__ is list:
+                for i in inp:
+                    if i.__class__ is not str:
+                        raise TypeError(
+                            "Inputs must be a provided as a list of strings"
```
rtimms: inputs is now a dict with string keys rather than a list of strings

pybamm/models/base_model.py
```diff
@@ -582,3 +608,38 @@ def default_solver(self):
...
+    dics = [
+        model.rhs,
+        model.algebraic,
+        model.variables,
+    ]
+    for dic in dics:
+        dic_return = find_symbol_in_dict(dic, name)
+        if dic_return:
+            return dic_return
+    print(name + " not found")
```
rtimms: do we want this to print?

pybamm/models/submodels/external_circuit/function_control_external_circuit.py
```diff
@@ -58,7 +58,7 @@ def __init__(self, param):
 
 def constant_voltage(variables):
     V = variables["Terminal voltage [V]"]
-    return V - pybamm.FunctionParameter("Voltage function [V]", pybamm.t)
+    return V - pybamm.FunctionParameter("Voltage function [V]", {"Times [s]": pybamm.t})
```
rtimms: Time [s]

squared.py
```diff
@@ -0,0 +1,3 @@
+
+def squared(x):
```
rtimms: this file can go?

tests/unit/test_models/test_submodels/test_external_circuit/test_function_control.py
Scottmar93: Happy to remove the seconds if want it to be dimensionless? Or thinking of changing input to be in seconds?

tinosulzer: I think we should do pybamm.t * self.param.timescale so that it is a dimensional time (that is how people will define it externally)

pybamm/models/base_model.py
```diff
@@ -614,6 +623,32 @@ def check_variables(self):
...
+            return None
+
+        print(div)
+        print(symbol_name, "\n")
+        print(type(symbol))
+
+        if isinstance(symbol, pybamm.FunctionParameter):
+            print("")
+            print("Inputs:")
+            symbol.input_names
```
tinosulzer: should this be `print("Inputs:\n{}".format(symbol.input_names)`?

tests/unit/test_expression_tree/test_parameter.py
```diff
@@ -22,22 +22,58 @@ def test_evaluate_for_shape(self):
...
+        a = pybamm.Parameter("a")
+        func = pybamm.FunctionParameter("func", {"2a": 2 * a})
+
+        new_func = func.new_copy()
+        self.assertEqual(func._input_names, new_func._input_names)
+
+    def test_print_input_names(self):
+        var = pybamm.Variable("var")
+        func = pybamm.FunctionParameter("a", {"var": var})
+        func.input_names
```
tinosulzer: not sure what this test is supposed to do?

pybamm/models/base_model.py
Scottmar93: `symbol.input_names` just prints the input names. Can change this though if you prefer it to return a list of the input names and print here instead.

tests/unit/test_expression_tree/test_parameter.py
Scottmar93: It tests that printing input names works without raising an error. As above `func.input_names` will print the imput names

pybamm/models/base_model.py
tinosulzer: I see, this is a little bit confusing because (as I understand it) a property should always return something, and a function that prints something should be called with `()`.
I would suggest doing this instead:
```python3
@property
def input_names(self):
    return self._input_names # a property should always return something

def print_input_names(self): # call this as .print_input_names() to print
    for inp in self._input_names:
         print(inp)
```

rtimms: yeah, I'm happy when this is changed

PR #891, opened 2020-03-15
Issue 609 broadcast to cell edges

### LINKED ISSUE #609: Broadcast to cell edges
We should allow for Broadcasts onto cell edges if required by the spatial method. This is needed e.g. for `fast_diffusion`, in `constant_stefan_maxwell_diffusion` and in `isothermal` models. At the moment these all broadcast zero flux to the appropriate domain, but this ends up being evaluated on cell centers, not edges.

tinosulzer:
# Description

Add `BroadcastToEdge` object and edit fluxes in relevant submodels to use this.
Also edited the particle submodels, hopefully a bit clearer now.

Fixes #609 

## Type of change

# Key checklist:

## Further checks:


pybamm/expression_tree/unary_operators.py
```diff
@@ -901,6 +901,9 @@ def x_average(symbol):
     :class:`Symbol`
         the new averaged symbol
     """
+    # Can't take average if the symbol evaluates on edges
```
rtimms: why can't we average on edges? is it just that it isn't implemented? (if so, no need to do this now. it can be a separate issue if/when we need it)

tinosulzer: No reason why we can't in theory. I'm not sure what the proper implementation would be in a finite volume context though? For nodes it's clearly just sum all the values, but for edges I don't know what it should be: maybe `N[-1] - N[0]`?

PR #926, opened 2020-03-27
Issue 892 install windows

### LINKED ISSUE #892: Update Windows installation instructions
The Windows installation instructions still suggest using WSL. We should add the standard Windows installation instructions (I would still keep the current instructions using WSL, but aimed maybe to more advanced users). Also, @rtimms suggested double-checking the other instructions.

ferranbrosa:
# Description

Added basic Windows installation instructions (very basic, just for the workshop). I kept the WSL instruction for more advanced users (maybe they can be removed, but I find WSL quite useful).

Fixes #892

## Type of change

Minor change in the documentation, so none of the points below applies.

# Key checklist:

## Further checks:

### DIFF
--- a/INSTALL-WINDOWS.md
+++ b/INSTALL-WINDOWS.md
@@ -2,7 +2,9 @@
 
 To use and/or contribute to PyBaMM, you must have Python 3.6 or 3.7 installed (note that 3.8 is not yet supported).
 
-To install Python 3 download the installation files from [Python's website](https://www.python.org/downloads/windows/).
+To install Python 3 download the installation files from [Python's website](https://www.python.org/downloads/windows/). Make sure
+to tick the box on `Add Python 3.X to PATH`. For more detailed instructions please see the 
+[official Python on Windows guide](https://docs.python.org/3.7/using/windows.html).
 
 ## Install PyBaMM
 
--- a/INSTALL-WINDOWS.md
+++ b/INSTALL-WINDOWS.md
@@ -1,89 +1,52 @@
-We recommend the use of Windows Subsystem for Linux (WSL) to install PyBaMM, see the
-instructions below to get PyBaMM working using Windows, WSL and VSCode.
+## Prerequisites
 
-## Install WSL
+To use and/or contribute to PyBaMM, you must have Python 3.6 or 3.7 installed (note that 3.8 is not yet supported).
 
-Follow the instructions from Microsoft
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -61,8 +61,7 @@ For instructions on installing PyBaMM on Mac OS distributions, please see [here]
 
 ### Windows
 
-We recommend using Windows Subsystem for Linux to install PyBaMM on a Windows OS, for
-instructions please see [here](INSTALL-WINDOWS.md)
+For instructions on installing PyBaMM on Windows distributions, please see [here](INSTALL-WINDOWS.md)
```
tinosulzer: Can you add a disclaimer that WSL is needed for the advanced install with scikit-odes and klu solvers?

ferranbrosa: I have added the disclaimer now.

INSTALL-WINDOWS.md
```diff
@@ -1,89 +1,52 @@
...
 
-## Install WSL
+To use and/or contribute to PyBaMM, you must have Python 3.6 or 3.7 installed (note that 3.8 is not yet supported).
 
-Follow the instructions from Microsoft
-[here](https://docs.microsoft.com/en-us/windows/wsl/install-win10). When given the
-option, choose the Ubuntu 18.04 LTS distribution to install. Don't forget to initialise
-the Ubuntu installation using the instructions given
-[here](https://docs.microsoft.com/en-us/windows/wsl/initialize-distro).
+To install Python 3 download the installation files from [Python's website](https://www.python.org/downloads/windows/).
```
tlestang: We could point to the [official python on Windows guide](https://docs.python.org/3.7/using/windows.html).
The most straightforward way would be to use the [full installer](https://docs.python.org/3.7/using/windows.html#windows-full).
Actually the command `python` is not automatically available, unless the `Add python 3.7 to PATH` option is ticked on the installer.
I think something that would always work is to use the [python launcher](https://docs.python.org/3.7/using/windows.html#python-launcher-for-windows) `py`.

So
```
py -m venv env
env\Scripts\activate.bat
pip install pybamm
```
I'm not sure if its worth putting in so much detail, we could maybe just point to the windows guide?

ferranbrosa: I have now added a link to the guide, and a line reminding them to tick the box, so the command `python` should work. Let me know if I should still change it to `py` and I will fix it.

ferranbrosa: BTW, feel free to merge the PR when you think it is fine, as I can't.

PR #931, opened 2020-03-31
Version 0.2.1 develop -> master

tinosulzer:
Merge develop into master for release


tinosulzer:
ah crap, ok well spotted, I'll just fix this for the next release so we can get this one out in time for the workshop

PR #932, opened 2020-04-02
Basic lead acid model

tinosulzer:
# Description

Add basic lead-acid model (with convection)

## Type of change

# Key checklist:

## Further checks:


pybamm/models/full_battery_models/lead_acid/basic_full.py
```diff
@@ -0,0 +1,286 @@
...
+    used instead.
+
+    Parameters
+    ----------
+    name : str, optional
+        The name of the model.
+
+    References
+    ----------
+    .. [2] V Sulzer, SJ Chapman, CP Please, DA Howey, and CW Monroe. Faster lead-acid
```
ferranbrosa, line 23: Minor typo: should the reference be [1]?

pybamm/models/full_battery_models/lead_acid/full.py
```diff
@@ -6,7 +6,7 @@
 
 
 class Full(BaseModel):
-    """Porous electrode model for lead-acid, from [1]_, based on the Full
+    """Porous electrode model for lead-acid, from [1]_, based on the Newman-Tiedemann
```
ferranbrosa, line 9: Minor typo: should the underscore be removed?

pybamm/models/full_battery_models/lead_acid/basic_full.py
tinosulzer: It has to be a different number because "full" and "basicfull" are in the same doc file and sphinx doesn't like it if the same ref number is used twice

pybamm/models/full_battery_models/lead_acid/full.py
tinosulzer: this is for sphinx cross-ref

PR #940, opened 2020-04-04
Issue 937 sensitivity

### LINKED ISSUE #937: Add sensitivity to casadi root-finder
The algebraic solver `CasadiAlgebraicSolver` seems like the easiest one to tackle first for adding sensitivities. Just dumping some thoughts in this issue.

### Casadi workflow
```python3
import casadi
x = casadi.MX.sym("x")
p = casadi.MX.sym("p")
alg = 2 * x - p
x0 = 1

roots = casadi.rootfinder("roots", "newton", dict(x=x, p=p, g=alg))
roots_p = roots(x0, p)
jac = casadi.jacobian(roots_p, p)
root_fn = casadi.Function("roots", [p], [roots_p])
jac_fn = casadi.Function("j", [p], [jac])

print(root_fn(3), jac_fn(3)) # returns 1.5, 0.5
```

### Proposed change

Optimizers like `scipy.optimize.minimize` have the option of setting `jac=True` which then assumes that the function being minimized returns both a value and a sensitivity. 
So the solution object needs to return both a value and a jacobian (e.g. via a method `Solution.value_and_sensitivity("variable")`)

Two options:
1. Compute both `sol.y` and `jacobian(sol.y, p)` in the solver, and evaluate at a given input, then store both of these in the solution object (i.e. a new `Solution` object is created at each time step)
```python3
def error_and_jacobian(x):
     sol = solver.solve(model, t_eval, inputs={"a": x}, sensitivity=True)
     value, sens = sol.value_and_sensitivity("variable")
     return value - data, jac
```
2. Return a new object type `SymbolicSolution`, which can be evaluated repeatedly with different input values, e.g.
```python3
sym_sol = solver.solve(model, t_eval, inputs={"a": "[symbolic]"})

def error_and_sensitivity(x):
     value, sens = sym_sol.value_and_sensitivity("variable", x)
     return value - data, sens
```

tinosulzer:
# Description

Added sensitivity to `CasadiAlgebraicSolver` class. To do this, added a `CasadiSolution` and `ProcessedCasadiVariable` class, which can hold a symbolic solution (i.e. inputs are symbolic, not yet provided). The value and sensitivity of a `ProcessedCasadiVariable` can then be found for a specific input by calling `.value(inputs)` or `.sensitivity(inputs)`.

I haven't added any examples or much extra docs for this, I think it can wait until we make this a more fleshed out feature. The next step would be to do the `CasadiSolver`.

Fixes #937 

## Type of change

# Key checklist:

## Further checks:


tests/unit/test_solvers/test_casadi_algebraic_solver.py
```diff
@@ -97,10 +116,96 @@ def test_solve_with_input(self):
...
+        model.initial_conditions = {var: 2}
+        model.variables = {"var": var}
+
+        # create discretisation
+        disc = pybamm.Discretisation()
+        disc.process_model(model)
+
+        # Solve
+        solver = pybamm.CasadiAlgebraicSolver()
+        solution = solver.solve(model, [0], inputs={"param": "[sym]"})
```
martinjrobins, line 138: I'm getting a bit confused about what is "symbolic" :)  I could consider a `pybamm.InputParameter` to be symbolic, so I don't see why I would need to specify again that the input parameter is symbolic. I think the default behaviour of the inputs should be symbolic. That is, if I don't provide a value when `solve` is called, then this input parameter is treated as symbolic

pybamm/solvers/processed_casadi_variable.py
```diff
@@ -0,0 +1,223 @@
+#
+# Processed Variable class
+#
+import casadi
+import numbers
+import numpy as np
+
+
+class ProcessedCasadiVariable(object):
```
martinjrobins, line 9: It is feasible that we could implement this (a symbolic processed variable) for non-casadi solvers, so I don't think we should use Casadi in the class name. Perhaps `ProcessedSymbolicVariable`?

pybamm/solvers/processed_casadi_variable.py
```diff
@@ -0,0 +1,223 @@
...
+
+        self.first_dim_pts = nodes
+
+    def value(self, inputs=None, check_inputs=True):
+        """
+        Returns the value of the variable at the specified input values
+
+        Parameters
+        ----------
+        inputs : float, array-like, or dict
```
martinjrobins, line 156: can you provide more information about how input parameters are ordered in the input array? A dict does not have any particular order, so how does the input parameter keys get mapped to the vector?

tests/unit/test_solvers/test_casadi_algebraic_solver.py
```diff
@@ -97,10 +116,96 @@ def test_solve_with_input(self):
...
+        np.testing.assert_array_equal(solution["var"].value(7), -7)
```
martinjrobins, line 139: Do you test the case where the input to value is a dict? I couldn't see this on a quick skim through the code

tests/unit/test_solvers/test_casadi_algebraic_solver.py
```diff
@@ -97,10 +116,96 @@ def test_solve_with_input(self):
...
+        np.testing.assert_array_equal(solution["var"].value(3), -3)
+        np.testing.assert_array_equal(solution["var"].sensitivity(3), -1)
```
martinjrobins, line 141: same here, test for case where input values are a dict

pybamm/solvers/processed_casadi_variable.py
tinosulzer: We could check `list(inputs.keys())` is the same as expected?

tests/unit/test_solvers/test_casadi_algebraic_solver.py
tinosulzer: Checked this in the test for `ProcessedCasadiVariable` but can add a test here too

tests/unit/test_solvers/test_casadi_algebraic_solver.py
tinosulzer: Ok - the solver has to know what shape the inputs have though, so I the model could keep track of this? Then the order in which the inputs were passed when solving will have to be read from the model to. Unless we always force `ProcessedSymbolicVariable.value()` to take a dictionary as input, and then the values get ordered in the same order as the model's inputs before being stacked?

tinosulzer:
To be honest, the lack of example/documentation was intentional as I want to keep this "semi-private" until we are settled on the API. Also, it currently only works with algebraic-only models, i.e. not any of the battery models

ferranbrosa:
Then it looks good to me :)

TomTranter:
> To be honest, the lack of example/documentation was intentional as I want to keep this "semi-private" until we are settled on the API. Also, it currently only works with algebraic-only models, i.e. not any of the battery models

But the issue is that I don't really know how to test it

pybamm/input/parameters/lead-acid/cells/BBOXX_Sulzer2019/parameters.csv
```diff
@@ -19,7 +19,7 @@ Positive tab centre z-coordinate [m],0.114,Tab at top,
 # Electrical,,,
 Cell capacity [A.h],17,Manufacturer,
 Typical current [A],1,,
-
+Current function [A],1,default current function,
```
martinjrobins, line 22: are these changes to that parameters supposed to be in this PR?

pybamm/parameters/parameter_values.py
```diff
@@ -265,64 +278,22 @@ def update(self, values, check_conflict=False, check_already_exists=True, path="
...
-    def check_and_update_parameter_values(self, values):
+    def check_parameter_values(self, values):
         # Make sure typical current is non-zero
         if "Typical current [A]" in values and values["Typical current [A]"] == 0:
             raise ValueError(
                 "'Typical current [A]' cannot be zero. A possible alternative is to "
                 "set 'Current function [A]' to `0` instead."
             )
-        if "C-rate" in values and "Current function [A]" in values:
+        if "C-rate" in values:
```
martinjrobins, line 291: are these supposed to be in this PR, doesn't seem like these are anything to do with sensitivities?

tinosulzer:
I merged in another branch which was about to be merged to develop, thought those changes would disappear once that branch got merged

PR #956, opened 2020-04-14
Issue 853 casadi safe

### LINKED ISSUE #853: casadi safe mode
The solution time when using casadi "safe" mode depends on the number of values in `t_eval`. For example
```
t_eval = np.linspace(0, 3600, 100)
sim.solve(t_eval=t_eval, solver=pybamm.CasadiSolver(mode="safe"))
```
takes around 1.6s but 
```
t_eval = np.linspace(0, 3600, 10)
sim.solve(t_eval=t_eval, solver=pybamm.CasadiSolver(mode="safe"))
```
only takes 0.2 s.

The current way safe mode is implemented is to integrate over each `dt` in `t_eval`. If there are lots of points in `t_eval` the result in the solver restarting a lot, which is slow.

There are some alternatives, each of which has pros and cons (here `t_eval = [t0, t1, ... tf]`):

1) Just integrate in fast mode until the end, and then truncate the solution at the first event after the fact. Without any extra checks this may give `SolverError`s e.g. if concentration goes negative
2) Try to integrate from `t0` to `tf` (outputting at all the points in `t_eval`): If success then truncate as above, if fail then try to integrate from `t0` to `tf/2` (again outputting at all points in `t_eval` that fall in this range). Then repeat, halving the window you are integrating over until the solver finishes successfully. If the event is near the start then this may lead you to unnecessarily integrate for longer than you need to.
3) Something similar to the above, but instead try to integrate in "sensible" chunks. What sensible means here is up for debate, but it could be something like try to integrate in blocks of 3600s/c_rate
4) Something else???

I think for most use cases all of those are reasonable. Does anyone have any input as to why one may be better than the other, or have other suggestions for tackling events in casadi?

rtimms:
# Description
Changes the behaviour of casadi "safe" mode. Instead of creating a new integrator for each `dt` in `t_eval` (which scales badly with the number of points in `t_eval`), it now integrates in larger "global" timesteps and checks if an event has been triggered. If an event is triggered in some interval `[t, t+dt]` then the earliest event time is found by a root finding algorithm and the state at that time is found by interpolation. The solution is then truncated and returned up to the event time. 

Fixes #853 

## Type of change

# Key checklist:

## Further checks:


rtimms:
question: at the moment `dt_max` is set somewhat arbitrarily. should we allow this to be passed as an option so that users have control over the “global” step?

pybamm/solvers/base_solver.py
```diff
@@ -593,10 +593,6 @@ def solve(self, model, t_eval=None, external_variables=None, inputs=None):
         # Set up external variables and inputs
         ext_and_inputs = self._set_up_ext_and_inputs(model, external_variables, inputs)
 
-        # Make sure t_eval is monotonic
-        if (np.diff(t_eval) < 0).any():
-            raise pybamm.SolverError("t_eval must increase monotonically")
-
```
tinosulzer, line 599: why did you remove this?

rtimms: the same check is performed on line 589

rtimms:
thanks, just need to fix a problem where sometimes casadi gets stuck doing `Linear solve failed psetup failed` forever when you step too far. It doesn’t get caught in the try/except, so I need to find the right option the catches that, but doesn’t cause it to bail out early in other cases.

tinosulzer:
Maybe some kind of timeout? https://medium.com/@chaoren/how-to-timeout-in-python-726002bf2291

PR #965, opened 2020-04-20
Issue 912 tab cooling

Scottmar93:
# Description

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:


pybamm/input/parameters/lead-acid/experiments/1C_discharge_from_full/parameters.csv
```diff
@@ -5,8 +5,11 @@ Name [units],Value,Reference,Notes
 Reference temperature [K],294.85,Room temperature,
 Maximum temperature [K],333.15,,
 Ambient temperature [K], 294.85,,
-Heat transfer coefficient [W.m-2.K-1],10,,
-Initial temperature [K],294.85,Room temperature,
+Negative current collector surface heat transfer coefficient [W.m-2.K-1],0,,
```
rtimms, line 8: is there a reason the defaults aren't just uniform cooling from all surfaces (I think this might be what people would expect?)

pybamm/models/submodels/thermal/pouch_cell/pouch_cell_1D_current_collectors.py
```diff
@@ -60,40 +60,82 @@ def set_rhs(self, variables):
...
 
     def set_boundary_conditions(self, variables):
         T_amb = variables["Ambient temperature"]
         T_av = variables["X-averaged cell temperature"]
-        T_av_left = pybamm.boundary_value(T_av, "negative tab")
-        T_av_right = pybamm.boundary_value(T_av, "positive tab")
+        T_av_top = pybamm.boundary_value(T_av, "right")
+        T_av_bottom = pybamm.boundary_value(T_av, "left")
+
+        # Tab cooling only implemented for both tabs at the top.
```
rtimms, line 97: is there a way to generalize this by using a linear combination of the two so that the appropriate cooling switches on/off depending on tab location? just a little concerned that this comment is buried in the model and people will use it with tabs in different locations without realizing it isn't implemented

pybamm/input/parameters/lead-acid/experiments/1C_discharge_from_full/parameters.csv
Scottmar93: I thought people would be more expecting the model to represent a single layer in the center of a cell as opposed to a single layer being cooled from each surface, since this is never really seen. 

I really don't mind what the defaults are though, so feel free to change them to whatever you think is clearest

rtimms: ah ok. I was thinking of an idealized single layer cell that might be used e.g. in a lab. Not sure I have a strong preference, but I think most people will expect to see a single value for h by default

pybamm/models/submodels/thermal/pouch_cell/pouch_cell_1D_current_collectors.py
rtimms: I was just thinking you could e.g. edge cooling everywhere then add something like (1-tab_z_coord)*tab_cool _correction (and all the other appropriate combinations to get the right thing) 

yeah seems like a lot of work. is there somewhere more explicit we can warn that this only does tab cooling from the top?

PR #971, opened 2020-04-28
Issue 963 misc bugs

### LINKED ISSUE #963: Better error messages
Some small changes / better error messages needed for increased usability:
- Raise error when taking gradient of an object with no domain (but gradient of broadcast should work and return zero)
- Raise error when trying to do boundary value of a gradient object
- ~~Allow re-processing discretised model~~ Raise error when trying to re-process discretised model, and recommend using `inplace=False`

tinosulzer:
# Description

- Add some extra checks when creating model
- Make grad (broadcast) and div(broadcast) automatically return broadcast(0)
- don't allow rediscretising a model once discretised (if allowed, it usually fails anyway)

Fixes #963 

## Type of change

# Key checklist:

## Further checks:


pybamm/solvers/casadi_solver.py
```diff
@@ -4,6 +4,22 @@
 import casadi
 import pybamm
 import numpy as np
+from scipy.interpolate import interp1d
```
rtimms, line 7: looks like this is out of date with the changes to casadi, but should be fixed by merging

tinosulzer: Ah thanks, I was wondering why the tests were failing

PR #995, opened 2020-05-14
Rk4 solver

pmohtat:
# Description

Added the fixed-step Runge–Kutta method solver (RK4). The main motivation is that with a manual solver we can then build on with better solvers (which we can then export to C)

Fixes # (not an issue)

## Type of change

PR #s 995 was added to [CHANGELOG.md](https://github.com/pybamm-team/PyBaMM/blob/master/CHANGELOG.md) 

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/solvers/rk4_ode_solver.py
+++ b/pybamm/solvers/rk4_ode_solver.py
@@ -74,17 +74,6 @@ class RK4Solver(pybamm.BaseSolver):
             )
         pybamm.logger.info("Start solving {} with {}".format(model.name, self.name))
 
-<<<<<<< HEAD
-=======
-        if (
-            pybamm.logger.getEffectiveLevel() == 10
-            or pybamm.settings.debug_mode is True
-        ):
-            show_eval_warnings = True
-        else:
-            show_eval_warnings = False
-
->>>>>>> 6a18a7dacaed63beb88326b26bb563139211691f
         # set up and solve
         U = inputs
         DT = casadi.MX.sym("DT")
--- a/pybamm/solvers/rk4_ode_solver.py
+++ b/pybamm/solvers/rk4_ode_solver.py
@@ -51,8 +51,9 @@ class RK4Solver(pybamm.BaseSolver):
         # convert inputs to casadi format
         inputs = casadi.vertcat(*[x for x in inputs.values()])
 
-        integrator = self.get_integrator(model, t_eval, inputs)
+        integrator, F = self.get_integrator(model, t_eval, inputs)
         solution = self._run_integrator(integrator, model, model.y0, inputs, t_eval)
+        solution.F = F
         return solution
 
     def get_integrator(self, model, t_eval, inputs):
@@ -173,7 +174,7 @@ class RK4Solver(pybamm.BaseSolver):
 
             solution.termination = "final"
 
-        return solution
+        return solution, F
 
     def _run_integrator(self, integrator, model, y0, inputs, t_eval):
         try:
--- a/pybamm/solvers/rk4_ode_solver.py
+++ b/pybamm/solvers/rk4_ode_solver.py
@@ -1,0 +1,120 @@
+#
+# Solver class using fixed step Runge-Kutta 4 integrator
+#
+import casadi
+import pybamm
+import numpy as np
+
+
+class RK4Solver(pybamm.BaseSolver):
+    """Solve a discretised model, using RK4.
+
+    **Extends**: :class:`pybamm.BaseSolver`
+
+    Parameters
+    ----------
+
+
+    """
+
+    def __init__(
+        self,
+        N,  # number of control intervals
+        rtol=1e-6,
+        atol=1e-6,
+        root_method="casadi",
+        root_tol=1e-6,
+        # extra_options_setup=None,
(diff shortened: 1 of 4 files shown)


martinjrobins:
hi @pmohtat, thank you for the PR! Can you explain a bit more about the reasoning behind adding another implementation of RK4? the scipy solver in pybamm already includes this method. I'm all for implementing better solvers, but I'd be reluctant to add methods we already have

pmohtat:
hi @martinjrobins, yes you are correct. But, the main motivation here was code generation (C++) for embedded controllers, simulink, etc. I wrote the integrator using CasAdi functions, so in theory we can use CasAdi's code generation to get the C code. Also, this way we are able to modify the solver freely and add needed functionality (e.g. adding algebraic equations to Rk4)

martinjrobins:
OK, so the aim is to implement a faster version of rk4 than the scipy one, using casadi code generation? Is the code gen implemented in this pr? Be good to test if this is possible before we merge it in

pmohtat:
we have tested the code generation on a separate project. I can add a code generation test in the test_rk4_solver.py.

pmohtat:
hi. I have added a test in test_rk4_solver.py. I was able to compile and run the code both from command line and using Matlab's mex function generator.
Let me know if you need more information.

tests/unit/test_solvers/test_rk4_solver.py
```diff
@@ -0,0 +1,247 @@
...
+        disc = pybamm.Discretisation()
+        model_disc = disc.process_model(model, inplace=False)
+        # Solve
+        solver = pybamm.RK4Solver(N=20)
+        t_eval = np.linspace(0, 1, 100)
+        solution = solver.solve(model_disc, t_eval)
+        opts = dict(main=True, mex=True)
+        solution.F.generate("gen.c", opts)
+        # To test the generated code run the following in the terminal
+        # gcc gen.c -o gen
```
martinjrobins, line 226: if code generation is going to be a feature of this solver, how to do this should be documented as part of the class

pybamm/solvers/rk4_ode_solver.py
```diff
@@ -0,0 +1,186 @@
...
+        solution.F = F
+        return solution
+
+    def get_integrator(self, model, t_eval, inputs):
+        # Only set up problem once
+        y0 = model.y0
+        if isinstance(y0, casadi.DM):
+            y0 = y0.full().flatten()
+        rhs = model.casadi_rhs
+        # algebraic = model.casadi_algebraic
```
martinjrobins, line 65: what if this solver is used on a dae model? Should fail gracefully with some sort of error message

pmohtat:
hi. thank you for the feedback, I'll hopefully start addressing them in coming days. I apologies for the many mistakes, I am pretty much a noob when it comes to pybamm and git...
I just want to reiterate that the motivation was code generation first and a new solver second. That's why I started with an easy to implement solver such as RK4.
If code generation is not a feature of interest l completely understand. Also, if you think there is better way to approach this without adding a custom solver let me know or point me in that direction. thanks again and happy to learn and contribute in any capacity!

martinjrobins:
no worries at all, thank you for all your work contributing to pybamm! I think the code generation would be a good feature, but your solver still needs to conform to the interface established in `pybamm.BaseSolver`. Looking at the API docs (https://pybamm.readthedocs.io/en/latest/source/solvers/base_solver.html), you can see that the constructor takes `atol` and `rtol` as absolute and relative error tolerances to use for the solver. Therefore each solver must satisfy these error tolerances, whatever the specific time-points given to the solver.

pybamm/solvers/rk4_ode_solver.py
```diff
@@ -0,0 +1,186 @@
...
+
+            solution.termination = "final"
+
+        return solution, F
+
+    def _run_integrator(self, integrator, model, y0, inputs, t_eval):
+        try:
+            sol = integrator
+            y_values = sol.y
+            return pybamm.Solution(sol.t, y_values)
```
tinosulzer, line 183: I don't thing the `get_integrator` / `run_integrator` structure is appropriate in this case. Here, the integrator is not a function, but already the solution, and `get_integrator` and `run_integrator` are only called once each. So you could put the whole thing in the `_integrate` function

pybamm/solvers/rk4_ode_solver.py
tinosulzer: @pmohtat you can set `self.ode_solver=True` in the init function to do this check automatically, and remove this commented-out line

martinjrobins:
Yes, I think this is not far off and will be a really good addition to the pybamm solvers. The main effort would be satisfying the tolerances, as this requires error estimation and interpolation to the points in `t_eval`. I've previously implemented a RK4(5) method with adaptive time-stepping and interpolation here (https://github.com/martinjrobins/swift-odeint). It's in a different language, but has citations you can read, and I'm happy to answer any questions you have about the algorithm. You can also have a look at the RK methods in scipy (https://github.com/scipy/scipy/blob/v1.4.1/scipy/integrate/_ivp/ivp.py#L157-L664), which have the advantage of being in python!

PR #996, opened 2020-05-14
Use Tox

tlestang:
# Description
This PR adds a `tox.ini` file facilitating testing in isolated environment and automating common development tasks.

Tox is a common python tool that creates and manages an ensemble python virtual environments for a set of predefined test environments (for instance different versions or python and/or dependencies).
Running the tests in isolation inside virtual environments separate to your development environment can reveal several tricky issues like dependency issues (ex: you added a dependency and forgot to update the `setup.py`), discrepancies behaviors between python versions, missing environment variables...
Basically, running your tests with Tox shields your PR against "It works on my machine" type of issues. Having a `tox.ini` is also a way to document the installation of your package!

Tox makes developers' life easier. For instance, typing 
```bash
tox
```
will create 2 venvs (one for python3.6, one for python3.7), install pybamm and extra dependencies (e.g. flake8) and run `python run-tests.py --quick` in each of them.

From a freshly cloned PyBaMM,
```bash
tox -e dev
```
will create a virtual environment `.tox/dev` with all the necessary software to start working on PyBaMM. It's an alternative to 
```bash
$ pip3 install --user virtualenv
$ virtualenv venv
(venv) $ pip install -e .[dev,docs]
```
Check the ["install from source" page](https://github.com/pybamm-team/PyBaMM/blob/tox/docs/install/install-from-source.rst) in the docs.

Another impact of Tox is the simplification of the CI config (see `.github/workflows/`).

Lastly, `tox.ini` can contain both the `.coveragerc` and `.flake8` configs, reducing the number of config files in the repo.

## Type of change

# Key checklist:

- [] The documentation builds: `$ cd docs` and then `$ make clean; make html`


tlestang:
Thanks for the review @tinosulzer 

By default, `tox` will always run tests for both python versions, resulting in an `InterpreterNotFound` error if the python version is not installed in your machine.
To run tox for only the python version of the current interpreter (`python --version`), use
```bash
tox -e py
```
To run the tests, specifically for 3.7, use `tox -e py37`.

tox.ini
```diff
@@ -0,0 +1,123 @@
...
+passenv = HOME
+whitelist_externals = git
+deps =
+     wget
+     cmake
+commands =
+         python {toxinidir}/scripts/install_KLU_Sundials.py
+	 - git clone https://github.com/pybind/pybind11.git {toxinidir}/pybind11
+
+[testenv:pybamm-dev]
```
tinosulzer, line 33: this should be `dev` instead of `pybamm-dev`

tox.ini
```diff
@@ -0,0 +1,123 @@
...
+[testenv:sundials]
+skip_install = true
+passenv = HOME
+whitelist_externals = git
+deps =
+     wget
+     cmake
+commands =
+         python {toxinidir}/scripts/install_KLU_Sundials.py
+	 - git clone https://github.com/pybind/pybind11.git {toxinidir}/pybind11
```
tinosulzer, line 31: shouldn't this install `scikits.odes` too? Or is it assumed that user will always run sundials then dev? What happens if other way round?

tlestang: Yes, it's assumed that `tox -e sundials` is executed before. It's to avoid going through the sundials build each time one wants to create a virtualenv...
We could proabably merge `tox -e sundials` into `tox -e dev`, but I should check that the sundials script is smart enough to detect a previous installation and not re-download everything! (I don't think it is atm)

tlestang:
@tinosulzer Many thanks for trying it out.
Re the CMake error, it looks like there's a conflict with a previous install. Did you have any CMakeCache.txt from a previous install?
Do you get the error when trying from a clean state? (i.e. freshly cloned repo)

tinosulzer:
Yep works fine now without CMakeCache.txt

tlestang:
@tinosulzer I updated the docs, I'm happy for it to be merged now.

.github/workflows/test_on_push.yml
```diff
@@ -64,51 +61,35 @@ jobs:
...
-    - name: Run unit tests
+    - name: Run unit tests on GNU/Linux and MacOS
       if: matrix.os != 'windows-latest'
+      run: tox -e unit
+
+    - name: Run unit tests on Windows
+      if: matrix.os == 'windows-latest'
       run: |
-        export LD_LIBRARY_PATH=$HOME/.local/lib:scikits.odes/sundials5/lib:$LD_LIBRARY_PATH
+        pip install -e .
```
tinosulzer, line 77: so does tox not work on windows?

CONTRIBUTING.md
```diff
@@ -12,7 +12,8 @@ Before you commit any code, please perform the following checks:
 - [All tests pass](#testing): `$ python run-tests.py --unit`
 - [The documentation builds](#building-the-documentation): `$ cd docs` and then `$ make clean; make html`
 
-You can even run all three at once, using `$ python run-tests.py --quick`.
+You can even run all three at once, using `$ tox`. This will run the above checks in two separate environments
+for python 3.6 and 3.7. To check for only your current python version, use `$ tox -e py`.
```
tinosulzer, line 16: Could also add 3.8 now

docs/install/install-from-source.rst
```diff
@@ -0,0 +1,214 @@
...
+.. code:: bash
+
+	  # in the PyBaMM/ directory
+	  tox -e build-requires
+
+This will download, compile and install the SuiteSparse and SUNDIALS libraries.
+Both libraries are installed in ``~/.local``.
+
+Using Homebrew (recommended for MacOS users)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
tinosulzer, line 73: not clear if this should be done instead of tox or as an alternative on MacOS

tox.ini
tinosulzer: still should be changed to `dev` (?)

docs/install/install-from-source.rst
tlestang: I specified the titled of the section above.

CONTRIBUTING.md
tlestang: This is no longer relevant.

.github/workflows/test_on_push.yml
tlestang: yes it should, but configuring tox to work on both *nix and windows isnt straightforward (we don't want to install sundials and/or scikits.odes on windows). Initially I struggled to write a tox.ini that would deal with all the cases without being very messy and intricate. Eventually I gave up on Windows.

I tried again today and I thing I've got something that works. I had to change the structure of the tox.ini a lot though.

Bottom line is:

GNU/Linux & Mac
```bash
tox -e dev
source .tox/dev/bin/actvate
```

Windows:
```bash
python -m tox -e windows-dev
.tox\dev\Scripts\activate.bat
```

PR #1008, opened 2020-05-20
Issue 992 plot arrays

### LINKED ISSUE #992: plot pybamm arrays
It would be nice to automate the plotting of `pybamm.Array` objects for doing things like the example below. I think just 1D is fine for now. More complicated will usually be of solution entries, so handled via `Quickplot`, but it would be nice to easily plot parameters that are functions. I guess 2D is useful since this depends on T and sto. Thoughts?

Edit: I know calling `.entries` isn't a huge deal, but think this would be easy to implement and avoid needing to import matplotlib 

```
import pybamm
import numpy as np

param = pybamm.ParameterValues(chemistry=pybamm.parameter_sets.Ecker2015)
D_fun = param["Negative electrode diffusivity [m2.s-1]"]
x = pybamm.Array(np.linspace(0,1,100))
D = D_fun(x, 300)
plt.plot(x.entries,D.entries) 
```

rtimms:
# Description
Added basic plots for 1D and 2D arrays (`pybamm.plot` and `pybamm.plot2D`). I've left these as separate methods, but I guess they could be combined and the appropriate plot called depending on what arrays you pass. On the other hand, maybe it is better to be explicit. 

Also added `pybamm.linspace` and `pybamm.meshgrid` to quickly create some common arrays.

Fixes #992 

## Type of change

# Key checklist:

## Further checks:


pybamm/__init__.py
```diff
@@ -223,7 +223,7 @@ def version(formatted=False):
 #
 # other
 #
-from .quick_plot import QuickPlot, dynamic_plot, ax_min, ax_max
+from .quick_plot import QuickPlot, plot, plot2D, dynamic_plot, ax_min, ax_max
```
tinosulzer, line 226: might be worth making a new plotting folder with a few different functions in different files?

tests/unit/test_plot.py
```diff
@@ -0,0 +1,43 @@
...
+            pybamm.plot("bad", x)
+        with self.assertRaisesRegex(TypeError, "y must be 'pybamm.Array'"):
+            pybamm.plot(x, "bad")
+
+    def test_plot2D(self):
+        x = pybamm.Array(np.array([0, 3, 10]))
+        y = pybamm.Array(np.array([6, 16, 78]))
+        X, Y = pybamm.meshgrid(x, y)
+
+        pybamm.plot2D(X, Y, Y, xlabel="x", ylabel="y", title="title", testing=True)
```
tinosulzer, line 24: could `plot2D` automatically do meshgrid(x,y) if they aren't given in meshgrid form?

pybamm/expression_tree/array.py
```diff
@@ -101,3 +101,16 @@ def new_copy(self):
     def _base_evaluate(self, t=None, y=None, y_dot=None, inputs=None):
         """ See :meth:`pybamm.Symbol._base_evaluate()`. """
         return self._entries
+
+
+def linspace(start, stop, num=50, **kwargs):
+    "Creates a linearly spaced array"
```
tinosulzer, line 107: why pick out `num` from the `kwargs`? Also, link to `np.linspace` docs for the kwargs

pybamm/expression_tree/array.py
```diff
@@ -101,3 +101,16 @@ def new_copy(self):
...
+    return pybamm.Array(np.linspace(start, stop, num, **kwargs))
+
+
+def meshgrid(x, y, **kwargs):
+    "Return coordinate matrices as from coordinate vectors"
```
tinosulzer, line 112: link `np.meshgrid` docs for the kwargs

tests/unit/test_plot.py
rtimms: turns out contourf handles this provided they are the correct shape. have added some lines to account for this

pybamm/__init__.py
rtimms: agreed

pybamm/expression_tree/array.py
rtimms: did this so you can do `pybamm.linspace(0,1,100)` without needing to do `num=100` as I think it's a pretty common use case and you can just do `np.linspace(0,1,100)` directly. that didn't work unless I picked out `num` but maybe there is a better/more common solution?

Issue #836, opened 2020-02-18
Add lithium plating model


Scottmar93:
Add the lithium plating model from: 

[Yang, X. G., Leng, Y., Zhang, G., Ge, S., & Wang, C. Y. (2017). Modeling of lithium plating induced aging of lithium-ion batteries: Transition from linear to nonlinear aging. Journal of Power Sources, 360, 28-40.](https://www.sciencedirect.com/science/article/pii/S0378775317307619)

felipe-salinas:
Hi @Scottmar93 and Pybamm team,

Thanks for the program, it is a great contribution! 

I've been trying to include the equations in the referenced paper and faced two problems. First, the model requires a total current j_tot =  j_n (already existing)+j_sei (which has a similar implementation than the one @Scottmar93 formulated in #834, but requires the definition of a new variable c_EC, the concentration of Ethylene Carbonate at the surface of graphite, and an algebraic equation that relates c_EC to j_sei and L_sei) + j_lpl (also similar to #834). This total current j_tot should be equal to the current in the solid phase i_s in the negative electrode. However, I can't find among the sub-models the interaction between the electrode (Ohms law) and interface (Butler-Volmer and associated kinetics), in which such modification should take place. I think that there was an attempt of this kind in the submodel "full_ohm.py" (pybamm/models/submodels/electrode/ohm/full_ohm.py) in the inclusion of the term sum_j (line 67), obtained from the dictionary "reactions" (apparently unused #933). What do you think that It would be the best way to include this modification? 

The second issue is that both Butler-Volmer for j_n and Tafel equations for j_sei and j_lpl should include a term in the overpotential of the type L_sei*R_sei*j_tot, which implies an iterative solution. Has this already been implemented in any of the submodels (ex. actual Butler-Bolmer formulation)?

rtimms:
Hi @felipe-salinas thanks for getting involved! There is currently a PR open (#948) which updates how the total current density is set (to allow for extra reactions etc.), so it might be worth checking there to see if that helps (@tinosulzer not sure how close this is to being finished?). I think after that you should be able to call up `j_tot` (or any of the other currents) more easily as an existing variable where you need it in the lithium plating submodel.

At the moment we haven't implemented having that resistance in the overpotential, but we were discussing it the other day. It requires updating the submodel to make j part of the state vector with a corresponding algebraic equation to solve for.

tinosulzer:
[...]

PR #1009, opened 2020-05-21
Issue 836 add lithium plating

felipe-salinas:
# Description

Adds SEI model based on Yang et. al. (see #836) and reduction in negative electrode porosity. 

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:


examples/scripts/cycling_ageing_yang.py
```diff
@@ -0,0 +1,36 @@
+import pybamm as pb
+
+options = {"sei": "ec reaction limited",
+           "sei film resistance": "distributed",
+           "porosity": "variable porosity"}
```
tinosulzer, line 5: Note that the "sei film resistance" is set automatically by the following lines in `BaseBatteryModel`:
```python3
        # Change the default for SEI film resistance based on which sei option is
        # provided
        extra_options = extra_options or {}
        sei_option = extra_options.get("sei", None)  # return None if option not given
        if sei_option is None:
            default_options["sei film resistance"] = None
        else:
            default_options["sei film resistance"] = "distributed"
```
Something similar could be done for porosity

pybamm/input/parameters/lithium-ion/seis/example/parameters.csv
```diff
@@ -15,5 +15,10 @@ Outer SEI open-circuit potential [V],0.8,,
...
-Initial inner SEI thickness [m], 7.5E-9, 7.5E-9 1/2 of initial thickness in Single paper,
-Initial outer SEI thickness [m], 7.5E-9, 1/2 of initial thickness in Single paper,
\ No newline at end of file
+Initial inner SEI thickness [m], 2.5E-9, 2.5E-9 1/2 of initial thickness in Safari paper,
+Initial outer SEI thickness [m], 2.5E-9, 1/2 of initial thickness in Safari paper,
+EC initial concentration in electrolyte [mol.m-3], 4.541E3, Safari paper,
+EC diffusivity [m2.s-1], 2E-18, adjusted parameter in Yang paper,
+SEI kinetic rate constant [m.s-1], 1e-12, adjusted parameter in Yang paper,
+SEI open-circuit potential [V], 0.4, Safari paper,
+
```
tinosulzer, line 24: thanks for updating these parameters!

pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -658,22 +660,6 @@ def set_voltage_variables(self):
...
-        ]
-        eta_sei_n_av_dim = self.variables[
-            "X-averaged negative electrode sei film overpotential [V]"
-        ]
-        eta_sei_p_av_dim = self.variables[
-            "X-averaged positive electrode sei film overpotential [V]"
-        ]
-        eta_sei_av = eta_sei_n_av + eta_sei_p_av
-        eta_sei_av_dim = eta_sei_n_av_dim + eta_sei_p_av_dim
-
```
tinosulzer, line 676: lines deleted accidentally?

pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -684,8 +670,6 @@ def set_voltage_variables(self):
                 "Measured open circuit voltage [V]": ocv_dim,
                 "X-averaged reaction overpotential": eta_r_av,
                 "X-averaged reaction overpotential [V]": eta_r_av_dim,
-                "X-averaged sei film overpotential": eta_sei_av,
-                "X-averaged sei film overpotential [V]": eta_sei_av_dim,
```
tinosulzer, line 688: lines deleted accidentally?

pybamm/models/full_battery_models/lithium_ion/dfn.py
```diff
@@ -53,7 +53,10 @@ def __init__(self, options=None, name="Doyle-Fuller-Newman model", build=True):
 
     def set_porosity_submodel(self):
 
-        self.submodels["porosity"] = pybamm.porosity.Constant(self.param)
+        if self.options["porosity"] is None:
+            self.submodels["porosity"] = pybamm.porosity.Constant(self.param)
+        elif self.options["porosity"] == "variable porosity":
+            self.submodels["porosity"] = pybamm.porosity.Full(self.param)
```
tinosulzer, line 59: equivalent changes need to be made in `SPM` and `SPMe` models

pybamm/models/submodels/interface/sei/ec_reaction_limited.py
```diff
@@ -0,0 +1,149 @@
...
+            domain=self.domain.lower() + " electrode",
+            auxiliary_domains={"secondary": "current collector"},
+        )
+        L_sei_av = pybamm.x_average(L_sei)
+        j_sei_av = pybamm.x_average(j_sei)
+        c_ec_av = pybamm.x_average(c_ec)
+
+        L_scale = pybamm.sei_parameters.L_sei_0_dim
+        # in this model the scale is identical to the intercalation current
+        j_scale = pybamm.sei_parameters.j_scale
```
tinosulzer, line 48: `j_scale` is different in the negaive and positive electrodes so you need to be a bit careful here (though at the moment an error will be raised if the positive electrode sei model isn't `NoSEI`, so it might be ok)

pybamm/models/submodels/interface/sei/ec_reaction_limited.py
```diff
@@ -0,0 +1,149 @@
...
+                          + " electrode total interfacial current density"]
+        C_sei_ec = pybamm.sei_parameters.C_sei_ec
+        C_ec = pybamm.sei_parameters.C_ec
+        R_sei = pybamm.sei_parameters.R_sei
+
+        # need to revise for thermal case
+
+        self.algebraic = {j_sei: j_sei + C_sei_ec * c_ec * pybamm.exp(- 0.5 * (
+            phi_s_n - phi_e_n - j * L_sei * R_sei)),
+            c_ec: c_ec - pybamm.Scalar(1) - j_sei * L_sei * C_ec}
```
tinosulzer, line 135: it looks like `j_sei` and `c_ec` can be defined explicitly in terms of other variables in `get_coupled_variables`, instead of being made `Variable` objects themselves. While this will give the same answer, it's best to try to reduce the number of equations, especially algebraic

pybamm/models/submodels/porosity/full_reaction_driven_porosity.py
```diff
@@ -30,15 +30,16 @@ def get_fundamental_variables(self):
...
-
-        deps_n_dt = -self.param.beta_surf_n * j_n
+        deps_n_dt = C_sei_eps * j_sei
         deps_s_dt = pybamm.FullBroadcast(
             0, "separator", auxiliary_domains={"secondary": "current collector"}
         )
-        deps_p_dt = -self.param.beta_surf_p * j_p
+        # for the moment is not needed but could be used for positive SEI
+        deps_p_dt = pybamm.FullBroadcast(0, "positive electrode",
+                                         "current collector")
```
tinosulzer, line 42: The `beta_surf_n * j_n` and `beta_surf_p * j_p` terms should be kept as they are used in the lead-acid models. You can do this by setting `beta_surf_n` and `beta_surf_p` to zero in `standard_parameters_lithium_ion`

pybamm/parameters/sei_parameters.py
```diff
@@ -51,25 +59,26 @@
...
 
-C_sei_reaction_n = (j_scale_n / m_sei_dimensional) * pybamm.exp(
-    -(F * U_n_ref / (2 * R * T_ref))
-)
-C_sei_reaction_p = (j_scale_p / m_sei_dimensional) * pybamm.exp(
-    -(F * U_n_ref / (2 * R * T_ref))
-)
+C_sei_reaction = (
+    F * L_sei_0_dim / (m_sei_dimensional * tau_discharge * V_bar_inner_dimensional)
+) * pybamm.exp(-(F * U_n_ref / (2 * R * T_ref)))
```
tinosulzer, line 67: please fix the changes overwritten in this file

examples/scripts/cycling_ageing_yang.py
```diff
@@ -0,0 +1,36 @@
...
+        "Electrolyte concentration [mol.m-3]",
+        "Negative electrode surface potential difference",
+        "Total negative electrode sei thickness",
+        "Negative electrode porosity",
+        "Negative electrode sei interfacial current density",
+        "Negative electrode interfacial current density",
+        "Sum of negative electrode interfacial current densities",
+        #"Negative electrode EC surface concentration",
+    ]
+)
```
tinosulzer, line 36: nice example!

examples/scripts/cycling_ageing_yang.py
felipe-salinas: Thanks for the tip: I will erase the redundant setting for sei film resistance in this example. In the case of porosity, It would be a nice feature, but the variable porosity is not a default option for SEI models (older publications like Ramadass et. al. 2003 don't include it), so I think that is better to leave it as a "visible" option.

pybamm/models/full_battery_models/base_battery_model.py
felipe-salinas: Sorry, I grabbed an older version of develop :/

pybamm/models/full_battery_models/base_battery_model.py
felipe-salinas: I think it is correct now.

pybamm/models/full_battery_models/lithium_ion/dfn.py
felipe-salinas: Done.

pybamm/models/submodels/interface/sei/ec_reaction_limited.py
felipe-salinas: Modified in the file to explicitly recover the negative scale.

pybamm/models/submodels/interface/sei/ec_reaction_limited.py
felipe-salinas: c_ec and j_sei are present in both equations, and j_sei depends on itself because it is included in j (the total current). I think that makes it necessary to implement them with algebraic equations. It is however possible to replace c_ec in the equation of j_sei to just make one larger equation, but I think that makes it difficult to understand if the equations in Yang et. al. where correctly implemented. I would  do the latter after it is clear that all scales are multiplied correctly, and that a reference experiment can be reproduced with the added models.

pybamm/models/submodels/porosity/full_reaction_driven_porosity.py
felipe-salinas: I created another class because they have different purposes.

pybamm/parameters/sei_parameters.py
felipe-salinas: Done.

examples/scripts/cycling_ageing_yang.py
felipe-salinas: :D

pybamm/models/full_battery_models/lithium_ion/spm.py
```diff
@@ -53,7 +53,10 @@ def __init__(self, options=None, name="Single Particle Model", build=True):
 
     def set_porosity_submodel(self):
 
-        self.submodels["porosity"] = pybamm.porosity.Constant(self.param)
+        if self.options["porosity"] is None:
+            self.submodels["porosity"] = pybamm.porosity.Constant(self.param)
+        elif self.options["porosity"] == "variable porosity":
+            self.submodels["porosity"] = pybamm.porosity.SeiFull(self.param)
```
tinosulzer, line 59: For the SPM, given that `j_sei` is uniform in x, using the "SeiFull" porosity submodel creates unnecessarily many states for the porosity (i.e. when discretised, there will be `n` ODEs where `n` is the number of grid points, instead of just one ODE for each domain). You could use the "leading order reaction driven porosity" model instead

pybamm/models/full_battery_models/lithium_ion/spme.py
```diff
@@ -56,7 +56,10 @@ def __init__(
 
     def set_porosity_submodel(self):
 
-        self.submodels["porosity"] = pybamm.porosity.Constant(self.param)
+        if self.options["porosity"] is None:
+            self.submodels["porosity"] = pybamm.porosity.Constant(self.param)
+        elif self.options["porosity"] == "variable porosity":
+            self.submodels["porosity"] = pybamm.porosity.SeiFull(self.param)
```
tinosulzer, line 62: Same comment as for the SPM

pybamm/models/submodels/porosity/sei_reaction_driven_porosity.py
```diff
@@ -0,0 +1,61 @@
...
+        deps_p_dt = pybamm.FullBroadcast(0, "positive electrode",
+                                         "current collector")
+
+        variables.update(
+            self._get_standard_porosity_change_variables(
+                deps_n_dt, deps_s_dt, deps_p_dt
+            )
+        )
+
+        return variables
```
tinosulzer, line 50: Still not convinced this warrants a whole separate submodel. All you would need to do to reuse the full submodel is to add these lines to `standard_parameters_lithium_ion.py`:
```python3
beta_surf_n = pybamm.Scalar(0)
beta_surf_p = pybamm.Scalar(0)
```
By the way, just so you are aware, `deps_dt` comes into the `dce_dt` equation (see e.g. `full_diffusion.py`) so a changing porosity will affect the electrolyte concentration (make the electrolyte more concentrated) - again, this comes from the lead-acid models

tinosulzer:
Now #1073

Issue #571, opened 2019-08-01
processed spatial variables


tobykirk:
We would like to process a SpatialVariable (into a "ProcessedSpatialVariable") to extract the limits and construct a np.array of points between them using linspace (or otherwise). This would allow easy plotting over any domain of choice.

martinjrobins:
might be able to make `ProcessedVariable` automatically detect if the variable is a spatial one and use a default np.array of points the same as the mesh on that domain?

rtimms:
you can now do this (see PR #1020) e.g.
```
import pybamm
import numpy as np

model = pybamm.lithium_ion.SPM()

sim = pybamm.Simulation(model)
solution = sim.solve()
r_n = solution["r_n [m]"]
r_n(r=np.linspace(0, 1, 10))
```
to get 10 linearly spaced points in the negative particle

PR #1020, opened 2020-05-26
Issue 1006 process space var

### LINKED ISSUE #1006: process space-only variables
At the moment `ProcessedVariable` only works if `len(t)>1` (except for the special case of  2D finite elements, which can handle "fixed t" cases). It would be good to set it up so that any variables that are functions of space only can be processed.

rtimms:
# Description
Extend `processed_variable` to work in the case when `len(t_sol)=1` and also makes it so you no longer need to pass `t` to the call in the cases where the variable is constant (doesn't depend on `t` or `y`) or when `len(solution.t)=1` (e.g. algebraic models)

Fixes #1006 and #571 

EDIT: Note that for variables discretised using finite elements (in y and z) the result returned by `ProcessedVariable` is transposed. This is consistent with other features, e.g. you can now do `plt.pcolormesh(x,y,z)` where `z` is some variable discretised with finite elements, whereas before you had to do `plt.pcolormesh(x,y,np.transpose(z))` 

## Type of change

# Key checklist:

## Further checks:


pybamm/solvers/processed_variable.py
```diff
@@ -297,40 +275,43 @@ def initialise_2D(self):
...
+                    second_dim = second_dim[:, 0]
+                    return interpolant(second_dim, first_dim)
+                elif isinstance(first_dim, np.ndarray):
+                    first_dim = first_dim[:, 0]
+                    return interpolant(second_dim, first_dim)[:, 0]
+                elif isinstance(second_dim, np.ndarray):
+                    second_dim = second_dim[:, 0]
+                    return interpolant(second_dim, first_dim)
+                else:
+                    return interpolant(second_dim, first_dim)[0]
```
tinosulzer, line 304: is there no way around the duplication of all these lines?

pybamm/solvers/processed_variable.py
```diff
@@ -367,23 +348,61 @@ def initialise_2D_scikit_fem(self):
...
+                        second_dim = second_dim[:, 0]
+                        return interpolant(first_dim, second_dim)[:, :, np.newaxis]
+                    else:
+                        first_dim = first_dim[:, 0]
+                        return interpolant(first_dim, second_dim)[:, np.newaxis]
+                elif isinstance(second_dim, np.ndarray):
+                    second_dim = second_dim[:, 0]
+                    return interpolant(first_dim, second_dim)[:, :, np.newaxis]
+                else:
+                    return interpolant(first_dim, second_dim)
```
tinosulzer, line 371: duplicated here

pybamm/solvers/processed_variable.py
rtimms: yep, have made a function to take the interpolant and inputs and returns the result of the correct shape

PR #1022, opened 2020-05-26
Issue 718 cooling area

### LINKED ISSUE #718: alternative geometries for lumped thermal models
At the moment the lumped thermal models assume a pouch cell geometry. We could get users to specify the cell surface area and volume in the parameters file so that the correct ratio can be applied in the Newton cooling. This would be useful if you wanted to do a 1D electrochemical model, with a lumped thermal model for a specific cell geometry.  

The Nplus1D thermal models also assume a pouch cell model and will probably remain this way for now as they account for cooling from the face of the current collector (which would be would up in a jelly roll and therefore not open to the air for cooling). It should probably be clearer that these models are for pouch cells only.

ferranbrosa:
# Description

The lumped thermal model now takes an option to choose between arbitrary geometry (default) or a pouch geometry. The arbitrary geometry takes a total heat transfer coefficient, surface cooling area and cell volume. The pouch geometry calculates it from the different parts of the pouch (including tabs). 

Fixes #718

## Type of change

# Key checklist:

## Further checks:


ferranbrosa:
For some reason all the commits from rebase appear as changes on the branch. Not sure if there is a way to fix that, otherwise there are a ton of files in the review that I have not changed.

tinosulzer:
You need to change the merge branch from "master" to "develop"

rtimms:
actually, could you make it raise an option error if people try to use an arbitrary geometry with the models that assume a pouch?

ferranbrosa:
> actually, could you make it raise an option error if people try to use an arbitrary geometry with the models that assume a pouch?

I thought about doing so, but I thought maybe people will want to use the Kokam parameters (for example) to model another set up so I thought it was better to not throw the error.

rtimms:
yeah I think we should allow using whatever parameters you like. I was thinking of the case where someone uses `dimensionality = 2` with an `x-lumped` model, which is specific to a pouch cell

ferranbrosa:
Oh yes, in that case it makes sense. Is it only for dimensionality = 2 x-lumped model? Do we want to throw an error or just a warning?

rtimms:
the `x-lumped` models assume a pouch cell for conditionality 1 and 2. I think it would be good to raise an option error. If people really want to to do it they can bypass the option error by manually changing submodels

ferranbrosa:
Actually, `x-lumped` models of higher dimension already assume pouch (I have only implemented arbitrary geometry for `lumped` models). Dimensionality 0 will fall back to the `lumped` model I think so then the arbitrary geometry is fine.

rtimms:
Yeah it is probably fine as is. I was just worried that people might choose the options `{"dimensionality": 2, "thermal": "x-lumped", "geometry": "arbitrary"}` which at the moment silently works ok, but people might not realise that choosing "arbitrary" has no effect. Maybe there could be a setter for geometry that picks the default based on some other options (if it isn't provided). But perhaps I am unnecessarily worried and that is too complicated!

PR #1028, opened 2020-05-28
Issue 920 dimensional interp

### LINKED ISSUE #920: Make t and x dimensional in interpolation
For consistency of the user-facing interface, values supplied to a parameter when interpolating should always be dimensional.

rtimms:
# Description
Calls to `ProcessedVariable` objects are made using *dimensional* time and space

Fixes #920 

## Type of change

# Key checklist:

## Further checks:


examples/scripts/SPMe_SOC.py
```diff
@@ -85,15 +85,16 @@
         xnsurf = sol["X-averaged negative particle surface concentration"]
         time = sol["Time [h]"]
         # Coulomb counting
-        time_hours = time(sol.t)
+        time_secs = sol.t * model.timescale_eval
```
tinosulzer, line 88: maybe do `sol["Time [s]"].entries` to avoid having to call `model.timescale_eval`?

pybamm/solvers/processed_variable.py
```diff
@@ -54,7 +57,27 @@ def __init__(self, base_variable, solution, known_evals=None):
...
-
+        self.warn = warn
+
+        # Set timescale
+        try:
+            self.timescale = solution.model.timescale_eval
+        except AttributeError:
+            if self.warn:
+                pybamm.logger.warning("No timescale provided. Using default of 1 [s]")
+            self.timescale = 1
```
tinosulzer, line 68: A model should always have a timescale, since the base model has the default timescale of 1s

pybamm/solvers/processed_variable.py
```diff
@@ -54,7 +57,27 @@ def __init__(self, base_variable, solution, known_evals=None):
...
+
+        # Store spatial variables to get scales
+        self.spatial_vars = {}
+        if solution.model:
+            for var in ["x", "y", "z", "r_n", "r_p"]:
+                if var and var + " [m]" in solution.model.variables:
```
tinosulzer, line 74: this doesn't check that both `var` and `var + " [m]"` are in `solution.model.variables`. It just checks that `var` is not `None` or `False`, and that `var + " [m]"` is in `solution.model.variables`. You have to check both separately.

pybamm/solvers/processed_variable.py
```diff
@@ -386,9 +437,9 @@ def __call__(self, t=None, x=None, r=None, y=None, z=None, warn=True):
         # an error
         if t is None:
             if len(self.t_sol) == 1:
-                t = self.t_sol
+                t = self.t_sol * self.timescale
             elif self.base_variable.is_constant():
-                t = self.t_sol[0]
+                t = self.t_sol[0] * self.timescale
```
tinosulzer, line 442: Am wondering whether `t_sol` should be stored in seconds. I guess you are either consistent with `sol.t`, or consistent with `first_dim_pts`, but you can't have both

pybamm/solvers/processed_variable.py
```diff
@@ -425,9 +476,30 @@ def call_2D(self, t, x, r, y, z):
...
+                self.spatial_vars[name + " [m]"] / self.spatial_vars[name]
+            ).evaluate()[-1]
+        else:
+            if self.warn:
+                pybamm.logger.warning(
+                    "No scale set for spatial variable {}. "
+                    "Using default of 1 [m]".format(name)
+                )
+            scale = 1
+        return scale
```
tinosulzer, line 501: maybe models should have a `r_scale`, `x_scale`, etc in the same way that they have a timescale. Probably a separate issue though. It would simplify processed variables and quickplot a lot.

tests/unit/test_solvers/test_processed_variable.py
```diff
@@ -43,10 +47,14 @@ def test_processed_variable_1D(self):
         t_sol = np.linspace(0, 1)
         y_sol = np.ones_like(x_sol)[:, np.newaxis] * np.linspace(0, 5)
 
-        processed_var = pybamm.ProcessedVariable(var_sol, pybamm.Solution(t_sol, y_sol))
+        processed_var = pybamm.ProcessedVariable(
+            var_sol, pybamm.Solution(t_sol, y_sol), warn=False
+        )
```
tinosulzer, line 52: Oh, I see, solution doesn't have a model here for timescale. Maybe `Solution.model` could be initialized to `BaseModel`?

examples/scripts/SPMe_SOC.py
rtimms: agreed that  the recommended way should be `sol["Time [s]"].entries`. I've made this consistent everywhere (I think)

pybamm/solvers/processed_variable.py
rtimms: thanks!

pybamm/solvers/processed_variable.py
rtimms: Yeah I went back and forth on this. I've added an attribute `t_pts` which is dimensional time. That way things ending `_sol` are the dimensionless values and things ending `_pts` are dimensional (and are the values used in the interpolant)

pybamm/solvers/processed_variable.py
rtimms: Yeah this would definitely be the neatest solution. I've made a new issue for this.

tests/unit/test_solvers/test_processed_variable.py
rtimms: yep, this worked thanks!

pybamm/solvers/processed_variable.py
rtimms: there is a bit of an issue with the shape when the timescale depends on the inputs. In `base_solver` the timescale is evaluated with the initial inputs. if you then try to do `model.timescale.evaluate(inputs=solution.inputs)` in `ProcessedVariable` then `inputs` is now of length `t`. For now I've done a try/except with using `timescale_eval` and then evaluating if that isn't found. Another option would be to store the initial values of the inputs in the solution and use that (since the timescale *should* be the same for all times).

tinosulzer: I think the safest option is to not allow the timescale to depend on inputs. i.e. do `model.timescale.evaluate()` instead of `model.timescale.evaluate(inputs=solution.inputs)` (and use a try except for a clear error message). Allowing the timescale to change in e.g. parameter fitting runs is probably asking for trouble

pybamm/solvers/processed_variable.py
tinosulzer: Great, that makes sense!

pybamm/solvers/processed_variable.py
rtimms: yeah I agree. I've added a check for this in `base_solver`

PR #1038, opened 2020-06-03
Issue 1031 jax

### LINKED ISSUE #1031: investigate use of JAX to compile and autodiff expression trees
this issue is to track investigation of compiling pybamm expression trees using [JAX](https://jax.readthedocs.io/en/latest/index.html)

martinjrobins:
# Description

Adds support for evaluating expression trees using JAX, and adds a new solver `JaxSolver` with two methods: A RK4(5) method using `jax.experimental.odeint`, and a custom BDF method written using jax. This solver is currently only useful for ode models with no termination events, and since JAX does not support sparse matrices they are all converted to dense arrays in the model. The latter limitation should hopefully be addressed over time by JAX (see https://github.com/google/jax/issues/765). 

This solver would be useful for running on a GPU/TPU, and, for very small state vectors (due to the restriction to dense arrays), is also very quick on a CPU. The next step after this PR would be to implement the adjoint sensitivities for the new BDF solver (already there for the RK45) and expose the raw jax solvers so a user can calculate sensitivities for parameter estimation.

UPDATE: just saw that jax is not supported under windows :(, so this would be linux/mac only....

Fixes #1031

## Type of change

# Key checklist:

## Further checks:


pybamm/solvers/jax_bdf_solver.py
```diff
@@ -0,0 +1,719 @@
...
+    t_eval: ndarray
+        time points to evaluate the solution, has shape (m,)
+    jac: (optional) callable
+        function with signature (t, y, in),returns the jacobian matrix of fun as an
+        ndarray with shape (n,n)
+    inputs: (optional) dict
+        dict mapping input parameter names to values
+    rtol: (optional) float
+        relative tolerance for the solver
+    rtol: (optional) float
```
tinosulzer, line 36: typo, should be `atol`

pybamm/solvers/jax_bdf_solver.py
```diff
@@ -0,0 +1,719 @@
...
+        function with signature (t, y), where t is a scalar time and y is a ndarray with
+        shape (n,), returns the rhs of the system of ODE equations as an nd array with
+        shape (n,)
+    jac: callable
+        function with signature (t, y), where t is a scalar time and y is a ndarray with
+        shape (n,), returns the jacobian matrix of fun as an ndarray with shape (n,n)
+    t0: float
+        initial time
+    y0: ndarray
+        initial state vector with shape (n,)
```
tinosulzer, line 148: missing rtol and atol docs

pybamm/solvers/jax_bdf_solver.py
```diff
@@ -0,0 +1,719 @@
...
+    f0 = fun(t0, y0)
+    state['atol'] = atol
+    state['rtol'] = rtol
+    order = 1
+    state['order'] = order
+    state['h'] = _select_initial_step(state, fun, t0, y0, f0, h0)
+    EPS = np.finfo(y0.dtype).eps
+    state['newton_tol'] = np.max((10 * EPS / rtol, np.min((0.03, rtol ** 0.5))))
+    state['n_equal_steps'] = 0
+    #state['jac'] = jac
```
tinosulzer, line 178: should this be removed?

pybamm/solvers/jax_solver.py
```diff
@@ -0,0 +1,189 @@
...
+
+        """
+        if model.convert_to_format != "jax":
+            raise RuntimeError("Model must be converted to JAX to use this solver"
+                               " (i.e. `model.convert_to_format = 'jax')")
+
+        if model.terminate_events_eval:
+            raise RuntimeError("Terminate events not supported for this solver."
+                               " Model has the following events:"
+                               " {}".format(model.events))
```
tinosulzer, line 106: Can you add some info to explain how to remove events from the model (`model.events = []`)? And perhaps suggest solving first with a different solver to figure out what the end time should be, then using this one for multiple fast iterations

pybamm/solvers/jax_bdf_solver.py
martinjrobins: forgot to update that docstring, done now

pybamm/solvers/jax_solver.py
martinjrobins: Yup, good idea. I've added a comment to this effect

martinjrobins:
Thanks Tino. Yea, the BDF code was a pain to write as you can't use any flow control or classes, it does make it difficult to read!

I'm not sure what changes need to be made to solve a DAE as I've not looked into that yet. I've not done the solution of daes before, but I would suspect you would have to code up a separate algorithm.

You can't generate C code from Jax as it never uses C. JAX traces your python code and produces its own expression tree. It uses that to build a computational graph using XLA. The input to XLA is an intermediate representation language called HLO IR, so rather than C, the language that you could emit would be HLO, this could then be compiled and run on any machine using XLA. I'm not sure exactly how to get the HLO using JAX, but in theory it should be a matter of taking any jax function (e.g. the jax_bdf_integrate function in this case), compiling it with JAX and then telling it to give you the generated HLO.

tinosulzer:
Thanks for the info. It looks like it's possible to adapt BDF to solve ODEs: https://www.cs.usask.ca/~spiteri/M314/notes/AP/chap10.pdf

martinjrobins:
Nice, thanks for the link. I'll have a look.

Cheers,
Martin

PR #1041, opened 2020-06-08
#1037 fix timescale

### LINKED ISSUE #1037: slider plots in notebooks
variables that depend on space aren't updating with the slider in notebooks, but work ok if you just run the scripts

rtimms:
Fixes issue in notebooks where slider didn't update. Problem was the timescale in `plot` (the sliders update ok in notebook when the timescale is seconds). It worked in scripts as that called `slider_update` rather than `plot`

Fixes #1037

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ New SEI models, simplification of submodel structure, as well as optimisations a
 
 ## Bug fixes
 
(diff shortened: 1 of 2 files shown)


rtimms:
it updates pretty quickly if you just click a point on the slider rather than slide. we use `continuous_update=False`, so it shouldn't update until you release the mouse, but it seems that it is...

PR #1055, opened 2020-06-14
#1051 fix 2D plot bug

### LINKED ISSUE #1051: plotting of 2D current collector variables
Doing `sim.plot` with a 2D current collector variable doesn't give the correct plot, but also doesn't raise an error. We should either fix it or raise an error.

e.g. doing 
```
options = {
    "current collector": "potential pair",
    "dimensionality": 2,  # 2D current collectors
    "thermal": "x-lumped",
}
model = pybamm.lithium_ion.SPM(options)
solution = sim.solve()

sim.plot(["Negative current collector potential [V]"])
```

gives the wrong plot. But

```
phi_s_cn = solution["Negative current collector potential [V]"]

L_z = param.evaluate(pybamm.standard_parameters_lithium_ion.L_z)
l_y = phi_s_cp.y_sol[-1]  # dimensionless width
l_z = phi_s_cp.z_sol[-1]  # dimensionless height
y_plot = np.linspace(0, l_y, 21) * L_z
z_plot = np.linspace(0, l_z, 21) * L_z
t_plot = 1000

phi_s_cn_plot = plt.pcolormesh(
    y_plot, z_plot, phi_s_cn(y=y_plot, z=z_plot, t=t_plot), shading="gouraud"
)
```

gives the correct result. The result from `sim.plot` is transposed, so we just need to catch/fix this in quickplot.

rtimms:
# Description

Fixes a bug where 2D variables that depend on y and z are transposed in `QuickPlot`. 

This was introduced in PR #1020 where the way 2D (y,z) variables are processed was changed, and I missed updating in quickplot as the scripts I had been using for 2+1D models all did the plotting manually... sorry!

Fixes #1051 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+# Unreleased
+
+## Features
+
+## Optimizations
+
+## Bug fixes
+
+-   Fix a bug where variables that depend on y and z were transposed in `QuickPlot` ([]())
+
+## Breaking changes
+
 # [v0.2.2](https://github.com/pybamm-team/PyBaMM/tree/v0.2.2) - 2020-06-01
 
 New SEI models, simplification of submodel structure, as well as optimisations and general bug fixes.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ## Bug fixes
 
--   Fix a bug where variables that depend on y and z were transposed in `QuickPlot` ([]())
+-   Fix a bug where variables that depend on y and z were transposed in `QuickPlot` ([#1055](https://github.com/pybamm-team/PyBaMM/pull/1055))
 
 ## Breaking changes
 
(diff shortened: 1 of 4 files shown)


rtimms:
the percentage coverage on the difference is too low. I can add a test to catch it

ferranbrosa:
But quickplot is already tested, right? If so, then it is fine by me :)

tinosulzer:
Yeah please add more tests for quickplot as it's easy to break things without realising atm

rtimms:
I've added some more tests to ensure the variables/plot data in quick plot get updated properly for different time units (and therefore tests shape, transposed vs not, etc.). For 1D plots you can get this info from the `Lines2D` object that gets stored in `self.plots[key]`. The 2D plots (contourf) create a  `QuadContourSet` object which doesn't store the z data as an attribute, so I've just stored the z_data directly in `self.plots[key][0][1]` and the  `QuadContourSet` object  in `self.plots[key][0][0]` (since you cant plot 2D vars with multiple solutions, you never actually use that final list, so we can just use it to store the object then the data -- if I have understood what the list contains correctly)

PR #1059, opened 2020-06-16
Issue 1015 algebraic solver

### LINKED ISSUE #1015: Adding constraints to algebraic solver
It seems the algebraic solver sometimes fails when a state goes outside an obvious bound (e.g. concentration going negative) which makes the algebraic equations return nan. Opening this issue to brainstorm some solutions, so suggestions welcome:

1. Simply replace `nan` with a very large number. Will try this first and hope it works as this would be the simplest solution. The other two solutions require manually specifying some bounds in the models (e.g. using the existing event functionality)
2. Impose hard constraints, but neither `scipy.optimize.root` nor casadi have a very clean way of providing constraints
3. Impose soft constraints: add a function of `c` with singularities as c-> 0 and c -> 1 to the function being optimized

N.B. the use case here is for fuel cell models, solving directly for the steady state.

tinosulzer:
# Description

I couldn't really get this working but thought I would merge what I have so far as it might be useful for another effort down the line (will require a better example)

- Added some new kinds of algebraic solvers (posing a minimization problem instead of root-finding), which can be constrained
- Added `bounds` attribute to the `Variable` object, to allow these constraints

The main problem is that the solvers tend to find local minima instead of roots.

Fixes #1015 

## Type of change

# Key checklist:

## Further checks:


tinosulzer:
It all works fine, yes, but I haven't yet found a case where rootfinding fails and minimization works instead. At least the bounds functionality is implemented, in case someone thinks of a hacky way to implement bounds for the rootfinder (e.g. artificially extending the function outside of the bounds)

PR #1090, opened 2020-06-30
Issue 1011 model new copy

### LINKED ISSUE #1011: model.new_copy() does not keep modified submodels
**Describe the bug**
When a standard model is defined (e.g. SPMe) but one of the submodels changed then `model.new_copy()` creates a copy of the standard model as it runs `new_model = self.__class__(options)`. An example of this is when trying to compare two different submodels with an experiment.

Not sure how easy it would be to fix it through `model.new_copy()`. In any case, I think it would be interesting to be able to change any submodel through model options. To avoid people needing to change how the model is defined, we could extend the model class so you can pass the option `{"electrolyte": pybamm.MyNewElectrolyteSubmodel()}`. Later on, if the model consolidates, we can name it and add it as a standard option `{"electrolyte": "my new electrolyte submodel"}`.

**To Reproduce**
1. Create two instances of a model (e.g. SPMe) and change a submodel for one of them.
2. Run a simulation with experiments 
3. Plot solutions

tinosulzer:
# Description

Reformat `model.new_copy()` to maintain submodels. Also removes some methods from `Simulation` to keep things simpler.

Fixes #1011 

## Type of change

# Key checklist:

## Further checks:


pybamm/models/submodels/external_circuit/experiment_events.py
```diff
@@ -0,0 +1,35 @@
...
+    def set_events(self, variables):
+        # add current and voltage events to the model
+        # current events both negative and positive to catch specification
+        n_cells = pybamm.electrical_parameters.n_cells
+        self.events.extend(
+            [
+                pybamm.Event(
+                    "Current cut-off (positive) [A] [experiment]",
+                    variables["Current [A]"]
+                    - abs(pybamm.InputParameter("Current cut-off [A]")),
```
ferranbrosa, line 22: Does the current cut-off need to be added to the parameter values or does it get a default value if unspecified?

tinosulzer: It's an input parameter so needs to be provided by the user when solving. But the experiment class does this automatically.

PR #1099, opened 2020-07-06
#1084 allow kwargs to be passed to sim.plot

### LINKED ISSUE #1084: pass keyword args to sim.plot()
It would be handy to be able to pass keyword args to `sim.plot()` (e.g. `time_unit`) in the same way you can when using `pybamm.dynamic_plot`

rtimms:
# Description
Allows users to pass keyword arguments to `Simulation.plot()`. For a full list of possible keyword arguments see `QuickPlot`.

Also renames `quick_plot_vars`  to `output_variables` in `Simulation` to be consistent with `QuickPlot`

Fixes #1084 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/simulation.py
+++ b/pybamm/simulation.py
@@ -64,7 +64,7 @@ class Simulation:
         domain (e.g. pybamm.FiniteVolume)
     solver: :class:`pybamm.BaseSolver` (optional)
         The solver to use to solve the model.
-    quick_plot_vars: list (optional)
+    output_variables: list (optional)
         A list of variables to plot automatically
     C_rate: float (optional)
         The C_rate at which you would like to run a constant current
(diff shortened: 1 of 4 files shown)


pybamm/simulation.py
```diff
@@ -531,28 +531,32 @@ def get_variable_array(self, *variables):
...
         ----------
-        quick_plot_vars: list, optional
+        output_variables: list, optional
             A list of the variables to plot.
-        testing, bool, optional
-            If False the plot will not be displayed
+        **kwargs
+            Additional keyword arguments passed to
+            :meth:`pybamm.QuickPlot.dynamic_plot`.
+            For a list of all possible keyword arguments see :class:`pybamm.QuickPlot`.
```
tinosulzer, line 547: this should still accept `quick_plot_vars` and raise an error if this is not None, saying to use `output_variables` instead

PR #1107, opened 2020-07-13
Issue 1104 mass matrix for bdf

### LINKED ISSUE #1104: modify JAX bdf solver to allow solution of semi-explicit DAEs
**Summary**
 Matlab allows for ode15s to solve semi-explicit DAEs by solving the equivalent ODE using a singular mass matrix. Need to modify the Jax BDF solver to do the same

 **Motivation**
So JAX BDF solver can solve semi-explicit DAEs (i.e. all DAE models in Pybamm). Currently limited to ODEs

**References**

[1] Shampine, L. F. and M. W. Reichelt, “The MATLAB ODE Suite,” SIAM Journal on Scientific Computing, Vol. 18, 1997, pp. 1–22.

See also comments by @tinosulzer in #1038

martinjrobins:
# Description

- adds mass matrix support for jax bdf solver, so it can solve semi-explicit dae models
- also adds sensitivity support for ODE and semi-explicit DAE models using the adjoint method
- limitations: currently limited to support for index 1 semi-explicit dae models, but this includes spm, spme and dfn models (that I've tested)
- while doing this, the bdf solver has been substantially refactored to remove lax.cond operators (which are slow), which also has the benefit of making the code more readable

Fixes #1104 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/solvers/jax_solver.py
+++ b/pybamm/solvers/jax_solver.py
@@ -122,7 +122,9 @@ class JaxSolver(pybamm.BaseSolver):
         y0 = jnp.array(model.y0).reshape(-1)
         mass = None
         if self.method == 'BDF':
-            mass = model.mass_matrix.entries.toarray()
+            mass = model.mass_matrix.entries.diagonal()
+            if onp.count_nonzero(mass) != model.mass_matrix.entries.nnz:
+                raise RuntimeError("Solver only supports a diagonal mass matrix")
 
         def rhs_ode(y, t, inputs):
             return model.rhs_eval(t, y, inputs),
@@ -146,7 +148,7 @@ class JaxSolver(pybamm.BaseSolver):
             return jnp.transpose(y)
 
         def solve_model_bdf(inputs):
-            y, stepper = pybamm.jax_bdf_integrate(
+            y = pybamm.jax_bdf_integrate(
                 rhs_dae,
                 y0,
                 t_eval,
--- a/pybamm/solvers/jax_solver.py
+++ b/pybamm/solvers/jax_solver.py
@@ -146,7 +146,7 @@ class JaxSolver(pybamm.BaseSolver):
             return jnp.transpose(y)
 
         def solve_model_bdf(inputs):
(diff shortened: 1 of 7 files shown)


pybamm/solvers/jax_solver.py
```diff
@@ -109,35 +118,55 @@ def _create_solve(self, model, t_eval):
...
+            #sstring = ''
+            #sstring += 'JAX {} solver - stats\n'.format(self.method)
+            #sstring += '\tNumber of steps: {}\n'.format(stepper.n_steps)
+            #sstring += '\tnumber of function evaluations: {}\n'.format(
+            #    stepper.n_function_evals)
+            #sstring += '\tnumber of jacobian evaluations: {}\n'.format(
+            #    stepper.n_jacobian_evals)
+            #sstring += '\tnumber of LU decompositions: {}\n'.format(
+            #    stepper.n_lu_decompositions)
+            #pybamm.logger.info(sstring)
```
tinosulzer, line 168: should these be removed?

martinjrobins: yup

martinjrobins:
yes, we should iterate on the API, I just put in something temporary for now. The solver can return a pure solve function that takes in inputs and returns a solution ndarray. This can then be differentiated using `jax.grad` or `jax.jacrev` (not `jax.jaxfwd`). Ideally you would take the solve function, and wrap it in some sort of scalar error measure, and then `jax.grad` that (see sensitivity tests in `test_jax_bdf_solver.py` and `test_jax.solver.py`

I've added in the jax solver to`compare-dae-solvers.py`, but be aware that it takes a long time. The first time you call the dfn solver jax needs to trace and compile the entire algorithm, along with the dfn equations. This takes 11 minutes for the dfn solver (its significantly quicker for simpler models such as spm or spme). Subsequent calls to the solver take 4 seconds

tinosulzer:
Ah ok, yeah 11 minutes to run an example isn't great, perhaps shouldn't include it in that light. Thanks for the info re: sensitivities :)

Issue #1100, opened 2020-07-07
Formulate symbolic DAEs for forward sensitivity


tinosulzer:
Casadi is turning out to be a bit slow for computing sensitivities. Explore the idea of formulating the sensitivity DAEs explicitly, for example as per Park et al (2018), "Optimal Experimental Design for Parameterization of an Electrochemical Lithium-Ion Battery Model"
<img width="654" alt="Screen Shot 2020-07-07 at 3 26 24 PM" src="https://user-images.githubusercontent.com/20817509/86832493-3bc22d80-c066-11ea-94c6-ba180c5babfb.png">

tinosulzer:
Closed by #1552

PR #1109, opened 2020-07-21
Issue 1082 casadi sensitivity

### LINKED ISSUE #1082: Reformat casadi solver again
Another reformatting of the casadi solver is needed with

- [ ] Better automatic choosing of `dt_max` (`dt_max = 0.1 * min(timescale, t_final)`
- [x]  Generating the integrator once with `[0,1]` and rescaling time - "safe without grid" mode
- [x] Include the event time and state in the solution being returned - done in #1300 
- [ ] Look into a smarter way of dealing with events ("The alternative is to try to be smarter about predicting when events will occur. At constant current you can see most of our events coming from a long way away (i.e. they are pretty much linear, definitely monotonic, in time). So you could envisage an algorithm where after each step you look at the gradient of the event function, look at where the function would cross zero, and choose the size of your next step based on that.")

Anything else?

tinosulzer:
# Description

Reformat casadi solver to allow sensitivities.
Sensitivity API will be reformatted in a subsequent PR.
Doesn't really fix issue #1082 so still leaving that open

## Type of change

# Key checklist:

## Further checks:


tinosulzer:
Yes, I'm reformatting the sensitivity API (getting rid of `ProcessedSymbolicVariable`) as part of #1100. Currently the model ignores events when doing sensitivity analysis. I can't think of a good way to handle events.

PR #1111, opened 2020-07-23
#1110 fix bug

### LINKED ISSUE #1110: Error when setting a function with an input parameter
If a function is supposed to vary in space and it is set with an input parameter instead, it sometimes raises an error during the parameter processing stage.

e.g. `param["Negative electrode porosity"] = pybamm.InputParameter("eps_n")` in the DFN

tinosulzer:
# Description

Fix setting a FunctionParameter with an InputParameter

Fixes #1110 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/examples/scripts/DFN.py
+++ b/examples/scripts/DFN.py
@@ -31,7 +31,7 @@ disc.process_model(model)
 # solve model
 t_eval = np.linspace(0, 3600, 100)
 solver = pybamm.CasadiSolver(atol=1e-6, rtol=1e-3)
-solution = solver.solve(model, t_eval, inputs={"eps_n": 0.3})
+solution = solver.solve(model, t_eval)
 
 # plot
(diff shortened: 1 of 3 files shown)


examples/scripts/DFN.py
```diff
@@ -31,7 +31,7 @@
 # solve model
 t_eval = np.linspace(0, 3600, 100)
 solver = pybamm.CasadiSolver(atol=1e-6, rtol=1e-3)
-solution = solver.solve(model, t_eval)
+solution = solver.solve(model, t_eval, inputs={"eps_n": 0.3})
```
ferranbrosa, line 34: I think it would be better to keep the DFN example without inputs, we can have a separate example for this.

tinosulzer: yes, forgot to remove this

PR #1116, opened 2020-07-28
#1056 added examples on how to pass custom drive cycle

### LINKED ISSUE #1056: Extend parameters-cli for drive cycles
**Summary**
Extend the capabilities of parameters-cli (or create a new function) so the drive cycle folder can be copied to a local directory and new drive cycles added locally.

ferranbrosa:
# Description

Added examples in the corresponding notebooks on how to pass a custom drive cycle to PyBaMM. Given that it is a change in the examples only, I didn't add any line to CHANGELOG.

Fixes #1056

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/examples/notebooks/Getting Started/Tutorial 4 - Setting parameter values.ipynb
+++ b/examples/notebooks/Getting Started/Tutorial 4 - Setting parameter values.ipynb
@@ -135,8 +135,8 @@
        " 'EC initial concentration in electrolyte [mol.m-3]': 4541.0,\n",
        " 'Electrode height [m]': 0.065,\n",
        " 'Electrode width [m]': 1.58,\n",
-       " 'Electrolyte conductivity [S.m-1]': <function electrolyte_conductivity_Nyman2008 at 0x7f234684e1e0>,\n",
-       " 'Electrolyte diffusivity [m2.s-1]': <function electrolyte_diffusivity_Nyman2008 at 0x7f2391b5a268>,\n",
+       " 'Electrolyte conductivity [S.m-1]': <function electrolyte_conductivity_Nyman2008 at 0x7f9afbc630d0>,\n",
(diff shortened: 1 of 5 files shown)


ferranbrosa:
> Looks good to me, thanks! I think that now that this works, it should be the only way to specify drive cycles (i.e. our examples should only show that way). And can you adapt `Simulation` to figure out that if current is an `Interpolant`, it should use the "drive cycle" operating mode and so passing `t_eval` is not required?

When you say that it should be the only way to specify drive cycles you mean to remove the examples where we do `"[current data]US06"`?

tinosulzer:
Yes, that's what I mean (@rtimms  do you agree?). And make it clear that the drive cycle data can be anywhere, `inputs/drive_cycles` is just where we happen to store the examples

ferranbrosa:
Shall I just remove the functionality or place a deprecation warning?

rtimms:
Yeah, I agree. It makes it much easier for people to use their own drive cycle data and gives you more control over the interpolant, which is nice. Thanks @ferranbrosa !

ferranbrosa:
I have now implemented Tino's comments. I haven't removed the old functionality or added a deprecation warning but it should be fairly easy to do.

examples/notebooks/Getting Started/Tutorial 4 - Setting parameter values.ipynb
```diff
@@ -18,17 +18,9 @@
...
-     "output_type": "stream",
-     "text": [
-      "Note: you may need to restart the kernel to use updated packages.\n"
-     ]
-    }
-   ],
+   "outputs": [],
    "source": [
-    "%pip install pybamm -q    # install PyBaMM if it is not installed\n",
+    "#%pip install pybamm -q    # install PyBaMM if it is not installed\n",
```
tinosulzer, line 23: why has this been commented out?

ferranbrosa: That was a typo, thanks!

PR #1122, opened 2020-08-04
Add BaseCoupledSubModel

dalonsoa:
# Description

Add BaseCoupledSubModel that allows to couple submodels that require a more intimate modification of their equations, boundary or initial conditions than that allowed via coupled variables. A classic example is the case of having two processes that can be considered in isolation but that result in more complex physics if they are considered together. Rather than modifying the underlying submodels, the coupling can be added via this submodel, overriding the parts of the equations or boundary conditions that are required in order to implement the coupling.

Eg. If var_x and var_y are variables of submodels X and Y respectively, then a possible coupling affecting the RHS could be:

```python
rhs[var_x] = rhs[var_x] + f(var_x, var_y)
rhs[var_y] = rhs[var_y] + g(var_x, var_y)
```

If nothing is overriden, the result of using this BaseCoupledSubModel with two or more submodels should be identical to adding those submodels independently. Also, care must be taken if one of the submodels depends on variables defined in the others. In that case, the order in which the submodels are added becomes essential.

Fixes # None - Impromptu coding

## Type of change

# Key checklist:

## Further checks:

- [ ] Tests added that prove fix is effective or that feature works -> To be done after discussion

### DIFF
--- a/pybamm/models/submodels/base_coupled_submodels.py
+++ b/pybamm/models/submodels/base_coupled_submodels.py
@@ -1,0 +1,192 @@
+#
+# Base submodel class
+#
+import pybamm
+from typing import Dict
+
+
+class BaseCoupledSubModel(pybamm.BaseSubModel):
+    """Base class for coupled submodels.
+
+    Couples submodels that require a more intimate modification of their equations,
+    boundary or initial conditions than that allowed via coupled variables. A classic
+    example is the case of having two processes that can be considered in isolation but
+    that result in more complex physics if they are considered together. Rather than
+    modifying the underlying submodels, the coupling can be added via this submodel,
+    overriding the parts of the equations or boundary conditions that are required in
+    order to implement the coupling.
+
+    Eg. If var_x and var_y are variables of submodels X and Y respectively, then a
+    possible coupling affecting the RHS could be:
+
+    rhs[var_x] = rhs[var_x] + f(var_x, var_y)
+    rhs[var_y] = rhs[var_y] + g(var_x, var_y)
+
+    If nothing is overriden, the result of using this BaseCoupledSubModel with two or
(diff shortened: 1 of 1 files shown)


tinosulzer:
I'm not yet convinced that this is necessary. For the above example, you could do it like this with the existing structure, with `ModelX` and `ModelY`.
`ModelX.get_fundamental_variables`: define `var_x`
`ModelY.get_fundamental_variables`: define `var_y`
`ModelX.get_coupled_variables`: define `rhs_of_var_x` and `f_of_var_x_var_y` and s
`ModelY.get_coupled_variables`: define `rhs_of_var_y` and `g_of_var_x_var_y`
Then you have everything you need. Adding the `BaseCoupledSubmodel` would work, but it just means there is more than one way of doing the same thing, which can create confusion

dalonsoa:
Sure, that can be done, of course, but it requires to modify the existing submodels, or submodel the submodels to add `rhs_of_var_x`, `f_of_var_x_var_y`, `rhs_of_var_y` and `g_of_var_x_var_y` to the corresponding `get_coupled_variables`. This approach seems cleaner, sepparating the coupling - which might be relevant or not, depending on the situation - from the base submodels.

tinosulzer:
Can you give an example of a model where this would be required? I am worried that the existing submodel structure is already a little hard to understand, and this would make it more complicated.

dalonsoa:
The degradation mechanisms, in general, are sub models that are likely to be interacting with each other. There’s already the SEI implemented by Felipe, Simon is working on Li platting and Weilong on cracking. If the three of them are present simultaneously, they are not independent processes and I think we will like to model such interactions independently of the individual sub models.

________________________________

dalonsoa:
I've convinced myself that this submodel is, indeed, unnecessary. So closing PR...

Issue #900, opened 2020-03-20
Proposal for a Spatial Method: Spectral Volumes


YannickNoelStephanKuhn:
Spectral Volumes are a higher-order generalization of Finite Volumes. They subdivide the same mesh Finite Volumes would use into so-called Control Volumes. These are used to compute a polynomial in each Finite Volume which is conservative in each Control Volume. Both the polynomial and its integral average over the Finite Volume are a higher-order approximation.

In contrast to higher-order Finite Volumes, which have all sorts of difficult numerical issues, Spectral Volumes are easily incorporated into any (1D, 2D, 3D) mesh. Additionally, they give higher-order approximations for boundary values which give much better resolution of e.g. Butler-Volmer flows. Compared to Orthogonal Collocation or other Spectral Methods, Spectral Volumes have the following additional benefits:
 - finer resolutions don't require an increase in polynomial order => higher numerical stability;
 - arbitrary subdivisions are possible (but Chebyshev nodes are the most stable);
 - they are a conservative method => suited for long-time simulations.

I implemented 2nd-order Spectral Volumes for my own battery code, and while I couldn't see that the convergence rate increases globally as I hoped for, the convergence rate of interfacial processes was indeed higher. My proposal is that I implement Spectral Volumes of arbitrary order in 1D as a Spatial Method for PyBaMM. A discussion about the storage of the additional internal states and possible output of these internal states for plotting would be necessary first, though.
Relevant paper: https://www.sciencedirect.com/science/article/pii/S0021999102970415

TL;DR: Spectral Volumes are the Finite Volume-equivalent to Spectral Methods for Finite Elements. Since they are not a Spectral Method per se, I didn't write this as a reply to issue #565.

tinosulzer:
Hi @YannickNoelStephanKuhn , that sounds great, thanks for suggesting this! I think the easiest way to tackle this will be to create equations of increasing complexity and make sure they get discretised properly. For example:
```python3
# define discretisation
# ....
var = pybamm.Variable("var", domain="negative electrode")
disc.set_variable_slices([var])
# first test
eqn = var
eqn_disc = disc.process_symbol(eqn)
# test discretised as expected
# ....
# second test
eqn = pybamm.grad(var)
eqn_disc = disc.process_symbol(eqn)
# test discretised as expected
# ....
# third test
eqn = pybamm.div(pybamm.grad(var))
eqn_disc = disc.process_symbol(eqn)
# test discretised as expected
# ....
```
[...]

PR #1123, opened 2020-08-04
Spectral volumes

YannickNoelStephanKuhn:
Added the Spectral Volumes spatial method (as /pybamm/spatial_methods/spectral_volumes.py) and the specific submesh that it works with (added to /pybamm/meshes/one_dimensional_submeshes.py), called SpectralVolumes1DSubMesh, as described in #900 .

### DIFF
--- a/pybamm/meshes/one_dimensional_submeshes.py
+++ b/pybamm/meshes/one_dimensional_submeshes.py
@@ -309,11 +309,12 @@ class UserSupplied1DSubMesh(SubMesh1D):
         super().__init__(edges, coord_sys=coord_sys, tabs=tabs)
 
 
-class SpectralVolumes1DSubMesh(SubMesh1D):
+class SpectralVolume1DSubMesh(SubMesh1D):
     """
     A class to subdivide any mesh to incorporate Chebyshev collocation
-    Control Volumes. Note that while the Spectral Volumes method can use
-    any mesh with the right number of nodes, in 1D the only sensible
+    Control Volumes. Note that the Spectral Volume method is optimized
+    to only work with this submesh. The underlying theory could use any
+    mesh with the right number of nodes, but in 1D the only sensible
     choice are the Chebyshev collocation points.
 
     Parameters
@@ -325,18 +326,47 @@ class SpectralVolumes1DSubMesh(SubMesh1D):
         each spatial variable. Note: the number of nodes (located at the
         cell centres) is npts, and the number of edges is npts+1.
     order : int, optional
-        The order of the Spectral Volumes method that is to be used with
+        The order of the Spectral Volume method that is to be used with
         this submesh. The default is 2, the same as the default for the
-        SpectralVolumes class. If the orders of the submesh and the
-        Spectral Volumes method don't match, the method will fail.
+        SpectralVolume class. If the orders of the submesh and the
+        Spectral Volume method don't match, the method will fail.
     **Extends:"": :class:`pybamm.SubMesh1D`
     """
 
-    def __init__(self, lims, npts, order=2):
+    def __init__(self, lims, npts, edges=None, order=2):
 
         spatial_var, spatial_lims, tabs = self.read_lims(lims)
         npts = npts[spatial_var.id]
-        edges = np.linspace(spatial_lims["min"], spatial_lims["max"], npts + 1)
+
+        # default: Spectral Volumes of equal size
+        if edges is None:
+            edges = np.linspace(spatial_lims["min"], spatial_lims["max"],
+                                npts + 1)
+        # check that npts + 1 equals number of user-supplied edges
+        elif (npts + 1) != len(edges):
+            raise pybamm.GeometryError(
+                "User-suppled edges should have length (npts + 1) but has len"
+                "gth {}. Number of points (npts) for domain {} is {}.".format(
+                    len(edges), spatial_var.domain, npts
+                )
+            )
+
+        # check end points of edges agree with spatial_lims
+        if edges[0] != spatial_lims["min"]:
+            raise pybamm.GeometryError(
+                """First entry of edges is {}, but should be equal to {}
+                 for domain {}.""".format(
+                    edges[0], spatial_lims["min"], spatial_var.domain
+                )
+            )
+        if edges[-1] != spatial_lims["max"]:
+            raise pybamm.GeometryError(
+                """Last entry of edges is {}, but should be equal to {}
+                for domain {}.""".format(
+                    edges[-1], spatial_lims["max"], spatial_var.domain
+                )
+            )
+
         coord_sys = spatial_var.coord_sys
 
         cv_edges = np.array([edges[0]] + [
--- a/pybamm/meshes/one_dimensional_submeshes.py
+++ b/pybamm/meshes/one_dimensional_submeshes.py
@@ -373,7 +373,10 @@ class SpectralVolume1DSubMesh(SubMesh1D):
             x
             for (a, b) in zip(edges[:-1], edges[1:])
             for x in np.flip(
-                self.chebyshev_collocation_points(order + 1, a, b)
+                a + 0.5 * (b - a) * (1 + np.sin(np.pi * np.array(
+                    [((order + 1) - 1 - 2 * i) / (2 * (order + 1) - 2)
+                     for i in range(order + 1)]
+                )))
             )[1:]
         ])
 
@@ -386,21 +389,3 @@ class SpectralVolume1DSubMesh(SubMesh1D):
         # "edges" and "nodes" properties. This makes some of the
         # code of FiniteVolume directly applicable.
         super().__init__(cv_edges, coord_sys=coord_sys, tabs=tabs)
-
-    def chebyshev_collocation_points(self, noe, a=-1.0, b=1.0):
-        """
-        Calculates Chebyshev collocation points in descending order.
-
-        Parameters
-        ----------
-        noe: integer
-            The number of the collocation points. "number of edges"
-
-        Returns
-        -------
-        :class:`numpy.array`
-        Chebyshev collocation points on [a,b].
-        """
-
-        return a + 0.5 * (b - a) * (1 + np.sin(np.pi * np.array(
-            [(noe - 1 - 2 * i) / (2 * noe - 2) for i in range(noe)])))
--- a/pybamm/meshes/one_dimensional_submeshes.py
+++ b/pybamm/meshes/one_dimensional_submeshes.py
@@ -340,12 +340,12 @@ class SpectralVolumes1DSubMesh(SubMesh1D):
         coord_sys = spatial_var.coord_sys
 
         cv_edges = np.array([edges[0]] + [
-                x
-                for (a, b) in zip(edges[:-1], edges[1:])
-                for x in np.flip(
-                    self.chebyshev_collocation_points(order + 1, a, b)
-                )[1:]
-            ])
+            x
+            for (a, b) in zip(edges[:-1], edges[1:])
+            for x in np.flip(
+                self.chebyshev_collocation_points(order + 1, a, b)
+            )[1:]
+        ])
 
         self.sv_edges = edges
         self.sv_nodes = (edges[:-1] + edges[1:]) / 2
--- a/pybamm/meshes/one_dimensional_submeshes.py
+++ b/pybamm/meshes/one_dimensional_submeshes.py
@@ -307,3 +307,70 @@ class UserSupplied1DSubMesh(SubMesh1D):
         coord_sys = spatial_var.coord_sys
 
         super().__init__(edges, coord_sys=coord_sys, tabs=tabs)
+
+
+class SpectralVolumes1DSubMesh(SubMesh1D):
+    """
+    A class to subdivide any mesh to incorporate Chebyshev collocation
+    Control Volumes. Note that while the Spectral Volumes method can use
+    any mesh with the right number of nodes, in 1D the only sensible
+    choice are the Chebyshev collocation points.
+
+    Parameters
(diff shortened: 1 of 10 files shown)


pybamm/spatial_methods/spectral_volumes.py
```diff
@@ -0,0 +1,617 @@
...
+
+    Parameters
+    ----------
+    mesh : :class:`pybamm.Mesh`
+        Contains all the submeshes for discretisation
+
+    **Extends:"": :class:`pybamm.FiniteVolume`
+    """
+
+    def __init__(self, options=None, order=2):
```
tinosulzer, line 52: `mesh` is not a parameter here? Also, when it is made a parameter, there should be a check that it is the `SpectralVolume1DSubMesh`, raising an error if not

pybamm/spatial_methods/spectral_volumes.py
```diff
@@ -0,0 +1,617 @@
...
+        super().__init__(options)
+
+    def chebyshev_collocation_points(self, noe, a=-1.0, b=1.0):
+        """
+        Calculates Chebyshev collocation points in descending order.
+
+        Parameters
+        ----------
+        noe: integer
+            The number of the collocation points. "number of edges"
```
tinosulzer, line 63: docstring a and b

pybamm/meshes/one_dimensional_submeshes.py
```diff
@@ -307,3 +307,70 @@ def __init__(self, lims, npts, edges=None):
         coord_sys = spatial_var.coord_sys
 
         super().__init__(edges, coord_sys=coord_sys, tabs=tabs)
+
+
+class SpectralVolumes1DSubMesh(SubMesh1D):
```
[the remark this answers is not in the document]
YannickNoelStephanKuhn: My mesh is basically a series of Chebyshev1DSubMeshes. From the perspective of the Spectral Volumes method, the Chebyshev1DSubMesh is only one Spectral Volume; that's like a Finite Volume method with only one Volume.

pybamm/spatial_methods/spectral_volumes.py
YannickNoelStephanKuhn: Well, it wasn't a parameter for the "FiniteVolume" constructor. I will overwrite the "build"-method where the "mesh" parameter is present and implement the check there.

pybamm/spatial_methods/spectral_volumes.py
YannickNoelStephanKuhn: Sorry, my bad.

YannickNoelStephanKuhn:
The convergence tests might not be much of a problem, since all I should have to do is copy test_finite_volume.py and regex-replace every instance of FiniteVolume with SpectralVolume and replace every mesh (except the 0D ones) with SpectralVolume1DSubMesh.

I'm struggling to find the part in PyBaMM that makes e.g. "pybamm.FiniteVolume" work instead of "pybamm.spatial_methods.FiniteVolume". Where do I add SpectralVolume and SpectralVolume1DSubMesh such that "pybamm.SpectralVolume" works?

YannickNoelStephanKuhn:
The Spectral Volume method almost works now. The issue that I still have is that it doesn't work correctly across concatenation boundaries. For example, with the SPMe, the electrolyte concentration profile looks after a short discharge looks like this:
```
-----
     \
      \
       ---
          \
           \
            \
             ------
```
The plateaus correspond to negative electrode, separator and positive electrode. The separator plateau has the right value, but the other two speed away. As far as I can tell, they have the right shape, just a (growing) wrong offset.

I will have a look into the concatenation method of FiniteVolume in order to fix this. If someone recognizes what happens here, please let me know.

Also, if the added optional argument to gradient_matrix truly is a problem, I'll rewrite that part to not rely on it.

tinosulzer:
I don't know what is happening there. For FiniteVolume, the subdomains aren't treated in a special way. Once you have a vector `[neg, sep, pos]`, that is just treated as one vector and the matrix multiplications are done directly on the concatenated vector.

YannickNoelStephanKuhn:
I checked that the following parts are right:
 - the gradient is correct
 - the divergence is correct
 - the treatment of boundary conditions is correct
 - the treatment of concatenations is correct
 - the solution of particle diffusion is correct
 - the solution of electrolyte diffusion is correct

I narrowed the problem with the incorrect solution of the SPMe down to the following: the solution of the electrolyte concentration is not enforced to be continuous. Which makes some sense, since I only implemented internal Neumann conditions when copying FiniteVolume.

So the reconstructed gradients in SpectralVolume are 0 from either side with the wrong solution from above, regardless of how far apart the variable values are at the boundary. That satisfies the internal Neumann conditions, and the solution of the SPMe with no continuous electrolyte is exactly the picture that I drew above. With FiniteVolume, different variable values from either side of the boundary result in a non-zero gradient. In a diffusion equation, this makes the two variable values at the boundary approach each other.

As I don't know all functions that get called when solving a model, I have the following question: is there an equivalent to "internal_neumann_condition" for internal Dirichlet conditions, i.e. continuity? If not, I'll see if I can add some sort of penalty function to the internal Neumann condition that enforces continuity.

tinosulzer:
> As I don't know all functions that get called when solving a model

The solver is just expecting functions that evaluate to the `rhs` and `algebraic` terms. I guess the gradients shouldn't be zero at the interface between subdomains? Maybe see here for some inspiration, https://github.com/davidhowey/Spectral_li-ion_SPM, and also Adrien Bizeray's PhD thesis

YannickNoelStephanKuhn:
Thank you very much for all your help. I'm pretty sure that the spectral method used there is orthogonal collocation, with which I have three issues: it is not conservative, has very sparse nodes in the middle of discretised domains and it has stability problems with high node counts. This is why I searched for another method in the first place.

I went ahead and used a simple finite difference to ensure proper handling of discontinuities. That at least ensures a performance that is as good as FiniteVolume, compared with the same amount of mesh nodes each time. I'm not sure if the first order finite difference kills the higher convergence order, but for now, I'm happy that it works. I didn't want it to be overly complicated with arbitrary-order finite differences at the boundaries.

The tests are implemented as you asked, apart from a variation of test_dfn.py; I just ran SpectralVolume in a DFN simulation and saw that it works. The tests only check for linear convergence though, since the higher convergence order, if present, only manifests in the laplacian.

pybamm/meshes/one_dimensional_submeshes.py
```diff
@@ -307,3 +307,100 @@ def __init__(self, lims, npts, edges=None):
...
+
+        Parameters
+        ----------
+        noe: integer
+            The number of the collocation points. "number of edges"
+
+        Returns
+        -------
+        :class:`numpy.array`
+        Chebyshev collocation points on [a,b].
```
tinosulzer, line 402: missing indentation

YannickNoelStephanKuhn:
As far as I can tell, I did implement the test for my 1D submesh. Why does codecov complain that it isn't tested?

pybamm/meshes/one_dimensional_submeshes.py
YannickNoelStephanKuhn: *internally facepalms* Thank you very much.

tinosulzer:
It is weird that that symbol breaks, since Ambient temperature and OCP entropic change are both parameters (i.e. scalars). Could you just do the isothermal example for now?

For codecov, that is weird. I tested your fork locally and see that it is calling the file, so I don't know what's going wrong.

YannickNoelStephanKuhn:
Well, I'll just add a separate test for the method and see if that fixes it.

Regarding the broken symbol: it was just a result of me using model.new_copy() instead of using two separate models all the way through. Interesting that it randomly breaks in the middle of discretisation though.

ferranbrosa:
> As far as I can tell, I did implement the test for my 1D submesh. Why does codecov complain that it isn't tested? Is it the method that the class contains? It is implicitly tested, since it is part of the constructor, but does codecov not recognise this as "tested"?

I think it complains because some lines are not executed when testing the code. For `one_dimensional_submeshes.py` the issue seems to be in line 343 as apparently the code never enters the following if statement
```
if edges is None:
    edges = np.linspace(spatial_lims["min"], spatial_lims["max"], npts + 1)
```

On the other hand, there are many lines not being tested in `spectral_volume.py` but that doesn't appear in red in the report, probably because it is a new file.

YannickNoelStephanKuhn:
Well, let's satisfy the codecov gods then.

If I didn't miss anything, the test coverage of spectral_volume.py is identical to the one of finite_volume.py, so I don't think there is a problem there.

ferranbrosa:
It seems that the issue is that the `test_spectral_volume.py` is not run because none of the lines of code is being run (not even `__init__`). When you run the integration tests locally, is that test run? Maybe the coverage is skipping that test for some reason...

YannickNoelStephanKuhn:
Here is the end of the output when I run `python run-tests.py --unit --folder=integration` locally:

```
test_cartesian_div_convergence (test_spatial_methods.test_finite_volume.TestFiniteVolumeConvergence) ... ok
test_cartesian_spherical_grad_convergence (test_spatial_methods.test_finite_volume.TestFiniteVolumeConvergence) ... ok
test_grad_div_broadcast (test_spatial_methods.test_finite_volume.TestFiniteVolumeConvergence) ... ok
test_p2d_spherical_convergence_quadratic (test_spatial_methods.test_finite_volume.TestFiniteVolumeConvergence) ... ok
test_p2d_with_x_dep_bcs_spherical_convergence (test_spatial_methods.test_finite_volume.TestFiniteVolumeConvergence) ... ok
test_spherical_div_convergence_linear (test_spatial_methods.test_finite_volume.TestFiniteVolumeConvergence) ... ok
test_spherical_div_convergence_quadratic (test_spatial_methods.test_finite_volume.TestFiniteVolumeConvergence) ... ok
test_cartesian_div_convergence (test_spatial_methods.test_spectral_volume.TestSpectralVolumeConvergence) ... ok
test_cartesian_spherical_grad_convergence (test_spatial_methods.test_spectral_volume.TestSpectralVolumeConvergence) ... ok
test_grad_div_broadcast (test_spatial_methods.test_spectral_volume.TestSpectralVolumeConvergence) ... ok
test_p2d_spherical_convergence_quadratic (test_spatial_methods.test_spectral_volume.TestSpectralVolumeConvergence) ... ok
test_p2d_with_x_dep_bcs_spherical_convergence (test_spatial_methods.test_spectral_volume.TestSpectralVolumeConvergence) ... ok
test_spherical_div_convergence_linear (test_spatial_methods.test_spectral_volume.TestSpectralVolumeConvergence) ... ok
test_spherical_div_convergence_quadratic (test_spatial_methods.test_spectral_volume.TestSpectralVolumeConvergence) ... ok
```

As far as I can tell, SpectralVolume is a carbon-copy of FiniteVolume here. Maybe the coverage only detects the methods as such as tested (hence the green markers next to their definitions), but not their individual lines, because there is only an integration test and no unit tests?

rtimms:
sorry, late to the conversation here, but I *think* only unit tests count towards the percentage on codecov, so you'll need to add something like the tests in 'tests/unit/test_spatial_methods/test_finite_volume/test_finite_volume.py'. The integration tests are obviously useful too, and the build fails if the integration tests fail, but they don't count towards coverage

YannickNoelStephanKuhn:
Now I'm especially grateful that I inherited from FiniteVolume, since that means I only need to adjust a small subset of the unit tests for SpectralVolume. I also had to add a flag to run-tests.py to run the tests locally, since my Python interpreter is named python3.7 instead of python.

The only notable differences between the unit tests for FiniteVolume and SpectralVolume are the SpectralVolume1DSubMeshes and that the gradient operators return a discretised symbol with length npts + 1 (internal nodepoints plus boundary conditions), whereas for FiniteVolume that length was npts - 1 (internal nodepoints).

tinosulzer:
Good find! Glad we figured this out :) I think the method names themselves are always read when a class is loaded in `__init__.py`, but the individual lines are only read when tested

YannickNoelStephanKuhn:
Thanks everyone! Apart from the exceptions that I didn't assert, I missed a few lines in the methods that apply the boundary conditions. For comparison, the corresponding code snippet from FiniteVolume (finite_volume.py, 663-668):

```
            if lbc_value.evaluates_to_number():
                left_ghost_constant = (
                    2 * lbc_value * pybamm.Vector(np.ones(second_dim_repeats))
                )
            else:
                left_ghost_constant = 2 * lbc_value
```

I don't have a test case where the boundary values don't "evaluates_to_number". Now that's annoying, since I have no idea how to set boundary conditions that result in the else-branch.

tinosulzer:
Interesting, it's possible that we can get away with removing that if statement, since
```
left_ghost_constant = (
                    2 * lbc_value * pybamm.Vector(np.ones(second_dim_repeats))
                )
```
should always give the right thing (worst case scenario, it just gets unecessarily multiplied by a vector of ones, and then simplified out later).
The use case here is (for example) if a variable depends on x and r, but is only differentiated in the r direction, so the boundary condition depends on x (i.e. is a vector, doesn't "evaluate to number"). In battery models this mainly comes in the form of a Neumann BC for the particle flux at the surface in the DFN.

YannickNoelStephanKuhn:
Out of fear of breaking anything at this stage, I will leave the else-branch there. But feel free to optimize those else branches out if they are removed from FiniteVolume eventually and nothing breaks.

For future reference: the lines where the if-else-branches are and could be replaced with just the content of the if-branch: 553-557 and 571-575. But they will stand out like a sore thumb anyway. Note: it is very much intended that it says "lbc_value" instead of "2 * lbc_value" (likewise for "rbc"), since the boundary conditions are just copied into the discretised symbol.

PR #1125, opened 2020-08-10
now works with python 3.8

tinosulzer:
# Description

Casadi now works with python 3.8, so pybamm should too. This PR updates the tests and install docs.


tlestang:
Looks like the CI is passing, so all good :+1:

PR #1130, opened 2020-08-21
Polynomial concentration

### LINKED ISSUE #1128: Add polynomial approximations to particle concentration
Add the two- and three-parameter models from the paper Efficient Macro-Micro Scale Coupled Modeling of Batteries Venkat R. Subramanian et al 2005 J. Electrochem. Soc. 152 A2002

rtimms:
# Description

Adds particle submodels that approximate the concentration profile within the particle as a polynomial in r. Note the "fast diffusion" option is now accessed as the "uniform profile" option (fast diffusion leads to uniform concentration in r, but it makes sense to have uniform concentration as the zero-order case of the polynomial model).

Fixes #1128 

## Type of change

# Key checklist:

## Further checks:


pybamm/models/submodels/particle/polynomial_many_particles.py
```diff
@@ -0,0 +1,299 @@
...
+        self.initial_conditions = {c_s_rav: c_init}
+
+        if self.order in [2, 4]:
+            # We also need to provide an initial condition (initial guess for the
+            # algebraic solver) for the surface concentration
+            c_s_surf = variables[self.domain + " particle surface concentration"]
+            self.initial_conditions.update({c_s_surf: c_init})
+        if self.order == 4:
+            # We also need to provide an initial condition (initial guess for the
+            # algebraic solver) for the average concentration gradient
```
rtimms, line 295: I will remove the comment about algebraic solver here, as it is an ODE for q (accidentally left the comment here from copy/paste above)

pybamm/models/submodels/particle/polynomial_single_particle.py
```diff
@@ -0,0 +1,356 @@
...
+            # The concentration is given by c = A + B*r**2
+            A = pybamm.PrimaryBroadcast(
+                (1 / 2) * (5 * c_s_rxav - 3 * c_s_surf_xav),
+                [self.domain.lower() + " particle"],
+            )
+            B = pybamm.PrimaryBroadcast(
+                (5 / 2) * (c_s_surf_xav - c_s_rxav), [self.domain.lower() + " particle"]
+            )
+            if self.domain == "Negative":
+                # TODO: figure out how to just use r from standard_spatial_vars
```
rtimms, line 173: I kept getting shape errors here and couldn't figure out the correct way to broadcast things so I could use `r` from `standard_spatial_vars` here. I thought I should broadcast A, B, (C), then use that to define `c_s`, and then average to get `c_s_av`, but I kept getting bogged down in shape errors

pybamm/models/submodels/particle/polynomial_single_particle.py
```diff
@@ -0,0 +1,356 @@
...
+
+        if self.domain == "Negative":
+            c_init = self.param.c_n_init(0)
+
+        elif self.domain == "Positive":
+            c_init = self.param.c_p_init(1)
+
+        self.initial_conditions = {c_s_rxav: c_init}
+        if self.order == 4:
+            # We also need to provide an initial condition (initial guess for the
```
rtimms, line 349: see above comment about removing the comment about algebraic solvers here

pybamm/models/submodels/particle/polynomial_single_particle.py
tinosulzer: Since `c_s_xav` doesn't depend on `x`, it makes sense that you should use a different `r` which only has "negative particle" and "current collector" as domains

pybamm/models/full_battery_models/lithium_ion/dfn.py
```diff
@@ -85,13 +85,23 @@ def set_particle_submodel(self):
...
+            elif self.options["particle"] == "quadratic profile":
+                order = 2
+            elif self.options["particle"] == "quartic profile":
+                order = 4
+            self.submodels[
+                "negative particle"
+            ] = pybamm.particle.PolynomialManyParticles(self.param, "Negative", order)
+            self.submodels[
+                "positive particle"
+            ] = pybamm.particle.PolynomialManyParticles(self.param, "Positive", order)
```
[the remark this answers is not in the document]
rtimms: agree

pybamm/models/submodels/particle/base_particle.py
```diff
@@ -49,6 +66,10 @@ def _get_standard_concentration_variables(self, c_s, c_s_xav):
             "R-averaged "
             + self.domain.lower()
             + " particle concentration [mol.m-3]": c_s_rav * c_scale,
+            "R-X-averaged " + self.domain.lower() + " particle concentration": c_s_av,
+            "R-X-averaged "
```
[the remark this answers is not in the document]
rtimms: I'm happy to change to "Average". I think it is obvious and less cumbersome.

TomTranter (approved):
Hey Rob, thanks for this, seems like quite a lot of work in the end. Looks great and works well. I added another notebook for the models which included comparing at different C-rates. Feel free to change bits or move it to another folder. I did get some ugly solver errors in the notebook at 5 A which I guess we could hide but are they anything to worry about? Still gets a solution...

rtimms:
thanks @TomTranter notebook looks great! made some minor adjustments. the solver errors were just because of the high current - the solver was taking steps/doing jacobian calculations that were later rejected and the step size decreased. in practice they are nothing to worry about, but I've changed the currents in the notebooks to avoid the warnings and still show off the same results. probably best to avoid potentially confusing solver errors in the examples

PR #1134, opened 2020-08-28
Issue 1126 upwind

### LINKED ISSUE #1126: Add upwinding and downwinding operators for convection-dominated flow
Consider a standard reaction-advection-diffusion equation,
```
dc/dt = D d2c/dx2 - d/dx( c * v ) + s
```
Currently, with the `FiniteVolume` discretisation, the `c` term in convection gets discretised with a central volume (in `process_binary_operators`). However, if transport is convection dominated (`D << 1`), then
1. The numerical solution is unstable, http://www.mathematik.uni-dortmund.de/~kuzmin/cfdintro/lecture5.pdf
1. the boundary conditions are almost entirely ignored, since boundary conditions only come in through the diffusion operator.

This can be fixed by introducing an upwinding (if v > 0) or downwinding (if v < 0) term, which should be explicitly stated when defining the equations (this could perhaps be automated at a later stage). The resulting discretisation of `upwind(c)` would be a concatenation `[bc_in, c]`, where `bc_in` is chosen to satisfy the Dirichlet boundary condition at c=0. Similarly, the discretisation of `downwind(c)` would be `[c, bc_in]`. This is the lowest-order discretisation possible, higher-order corrections are possible, see http://www.mathematik.uni-dortmund.de/~kuzmin/cfdintro/lecture5.pdf (e.g. "Quadratic upwind difference scheme (QUICK)") and may be required to avoid smearing.

tinosulzer:
# Description

Add upwind and downwind operators for convection.

Fixes #1126 

## Type of change

# Key checklist:

## Further checks:


rtimms (approved):
thanks @tinosulzer !

Issue #1117, opened 2020-07-29
Total lithium in electrodes and electrolyte


ferranbrosa:
**Summary**
Add a new model variable that it is the total lithium concentration in each electrode and in the electrolyte.

 **Motivation**
These new variables will be used to verify that lithium is conserved.

PR #1136, opened 2020-09-04
Issue 1112 ec sei params

### LINKED ISSUE #1112: Some parameters not getting set by EC SEI submodel
The EC SEI submodel doesn't set things like loss of lithium

rtimms:
# Description
 - Adds new variables used to track the total lithium concentration (thanks @ferranbrosa !)
 - Fix a bug where not all parameters are set by the EC SEI model (e.g. loss of lithium)
 - Slightly reformat the way variables related to concentration are set by the sei models  

Fixes #1112  and #1117 

## Type of change

# Key checklist:

## Further checks:


pybamm/models/submodels/electrolyte_diffusion/base_electrolyte_diffusion.py
```diff
@@ -105,6 +105,37 @@ def _get_standard_flux_variables(self, N_e):
...
+        Returns
+        -------
+        variables : dict
+            The variables which can be derived from the flux in the
+            electrolyte.
+        """
+
+        c_e_typ = self.param.c_e_typ
+        L_x = self.param.L_x
+
```
rtimms, line 128: should this be epsilon * c_e here?

tinosulzer: Yep

pybamm/models/submodels/interface/sei/ec_reaction_limited.py
```diff
@@ -131,3 +111,48 @@ def set_initial_conditions(self, variables):
         j_sei_0 = pybamm.Scalar(0)
 
         self.initial_conditions = {L_sei: L_sei_0, j_sei: j_sei_0}
+
+    def _get_standard_concentraion_variables(self, variables):
```
tinosulzer, line 115: If this function is only ever called from `get_coupled_variables`, there is no need to make it a separate function. Ideally, we should find a way to make it use the method from `base_sei` (e.g. with inner set to 0 and outer representing the total)

tinosulzer (commented):
Looks good, but I think there should be a better way to do #1112 .

Also, there are some `.pkl` files in `examples/notebooks` to remove. They should probably be removed when running the notebooks (using `os.remove`)

tinosulzer (approved):
Thanks @rtimms , merging

PR #1165, opened 2020-09-11
Add readthedocs config file

### LINKED ISSUE #1164: API documentation missing on RTD
**Describe the bug**
The API documentation on readthedocs is missing/empty.

**To Reproduce**
1. Go to https://pybamm.readthedocs.io/en/latest/#module-pybamm
2. Click on any class
3. No documentation to be seen

**Expected behaviour**
The API documentation should contain the same information found in class and method docstrings, as it does when building the documentation locally.

**Screenshots**
On readthedocs:

Local build:

tlestang:
This adds a `.readthedocs.yml` at the root of the PyBaMM repo.
This makes configuring Read The Docs easier, without the need to log in to RTD and change the settings there.

This change originates from issue #1164 , caused by the file `.requirements-docs` not being up to date with the current dependencies.
This file was updated, renamed to `docs/requirements.txt` and set as the requirement file for readthedocs in `.readthedocs.yml`.

Fixes #1164


tinosulzer (approved):
Thanks for fixing this @tlestang

PR #1175, opened 2020-09-22
Issue 1174 raise error bcs sphere other than no-flux

### LINKED ISSUE #1174: "spherical polar" coordinates ignore left boundary condition
**Describe the bug**
When using "spherical polar" coordinates, the left boundary condition seems to always be zero flux, regardless of what the user specifies.

**To Reproduce**
1. Run the 'Creating Models/2-a-pde-model.ipynb' notebook.
2. Change the left boundary condition type and/or value to anything.
3. Rerun notebook.
4. Final result is the same.

**Expected behaviour**
The left boundary condition should be user-changeable. Alternatively, if zero flux is intentionally enforced on the left boundary in spherical coordinates, the user should be warned of this fact if they request something else.

ferranbrosa:
# Description

The code now raises an error if the boundary condition at the origin in a sphere geometry is other than a no-flux condition.

Fixes #1174

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ## Bug fixes
 
 -   Raise error if the boundary condition at the origin in a spherical domain is other than no-flux ([#1175](https://github.com/pybamm-team/PyBaMM/pull/1175))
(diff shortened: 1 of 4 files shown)


tinosulzer (approved):
Perfect, thanks!

PR #1201, opened 2020-10-19
Install scikits in coverage tox

brosaplanella:
# Description

Install `scikits.odes` when running `tox -e coverage`.

Before merging it would be good that coverage installs all the dependencies required though.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -45,7 +45,9 @@ deps = flake8>=3
 commands = python -m flake8
 
 [testenv:coverage]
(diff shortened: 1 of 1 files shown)


brosaplanella:
Coverage went back to what we had before it broke.

tlestang:
Okay, good to hear. I think scikit.odes is the only (optional) dependency that is not listed in the setup.py

tlestang (approved):

PR #1206, opened 2020-10-21
Issue 1203 save to matlab

### LINKED ISSUE #1203: Raise error when saving bad variable name to mat file
If a `.mat` file has a variable name with a space in the name, MATLAB won't be able to load it. So we shouldn't save it. Raise error instead and suggest making a new variable with a more MATLAB-compatible name.
We could try converting spaces to underscores, etc, but units (`[mol.m-3]`) would not work (none of [, ., - are ok)

tinosulzer:
# Description

Add error message and option to use different name if variable name contains spaces or special characters, so that matlab can read it
Fixes #1203 

## Type of change

# Key checklist:

## Further checks:


brosaplanella (approved):
Looks good, thanks Tino!

PR #1208, opened 2020-10-22
Put evaluated timescale and length_scales on solution

TomTranter:
# Description

Previously models would not evaluate timescales if they are functions of input parameters. If an input parameter doesn't change during a solution but was created for purposed of reusing models with different input parameters then it should result in consistent scaling and should be allowed. To address the problem the scaling are saved to the solution and used here for processed variables. A check is made when stepping that the inputs that affect scaling do not change

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/examples/scripts/compare_comsol/compare_comsol_DFN.py
+++ b/examples/scripts/compare_comsol/compare_comsol_DFN.py
@@ -140,9 +140,7 @@ comsol_model.variables = {
 
 # Make new solution with same t and y
 comsol_solution = pybamm.Solution(pybamm_solution.t, pybamm_solution.y)
-# Update model and solution scales to match the pybamm model
-comsol_model.timescale = pybamm_model.timescale
-comsol_model.length_scales = pybamm_model.length_scales
+# Update solution scales to match the pybamm model
 comsol_solution.timescale_eval = pybamm_model.timescale_eval
 comsol_solution.length_scales_eval = pybamm_model.length_scales_eval
 comsol_solution.model = comsol_model
--- a/examples/scripts/compare_comsol/compare_comsol_DFN.py
+++ b/examples/scripts/compare_comsol/compare_comsol_DFN.py
@@ -140,9 +140,11 @@ comsol_model.variables = {
 
 # Make new solution with same t and y
 comsol_solution = pybamm.Solution(pybamm_solution.t, pybamm_solution.y)
-# Update model scales to match the pybamm model
+# Update model and solution scales to match the pybamm model
 comsol_model.timescale = pybamm_model.timescale
 comsol_model.length_scales = pybamm_model.length_scales
+comsol_solution.timescale_eval = pybamm_model.timescale_eval
+comsol_solution.length_scales_eval = pybamm_model.length_scales_eval
 comsol_solution.model = comsol_model
 
 # plot
--- a/pybamm/models/base_model.py
+++ b/pybamm/models/base_model.py
@@ -118,6 +118,11 @@ class BaseModel(object):
         # Default timescale is 1 second
         self.timescale = pybamm.Scalar(1)
         self.length_scales = {}
+        self.timescale_eval = self.timescale.evaluate()
+        self.length_scales_eval = {
+            domain: scale.evaluate()
+            for domain, scale in self.length_scales.items()
+        }
 
     @property
     def name(self):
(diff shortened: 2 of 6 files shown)


pybamm/models/base_model.py
```diff
@@ -118,6 +118,11 @@ def __init__(self, name="Unnamed model"):
         # Default timescale is 1 second
         self.timescale = pybamm.Scalar(1)
         self.length_scales = {}
+        self.timescale_eval = self.timescale.evaluate()
+        self.length_scales_eval = {
+            domain: scale.evaluate()
+            for domain, scale in self.length_scales.items()
+        }
```
tinosulzer, line 125: This is scary. With a model that isn't the base model, you would end up with timescale being some parameter and timescale_eval being equal to 1 (until you update it in the solver) which could get confusing

pybamm/solvers/base_solver.py
```diff
@@ -775,6 +777,28 @@ def step(
...
+                    "The model timescale is a function of an input parameter "
+                    "and the value has changed between steps!"
+                )
+            for domain in temp_length_scales_eval.keys():
+                old_dom_eval = old_solution.length_scales_eval[domain]
+                if temp_length_scales_eval[domain] != old_dom_eval:
+                    pybamm.logger.error(
+                        "The model lengthscale is a function of an input "
+                        "parameter {} and the value has changed between "
+                        "steps!".format(domain)
```
tinosulzer, line 800: Check this string ("domain" is not an input parameter). Should it say "an input parameter affecting the length scale in {}"?

pybamm/solvers/solution.py
```diff
@@ -70,6 +70,10 @@ def __init__(
         for time in t:
             self._known_evals[time] = {}
 
+        # Copy the timescale_eval and lengthscale_evals
+        self.timescale_eval = self._model.timescale_eval
+        self.length_scales_eval = self._model.length_scales_eval
+
```
tinosulzer, line 76: I guess this is the bit that required you to define `model.timescale_eval` at the start? I think we just need to be a bit more careful here. Maybe these two lines could be:
```python3
if hasattr(self._model, "timescale_eval"):
         self.timescale_eval = self._model.timescale_eval
else:
         self.timescale_eval =self._model.timescale.evaluate()
```
or even better put that logic into a class property `model.timescale_eval`

tinosulzer (commented):
Looks good, I think this is a neat solution. I just have a comment on `model.timescale_eval`

pybamm/solvers/solution.py
TomTranter: I've changed this

pybamm/models/base_model.py
TomTranter: There were some tests that failed without adding this that weren't battery models and didn't have a solver. I agree though it's a bit dodgy and is now solved by checking for the attribute

pybamm/solvers/base_solver.py
TomTranter: Re-worded

tinosulzer (approved):
Cool, looks good to me now

TomTranter:
@tinosulzer @tlestang - the script now runs. It is a bit hacky but I think the way this example was written was already hacky copying data into a solution from an external model - might want to think about supporting this better with a nice wrapper. Also it only failed on Linux but the error was also there on windows so there is some inconsistency between which things get tested on various platforms

examples/scripts/compare_comsol/compare_comsol_DFN.py
```diff
@@ -140,9 +140,11 @@ def myinterp(t):
 
 # Make new solution with same t and y
 comsol_solution = pybamm.Solution(pybamm_solution.t, pybamm_solution.y)
-# Update model scales to match the pybamm model
+# Update model and solution scales to match the pybamm model
 comsol_model.timescale = pybamm_model.timescale
 comsol_model.length_scales = pybamm_model.length_scales
```
tinosulzer, line 145: you probably don't need these two lines any more now that the solution contains the timescale

tinosulzer (approved):
Yeah no problem, agreed it was already hacky

examples/scripts/compare_comsol/compare_comsol_DFN.py
TomTranter: I've removed the unnecessary lines now

TomTranter:
@tinosulzer just mac failing on this one now

PR #1214, opened 2020-10-26
Workaround for macos brew update error

brosaplanella:
# Description

This workaround allows the macos tests to work until the issue with GitHub actions. This change can be reverted once [this PR](https://github.com/actions/virtual-environments/pull/1808) is merged.

PR #1216, opened 2020-10-27
Apply workaround for brew update in publishing workflow

tlestang:
PR #1214 only modified the test workflow, but the `brew update` issue is also affecting the publishing workflow in the same way.
This is basically a duplicate of #1214


brosaplanella (approved):
Good catch, thanks Thibault!

PR #1222, opened 2020-10-30
Issue 1220 optimizations

### LINKED ISSUE #1220: Make `Symbol.is_constant` and `Symbol.shape` more efficient
These are bottlenecks in the generation of python or julia code.

`is_constant` should behave more similarly to `evaluates_on_edges`, and cache the value once it has calculated it the first time. Looking for symbols of a certain type is inefficient. `evaluates_on_edges` should also cache.

`shape` and `size` should cache the value once they have calculated it the first time.

tinosulzer:
# Description

Various optimizations based on running some profiling
- Cache shape and size calculations
- Rewrite `is_constant`
- Only create one instance of each of geometric, electrical, and thermal parameters
- Some small checks to avoid unnecessary recomputations wherever possible

Highlighted effects:
- 4x speedup on creating the DFN model object (0.4s -> 0.1s)
- The casadi solver with "fast" mode is now much faster on the second run, since it avoids calculating sizes and re-creating the integrator (around 2-3x speedup, but this depends on the grid size and parameters and model)

Fixes #1220 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Optimizations
 
+-   Only instantiate the geometric, electrical and thermal parameter classes once
 
 ## Bug fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,10 @@
 
 ## Optimizations
 
--   Only instantiate the geometric, electrical and thermal parameter classes once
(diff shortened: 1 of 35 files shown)


brosaplanella (approved):
Looks good, thanks Tino! Happy to merge when coverage has been fixed.

PR #1223, opened 2020-10-31
Issue 1219 smooth approximations

### LINKED ISSUE #1219: Look into using smooth approximations for min, max, heaviside
Functions with discontinuous derivatives can cause problems for adaptive time-stepping algorithms: https://discourse.julialang.org/t/handling-instability-when-solving-ode-problems/9019/5

Need to look into whether this is actually a problem for the solvers in pybamm. I suspect it might be given that linear interpolants have been giving similar issues.

If so, we could add a setting where pybamm automatically provides the smooth form instead of the exact form. e.g. by setting `pybamm.settings.smooth_approximation_steepness`, which is `k` in the link above, and setting `k` to zero could give the exact form.

tinosulzer:
# Description

Add smooth versions of the heaviside, minimum, and maximum operators, which can all have discontinuous values and/or derivatives. This clearly improves the performance of the solver, when applicable. A new setting `min_max_heaviside_smoothing_parameter` can be set to choose the degree of smoothing (or "exact" for no smoothing).

Fixes #1219 

## Type of change

# Key checklist:

## Further checks:


CHANGELOG.md
```diff
@@ -2,13 +2,17 @@
...
+-   Improve string printing of `BinaryOperator`, `Function`, and `Concatenation` objects ([#1223](https://github.com/pybamm-team/PyBaMM/pull/1223))
+-   Added `Solution.integration_time`, which is the time taken just by the integration subroutine, without extra setups ([#1223](https://github.com/pybamm-team/PyBaMM/pull/1223))
 -   Added parameter set for an A123 LFP cell ([#1209](https://github.com/pybamm-team/PyBaMM/pull/1209))
 -   Added variables related to equivalent circuit models ([#1204](https://github.com/pybamm-team/PyBaMM/pull/1204))
 -   Added an example script to check conservation of lithium ([#1186](https://github.com/pybamm-team/PyBaMM/pull/1186))
 -   Added `erf` and `erfc` functions ([#1184](https://github.com/pybamm-team/PyBaMM/pull/1184))
 
 ## Optimizations
 
+-   Add (optional) smooth approximations for the `minimum`, `maximum`, `Heaviside`, and `AbsoluteValue` operators ([#1223](https://github.com/pybamm-team/PyBaMM/pull/1223))
```
brosaplanella, line 15: Very minor thing: should `minimum` and `maximum` be capitalised or is it because the smooth approximation is for the method but not for the class?

examples/scripts/DFN.py
```diff
@@ -9,7 +9,7 @@
 
 
 # load model
-model = pybamm.lithium_ion.DFN()
+model = pybamm.lithium_ion.SPM()
```
brosaplanella, line 12: This should be DFN

brosaplanella (changes requested):
Looks good, only two minor comments (the one on the CHANGELOG might not require any action).

brosaplanella (approved):
Looks great, thanks Tino!

PR #1227, opened 2020-11-02
Issue 1225 slow sei model

### LINKED ISSUE #1225: Model with EC SEI submodel takes long time to build
There is a parameter missing, so it is calling our (slow) implementation of the levenshtein ratio. Either speed up levenshtein ratio, or write the EC submodel in such a way that it does not need to call that (e.g. bypass the levenshtein ratio, which is only used for a nicer error message)

tinosulzer:
# Description

Avoid try-except block which was causing the EC SEI submodel to slow down because of the levenshtein ratio search

Fixes #1225 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 ## Bug fixes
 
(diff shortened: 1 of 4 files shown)


brosaplanella (approved):
Looks good, thanks Tino!

PR #1231, opened 2020-11-04
Revert workaround macos errors

brosaplanella:
# Description

The issue with `brew update` on GitHub Actions has been fixed so the workaround can be reverted. It works fine with the PR workflow, I assume it will be fine with the publish workflow.


tinosulzer (approved):
On top of it! Thanks ferran

PR #1234, opened 2020-11-06
make quickplot use scaling from solutions

TomTranter:
# Description

Quickplot now uses timescale and lengthscale evals from the solution instead of evaluating them.

Fixes # 1233

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 -   Raise error if the boundary condition at the origin in a spherical domain is other than no-flux ([#1175](https://github.com/pybamm-team/PyBaMM/pull/1175))
 -   Fix boundary conditions at r = 0 for Creating Models notebooks ([#1173](https://github.com/pybamm-team/PyBaMM/pull/1173))
 -   Make sure simulation solves when evaluated timescale is a function of an input parameter ([#1218](https://github.com/pybamm-team/PyBaMM/pull/1218))
+-   Quickplot now works when timescale or lengthscale is a function of an input parameter ([#1234](https://github.com/pybamm-team/PyBaMM/pull/1234))
 
 ## Breaking changes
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -27,6 +27,7 @@
 -   Raise error if the boundary condition at the origin in a spherical domain is other than no-flux ([#1175](https://github.com/pybamm-team/PyBaMM/pull/1175))
 -   Fix boundary conditions at r = 0 for Creating Models notebooks ([#1173](https://github.com/pybamm-team/PyBaMM/pull/1173))
 -   Make sure simulation solves when evaluated timescale is a function of an input parameter ([#1218](https://github.com/pybamm-team/PyBaMM/pull/1218))
```
brosaplanella, line 29: Can you move these two lines of CHANGELOG so they are in descending order of PR?

brosaplanella (approved):
Looks good, just the minor point on the CHANGELOG. Thanks Tom!

CHANGELOG.md
TomTranter: Done

TomTranter:
@tinosulzer @rtimms some of the lines I changed don't actually seem to be used for anything. self.length_scales is defined but as far as I can tell never used anywhere. I commented them out and ran tests locally and they all passed. I also ran a 1+1 script and plotting the heat maps works ok without these lines. It does fail when the height is an input though. Not sure if worth fixing as I can't think of a scenario when  you'd want to have height as input when doing 1+1. Similarly I guess 2+1 will not like when height and width are inputs. I could add some more informative error messages

rtimms:
yeah it looks like the length scale stuff gets handled in processed variable. and I agree it doesn't make sense for height to be an input in 1+1D as you're changing the geometry. maybe in `battery_geometry` we should raise an error if any of the parameters are inputs since you can change the geometry on the fly

PR #1240, opened 2020-11-10
#1239 add python requires

### LINKED ISSUE #1239: Check python version in setup.py
When running setup.py, check that the python version is compatible (currently 3.6, 3.7, or 3.8)
Can be done via `python_requires`. See https://stackoverflow.com/questions/19534896/enforcing-python-version-in-setup-py

tinosulzer:
# Description

Add `python_requires` to `setup.py` to enforce using correct version of python

Fixes #1239 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -185,6 +185,8 @@ setup(
         "install": CustomInstall,
     },
     package_data={"pybamm": pybamm_data},
+    # Python version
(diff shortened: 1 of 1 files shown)


tlestang (approved):
Oh, good catch 👍

PR #1241, opened 2020-11-11
move length scale evaluation

tinosulzer:
# Description

Make sure solution timescale and lengthscale are updated when the model is assigned as an attribute, instead of just in the init

## Type of change

# Key checklist:

## Further checks:


TomTranter (approved):
I've tested the scripts that used inputs before which didn't work without the original change. They still work. The tests seem to pass now and the plot scales look ok but I don't know if all or a specific set were not working

tinosulzer:
Yeah I think this fixes it, will merge once tests pass

PR #1242, opened 2020-11-12
Issue 1235 jax sparse mat vec

### LINKED ISSUE #1235: sparse mat-vec support for EvaluatorJax
**Summary**

currently `pybamm.EvaluatorJax` does not support sparse matrices, because jax does not support them. It just replaces all sparse mats in expression tree with dense ones. However, you can implement a sparse mat-vec multiplication in jax easily enough (see https://gist.github.com/romanodev/34a471e8914989dbedee85febdbe7c77), so I propose replacing all sparse mat-vec mults with this, should be a lot faster.

 **Motivation**

a sparse mat-vec mult will be much faster than the equivalent dense mat-vec

martinjrobins:
# Description

`EvaluatorJax` uses a custom COO sparse matrix class and implements mat-vec multiply, as an alternative to converting everything to dense matrices

Fixes #1235 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/expression_tree/operations/evaluate.py
+++ b/pybamm/expression_tree/operations/evaluate.py
@@ -18,6 +18,25 @@ if system() != "Windows":
     config.update("jax_enable_x64", True)
 
     class JaxCooMatrix:
+        """
+        A sparse matrix in COO format, with internal arrays using jax device arrays
+
+        This matrix only has two operations supported, a multiply with a scalar, and a
+        dot product with a dense vector. It can also be converted to a dense 2D jax
+        device array
+
+        Parameters
+        ----------
+
+        row: arraylike
+            1D array holding row indices of non-zero entries
+        col: arraylike
+            1D array holding col indices of non-zero entries
+        data: arraylike
+            1D array holding non-zero entries
+        shape: 2-element tuple (x, y)
+            where x is the number of rows, and y the number of columns of the matrix
+        """
         def __init__(self, row, col, data, shape):
             self.row = jax.numpy.array(row)
             self.col = jax.numpy.array(col)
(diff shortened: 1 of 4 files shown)


pybamm/expression_tree/operations/evaluate.py
```diff
@@ -17,6 +17,61 @@
...
+
+        @jax.partial(jax.jit, static_argnums=(0,))
+        def multiply(self, b):
+            raise NotImplementedError
+
+        @jax.partial(jax.jit, static_argnums=(0,))
+        def __matmul__(self, b):
+            return self.dot_product(b)
+
+    def createJaxCooMatrix(value):
```
tinosulzer, line 55: function names should be lowercase_with_underscores, not CamelCase

tinosulzer (changes requested):
Code looks good, can you add more tests to improve coverage?

martinjrobins:
while I was writing tests, I noticed that the `pybamm.Inner` operation just uses `multiply` or `*` for the `evaluate` depending on if its a dense or sparse vector/matrix . But this should be an inner product right, so this would be incorrect?

tinosulzer:
`Inner` might not be the best name for that operator. It is almost identical to an elementwise multiplication, but with an extra step in the [finite volume discretisation](https://github.com/pybamm-team/PyBaMM/blob/2d577490bb38ede420b1c45975fc81a7e80c0d9f/pybamm/spatial_methods/finite_volume.py#L1103) which takes it from edges to nodes. This is applied to source terms for the thermal problem, which depend on (grad(phi))^2 - evaluates on edges - but need to evaluate on nodes.

Even if we don't rename, we could simplify the code by having `Inner` inherit from `Multiplication`

martinjrobins:
@tinosulzer, should be all good now. The only lines uncovered are those for windows-specific code, which I don't think we need to test, what do you think?

pybamm/expression_tree/operations/evaluate.py
```diff
@@ -17,6 +17,57 @@
 
     config.update("jax_enable_x64", True)
 
+    class JaxCooMatrix:
```
tinosulzer, line 20: should this class have a docstring (and corresponding rst file in the docs) or is it deep enough to never be user-facing?

tinosulzer (approved):
Thanks, looks good! Looking forward to trying this out with the models.
Re: windows coverage, that's fine, you can add `# pragma: no cover` at the end of an if or elseif line to tell it to ignore the entire branch

pybamm/expression_tree/operations/evaluate.py
martinjrobins: users should never have to use this. The only way they would actually get a JaxCooMatrix is if they define an expression to return a sparse matrix, which I don't think we need to consider?

martinjrobins: however, I really should add docstrings to this! which is now done

Issue #849, opened 2020-02-21
Parallel Processing with SharedMemory


TomTranter:
**Summary**
Doing many 1D models in parallel is a good way to solve my ECM problem and also parametric sweeps. 

 **Motivation**
The problem when you have hundreds or thousands of tasks becomes sending the data back and forth, especially if you step the solutions a little bit and make adjustments on the fly as I am doing. I think this can be addressed by using shared memory. I found a package that just works on linux (or maybe POSIX systems) which seems to give great performance.

The below script demonstrates stepping 1300 SPMs with slightly different applied currents using shared arrays and a processpoolexecutor.
There is also a line commented out that evaluates the terminal voltage. This slows things down a lot and highlights that evals are in need of vectorization

**Additional context**
 ```
import pybamm
import numpy as np
import matplotlib.pyplot as plt
from concurrent.futures import ProcessPoolExecutor
import SharedArray
import time as time_module

plt.close('all')
pybamm.set_logging_level("WARNING")

def current_function(t):
    return pybamm.InputParameter("Current")

if __name__ == '__main__':
    # load model
    model = pybamm.lithium_ion.SPMe()
    
    # create geometry
    geometry = model.default_geometry
    
    # load parameter values and process model and geometry
    param = model.default_parameter_values
    
    param.update({"Current function [A]": current_function,})
    param.update({"Current": "[input]"}, check_already_exists=False)
    param.process_model(model)
    param.process_geometry(geometry)
    
    # set mesh
    mesh = pybamm.Mesh(geometry, model.default_submesh_types, model.default_var_pts)
    
    # discretise model
    disc = pybamm.Discretisation(mesh, model.default_spatial_methods)
    disc.process_model(model)
    inputs = {"Current": 0.67}
    # solve model
    t_eval = np.linspace(0, 1, 2)
    solver = pybamm.CasadiSolver()
    sol_init = solver.solve(model, t_eval, inputs=inputs)
    i_app = 1.0
    
    Nspm = 1300
    Nsteps = 10
    try:
        shm_y = SharedArray.create('shm://y', [sol_init.y.shape[0], Nspm], dtype=float)
        shm_t = SharedArray.create('shm://t', [Nspm], dtype=float)
        shm_i_app = SharedArray.create('shm://i_app', [Nspm], dtype=float)
        shm_V = SharedArray.create('shm://V', [Nsteps, Nspm], dtype=float)
    except:
        SharedArray.delete('shm://y')
[...]

PR #1261, opened 2020-11-27
Solving for an ensemble of input parameters in parallel

tlestang:
# Description

This makes it possible for `base_solver.solve()` to solve for an ensemble of inputs in parallel, using multiprocessing.
This is inspired by some work by @TomTranter, see [#849](https://github.com/pybamm-team/PyBaMM/issues/849). This feature could be useful for solving models that can be decomposed into a ensemble of smaller models, and parameter sweeping.

This PR shouldn't break the API, and `solve` can still be used by passing a single `inputs` dictionary, returning an instance of `pybamm.solution`.  In addition, it is now possible to pass a list of dictionaries, representing an ensemble of, say, `N` set of input parameters to solve for. In this case the `N` calls to `base_solver._integrate` are spread across a pool of Python processes, leveraging multiple processors if available.  By default the number of processes is the number of cpu cores available (returned by `os.cpu_count()`), but this can be specified using the `nproc` keyword argument of `solve`.

These changes do not introduce significant change in terms of single-processor performance when passing a single dictionary to `solve`.  When passing multiple sets of inputs parameters, the speed up due to parallel processing can be visualized in the following weak scaling plots (**N** denotes the length of the list of input sets to process):

**Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz (6 cores)**
image

**Intel(R) Core(TM) i5-3470 CPU @ 3.20GHz (4 cores)** 
image

I suppose the slower processing time could explain the better scaling in the second case, but I don't have a definitive conclusion yet.

**Overall**, this is a first step towards parallel processing of an ensemble of models.  Changes in the `base_solver` class aim at modifying as little code as possible, so as to not entangle performance and software structure and make review of this PR easier. Once we're happy with the performance, we could probably think of some refactoring for this module. **On aspect that remains to be addressed before this PR is merged is testing**, and suggestions on how to best do this are welcome!

In terms of performance, my guess is that the main limiter is passing `self._integrate` to `Pool.starmap`. This causes the whole solver instance to be pickled/unpickled along with the function arguments. This certainly hurts performance, but I can't tell by how much. The pickling of the whole object could be avoided by moving the call to `starmap` inside the concrete solver classes (for instance `scipysolver`) so as to avoid passing an instance method to `starmap`.  [Read more on how passing instance methods to Pool.map() can hurt performance](https://thelaziestprogrammer.com/python/a-multiprocessing-pool-pickle).  But, again, that's for another PR.

Fixes # (issue)
#849 
## Type of change

# Key checklist:

## Further checks:


pybamm/solvers/base_solver.py
```diff
@@ -555,14 +562,23 @@ def solve(self, model, t_eval=None, external_variables=None, inputs=None):
...
+            for inputs in inputs_list
+        ]
 
         # Set up
         timer = pybamm.Timer()
 
         # Set up (if not done already)
         if model not in self.models_set_up:
-            self.set_up(model, ext_and_inputs, t_eval)
+            self.set_up(model, ext_and_inputs_list[0], t_eval)
```
tlestang, line 581: Assuming we models don't need to be set up for every value of the inputs, we just set it up using the first value.

pybamm/solvers/base_solver.py
```diff
@@ -573,22 +589,22 @@ def solve(self, model, t_eval=None, external_variables=None, inputs=None):
...
-        self._set_initial_conditions(model, ext_and_inputs, update_rhs=True)
+        self._set_initial_conditions(model, ext_and_inputs_list[0], update_rhs=True)
 
         # Non-dimensionalise time
         t_eval_dimensionless = t_eval / model.timescale_eval
 
         # Calculate discontinuities
         discontinuities = [
-            event.expression.evaluate(inputs=inputs)
+            event.expression.evaluate(inputs=inputs_list[0])
```
tlestang, line 607: Generally, the whole preprocessing (setup + calculation of discontinuities) is made using the first element of the `inputs` list given as an argument to solve.

Maybe a comment explaining why only the first element of both `inputs_list` in `ext_and_inputs_list` could help

pybamm/solvers/base_solver.py
```diff
@@ -640,73 +663,98 @@ def solve(self, model, t_eval=None, external_variables=None, inputs=None):
...
                 )
             )
-            new_solution = self._integrate(
-                model, t_eval_dimensionless[start_index:end_index], ext_and_inputs
-            )
-            new_solution.solve_time = timer.time()
-            if solution is None:
-                solution = new_solution
+            ninputs = len(ext_and_inputs_list)
+            if ninputs == 1:
```
tlestang, line 675: This is the important bit where we solve one model if `inputs` was a dictionary or an ensemble of models using multiprocessing if it was a list of dicts

pybamm/solvers/base_solver.py
```diff
@@ -613,6 +629,13 @@ def solve(self, model, t_eval=None, external_variables=None, inputs=None):
             pybamm.logger.info(
                 "Discontinuity events found at t = {}".format(discontinuities)
             )
+            if isinstance(inputs, list):
```
tlestang, line 632: For now, it is not possible to solve for several sets of inputs parameters and have discontinuities.
this is because discontinuities where computed only once (using the first element of `inputs_list`) and setting the restart state after a discontinuity isn't obvious since `model.y0` only contains a single state (see below)

pybamm/solvers/base_solver.py
```diff
@@ -640,73 +663,98 @@ def solve(self, model, t_eval=None, external_variables=None, inputs=None):
...
+                solutions = [sol for sol in new_solutions]
+            else:
+                for i, new_solution in enumerate(new_solutions):
+                    solutions[i].append(new_solution, start_index=0)
 
-            if solution.termination != "final time":
+            if solutions[0].termination != "final time":
                 break
 
             if end_index != len(t_eval_dimensionless):
```
tlestang, line 707: Again, the content of this `if` statement assumes that, if there are discontinuities, the `inputs` was a single dictionary, and therefore we're fine with just using `solutions[0]` and `ext_and_inputs_list[0]` to compute the restart state.

pybamm/solvers/base_solver.py
```diff
@@ -640,73 +663,98 @@ def solve(self, model, t_eval=None, external_variables=None, inputs=None):
...
+                        model, t_eval_dimensionless[end_index], ext_and_inputs_list[0]
                     )
 
-        # Assign times
-        solution.set_up_time = set_up_time
-        solution.solve_time = timer.time()
+        solve_time = timer.time()
+        for i, solution in enumerate(solutions):
+            # Assign times
+            solution.set_up_time = set_up_time
```
tlestang, line 721: All the solutions get the same times

tinosulzer (commented):
Thanks @tlestang , looks good! For the testing, what happens if you give it a test with multiple inputs?

tlestang:
For the testing, I added a new test `test_model_solver_with_multiple_inputs` in `test_scipy_solver`. it mimicks `test_model_solver_inputs` but solves the model in parallel over a range of inputs (using 4 processes).

pybamm/solvers/base_solver.py
```diff
@@ -573,22 +600,30 @@ def solve(self, model, t_eval=None, external_variables=None, inputs=None):
...
         set_up_time = timer.time()
         timer.reset()
 
         # (Re-)calculate consistent initial conditions
-        self._set_initial_conditions(model, ext_and_inputs, update_rhs=True)
+        # Assuming initial conditions do not depend on input parameters
+        # when len(inputs_list) > 1, only `ext_and_inputs_list[0]`
+        # is passed to `_set_initial_conditions`.
+        # See https://github.com/pybamm-team/PyBaMM/pull/1261
+        self._set_initial_conditions(model, ext_and_inputs_list[0], update_rhs=True)
```
tinosulzer, line 615: There should be an error check for this. You can check what input parameters are in `model.concatenated_initial_conditions.pre_order()` and make sure none of them appear in ext_and_inputs_list.

pybamm/solvers/base_solver.py
```diff
@@ -640,77 +680,102 @@ def solve(self, model, t_eval=None, external_variables=None, inputs=None):
...
+            if ninputs == 1:
+                new_solution = self._integrate(
+                    model,
+                    t_eval_dimensionless[start_index:end_index],
+                    ext_and_inputs_list[0],
+                )
+                new_solutions = [new_solution]
             else:
-                solution.append(new_solution, start_index=0)
+                with mp.Pool(processes=nproc) as p:
```
tinosulzer, line 700: does using `Pool` with `nproc=1` come with any performance losses? If so, should we just have a simple for loop in that case?

tinosulzer (approved):
Thanks @tlestang , looks good to me!

pybamm/solvers/base_solver.py
tlestang: I actually compared solving an ensemble of models using either a loop over `ìnputs_list` or a call to `map`.

| implementation | average(s) | std(s) |
|:--------------------:|:---------------:|:-------:|
| loop           |      0.174 |  0.007 |
| starmap        |      0.175 |  0.006 |

Stats over 32 models.

however I used the built-in `map` at the time... so not sure about the overhead induced by creating an extra process.

pybamm/solvers/base_solver.py
tlestang: Done!

tlestang:
Switching to draft, waiting for #1298 to be merged in this branch.

PR #1262, opened 2020-11-27
Issue 1245 active material

### LINKED ISSUE #1245: Increase in a active material during aging
I tried running cycling_ageing_yang.py for 50 cycles. I plotted the following quantities:
        "X-averaged negative electrode active material volume fraction",
        "X-averaged separator active material volume fraction",
        "X-averaged positive electrode active material volume fraction",
Each plot gave an increasing trend. AS per literature, with an increase in the number of cycles, the active material reduces which is one of the causes for degradation of battery or capacity fading.
All other values like terminal voltage and the dicharge capacity looked fine apart from the increase in the active material.

**To Reproduce**
Run the cycling_ageing_yang.py by changing the number "*2" to "*50 "in line 15 and run it. Plot the above quantities and check for the trend.

**Expected behavior**
An increase in active material will be seen.

**Possible solution**
The capacity fade can be computed as a space-time integral of ionic flux. Basically fade is the function of side reaction flux. The capacity fade computed can be used to update the initial active material at each cycle.

tinosulzer:
# Description

Set up active material models ready to model loss of active material
Fixes #1245 

## Type of change

# Key checklist:

## Further checks:


pybamm/parameters/lithium_ion_parameters.py
```diff
@@ -368,31 +376,19 @@ def R_p_dimensional(self, x):
...
-        through-cell distance
-        """
-        inputs = {"Through-cell distance (x_n) [m]": x}
-        if self.options["particle shape"] == "spherical":
-            return 3 * self.epsilon_s_n / self.R_n_dimensional(x)
-        elif self.options["particle shape"] == "user":
-            return pybamm.FunctionParameter(
-                "Negative electrode surface area to volume ratio [m-1]", inputs
-            )
+    def epsilon_s_n(self, x):
```
rtimms, line 379: If you set a model for LAM then the variable corresponding to `epsislon_s_n` and the parameter `epsilon_s_n` will be different (I think?) - the latter will just give the initial parameter value/function provided. Will this cause confusion? What you really provide as a parameter is the initial active material volume fraction, but I think it would be messy to append "Initial" everywhere. The same is true for porosity. 

I don't think this needs changing here, but something to think about in the future. I think my opinion is to leave it as is, and that it is understood that if you have a model describing how a parameter changes then that parameter should be treated as a variable.

rtimms (approved):
thanks @tinosulzer !

pybamm/parameters/lithium_ion_parameters.py
tinosulzer: Yes, this is something we should mention when we put together the workflow for degradation modeling. I agree putting "Initial" into the parameter names would be messy

weilongai:
Hi, I have made a los of active material submodel in pybamm. Would you mind if I push it here?

tinosulzer:
Yes, go for it! This branch has now been merged though so you'll need to open a PR to develop

weilongai:
I see there is a place holder already, but my implementation is not based on this. How do you think is the best way to do? Is it good to push my current version (tracing back to the last commit on Nov 17, 2020) to develop?

PR #1267, opened 2020-12-03
Issue 1266 add enertech data

### LINKED ISSUE #1266: Enertech cell data not included in the release
Do we need to edit MANIFEST.in? Are there other things we forgot to add to this @tlestang ?
Screen Shot 2020-12-02 at 9 22 44 PM

tlestang:
# Description

Modify the `setup.py` to include `.txt` files as package data. It also removes the file `MANIFEST.in`, which I think isn't needed anymore.

Fixes #1266


tinosulzer (approved):
Thanks @tlestang . Can you also cherry-pick this to the release-0.3.0 branch and make a 0.3.0.post1 release?

tlestang:
Will do now. I'm surprised why this wasn't caught by the tests?

tinosulzer:
The tests use the local install, this is an issue with the pip install. Maybe we should add a test for the pip install (e.g. to be run when a PR to master is made?)

PR #1273, opened 2020-12-04
[>master] Switch to manylinux2014

tlestang:
Building of manylinux wheels inside the manylinux2010 image breaks because the package repos are off following CentOS 6 reaching its end of life on November, 30th. See https://github.com/pypa/manylinux/issues/836

This updates the Dockerfile, to use the manylinux2014 image instead, based on CentOS 7.


tinosulzer (approved):
thanks!

PR #1274, opened 2020-12-04
Switch to manylinux2014

tlestang:
Building of manylinux wheels inside the manylinux2010 image breaks because the package repos are off following CentOS 6 reaching its end of life on November, 30th. See pypa/manylinux#836

This updates the Dockerfile, to use the manylinux2014 image instead, based on CentOS 7.


tinosulzer (approved):
thanks @tlestang !

tlestang:
Just realised I closed this without merging!

Issue #1265, opened 2020-12-02
Unable to Update Initial Conditions for Simulation


alibh95:
I am trying to simulate a driving cycle followed by CCCV charging.  Driving cycles are simulated by creating a current function parameter. And CCCV charging is simulated using Experiment class. To do so, I am solving the simulation with current function and trying to concatenate initial conditions to the model  and again starting the simulation with the Experiment for CCCV Charging. 

**Code**
`simulation1 = pybamm.Simulation(model, parameter_values=param, solver=solver)`
`simulation1.solve() #Driving Cycle`
`model.concatenated_initial_conditions = pybamm.Vector(solution1.y[:, -1][:, np.newaxis])`
`simulation2= pybamm.Simulation(model, experiment=experiment, solver=pybamm.CasadiSolver())`
`simulation2.solve() #CCCV Charging`

**Issue**
Unable to Update Initial Conditions for Simulation using 
`model.concatenated_initial_conditions = pybamm.Vector(solution1.y[:, -1][:, np.newaxis])`
Both Simulations starts from default initial conditions and at time t=0.

**Expected behaviour**
simulation2 must start with the initial conditions and time at the end of the simulation1

tinosulzer:
Yeah, this should work but unfortunately doesn't right now because of the way `model.new_copy` is currently set up.
Keep an eye on:
#1205 for a fix to new_copy
#1194 for a workaround on updating initial conditions (this is probably the easiest fix)
#1193 to allow drive cycles directly in experiments so you don't need to re-initialize

Any one of these would fix your problem or provide a workaround, so I'm going to close this issue as duplicate. If you are willing to help that would be great, but unfortunately these aren't the most beginner-friendly issues.

ssingh-ipa:
Hello,

I cannot follow if the issue has been resolved or now. 
As mentioned above, in order to do the following:
"simulation2 must start with the initial conditions and time at the end of the simulation1", 
what is the method of doing this for now? Could you please give an example showing how to implement #1193, if that's the way for going forward with this? 

Thank you.

tinosulzer:
`simulation2.solve(starting_solution=simulation1.solution)`

ssingh-ipa:
"ValueError: starting_solution can only be provided if simulating an Experiment."

[...]

Issue #1193, opened 2020-10-17
Allow drive cycles in experiments


tinosulzer:
Add functionality to allow running a drive cycle as part of an experiment. For example:
```python3
pybamm.Experiment(["Run US06 for 1 hour"], functions={"US06": US06_function})
```
Currently, the way experiments are implemented require current/voltage/power to be constant (they are set up as `InputParameter` objects). So this could be challenging to implement, and would depend on #895 being possible.

maurosgroi:
Dear @tinosulzer,
I'm trying to use this functionality to import an extended WLTP cycle for HD vehicles lasting more than 2800s. The sampling time is 0.1s. 
Essentially I'm substuting the original csv file with mine but the solution is cut at about 770 s. 
Is there any limitation in the number of point that can be imported?
My csv file for reference is uploaded here:  https://drive.google.com/file/d/1MJBdq-g5RjednvSJ8e-LMkQnnxe1OWvj/view?usp=sharing

Thanks a lot in advance and best regards,
Mauro Sgroi.

tinosulzer:
This functionality is not implemented yet as described above. Can you show me the code you are using to load your drive cycle?

maurosgroi:
Here it is. I tried to reuse part of tutorial4.

import pybamm
import pandas as pd    # needed to read the csv data file
chemistry = pybamm.parameter_sets.Chen2020
parameter_values = pybamm.ParameterValues(chemistry=chemistry)
model = pybamm.lithium_ion.SPM()
drive_cycle = pd.read_csv("D:/WLTC.csv", comment="#", header=None).to_numpy()
timescale = parameter_values.evaluate(model.timescale)
current_interpolant = pybamm.Interpolant(drive_cycle, timescale * pybamm.t)

parameter_values["Current function [A]"] = current_interpolant
sim = pybamm.Simulation(model, parameter_values=parameter_values)
sim.solve(t_eval=None)
sim.plot(["Current [A]", "Terminal voltage [V]"], time_unit='seconds', variable_limits='tight')

Thanks a lot and best regards,
Mauro Sgroi.

maurosgroi:
Sorry, I misunderstood your initial post.
Anyway I cannot my cycle work.
Thanks a lot and best regards,
Mauro Sgroi.

tinosulzer:
Can you describe what you are trying to do? Would be good to know, in case it is easy to implement

maurosgroi:
[...]

PR #1278, opened 2020-12-07
Issue 1194 initial conditions

tinosulzer:
# Description

Add a function to initialize a model with a different model, fixes #1194 
Reformat model.new_copy so that any changes to the model are kept, fixes #1205 

## Type of change

Please add a line in the relevant section of [CHANGELOG.md](https://github.com/pybamm-team/PyBaMM/blob/master/CHANGELOG.md) to document the change (include PR #) - note reverse order of PR #s. If necessary, also add to the list of breaking changes.

- [x] New feature (non-breaking change which adds functionality)
- [ ] Optimization (back-end change that speeds up the code)
- [x] Bug fix (non-breaking change which fixes an issue)

# Key checklist:

- [ ] No style issues: `$ flake8`
- [ ] All tests pass: `$ python run-tests.py --unit`
- [ ] The documentation builds: `$ cd docs` and then `$ make clean; make html`

You can run all three at once, using `$ python run-tests.py --quick`.

## Further checks:

- [ ] Code is commented, particularly in hard-to-understand areas
- [ ] Tests added that prove fix is effective or that feature works

PR #1279, opened 2020-12-07
Drive Cycle in Experiment

alibh95:
# Description

Added the functionality to simulate drive cycle using the experiment class.

Fixes # (issue)
#1193 
#1265 

## Type of change

- [✓] New feature (non-breaking change which adds functionality)

# Key checklist:

- [✓] No style issues: `$ flake8`

## Further checks:

- [✓] Code is commented, particularly in hard-to-understand areas


pybamm/experiments/experiment.py
```diff
@@ -134,9 +143,18 @@ def read_string(self, cond):
             electric = self.convert_electric(cond_list[:idx])
             time = None
             events = self.convert_electric(cond_list[idx + 1 :])
+        elif "Run" in cond:
+            # e.g. Run US06
+            cond_list = cond.split()
+            electric = (drive_cycles[cond_list[1]][:, 1], "Drive")
+            # Set time and period to 1 second
+            # because drive cycles are sampled at 1Hz frequency
```
tinosulzer, line 151: This is not always true. For drive cycles, the time and period should be set to "None" and then the difference between times in the drive cycle itself should be used.

pybamm/simulation.py
```diff
@@ -211,39 +211,84 @@ def set_up_experiment(self, model, experiment):
...
+                    driving_inputs = {
+                        "Current switch": 1,
+                        "Voltage switch": 0,
+                        "Power switch": 0,
+                        "Current input [A]": I_drive,
+                        "Voltage input [V]": 0,  # doesn't matter
+                        "Power input [W]": 0,  # doesn't matter
+                    }
+                    # Update period
+                    driving_inputs["period"] = op[3]
```
tinosulzer, line 229: here period should also be a list, which is updated based on the times in the drive cycle

pybamm/simulation.py
```diff
@@ -417,38 +462,69 @@ def solve(
             self._solution = None
             # Step through all experimental conditions
             inputs = inputs or {}
-            pybamm.logger.info("Start running experiment")
```
tinosulzer, line 420: why was this removed?

tinosulzer (changes requested):
@alibh95 , thanks very much for taking this on! 

Can you add tests for the changes you have made?
After we merge #1278 , there will be a way to do this in a more efficient way, but this is a good first step.

pybamm/simulation.py
alibh95: It was removed by mistake. I will add this line in next commit.

tinosulzer:
@alibh95 I have an idea for how to do this slightly differently. Are you still working on it?

alibh95:
@tinosulzer Yes, I am working on this issue. Till now I am able to add the functionality of "for event" but unable to add the functionality of "until event".

PR #1280, opened 2020-12-08
Issue 1244 algebraic j

### LINKED ISSUE #1244: Algebraic equation for j
Investigate whether using an explicit algebraic equation for `j`, instead of making it a symbol function of other variables, makes a difference to the runtime and/or solution

tinosulzer:
# Description

Add an option to make the interfacial current density a state with an algebraic equation

Fixes #1244 

## Type of change

# Key checklist:

## Further checks:


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Features
 
+-   Added option to make the total interfacial current density a state
```
brosaplanella, line 5: Needs a link to this PR

CHANGELOG.md
```diff
@@ -13,6 +14,7 @@
 
 ## Breaking changes
 
+-   Boolean model options ('sei porosity change', 'convection') must now be given in string format ('true' or 'false' instead of True or False)
```
brosaplanella, line 17: Needs a link to this PR

pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -16,46 +16,47 @@ class BaseBatteryModel(pybamm.BaseModel):
...
                 (if thermal effects are included) solves a lumped thermal model
                 with prescribed surface area for cooling.
-            * "convection" : bool or str, optional
+            * "convection" : str
                 Whether to include the effects of convection in the model. Can be
-                False (default), "differential" or "algebraic". Must be 'False' for
-                lithium-ion models.
-            * "current collector" : str, optional
+                False (default), "uniform transverse" or "full transverse".
+                Must be "none" for lithium-ion models.
```
brosaplanella, line 31: Is `"none"` the same as `False` (default)? If yes, then they should be called the same.

pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -102,18 +103,24 @@ class BaseBatteryModel(pybamm.BaseModel):
...
-            * "sei porosity change" : bool
-                Whether to include porosity change due to SEI formation (default False)
-            * "side reactions" : list, optional
+            * "sei porosity change" : str
+                Whether to include porosity change due to SEI formation, can be "false"
+                (default) or "true".
+            * "side reactions" : list
                 Contains a list of any side reactions to include. Default is []. If this
                 list is not empty (i.e. side reactions are included in the model), then
                 "surface form" cannot be 'False'.
```
brosaplanella, line 112: Should be 
```
"surface form" cannot be "false"
```

pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -202,8 +209,8 @@ def options(self, extra_options):
         default_options = {
             "operating mode": "current",
             "dimensionality": 0,
-            "surface form": False,
-            "convection": False,
+            "surface form": "false",
+            "convection": "none",
```
brosaplanella, line 213: Linked to my previous comment, how come here we use `"none"` instead of `"false"`?

brosaplanella (approved):
Looks good, just some minor comments. Thanks Tino!

pybamm/models/full_battery_models/base_battery_model.py
tinosulzer: It is a bit arbitrary but I am basically using "false" if the other option is "true" and "none" otherwise, except for "surface form". "There is no convection" -> "convection: none" vs "We don't use the surface form" -> "surface form: false"

PR #1287, opened 2020-12-09
Lithium plating

DrSOKane:
# Description

- The reversible Li plating model of O'Kane et al. (2020) is included as a new submodel
- An irreversible plating model with similar physics is also included
- New parameters that are used by these submodels are defined
- A new variant of the existing Chen 2020 parameter set that includes these new parameters has been added
- The new parameter set also has temperature-dependent diffusivity in the anode
- The submodel accounts for SEI layer resistance
- The submodel does NOT account for temperature variation at this time, but this should be easy to fix

Fixes # 836

## Type of change

# Key checklist:

## Further checks:


pybamm/models/submodels/interface/li_plating/irreversible_plating.py
```diff
@@ -0,0 +1,79 @@
...
+
+        return variables
+
+    def get_coupled_variables(self, variables):
+        param = self.param
+        phi_s_n = variables[f"{self.domain} electrode potential"]
+        phi_e_n = variables[f"{self.domain} electrolyte potential"]
+        c_e_n = variables[f"{self.domain} electrolyte concentration"]
+        C_plating = param.C_plating
+        phi_ref = param.U_n_ref / param.potential_scale
```
tinosulzer, line 39: this seems a bit weird for the reference potential, need to double check

pybamm/models/submodels/interface/li_plating/irreversible_plating.py
```diff
@@ -0,0 +1,79 @@
...
+
+        if f"{self.domain} electrode sei film overpotential" in variables:
+            eta_sei = variables[f"{self.domain} electrode sei film overpotential"]
+        else:
+            eta_sei = pybamm.Scalar(0)
```
tinosulzer, line 44: Ideally the sei film overpotential should always be defined, sometimes to zero. Otherwise, if the submodels are called in the wrong order, the sei film overpotential might be non-zero but this would miss it

pybamm/models/submodels/interface/li_plating/irreversible_plating.py
```diff
@@ -0,0 +1,79 @@
...
+class IrreversiblePlating(BasePlating):
+    """Base class for irreversible Li plating.
+    Parameters
+    ----------
+    param : parameter class
+        The parameters to use for this submodel
+    domain : str
+        The domain of the model either 'Negative' or 'Positive'
+    **Extends:** :class:`pybamm.li_plating.BasePlating`
+    """
```
tinosulzer, line 17: add ref for the model

pybamm/models/submodels/interface/li_plating/irreversible_plating.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def __init__(self, param, domain):
+        super().__init__(param, domain)
+
```
tinosulzer, line 21: register citation for the model

pybamm/models/submodels/interface/li_plating/reversible_plating.py
```diff
@@ -0,0 +1,80 @@
...
+class ReversiblePlating(BasePlating):
+    """Base class for reversible Li plating.
+    Parameters
+    ----------
+    param : parameter class
+        The parameters to use for this submodel
+    domain : str
+        The domain of the model either 'Negative' or 'Positive'
+    **Extends:** :class:`pybamm.li_plating.BasePlating`
+    """
```
tinosulzer, line 17: add ref for the model

pybamm/models/submodels/interface/li_plating/reversible_plating.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def __init__(self, param, domain):
+        super().__init__(param, domain)
```
tinosulzer, line 20: register citation for the model

pybamm/models/submodels/interface/li_plating/irreversible_plating.py
```diff
@@ -0,0 +1,79 @@
...
+            variables.update(
+                self._get_standard_whole_cell_interfacial_current_variables(variables)
+            )
+
+        return variables
+
+    def set_rhs(self, variables):
+        c_plated_Li = variables[f"{self.domain} electrode Li plating concentration"]
+        j_stripping = variables[
+            f"{self.domain} electrode Li plating " f"interfacial current density"
```
tinosulzer, line 69: second f string not needed

examples/notebooks/lithium-plating.ipynb
```diff
@@ -0,0 +1,488 @@
...
+    "parameter_values = pybamm.ParameterValues(chemistry=chemistry)\n",
+    "parameter_values.update({\"Reference temperature [K]\": 268.15})\n",
+    "parameter_values.update({\"Ambient temperature [K]\": 268.15})\n",
+    "parameter_values.update({\"Initial temperature [K]\": 268.15})\n",
+    "parameter_values.update({\"Upper voltage cut-off [V]\": 4.21})\n",
+    "parameter_values.update({\"Li plating kinetic rate constant [m.s-1]\": 1E-9})\n",
+    "\n",
+    "# add plating submodels\n",
+    "model1.submodels[\"Negative Li plating\"] = pybamm.li_plating.ReversiblePlating(model1.param, \"Negative\")\n",
+    "model1.submodels[\"Positive Li plating\"] = pybamm.li_plating.NoPlating(model1.param, \"Positive\")\n",
```
tinosulzer, line 54: need to add these to the model options, etc

tinosulzer (commented):
Thanks @DrSOKane , this looks almost ready already! 
Just a few minor comments. I'd also like to double-check the non-dimensionalization.

Am I right in reading that the "irreversible" and "reversible" plating models are identical except for the form of `j_stripping`? If so, we could remove some repeated code by passing `options` to the submodel and then using the reversible or irreversible form based on that.

Also, the plating options should be added to `BaseModel.options` and there should be a corresponding `set_li_plating_models` function

pybamm/models/submodels/interface/li_plating/irreversible_plating.py
brosaplanella: That is what you get if you set the dimensional electrolyte potential to zero (I got the same in the half-cell model).

pybamm/parameters/parameter_sets.py
```diff
@@ -113,6 +113,18 @@
...
+Chen2020_plating = {
+    "chemistry": "lithium-ion",
+    "cell": "LGM50_Chen2020",
+    "anode": "graphite_Chen2020_plating",
+    "separator": "separator_Chen2020",
+    "cathode": "nmc_Chen2020",
+    "electrolyte": "lipf6_Nyman2008",
+    "experiment": "1C_discharge_from_full_Chen2020",
+    "sei": "example",
+    "citation": "Chen2020",
```
brosaplanella, line 125: Apart from `Chen2020` you should also add the reference for the plating parameters.

pybamm/input/parameters/lithium-ion/anodes/graphite_Chen2020_plating/README.md
```diff
@@ -0,0 +1,7 @@
+# LG M50 Graphite anode parameters
+
+Parameters for a LG M50 graphite anode, from the paper
+
+> Chang-Hui Chen, Ferran Brosa Planella, Kieran O’Regan, Dominika Gastol, W. Dhammika Widanage, and Emma Kendrick. ["Development of Experimental Techniques for Parameterization of Multi-scale Lithium-ion Battery Models."](https://iopscience.iop.org/article/10.1149/1945-7111/ab9050) Journal of the Electrochemical Society 167 (2020): 080534
+
+and references therein.
```
brosaplanella, line 7: Add the reference you got the plating parameters from

pybamm/input/parameters/lithium-ion/anodes/graphite_Chen2020_plating/parameters.csv
```diff
@@ -0,0 +1,36 @@
...
+# Density,,,
+Negative electrode density [kg.m-3],1657,default,
+,,,
+# Thermal parameters,,,
+Negative electrode specific heat capacity [J.kg-1.K-1],700,default,
+Negative electrode thermal conductivity [W.m-1.K-1],1.7,default,
+Negative electrode OCP entropic change [V.K-1],0,,
+,,,
+# Li plating parameters,,,
+Li metal partial molar volume [m3.mol-1],1.3E-5,,
```
brosaplanella, line 34: For the plating parameters, add the reference where you got them from in the third column (like in the Negative electrode charge transfer coefficient, for example).

pybamm/input/parameters/lithium-ion/anodes/graphite_Chen2020_plating/README.md
brosaplanella: General comment (doesn't apply to this PR only): we might need to reconsider how we structure the parameters as the way it is now we repeat ourselves which makes it easier for inconsistencies to appear. 

I would say for the moment is fine, but now that we have several different types of models (SEI, plating, mechanics...) it would be good to think if we can improve. Maybe that is a topic for the new Discussion channel?

tests/unit/test_models/test_submodels/test_interface/test_li_plating/test_irreversible_plating.py
```diff
@@ -0,0 +1,43 @@
+#
+# Test reversible Li plating submodel
+#
+
+import pybamm
+import tests
+import unittest
+
+
+class TestReversiblePlating(unittest.TestCase):
```
brosaplanella, line 10: I think you have a typo here and you want to change all the `ReversiblePlating` by `IrreversiblePlating`. This would explain why the coverage of irreversible plating is so poor even with these tests.

tests/unit/test_models/test_submodels/test_interface/test_li_plating/test_reversible_plating.py
```diff
@@ -0,0 +1,44 @@
+#
+# Test reversible Li plating submodel
+#
+
+import pybamm
+import tests
+import unittest
+
+
+class TestReversiblePlating(unittest.TestCase):
```
brosaplanella, line 10: To get the coverage of reversible plating to 100% you need to add a couple of tests that cover:

lines 44-45:
```python3
else:
    eta_sei = pybamm.Scalar(0)
```

lines 56-63:
```python3
if (
    "Negative electrode Li plating interfacial current density" in variables
    and "Positive electrode Li plating interfacial current density" in variables
    and "Li plating interfacial current density" not in variables
 ):
    variables.update(
        self._get_standard_whole_cell_interfacial_current_variables(variables)
    )
```

You might need to do the same for the irreversible tests.

brosaplanella (changes requested):
Hi Simon, that looks very nice! 😃 Just some minor comments.

Apart from the inline comments, you should also add tests for `no_plating.py` and `base_plating.py` to improve the coverage.

pybamm/input/parameters/lithium-ion/anodes/graphite_Chen2020_plating/README.md
rtimms: can we just append the SEI parameters to the existing set and update the references? or do some of the other parameters change? For the functions that change can they just live in the Chen folder with a new name, and then users manually update that parameter?

DrSOKane:
Hi all, thanks for your suggestions. I agree we need more tests, more citations and to merge the reversible and irreversible submodels since they are mostly the same.

I've also found that the Li plating model does not converge when SEI is present in large amounts. This is the case even when the code adding eta_sei is commented out. A small amount of SEI do not cause this and produce a reasonable result even after 1000 cycles.

tinosulzer:
Is this with constant SEI or varying?

DrSOKane:
It works with constant SEI but fails with varying.

tinosulzer:
If I understand correctly, when you have both SEI and plating, the model is basically trying to solve an equation of the form
`j(phi_s - phi_e) + j_sei(phi_s - phi_e) + j_pl(phi_s - phi_e) = I`
for `phi_s - phi_e`, where each `j` is an exponential or sinh.  I can see this being numerically challenging. I suggest we deal with it separately from this PR.

DrSOKane:
Two points:
1) It is true that the two models in this release are identical save for the expression for j_stripping. However, I also have an unpublished model that differs more substantially. My plan is to keep the two separate submodels but add an option to allow the user to easily pick one.
2) I do not actually have a meaningful reference for the plating constant. It's true that my paper does provide a value for it, but that's not only for a different cell but for a more physically accurate nonlinear diffusion function, which I've been unable to implement in PyBaMM thus far.

tinosulzer:
Ok makes sense. Would be good to just get something merged (I think coverage is all that is needed) and then we can iterate on it

pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -42,6 +42,9 @@ class BaseBatteryModel(pybamm.BaseModel):
             * "interfacial surface area" : str, optional
                 Sets the model for the interfacial surface area. Can be "constant"
                 (default) or "varying". Not currently implemented in any of the models.
+            * "Li plating" : str, optional
```
tinosulzer, line 45: Very minor: I would prefer "lithium plating" to "Li plating"

pybamm/models/full_battery_models/lithium_ion/base_lithium_ion_model.py
```diff
@@ -85,6 +85,29 @@ def set_sei_submodel(self):
         # positive electrode
         self.submodels["positive sei"] = pybamm.sei.NoSEI(self.param, "Positive")
 
+    def set_li_plating_submodel(self):
```
tinosulzer, line 88: in general I don't think anything is lost by calling all these things `lithium_plating`, in fact makes it easier to read

pybamm/models/submodels/interface/li_plating/irreversible_plating.py
tinosulzer: this still needs to be done (unless you don't want to of course)

pybamm/models/submodels/interface/li_plating/reversible_plating.py
tinosulzer: still needs to be done

tinosulzer (approved):
Looks good to me, just consider using the full "lithium plating" instead of "Li plating" for legibility, and do you want to register your paper for citations?

pybamm/models/full_battery_models/lithium_ion/base_lithium_ion_model.py
```diff
@@ -85,6 +85,29 @@ def set_sei_submodel(self):
...
+
+        # negative electrode
+        if self.options["Li plating"] == "none":
+            self.submodels["negative Li plating"] = pybamm.sei.NoPlating(
```
tinosulzer, line 92: this should say `li_plating.NoPlating`

tests/unit/test_models/test_submodels/test_interface/test_li_plating/test_irreversible_plating.py
brosaplanella: This still needs to be done (minor fix)

brosaplanella (approved):
I agree with @tinosulzer comments. Once this and the additional comment on `test_irreversible_plating.py` are fixed we can merge. Thanks @DrSOKane!

brosaplanella:
Also, note that there are some conflicts with the main branch (in 'base_battery_model.py') that also need to be addressed before we can merge it.

DrSOKane:
I'm in the process of renaming everything to lithium_plating or "lithium plating". While I'm doing that, do you want me to add the stripping current to inverse_butler_volmer.py?

tinosulzer:
> do you want me to add the stripping current to inverse_butler_volmer.py?

What does this refer to again? In the `CurrentForInverseButlerVolmer` class?

DrSOKane:
I am referring to the CurrentForInverseButlerVolmer class, yes.

tinosulzer:
Yes that would be good to update too

DrSOKane:
Added plating/stripping current density to CurrentForInverseButlerVolmer and fixed remaining conflicts. I've also added my paper to CITATIONS.txt but not sure how to use the BibTeX entry now that it's there!

DrSOKane:
Hi all, I'm trying to test if examples/lithium-plating.ipynb still runs and I'm getting the pybtex error:

---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-1-992448a06289> in <module>
      1 get_ipython().run_line_magic('pip', 'install pybamm -q    # install PyBaMM if it is not installed')
----> 2 import pybamm
      3 import os
      4 import numpy as np
      5 import matplotlib.pyplot as plt

~/PyBaMM/pybamm/__init__.py in <module>
     70 from .logger import logger, set_logging_level
     71 from .settings import settings
---> 72 from .citations import Citations, citations, print_citations
     73 
     74 #

~/PyBaMM/pybamm/citations.py in <module>
      6 import pybamm
      7 import os
----> 8 import pybtex
      9 
     10 

ModuleNotFoundError: No module named 'pybtex'

tinosulzer:
You need to install pybtex, just do `pip install -e .` and it will do it for you

brosaplanella:
I think the example tests will still fail because `lithium_plating.ipynb` does not include `pybamm.print_citations()` at the end (this is a new test we added recently). You can copy the last block from another notebook (for example Tutorial 2 - Compare models.ipynb, don't copy the block in Tutorial 1 as it is slightly different).

examples/scripts/custom_model.py
```diff
@@ -59,6 +59,12 @@
 ] = pybamm.electrolyte_conductivity.LeadingOrder(model.param)
 model.submodels["negative sei"] = pybamm.sei.NoSEI(model.param, "Negative")
 model.submodels["positive sei"] = pybamm.sei.NoSEI(model.param, "Positive")
+model.submodels[
+    "negative lithium plating"
+] = pybamm.li_plating.NoPlating(model.param, "Negative")
```
brosaplanella, line 64: This should be `pybamm.lithium_plating` (same in the line below). I think it is what is breaking the example tests.

brosaplanella:
@all-contributors add @DrSOKane for code, examples, documentation and tests

brosaplanella:
Great work @DrSOKane, thank you very much!

PR #1294, opened 2020-12-11
#1251 fix function parameter diff bug

### LINKED ISSUE #1251: Diff of a ParameterFunction doesn't work when evaluated at a constant
**Describe the bug**
When defining a parameter as the `diff` of a `ParameterFunction` sometimes it returns 0. I think the issue is that the expression gets simplified first and, because the argument is a constant, we loose track of the differentiation variable. Then, when taking the derivative, the variable is not in the tree and therefore it returns 0.

**To Reproduce**
1. Add a `diff` parameter to your model, for example
```python3
            def Uprime(sto, T):
                inputs = {"Positive particle stoichiometry": sto}
                Uprime_ref = pybamm.FunctionParameter(
                    "Positive electrode OCP [V]", inputs, diff_variable=sto
                )
                return Uprime_ref  / pot
```
2. Set the positive OCV from a function (for example a linear function).
3. Evaluate at `c_0 = param.c_p_init(1)`, the function returns zero.
4. Evaluate at `c_0 + 0 * pybamm.t`, the function now returns the right value.

tinosulzer:
# Description

Fixes #1251 

## Type of change

# Key checklist:

## Further checks:


tinosulzer:
Coverage error is something that was already failing before ...

pybamm/models/submodels/interface/kinetics/base_kinetics.py
```diff
@@ -207,12 +207,13 @@ def _get_dj_ddeltaphi(self, variables):
         return j.diff(delta_phi)
 
     def _get_interface_variables_for_first_order(self, variables):
-        # This is a bit of a hack, but we need to multiply electrolyte concentration by
-        # one to differentiate it from the electrolyte concentration inside the
+        # This is a bit of a hack, but we need to wrap electrolyte concentration with
+        # the NotConstant class
+        # to differentiate it from the electrolyte concentration inside the
         # surface potential difference when taking j.diff(c_e) later on
         # use explicit Multiplication instead of * to avoid simplification
```
brosaplanella, line 214: Can this line be removed now?

brosaplanella (approved):
Looks good to me, thanks Tino!

Issue #734, opened 2019-11-15
Pickle python code strings


tinosulzer:
We can't pickle python code so can't save the model if `model.convert_to_format="python"`. See [#732 (comment)](https://github.com/pybamm-team/PyBaMM/pull/732#issuecomment-553765895) for workaround

tinosulzer:
Closing as duplicate of #1283 with less detail

PR #1298, opened 2020-12-14
Enable pickling of models in "python" format

### LINKED ISSUE #1283: Cannot pickle model when format is "python"
**Description**

Instances of `BaseModel` and derived classes cannot be pickled when attribute `convert_to_format` is `"python"`.  This hampers use of multiprocessing, since it relies on pickling objects for passing them between processes, see #1261 .

**To reproduce**

```python
import pickle
import pybamm

param.process_model(model)
# discretise model
disc.process_model(model)

model.convert_to_format = "python"
model.default_solver.set_up(model)

try:
    with open("model.pickle", mode="wb") as f:
	pickle.dump(model, f)
except pickle.PicklingError as e:
    print(e)
```

    Can't pickle <function evaluate at 0x7f1c4c5d3280>: attribute lookup evaluate on pybamm.expression_tree.operations.evaluate failed

**Details**

Function [`base_solver.set_up`](https://github.com/pybamm-team/PyBaMM/blob/9d42dfe69834c436f6aae40c756d2f8085df958d/pybamm/solvers/base_solver.py#L120) does several things but in particular is creates the `SolverCallable` objects the solver relies on to evaluate the RHS, initial conditions and jacobian.

```python
# base_solver.py

# Add the solver attributes
model.init_eval = init_eval # SolverCallable instance
model.rhs_eval = rhs_eval
model.algebraic_eval = algebraic_eval
model.jac_algebraic_eval = jac_algebraic
model.terminate_events_eval = terminate_events_eval
model.discontinuity_events_eval = discontinuity_events_eval

# Calculate initial conditions
model.y0 = init_eval(inputs)
```
When `model.convert_to_format == "python"`, `SolverCallable.__call__` is a wrapper around [`EvaluatorPython.evaluate()`](https://github.com/pybamm-team/PyBaMM/blob/9d42dfe69834c436f6aae40c756d2f8085df958d/pybamm/expression_tree/operations/evaluate.py#L494) which evaluates the Python code generated in `EvaluatorPython.__init__` to translate the expression tree. When an `EvaluatorPython` instance is created, a global function `evaluate` is defined by compiling its definition on the fly. This function is then pointed to by `EvaluatorPython._evaluate`.

According to the [docs for Pickle](https://docs.python.org/3.8/library/pickle.html#what-can-be-pickled-and-unpickled), only functions defined at the top level of a module can be pickled and *I think* this is the reason why `evaluate` cannot be pickled.

> Note that functions (built-in and user-defined) are pickled by
> “fully qualified” name reference, not by value. This means that
> only the function name is pickled, along with the name of the
> module the function is defined in.

Note that `evaluate` is a global function and not a local function in `EvaluatorPython.__init__`, *i.e* `print(self._evaluate)` evaluates to `<function evaluate at 0x7fa4fee843a0>` and not `<function EvaluatorPython.__init__.<locals>.evaluate 0x7fa4fee843a0>`.  However since `evaluate` is compiled on the fly, I don't think `Pickle` can find the name of the module it is defined in&#x2026;

Minimal example:

```python
import pickle

python_str = """def dummy_function():
    return "Hello"
self._method = dummy_function
"""

class myclass():
    def __init__(self):
	compiled_function = compile(python_str, "", "exec")
	exec(compiled_function)

inst = myclass()
with open("data.pickle", mode="w+b") as f:
    try:
	pickle.dump(inst, f)
    except pickle.PicklingError as e:
	print(e)
```

    Can't pickle <function dummy_function at 0x7fc39202ec10>: attribute lookup dummy_function on __main__ failed

Models can be pickled fine when `mode.convert_to_format == "casadi"`, since in this case `SolverCallable.function` calls an instance of `Casadi.Function`, which can pickled.

tlestang:
# Description

This defines the two methods `__getstate__` and `__setstate__` that can be used to control what is pickled when pickling a class
instance, as well as what happens when the object in unpickled.

In this case this is useful to avoid pickling the `_evaluate` attribute, which is a reference to a function that cannot be pickled.
This attribute is restored when unpicking by re-compiling the definition of the function `evaluate`. 

Note that this does _not_ make it possible to pickle model in "jax" format.

Fixes #1283 #734 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/expression_tree/operations/evaluate.py
+++ b/pybamm/expression_tree/operations/evaluate.py
@@ -485,6 +485,7 @@ class EvaluatorPython:
         python_str = python_str + "\nself._evaluate = evaluate"
 
         self._python_str = python_str
+        self._result_var = result_var
         self._symbol = symbol
 
         # compile and run the generated python code,
@@ -506,6 +507,23 @@ class EvaluatorPython:
             return result, known_evals
         else:
             return result
+
+    def __getstate__(self):
+        # Control the state of instances of EvaluatorPython
+        # before pickling. Method "_evaluate" cannot be pickled.
+        # See https://github.com/pybamm-team/PyBaMM/issues/1283
+        state = self.__dict__.copy()
+        del state["_evaluate"]
+        return state
+
+    def __setstate__(self, state):
+        # Restore pickled attributes and
+        # compile code from "python_str"
+        # Execution of bytecode (re)adds attribute
+        # "_method"
+        self.__dict__.update(state)
+        compiled_function = compile(self._python_str, self._result_var, "exec")
+        exec(compiled_function)
 
 
 class EvaluatorJax:
--- a/pybamm/solvers/base_solver.py
+++ b/pybamm/solvers/base_solver.py
@@ -545,12 +545,6 @@ class BaseSolver(object):
         """
         pybamm.logger.info("Start solving {} with {}".format(model.name, self.name))
 
-        # Cannot use multiprocessing with model in "jax" format
-        if(len(inputs) > 1) and model.convert_to_format == "jax":
-            raise pybamm.SolverError(
-                "Cannot solve list of inputs with multiprocessing "
-                "when model in format \"jax\"."
-            )
         # Make sure model isn't empty
         if len(model.rhs) == 0 and len(model.algebraic) == 0:
             if not isinstance(self, pybamm.DummySolver):
@@ -594,6 +588,13 @@ class BaseSolver(object):
             self._set_up_ext_and_inputs(model, external_variables, inputs)
             for inputs in inputs_list
         ]
+
+        # Cannot use multiprocessing with model in "jax" format
+        if(len(inputs_list) > 1) and model.convert_to_format == "jax":
+            raise pybamm.SolverError(
+                "Cannot solve list of inputs with multiprocessing "
+                "when model in format \"jax\"."
+            )
 
         # Set up
         timer = pybamm.Timer()
--- a/pybamm/solvers/base_solver.py
+++ b/pybamm/solvers/base_solver.py
@@ -738,6 +738,8 @@ class BaseSolver(object):
                             ext_and_inputs_list,
                         ),
                     )
+                    p.close()
+                    p.join()
             # Setting the solve time for each segment.
             # pybamm.Solution.append assumes attribute
             # solve_time.
--- a/pybamm/solvers/base_solver.py
+++ b/pybamm/solvers/base_solver.py
@@ -545,6 +545,12 @@ class BaseSolver(object):
         """
         pybamm.logger.info("Start solving {} with {}".format(model.name, self.name))
 
+        # Cannot use multiprocessing with model in "jax" format
+        if(len(inputs) > 1) and model.convert_to_format == "jax":
+            raise pybamm.SolverError(
+                "Cannot solve list of inputs with multiprocessing "
+                "when model in format \"jax\"."
+            )
         # Make sure model isn't empty
         if len(model.rhs) == 0 and len(model.algebraic) == 0:
             if not isinstance(self, pybamm.DummySolver):
--- a/tests/unit/test_solvers/test_scipy_solver.py
+++ b/tests/unit/test_solvers/test_scipy_solver.py
@@ -200,10 +200,9 @@ class TestScipySolver(unittest.TestCase):
         var = pybamm.Variable("var", domain=domain)
         model.rhs = {var: -pybamm.InputParameter("rate") * var}
         model.initial_conditions = {var: 1}
-        model.events = [pybamm.Event("var=0.5", pybamm.min(var - 0.5))]
         # No need to set parameters; can use base discretisation (no spatial
         # operators)
-
+        model.events = [pybamm.Event("var=0.5", pybamm.min(var - 0.5))]
         # create discretisation
         mesh = get_mesh_for_testing()
         spatial_methods = {"macroscale": pybamm.FiniteVolume()}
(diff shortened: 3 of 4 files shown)


tests/unit/test_solvers/test_scipy_solver.py
```diff
@@ -218,6 +218,39 @@ def test_model_solver_with_inputs(self):
         np.testing.assert_allclose(solution.y[0], np.exp(-0.1 * solution.t))
 
     def test_model_solver_multiple_inputs_happy_path(self):
+        for convert_to_format in ["python", "casadi"]:
+            # Create model
+            model = pybamm.BaseModel()
+            # Covert to casadi instead of python to avoid pickling of
+            # "EvaluatorPython" objects.
```
tinosulzer, line 225: comment out of date

tests/unit/test_solvers/test_scipy_solver.py
```diff
@@ -218,6 +218,39 @@ def test_model_solver_with_inputs(self):
...
+            model.convert_to_format = convert_to_format
+            domain = ["negative electrode", "separator", "positive electrode"]
+            var = pybamm.Variable("var", domain=domain)
+            model.rhs = {var: -pybamm.InputParameter("rate") * var}
+            model.initial_conditions = {var: 1}
+            # No need to set parameters; can use base discretisation (no spatial
+            # operators)
```
tinosulzer, line 232: comment out of date

tests/unit/test_solvers/test_scipy_solver.py
```diff
@@ -258,7 +298,7 @@ def test_model_solver_multiple_inputs_discontinuity_error(self):
         domain = ["negative electrode", "separator", "positive electrode"]
         var = pybamm.Variable("var", domain=domain)
         model.rhs = {var: -pybamm.InputParameter("rate") * var}
-        model.initial_conditions = {var: 1}
+        model.initial_conditions = {var: 2 * pybamm.InputParameter("rate")}
         # No need to set parameters; can use base discretisation (no spatial
         # operators)
```
tinosulzer, line 303: comment out of date

tests/unit/test_solvers/test_scipy_solver.py
```diff
@@ -272,28 +312,18 @@ def test_model_solver_multiple_inputs_discontinuity_error(self):
...
+            ("Input parameters cannot appear in expression " "for initial conditions."),
         ):
             solver.solve(model, t_eval, inputs=inputs_list, nproc=2)
 
-    def test_model_solver_multiple_inputs_initial_conditions_error(self):
+    def test_model_solver_multiple_inputs_jax_format_error(self):
         # Create model
         model = pybamm.BaseModel()
         # Covert to casadi instead of python to avoid pickling of
         # "EvaluatorPython" objects.
```
tinosulzer, line 325: comment out of date

pybamm/expression_tree/operations/evaluate.py
```diff
@@ -507,6 +508,23 @@ def evaluate(self, t=None, y=None, y_dot=None, inputs=None, known_evals=None):
...
+        return state
+
+    def __setstate__(self, state):
+        # Restore pickled attributes and
+        # compile code from "python_str"
+        # Execution of bytecode (re)adds attribute
+        # "_method"
+        self.__dict__.update(state)
+        compiled_function = compile(self._python_str, self._result_var, "exec")
+        exec(compiled_function)
```
tinosulzer, line 526: can this be tested? if not, add # pragma: no cover

tinosulzer (approved):
Thanks, neat fix :) I know you didn't introduce this but I just noticed lots of the comments in test_scipy_solver contradict the associated code ...

pybamm/expression_tree/operations/evaluate.py
tlestang: I'm surprised it's not, since this should run every time a model is unpickled.

tests/unit/test_solvers/test_scipy_solver.py
tlestang: Thanks, good catch

pybamm/solvers/base_solver.py
```diff
@@ -589,6 +589,13 @@ def solve(
             for inputs in inputs_list
         ]
 
+        # Cannot use multiprocessing with model in "jax" format
```
martinjrobins, line 592: Note that Jax is designed to support parallel solution of models with different inputs, so this should be doable. But it can be left for a PR later on

martinjrobins (approved):
looks great @tlestang , happy to merge. Had a couple of points below.

pybamm/expression_tree/operations/evaluate.py
```diff
@@ -507,6 +508,23 @@ def evaluate(self, t=None, y=None, y_dot=None, inputs=None, known_evals=None):
         else:
             return result
 
+    def __getstate__(self):
```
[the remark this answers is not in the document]
tlestang: I naively tried the same trick with the Jax evaluator but couldn't get it to work as smoothly.
Something in `jax.jit` can't be pickle, but I don't know what.
```
AttributeError: Can't pickle local object 'jit.<locals>.f_jitted'
```

I tried not pickling attributes `jit_evaluate` and `jac_evaluate` in the same way, i.e.
```python
def __getstate__(self):
    state = self.__dict__.copy()
    del state["_evaluate_jax"]
    del state["_jit_evaluate"]
    del state["_jac_evaluate"]
    return state
```
but still get the `AttributeError`..

pybamm/solvers/base_solver.py
tlestang: Okay good, would be better to rely on Jax's native behavior anyway :)

tlestang:
Thanks @tinosulzer and @martinjrobins  for the review. I will update the comments but before we merge that PR I'd like to investigate why the content of `__getstate__` isn't tested..

tlestang:
This brought coverage to 100%. Now waiting to merge #1319 to fix CI on MacOS

pybamm/expression_tree/operations/evaluate.py
tlestang: See 43da47a and 7729e87 below

PR #1302, opened 2020-12-21
Update reference

kinnala:
We recently published a paper on scikit-fem. This PR uses the new reference.


tinosulzer (approved):
Sure, thanks!

PR #1304, opened 2020-12-23
Experiment Drive Cycle with Events

alibh95:
# Description
Added the functionality of simulating drive cycles in Experiment class along with the 'for' event.
If the time specified in the 'for' event is less than the drive cycle time, then the drive cycle is stopped at the time specified in the 'for' event.
Else If the time specified in the 'for' event is greater than the drive cycle time, the drive cycle first extended for one week and then stopped at the time specified in the 'for' event.

### Fixes # (issue)

#1193 
#1293 (Can extend single and/or multiple drive cycles for long time durations)

### Note: 
I am new to GitHub. I was not able to manage my previous pull request (#1279) effectively. Therefore, I am opening this new pull request. I am sorry for any inconvenience I caused.

## Warning
In this PR, Only the Experiment class is modified to support drive cycle functionality. This functionality is not actually implemented in Simulation yet.

## Type of change

# Key checklist:

## Further checks:


brosaplanella:
Hi @alibh95! The errors in the MacOs tests should now be fixed if you merge in the develop branch. Let me know if the PR is ready to be reviewed.

brosaplanella:
Hi @alibh95! What's the state of this PR?

alibh95:
Hi @brosaplanella . This pull request successfully simulates drive cycles using experiment class. The implementation of for event have been completed. The drive cycle is clipped or extended depending upon the time period provided in the for event. 
Unfortunately, I am unable to spare much time for the successful implementation of until event because I am heavily occupied in completing my thesis. 
The issue I am facing in the implementation of until event is that whenever until event is triggered, the solver continues to solve the drive cycle instead of stoping. 
Any tips/suggestions would be helpful for me to complete this PR.

tinosulzer:
Hi @alibh95 , thanks for taking this on. We're going to implement this slightly differently in the simulation now that #1408 allows a better approach. Still, what you have added to the `Experiment` class is useful - could you change the PR to include only those changes (not the changes to the `Simulation` class)? Also add a warning that the drive cycle functionality is not actually implemented in Simulation yet.

examples/scripts/experiment_drive_cycle.py
```diff
@@ -0,0 +1,44 @@
+#
+# Constant-current constant-voltage charge with US06 Drive Cycle using Experiment Class.
```
tinosulzer, line 2: this example can be removed for now

pybamm/experiments/experiment.py
```diff
@@ -143,37 +162,112 @@ def read_string(self, cond):
...
+                # e.g. for 3 hours or until 4.2 V
+                idx_for = cond_list.index("for")
+                idx_until = cond_list.index("or")
+                end_time = self.convert_time_to_seconds(
+                    cond_list[idx_for + 1:idx_until])
+                ext_drive_cycle = self.extend_drive_cycle(drive_cycles[cond_list[1]],
+                                                          end_time)
+                electric = (ext_drive_cycle[:, 1], "Drive")
+                time = self._np.append(1, self._np.diff(ext_drive_cycle[:, 0]))
+                period = time
```
tinosulzer, line 185: I think period should just be a scalar, the min of the diff of the drive cycle (and time should also just be a scalar)

pybamm/experiments/experiment.py
```diff
@@ -143,37 +162,112 @@ def read_string(self, cond):
...
+                # e.g. for 3 hours
+                idx = cond_list.index("for")
+                end_time = self.convert_time_to_seconds(cond_list[idx + 1 :])
+                ext_drive_cycle = self.extend_drive_cycle(drive_cycles[cond_list[1]],
+                                                          end_time)
+                electric = (ext_drive_cycle[:, 1], "Drive")
+                time = self._np.append(1, self._np.diff(ext_drive_cycle[:, 0]))
+                period = time
+                events = None
+            elif "until" in cond:
```
tinosulzer, line 197: Dealing with events in drive cycles is challenging numerically. When running real experiments people usually choose a drive cycle that is short enough that the voltage limits won't be triggered (e.g. only 20% discharge). So I don't think we need the "until" condition for drive cycles.

tinosulzer (changes requested):

pybamm/experiments/experiment.py
alibh95: The implementation of the 'until' condition in the simulation class was the only limitation in commit (ffd340c) of this PR. Nevertheless, I will remove the implementation of until event for drive cycles from the experiment class in the next commit

examples/scripts/experiment_drive_cycle.py
alibh95: Noted

pybamm/experiments/experiment.py
alibh95: I will do so in the next commit.

pybamm/experiments/experiment.py
```diff
@@ -51,7 +63,12 @@ def __init__(
         parameters=None,
         period="1 minute",
         use_simulation_setup_type="new",
+        drive_cycles={},
     ):
+
+        import numpy as np
+        self._np = np
```
tinosulzer, line 70: I've never seen this style of import before. Usually we just import modules globally at the top of the file. Is there a style guide somewhere that says this is a better way of doing it?

alibh95: This was actually a style issue.

pybamm/experiments/experiment.py
```diff
@@ -66,7 +68,6 @@ def __init__(
         drive_cycles={},
     ):
 
-        import numpy as np
         self._np = np
```
tinosulzer, line 71: since we now import numpy at the top we should use that everywhere instead of declaring self._np

alibh95: Done.

tinosulzer (approved):
Thanks @alibh95 , looks good now. We're about to do a release, I will wait until that is done before merging as otherwise it might be confusing why you can specify a drive cycle in the experiment but not solve it

tinosulzer:
@all-contributors add @alibh95 for code, tests

PR #1309, opened 2020-12-28
Issue 1195 track cycles in solution

### LINKED ISSUE #1195: Track cycles in solution
Change experiment and solution classes so that `solution.cycles` is a list of the cycles (e.g. 1 CCCV charge + CC discharge)
See #1176

lonnbornj:
# Description

Altered the Experiment class to parse experiments expressed in terms of cycles (in the form of tuples of operating conditions). The Simulation class creates `solution.cycles` -- a list of cycles. Each cycle is a tuple of `sub_solution` objects.

Fixes #1195 

## Type of change

# Key checklist:

## Further checks:


pybamm/experiments/experiment.py
```diff
@@ -45,6 +45,13 @@ class Experiment:
 
     def __init__(self, operating_conditions, parameters=None, period="1 minute"):
         self.period = self.convert_time_to_seconds(period.split())
+        if all([isinstance(cycle, tuple) for cycle in operating_conditions]):
+            self.cycle_lengths = [len(cycle) for cycle in operating_conditions]
+            operating_conditions = [
+                cond for cycle in operating_conditions for cond in cycle
+            ]
+        elif all([isinstance(cond, str) for cond in operating_conditions]):
+            self.cycle_lengths = [len(operating_conditions)]
```
tinosulzer, line 54: What about the case where there is a mix of tuples and strings? This should either raise an error, or strings should be made into 1-tuples (i.e. cycles of length 1). Second option slightly better IMO.

pybamm/simulation.py
```diff
@@ -473,6 +473,18 @@ def solve(
                         "or reducing the period.\n\n"
                     )
                     break
+            if hasattr(self.solution, "_sub_solutions"):
```
tinosulzer, line 476: can you add a comment explaining what is being done here?

tests/unit/test_experiments/test_experiment.py
```diff
@@ -82,6 +82,24 @@ def test_read_strings_repeat(self):
...
+        self.assertEqual(
+            experiment.operating_conditions,
+            [
+                (0.05, "C", 1800.0, 60.0),
+                (-0.2, "C", 2700.0, 60.0),
+                (0.05, "C", 1800.0, 60.0),
+                (-0.2, "C", 2700.0, 60.0),
+            ],
+        )
+        self.assertEqual(experiment.cycle_lengths, [2, 2])
```
tinosulzer, line 101: add test for cycles of different lengths, and mixed cycle/string (see comment above)

tinosulzer (changes requested):
Thanks @lonnbornj , impressed how few changes you managed to do this in.
Just some edge cases to address. Also, can you edit the `cccv.py` and `gitt.py` examples to use cycles, since we want to encourage this in general, and add this change to the changelog.

pybamm/experiments/experiment.py
lonnbornj: I thought it made sense to keep all the type-checking together, so I moved the check for strings out of experiment.read_operating_conditions and do it all in __init__. Hopefully my solution is not too opaque (it got a bit complicated checking both cycles and the operating conditions therein).

rtimms:
This is a really neat solution, thanks @lonnbornj ! I think it would be good to add a "Getting Started" guide on how to use experiments, explaining the syntax for cycles etc., but this can be done as a separate issue.

rtimms:
Actually, just realised we already have an experiment guide [here](https://github.com/pybamm-team/PyBaMM/blob/develop/examples/notebooks/Getting%20Started/Tutorial%205%20-%20Run%20experiments.ipynb), so it can just be updated to show how to use cycles.

tinosulzer (approved):
Thanks @lonnbornj , looks good! Happy to merge, unless you want to update the tutorial @rtimms linked to

tinosulzer:
@all-contributors add @lonnbornj for code, test, example

Issue #989, opened 2020-05-11
Warning when "extrapolating" fitted parameter functions


ferranbrosa:
**Summary**
The parameter functions usually are provided with a range of validity, but this range is included in the parameter sets. Could we throw a warning if the functions are evaluated outside of this range? Not sure if that can be tackled jointly with #976 

 **Motivation**
Throwing a warning would make users more aware of the limitations of the parameters they are using and might help debugging the code.

ferranbrosa:
An easy way here would be to add an if statement inside the function that throws a warning when the fitting range is exceeded, but I suspect that would be slow and it would clutter the terminal with warnings. Not sure if we want that.

A better approach could be to add a non-terminating event (maybe new event type) that registers a warning which is printed after solving. The question here is where to define the event and the range limits.

Thoughts?

tinosulzer:
Was this fixed by #1315?

brosaplanella:
No, needs some extra work as at the moment the warnings are only defined by the interpolants

tinosulzer:
Ah I see the difference now. To define the range, I think we need a new class `BoundedFunctionOutput(output, range)`, and when `ParameterValues` sees this class it puts `output` into the expression tree and `range` into the events. This way we don't have some functions returning just one output (the function) and some returning two (function and range). Unless you had something else in mind :)

brosaplanella:
I had a similar idea. Wouldn't it be better to define the class with a `call` method (hopefully then it works the same way as with functions) and an attribute for the bounds? Then your function doesn't output the bounds all the time.

PR #1315, opened 2020-12-30
Issue 976 casadi extrapolate warning

### LINKED ISSUE #976: Casadi extrapolate warning/error
When Casadi extrapolates you get errors like
```
At t = 0.300171, , mxstep steps taken before reaching tout.
psetup failed: .../casadi/interfaces/sundials/idas_interface.cpp:830: Linear solve failed
psetup failed: .../casadi/interfaces/sundials/idas_interface.cpp:830: Linear solve failed
psetup failed: .../casadi/interfaces/sundials/idas_interface.cpp:830: Linear solve failed
psetup failed: .../casadi/interfaces/sundials/idas_interface.cpp:830: Linear solve failed
```
which are hard to trace back to the interpolant. It would be useful to throw a warning/error when this happens which is more insightful.

brosaplanella:
# Description

When an interpolant is created from data, a new type of event (`INTERPOLANT_EXTRAPOLATION`) is added to ensure that we can catch when extrapolation occurs. This is particularly important in CasADI as you get a lot of warnings that are not very insightful.

I haven't managed to remove all these warnings as, before hitting the event and raising the error, the integration step where extrapolation happens takes place. However, I don't know how to get rid of them if we can't catch the CasADI warnings.

Fixes #976

## Type of change

# Key checklist:

## Further checks:


pybamm/solvers/casadi_solver.py
```diff
@@ -71,6 +71,7 @@ def __init__(
         root_tol=1e-6,
         max_step_decrease_count=5,
         dt_max=None,
+        extrap_tol=1e-3,
```
rtimms, line 74: can you add this to the docstring? also, do other solvers have this problem with interpolants (i.e. they fail without giving much info)?

pybamm/solvers/casadi_solver.py
```diff
@@ -143,6 +145,33 @@ def _integrate(self, model, t_eval, inputs=None):
                     [event(t, y0, inputs) for event in model.terminate_events_eval]
                 )
             )
+
+            extrap_event = [
```
rtimms, line 149: do we need to check the initial conditions? I guess they could lead to extrapolation, but it would be nice to not repeat this code

rtimms:
thanks @brosaplanella looks good! my only question is should we also raise a warning when extrapolation occurs with the other solvers?

pybamm/parameters/parameter_values.py
```diff
@@ -410,6 +411,17 @@ def process_model(self, unprocessed_model, inplace=True):
                     event.name, self.process_symbol(event.expression), event.event_type
                 )
             )
+
+        for event in self.parameter_events:
+            pybamm.logger.debug(
+                "Processing parameters for event'{}''".format(event.name)
```
tinosulzer, line 417: missing space

pybamm/solvers/casadi_solver.py
```diff
@@ -215,6 +244,35 @@ def _integrate(self, model, t_eval, inputs=None):
...
+                                        inputs=inputs,
+                                    )
+                                    < self.extrap_tol
+                                ).any()
+                            ):
+                                extrap_event_names.append(event.name[12:])
+
+                        raise pybamm.SolverError(
+                            "CasADI solver failed because the following interpolation"
+                            " bounds were exceeded: {}".format(extrap_event_names)
```
tinosulzer, line 273: Maybe indicate here how this can be fixed? e.g. "You may need to provide additional interpolation points outside these bounds"

tinosulzer (approved):
Looks good to me, maybe @rtimms wants to take another look though

rtimms (commented):
Looks good to me, thanks! 

I was wondering if we should make all the solvers give at least a warning when you extrapolate. At the moment e.g. the scipy solver only stops for the terminal events (as expected), but should we also check and warn if extrapolation has occurred so that users are aware and can check if they trust the results. Maybe this should be a separate issue though.

pybamm/solvers/casadi_solver.py
brosaplanella: Done

pybamm/solvers/casadi_solver.py
brosaplanella: I did it so it fails faster when you give an initial condition that extrapolates. What do you prefer?

pybamm/parameters/parameter_values.py
brosaplanella: Done

pybamm/solvers/casadi_solver.py
brosaplanella: Done

brosaplanella:
@rtimms Sorry, you are totally right. I had in mind adding the warning for the other solvers, but I wanted first to check if this made sense or not.

I have tried what happens with IDAKLU and Scikits DAE solvers and they take a very long time when there is extrapolation. For example, if you try this with `my_solver` equal to the solver of your choice you will see it (can you check in your laptops please, just in case?)
```python3

param = pybamm.ParameterValues(chemistry=pybamm.parameter_sets.Chen2020)
ci = param["Initial concentration in positive electrode [mol.m-3]"]
param["Initial concentration in positive electrode [mol.m-3]"] = 0.8 * ci

sim = pybamm.Simulation(
    model,
    parameter_values=param,
    solver=my_solver,
)

sim.solve()
```

In this case, should I add the warning or should I add an error? Note that I will implement the warnings anyway in issue #989 so we can change our minds later.

pybamm/solvers/casadi_solver.py
rtimms: yeah probably better to fail faster

rtimms:
IMO a warning would be good. At least that way you get the solution back and can make sure it makes sense (and check how much you extrapolated by). Happy to leave this for another issue though and get this PR merged.

brosaplanella:
Cool, then I will add the warnings on this PR.

pybamm/solvers/casadi_solver.py
brosaplanella: I will make this sentence my motto from now on haha

rtimms (approved):
thanks @brosaplanella !

PR #1316, opened 2020-12-31
Issue 1221 convert to casadi

### LINKED ISSUE #1221: Optionally convert variable to casadi format before evaluating
Right now, for fast models such as the SPM, solving the model takes 2 ms but evaluating the voltage takes 160 ms. 

Under the right combination of conditions, it might be faster to convert a variable to casadi before evaluating it. The converted variable could be saved to a dictionary such as `model.casadi_variables` which would then be available to the next solution. This could also bypass the need for doing `y.full()` after solving to convert from casadi format to numpy format (`y.full()` takes as much time as doing the integration itself, 0.5 ms).

Could also try vectorizing as suggested in #788

tinosulzer:
# Description

Convert variables to casadi for faster post-processing.
The conversion is done the first time the variable is called, and the casadi function is then stored in the _model_ (in `model._variables_casadi`. Therefore, when doing repeated solutions of the same model (e.g. for parameter sweeps) the conversion is only performed the first time.

Fixes #1221

## Type of change

# Key checklist:

## Further checks:


tests/unit/test_solvers/test_casadi_solver.py
```diff
@@ -121,43 +121,43 @@ def test_model_solver_events(self):
...
         )
 
         # Solve using "safe" mode with debug off
         pybamm.settings.debug_mode = False
         solver = pybamm.CasadiSolver(mode="safe", rtol=1e-8, atol=1e-8, dt_max=1)
         t_eval = np.linspace(0, 5, 100)
         solution = solver.solve(model, t_eval)
-        np.testing.assert_array_less(solution.y[0], 1.5)
-        np.testing.assert_array_less(solution.y[-1], 2.5 + 1e-10)
+        np.testing.assert_array_less(solution.y.full()[0], 1.5)
```
brosaplanella, line 138: This test fails but I can't understand why it doesn't take the `.full()`.

brosaplanella (approved):
Looks good to me, happy to merge once the test passes. Thanks Tino!

tests/unit/test_solvers/test_casadi_solver.py
tinosulzer: `get_termination_reason` had changed in #1300 and was returning a numpy array

PR #1319, opened 2021-01-05
Fix GitHub actions MacOS

brosaplanella:
# Description

Looks like our issue with the MacOs tests evolved and now it doesn't randomly fail anymore. I have found that adding solves it 
`brew unlink gcc@8 gcc@9` (I hope it doesn't break anything down the line...).

### DIFF
--- a/.github/workflows/test_on_push.yml
+++ b/.github/workflows/test_on_push.yml
@@ -50,7 +50,6 @@ jobs:
     - name: Install MacOS system dependencies
       if: matrix.os == 'macos-latest'
       run: |
-#       rm -f /usr/local/bin/2to3
         brew update
         brew install graphviz
         brew install openblas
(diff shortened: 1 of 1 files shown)


tinosulzer (approved):
Looks good, thanks @brosaplanella

tlestang:
Thanks Ferran! Would it be possible to add one or two comments explaining why these lines where added? With a link with relevant references if this is inspired from something else.

tlestang:
Two relevant issues are:
- [brew upgrade fails for Python's 2to3 conflict](https://github.com/actions/virtual-environments/issues/2322)
- [brew upgrade fails on gcc](https://github.com/actions/virtual-environments/issues/2391)

brosaplanella:
> Two relevant issues are:
> 
> * [brew upgrade fails for Python's 2to3 conflict](https://github.com/actions/virtual-environments/issues/2322)
> * [brew upgrade fails on gcc](https://github.com/actions/virtual-environments/issues/2391)

I have now added the extra information. I only implemented the fix for gcc, which was basically what it was suggested by the error message. Let me know if I should add anything else.

tlestang (approved):

PR #1323, opened 2021-01-11
Issue 1322 thermal parameters dependent temperature

### LINKED ISSUE #1322: Make thermal parameters a function of temperature
**Summary**
Allow the parameters in the thermal models (density, heat capacity and thermal conductivity) to depend on temperature.

 **Motivation**
For the upcoming thermal parameters for the LG M50, heat capacity and thermal conductivity are temperature dependent. Density is not but, for consistency, it should be implemented too.

brosaplanella:
# Description

The thermal parameters (density, heat capacity and conductivity) can now be functions of temperature.

Fixes #1322

## Type of change

# Key checklist:

## Further checks:


pybamm/models/submodels/thermal/x_full.py
```diff
@@ -48,13 +48,27 @@ def set_rhs(self, variables):
         T = variables["Cell temperature"]
         Q = variables["Total heating"]
 
+        # Define volumetric heat capacity
+        rho_k = pybamm.Concatenation(
```
brosaplanella, line 52: Not sure if `rho_k` should be assembled here or in the parameters scripts...

pybamm/parameters/lead_acid_parameters.py
```diff
@@ -771,6 +762,16 @@ def a_p(self, x):
         x_dim = x * self.L_x
         return self.a_p_dimensional(x_dim) / self.a_p_typ
 
+    def rho(self, T):
```
brosaplanella, line 765: Due to the definition of the current collectors this is not equal to one (as it is with lead-acid...).

pybamm/input/parameters/lead-acid/anodes/lead_Sulzer2019/parameters.csv
```diff
@@ -36,7 +36,7 @@ Hydrogen reference OCP vs SHE [V],0,srinivasan2003mathematical,
 Negative electrode double-layer capacity [F.m-2],0.2,,
 ,,,
 # Density,,,
-Negative electrode density [kg.m-3],113400,CRC Handbook of Chemistry and Physics,
+Negative electrode density [kg.m-3],11300,CRC Handbook of Chemistry and Physics,
```
brosaplanella, line 39: It wasn't consistent with the value for the current collectors. Given that this one was extremely large, I assumed it was the incorrect one. @tinosulzer, does it make sense?

tinosulzer: Yep, just looked it up, thanks

pybamm/models/submodels/thermal/x_full.py
tinosulzer: Here is fine, but could you do this instead of calling `standard_variables`?
```python3
T_n, T_s, T_p = T.orphans
```

pybamm/parameters/lithium_ion_parameters.py
```diff
@@ -848,6 +839,16 @@ def c_p_init(self, x):
...
 
+    def rho(self, T):
+        "Dimensionless effective volumetric heat capacity"
+        return (
+            self.rho_cn(T) * self.l_cn
+            + self.rho_n(T) * self.l_n
+            + self.rho_s(T) * self.l_s
+            + self.rho_p(T) * self.l_p
+            + self.rho_cp(T) * self.l_cp
+        ) / self.l
```
tinosulzer, line 850: shouldn't this take separate temperatures for neg, sep, pos?

pybamm/parameters/thermal_parameters.py
```diff
@@ -118,59 +55,178 @@ def _set_dimensional_parameters(self):
...
     def T_amb_dim(self, t):
         "Dimensional ambient temperature"
         return pybamm.FunctionParameter("Ambient temperature [K]", {"Times [s]": t})
 
-    def _set_dimensionless_parameters(self):
-        "Defines the dimensionless parameters"
+    def rho_cn_dim(self, T):
+        "Negative current collector density [kg.m-3]"
+        inputs = {
+            "Temperature [K]": self.Delta_T * T + self.T_ref,
```
tinosulzer, line 71: for consistency with other functions in this script, dimensional functions should take dimensional temperature (i.e. convert from T to T_dim in the dimensionless function then pass T_dim to this function)

tinosulzer (changes requested):
Thanks @brosaplanella , just a couple of things to change / clarify

pybamm/parameters/lithium_ion_parameters.py
brosaplanella: This only appears on the lumped models where T is the same everywhere, but I agree that it can be misleading to define it here. Maybe it would be better to define it in the model script, like we did for the `x_full`.

pybamm/parameters/thermal_parameters.py
brosaplanella: Done (I think, please double check when I push).

pybamm/parameters/lithium_ion_parameters.py
tinosulzer: ok, that's fine then

pybamm/parameters/thermal_parameters.py
```diff
@@ -179,5 +235,45 @@ def T_amb(self, t):
         "Dimensionless ambient temperature"
         return (self.T_amb_dim(t) - self.T_ref) / self.Delta_T
 
+    def rho_cn(self, T):
+        "Dimensionless negative current collector density"
+        return self.rho_cn_dim(T) * self.c_p_cn_dim(T) / self.rho_eff_dim(self.T_ref)
```
tinosulzer, line 240: need to make `T` dimensional here and pass `T_dim` to the functions

tinosulzer (changes requested):
Each dimensionless function should calculate `T_dim` from `T` and then pass this to the dimensional function

tinosulzer (approved):
Thanks, all good now

Issue #1318, opened 2021-01-04
Casadi event handling 


rtimms:
Explore some other options for how events are handled by the casadi solver 

This is motivated by an issue @DrSOKane is having where the cut off voltage isn’t being hit:

![4D32B724-0541-4AE7-999E-684490B0BE0D](https://user-images.githubusercontent.com/43040151/103511638-de763000-4e80-11eb-99c0-70ba15a30297.jpeg)

I think this is related to having to do “step-and-check” integration to handle events in casadi. The solver returns the times in t_eval (which is controlled by the period). When an event is crossed we create a linear interpolant, then use a root finding algorithm to find t_event. We then go back and solve up to that time to get y_event. Obviously when the solution is highly nonlinear the t_event we found isn’t correct and therefore the event state is out too. One solution would be to reduce the period, but then you end up with more data. We could also try using a higher order intepolant on line 229 of casadi_solver.py. This kind of event checking happens internally for the other solvers, so they can use the internal adaptive steps taken by the solver rather than the times in t_eval, and therefore hit the event more accurately. Ideally we would be able to access the internal steps used by the casadi solver for use in the root finding. 

An alternative would be to do some iteration in the event location part of the solver. I.e. find t_event using interpolation, solve up to this time, check again if the event has been crossed. If it has, use root finding again. If it hasn’t, solve a little further. Repeat. 

Another option would be the following. When an event is crossed, go back and integrate over that window of time again requesting the solution at a more dense set of points before doing the root finding.

tinosulzer:
I like the iterative solution. Hopefully converges in 2-3 iterations since Newton-Raphson has quadratic convergence?

Another similar option is to directly pass the solver to the rootfinding algorithm, so the rootfinder itself is calling the solver. But I don't know whether this would work since we are looking for one event to be zero, not trying to find the point where they are all zero

rtimms:
Yeah it will be interesting to see what the trade off is in doing multiple integrations in an iteration loop vs just doing one dense solve to get a solution you trust everywhere in that window of time (to some tol) and using that to find the event time. The problem with the latter is how many points is enough?

I agree iteration should converge quickly.

rtimms:
The issue seems to be related to the period, e.g.

```python3
import matplotlib.pyplot as plt
import pybamm


exp = pybamm.Experiment(
    ["Hold at 4.2V until C/100", "Rest for 4 hours (3 minute period)"]
[...]

PR #1328, opened 2021-01-18
Re-solve before root finding when handling events in casadi 

rtimms:
# Description

Currently when an event is triggered in casadi we construct a linear interpolant, use a root finding algorithm to find the event time, and then go back and solve again up to the event time. This can give the wrong event time when the linear interpolant isn't a good fit to the actual solution (see #1318). 

This PR changes it so that when an event is triggered we first go back and re-solve in the current time window using a more dense array of times. This solution is then used to construct a higher-order interpolant for use in the root finding algorithm. This allows the event time to be found more accurately.

I'm leaving #1318 open as there are some other (more sophisticated) methods for handling events that could be explored. 

## Type of change

# Key checklist:

## Further checks:


tinosulzer (approved):
thanks @rtimms !

PR #1331, opened 2021-01-20
#1329 reformat Solution to avoid concatenating y

### LINKED ISSUE #1329: "Adding" solutions is slow (casadi.horzcat)
"Adding" solutions can take up a large portion of the overall solution time. The time is spend calling `casadi.horzcat`.

E.g. in the below example `casadi.horzcat` takes about 9% of the overall time.

```python3
import pybamm
options = {
    "sei": "solvent-diffusion limited",
    "sei film resistance": "distributed",
    "sei porosity change": "true",
}
model = pybamm.lithium_ion.DFN(options=options)
params = pybamm.ParameterValues(chemistry=pybamm.parameter_sets.Chen2020)
params.update(
    {
        "Reference temperature [K]": 278.15,
        "Ambient temperature [K]": 278.15,
        "Initial temperature [K]": 278.15,
        "Upper voltage cut-off [V]": 4.21,
    }
)
exp = pybamm.Experiment(
    ["Hold at 4.2V until C/100", "Rest for 4 hours (3 minute period)"]
    + [
        "Discharge at 1C until 2.5V",
        "Charge at 0.3C until 4.2V (3 minute period)",
        "Hold at 4.2V until C/100 (3 minute period)",
    ]
    * 50
)
sim = pybamm.Simulation(model=model, parameter_values=params, experiment=exp)
sim.solve()
```

tinosulzer:
# Description

Reformat `Solution` to avoid concatenating `y` (for performance)

Fixes #1329 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/simulation.py
+++ b/pybamm/simulation.py
@@ -455,11 +455,11 @@ class Simulation:
                 self.step(dt, solver=solver, npts=npts, **kwargs)
 
                 # Extract the new parts of the solution to construct the entire "phase"
-                new_num_subsolutions = len(self.solution.sub_solutions)
+                sol = self.solution
+                new_num_subsolutions = len(sol.sub_solutions)
                 diff_num_subsolutions = new_num_subsolutions - previous_num_subsolutions
                 previous_num_subsolutions = new_num_subsolutions
 
-                sol = self.solution
                 phase_solution = pybamm.Solution(
                     sol.all_ts[-diff_num_subsolutions:],
                     sol.all_ys[-diff_num_subsolutions:],
@@ -496,6 +496,9 @@ class Simulation:
                 cycle_solution = phases[cycle_start_idx]
                 for idx in range(cycle_length - 1):
                     cycle_solution = cycle_solution + phases[cycle_start_idx + idx + 1]
+                cycle_solution.phases = phases[
+                    cycle_start_idx : cycle_start_idx + cycle_length
+                ]
                 self.solution.cycles.append(cycle_solution)
             pybamm.logger.info(
                 "Finish experiment simulation, took {}".format(timer.time())
--- a/pybamm/simulation.py
+++ b/pybamm/simulation.py
@@ -443,7 +443,7 @@ class Simulation:
             pybamm.logger.info("Start running experiment")
             timer = pybamm.Timer()
 
-            phases = []
+            steps = []
             for idx, (exp_inputs, dt) in enumerate(
                 zip(self._experiment_inputs, self._experiment_times)
             ):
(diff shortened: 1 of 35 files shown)


pybamm/simulation.py
```diff
@@ -438,10 +437,13 @@ def solve(self, t_eval=None, solver=None, check_model=True, **kwargs):
             # Re-initialize solution, e.g. for solving multiple times with different
             # inputs without having to build the simulation again
             self._solution = None
+            previous_num_subsolutions = 0
             # Step through all experimental conditions
             inputs = kwargs.get("inputs", {})
             pybamm.logger.info("Start running experiment")
             timer = pybamm.Timer()
+
+            phases = []
```
rtimms, line 446: might be useful to add a comment about what a "phase" is. is the number of phases the same as the numbers of steps?

pybamm/solvers/solution.py
```diff
@@ -11,132 +10,145 @@
 from scipy.io import savemat
 
 
-class _BaseSolution(object):
+class Solution(object):
     """
     (Semi-private) class containing the solution of, and various attributes associated
```
rtimms, line 15: No longer semi-private?

pybamm/solvers/solution.py
```diff
@@ -11,132 +10,145 @@
...
+        self._sub_solutions = [self]
+
+        # Solution now uses CasADi
+        pybamm.citations.register("Andersson2019")
+
     @property
     def t(self):
         "Times at which the solution is evaluated"
-        return self._t
+        try:
```
rtimms, line 112: what happens if I call `solution.t` (or `solution.y`) then take a solver step and call `solution.t` again? does it get updated to include the new times/states or does it miss them since it already has the attribute `_t`?

rtimms (commented):
neat solution! just a few comments, thanks @tinosulzer !

brosaplanella (approved):
Looks good to me, thanks Tino!

pybamm/simulation.py
brosaplanella: I agree. I am not too sure what it means, so a comment would be helpful.

pybamm/solvers/solution.py
tinosulzer: I've actually removed `append` and made it so `__add__` returns a new object (i.e. the two solutions being added are unchanged), which I think is more intuitive and not slower

pybamm/simulation.py
tinosulzer: I'm thinking one cycle would consist of e.g. "discharge phase", "CC phase", "CV phase". Would you prefer "step" instead?

pybamm/solvers/solution.py
rtimms: ah yeah I see that now, thanks

rtimms (approved):

PR #1334, opened 2021-01-22
Nightly benchmarks with asv

tlestang:
# Description
This adds a GitHub action workflow `Benchmarks` (`periodic_benchmarks.yml) that periodically - for now once a day - runs the benchmarks suite in `benchmarks/` using [airspeed velocity (asv)](https://asv.readthedocs.io/en/stable/). It also updates the visualisation available at [pybamm-bench](https://tlestang.github.io/pybamm-bench/).

asv is a nice python utility that automates running benchmarks over the history of a Python project, as well as visualising the results and comparing benchmarks results across revisions.

asv relies on two things:
- A collection of benchmarks, located in the `benchmarks` directory.  Writing benchmarks is very similar to writing tests.
  See [Writing Benchmarks](https://asv.readthedocs.io/en/stable/writing_benchmarks.html).
- A configuration file [asv.conf.json](asv.conf.json).

Both are located at the root of the PyBaMM repo.

Everytime the workflow is triggered, for instance every day at 03:00 UTC, asv will run the benchmarks for every commit on
`develop` that hasn't been benchmarked yet.  Benchmarks results are _not_ stored in the PyBaMM repo, but rather in a separate repo [pybamm-bench](https://github.com/tlestang/pybamm-bench). The branch `gh-pages` in this repo holds the content of the visualisatin website hosted on github-pages, available at https://tlestang.github.io/pybamm-bench/.  This
website is updated each time the workflow is run.  The pybamm-bench repo currently lives on my personal GitHub account.
Once this PR is approved, we should migrate it under the pybamm-team organisation.

To populate the pybamm-bench repo with results for the previous commits in PyBaMM's history, there is another workflow
`Manual Benchmarks` (`run_benchmarks_over_history.yml`) which can be triggered manually from the github interface.  I
already did this from my fork of PyBaMM.

## Type of change
- [x] New feature (non-breaking change which adds functionality)

## Key checklist:

- [ ] No style issues: `$ flake8`
- [ ] All tests pass: `$ python run-tests.py --unit`
- [ ] The documentation builds: `$ cd docs` and then `$ make clean; make html`

## Further checks:

- [ ] Code is commented, particularly in hard-to-understand areas
- [ ] Tests added that prove fix is effective or that feature works

PR #1335, opened 2021-01-22
Check for any performance regression on pull request

tlestang:
# Description
_This comes on top of #1334 (Nightly benchmarks with asv).
Better to first review and merge #1334 then review this PR._

This adds a new GitHub action workflow `Run benchmarks on PR` (`benchmark_on_PR.yml`) that runs each time a pull request is opened and reports any performance regression compared to the tip of the base branch of the PR.
This workflow uses [airspeed velocity](https://asv.readthedocs.io/en/stable/) to run the benchmark suite for both the new commit and the tip of the base branch for the PR. If any of the benchmark run results in a performance regression of more than 10% compared to the base commit, the workflow reports a failure. Here "performance" means wall-clock time. The value of 10% for the regression ratio is arbitrary, and can be tuned at will.

I manually tested this feature on my fork, see for instance [this CI log](https://github.com/tlestang/PyBaMM/runs/1748192484?check_suite_focus=true). In  this case I introduced the Scipy Solver sleep for 2 seconds, in order to simulate a regression.
However I'm not sure how to automatically test that behaviour, so it might take a couple of PRs to get this right.
## Type of change

## Key checklist:


tlestang:
Close as the same PR was open from the same branch in this repo.
See #1342

PR #1344, opened 2021-01-25
#1343 update get_termination_reason to return solution

### LINKED ISSUE #1343: Event time and state missing from solution
The event time and state are no longer included in the solution. This is because "adding" solutions creates a new solution object. The event time and state are added in `get_termination_reason` which only returns the termination reason and _not_ the updated solution object. Previously the solution was modified in-place.

rtimms:
# Description
Return the solution as well as the termination reason in `get_termination_reason`. This is so that the updated solution (`solution = solution + event_sol`) which includes the event state gets returned by the solver.

Fixes #1343 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/solvers/base_solver.py
+++ b/pybamm/solvers/base_solver.py
@@ -574,6 +574,7 @@ class BaseSolver(object):
                 t_eval = np.array([0])
             else:
                 raise ValueError("t_eval cannot be None")
+
         # If t_eval is provided as [t0, tf] return the solution at 100 points
         elif isinstance(t_eval, list):
             if len(t_eval) == 1 and self.algebraic_solver is True:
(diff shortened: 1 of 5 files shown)


pybamm/solvers/base_solver.py
```diff
@@ -797,10 +797,18 @@ def solve(
...
 
-        # Identify the event that caused termination
-        termination = self.get_termination_reason(solutions[0], model.events)
+        # Identify the event that caused termination and update the solution to
+        # include the event time and state
+        # TODO: when solving with a list of inputs each solution may terminate due
+        # to a different event - this isn't captured here (or in the logger info)
+        solutions[0], termination = self.get_termination_reason(
+            solutions[0], model.events
+        )
```
tinosulzer, line 806: isn't this a trivial for loop?

tinosulzer (approved):
Ahh that makes sense, thanks for fixing

pybamm/solvers/base_solver.py
rtimms: Yeah, I’ll add it. For the logger info shall i just give different info when it’s a list of inputs (ie don’t specify a termination reason since it may be different for each solution).

tinosulzer: Yeah sounds good

PR #1345, opened 2021-01-26
add lines to remove 'tests' from site-packages in setup.py

tinosulzer:
# Description

`pip install pybtex` installs an unnecessary `tests` directory which messes up running some of the tests in pybamm individually.
This change removes that directory when it is installed via setup.py


brosaplanella (approved):
Looks good, thanks Tino!

PR #1346, opened 2021-01-27
update logging info for solutions list

rtimms:
# Description

I previously updated the logger info so that when calling `solve` for a list of inputs it didn't print the termination reason (since it may be different for each set of inputs). At the same time I changed it so that it returned the sum of the integration, solve and total times. This was the wrong thing to do as it gives the impression that the call to `solve` to ~N times as long as it actually did (for a list on N inputs). I've reverted it back to just print a single set up, solve and total time (note all solutions get given the same solve time, see https://github.com/pybamm-team/PyBaMM/pull/1261). I also removed the integration time info from the logger in this case since it is also different for each set of inputs. Users can still access the integration time for each solution in the list of solutions if they want to.

The logger info when returning a single solution remains unchanged.

### DIFF
--- a/pybamm/solvers/base_solver.py
+++ b/pybamm/solvers/base_solver.py
@@ -821,14 +821,10 @@ class BaseSolver(object):
         else:
             pybamm.logger.info("Finish solving {} for all inputs".format(model.name))
             pybamm.logger.info(
-                (
-                    "Set-up time: {}, Solve time: {} (of which integration time: {}), "
-                    "Total time: {}"
-                ).format(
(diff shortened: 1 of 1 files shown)


brosaplanella (approved):
Looks good, thanks Rob! Another option could be to pass the average time (pm std if we want) to give an idea on how much each run took.

rtimms:
Yeah I did think about that, but thought it might clutter up the info. I'll leave it for now and people can inspect the integration times for each solution manually. Maybe we can revisit this though as people start using this functionality more.

PR #1350, opened 2021-02-01
Electrolyte overpotentials

### LINKED ISSUE #1349: Implement electrolyte overpotentials for full_conductivity model
**Summary**
Implement the definitions of electrolyte overpotentials for the full_conductivity model.

brosaplanella:
# Description

Added the electrolyte overpotential and ohmic losses to the full_conductivity model.

Fixes #1349 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Features
 
+-   Added the electrolyte overpotential and Ohmic losses for full conductivity, including surface form ([#1350](https://github.com/pybamm-team/PyBaMM/pull/1350))
 -   Added functionality to `Citations` to print formatted citations ([#1340](https://github.com/pybamm-team/PyBaMM/pull/1340))
 -   Updated the way events are handled in `CasadiSolver` for more accurate event location ([#1328](https://github.com/pybamm-team/PyBaMM/pull/1328))
 -   Added error message if initial conditions are outside the bounds of a variable ([#1326](https://github.com/pybamm-team/PyBaMM/pull/1326))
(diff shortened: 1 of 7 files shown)


tinosulzer:
We should also be able to replace the `get` in these lines with a standard dictionary read https://github.com/pybamm-team/PyBaMM/blob/fa076b8a571379410c42f4dced7f8ddbecb51338/pybamm/models/full_battery_models/base_battery_model.py#L851

brosaplanella:
@tinosulzer I have tried what you suggested but it causes issues with the surface form equations. Because they are defined on each electrode and separator separately it is not clear to me where the overpotentials (which are global) should be defined. One way would be to do it where the DFN (and similarly for the others) is defined but that seems an unnecessary repetition of the code. 

Alternatively, I can leave the `get` there and open another issue for the surface form.

rtimms:
Could you add an extra submodel in `surface_potential_form` that is used for calculating the overpotentials? Not ideal, but not entirely dissimilar to e.g. how we have `CurrentForInverseButlerVolmer`.

Alternatively, could you make calculating the overpotentials a new method in `BaseElectrolyteConductivity` that has a try/except in to get the variables it needs. Then the surface model on each domain would call it in turn - calculating the overpotentials would be skipped until all the necessary variables are defined. Not sure this is a very nice solution though...

rtimms (approved):
Thanks @brosaplanella ! If you wanted to avoid repeated code you could add a method `_get_overpotentials` which you can call from both `_get_whole_cell_variables` in the base model and `get_coupled_variables` in the full model.

PR #1355, opened 2021-02-03
Disable compilation of sundials and scikits.odes on mac

tlestang:
# Description
Since 03/02/2021 the scheduled "PyBaMM" workflow fails on mac for all python versions.
See https://github.com/pybamm-team/PyBaMM/actions/runs/532703527
The logs indicate that CMake cannot find the fortran compiler, and therefore Sundials isn't compiled... and the installation of scikits.odes fails.

Until a solution is found, this disables the compilation of sundials, as well as the installation of scikits.odes, on macos.

### DIFF
--- a/.github/workflows/benchmark_on_PR.yml
+++ b/.github/workflows/benchmark_on_PR.yml
@@ -21,8 +21,8 @@ jobs:
         run: |
           asv machine --machine "GitHubRunner"
           # Get IDs of branch and PR commits
-          BASE_COMMIT=$(git rev-parse $GITHUB_BASE_REF)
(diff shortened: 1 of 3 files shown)


tinosulzer:
Passed the mac tests by skipping every test 😂

tlestang:
Thanks @tinosulzer for noticing that. I think it looks ok now

.github/workflows/benchmark_on_PR.yml
```diff
@@ -1,6 +1,6 @@
 name: Run benchmarks on PR
```
tinosulzer, line 1:
```suggestion
name: Run benchmarks on push
```

tinosulzer (approved):
awesome, thanks @tlestang

PR #1364, opened 2021-02-05
Issue 1363 length scale bug

### LINKED ISSUE #1363: Incorrect values for y axis in 2D simulation?
**Describe the bug**

This might not be a bug but just me not understanding how to do things properly (PyBaMM beginner...).
When using a 2+1D pouch cell model, the plot uses values for the y (horizontal) axis which are not what I would expect based on the `Electrode width` parameter setting.

**To Reproduce**

```python
import pybamm

options = {"cell geometry": "pouch", "thermal": "x-lumped", "dimensionality": 2}
model = pybamm.lithium_ion.SPM(options=options)

parameter_values = model.default_parameter_values
parameter_values.update(
    {
        "Electrode height [m]": 0.3,
        "Electrode width [m]": 0.1,
        "Negative tab width [m]": 0.03,
        "Positive tab width [m]": 0.03,
        "Negative tab centre y-coordinate [m]": 0.03,
        "Positive tab centre y-coordinate [m]": 0.07,
        "Negative tab centre z-coordinate [m]": 0.3,
        "Positive tab centre z-coordinate [m]": 0.3,
    }
)
parameter_values.process_model(model)
parameter_values.process_geometry(geometry)

sim = pybamm.Simulation(model, parameter_values=parameter_values, solver=pybamm.CasadiSolver("fast"))
sim.solve([0, 60])
```
**Expected behaviour**

I expected the horizontal axis (y) to extend from 0 to 10 cm.

**Screenshots**

image

**Additional context**

Using PyBaMM 0.3.0 or current master (no difference). Python 3.8.5.

tinosulzer:
# Description

Changes the "current collector y" length scale from `L_y` to `L_z`. This is because both the y- and z-directions should be scaled with `L_z` when non-dimensionalising (otherwise there is an awkward extra factor of `L_y/L_z` in the 2D Laplacian for the current collector problem).

Fixes #1363 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -302,7 +302,7 @@
         "code",
         "example",
         "doc",
-        "test",
+        "test"
       ]
     },
     {
--- a/pybamm/models/full_battery_models/lead_acid/base_lead_acid_model.py
+++ b/pybamm/models/full_battery_models/lead_acid/base_lead_acid_model.py
@@ -30,7 +30,7 @@ class BaseModel(pybamm.BaseBatteryModel):
             "negative electrode": self.param.L_x,
             "separator": self.param.L_x,
             "positive electrode": self.param.L_x,
-            "current collector y": self.param.L_y,
+            "current collector y": self.param.L_z,
             "current collector z": self.param.L_z,
         }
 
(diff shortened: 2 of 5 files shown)


rtimms (approved):
thanks @tinosulzer

PR #1374, opened 2021-02-10
Update benchmark workflows

tlestang:
`periodic_benchmarks.yml`: Fetching the develop branch made sense when developing the workflow on another branch, but doesn't anymore now that the workflow runs on develop. Also some syntax needed to be fixed in order to properly extract committer identity from secrets.

`benchmark_on_push.yml` (formerly `benchmark_on_PR`): Environment variables `GITHUB_BASE_REF` and `GITHUB_HEAD_REF` are only defined for pull requests event. Instead, we just want to fetch the `develop` branch.


tinosulzer (approved):
Thanks for fixing @tlestang

PR #1375, opened 2021-02-10
Update contibuting guidelines

tlestang:
Add a section about running benchmarks with `asv` as well as adding new benchmarks.

Update the section on CI to mention GitHub actions instead of Travis CI


tinosulzer (approved):
Thanks, looks good to me

PR #1380, opened 2021-02-11
Remove OCV from ECM resistance definition

### LINKED ISSUE #1377: Local ECM resistance
**Describe the bug**

Hi, I am trying to understand the "Local ECM resistance" variable. It seems like this is [defined](https://github.com/pybamm-team/PyBaMM/blob/develop/pybamm/models/full_battery_models/base_battery_model.py#L904-L934) as the sum of all overpotentials, plus the difference between current OCV and initial OCV, divided by the current. If I understand correctly, this corresponds to treating the entire cell as a single Ohmic resistor with a time-dependent resistance. However, what I would have expected is that "ECM" refers to a model with a voltage source for OCV in series with an Ohmic resistor, so the OCV is not considered in the voltage difference.

It could be that this is intentional, so I would just like to clarify.

**To Reproduce**

```python
import pybamm
sim.solve([0, 3600])
sim.plot(["Local ECM resistance [Ohm]"])
```

**Expected behaviour**

I would have expected the internal resistance (defined as the resistance in series with the OCV voltage source) to not vary drastically during discharge.

**Screenshots**

image

**Additional context**
Reproduced under v0.3.0 and on master (903b4a05ef5a4f91633e990d4aec12c53df723a2).

DavidMStraub:
# Description

To address #1377, this removes the OCV (difference between current OCV and initial OCV) from the definition of the `Local ECM resistance`. The `Change in measured open circuit voltage` variable is left untouched.

Fixes #1377

## Type of change

# Key checklist:

## Further checks:

## TO DO

- [ ] Add to changelog

## Screenshot

Same plot as in #1377, after change:

grafik

### DIFF
--- a/pybamm/models/full_battery_models/base_battery_model.py
+++ b/pybamm/models/full_battery_models/base_battery_model.py
@@ -901,14 +901,8 @@ class BaseBatteryModel(pybamm.BaseModel):
         i_cc = self.variables["Current collector current density"]
         i_cc_dim = self.variables["Current collector current density [A.m-2]"]
         # Gather all overpotentials
-        v_ecm = -(eta_ocv + eta_r_av + eta_c_av + eta_e_av + delta_phi_s_av)
-        v_ecm_dim = -(
-            eta_ocv_dim
-            + eta_r_av_dim
-            + eta_c_av_dim
-            + eta_e_av_dim
-            + delta_phi_s_av_dim
-        )
+        v_ecm = -(eta_r_av + eta_c_av + eta_e_av + delta_phi_s_av)
+        v_ecm_dim = -(eta_r_av_dim + eta_c_av_dim + eta_e_av_dim + delta_phi_s_av_dim)
         # Current collector area for turning resistivity into resistance
         A_cc = self.param.A_cc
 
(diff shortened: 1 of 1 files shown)


tinosulzer (commented):
Changes are good but this is making me realize it might be cleaner to just define `v_ecm` as `V - ocv`?

DavidMStraub:
That's true. Implemented this now.

tinosulzer (approved):
Thanks @DavidMStraub , just checking with @TomTranter for approval too

tinosulzer:
@all-contributors add @DavidMStraub for code

TomTranter:
I think I originally did it this way with component as a sort of double check on whether it matched v - v_ocv. To be honest I don't actually use this variable in my external code anyway as I want all the components individually to output. I guess it's still useful - if a little confusing the way I'd done it. I sort of think of change in ocv as an extra overpotential

TomTranter (approved):

tinosulzer:
So maybe we should keep that sum as a "sum of overpotentials" variable?

TomTranter:
Probably not necessary to keep both. I approve this change

PR #1382, opened 2021-02-12
run black and flake8

priyanshuone6:
# Description

I found various formatting and styling mistakes so I ran `black .` and `flake8`

# Key checklist:

## Further checks:


tinosulzer (approved):
Ok. This doesn't matter that much tbh. As part of the tests we already check flake8, then there is a little bit of play in what qualifies under flake8.

PR #1395, opened 2021-02-19
#1393 end simulation when experiment infeasible

### LINKED ISSUE #1393: Stop simulation when experiment infeasible
When running a simulation with an experiment if you hit an event that wasn't defined by the experiment you get a warning "Experiment is infeasible". But then the simulation proceeds to the next cycle. I would expect the _whole_ simulation to stop when you hit an event like this (typically min/max concentration), not just that cycle. 

e.g. here you hit minimum voltage in cycle 3, but then it goes ahead and tries to solve for cycle 4. It starts from t=0 and uses the current state as initial conditions, so then hits the same event again.

```python3
import pybamm

    [
        ("Discharge at 1C for 0.5 hours",),
    ]
    * 4
)
sim = pybamm.Simulation(model, experiment=experiment, solver=pybamm.CasadiSolver())
sol = sim.solve()
```

rtimms:
# Description
Makes it so that simulations stop when an experiment becomes infeasible 

Fixes #1393 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
 
 ## Bug fixes
 
+-   Simulations now stop when an experiment becomes infeasible ([#1395](https://github.com/pybamm-team/PyBaMM/pull/1395))
 -   Added a check for domains in `Concatenation` ([#1368](https://github.com/pybamm-team/PyBaMM/pull/1368))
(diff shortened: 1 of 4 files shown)


tinosulzer (approved):
Looks good, thanks!

PR #1400, opened 2021-02-25
Add new workflow to build macos wheels

### LINKED ISSUE #1296: Mac OS Sundials bug
**Descirbe the bug**
scikits.odes does not work ("image not found") when following installation instructions

**Explanation**
As I understand it the problem is that scikits.odes looks for `libsundials_sunlinsolklu.3.dylib`, but `brew install sundials` installs a more recent version, which only has
```
libsundials_sunlinsolklu.dylib
libsundials_sunlinsollapackband.3.5.0.dylib
```
so `libsundials_sunlinsolklu.3.dylib` cannot be found.

This is not caught by the Mac OS tests since in CI the Mac OS tests use the script `install_KLU_Sundials.py`

Furthermore, if installing locally, with `tox -e pybamm-requires`, the script is used (which installs the correct version of sundials) but only `LD_LIBRARY_PATH` is updated, and not `DYLD_LIBRARY_PATH` as should be the case for mac.
This means that scikits.odes is still looking in `/usr/local/opt/` for sundials, where it finds the incorrect brew installation, instead of in `~/.local/lib`

** Solution **
- Remove the requirement for `brew install sundials` from docs
- Use the KLU script to install sundials on mac, and also update `DYLD_LIBRARY_PATH`

This could also be an opportunity to clean up the install docs, which I still find a little bit confusing

tlestang:
_This is the first PR in a series of 3. DO NOT MERGE until all PRs in the series are approved. PRs must be merged from top to bottom_

3 - #1402
2 - #1401
1 - #1400

## Description

Currently, PyBaMM wheels for macos contain the `idaklu` extension but not its (runtime) dependencies sundials and suitesparse. This means mac users must install these manually, most likely through Homebrew.
the currently available wheels where compiled using a version of sundials that is much older than the one installed by Homebrew.
This leads to a runtime error whenever the idaklu solver is imported... which basically means pybamm can't be imported.
See #1296 

We could always make sure to compile the wheel with the right version but an even better option is to bundle the dependencies in the wheel itself, using [delocate](https://pypi.org/project/delocate/). This means users don't have even have to know about sundials and suitesparse. The new workflow relies on the amazing [cibuildwheel](https://pypi.org/project/cibuildwheel/) utility to automate the wheel building and repair (i.e. it calls delocate under the hood).

This is the first PR in a series of 4. The next PR enrich this workflow, using cibuildwheel to build GNU/Linux wheels in addition to macos.

Fixes #1296

## Type of change

_This is the first PR in a series of 3. DO NOT MERGE until all PRs in the series are approved. PRs must be merged from top to bottom_

3 - #1402
2 - #1401
1 - #1400


.github/workflows/build_wheels.yml
```diff
@@ -0,0 +1,35 @@
...
+      # Used to host cibuildwheel
+      - uses: actions/setup-python@v2
+
+      - name: Install cibuildwheel
+        run: python -m pip install cibuildwheel==1.9.0
+
+      - name: Install sundials
+        run: |
+          brew update
+          brew install sundials
```
tinosulzer, line 22: why do we still need to brew install sundials if we now have the wheel?

tinosulzer (approved):
Thanks @tlestang , reviewing without completely understanding everything but I was able to install locally even after uninstalling sundials

.github/workflows/build_wheels.yml
tlestang: That's because this workflow is building the wheel. In order to compile the idaklu module (written in C++), this workflow (or anyone building pybamm from source) must have sundials installed. The resulting (compiled/binary) module still depends on sundials, but the required sundials libraries are bundled into the wheel, meaning that anyone with the wheel doesn't need to install sundials. But to build the wheel, it is needed.

PR #1407, opened 2021-02-27
#1406 fixed bug with positive electrode ohmic losses

### LINKED ISSUE #1406: Bug positive electrode ohmic losses
**Describe the bug**
The positive electrode Ohmic losses are positive, but they should be negative. This is because the definition on line 55 of `base_electrode.py` has the wrong sign compared to eq [8] in Marquis et al. (2019).

**To Reproduce**
1. Just run the default DFN model
2. Plot "X-averaged positive electrode ohmic losses"
3. See that it is positive

**Expected behaviour**
The ohmic losses should be negative.

brosaplanella:
# Description

Fixed the sign in the definition of positive electrode overpotential.

Fixes #1406

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
 
 ## Bug fixes
 
+-   Fixed a sign error in the positive electrode ohmic losses ([#1407](https://github.com/pybamm-team/PyBaMM/pull/1407))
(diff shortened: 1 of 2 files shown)


tinosulzer (approved):
Good catch, thanks @brosaplanella

PR #1408, opened 2021-03-01
Issue 1276 add models

### LINKED ISSUE #1276: "Add" solutions from different models
Add functionality to "add" (i.e. concatenate) solutions that come from different models, e.g. using a DFN for the charge and a SPM for the discharge.

tinosulzer:
# Description

- Allow creating a `Solution` that consists of stepping *different* models
- Reformat the simulation with experiment using this to be more efficient (at the cost of slightly longer set-up time)

Fixes #1276 

## Type of change

# Key checklist:

## Further checks:


pybamm/models/base_model.py
```diff
@@ -382,13 +382,25 @@ def set_initial_conditions_from(self, solution, inplace=True):
...
+                        "To update a model from a solution, each variable in "
+                        "model.initial_conditions must appear in the solution with "
+                        "the same key as the variable name. In the solution provided, "
+                        f"{e.args[0]}"
+                    )
                 if isinstance(solution, pybamm.Solution):
                     final_state = final_state.data
                 if final_state.ndim == 1:
-                    final_state_eval = np.array([final_state[-1]])
+                    final_state_eval = final_state[-1:]
```
brosaplanella, line 403: Probably a stupid question, but is this colon necessary?

brosaplanella (approved):
Looks good Tino, thanks!

pybamm/models/base_model.py
tinosulzer: Yeah, just discovered this in python, it returns an array (which is needed) instead of a scalar
```python3
np.array([1,2,3])[-1] # --> 3
np.array([1,2,3])[-1:] # --> np.array([3])
```

PR #1417, opened 2021-03-05
Add infrastructure and CI worklfow to build windows wheels

tlestang:
# Description

This builds on top on #1402 to generate python wheels for Windows. Both SuiteSparse and SUNDIALS are installed through
[vcpkg](https://github.com/Microsoft/vcpkg), a (relatively) recent package manager for C++ libraries.

## Automatically installing sundials with vcpkg in "manifest mode"

Installing a library with vcpkg is as simple as `vcpkg install sundials`. In addition, vcpkg can be automatically called to install missing dependencies when building PyBaMM from CMake, provided than a "manifest file" `vcpkg.json` is found at the root or he project (in this case `PyBaMM/`):
```json
# vcpkg.json
{
  "name": "pybamm",
  "version-string": "0.3.0",
  "dependencies": [
    {
      "name": "sundials",
      "default-features": false,
      "features": [ "klu" ]
    }
  ]
}
```
See [Using a manifest file to declare your dependencies](https://vcpkg.readthedocs.io/en/latest/users/manifests/).

## Using a separate registry for sundials with klu support

The `idaklu` solver provided by PyBaMM requires that the sundials library is compiled with KLU support, a sparse matrix factorization algorithm provided by the [suitesparse](https://people.engr.tamu.edu/davis/suitesparse.html) library.
Currently, the version of sundials available through vcpkg doesn't support KLU. Luckily, vcpkg makes it easy to add "features" to ported libraries, allowing some flexibility with regards to different build options. In this case, I added a "klu" feature to the currently available sundials port, so that, if that feature is specified (e.g. `vcpkg install sundials[klu]`) sundials is linked to suiteparse and built with KLU support enabled.

I'm currently in the process of contributing these changes to the main vcpkg repo (see [[sundials] Add klu feature]) (https://github.com/microsoft/vcpkg/pull/16514) but must sign a contributor license agreement before my PR is merged. In the meantime, I've set up a separate vcpkg registry at [pybamm-team/sundials-vcpkg-registry](https://github.com/pybamm-team/sundials-vcpkg-registry) which provides sundials with KLU enabled. 

To tell vcpkg to look for the sundials port in the PyBaMM registry instead of the main one, we can write a `vcpkg-configuration.json` file at the root of the PyBaMM repo:
```json
# vcpkg-configuration.json
{
  "registries": [
    {
      "kind": "git",
      "repository": "https://github.com/pybamm-team/sundials-vcpkg-registry.git",
      "packages": [ "sundials" ]
    }
  ]
}
```
## Required environment variables

To build PyBaMM (and particularly the `idaklu` extension) using CMake and vcpkg, it is required that 4 environment variables are defined:
- `PYBAMM_USE_VCPKG`: Can be anything. If empty, vcpkg isn't used (for instance on platforms other than windows)
- `VCPKG_ROOT_DIR`: The full path to the `vcpkg` installation. On GitHub runners, this is contained in `VCPKG_INSTALLATION_ROOT`.
- `VCPKG_DEFAULT_TRIPLET`: The default triplet, e.g. `x64-windows-static` or `x86-windows-static`. See [Triplet selection](https://vcpkg.readthedocs.io/en/latest/users/integration/).
- `VCPKG_FEATURE_FLAGS`: Should be set to `manifests,registries`. See [Environment and Configuration](https://vcpkg.readthedocs.io/en/latest/users/config-environment/).

## Statically linking SuiteSparse and sundials to the `idaklu` extension

The `idaklu` solver has a runtime dependency on some libraries in both SuiteSparse and sundials (e.g. `sunlinsolklu`).
On Linux and macos, there is a clear way of distributing the required share libraries along with the idaklu extension, by including the `.so` (or `.dylib`) binaries inside the python wheel. On GNU/Linux, this can be achieved using `auditwheel` or on macos, using `delocate`. There is no such tool for windows and I couldn't find a widely accepted way of distributing shared library dependencies. An alternative approach is to link these dependencies statically to the final Python extension. This leads to a bigger extension binary, but with only a few libraries, this is largely acceptable.

To link statically, the static version of sundials and suitesparse are installed using the triplet `x64-windows-static` with vcpkg, for instance:
```shell
vcpkg install sundials[klu]:x64-windows-static
```
Lastly, we set the variable `VCPKG_TARGET_TRIPLET` to e.g. `x64-windows-static` in the `CMakeLists.txt`:
```cmake
# CMakeLists.txt
if(DEFINED ENV{VCPKG_DEFAULT_TRIPLET} AND NOT DEFINED VCPKG_TARGET_TRIPLET)
  set(VCPKG_TARGET_TRIPLET "$ENV{VCPKG_DEFAULT_TRIPLET}"
    CACHE STRING "Vcpkg target triplet")
endif()
```


tinosulzer (approved):

tlestang:
Switching this back to draft mode as this only covers x64 wheels. this will not work for users using 32-bit python.
Currently trying to install sundials for both architectures, see https://github.com/microsoft/vcpkg/issues/16554

tlestang:
Rebased branch `windows_build` onto `main`

tinosulzer (approved):
Thanks @tlestang , can those json files be put into tox.ini so we reduce the number of files in root or not?

tinosulzer:
For the benchmarks you might just need to merge the latest version of develop

.github/workflows/build_wheels.yml
```diff
@@ -1,4 +1,4 @@
-name: Build
+name: "Build and publish Python wheels"
```
fcooper8472, line 1: Not a problem, but quoting strings is only needed in yml if they include special characters such as :

.github/workflows/build_wheels.yml
```diff
@@ -15,38 +15,52 @@ jobs:
...
       - uses: actions/checkout@v2
 
-      # Used to host cibuildwheel
       - uses: actions/setup-python@v2
 
       - name: Install cibuildwheel
         run: python -m pip install cibuildwheel==1.9.0
 
+      - name: Clone pybind11 repo
+        run: git clone https://github.com/pybind/pybind11.git
```
fcooper8472, line 28: unlikely, but possible that getting the latest pybind might cause problems.  Perhaps clone a tagged release instead?

.github/workflows/build_wheels.yml
```diff
@@ -15,38 +15,52 @@ jobs:
...
-          python -m pip install wheel
-          python -m pip wheel --no-deps --wheel-dir wheelhouse .
+      - name: Cache packages installed through vcpkg on windows
+        if: matrix.os == 'windows-latest'
+        uses: actions/cache@v2
+        env:
+          cache-name: vckpg_binary_cache
+        with:
+          path: C:\Users\runneradmin\AppData\Local\vcpkg\archives
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('vcpkg.json') }}
```
fcooper8472, line 49: May (?) also be worth adding the compiler version to the cache key

.github/workflows/build_wheels.yml
```diff
@@ -15,38 +15,52 @@ jobs:
...
         run: python -m pip install cibuildwheel==1.9.0
 
+      - name: Clone pybind11 repo
+        run: git clone https://github.com/pybind/pybind11.git
+
       - name: Install sundials on macOS
         if: matrix.os == 'macOS-10.15'
         run: |
           brew update
           brew install sundials
```
fcooper8472, line 34: Perhaps later all OSs should get sundials from the vcpkg registry.  Homebrew is very trigger happy on updating to potentially incompatible versions

CMakeBuild.py
```diff
@@ -60,21 +88,35 @@ def run(self):
                 "-DSUNDIALS_ROOT={}".format(os.path.abspath(self.sundials_root))
             )
 
-        if not os.path.exists(self.build_temp):
-            os.makedirs(self.build_temp)
+        build_dir = self.get_build_directory()
+        if not os.path.exists(build_dir):
```
fcooper8472, line 92: This if statement is equivalent to
```
os.makedirs(build_dir, exist_ok=True)
```

fcooper8472 (approved):
Looks good to me!

None of the comments I have left are in any way major, so feel free to either address them or not.

.github/workflows/build_wheels.yml
tlestang: Good to know!

.github/workflows/build_wheels.yml
tlestang: Very good point, will make that change

.github/workflows/build_wheels.yml
tlestang: Good idea. Hashing `vcpkg.json` might also not be a great, since it contains the version of PyBaMM. We don't want to clear the cache each time PyBaMM changes version...
I will probably look at this in more details after this is merged.

.github/workflows/build_wheels.yml
tlestang: Yes. Building the `idaklu` extension (which depends on sundials) is the only thing that made me have a look at vcpkg. But now that it is setup, I can't think of a reason why this shouldn't be used on other platforms as well. This would really simplify the process.. currently the sundials build is different on the the 3 OSs!

CMakeBuild.py
tlestang: Good catch

tlestang: There are a few things that need refactoring in this module, so will do that then

tlestang:
> For the benchmarks you might just need to merge the latest version of develop

@tinosulzer I'll rebase this onto develop then. Do you have an idea of what might cause the performance decrease?

tlestang:
> can those json files be put into tox.ini so we reduce the number of files in root or not?

Unfortunately no, as far as I know

tinosulzer:
My guess is the `windows_build` branch is running an older version of the benchmarks that is slower (some recent PRs made the benchmarks faster). Otherwise I don't know why it would have changed

tlestang:
@tinosulzer Okay, rebasing on the latest develop did it

PR #1418, opened 2021-03-05
Options class in base_battery_model

### LINKED ISSUE #894: Model options class
**Summary**
Create a class to set model options with helper methods such as `info` to view the current set of model options. 

 **Motivation**
Makes it easier to set model options without having to look through the documentation. Also creates a natural place in the documentation for putting model options (instead of base_battery_model).

priyanshuone6:
# Description
Create class Options to modularize model options

Fixes #894 

## Type of change

# Key checklist:

## Further checks:


priyanshuone6:
Adding `Options.print_options()`, `Options.print_detailed_options()` and `Options.update` functionalities are still remaining.

pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -474,6 +358,140 @@ def options(self, extra_options):
                 )
             )
 
+        super().__init__(options.items())
+
+    def print_options(self):
+        print(self.items())
```
tinosulzer, line 364: For this part I was thinking the options class could have a `possible_options` attribute which would look something like
```python3
self.possible_options = {
"surface form": ["false", "differential", "algebraic"], 
"thermal": ["isothermal", "lumped", "x-lumped", "x-full"],
...
}
```
And then instead of having to check each option individually, we can automate that check, something like
```python3
for option, value in self.options.items():
    if value not in self.possible_options[option]:
         raise pybamm.OptionError(f"{option} '{value}' not recognized. Possible options are {self.possible_options[option]}")
```
Then `print_options` should print both the possible options and which one was selected, e.g.
```python3
 {
"surface form": "false" (possible: ["false", "differential", "algebraic"]), 
"thermal": "lumped" (possible: ["isothermal", "lumped", "x-lumped", "x-full"]),
...
}
```

priyanshuone6: @tinosulzer This is what I came up with
```
self.possible_options = {
"surface form": ["false", "differential", "algebraic"], 
"convection": ["none", "uniform transverse", "full transverse"], 
"current collector": ["uniform", "potential pair", "potential pair quite conductive"], 
"dimensionality": [0, 1, 2], 
"interfacial surface area": ["constant", "varying"], 
"thermal": ["isothermal", "lumped", "x-lumped", "x-full"], 
"cell geometry": ["arbitrary", "pouch"], 
"SEI": ["none", "constant", "reaction limited", "solvent-diffusion limited", "electron-migration limited", "interstitial-diffusion limited", "ec reaction limited"], 
"SEI film resistance": ["none", "distributed", "average"], 
"SEI porosity change": ["true", "false"], 
"lithium plating": ["none", "reversible", "irreversible"], 
"loss of active material": ["none", "negative", "positive", "both"], 
"operating mode": ["current", "voltage", "power"], 
"particle cracking": ["none", "no cracking", "negative", "positive", "both"], 
"particle": ["Fickian diffusion", "fast diffusion", "uniform profile", "quadratic profile", "quartic profile"], 
"particle shape": ["spherical", "user", "no particles"], 
"electrolyte conductivity": ["default", "full", "leading order", "composite", "integrated"]
}
```
I have few doubts regarding the `possible_options`
- The doc for `electrolyte conductivity` is not written but the code is present, so should this be included?
- What are the possible options for `"side reactions"` and `"external submodels"`
- Possible options for `"total interfacial current density as a state"`, it is false by default and is dependent on ` "SEI film resistance"`
- `"operating mode"` has a `callable(options["operating mode"])` in it, if we automate the check, it may cause errors

tinosulzer: That looks good
- Need to check with @brosaplanella for the electrolyte conductivity
- You can skip the check for side reactions and external submodels
- "total interfacial current density as a state" can be "true" or "false"
- You can have a separate check for callable with operating mode (... `or option == "operating mode" and callable(value)` ...)

tinosulzer (approved):
Looks great, thanks @priyanshuone6 !

PR #1421, opened 2021-03-07
Fixed bcs for the particle models

### LINKED ISSUE #1420: Fix definition of bcs in particle models
**Describe the bug**
With the current implementation the particle model doesn't conserve mass when the diffusion coefficient is not constant. This is because the diffusivity at the boundary is evaluated as `D(pybamm.surf(c))` which is inconsistent with the definition in the divergence `pybamm.surf(D(c))`. 

We basically need to change `D(pybamm.surf(c))` into `pybamm.surf(D(c))`.

brosaplanella:
# Description

Changed how `D` is evaluated at the boundary condition to ensure that mass is conserved. I also changed some of the `standard_output_tests` to catch when mass is not conserved (before it didn't) and fixed some others.

Fixes #1420 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/tests/integration/test_models/standard_output_tests.py
+++ b/tests/integration/test_models/standard_output_tests.py
@@ -265,11 +265,14 @@ class ParticleConcentrationTests(BaseOutputTest):
         self.N_s_n = solution["Negative particle flux"]
         self.N_s_p = solution["Positive particle flux"]
 
-        self.n_SEI_n_av = solution[
-            "X-averaged negative electrode SEI concentration [mol.m-3]"
+        self.c_SEI_n_tot = solution["Loss of lithium to negative electrode SEI [mol]"]
+        self.c_SEI_p_tot = solution["Loss of lithium to positive electrode SEI [mol]"]
+
+        self.c_Li_n_tot = solution[
+            "Loss of lithium to negative electrode lithium plating [mol]"
         ]
-        self.n_SEI_p_av = solution[
-            "X-averaged positive electrode SEI concentration [mol.m-3]"
+        self.c_Li_p_tot = solution[
+            "Loss of lithium to positive electrode lithium plating [mol]"
         ]
 
     def test_concentration_increase_decrease(self):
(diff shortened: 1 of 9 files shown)


tests/integration/test_models/standard_output_tests.py
```diff
@@ -293,11 +293,11 @@ def test_concentration_increase_decrease(self):
             )
 
         if self.operating_condition == "discharge":
-            np.testing.assert_array_less(neg_end_vs_start, 0)
-            np.testing.assert_array_less(-pos_end_vs_start, 0)
+            np.testing.assert_equal(np.any(neg_end_vs_start <= 0), True)
```
brosaplanella, line 296: Sometimes it is equal to 0. Not 100% sure why (probably diffusion gets very low) but I think that the new constraint makes sense physically. Thoughts?

tests/integration/test_models/standard_output_tests.py
```diff
@@ -328,7 +328,14 @@ def test_conservation(self):
             + self.n_SEI_p_av(self.solution.t) * L_p * A
         )
         diff = (self.c_s_tot[1:] - self.c_s_tot[:-1]) / self.c_s_tot[:-1]
-        np.testing.assert_array_almost_equal(diff, 0)
+        if "profile" in self.model.options["particle"]:
+            np.testing.assert_array_almost_equal(diff, 0, decimal=10)
+        elif self.model.options["surface form"] == "differential":
+            np.testing.assert_array_almost_equal(diff, 0, decimal=10)
+        elif self.model.options["SEI"] == "ec reaction limited":
```
brosaplanella, line 335: I am not 100% that the `ec reaction limited` is conservative. In any case, with this new tolerance it should be fine.

tinosulzer: need to think a little bit more carefully about factors of 2 in the SEI layer growth equation

tests/integration/test_models/standard_output_tests.py
tinosulzer: So the concentration at the end is the same as it was at the start? That seems weird. Also, a nicer way to test "less than or equal to zero" would be to test "less than a small number (e.g. 1e-6)"

tests/integration/test_models/standard_output_tests.py
```diff
@@ -315,20 +324,23 @@ def test_concentration_limits(self):
...
+            + self.c_Li_n_tot(self.solution.t)
+            + self.c_Li_p_tot(self.solution.t)
         )
         diff = (self.c_s_tot[1:] - self.c_s_tot[:-1]) / self.c_s_tot[:-1]
-        np.testing.assert_array_almost_equal(diff, 0)
+        if "profile" in self.model.options["particle"]:
+            np.testing.assert_array_almost_equal(diff, 0, decimal=10)
+        elif self.model.options["surface form"] == "differential":
+            np.testing.assert_array_almost_equal(diff, 0, decimal=10)
+        elif self.model.options["SEI"] == "ec reaction limited":
```
tinosulzer, line 340: shouldn't this be fine up to 1e-15 now that you are using the ""Loss of lithium to negative electrode SEI [mol]" variable?

tinosulzer (approved):
Thanks @brosaplanella !

tests/integration/test_models/standard_output_tests.py
brosaplanella: Now it is of the order of 1e-13 (it went down from something like 1e-9) so it is a significant improvement.

PR #1422, opened 2021-03-08
Wrong argument type in `ParameterValues.update`

DavidMStraub:
# Description

I didn't open an issue since this is such a tiny change: in the constructor of `ParameterValues`, if the `values` argument is neither `None` nor a string, the `path` argument that is fed to `ParameterValues.update` is set to `None`. However this is incorrect as that method expects it to be a string; passing `None` will lead to a `ValueError` in `os.path.join`.

This happened to me when creating a dictionary of parameter values where one value was `[function]/absolute/path/to/function`. With this fix, it works.


rtimms (approved):
thanks @DavidMStraub !

PR #1423, opened 2021-03-10
Newman tobias model

rtimms:
# Description

Adds a new lithium-ion model based on the formulation in _JS Newman and CW Tobias. "Theoretical Analysis of Current Distribution in Porous Electrodes". Journal of The Electrochemical Society,1962_. 

The model assumes uniform (constant) concentration in the electrolyte and includes a single ODE in each electrode for the average particle concentration. Options can be passed to allow different particle models, but there is only ever a single particle in each electrode (i.e. we solve for the x-averaged concentration).

## Type of change

# Key checklist:

## Further checks:


pybamm/models/submodels/electrolyte_diffusion/composite_diffusion.py
```diff
@@ -95,14 +95,3 @@ def set_initial_conditions(self, variables):
         c_e = variables["Electrolyte concentration"]
 
         self.initial_conditions = {c_e: self.param.c_e_init}
-
-    def set_boundary_conditions(self, variables):
```
tinosulzer, line 99: I think at some point we moved these into the submodels to be clearer what the submodels are. It's not too much repeated code.

tests/integration/test_models/test_full_battery_models/test_lithium_ion/test_dfn.py
```diff
@@ -168,6 +168,10 @@ def positive_radius(x):
 
         param["Negative particle radius [m]"] = negative_radius
         param["Positive particle radius [m]"] = positive_radius
+        # Only get 3dp of accuracy in some tests at 1C with particle distribution
+        # TODO: investigate if there is a bug or some way to improve the
+        # implementation
```
tinosulzer, line 173: is this related to what @brosaplanella was seeing?

tinosulzer (approved):
Looks great, thanks!

pybamm/models/submodels/electrolyte_diffusion/composite_diffusion.py
rtimms: Makes sense, I will put them back. I did it this way as you also need to provide boundary conditions for the constant concentration model so that `grad(c_e)` ends up having the correct shape.

tests/integration/test_models/test_full_battery_models/test_lithium_ion/test_dfn.py
rtimms: looks like it. you can get 4dp of accuracy everywhere else.

PR #1432, opened 2021-03-15
Fix sundials download link

tlestang:
The website for sundials changed, and the download url with it.
Thanks @brosaplanella for reporting the issue


brosaplanella (approved):
Looks good, thanks Thibault!

PR #1444, opened 2021-03-25
Lithium plating

DrSOKane:
# Description

-A thermal prefactor has been added to the reversible and irreversible plating submodels, allowing them to be used for models where the temperature varies
-These models now use delta_phi as an input instead of phi_s and phi_e, which should lead to minor performance improvements
-The "Lithium plating kinetic rate constant [m.s-1]" parameter has been restored and integrated into the new framework, so people who aren't plating experts can adjust it.

## Type of change

# Key checklist:

## Further checks:


tinosulzer (approved):
Thanks @DrSOKane , changes look good, not immediately obvious to me why the test is failing

(approving but should fix test before merging)

tinosulzer:
Can you merge develop? It should get rid of all the lines in the unit test so we can see what's going on

PR #1454, opened 2021-03-30
Added warning for 'casadi' format

### LINKED ISSUE #887: add warning to klu if using model converted to casadi
the `IDAKLUSolver` can be considerably slower when using `model.convert_to_format = "casadi"` compared with `model.convert_to_format = "python"`, we should add a warning and/or explain in the docs for the solver that we recommend converting to python for this solver

Saransh-cpp:
# Description

Added warning to Idaklu solver if the model is converted to the format "casadi". 

A question - should I add
```py
if self.root_method != "CasadiAlgebraicSolver":
```
condition before adding the warning?

I also looked in the unit tests directory but couldn't find a suitable place for adding any sort of test for this warning (Please let me know where to put the test, if it is required)

Fixes #887 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/solvers/idaklu_solver.py
+++ b/pybamm/solvers/idaklu_solver.py
@@ -152,6 +152,13 @@ class IDAKLUSolver(pybamm.BaseSolver):
         if model.rhs_eval.form == "casadi":
             # stack inputs
             inputs = casadi.vertcat(*[x for x in inputs_dict.values()])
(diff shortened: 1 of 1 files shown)


tinosulzer (approved):
Thanks :) No need to add the casadi root solver check since the message mentions it already. This will probably already be caught in the coverage so no need to add a test for now

PR #1457, opened 2021-04-08
Type of Driving Cycle ('A', 'V', 'W') in Experiment Class

alibh95:
# Description

The experiment.py file is modified to read the operating conditions string which includes the type of driving cycle ('A', 'V', 'W').
This change is suggested by @tinosulzer in #1193. 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/experiments/experiment.py
+++ b/pybamm/experiments/experiment.py
@@ -174,16 +174,20 @@ class Experiment:
                 end_time = self.convert_time_to_seconds(cond_list[idx + 1 :])
                 ext_drive_cycle = self.extend_drive_cycle(drive_cycles[cond_list[1]],
                                                           end_time)
-                interp = ext_drive_cycle[:, 1] # Electric Part of the Drive Cycle as Numpy Array
-                typ =  cond_list[2][1] # Find the Type of Drive Cycle ("A", "V", or "W")
+                # Electric Part of the Drive Cycle as Numpy Array
+                interp = ext_drive_cycle[:, 1]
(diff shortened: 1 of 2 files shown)


Saransh-cpp:
Hi @alibh95, I was also working on this issue and most of the code in this PR is almost the same (even better I'd say) as the one I added in my local files, one addition that I had that is missing here was a value error for the user if the type of data is not specified. Something like this - 
```py
raise ValueError(
                    """type of data provided using a drive cycle should be
                    specified using 'A', 'V' or 'W'.
                    For example: {}
                    """.format(
                        examples
                    )
                )
```
Maybe we can add something like this?

As, for example, if an experiment `Run US06` is simulated, the code gives an error  -
```terminal
typ = cond_list[2][1]
IndexError: list index out of range
```
because of the absence of a specifier like `(A)`

alibh95:
@tinosulzer to create an interpolant in the experiment class, we have to pass the model into it. 
As the model is already passed into the simulation class. I think it would be better to create interpolant in that class.

alibh95:
@Saransh-cpp I didn't know that you were also working on this part of #1193. That's a great suggestion to raise error if the type of driving cycle is not provided.

tinosulzer:
Ok - good point about where you can define the interpolant. In that case you should also store the time points to pass to the simulation so it can create the interpolant

alibh95:
@tinosulzer Kindly review this PR.

tinosulzer (approved):
Thanks @alibh95 , merging

PR #1461, opened 2021-04-14
#1460 update jax versions

### LINKED ISSUE #1460: update pinned jax versions
update to 
        "jax==0.2.12",
        "jaxlib==0.1.65",

martinjrobins:
# Description

Fixes #1460

Question, does this need an addition in the `CHANGELOG.md`?

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/solvers/jax_bdf_solver.py
+++ b/pybamm/solvers/jax_bdf_solver.py
@@ -9,9 +9,11 @@ from jax import dtypes
 from jax.util import safe_map, cache, split_list
 from jax.api_util import flatten_fun_nokwargs
 from jax.flatten_util import ravel_pytree
-from jax.tree_util import tree_map, tree_flatten, tree_unflatten, tree_multimap, partial
+from jax.tree_util import tree_map, tree_flatten, tree_unflatten, tree_multimap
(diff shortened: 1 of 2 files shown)


pybamm/solvers/jax_bdf_solver.py
```diff
@@ -26,6 +27,45 @@
 MAX_FACTOR = 10
 
 
+# https://github.com/google/jax/issues/4572#issuecomment-709809897
+def some_hash_function(x):
+    return int(onp.sum(x))
```
tinosulzer, line 32: an alternative is to use `to_bytes`, see https://github.com/pybamm-team/PyBaMM/blob/e025cedfd3de4d1185d88adf3c7e22921bcca3ae/pybamm/expression_tree/array.py#L70-L86

tinosulzer (approved):
thanks @martinjrobins !

pybamm/solvers/jax_bdf_solver.py
martinjrobins: yea, good point. I think `tobytes` is better, so swapped to that.

PR #1471, opened 2021-04-28
Add parameterization example notebook

### LINKED ISSUE #1458: Parameterization example
Write a new example for parameterization showing:
- where the different parameter examples are
- how to find what are the parameters in a model
- how to add your own parameter set (in a dictionary)
- how to plot parameter functions

Saransh-cpp:
# Description

Added an example notebook explaining different concepts of parameterization using a simple model, and then with `DFN` and `SPM` models.

Overall the example is getting pretty big, please let me know if there are some steps that can be omitted or some steps that I missed and should be added.

View the notebook on my fork [here](https://github.com/Saransh-cpp/PyBaMM/blob/issue-1458-parameterization-example/examples/notebooks/parameterization.ipynb)

Fixes #1458 

## Type of change

# Key checklist:

## Further checks:


tinosulzer:
Thanks @Saransh-cpp , I pushed some updates onto the same branch in the main repo, can you see what you think?

Saransh-cpp:
@tinosulzer yes, adding a function parameter to the model defined, using `print_parameter_info` and setting up the `Chen2020` parameter values manually first and then using the one available in `pybamm` does make it more clear.

tinosulzer:
Ok, if you merge those changes into your branch I will approve and merge

Saransh-cpp:
@tinosulzer I have updated the example and the changes mentioned by @rtimms have also been added to this PR. Could you please let me know why the tests are failing? I tried fixing them locally but couldn't.

pybamm/models/base_model.py
```diff
@@ -274,37 +275,60 @@ def timescale(self, value):
...
+        """Find all the instances of `typ` in the model"""
+        unpacker = pybamm.SymbolUnpacker(typ)
         all_input_parameters = unpacker.unpack_list_of_symbols(
             list(self.rhs.values())
             + list(self.algebraic.values())
             + list(self.initial_conditions.values())
+            + [
+                x[side][0]
+                for x in self.boundary_conditions.values()
+                for side in ["left", "right"]
```
rtimms, line 326: I think the issue with the tests is here. Most models give boundary conditions using "left" and "right" as the keys, but some models use "positive tab", "negative tab" or "no tab", so I guess here we need to check for all these keys (and check if they are provided) in the boundary conditions dicts

Saransh-cpp: This is a rough code, when added, passes all the tests except 1 - 
```py
keys = self.boundary_conditions.values()
        side_list = []
        flag = True
        for key in keys:
            if isinstance(key, dict):
                if "positive tab" in key.keys():
                    side_list.append("positive tab")
                    flag = False
                if "negative tab" in key.keys():
                    side_list.append("negative tab")
                    flag = False
                if "no tab" in key.keys():
                    side_list.append("no tab")
                    flag = False

        if flag:
            side_list = ["left", "right"]
```
Editing `for side in ["left", "right"]` to `for side in side_list`

The one not passing is `test_plot_2plus1D_spm (test_plotting.test_quick_plot.TestQuickPlot` giving the error `KeyError: 'negative tab'`

Should it be checked like this or is there a better alternative?

rtimms: would this do the same?
```python3
[
                x[side][0]
                for x in model.boundary_conditions.values()
                for side in x.keys()
]
```

Saransh-cpp: The tests did pass locally using this, pushed the changes. Thanks!

examples/notebooks/parameterization.ipynb
```diff
@@ -0,0 +1,2439 @@
...
+   ],
+   "source": [
+    "spm.print_parameter_info()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Note that there are no `InputParameter` objects in the default SPM"
```
tinosulzer, line 438:
```suggestion
    "Note that there are no `InputParameter` objects in the default SPM. Also, note that if a `FunctionParameter` is expected, it is ok to provide a scalar (parameter) instead. However, if a `Parameter` is expected, you cannot provide a function instead."
```

pybamm/models/base_model.py
```diff
@@ -274,37 +275,60 @@ def timescale(self, value):
     def parameters(self):
         """Returns all the parameters in the model"""
         if self._parameters is None:
-            self._parameters = self._find_parameters()
+            self._parameters = self._find_symbols(
+                (pybamm.Parameter, pybamm.InputParameter)
```
rtimms, line 279: should this find `pybamm.FunctionParameter`s too?

Saransh-cpp: Oh yes, it should, that will make `print_parameter_info` and `parameter`  consistent with each other

tests/unit/test_solvers/test_algebraic_solver.py
```diff
@@ -295,6 +295,7 @@ def test_solve_with_input(self):
         # Solve
         solver = pybamm.AlgebraicSolver()
         solution = solver.solve(model, np.linspace(0, 1, 10), inputs={"value": 7})
+        model.print_parameter_info()
```
tinosulzer, line 298: I think this line can be deleted

tinosulzer (approved):
Thanks @Saransh-cpp , looks good!

tests/unit/test_solvers/test_algebraic_solver.py
tinosulzer: Both the property `model.parameters` and the function `model.print_parameter_info()` use caching to avoid having to recalculate the parameters when the function is called several times, see for example https://github.com/pybamm-team/PyBaMM/pull/1471/files#diff-579718d1877c511c7d203b3e4dfbbb093d1fb0b728efeb2069431ee27aa57250R275-R281 (the first time, `self._parameters` is None and then gets updated, the second time it already has been updated and so doesn't change).
Therefore this doesn't keep track of any changes in the model and just returns whatever it was the first time you called it.
This is probably overkill as this function will only be used in model building and model exploration, so doesn't need to be optimized like this. So can you edit the code to remove the caching?

tinosulzer: Regarding the original comment, it would be clearer if the line
```python
if input_param.domain == []:
    self._parameter_info += f"{input_param.name} (InputParameter)\n"
```
was tested by test_base_model (i.e. add an inputparameter without domain to that test model)

Saransh-cpp: I've made the necessary changes, I think everything should work fine now

tinosulzer: thanks! merging

PR #1472, opened 2021-04-29
#1465 allow evaluate parameter to return arrays

### LINKED ISSUE #1465: Plot parameter functions
Some parameters are provided as functions (e.g. diffusivity as a function of concentration and temperature). Currently, to plot them you have to do something like this 
```python3
import pybamm

params = pybamm.ParameterValues(chemistry=pybamm.parameter_sets.Chen2020)
i0 = params["Negative electrode exchange-current density [A.m-2]"]
x = pybamm.linspace(3000,6000,100)
i0_eval = i0(1000,x,300)
i0_processed = params.process_symbol(i0_eval)
pybamm.plot(x, i0_processed)
```
We have to process the function "Negative electrode exchange-current density [A.m-2]" since its definition contains a `pybamm.Parameter`. It would be neat to be able to do something like `params.plot("Negative electrode exchange-current density [A.m-2]", (1000, x, 300))` instead so you can quickly plot the functional dependence of parameters. We can use `pybamm.plot` and `pybamm.plot2D` to plot the resulting arrays.

Maybe you could pass a list or dict of parameter names and values so you can plot multiple parameters on the same plot.

rtimms:
# Description

`ParameterValues.evaluate()` can now return arrays so that you can evaluate parameter functions as 
```python3
import pybamm

params = pybamm.ParameterValues(chemistry=pybamm.parameter_sets.Chen2020)
i0 = params["Negative electrode exchange-current density [A.m-2]"]
x = pybamm.linspace(3000,6000,100)
params.evaluate(i0(1000,x,300))
```

Also, `plot` and `plot2D` now take and return the axis `ax` for easier customization
. 
Fixes #1465 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Features
 
+-   `plot` and `plot2D` now take and return a matplotlib Axis to allow for easier customization ([#1472](https://github.com/pybamm-team/PyBaMM/pull/1472))
+-   `ParameterValues.evaluate` can now return arrays to allow function parameters to be easily evaluated ([#1472](https://github.com/pybamm-team/PyBaMM/pull/1472))
 -   Added Batch Study class ([#1455](https://github.com/pybamm-team/PyBaMM/pull/1455))
 -   Added `ConcatenationVariable`, which is automatically created when variables are concatenated ([#1453](https://github.com/pybamm-team/PyBaMM/pull/1453))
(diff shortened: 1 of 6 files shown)


tinosulzer (approved):
Looks great, thanks @rtimms ! @Saransh-cpp can you also edit the example in #1458 to use this?

tinosulzer:
just needs a changelog before merging

PR #1475, opened 2021-04-30
Make parameters importable

### LINKED ISSUE #960: Loading simulations
When loading a simulation using `sim.load("something.p")` I get the error:
```
No module named 'graphite_diffusivity_Kim2011'
```
But if I add the lines:
```
chemistry = pybamm.parameter_sets.NCA_Kim2011
```
before loading the simulation, it loads and works fine. 

Can anyone else can reproduce this? Might just be an issue with my own python setup.

priyanshuone6:
# Description

- Modify load_function as it imports the function without the complete module path and that causes issues while unpickling. It also removes the (incomplete) module path from `sys.modules` to avoid namespace clashes.
- Edit tests
- Add `__init__.py` in parameters to convert them to modules and make them importable through pybamm

Fixes #960

## Type of change

# Key checklist:

## Further checks:


tinosulzer (approved):
Thanks @priyanshuone6 , changes look good. Is there a way to test that saving and loading a sim now works in the case it didn't before?

pybamm/util.py
```diff
@@ -227,95 +226,48 @@ def __eq__(self, other):
...
-            raise ValueError(
-                "{} is an absolute path, but the file is not found".format(filename)
-            )
-
-        valid_filename = filename
+    _, tail = os.path.split(filename)
 
-    # Else, search in the whole PyBaMM directory for matches
+    if "PyBaMM" in filename:
+        root_path = filename[filename.rfind("PyBaMM") :][7:]
```
tlestang, line 262: Maybe worth adding a comment explaining what this does and why?
Is it because you want to strip the full path to the module from `../PyBaMM/pybamm/input/example.py` to `pybamm/input/example.py` ?

tests/unit/test_simulation.py
```diff
@@ -265,6 +265,20 @@ def test_save_load(self):
         ):
             sim.save("test.pickle")
 
+    def test_load_param(self):
+        # Test load_sim for parameters imports
+        pkl_obj = pybamm.load_sim(
+            os.path.join(
+                "tests", "unit", "test_parameters", "data", "test_load_param.p"
```
tlestang, line 272: You committed the pickle file `test_load_param.p` but we don't really know what it corresponds to. How about, instead, this tests saves a simulation before trying to load it?

pybamm/util.py
priyanshuone6: Yes, I'll add a comment

tests/unit/test_simulation.py
priyanshuone6: @tlestang The bug is reproducible only when the pickle file is loaded afresh in a new terminal. If we'll save and then load it then the parameter will already be imported and the error won't occur.

pybamm/util.py
tlestang: So that works if `pybamm` is imported from a directory named `PyBaMM` (which is indeed default when cloning from github). But what happens if `pybamm` is imported from its installed location (e.g. `../site-packages/pybamm`)? This would be the case when installing PyBaMM with pip instead of importing it directly from the sources

tests/unit/test_simulation.py
tlestang: Ah yes. Alternatively we could write the pickle from another process, something like
```
save_sim = "import pybamm; model = pybamm.lithium_ion.SPM(); pybamm.Simulation(model).save("test_load_param.p")"
subprocess.run("python -c {save_sim}")
```
Relying on the binary pickle only may make debugging hard in the future, when nobody remembers what is contains :) 
Which makes me think another option is adding a comment that describes how you generated the pickle file.

priyanshuone6:
@tlestang Thank you for all the suggestions, I have added the subprocess.run in the test and also tested it with the develop branch and it worked.

tests/unit/test_simulation.py
```diff
@@ -265,6 +268,26 @@ def test_save_load(self):
...
 
+    def test_load_param(self):
+        # Test load_sim for parameters imports
+        filename = f"{uuid.uuid4()}.p"
+        save_sim = f"import pybamm; model = pybamm.lithium_ion.SPM(); params = pybamm.ParameterValues(chemistry=pybamm.parameter_sets.Chen2020); sim = pybamm.Simulation(model, parameter_values=params); sim.solve([0, 3600]); sim.save('{filename}')"  # noqa
+        subprocess.run([sys.executable, "-c", save_sim])
+
+        try:
+            pkl_obj = pybamm.load_sim(os.path.join(filename))
+        except Exception as excep:
```
tlestang, line 279: Could you use a specific pickle-related exception class instead of the general `Exception`?

tlestang:
Looks good @priyanshuone6 , nice one :)

PR #1476, opened 2021-05-03
Issue 1456 concentration

### LINKED ISSUE #1456: Replace "concentration" variable with "porosity times concentration"
Since "porosity times concentration" is actually what is conserved in the electrolyte diffusion equation, it might make more sense to have that be the variable, instead of "concentration"

tinosulzer:
# Description

Makes "porosity times concentration" the variable in the full diffusion model. This could also be changed in the other diffusion models, but I don't need it right now and am lazy. At some point we will need to change the diffusion model for the SPM to account for the effect of porosity changes on average concentration.

Fixes #1456 

## Type of change

# Key checklist:

## Further checks:


rtimms (approved):
thanks @tinosulzer !

PR #1480, opened 2021-05-07
Issue 1478 ensemble inputs casadi

TomTranter:
# Description

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:


pybamm/solvers/casadi_solver.py
```diff
@@ -499,6 +515,8 @@ def create_integrator(self, model, inputs, t_eval=None, use_event_switch=False):
                     method, problem, options = self.integrator_specs[model]
                     options["grid"] = t_eval_shifted
                     integrator = casadi.integrator("F", method, problem, options)
+                    if ninputs > 1:
+                        integrator = integrator.map(ninputs, 'thread', nproc)
```
tinosulzer, line 519: This line should be tested. You need to solve the same model a second time with different times (i.e. an experiment) to enter this branch

pybamm/solvers/casadi_solver.py
```diff
@@ -591,25 +611,48 @@ def _run_integrator(self, model, y0, inputs_dict, inputs, t_eval, use_grid=True)
...
+                    sol.integration_time = integration_time
+                else:
+                    nt = len(t_eval)
+                    xf = casadi_sol["xf"]
+                    # zf = casadi_sol["zf"]
+                    sol = []
+                    for i in range(N):
+                        start = i * nt
+                        y_sol = xf[:, start:start + nt]
+                        # Not sure how to index into zf - need an example
```
tinosulzer, line 647: it should just be the same way as xf?

pybamm/solvers/casadi_solver.py
```diff
@@ -591,25 +611,48 @@ def _run_integrator(self, model, y0, inputs_dict, inputs, t_eval, use_grid=True)
         len_rhs = model.concatenated_rhs.size
         y0_diff = y0[:len_rhs]
         y0_alg = y0[len_rhs:]
+        in_size = inputs.shape[1]
+        if in_size > 1:
+            N = in_size
```
tinosulzer, line 616: why not just use `in_size` everywhere instead of renaming to `N`?

pybamm/solvers/casadi_solver.py
```diff
@@ -104,7 +104,7 @@ def __init__(
 
         pybamm.citations.register("Andersson2019")
 
-    def _integrate(self, model, t_eval, inputs_dict=None):
+    def _integrate(self, model, t_eval, inputs_dict=None, ninputs=1, nproc=1):
```
tinosulzer, line 107: Why have a separate `ninputs` here instead of reading the length of `inputs_dict`?
Also if `inputs_dict` can now be a list or a dict, we should rename it to avoid confusion (maybe even just to `input_dicts`?)

pybamm/solvers/casadi_solver.py
TomTranter: I don't think experiments work with multidimensional inputs due to the event handling - I can try to get it to work but looks complex. Might be simpler to implement parallel processing outside of the solver rather than using the map function from casadi.

TomTranter: @tinosulzer this looks too hacky to implement now with the way that experiments call the solver step function. It would be worth doing though I think. Might be worth thinking about a refactor of solver, simulation, experiment.

tinosulzer: Oh yeah you're right it won't work with experiment. I don't see how it could work but maybe I don't know about some casadi features. Any specific reason for a refactor?

TomTranter: Well the simplest thing would be to parallelize steps which means the experiment would have to control the solver but it kinda seems like it's the other way round... I am only really just looking at the solver codes in detail now but they are complicated and it might be possible to make them simpler with a refactor too but it's probably  a lot of work

tinosulzer: The casadi solver code is really complicated because of events

TomTranter: So we can get this code running without experiments or shelve it until casadi handles events

TomTranter:
@tinosulzer I think this feature is now addressed by liionpack. I'd be happy to get the code running without experiments in PyBaMM but it might be cleaner to just leave this feature out of PyBaMM and delete this branch. Up to you?

tinosulzer:
I'll take a look at how it's done in liionpack, ideally we could reimplement in pybamm and call that from liionpack?

TomTranter:
Not sure that would work to be honest. The solver in liionpack is a lot more stripped down. The code you'd need to look at is `-`_mapped_step` [here](https://github.com/pybamm-team/liionpack/blob/main/liionpack/solver_utils.py)

kratman:
@TomTranter Can this be closed?

rtimms:
Closing as this has been hanging around a while. Can reopen if needed.

PR #1490, opened 2021-05-21
Issue 1405 xaverage

### LINKED ISSUE #1405: Use x-averaged SEI models for SPM and SPMe
SPM and SPMe use the full SEI (and plating) submodels which unnecessarily repeats states.
They could instead use "x-averaged" submodels.
This could be passed as an option to the existing submodels to avoid having to rewrite the whole submodel twice.
Similarly the LAM submodel could be formatted as a single submodel using an option

tinosulzer:
# Description

Reformats the degradation models so that the x-averaged form can be used
Adds a reaction-driven LAM model from Reniers 2019

Fixes #1405 

## Type of change

# Key checklist:

## Further checks:


pybamm/models/full_battery_models/lithium_ion/dfn.py
```diff
@@ -32,6 +32,7 @@ class DFN(BaseModel):
 
     def __init__(self, options=None, name="Doyle-Fuller-Newman model", build=True):
         super().__init__(options, name)
+        self.x_average = False
```
rtimms, line 35: might be worth a comment here explaining why `x_average` is True/False for different models (just in dfn, spm etc., not in every submodel)

pybamm/models/submodels/active_material/reaction_driven_active_material.py
```diff
@@ -0,0 +1,124 @@
...
+            beta_LAM_sei = self.param.beta_LAM_sei_p
+
+        deps_solid_dt = beta_LAM_sei * a * j_sei
+        variables.update(
+            self._get_standard_active_material_change_variables(deps_solid_dt)
+        )
+        return variables
+
+    def set_rhs(self, variables):
+        Domain = self.domain + " electrode"
```
rtimms, line 86: very minor style point, but odd for this to be capitalized. I guess to distinguish from `self.domain`

rtimms (approved):
neat! thanks, @tinosulzer !

PR #1496, opened 2021-05-27
add half cell notebook

### LINKED ISSUE #1127: Extra documentation and notebooks
Add notebooks on:
1. changing the mesh
2. changing solver options 
3. thermal models
4. simulation vs "the long way"
5. basic DFN /  show DFN in dimensional form and give parameter names (strings)

1 & 2 will show how to do a graphical convergence study (just showing how to plot the solutions for different numbers mesh points etc) and encourage users to do a proper convergence study when using pybamm. 

3 will give details of the governing equation used for each thermal option, along with the parameters used (in particular we need to explain the different cooling coefficients used in pouch/arbitrary geometries)

4 - I think we should simplify most/all existing examples to use the simulation class as it is much more straightforward and should be the preferred way to do things in most cases. then we can have a notebook showing using simulation and "the long way" so users know both approaches exist

5 - I think lots of people who are less familiar with modelling are likely to jump in with a basic model, and be more comfortable with dimensional models. We could use this notebook to give the basic DFN model in dimensional form and give the string names for all variables as used in pybamm. This way people can look here for the most common parameter names and get a really clear exposition of the model being solved. We can point out that the basic DFN is the same as using just DFN with the default options, and that they are both non-dimensionalalised using the scales in REF/table

Any others??

rtimms:
# Description

Adds a new half cell model notebook to demonstrate how to set up a model which has multiple domains.

Fixes #1127 

## Type of change

# Key checklist:

## Further checks:


tinosulzer:
is this ready for review?

rtimms:
yes please, forgot to request

tinosulzer (approved):
Looks good, thanks!

PR #1502, opened 2021-06-04
Add to_equation in Binary and Scalar

priyanshuone6:
# Description

- Add `to_equation` in Scalar, Binary Operators and Array
- Change `Heaviside` to `_Heaviside`
- Add tests

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/expression_tree/binary_operators.py
+++ b/pybamm/expression_tree/binary_operators.py
@@ -62,7 +62,8 @@ def get_binary_children_domains(ldomain, rdomain):
 
 
 class BinaryOperator(pybamm.Symbol):
-    """A node in the expression tree representing a binary operator (e.g. `+`, `*`)
+    """
+    A node in the expression tree representing a binary operator (e.g. `+`, `*`)
 
     Derived classes will specify the particular operator
 
@@ -77,7 +78,6 @@ class BinaryOperator(pybamm.Symbol):
         lhs child node (converted to :class:`Scalar` if Number)
     right : :class:`Symbol` or :class:`Number`
         rhs child node (converted to :class:`Scalar` if Number)
-
     """
 
     def __init__(self, name, left, right):
@@ -95,7 +95,7 @@ class BinaryOperator(pybamm.Symbol):
         self.right = self.children[1]
 
     def __str__(self):
-        """ See :meth:`pybamm.Symbol.__str__()`. """
+        """See :meth:`pybamm.Symbol.__str__()`."""
         # Possibly add brackets for clarity
         if isinstance(self.left, pybamm.BinaryOperator) and not (
             (self.left.name == self.name)
@@ -115,7 +115,7 @@ class BinaryOperator(pybamm.Symbol):
         return "{} {} {}".format(left_str, self.name, right_str)
 
     def new_copy(self):
-        """ See :meth:`pybamm.Symbol.new_copy()`. """
+        """See :meth:`pybamm.Symbol.new_copy()`."""
 
         # process children
         new_left = self.left.new_copy()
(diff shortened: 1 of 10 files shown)


tinosulzer (changes requested):
Looks good, you just need to add tests for the binary operators to equation. Also it looks like your formatter might be set to 79 characters for max line length? We use 88

tests/unit/test_expression_tree/test_binary_operators.py
```diff
@@ -668,6 +668,24 @@ def test_inner_simplifications(self):
...
+        # Test Matrix Multiplication
+        arr1 = pybamm.Array([[1, 0], [0, 1]])
+        arr2 = pybamm.Array([[4, 1], [2, 2]])
+        self.assertEqual(
+            pybamm.MatrixMultiplication(arr1, arr2).to_equation(),
+            sympy.Matrix([[4.0, 1.0], [2.0, 2.0]]),
+        )
+
+        # Test Heaviside
+        self.assertEqual(pybamm.Heaviside("test", 4, 2).to_equation(), 1)
```
priyanshuone6, line 687: The same code works on the jupyter notebook and gives `1` but here gives an error

tinosulzer: It works in notebook but not in test because these lines are only called in the tests when debug mode is true https://github.com/pybamm-team/PyBaMM/blob/08de41714aa093fa2b705a649d4132d3418d43b3/pybamm/expression_tree/symbol.py#L226-L232

tinosulzer: `_binary_evaluate` isn't implemented for `Heaviside`. Instead of calling `Heaviside` directly just test both `EqualHeaviside` and `NotEqualHeaviside`. Can you also change the class name `Heaviside` to `_Heaviside` to indicate that it is a semi-private class (shouldn't be called directly)?

priyanshuone6: Thanks, @tinosulzer, I have removed the `to_equation` from the Heaviside as it is a semi-private class and won't be called directly

tinosulzer: `to_equation` can still be in the Heaviside if the behavior of EqualHeaviside and NotEqualHeavside is intented to be the same (saves writing the same code twice). Then `EqualHeaviside.to_equation` will just call `_Heaviside.to_equation`

priyanshuone6: The `_binary_evaluate` in EqualHeaviside and NotEqualHeavside already does what `sympy.Heaviside` is intended to do, so I think we can probably skip overriding the `to_equation` for the _Heaviside class

pybamm/expression_tree/binary_operators.py
```diff
@@ -473,8 +500,9 @@ def inner(left, right):
     return pybamm.simplify_if_constant(pybamm.Inner(left, right))
 
 
-class Heaviside(BinaryOperator):
-    """A node in the expression tree representing a heaviside step function.
+class _Heaviside(BinaryOperator):
+    """
+    A node in the expression tree representing a heaviside step function.
```
tinosulzer, line 505:
```suggestion
    A node in the expression tree representing a heaviside step function.
    This class is semi-private and should not be called directly, use `EqualHeaviside`
    or `NotEqualHeaviside` instead, or `<` or `<=`.
```

pybamm/expression_tree/binary_operators.py
```diff
@@ -586,60 +614,60 @@ def __init__(self, left, right):
...
 
     def _binary_new_copy(self, left, right):
-        "See :meth:`pybamm.BinaryOperator._binary_new_copy()`. "
+        """See :meth:`pybamm.BinaryOperator._binary_new_copy()`."""
         return pybamm.minimum(left, right)
 
 
 class Maximum(BinaryOperator):
-    """Returns the smaller of two objects."""
+    """Returns the greater of two objects."""
```
tinosulzer, line 643: good catch! 😅

tinosulzer (approved):
Thanks @priyanshuone6 , just one suggestion

PR #1510, opened 2021-06-11
#1507 remove algebraic equation check

### LINKED ISSUE #1507: Voltage control when using "surface form"
If you try to use voltage control with the "surface form" option anything other than "false" you get the error `ModelError: each variable in the algebraic eqn keys must appear in the eqn` due to the check [here](https://github.com/pybamm-team/PyBaMM/blob/develop/pybamm/models/base_model.py#L655). If you comment out the check then it works fine. Should we relax this check further? 

The problem, in this case, is that the "Total current density" variable doesn't appear in the equation `V - pybamm.FunctionParameter("Voltage function [V]", {"Time [s]": pybamm.t * self.param.timescale})`

Full example:
 
```python3
import pybamm 

model = pybamm.lithium_ion.SPM(options={"surface form": "algebraic"})

for var in [
    "Terminal voltage", 
    "X-averaged positive electrode surface potential difference", 
    "X-averaged negative electrode surface potential difference",
]:
    print(model.variables[var])

#experiment = pybamm.Experiment(["Hold at 3.8 V until 10 mA"])
    [
        (
            "Discharge at C/5 for 10 hours or until 3.3 V",
            "Rest for 1 hour",
            "Charge at 1 A until 4.1 V",
            "Hold at 4.1 V until 10 mA",
            "Rest for 1 hour",
        ),
    ]
)

sim.solve()
```

rtimms:
# Description
Remove the check "each variable in the algebraic eqn keys must appear in the eqn"

Fixes #1507 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 
 ## Bug fixes
 
+-   Removed the overly-restrictive check "each variable in the algebraic eqn keys must appear in the eqn" ([#1510](https://github.com/pybamm-team/PyBaMM/pull/1510))
 -   Made parameters importable through pybamm ([#1475](https://github.com/pybamm-team/PyBaMM/pull/1475))
 
 ## Breaking changes
--- a/pybamm/models/base_model.py
+++ b/pybamm/models/base_model.py
@@ -639,40 +639,20 @@ class BaseModel(object):
 
     def check_algebraic_equations(self, post_discretisation):
         """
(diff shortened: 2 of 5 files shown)


tinosulzer (approved):
Cheers @rtimms . If you're feeling brave, there might be some hacks in the potential pair models that were done to pass this test and can now be reversed. No worries if not.

PR #1524, opened 2021-06-25
Solve Drive Cycle (A,V,W) in Experiment

alibh95:
# Description

PR Status: (Work in Progress)

This PR provides the functionality of solving the drive cycle using the experiment only if the first input of the experiment is the drive cycle. 

Issues in PR:

- When two consecutive drive cycles are given in the experiment, only the first one solves but the second one only returns 0 current. 

- If the drive cycle is not the first input of the experiment, the drive cycle part only returns 0 current. 

Possible Cause of this Issue: 

The driving cycles are given to the solver as interpolant. I think there is an issue with the timing of the interpolant which I am unable to overcome. For example, The US06 Drive cycle data have time range [0:600], When the drive cycle is the first input of the experiment is drive cycle (solution start time = 0), the cycle is solved successfully, but when the same cycle is repeated, (solution time > 600), the interpolant returns 0 A.

@tinosulzer and @Saransh-cpp suggestions on overcoming this issue are highly appreciated. 

Fixes # (issue)
#1279 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/simulation.py
+++ b/pybamm/simulation.py
@@ -170,44 +170,82 @@ class Simulation:
         self._experiment_inputs = []
         self._experiment_times = []
         for op, events in zip(experiment.operating_conditions, experiment.events):
-            if op[1] in ["A", "C"]:
-                # Update inputs for constant current
+            if isinstance(op[0], np.ndarray):
+                # self.operating_mode = "drive cycle" ########!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+                # If ndarray is recived from, create interpolant
+                # create interpolant
+                timescale = self._parameter_values.evaluate(model.timescale)
+                drive_cycle_interpolant = pybamm.Interpolant(
+                    op[0][:, 0], op[0][:, 1], timescale * pybamm.t
+                )
+                # print(type(drive_cycle_interpolant))  #!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
                 if op[1] == "A":
-                    I = op[0]
-                else:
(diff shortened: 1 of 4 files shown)


pybamm/simulation.py
```diff
@@ -170,44 +170,79 @@ def set_up_experiment(self, model, experiment):
...
         self._experiment_times = []
         for op, events in zip(experiment.operating_conditions, experiment.events):
-            if op[1] in ["A", "C"]:
-                # Update inputs for constant current
+            if isinstance(op[0], np.ndarray):
+                # If ndarray is recived from, create interpolant
+                # create interpolant
+                timescale = self._parameter_values.evaluate(model.timescale)
+                drive_cycle_interpolant = pybamm.Interpolant(
+                    op[0][:, 0], op[0][:, 1], timescale * pybamm.t
```
tinosulzer, line 178: The last input here is the time that is evaluated by the interpolant. You could do the following:
```python3
drive_cycle_interpolant = pybamm.Interpolant(
                    op[0][:, 0], op[0][:, 1], timescale * pybamm.t - pybamm.InputParameter("start time")
)
```
and pass the appropriate "start time" to the model inputs when solving the model at each step

brosaplanella:
Hi @alibh95! What is the status of this PR?

alibh95:
Dear @brosaplanella, 
Unfortunately I am unable to spare much time these days to work on this PR. However, I would be grateful if any other contributor works on this PR as it would be valuable addition to this project. 
Again, I am sorry if any I cause any inconvenience.

brosaplanella:
No worries! I can try to pick it up in a few days. What's the current status? Is it ready and just needs testing before merging or are there still features to be implemented?

alibh95:
This PR is currently incomplete and I haven't tried @tinosulzer's suggestion yet.

PR #1528, opened 2021-06-30
Add to_equation for SPM voltage

priyanshuone6:
# Description

- Add to_equation in Function, Arcsinh, Parameter, BoundaryValue
- Automate print_name using regex
- Add tests

## SPM model equations
spm-1

## SPM Voltage expression
spm-2

## Type of change

# Key checklist:

## Further checks:


pybamm/expression_tree/unary_operators.py
```diff
@@ -952,6 +952,9 @@ def _unary_new_copy(self, child):
         """See :meth:`UnaryOperator._unary_new_copy()`."""
         return boundary_value(child, self.side)
 
+    def _sympy_operator(self, child):
+        return child
```
tinosulzer, line 956: this should actually return an indication of the boundary value. The mathematical notation looks something like this:
Screen Shot 2021-06-30 at 2 08 56 PM
That's hard to replicate so we could instead do something like
```python3
if child.domain in ["negative particle", "positive particle"] and self.side == "right":
   return child + "^{surf}" # value on the surface of the particle
else:
   return child + f"^{self.side}"
```

tinosulzer (changes requested):
Looks good just one suggestion for boundary value

pybamm/models/standard_variables.py
```diff
@@ -1,9 +1,11 @@
 #
 # Standard variables for the models
 #
-import pybamm
 import numpy as np
 
+import pybamm
+from pybamm.expression_tree.print_name import prettify_print_name
```
brosaplanella, line 7:
```suggestion
import pybamm
import numpy as np
from pybamm.expression_tree.print_name import prettify_print_name
```

brosaplanella: Very minor cosmetic change, but to keep consistency across files

pybamm/parameters/lithium_ion_parameters.py
```diff
@@ -1,8 +1,10 @@
 #
 # Standard parameters for lithium-ion battery models
 #
-import pybamm
 import numpy as np
+
+import pybamm
+
 from .base_parameters import BaseParameters
```
brosaplanella, line 8:
```suggestion
import pybamm
import numpy as np
from .base_parameters import BaseParameters
```

tests/unit/test_expression_tree/test_functions.py
```diff
@@ -1,12 +1,14 @@
...
-
 import unittest
+
 import numpy as np
-from scipy.interpolate import interp1d
+import sympy
 from scipy import special
+from scipy.interpolate import interp1d
+
+import pybamm
```
brosaplanella, line 11: Same as my suggestions before (here I can't make a suggestion though because some lines appear as deleted)

tests/unit/test_expression_tree/test_functions.py
```diff
@@ -127,6 +129,21 @@ def test_function_unnamed(self):
             fun.name, "function (<class 'scipy.interpolate.interpolate.interp1d'>)"
         )
 
+    def test_to_equation(self):
+        a = pybamm.Symbol("a", domain="test")
+
+        # Test _sympy_operator
+        with self.assertRaises(NotImplementedError):
+            pybamm.Erf(a).to_equation()
```
brosaplanella, line 137: Is there a symbol that we know for sure will never be implemented? Because `Erf` should be fairly easy to bring in.

brosaplanella (approved):
Looks good to me, just a couple of cosmetic comments. Also, note that one test is not passing ATM (I think it is a minor thing though).

priyanshuone6:
Thanks, @brosaplanella, I added `Erf` temporarily to test the `NotImplementedError` as it does not have a `_sympy_operator` yet, I'll change it and put something else there.
I formatted all the imports using [isort](https://github.com/PyCQA/isort), I will revert them to as they were before.

pybamm/expression_tree/print_name.py
```diff
@@ -0,0 +1,87 @@
...
+            r"([\da-zA-Z]+)_?(.*?)_?(?:dim|dimensional)", name
+        )[0]
+
+        if "^" in name:
+            name = r"\hat{" + dim_re1 + r"}" + dim_re2.replace("_", "\,")
+        else:
+            name = r"\hat{" + dim_re1 + r"}_{" + dim_re2.replace("_", "\,") + r"}"
+
+    # Bar with comma separated
+    # e.g. c_s_n_xav --> \bar{c}_{s\,n\,xav}
```
tinosulzer, line 77: remove xav from the subscript

pybamm/expression_tree/print_name.py
```diff
@@ -0,0 +1,87 @@
...
+    if any(elem in name for elem in ["_av", "_xav"]):
+        bar_re1, bar_re2 = re.findall(r"^([a-zA-Z]+)_*(\w*(?:av|xav))", name)[0]
+        name = r"\bar{" + bar_re1 + r"}_{" + bar_re2.replace("_", "\,") + r"}"
+
+    # Greek letters
+    # e.g. delta_phi_n --> \delta_\phi_n
+    greek_re = f"({'|'.join(GREEK_LETTERS)})"
+    name = re.sub(greek_re, r"\\\1", name, flags=re.IGNORECASE)
```
tinosulzer, line 85: case is important, if a greek letter starts with an uppercase (e.g. Delta) we should print the uppercase greek letter

pybamm/models/standard_variables.py
```diff
@@ -312,7 +313,7 @@ def __init__(self):
         )
 
     def __setattr__(self, name, value):
-        value.print_name = name
+        value.print_name = prettify_print_name(name)
```
tinosulzer, line 316: we could make `print_name` a property in `symbol` and then have the setter class for the property automatically call `prettify_print_name`

pybamm/expression_tree/print_name.py
```diff
@@ -0,0 +1,87 @@
...
+    "tau",
+    "upsilon",
+    "phi",
+    "chi",
+    "psi",
+    "omega",
+]
+
+
+def prettify_print_name(name):
```
tinosulzer, line 41: awesome!

tinosulzer (approved):
Great work - maybe the `print_name` function and also the sympy overload from #1532 could go into a new subfolder of expression_tree called `printing`?

pybamm/expression_tree/print_name.py
priyanshuone6: Yes, it will return `\Delta` if the name contains Delta and `\delta` if the name contains delta

pybamm/expression_tree/symbol.py
```diff
@@ -195,6 +196,8 @@ class Symbol(anytree.NodeMixin):
         collector"}`).
     """
 
+    _print_name = None
```
tinosulzer, line 199: I didn't actually know you could do this here 😆 but for consistency let's put it in the `__init__`

tinosulzer (approved):
Looks great!

PR #1556, opened 2021-07-20
#1554 fix indexing for FEM entries

### LINKED ISSUE #1554: Dimensions interchanged in 2D processed variable
**Describe the bug**

I'm plotting a 2D processed variable and finding that when I specify the spatial variables y and z, they seem to be interchanged.

**To Reproduce**

```python
import pybamm
options = {"cell geometry": "pouch", "thermal": "x-lumped", "dimensionality": 2, "current collector": "potential pair"}
model = pybamm.lithium_ion.SPMe(options=options)
experiment = pybamm.Experiment(["Discharge with 3 C for 10 seconds"])
solver = pybamm.CasadiSolver("fast")
sim = pybamm.Simulation(model, experiment=experiment, solver=solver)
sim.solve()
sim.plot(["X-averaged Ohmic heating [W.m-3]"], spatial_unit="m")
print(sim.solution["X-averaged Ohmic heating [W.m-3]"](t=10, y=0.147, z=0.137))
print(sim.solution["X-averaged Ohmic heating [W.m-3]"](t=10, y=0.207, z=0.1))
```

grafik

```
array(9830.35092739)
array(17052.65235706)
```

**Expected behaviour**

The first printed value should be directly on top of the positive tab, which - as expected and shown in the plot - has a larger heat generation. The second printed value, which has the large heat generation, is a (y, z) point on the right edge of the pouch. So it seems that y and z are interchanged.

rtimms:
# Description

Fixes the order of the indexing for the entries of variables solved using FEM

Fixes #1554

## Type of change

# Key checklist:

## Further checks:


tests/unit/test_plotting/test_quick_plot.py
```diff
@@ -429,6 +429,7 @@ def test_plot_2plus1D_spm(self):
         quick_plot.slider_update(1)
 
         # check 2D (y,z space) variables update properly for different time units
+        # Note: these should be the transpose of the entries in the processed variable
```
brosaplanella, line 432: I am confused: the comment says it should be the transpose but below it is not transposed

brosaplanella (approved):
Looks good, thanks Rob! Just see my comment on the commented line.

tests/unit/test_plotting/test_quick_plot.py
rtimms: In the test we do, e.g., `np.testing.assert_array_almost_equal(qp_data.T, phi_n[:, :, 0])` so that `qp_data` is transposed

PR #1578, opened 2021-07-28
try ignoring some files in coverage

tinosulzer:
# Description

Ignore some files not being covered

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -46,6 +46,10 @@ commands = python -m flake8
 deps = 
      coverage
      scikits.odes
+ignore = 
+     input
+     parameters_cli.py
+     install_odes.py
 commands = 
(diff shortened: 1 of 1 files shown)


brosaplanella (approved):
Looks good, thanks Tino!

tinosulzer:
It doesn't change coverage 😭

brosaplanella:
May it be because coverage is calculated vs the one on develop and there those files are not ignored?

brosaplanella:
This [StackOverflow post](https://stackoverflow.com/questions/1559424/coverage-py-exclude-files) suggest other ways, if I understand correctly

brosaplanella:
Also, checking how flake8 works, it could be that the command is `exclude` rather than `ignore`.

PR #1581, opened 2021-07-29
Issue 1571 1 d cc opposite tabs

### LINKED ISSUE #1571: 1D current collectors with opposite-sided tabs
**Summary**
 
Currently, the 1+1D pouch cell model only works for both cells on the top.

 **Motivation**

For pouch cells with tabs on opposite sides, the 1D model is particularly useful as they are symmetric in the y direction.

**Additional context**

https://github.com/pybamm-team/PyBaMM/blob/develop/pybamm/models/full_battery_models/base_battery_model.py#L302-L306

https://github.com/pybamm-team/PyBaMM/blob/develop/pybamm/models/submodels/thermal/pouch_cell/pouch_cell_1D_current_collectors.py#L98

rtimms:
# Description

Allows tabs to be placed at the bottom of the cell in 1+1D thermal models. Adds a new `Equality` binary operator that returns 1 if the nodes evaluate to the same thing and 0 otherwise.

Fixes #1571 

## Type of change

# Key checklist:

## Further checks:


tinosulzer (approved):
Nice. You could also add the `__eq__` magic function to `Symbol` to return `Equality(a,b)` for `a==b`

rtimms:
I did try adding `__eq__` but ran into an issue. In some places (e.g. [here](https://github.com/pybamm-team/PyBaMM/blob/develop/pybamm/expression_tree/binary_operators.py#L1019)) we use `==` to check if two symbols are equal. I guess we could check the symbol `id` here and use `==` for `Equality`?

tinosulzer:
Ah right, in that case we do want to check actual equality because otherwise the clear_domain a couple of lines below causes issues. So let's leave it

rtimms:
Hey, @tobykirk some of the many-particle submodel tests are failing on this PR, but only on certain builds (e.g. [here](https://github.com/pybamm-team/PyBaMM/pull/1581/checks?check_run_id=3239519517#step:9:2044)). They pass ok locally and I haven't changed anything related to them - have you seen this error before?

```
pybamm.expression_tree.exceptions.ShapeError: Cannot find shape (original error: zero-size array to reduction operation minimum which has no identity)
```

tobykirk:
> Hey, @tobykirk some of the many-particle submodel tests are failing on this PR, but only on certain builds (e.g. [here](https://github.com/pybamm-team/PyBaMM/pull/1581/checks?check_run_id=3239519517#step:9:2044)). They pass ok locally and I haven't changed anything related to them - have you seen this error before?
> 
> ```
> pybamm.expression_tree.exceptions.ShapeError: Cannot find shape (original error: zero-size array to reduction operation minimum which has no identity)
> ```

I've seen this error once before and it was similar to here: only occurring on certain builds, but I think they were different builds to these - weird. 

It resolved itself after I pushed some unrelated commits, so never got to the bottom of it. Looks like something to do with `set_events` of the many-particle submodels, i.e., the max / min concentration limits. Actually, are these enforced differently now? (I can't find any particle concentration events in the code for the other submodels.) If so, you could try removing those events.

rtimms:
Thanks, @tobykirk, removing `set_events` fixed this! As you say, they shouldn't be needed since the variable has bounds. It's weird that the test only failed sometimes, though?

tobykirk:
> Thanks, @tobykirk, removing `set_events` fixed this! As you say, they shouldn't be needed since the variable has bounds. It's weird that the test only failed sometimes, though?

Great! Strange error indeed, hopefully it's gone for good now.

PR #1590, opened 2021-08-03
1580 issue more domains

### LINKED ISSUE #1580: Allow variables to have more than 3 domains
**Summary**
 Currently variables can have a maximum of 3 domains, i.e. a `"primary"` domain and two auxiliary domains, `"secondary"` and `"tertiary"`. This could be extended to allow `"quaternary"` etc. domains (or changed to allow any number of domains).

Also, a `TertiaryBroadcast()` (or a generalization) could be defined, similar to `PrimaryBroadcast()` and `SecondaryBroadcast()`.

 **Motivation**
 Necessary in order to add particle-size distributions to the DFN framework, where some variables live on 4 different domains , e.g.
```python
 c_s_distribution = pybamm.Variable(
    "Negative particle concentration distribution",
    domain="negative particle",
    auxiliary_domains={
        "secondary": "negative particle size",
        "tertiary": "negative electrode",
        "quaternary": "current collector",
    },
    bounds=(0, 1),
)
```
(Even if some domains are 0D in practice, they can push others off the domain list before the domains are discretized.)

A `TertiaryBroadcast()` would then be used for e.g. broadcasting the "X-averaged" concentration distribution to the "electrode" domains (which would be added in the `"tertiary"` position).

**Additional context**

tobykirk:
# Description
Fixes #1580

Extended the domain list of symbols to allow a fourth ("quaternary") domain, and updating the spatial operators that add or remove domains (thus shifting their position up and down) to reflect this: 
 -  `PrimaryBroadcast` and `SecondaryBroadcast`
 -  `Integral`
 -  `BoundaryValue`

Also added a `TertiaryBroadcast` class, similar to `SecondaryBroadcast` but puts the broadcast domain in the `"tertiary"` rather than `"secondary"` position.

## Type of change

# Key checklist:

## Further checks:


pybamm/expression_tree/binary_operators.py
```diff
@@ -827,7 +827,7 @@ def simplified_addition(left, right):
             )
         ) and all(
             left.evaluates_on_edges(dim) == right.evaluates_on_edges(dim)
-            for dim in ["primary", "secondary", "tertiary"]
+            for dim in ["primary", "secondary", "tertiary", "quaternary"]
```
tinosulzer, line 830: We should make `evaluates_on_edges("all")` check all the dimensions to avoid this repetitive construction

tinosulzer (approved):
Thanks @tobykirk ! Just need to fix coverage. I think we could also reduce some code duplication in places with a few tricks ... but this is fine for now.

For example: we could have a list of domain orders (particle, size, electrodes, collector) and then automatically check/adjust domains when broadcasting based on that. Also, `domains` could be a specialized dict that understands that "primary"===1, "secondary"===2, etc and adjusts based on that

pybamm/expression_tree/binary_operators.py
tobykirk: This is checking that `evaluates_on_edges` match for left and right in each dimension (rather than all on edges or all on nodes), right? After playing around, keeping the domain loop outside of `evaluate_on_edges` seemed a lot cleaner, just replacing the list `["primary", "secondary", ...]` with `left.domains.keys()`. (Both `left` and `right` have the same domains at this point, I think? So can just use the domain list of one of them.)

tobykirk:
> Thanks @tobykirk ! Just need to fix coverage. I think we could also reduce some code duplication in places with a few tricks ... but this is fine for now.
> 
> For example: we could have a list of domain orders (particle, size, electrodes, collector) and then automatically check/adjust domains when broadcasting based on that. Also, `domains` could be a specialized dict that understands that "primary"===1, "secondary"===2, etc and adjusts based on that

Indeed, I like these ideas! I had a go at implementing the specialised dict for `domains`, subclassing `dict` and overwriting `__setitem__` and `__getitem__` to recognize 1 as `"primary"`, etc. Was causing issues with the domain properties/setters in `Symbol` though, and was getting myself confused. Probably needs a more careful implementation than I can do at the moment, so decided to leave it for now.

tinosulzer:
Alright, let's work on those other cases once this is merged

tinosulzer (approved):
All looks good now, ready to merge?

tobykirk:
> All looks good now, ready to merge?

Ah, still have to update Changelog. Will do that now.

Issue #1479, opened 2021-05-07
pybamm fails when running script a second time in spyder on windows


TomTranter:
**Describe the bug**
Running any script more than once seems to result in an import error because of the recent changes to parameters

  File "<frozen importlib._bootstrap>", line 908, in _find_spec

AttributeError: 'ImportDenier' object has no attribute 'find_spec'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):

  File "C:\Users\tom\code\PyBaMM\examples\scripts\drive_cycle.py", line 14, in <module>
    param = model.default_parameter_values

  File "c:\users\tom\code\pybamm\pybamm\models\full_battery_models\base_battery_model.py", line 346, in default_parameter_values
    return pybamm.ParameterValues(chemistry=pybamm.parameter_sets.Marquis2019)

  File "c:\users\tom\code\pybamm\pybamm\parameters\parameter_values.py", line 73, in __init__
    self.update_from_chemistry(chemistry)

  File "c:\users\tom\code\pybamm\pybamm\parameters\parameter_values.py", line 216, in update_from_chemistry
    self.update(

  File "c:\users\tom\code\pybamm\pybamm\parameters\parameter_values.py", line 301, in update
    loaded_value = pybamm.load_function(os.path.join(path, value[10:]))

  File "c:\users\tom\code\pybamm\pybamm\util.py", line 273, in load_function
    module_object = importlib.import_module(path)

  File "C:\Users\tom\anaconda3\lib\importlib\__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)

  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import

  File "<frozen importlib._bootstrap>", line 991, in _find_and_load

  File "<frozen importlib._bootstrap>", line 961, in _find_and_load_unlocked

  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed

  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import

  File "<frozen importlib._bootstrap>", line 991, in _find_and_load

  File "<frozen importlib._bootstrap>", line 971, in _find_and_load_unlocked

  File "<frozen importlib._bootstrap>", line 910, in _find_spec

  File "<frozen importlib._bootstrap>", line 884, in _find_spec_legacy

  File "C:\Users\tom\anaconda3\lib\site-packages\IPython\external\qt_loaders.py", line 48, in find_module
    if path:

[...]

Issue #1433, opened 2021-03-16
Composite anode of graphite and silicon


weilongai:
**Summary**
There is a trend of adding silicon to graphite to increase the energy density of batteries, so we are considering to develop a composite anode model. It is based on the fickan's diffusion equation for the two phases (silicon and graphite), but PyBaMM considers only one phase. What are the best ways to add this feature?

It seems not possible to reuse the particle submodel for lithium diffusion in both phases. Do we need to rewrite the particle submodel to make one particle_composite model?

I cannot explain the theoretical model too much, because it is under development.

tinosulzer:
One option is to add the particle_composite submodel, another option (more general) is to make the existing particle submodel more general and then call it twice, once for silicon and once for graphite. For example, see how the [base interface submodel](https://github.com/pybamm-team/PyBaMM/blob/develop/pybamm/models/submodels/interface/base_interface.py) takes a `reaction` argument.

We will also need to think about how to deal with the two interfacial current densities for silicon and graphite, but this shouldn't be too bad as it's similar to SEI.

Either way I recommend you start by editing the `BasicDFN` model where you can easily just add another `Variable` for silicon concentration and edit the current equations

weilongai:
Thanks for your suggestions. I think I will make a new `particle_composite` submodel, because reusing the existing particle submodel needs a lot of modifications and is not an easy task.

weilongai:
Is there any example on how to use the `BasicDFN` model? I tried to use it but got the error of `KeyError: "'Total current density' not found` using the below codes.

```python3
import pybamm
import os
import numpy as np
import matplotlib.pyplot as plt
import timeit
start = timeit.default_timer()
os.chdir(pybamm.__path__[0]+'/..')


model = pybamm.lithium_ion.BasicDFNComposite()
chemistry = pybamm.parameter_sets.Chen2020_composite
param = pybamm.ParameterValues(chemistry=chemistry)
param.update({"Upper voltage cut-off [V]": 4.21})
param.update({"Lower voltage cut-off [V]": 2.49})

exp = pybamm.Experiment(["Discharge at 1C until 2.5V"], use_simulation_setup_type='old')
sim1 = pybamm.Simulation(
    model, 
    experiment=exp,
    parameter_values=param,
    solver=pybamm.CasadiSolver(),
)
solution1 = sim1.solve()

stop = timeit.default_timer()
print('running time: ' + str(stop - start) +'s')
```

[...]

PR #1598, opened 2021-08-09
Issue 1594 add O'Regan 2021 parameters

### LINKED ISSUE #1594: Add LG M50 thermal parameter set from O'Regan 2021
**Summary**
Add the parameter set for the LG M50 (including temperature dependence) from the O'Regan 2021 paper (in progress).

brosaplanella:
# Description

Added LG M50 parameter set from O'Regan 2021 (submitted).

Fixes #1594

## Type of change

# Key checklist:

## Further checks:


pybamm/input/parameters/lithium_ion/cells/LGM50_ORegan2021/parameters.csv
```diff
@@ -0,0 +1,34 @@
+Name [units],Value,Reference,Notes
+# Empty rows and rows starting with ‘#’ will be ignored,,,
+,,,
+# Macroscale geometry,,,
+Negative current collector thickness [m],12E-6,Chen 2020,
+Negative electrode thickness [m],85.2E-6,Chen 2020,
+Separator thickness [m],12E-6,Chen 2020,
+Positive electrode thickness [m],75.6E-6,Chen 2020,
+Positive current collector thickness [m],16E-6,Chen 2020,
+Electrode height [m],6.5E-2,Chen 2020,Not needed for 1D
```
rtimms, line 10: I think we also have the comment "Not needed for 1D" in the Chen2020 set, but the height and width are used to calculate the electrode cross-sectional area, which is used in 1D. Maybe we should update the comments to say the width is the "unwound" length of the jelly roll?

pybamm/input/parameters/lithium_ion/negative_electrodes/graphite_ORegan2021/graphite_LGM50_electrolyte_exchange_current_density_ORegan2021.py
```diff
@@ -0,0 +1,45 @@
...
+        Temperature [K]
+
+    Returns
+    -------
+    :class:`pybamm.Symbol`
+        Exchange-current density [A.m-2]
+    """
+
+    i_ref = 2.668  # (A/m2)
+    alpha = 0.792
```
rtimms, line 32: Is this consistent with the fact we only have symmetric Butler-Volmer?

pybamm/parameters/parameter_sets.py
```diff
@@ -207,6 +207,17 @@
...
 
+ORegan2021 = {
+    "chemistry": "lithium_ion",
+    "cell": "LGM50_ORegan2021",
+    "negative electrode": "graphite_ORegan2021",
+    "separator": "separator_ORegan2021",
+    "positive electrode": "nmc_ORegan2021",
+    "electrolyte": "lipf6_EC_EMC_3_7_Landesfeind2019",
+    "experiment": "1C_discharge_from_full_ORegan2021",
+    "citation": "ORegan2021",
```
rtimms, line 218: Should you cite Chen2020 as well?

rtimms (approved):
great, thanks @brosaplanella !

TomTranter (changes requested):
As you need to change to quite a fine mesh for this set to give decent results it might be work printing a warning when using it or maybe include an example. Also I'm getting the kernel problem again where I needed to add empty inits to the parameter subfolders so I can run a script twice without restarting the kernel but I checked them all and can't see any missing. Very strange

pybamm/input/parameters/lithium_ion/negative_electrodes/graphite_ORegan2021/graphite_LGM50_electrolyte_exchange_current_density_ORegan2021.py
brosaplanella: Probably not, from the thermodynamical point of view, but given that this was the measured exchange current density I think it is better to leave it as it is.

brosaplanella:
> As you need to change to quite a fine mesh for this set to give decent results it might be work printing a warning when using it or maybe include an example. 

I am happy either way, but the example will take a few minutes to run so that will make GitHub actions even slower. Where would it be best to include the warning?

> Also I'm getting the kernel problem again where I needed to add empty inits to the parameter subfolders so I can run a script twice without restarting the kernel but I checked them all and can't see any missing. Very strange

Where do you get this? I mean, is it related to how the parameters are implemented or is it a separate issue?

TomTranter:
It's my guess that it relates to the parameters as this is what it was last time. #1479 I think a notebook example would be best - maybe just run it for a few hundred seconds

tinosulzer:
If the parameter set is graphite + SiOx don't you need #1433 ?

brosaplanella:
> If the parameter set is graphite + SiOx don't you need #1433 ?

No, because the parameterization is for a DFN model with one component, so the parameters capture the complexity of the two phases. When #1433 is merged, we will need separate parameter sets (OCVs, diffusivities, etc) for each one independently.

tinosulzer (commented):
Looks good but seems like this PR changes the test time from 1h to 2h, can we make it quicker? https://github.com/brosaplanella/PyBaMM/actions

brosaplanella:
> Looks good but seems like this PR changes the test time from 1h to 2h, can we make it quicker? https://github.com/brosaplanella/PyBaMM/actions

Is it from this PR? Because I have only added 2 tests (one is citation where the addition is negligible and the other runs in 6s in my laptop). For the notebook I can shorten the solving time (or even comment it out) but that was 20s in my laptop. Also, checking the Actions in this repo I have noticed that other tests took nearly 2h too (e.g. [this one](https://github.com/pybamm-team/PyBaMM/actions/runs/1143550953)). Is there a way of knowing where the bottleneck is?

tinosulzer (approved):

tinosulzer:
@TomTranter needs to approve before we can merge

PR #1617, opened 2021-08-23
Issue 1505 porosity

### LINKED ISSUE #1505: Reformat porosity model for SEI/plating
Should not be its own ODE but instead proportional to SEI thickness and plating thickness

brosaplanella:
# Description

Write reaction driven porosity as a function of the plated & SEI thicknesses for lithium-ion models (lead-acid stays the same).

Fixes #1505

## Type of change

# Key checklist:

## Further checks:


examples/scripts/cycling_ageing.py
```diff
@@ -30,24 +30,28 @@
...
             "Discharge at 2 C until 2.8 V",
+            "Rest for 30 minutes",
         ),
         (
             "Charge at 1 C until 4.2 V",
             "Hold at 4.2 V until C/20",
             "Rest for 30 minutes",
-            "Discharge at 3 C until 2.8 V",
+            "Discharge at 3 C until 2.8 V (10 second period)",
+            "Rest for 30 minutes",
```
brosaplanella, line 54: Had to add the rest periods and reduce the sampling period at discharge, otherwise simulations didn't converge.

pybamm/models/submodels/porosity/reaction_driven_porosity_ode.py
```diff
@@ -0,0 +1,111 @@
...
+            )
+
+        deps_n_dt = -self.param.beta_surf_n * j_n
+        # Reaction driven porosity ODE for lithium-ion is not supported at the moment
+        # if self.options["SEI porosity change"] == "true":
+        #     beta_sei_n = self.param.beta_sei_n
+        #     deps_n_dt += beta_sei_n * j_sei_n
+        # if self.options["lithium plating porosity change"] == "true":
+        #     beta_plating = self.param.beta_plating
+        #     deps_n_dt += beta_plating * j_plating
```
brosaplanella, line 75: I commented it out as it is not used. Should I delete it? Because it is checking the options rather than the submodels at the moment there is no way we can enter these `if` statements.

pybamm/models/full_battery_models/lead_acid/higher_order.py
```diff
@@ -198,7 +198,7 @@ def set_full_porosity_submodel(self):
         Update porosity submodel, now that we have the spatially heterogeneous
         interfacial current densities
         """
-        self.submodels["full porosity"] = pybamm.porosity.ReactionDriven(
+        self.submodels["full porosity"] = pybamm.porosity.ReactionDrivenODE(
```
brosaplanella, line 201: At the moment this escapes coverage (and I think it does in the main branch too) because all the other classes overwrite it. What should I do?

tinosulzer: don't worry about it, it's a bug but an extremely low priority one (this is just to get the right porosity distribution in lead-acid models)

pybamm/models/submodels/porosity/reaction_driven_porosity_ode.py
tinosulzer: Yes can be removed

tinosulzer (approved):
Thanks @brosaplanella ! Just remove the commented out lines then we can merge

PR #1624, opened 2021-08-26
Remove Matplotlib style

### LINKED ISSUE #1618: Importing pybamm affects style of Matplotlib figures
**Describe the bug**
I prefer to create my own Matplotlib figures without using the `sim.plot()` function in PyBaMM. Unfortunately, importing `pybamm` will affect the style of the Matplotlib figures even if I don't use PyBaMM to create the plots.  

**To Reproduce**
In this example, `pybamm` is imported. Notice the small font size used for axis labels and the red color applied to the line plot.

```python
import pybamm

_, ax = plt.subplots()
ax.plot([1, 3, 5, 3, 8.4])
ax.set_xlabel('X axis')
ax.set_ylabel('Y axis')

plt.show()
```

fig-pybamm

In this example, `pybamm` is not imported. Notice the default (larger) font size for axis labels and the blue color for the line plot.

```python

_, ax = plt.subplots()
ax.plot([1, 3, 5, 3, 8.4])
ax.set_xlabel('X axis')
ax.set_ylabel('Y axis')

plt.show()
```

fig-no-pybamm

**Expected behaviour**
I expect the default figure style to be used when creating plots with Matplotlib.

wigging:
# Description

Remove the imposed Matplotlib style as discussed in the issue mentioned below. Also adjust the title `fontsize` when using `sim.plot()` for better readability.

Fixes #1618 

## Type of change

# Key checklist:

## Further checks:


tinosulzer:
I'm a bit reluctant to merge this just before a release, want to give it a bit of time to check the plots look ok. So will do the release first then merge

tinosulzer (approved):

PR #1644, opened 2021-09-07
back to the macos fix

tinosulzer:
trying to fix the macos failing tests


tinosulzer:
https://github.com/pybamm-team/PyBaMM/runs/3531296211?check_suite_focus=true#step:5:568

PR #1649, opened 2021-09-09
Build and publish python sdist on PyPI

priyanshuone6:
# Description

Add a workflow to build and publish PyBaMM's python source distribution on PyPI for Windows (.zip) and Unix (.tar.gz)

### DIFF
--- a/.github/workflows/build_sdist.yml
+++ b/.github/workflows/build_sdist.yml
@@ -1,8 +1,6 @@
 name: "Build and publish Python Source Distribution"
 
 on:
-  push:
-    branches: main
   workflow_dispatch:
     inputs:
       target:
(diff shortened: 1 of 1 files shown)


priyanshuone6:
Not sure on how to test this

tinosulzer:
What's the difference between this and the existing workflow to publish releases?

priyanshuone6:
A source distribution is basically an archive containing the source code (both Python and C++). In contrast a wheel is what we call a built distribution: it contains the "installed" version of a package - particularly the compiled extension modules.
I talked with @tlestang and he suggested creating a build_sdist.yml that builds the tar.gz and publishes it to PyPI. The tar.gz file published to PyPI will then be used to automatically bump the version on conda-forge/pybamm-feedstock

tinosulzer:
Ok, I don't know how to test this either, I guess @tlestang should review it

tlestang:
That all looks very good. @priyanshuone6 to test this, you can trigger this action from your own fork of PyBaMM. The `workflow_dispatch` at the top of the workflow file entry makes it possible to manually trigger the workflow. When doing so, make sure to choose testpypi to upload the sdist to the test pypi server. you'll need to create an account on test.pypi.org first, and save the token on your fork's secret list.

priyanshuone6:
@tinosulzer @tlestang 
I think it's ready to merge now. You can check the test workflow run [here](https://github.com/priyanshuone6/PyBaMM/runs/3564070355?check_suite_focus=true)

PR #1658, opened 2021-09-10
Fix workflow_dispatch in build_sdist

priyanshuone6:
# Description

Fix workflow_dispatch and the `if` condition due to which few steps were getting skipped [here](https://github.com/pybamm-team/PyBaMM/runs/3569181081?check_suite_focus=true)

### DIFF
--- a/.github/workflows/build_sdist.yml
+++ b/.github/workflows/build_sdist.yml
@@ -4,10 +4,6 @@ on:
   push:
     branches: main
   workflow_dispatch:
-    inputs:
-      target:
-        description: 'Deployment target. Can be "pypi" or "testpypi"'
-        default: 'pypi'
 
 jobs:
   build_sdist:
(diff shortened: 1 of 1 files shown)


tinosulzer:
So does this automatically send to both pypi and testpypi now?

priyanshuone6:
Yes

tlestang:
@priyanshuone6 I don't think we want to get rid of this. We still need a way of manually setting the deployment target to test.pypi.org for testing purposes. Not sure I remember correctly but I think the default target pypi isn't set properly as the default when this workflow is ran automatically after a merge event.

priyanshuone6:
Also, the workflow_dispatch is not showing up in the actions tab, which might be because the workflow is not there in the default develop branch. Closing this and will make a new PR to merge main to develop
image

PR #1666, opened 2021-09-13
Add multi-line if condition in workflow

### LINKED ISSUE #1663: "Publish to PyPI" job skips actual deployment step
- PyBaMM version - 21.08
- Python version - na

On both workflows [Build and publish Python wheels](https://github.com/pybamm-team/PyBaMM/blob/develop/.github/workflows/build_wheels.yml) and [Build and publish python sdist](https://github.com/pybamm-team/PyBaMM/blob/develop/.github/workflows/build_sdist.yml), the job that should upload the wheels (or sdist) to PyPI actually doesn't do anything when the workflow is triggered by a merge event on `main`. Both steps "Publish wheels/sdist on PyPI" or "Publish wheels/sdist on TestPyPI" are skipped. 

**Expected behavior**: Only step "Publish wheelsdist to TestPyPI" is skipped and wheels/sdist are uploaded to PyPI.

## Details

When the workflow is triggered from a merge event, it seems like both the conditionals
```
if: github.event.inputs.target == 'pypi'
```
or 
```
if: github.event.inputs.target == 'testpypi'
```
evaluate to false. Probably because `github.event.inputs.target` isn't defined?

This workflow works as expecred when manually triggered (using workflow dispatch). In this case if: `github.event.inputs.target == 'pypi'` is properly set to `pypi` or `testpypi` according to whichever is selected when triggering the workflow. I initially thought that defining `pypi` as the `default` entry would be enough for deploying to PyPI when triggering the workflow from a merge event.

Being able to choose between TestPyPI and PyPI when triggering the workflow manually is very useful for testing. Unless there is a better of doing something similar, I'd like to keep this in place. The question is then how not to skip the PyPI step when running after a merge event? Can `github.event.inputs.target` be defined in this case? Can we maybe check it's undefined?

Useful links:

- https://docs.github.com/en/actions/reference/events-that-trigger-workflows
- https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context

**To Reproduce**

Push or merge something to `main` (preferentially on your own fork of PyBaMM). 
Go to the "Actions" tab on GitHub and observe how PyBaMM is built but actual deploy step is skipped.

priyanshuone6:
# Description

Add multi line if condition to trigger the workflow from a merge event in the main branch

Fixes #1663 

See test workflow run [here](https://github.com/priyanshuone6/PyBaMM/runs/3587866795?check_suite_focus=true)
(It skipped the pypi event and only ran the testpypi event from a merge because I added the multi-line condition to testpypi)

### DIFF
--- a/.github/workflows/build_sdist.yml
+++ b/.github/workflows/build_sdist.yml
@@ -53,7 +53,9 @@ jobs:
           path: sdist
 
       - name: Publish sdist on PyPI
-        if: github.event.inputs.target == 'pypi'
+        if: |
+          github.event.inputs.target == 'pypi' ||
+          (github.event_name == 'push' && github.ref == 'refs/heads/main')
         uses: pypa/gh-action-pypi-publish@master
         with:
           user: __token__
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -81,7 +81,9 @@ jobs:
           name: wheels
 
       - name: Publish wheels on PyPI
-        if: github.event.inputs.target == 'pypi'
+        if: |
+          github.event.inputs.target == 'pypi' ||
+          (github.event_name == 'push' && github.ref == 'refs/heads/main')
         uses: pypa/gh-action-pypi-publish@master
         with:
           user: __token__


tlestang:
Looking good!

PR #1670, opened 2021-09-14
Remove .zip sdist

priyanshuone6:
# Description

Remove .zip from sdist as only one sdist may be uploaded per release.
https://github.com/pybamm-team/PyBaMM/runs/3596099765?check_suite_focus=true


brosaplanella (approved):
Looks good, thanks Priyanshu!

PR #1675, opened 2021-09-15
Issue 1673 import parameters chemistry error

### LINKED ISSUE #1673: Import parameters from chemistry fails when importing from local parameter set
- PyBaMM version - `$ python -m pip show pybamm`: 21.8
- Python version - `$ python --version`: 3.7.4

**Describe the bug**
When there is a local folder `lithium_ion` with the parameter set, importing parameters from chemistry fails.

**To Reproduce**
1. Run on bash `pybamm_edit_parameter lithium_ion`
2. Then, in the same folder run the following pybamm command: `pybamm.ParameterValues(chemistry=pybamm.parameter_sets.Chen2020`

**Expected behaviour**
It should import the parameters as usual.

brosaplanella:
# Description

This allows importing parameters via chemistry that is in the local directory. I don't think this fixes importing parameters from other folders via chemistry. This is a much more complicated issue (I believe), but this fix should be enough for the workshop.

Fixes #1673 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,7 @@ example notebook ([#1602](https://github.com/pybamm-team/PyBaMM/pull/1602))
 
 ## Bug fixes
 
+-   Fixed bug with `load_function` ([#1675](https://github.com/pybamm-team/PyBaMM/pull/1675))
 -   Updated documentation to include some previously missing functions, such as `erf` and `tanh` ([#1628](https://github.com/pybamm-team/PyBaMM/pull/1628))
 -   Fixed reading citation file without closing ([#1620](https://github.com/pybamm-team/PyBaMM/pull/1620))
 -   Porosity variation for SEI and plating models is calculated from the film thickness rather than from a separate ODE ([#1617](https://github.com/pybamm-team/PyBaMM/pull/1617))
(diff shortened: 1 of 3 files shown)


brosaplanella:
PS: we need to include this fix in the released version for the workshop, so not sure what is the right pipeline for this.

tinosulzer:
Let's make a patch release for this. Can you make a PR to main with version 21.8.1?

tinosulzer (approved):

brosaplanella:
So, do I merge this to main first with the new version number and then we merge main to develop?

tinosulzer:
Yep, but don't merge this directly to main since it will also merge the other recent changes to develop. You need to make a new branch from main with just these changes. Might be possible with git cherry pick or just do manually

brosaplanella:
When we switched to calendar versioning we removed the patch number (it is just 21.08 now). Should I bring it in? Technically it is not a new version but just a patch to fix a bug mostly because of the workshop (otherwise we could wait until the next release).

tinosulzer:
Hmm ok yeah change it to 21.08.post2 then (in setup.py), and just add the line to changelog but don't make it a new section

PR #1681, opened 2021-09-17
Merge main to develop

priyanshuone6:
# Description
Merge main branch to develop branch


priyanshuone6:
I am unable to resolve the conflicting files, someone with the proper access has to do it

brosaplanella:
The conflict is on the baseline for the configuration. Should we keep the one in main or the one in develop?

brosaplanella (approved):
Looks good, thanks Priyanshu!

priyanshuone6:
@brosaplanella can you share the commit id for both the branches

priyanshuone6:
It should look like this
```
{
  "default-registry": {
    "kind": "builtin",
    "baseline": "e201002b4f4827d7b7b672f0c5c672a77fc3b77d"
  },
  "registries": [
    {
      "kind": "git",
      "repository": "https://github.com/pybamm-team/sundials-vcpkg-registry.git",
      "baseline": "3cdbafc061c4cdda2377c77c86aeb72cfcfbada6",
      "packages": [ "sundials" ]
    }
  ]
}
```

brosaplanella:
```
<<<<<<< main
    "baseline": "e201002b4f4827d7b7b672f0c5c672a77fc3b77d"
=======
    "baseline": "642d1ccb276ce13d8a8521b1ef2856f5f7bc9cdd"
>>>>>>> develop
```
So from what you poster I imagine we want to keep the one from main.

priyanshuone6:
Yup

brosaplanella:
I think I resolved it... (also, I killed the build weels workflow)

priyanshuone6:
I think you merged all the unreleased changes from develop to main (https://github.com/pybamm-team/PyBaMM/commit/830093476ce8e1a213cf24d628c516dbb69bcd23), but that shouldn't be a problem as in few days we are going to release a new version anyway

brosaplanella:
Let me revert it then...

tinosulzer:
Yes please. But interestingly the release was not triggered on pypi, since the version number did not change. So maybe we don't actually need a separate main and develop, we just update the version number when we want to make a new release

brosaplanella:
Ok, I tried to revert the commit but I might have chosen the wrong branch as no changes appear. Can someone point me how to undo it? I don't want to mess it up more...

priyanshuone6:
Yes, and if anything goes wrong we can add new commits to the existing Git tag
https://stackoverflow.com/questions/21459540/add-new-commit-to-the-existing-git-tag/21466838

priyanshuone6:
@brosaplanella instead of reverting could you try dropping both the commits and then force push the branch

brosaplanella:
The problem is that because it is a merge there are many commits to remove and not sure the best way of doing so without messing it up even more.

priyanshuone6:
Drop all the commits till you see `Fix error` (https://github.com/pybamm-team/PyBaMM/commit/17b40edb91bd4b4c3af8192b9b090cb567681623)

priyanshuone6:
git rebase --interactive --autosquash HEAD~2

brosaplanella:
Then I get a text file to choose the commits. Should I leave it as it is? It seems to pick all the commits we don't want...

priyanshuone6:
If you are using vscode, do install git lens, it would make the process much easier and instead of opening text file, it will open an interactive shell

priyanshuone6:
`git rebase --interactive --autosquash HEAD~2`
and then drop all the 87 commits

brosaplanella:
Ok. I think now I reverted to the previous state, but the problem is that merging will delete all the progress done in `develop`.

priyanshuone6:
Yep, that is why we need to completely drop the commits from the PR

PR #1706, opened 2021-10-01
release

tinosulzer:


priyanshuone6:
Does pypi automatically remove 0 from the version as v21.08 is v21.8 in pypi

tinosulzer:
I guess so

priyanshuone6:
If that's the case then we should match the name of this release with pypi

pybamm/version
```diff
@@ -1 +1 @@
-21, 8
+21, 9
```
priyanshuone6, line 1: Or maybe this should be 21, 09

brosaplanella: I think it is read as an `int` and the formatted when it is printed. About PyPi, I think I saw some warning in the actions (probably the one that builds the wheels) that said that PyPi would ignore the 0, so we might as well remove it...

brosaplanella (approved):

priyanshuone6:
@tinosulzer You reverted `pybamm/version` back to 21, 8 in the latest commit

PR #1721, opened 2021-10-06
Merge main into develop

tinosulzer:
Merge main into develop, hopefully will avoid future conflicts like we had with the last release


brosaplanella (approved):
We will still have to change the baseline to the latest commit. Might be worth trying to automate it.

brosaplanella:
@priyanshuone6 if the only changes from merging `main` into `develop` are in `vcpkg-configuration.json`, do we still have to merge it?

PR #1735, opened 2021-10-10
Power mode bug

valentinsulzer:
# Description

Fix a bug in `is_matrix_one` that was causing an incorrect result when using power mode (and probably also voltage mode).

## Type of change

# Key checklist:

## Further checks:


tinosulzer:
Could probably do with adding more tests to catch this kind of bug earlier

PR #1743, opened 2021-10-11
Exit sdist workflow if wheels fail

priyanshuone6:
Exit sdist workflow if wheels fail


tinosulzer:
What is it doing currently if the build fails?

tinosulzer:
Ok makes sense, thanks

PR #1767, opened 2021-10-29
Make jax optional

priyanshuone6:
# Description

Make jax and jaxlib optional

Fixes #1701 
Fixes #1775 

## Type of change

Please add a line in the relevant section of [CHANGELOG.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CHANGELOG.md) to document the change (include PR #) - note reverse order of PR #s. If necessary, also add to the list of breaking changes.

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (back-end change that speeds up the code)
- [ ] Bug fix (non-breaking change which fixes an issue)

# Key checklist:

- [ ] No style issues: `$ flake8`
- [ ] All tests pass: `$ python run-tests.py --unit`
- [ ] The documentation builds: `$ cd docs` and then `$ make clean; make html`

You can run all three at once, using `$ python run-tests.py --quick`.

## Further checks:

- [ ] Code is commented, particularly in hard-to-understand areas
- [ ] Tests added that prove fix is effective or that feature works

PR #1770, opened 2021-10-31
Changes release 21.10

brosaplanella:
# Description

Updated the version number for release 21.10. I guess we are keeping `jax` pinned to whatever it is now, as in the next release we will make it optional (#1767)?


tinosulzer (approved):
Thanks! Let's leave jax as it is for now

PR #1771, opened 2021-10-31
Release v21.10

tinosulzer:


brosaplanella:
@priyanshuone6 can you remind us which commit ID we need to use in `vcpgk-configuration.json`?

brosaplanella (approved):

priyanshuone6:
`699c8779f1b0feb4ed3564716d1ed31f69663ea6` should work

CITATION.cff
```diff
@@ -24,6 +24,6 @@ keywords:
   - "expression tree"
   - "python"
   - "symbolic differentiation"
-version: 21.9
+version: 21.10
```
Saransh-cpp, line 27:
```suggestion
version: "21.10"
```

`YAML` truncates the trailing 0s (which would also be the case if we add python `3.10`).
The citation right now in GitHub's citations tab comes out to be ("`Version 21.1`") -
```
Sulzer, V., Marquis, S. G., Timms, R., Robinson, M., & Chapman, S. J. (2021). Python Battery Mathematical Modelling (PyBaMM) (Version 21.1) [Computer software]. https://doi.org/10.5334/jors.309
```

Saransh-cpp: Just realized this is develop going into main, should a make a PR to develop for this?

priyanshuone6: Committing here should be fine as this PR is made from the develop branch so changes will be committed to develop first before appearing here

Issue #1075, opened 2020-06-24
Increase default r_pts


rtimms:
I think we should increase the default r_pts from 10. E.g.
```python3
import pybamm
import numpy as np

model = pybamm.lithium_ion.DFN()

chemistry = pybamm.parameter_sets.Ecker2015
parameter_values = pybamm.ParameterValues(chemistry=chemistry)

var = pybamm.standard_spatial_vars
npts = [10, 20, 25, 30, 40, 50]

solver = pybamm.CasadiSolver(mode="fast")
t_eval = np.linspace(0, 3800, 200)

sols = []
for N in npts:
    var_pts = {var.x_n: 20, var.x_s: 20, var.x_p: 20, var.r_n: N, var.r_p: N}
    sim = pybamm.Simulation(
        model, solver=solver, parameter_values=parameter_values, var_pts=var_pts
    )
    sol = sim.solve(t_eval=t_eval)
    print("N={}: {} s".format(N, sol.solve_time))
    sols.append(sol)

pybamm.dynamic_plot(sols, ["Terminal voltage [V]"], time_unit="seconds", labels=npts)
```

gives 

N=10: 0.15861007800413063 s
N=20: 0.235911022995424 s
N=25: 0.29073380999761866 s
N=30: 0.31032642299396684 s
N=40: 0.4951283990012598 s
N=50: 0.5220356229983736 s

![Figure_2](https://user-images.githubusercontent.com/43040151/85523692-37762900-b5ff-11ea-9139-933e013cc036.png)

My vote would be somewhere between 20 and 30

tinosulzer:
Wow, that is bad. Definitely at least 30 required.

PR #1782, opened 2021-11-03
Spherical finite volume

tinosulzer:
# Description

Fixes a bug in `finite_volumes.py` for spherical diffusion.
`(r_edge[i+1]**3 - r_edge[i]**3)/3` is now used instead of `r_node**2`. This is what it should be according to Finite Volume theory e.g. https://journals.aps.org/prapplied/abstract/10.1103/PhysRevApplied.6.064010

Screen Shot 2021-11-03 at 3 48 05 PM

Previous convergence: 

Figure_2

New convergence:

Figure_1

Code: 
```python3
import pybamm
import numpy as np

chemistry = pybamm.parameter_sets.Ecker2015

npts = [10, 20, 25, 30, 40, 50]

t_eval = np.linspace(0, 3800, 200)

sols = []
for N in npts:
    var_pts = {var.x_n: 20, var.x_s: 20, var.x_p: 20, var.r_n: N, var.r_p: N}
    sim = pybamm.Simulation(
        model,
        solver=solver,
        var_pts=var_pts,
        # C_rate=5,
    )
    sol = sim.solve(t_eval=t_eval)
    print("N={}: {} s".format(N, sol.solve_time))
    sols.append(sol)

pybamm.dynamic_plot(sols, ["Terminal voltage [V]"], time_unit="seconds", labels=npts)
```

## Type of change

# Key checklist:

## Further checks:


tinosulzer:
Since the convergence is better, we could reduce the default number of r points to 20 (10 still gives inaccurate results at higher C-rates). See #1075

rtimms (approved):
Thanks, @tinosulzer , nice catch! This has been lurking for a while...

Yes, let's reduce the default points, and maybe we can also check to see if we can now use fewer points in [this notebook](https://github.com/pybamm-team/PyBaMM/blob/develop/examples/notebooks/models/simulating-ORegan-2021-parameter-set.ipynb)? I'm guessing we might be able to since the method is correct now.

brosaplanella (approved):
Looks good, thanks Tino!

tinosulzer:
How did you come up with 150 points for the ORegan parameter set
before:
Figure_2

now:
Figure_1
?

brosaplanella:
There were some issues with the convergence with the low diffusion areas, but maybe with the final parameter set is not needed any more. If it converges, happy to reduce the number of points 😄

PR #1784, opened 2021-11-04
Remove empty method

wigging:
The `clear_casadi_attributes()` method in the `Solution` class doesn't do anything. Apparently the method is supposed to clear Casadi objects for pickling but this is not necessary since the `Solution` object can already be pickled. So I removed the `clear_casadi_attributes()` method and references to the method.


tinosulzer:
Test errors are clearly unrelated so I think it's fine to merge this

PR #1790, opened 2021-11-09
#1489 add surface area to volume ratio factor

### LINKED ISSUE #1489: Multiply by surface area to volume ratio
In all the degradation models that depend on a current (d.../dt = j), the RHS should be multiplied by surface area to volume ratio (a) to account for possible changes in active material

tinosulzer:
# Description

Add surface area to volume ratio factor

Fixes #1489 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/examples/scripts/compare_lithium_ion.py
+++ b/examples/scripts/compare_lithium_ion.py
@@ -7,10 +7,10 @@ pybamm.set_logging_level("INFO")
 
 # load models
 models = [
-    pybamm.lithium_ion.SPM(),
-    pybamm.lithium_ion.SPMe(),
+    # pybamm.lithium_ion.SPM(),
+    # pybamm.lithium_ion.SPMe(),
     pybamm.lithium_ion.DFN(),
-    pybamm.lithium_ion.NewmanTobias(),
+    # pybamm.lithium_ion.NewmanTobias(),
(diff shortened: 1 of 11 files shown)


examples/scripts/compare_lithium_ion.py
```diff
@@ -7,10 +7,10 @@
 
 # load models
 models = [
-    pybamm.lithium_ion.SPM(),
-    pybamm.lithium_ion.SPMe(),
+    # pybamm.lithium_ion.SPM(),
+    # pybamm.lithium_ion.SPMe(),
     pybamm.lithium_ion.DFN(),
-    pybamm.lithium_ion.NewmanTobias(),
+    # pybamm.lithium_ion.NewmanTobias(),
```
brosaplanella, line 13: I think we should restore this script to its previous version.

brosaplanella (approved):
Looks good, just a question: this is a dimensionless surface area to volume so it is `1` when there is no degradation, right?

examples/scripts/compare_lithium_ion.py
tinosulzer: oops yes

tinosulzer:
Yes it's dimensionless

PR #1803, opened 2021-11-13
Added a condition for `jaxlib` imports

### LINKED ISSUE #1801: Jaxlib not supported on windows so can we make the imports optional
- PyBaMM version - develop

Saransh-cpp:
# Description

If `jax` is already installed in an environment where the user updates `pybamm` to develop, importing `pybamm` gives the following error (as `jaxlib` is not available on windows but `jax` is) -
```
raise ModuleNotFoundError(
ModuleNotFoundError: jax requires jaxlib to be installed. See https://github.com/google/jax#installation for installation instructions.
```

Fixes #1801 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/util.py
+++ b/pybamm/util.py
@@ -343,8 +343,10 @@ def get_parameters_filepath(path):
 
 
 def have_jax():
-    """Check if jax is installed"""
-    return importlib.util.find_spec("jax") is not None
+    """Check if jax and jaxlib are installed"""
+    return (importlib.util.find_spec("jax") is not None) and (
+        importlib.util.find_spec("jaxlib") is not None
+    )
 
 
 def install_jax():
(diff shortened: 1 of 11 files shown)


pybamm/util.py
```diff
@@ -343,8 +343,10 @@ def get_parameters_filepath(path):
 
 
 def have_jax():
-    """Check if jax is installed"""
-    return importlib.util.find_spec("jax") is not None
+    """Check if jax and jaxlib are installed"""
+    return (importlib.util.find_spec("jax") is not None) and (
+        importlib.util.find_spec("jaxlib") is not None
```
Saransh-cpp, line 348: Another approach would be to create a new `have_jaxlib` function to carry out this condition but I am not sure if that would be better

priyanshuone6: Creating `have_jaxlib` would complicate things unnecessarily

Saransh-cpp: Yes

priyanshuone6 (approved):

tinosulzer (approved):
Thanks!

Issue #1787, opened 2021-11-04
How to add a custom parameter function that requires variables to be evaluated


ehtec:
I have written some smooth code that can train a model using machine learning methods (for example random forest regression) from supplied measured parameter values from a CSV file and save it to a PKL file. The code works with any number of independent variables, as long as there are enough measurements. I want to add this code to PyBaMM so that parameter value functions can easily be loaded from the PKL file. Here is the class taking care of that:

```python
class PybammStandardParameterClass:

    def __init__(self, pkl_filename):

        with open(pkl_filename, 'rb') as modelfile:
            self.model = pickle.load(modelfile)

    def pybamm_standard_parameter_function(self, params):

        print("params: {0}".format(params))

        params = np.array([[params]])

        res = float(self.model.predict(params))

        return res
```

I can call the function successfully using

```python
my_pspc = PybammStandardParameterClass(PATH_TO_PKL)

my_pspc.pybamm_standard_parameter_function(0.5)
```

for example.

If I try now to use this parameter, for example like this:

```python
pv = pb.ParameterValues({'Negative electrode volume change': my_pspc.pybamm_standard_parameter_function})
```

I get an error when running the simualtion:

```
TypeError: float() argument must be a string or a number, not 'Division'
```

When I look at the output from my debug message, I see this:

```
params: integral dr ['negative particle'](Negative particle concentration) / integral dr ['negative particle'](broadcast(1.0))
```

It looks like Pybamm is supplying my function with some abstract expression tree values instead of numerical values, and the model from the PKL file does not know how to use these abstract expressions, of course.

I played around with the `pb.Function` class, trying to fix this. I also had a look at this file, which has a similar functionality

https://pybamm.readthedocs.io/en/latest/_modules/pybamm/expression_tree/interpolant.html#Interpolant

But I could not figure out a working solution.

How do I need to handle the input parameter inside my function to be able to evaluate it?

tinosulzer:
pybamm has its own symbolic representation so your model would need to be able to be expressed in terms of these functions. This is necessary for conversion to casadi and jacobian calculations. Can you give an example of what `self.model` looks like?

ehtec:
```python
import numpy as np
from sklearn.ensemble import RandomForestRegressor
import pickle
import os


def train_model(csvpath, pkl_filename, choice_number=None):
    input_data = np.genfromtxt(csvpath, delimiter=',')

    if bool(choice_number):
[...]

PR #1805, opened 2021-11-15
Add support for 2D parameter data

ehtec:
This pull request is created to fix Issue https://github.com/pybamm-team/PyBaMM/issues/1787


tinosulzer:
Can you fix conflicts before we review? Also, I know this is personal preference, but I am not a fan of leaving a blank line between every line (less code visible on screen)

ehtec:
Please have a look again now.

ehtec:
The static code analysis warning is because it does not like the pickle model... it is good to use in this case though. People just should be aware that when using other people's pickle files, they may contain malicious content... and the same goes with any github code made by other people.

tinosulzer (changes requested):
These changes introduce too much complexity (will require maintenance) for what they really do. We should have a very specific csv format in which 2D data can be accepted, which can then be read by the `ParameterValues` class. If you want to host a notebook on your github on how to generate this csv file from a random forest that's fine and we can advertise it, I just don't think this belongs in the main repo

ehtec:
> These changes introduce too much complexity (will require maintenance) for what they really do. We should have a very specific csv format in which 2D data can be accepted, which can then be read by the `ParameterValues` class. If you want to host a notebook on your github on how to generate this csv file from a random forest that's fine and we can advertise it, I just don't think this belongs in the main repo

To be honest, I'd rather leave it as is... let me explain the reasons. This code is very dynamic and adaptable right now. You can supply any model through the PKL file, as long as it has a predict function. And the JSON file stores only the parameters for hte linspaces (start, end, number of values to use). The code in `pkl_input.py` and `csv_model_generator.py` is not bound to a specific dimension. It will correctly handle 1D, 2D, 3D data and data of any dimension. This is only limited inside the interpolator (changes would need to be made to `interpolant.py` and `convert_to_casadi.py`. To my knowledge, casadi and scipy don't have 3D+ interpolants yet).

And as you said yourself before, CSV is simply not made for this kind of data, it is made for 1D data. We would have a very large number of columns.

Furthermore, nobody can really use this feature when it is only accepting data in a perfect grid. It is practically not possible to produce this data at a reasonable accuracy (I work for AVL List GmbH, and even we don't produce this data, so I came up with the Random Forest Regressor idea). Even if you could, it would not be optimal at all. You always have less interesting, constant regions and also more interesting areas in every measuring range. As an experimental scientist, you will for sure want to focus on the latter and make more measurements where necessary, instead of preserving the same measurement density over the whole range. It is much easier to produce data of the format (independent variable 1, independent variable 2, dependent variable) like `csv_model_trainer.py` is taking as input.

So if I replace the PKL input with a CSV input, I'd basically add an unnecessary step, adding complexity.

rtimms:
I agree, the code in pybamm should just allow for 2D interpolants on a structured grid, and we should allow the user to generate uniform data from their own non-uniform data in whichever way they like. We could add a really simple example notebook with a small data set showing how to do this, and you could host an example using your random forest on your own github and we can link to it. I think the random forest is too specific to be in the core code.

rtimms:
Also note you don't necessaily _need_ to create a csv. You could just have some code that takes you nonuniform data and turns it into arrays on a uniform grid, then create the `Interpolant` directly rather than using the `[data]` tag. See "Loading in current data" [here](https://github.com/pybamm-team/PyBaMM/blob/develop/examples/notebooks/parameterization/change-input-current.ipynb).

ehtec:
> Also note you don't necessaily _need_ to create a csv. You could just have some code that takes you nonuniform data and turns it into arrays on a uniform grid, then create the `Interpolant` directly rather than using the `[data]` tag. See "Loading in current data" [here](https://github.com/pybamm-team/PyBaMM/blob/develop/examples/notebooks/parameterization/change-input-current.ipynb).

It is still possible to do that. Just call `pybamm.Interpolant([x1, x2], y)` where `x1` and `x2` are the linspaces and `y` is the 2D output grid, the same way the casadi interpolant takes it.

ehtec:
> I agree, the code in pybamm should just allow for 2D interpolants on a structured grid, and we should allow the user to generate uniform data from their own non-uniform data in whichever way they like. We could add a really simple example notebook with a small data set showing how to do this, and you could host an example using your random forest on your own github and we can link to it. I think the random forest is too specific to be in the core code.

Of course I can only speak for our use case, but including this will make it a lot easier to use for scientists here at AVL List GmbH. It is a lot simpler to just write `[ml data]<filename>` instead of creating the `Interpolant` over and over again and manually reading the PKL files if you have multiple 2D parameters.

And the code does not add any dependencies (pickle is in the standard lib), `pkl_input` is very short in general, and the code that reads the `[ml data]` tag is bound to an if statement, so even if something was broken here, you would not notice unless you actively use `[ml data]`.

I can move the `csv_model_trainer.py` to my own repo instead of your examples, sure, I don't care about this one. But I'd rather prefer to leave the `pkl_input.py` in there, if possible, as it does not interfere with anything and will make life easier for us or anyone measuring 2D data.

rtimms:
Sure, I just meant if you were to make an example you don't need to create a 2D csv, just create the data and pass it.

> It is a lot simpler to just write [ml data]<filename> instead of creating the Interpolant over and over again and manually reading the PKL files if you have multiple 2D parameters.

Can't this be automated though, just as a local function? Just pass the parameter name and the filename and then it automatically creates the `Interpolant` and updates the `ParameterValues`?

Thanks for taking this on by the way! I think it looks good, but is somewhat separate from the core code.

ehtec:
> Sure, I just meant if you were to make an example you don't need to create a 2D csv, just create the data and pass it.
> 
> > It is a lot simpler to just write [ml data] instead of creating the Interpolant over and over again and manually reading the PKL files if you have multiple 2D parameters.
> 
> Can't this be automated though, just as a local function? Just pass the parameter name and the filename and then it automatically creates the `Interpolant` and updates the `ParameterValues`?
> 
> Thanks for taking this on by the way! I think it looks good, but is somewhat separate from the core code.

Sure it could be automated in a local function as well... but it is just simpler to have this in the pybamm integrated for the end users.

We have 8 different simulation scenarios here in different notebooks (cycle and calendar ageing with different geometries and different submodels enabled) which would either require to copy-paste that function to each notebook, or to locally create a Python module with this function that can be imported from each file (the notebooks are not all in the same directory, so I can't just write a single python file and import it). I can still do all of this locally, sure, but the end users of this will be chemists and not programmers (at least in our case) and it will make their life a lot easier to have it integrated.

And I don't think it is so much off-topic to have this here, as I said it is completely impossible practically to produce a uniform grid of measurement values, and people will be glad I found I way to tackle that problem with the random forest regressor.

tinosulzer:
Why not just save the output of `PybammStandardParameterClass.get_interpolant_data` to a csv? Then there is a single pre-processing step the user needs to do locally if adding their own data, and if using other people's data it has already been transformed to the uniform grid (which is what we need to input into casadi anyway) and then they can use a `[2D data]` tag without needing to write the interpolant manually. 

> the code does not add any dependencies (pickle is in the standard lib)

scikit-learn is required as a dependency to unpickle the file

> Sure it could be automated in a local function as well... but it is just simpler to have this in the pybamm integrated for the end users.

There are plenty of other cases in which this is true, but as maintainers we have to be careful about scope creep / feature bloat and make decisions about what to include and not to include. This falls into the category of things that people might want to do themselves in different ways. Similarly, we don't provide tools for fitting functions to OCV data, even though that might be convenient, since there are different ways of doing it. 

> or to locally create a Python module with this function that can be imported from each file

It sounds like this is worth doing, it's a useful functionality that might be of interest to people outside of pybamm

Finally, if you really want this feature included in a single package, you are free to fork and install that version as your own pybamm

ehtec:
> Why not just save the output of `PybammStandardParameterClass.get_interpolant_data` to a csv? Then there is a single pre-processing step the user needs to do locally if adding their own data, and if using other people's data it has already been transformed to the uniform grid (which is what we need to input into casadi anyway) and then they can use a `[2D data]` tag without needing to write the interpolant manually.
> 
> > the code does not add any dependencies (pickle is in the standard lib)
> 
> scikit-learn is required as a dependency to unpickle the file
> 
> > Sure it could be automated in a local function as well... but it is just simpler to have this in the pybamm integrated for the end users.
> 
> There are plenty of other cases in which this is true, but as maintainers we have to be careful about scope creep / feature bloat and make decisions about what to include and not to include. This falls into the category of things that people might want to do themselves in different ways. Similarly, we don't provide tools for fitting functions to OCV data, even though that might be convenient, since there are different ways of doing it.
> 
> > or to locally create a Python module with this function that can be imported from each file
> 
> It sounds like this is worth doing, it's a useful functionality that might be of interest to people outside of pybamm
> 
> Finally, if you really want this feature included in a single package, you are free to fork and install that version as your own pybamm

Ok, if you really don't want to use the PKL file I will do on Friday as you suggested... personally I don't have a problem with the PKL files used, but if you really don't want that I will add the extra 2D CSV step, it's your software in the end.

sklearn would be required as a dependency to unpickle the file, this is true. But people who do not want to use the feature at all would not need to install it.

Creating and maintaining an own fork is not really worth it for that, so I'll implement the changes later this week

tinosulzer:
Great, thanks!

ehtec:
Done. I am reading the data from JSON now, because CSV is not really suitable, we would need to do some error prone reshaping, JSON can handle nested lists much easier.

Format is `{'data': [[x1_values_1D, x2_values_1D], y_values_2D]}`. It is exactly what the `PybammStandardParameterClass.get_interpolation_parameters` method was returning.

ehtec:
https://github.com/ehtec/pybamm-param-ml

Here are the external tools.

pybamm/expression_tree/interpolant.py
```diff
@@ -93,7 +96,9 @@ def __init__(
                         x1, y.T, bounds_error=False, fill_value="extrapolate"
                     )
             elif len(x) == 2:
+                self.dimension = 2
                 interpolating_function = interpolate.interp2d(x1, x2, y)
+                # interpolating_function = lambda *params: grid_interpolating_function(*params).flatten()
```
tinosulzer, line 101: can this line be removed?

pybamm/parameters/parameter_values.py
```diff
@@ -323,6 +324,29 @@ def update(self, values, check_conflict=False, check_already_exists=True, path="
...
                     values[name] = (function_name, data)
+
+                # Machine learning data from Random Forest Regressor
+                elif value.startswith("[current 2D data]") or value.startswith("[2D data]"):
+                    if value.startswith("[current 2D data]"):
+                        data_path = os.path.join(
+                            pybamm.root_dir(), "pybamm", "input", "drive_cycles"
+                        )
+                        filename = os.path.join(data_path, value[17:] + ".json")
+                        function_name = value[17:]
```
tinosulzer, line 335: What is the use case for current 2D data? Won't it always be a function of time only?

pybamm/parameters/parameter_values.py
```diff
@@ -638,28 +662,60 @@ def _process_symbol(self, symbol):
...
+                            )
+                        )
+                        self.parameter_events.append(
+                            pybamm.Event(
+                                "Interpolant {} upper bound".format(name),
+                                pybamm.min(max(data[0][data_index]) - new_children[data_index]),
+                                pybamm.EventType.INTERPOLANT_EXTRAPOLATION,
+                            )
+                        )
+
```
tinosulzer, line 715: no longer needed

pybamm/parameters/parameter_values.py
```diff
@@ -638,28 +662,60 @@ def _process_symbol(self, symbol):
...
+                else:
+                    raise Exception("Invalid function name length: {0}".format(len(function_name)))
```
tinosulzer, line 717: I don't think this exception is needed, since this is an internal (not user-facing) function.

tinosulzer (changes requested):
Thanks, code looks good. Can you add a test to https://github.com/pybamm-team/PyBaMM/blob/develop/tests/unit/test_parameters/test_parameter_values.py

pybamm/parameters/parameter_values.py
ehtec: To be honest I am not 100% sure what the `current data` does. I just copied the code above for 1D and adapted it for my purposes.

ehtec: It looks to me like the `current data` uses a different root path when searching for the file from what I see from the source code...

pybamm/parameters/parameter_values.py
ehtec: This is needed. Removing it will break my feature. Why do you think it is not needed?

pybamm/parameters/parameter_values.py
ehtec: It is an internal exception, but I like to use such exceptions. If I have multiple `if` - `elif` - `elif` statements I always consider the possibility that there might somewhere be a problem or bug, and without the `else` statement we have undefined behavior. Here it is not so critical but especially in Web apps it is important, not doing it can lead to vulnerabilities like DoS

pybamm/parameters/parameter_values.py
tinosulzer: Ah yeah my bad I thought you were doing it differently

pybamm/parameters/parameter_values.py
tinosulzer: Yeah, since current profiles don't need to be associated with a particular chemistry unlike other parameters. You can get rid of this branch

pybamm/parameters/parameter_values.py
```diff
@@ -638,28 +662,60 @@ def _process_symbol(self, symbol):
...
-                )
+                    self.parameter_events.append(
+                        pybamm.Event(
+                            "Interpolant {} upper bound".format(name),
+                            pybamm.min(max(data[:, 0]) - new_children[0]),
+                            pybamm.EventType.INTERPOLANT_EXTRAPOLATION,
+                        )
+                    )
+
+                elif len(function_name) == 3:  # ML data
```
tinosulzer, line 690: Instead of adding "padding" and then branching on the length of "function_name", can you add an entry to both 1D data and 2D data like "1D" and "2D"

pybamm/parameters/parameter_values.py
```diff
@@ -638,28 +662,60 @@ def _process_symbol(self, symbol):
...
+                                pybamm.EventType.INTERPOLANT_EXTRAPOLATION,
+                            )
+                        )
+                        self.parameter_events.append(
+                            pybamm.Event(
+                                "Interpolant {} upper bound".format(name),
+                                pybamm.min(max(data[0][data_index]) - new_children[data_index]),
+                                pybamm.EventType.INTERPOLANT_EXTRAPOLATION,
+                            )
+                        )
```
tinosulzer, line 714: these lines could be written in such a way as to add these events for both 1D and 2D data, instead of repeating as they are now

pybamm/parameters/parameter_values.py
tinosulzer: If you can write a test that triggers this line of code then we can keep it. Also, please use a more specific exception

tinosulzer (changes requested):
Test still needs to be added and CI needs to pass, including style checks. You can test this locally using `python run-tests.py --quick`

ehtec:
Ok, I will use a more specific exception. Here is how you trigger it:

```python
import pybamm as pb
pv = pb.ParameterValues({"testparam": (1, 2, 3, 4, 5)})
the_symbol = pb.FunctionParameter("testparam", {})
pv._process_symbol(the_symbol)
```

Should this also be added to a test? This is just an example, there are for sure other ways also how this can be triggered.

ehtec:
@tinosulzer do you want me to add only a `test_process_interpolant_2d` test or do you also need an equivalent to the `test_interpolant_against_function` test? This would require a 2d example JSON data file to be added. Should I do this or is the first test sufficient?

ehtec:
`test_process_interpolant_2d` has been added.

pybamm/parameters/parameter_values.py
```diff
@@ -321,8 +324,24 @@ def update(self, values, check_conflict=False, check_already_exists=True, path="
                         filename, comment="#", skip_blank_lines=True, header=None
                     ).to_numpy()
                     # Save name and data
+                    self._dict_items[name] = (function_name, ([data[:, 0]], data[:, 1]))
+                    values[name] = (function_name, ([data[:, 0]], data[:, 1]))
+
+                # Machine learning data from Random Forest Regressor
```
tinosulzer, line 330: comment should be more general

pybamm/parameters/parameter_values.py
```diff
@@ -321,8 +324,24 @@ def update(self, values, check_conflict=False, check_already_exists=True, path="
...
+                    filename = os.path.join(path, value[9:] + ".json")
+                    function_name = value[9:]
+                    filename = pybamm.get_parameters_filepath(filename)
+                    with open(filename, 'r') as jsonfile:
+                        json_data = json.load(jsonfile)
+                    data = json_data['data']
+                    data[0] = [np.array(el) for el in data[0]]
+                    data[1] = np.array(data[1])
+                    # Save name and data
+                    # Add padding to see a difference from CSV imported data
```
tinosulzer, line 341: comment should be more general

pybamm/parameters/parameter_values.py
```diff
@@ -637,29 +656,47 @@ def _process_symbol(self, symbol):
...
-                        pybamm.min(new_children[0] - min(data[:, 0])),
-                        pybamm.EventType.INTERPOLANT_EXTRAPOLATION,
-                    )
-                )
-                self.parameter_events.append(
-                    pybamm.Event(
-                        "Interpolant {} upper bound".format(name),
-                        pybamm.min(max(data[:, 0]) - new_children[0]),
-                        pybamm.EventType.INTERPOLANT_EXTRAPOLATION,
+                if len(function_name) == 2:  # ML data
```
tinosulzer, line 659: not necessarily ML data

tests/unit/test_parameters/test_parameter_values.py
```diff
@@ -507,6 +507,68 @@ def test_process_interpolant(self):
...
+
+        # process differentiated function parameter
+        # diff_func = func.diff(a)
+        # processed_diff_func = parameter_values.process_symbol(diff_func)
+        # self.assertEqual(processed_diff_func.evaluate(), 2)
+
+        # interpolant defined up front
+        interp2 = pybamm.Interpolant(data[0], data[1], children=(a, b))
+        processed_interp2 = parameter_values.process_symbol(interp2)
+        self.assertEqual(processed_interp2.evaluate(), 14.82)  # 6.02
```
tinosulzer, line 544: comments at the end of these lines are redundant, and confusing when out of date

tests/unit/test_parameters/test_parameter_values.py
```diff
@@ -507,6 +507,68 @@ def test_process_interpolant(self):
...
+
+        data3 = x_, Y3
+
+        parameter_values = pybamm.ParameterValues(
+            {"a": 3.01, "b": 4.4, "Times two": ("times two", data3)}
+        )
+
+        a = pybamm.Parameter("a")
+        b = pybamm.Parameter("b")
+        func = pybamm.FunctionParameter("Times two", {"a": a, "b": b})
```
tinosulzer, line 558: times three?

pybamm/parameters/parameter_values.py
```diff
@@ -637,29 +656,47 @@ def _process_symbol(self, symbol):
...
+                    name, data = function_name
+
+                    if isinstance(data, np.ndarray):
+                        data = [data[:, 0]], data[:, 1]
+
+                    if len(data[0]) == 1:
+                        input_data = data[0][0], data[1]
+
+                    else:
+                        input_data = copy.deepcopy(data)
```
tinosulzer, line 670: why deepcopy here? `data` is not modified in-place anywhere as far as I can see

tinosulzer (changes requested):
Getting there. All lines of code should be covered. This will require storing some `json` file for testing purposes for the 2D date, which is fine

ehtec:
Where should I store the JSON file (which directory is suitable for this?) And should I just try to create the interpolant object, or should I test it also with a simulation example script? The latter would test really all lines of code, but increase the test execution time by several minutes with the currently used cycling example. Is this a problem?

pybamm/parameters/parameter_values.py
ehtec: Would not be necessary but I have the habit of only passing references to objects if I really intend to pass references. I would do it like this in my own projects, if you need it removed I can do so...

ehtec:
Ok, I have managed to simplify the test experiment using the interpolant so it takes only 1 min to execute. Is this ok for a test? Where should I put the test function? Here?

https://github.com/pybamm-team/PyBaMM/blob/develop/tests/unit/test_simulation.py

tinosulzer:
1 minute is a bit long for a single test. Just testing the parameter can be read should be fine for now. You can add the json file to the folder the test is in

tinosulzer:
For the other error messages that can't be tested, add `#pragma: no cover` after the `else` statement

pybamm/parameters/parameter_values.py
tinosulzer: I see, yes in that case remove it please

ehtec:
Done.

tinosulzer:
Still need to add tests/skips for the remaining coverage misses

ehtec:
> For the other error messages that can't be tested, add `#pragma: no cover` after the `else` statement

You mean the `convert_to_casadi.py` tests with the Interpolant in 2D? Where should these be added?

I think I have tested everything else (reading and using 2D data is done in the test, this should cover all the new lines in `parameter_values.py` and `interpolant.py`).

tinosulzer:
Casadi tests here: https://github.com/pybamm-team/PyBaMM/blob/develop/tests/unit/test_expression_tree/test_operations/test_convert_to_casadi.py
Generally, the test directory structure is the same as the source code

For the remaining missing coverage, are you able to access this page to see the misses (by clicking on the x next to codecov/patch)? https://github.com/pybamm-team/PyBaMM/pull/1805/checks?check_run_id=4337573365

ehtec:
@tinosulzer I think it is fine now.

tinosulzer:
Just need to remove that deepcopy, it's not needed

tinosulzer (approved):

tinosulzer:
Thanks @ehtec !

tinosulzer:
@all-contributors add @ehtec for code

PR #1821, opened 2021-11-22
Issue 1781 utilization

### LINKED ISSUE #1781: Scaling of interfacial currents with particle shape
### Discussed in https://github.com/pybamm-team/PyBaMM/discussions/1769

<sup>Originally posted by **dion-w** October 31, 2021</sup>
hello everyone,
i have a question concerning the submodel "particle shape" : "user". I know that with this parameter you can reduce the amount of surface of your active material. For example if the particle shape is non-spherical or if some parts are blocked by the conductive binder fraction.

What i wonder: I tried some simulations with the user-supplied particle shape. Normally when you supply a very low current to your cell you get almost the theorethical capacity out of your cell, but this is not the case here. For a setup with an electrode capacity of 6mAh/cm² (positive electrode thickness ~100 µm) i almost loose 2 mAh/cm² at C/10 when using a user-supplied particle shape instead of the standard 3 e_s / R.

I think this is due to incorrectly scaling the molar flux in the microscale. Normally (and for this i refer to the [supporting_information.pdf](https://github.com/pybamm-team/PyBaMM/files/7448901/am8b08993_si_001.pdf) of [this paper](https://pubs.acs.org/doi/abs/10.1021/acsami.7b17771)) when you use a user-supplied surface area  you only set the surface area of the macroscale to this specific value. To ensure **charge conservation** the molar flux in the boundary needs to be scaled withe the factor of user-supplied-particle shape divided by 3*eps_s/R (see **equation S22 in the attached supporting information**).

Nonetheless i don't know how to fix this in the code. I only have an idea so far which i will show for the positive electrode (same for negative electrode):

- _pybamm/parameters/lithium_ion_parameters.py_: In method _set_scales() starting at Line 532, introduce not only a_p_typ and but also a blockage quotient (maybe called "a_block") being calculated  with  . 

- _pybamm/models/submodels/interface/kinetics/butler_volmer.py_:  could be used here in line 35 as an additional prefactor. Maybe it can used the same in other _get_kinetics methods of the interface subomdell, i don't know.

There may be a better name than  since if  is 1, no current is blocked but i didn't come up with a better name so far.

So what do you think about it?

</div>

tinosulzer:
# Description

- Remove the "user" option for particle shape. Particle shape is now determined by geometry
- Add "utilisation" submodels

Fixes #1781 

## Type of change

# Key checklist:

## Further checks:


tests/integration/test_models/test_full_battery_models/test_lithium_ion/base_lithium_ion_tests.py
```diff
@@ -93,6 +93,38 @@ def test_particle_quartic(self):
         options = {"particle": "quartic profile"}
         self.run_basic_processing_test(options)
 
+    def test_constant_utilisation(self):
+        options = {"interface utilisation": "constant"}
+        parameter_values = pybamm.ParameterValues(
+            chemistry=pybamm.parameter_sets.Marquis2019
```
brosaplanella, line 99: Hasn't the keyword argument been deprecated?

brosaplanella (approved):
Looks good, just some small comments.

PR #1837, opened 2021-11-30
fixed downloading of wheels

brosaplanella:
# Description

This should make the workflow to publish the wheels work as before (see details [here](https://github.com/actions/download-artifact/tree/v2#compatibility-between-v1-and-v2)).

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -79,6 +79,7 @@ jobs:
         uses: actions/download-artifact@v2
         with:
           name: wheels
+          path: wheels
 
       - name: Publish wheels on PyPI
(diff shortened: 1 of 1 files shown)


brosaplanella:
Link to test run: https://github.com/pybamm-team/PyBaMM/actions/runs/1522183651

tinosulzer (approved):
Thanks! Why does something break every time :(

brosaplanella:
> Thanks! Why does something break every time :(

That's the fun of open-source!

Alt Text

brosaplanella:
The fix worked. The test just failed as I don't have a PyPI account to publish the wheels, so I am merging it.

PR #1843, opened 2021-12-03
Averages classes

tinosulzer:
# Description

Make classes for averages (`XAverage`, `YZAverage`, ...) so that they can be recognized more easily that trying to find a division of integrals

## Type of change

# Key checklist:

## Further checks:


pybamm/expression_tree/averages.py
```diff
@@ -0,0 +1,366 @@
...
+        self.f_a_dist = f_a_dist
+
+    def _unary_new_copy(self, child):
+        """See :meth:`UnaryOperator._unary_new_copy()`."""
+        return size_average(child, f_a_dist=self.f_a_dist)
+
+
+def x_average(symbol):
+    """
+    convenience function for creating an average in the x-direction.
```
brosaplanella, line 95:
```suggestion
    Convenience function for creating an average in the x-direction.
```

Not sure if what I am suggesting goes against the standard format. If not, note that the change also applies to all the other functions below.

pybamm/expression_tree/averages.py
```diff
@@ -0,0 +1,366 @@
...
+            )
+        ):
+            aux = {"secondary": symbol.auxiliary_domains["secondary"]}
+            if "quaternary" in symbol.auxiliary_domains:
+                aux["tertiary"] = symbol.auxiliary_domains["quaternary"]
+            return pybamm.FullBroadcast(symbol.orphans[0], symbol.broadcast_domain, aux)
+        else:  # pragma: no cover
+            # It should be impossible to get here
+            raise NotImplementedError
+    # If symbol is a concatenation of Broadcasts, its average value is its child
```
brosaplanella, line 158: Probably a very dumb question but I don't get why that is the case should you take the average of each child and then average them together? (maybe it is what you are doing already, but didn't understand it).

brosaplanella (approved):
Looks good, thanks Tino! Just a couple of minor comments. You also need to add a line to CHANGELOG before merging.

pybamm/expression_tree/averages.py
tinosulzer: That is what I'm doing, I'll change the comment

PR #1859, opened 2021-12-15
default stress-induced diffusion 

rtimms:
# Description

If no option is provided for "particle mechanics" or "stress-induced diffusion" then the default stress-induced diffusion option is set based on the default particle mechanics option. For example, if the user sets up a model with stress-driven LAM
```python3
model = pybamm.lithium_ion.SPM({"loss of active material": "stress-driven"})
```
then the default particle mechanics option is automatically changed to "swelling only". In <= v21.10 stress-induced diffusion was hardcoded, so was automatically included with particle mechanics. As of v21.11 "stress-induced diffusion" is a separate option, so wasn't included in the above example. After this PR stress-induced diffusion _will_ be included in the above model since the default mechanics option has changed to "swelling only".

In summary the default options for the above model will be:

- <= v21.10: `{"particle mechanics": "swelling only"}` (stress-induced diffusion included since it is hardcoded)
- v21.11:  `{"particle mechanics": "swelling only", "stress-induced diffusion": "false"}`
- After this PR: `{"particle mechanics": "swelling only", "stress-induced diffusion": "true"}` (i.e. same behaviour as in <= v21.10)

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/models/full_battery_models/base_battery_model.py
+++ b/pybamm/models/full_battery_models/base_battery_model.py
@@ -243,9 +243,15 @@ class BatteryModelOptions(pybamm.FuzzyDict):
         # provided
 
         # Change the default for stress-induced diffusion based on which particle
-        # mechanics option is provided
+        # mechanics option is provided. If the user doesn't supply a particle mechanics
+        # option set the default stress-induced diffusion option based on the default
+        # particle mechanics option which may change depending on other options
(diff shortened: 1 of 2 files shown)


tinosulzer (approved):
Thanks!

PR #1865, opened 2021-12-17
Issue 1749 lam

### LINKED ISSUE #1749: Option to have LAM both due to SEI and stress
**Summary**
 For now, the LAM in the electrode can either be SEI driven or Particle stress driven. But LAM might be due to both simultaneously.   I would like the option to do that.

 **Motivation**
Looking at the data I have, particle stress LAM doesn't match data at lower C-rates if the parameter was tuned for higher C-rates. This is because LAM at lower C-rates might be dominated by SEI growth. So having both mechanisms simultaneously cause LAM might explain my data correctly.

tinosulzer:
# Description

Fixes #1749 

Use "stress and reaction-driven" for both models together.

## Type of change

# Key checklist:

## Further checks:


pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -423,6 +428,32 @@ def print_detailed_options(self):
...
+        return BatteryModelDomainOptions(self.items(), 0)
+
+    @property
+    def positive(self):
+        "Returns the options for the positive electrode"
+        # index 1 in a 2-tuple for the positive electrode
+        return BatteryModelDomainOptions(self.items(), 1)
+
+
+class BatteryModelDomainOptions(dict):
```
rtimms, line 444: neat :)

rtimms (approved):
thanks!

TomTranter (approved):
Looks good to me. I ran a script using Ai2020 params and borrowed some SEI and plating ones from Chen2020 to test all the deg models together. Might be worth adding as a notebook to show the effect of turning each one on and off with a few typical values. I can look into adding this in the new year

PR #1874, opened 2021-12-27
Make `PyBaMM` importable in `Linux` systems if `jax` is already installed

Saransh-cpp:
# Description

- Check for the version of `jax` in `have_jax()`
- Give an error in `install_jax()` if `jax` is already available with a different version

Additionally, the cell with `pybamm.print_citations()` gives an error which is described here - https://github.com/pybamm-team/BattBot/issues/47 - and is most probably an error within the `pybtex` library.

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/util.py
+++ b/pybamm/util.py
@@ -370,8 +370,7 @@ def install_jax():
     elif importlib.util.find_spec("jax") is not None:
         if not is_jax_compatible():
             raise ValueError(
-                "Jax is already installed",
-                "but the installed version of jax or jaxlib is not supported by PyBaMM",
+                "Jax is already installed but the installed version of jax or jaxlib is not supported by PyBaMM",  # noqa: E501
             )
     else:
         subprocess.check_call(
--- a/pybamm/util.py
+++ b/pybamm/util.py
@@ -21,6 +21,10 @@ import numpy as np
 
 import pybamm
 
+# versions of jax and jaxlib compatible with PyBaMM
+JAX_VERSION = "0.2.12"
+JAXLIB_VERSION = "0.1.70"
+
 
 def root_dir():
     """return the root directory of the PyBaMM install directory"""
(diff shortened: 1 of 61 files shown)


brosaplanella (approved):
Looks good, thanks Saransh! For a more robust fix, would it be possible to check if the `jax` version is the correct one and if not, throw a warning and treat the code as if `jax` wasn't installed?

Saransh-cpp:
> Looks good, thanks Saransh! For a more robust fix, would it be possible to check if the `jax` version is the correct one and if not, throw a warning and treat the code as if `jax` wasn't installed?

I was trying to uninstall the existing version of `jax` while installing `pybamm` through a function but it felt weird editing a user's environment. The warning sounds really good! I'll try adding that instead of editing the notebooks.

pybamm/util.py
```diff
@@ -356,6 +367,11 @@ def install_jax():
 
     if system() == "Windows":
         raise NotImplementedError("Jax is not available on Windows")
+    elif importlib.util.find_spec("jax") is not None:
+        if not is_jax_compatible():
+            raise ValueError(
```
Saransh-cpp, line 372: Codecov shows that this complete function is not included in the coverage, should I add `# pragma: no cover` for this function?

priyanshuone6:
We should create a global variable for jax version, jaxlib version and use it in have_jax, is_jax_compatible

pybamm/util.py
brosaplanella: That sounds good

brosaplanella: Also, this is now an error. Should we make it a warning and proceed without `jax`?

Saransh-cpp: PyBaMM installation (and importing PyBaMM) proceeds without `jax` if the installed `jax` version is not compatible. But this function throws an error if a user tries to install `jax` and has an incompatible `jax` version already installed.

Should it instead uninstall the existing version and install the compatible version?

brosaplanella: Sorry, I misunderstood! That works as it is :)

brosaplanella:
Looks great now, I will wait until tomorrow in case someone else wants to give it a look.

brosaplanella:
@all-contributors please add @Saransh-cpp for maintenance

PR #1878, opened 2021-12-30
Fix sdist workflow

priyanshuone6:
Fail the workflow if build_wheels fails


brosaplanella (approved):
Looks good, thanks Priyanshu!

PR #1880, opened 2021-12-31
Callback

tinosulzer:
# Description

Add callback hooks to `Simulation.solve` and a move some logging steps to the callback

## Type of change

# Key checklist:

## Further checks:


tinosulzer:
@martinjrobins I think this could be useful for automatically uploading simulation results from a script

tests/unit/test_experiments/test_simulation_with_experiment.py
```diff
@@ -176,16 +177,50 @@ def test_run_experiment_drive_cycle(self):
...
+            model,
+            experiment=experiment,
+            parameter_values=parameter_values,
+        )
+        sol = sim.solve()
+        self.assertEqual(len(sol.cycles), 1)
+        self.assertEqual(len(sol.cycles[0].steps), 1)
+
+        # Different callback
+        sol = sim.solve(callbacks=pybamm.callbacks.Callback())
```
brosaplanella, line 223: What are you testing with this callback here?

brosaplanella (approved):
Looks good, thanks Tino! I guess we should extend the example notebook but we can do this in the future. Before merging, you need to update CHANGELOG.

pybamm/callbacks.py
```diff
@@ -0,0 +1,243 @@
...
+        pass
+
+    def on_experiment_infeasible(self, logs):
+        """
+        Called when an experiment simulation is infeasible.
+        """
+        pass
+
+
+########################################################################################
```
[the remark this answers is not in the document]
tinosulzer: I added these because it indicates that the functions below the class affect only this class

tests/unit/test_experiments/test_simulation_with_experiment.py
tinosulzer: I think it was just coverage

PR #1890, opened 2022-01-07
update license year and add action

tinosulzer:
The action automatically updates the license every year

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,4 +1,4 @@
-Copyright (c) 2022, the PyBaMM team.
+Copyright (c) 2018-2021, the PyBaMM team.
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,4 +1,4 @@
-Copyright (c) 2018-2021, the PyBaMM team.
+Copyright (c) 2022, the PyBaMM team.
 All rights reserved.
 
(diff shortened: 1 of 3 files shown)


priyanshuone6:
Should we update this too?
https://github.com/pybamm-team/PyBaMM/blob/c8a01ba72065a81e3b1c95586361cd76d326adc4/docs/conf.py#L25

tinosulzer:
@priyanshuone6 can you set up automatic updating for the file here?

LICENSE.txt
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2018-2021, the PyBaMM team.
+Copyright (c) 2022, the PyBaMM team.
```
priyanshuone6, line 1: @tinosulzer I tried running the `FantasticFiasco/action-update-license-year@v2` action on my fork and it is creating a range of years instead of a single year (See https://github.com/priyanshuone6/PyBaMM/pull/22/files). Should we include a single year or a range of years?

tinosulzer: Yeah let's do the range. I should have kept it as a range

priyanshuone6:
Ready to merge now, See https://github.com/priyanshuone6/PyBaMM/pull/25 for test workflow run

tinosulzer:
Thanks! Somebody else needs to approve, since I opened

brosaplanella (approved):
Looks great, thanks guys!

PR #1893, opened 2022-01-08
#1889 fix some parameter bugs and add timescale option

### LINKED ISSUE #1889: Scalar timescale
I'd like to implement a way to set a scalar or typical timescale. This is because when solving a jellyroll model or liionpack model if I set electrode height to an input then this changes the timescale along with a few other parameters. It would simplify things to just have a constant value across all the simulations using the model. If I replace the model.timescale with a Scalar though I see different results when changing the scalar value. See below for example:
```
import pybamm

model1 = pybamm.lithium_ion.SPM()
model2 = pybamm.lithium_ion.SPM()

symbolic_tau = pybamm.LithiumIonParameters().tau_discharge
tau = parameter_values.process_symbol(symbolic_tau)
tau_eval = tau.evaluate()
model2.timescale = pybamm.Scalar(2*tau_eval)

sim1 = pybamm.Simulation(model=model1, parameter_values=parameter_values)
sim2 = pybamm.Simulation(model=model2, parameter_values=parameter_values)

sol1 = sim1.solve([0, 1800])
sol2 = sim2.solve([0, 1800])

pybamm.QuickPlot(solutions=[sol1, sol2]).dynamic_plot()
```

tinosulzer:
# Description

Fixes #1889 

`model.timescale` cannot be changed any more. `model._timescale` can in theory but users really shouldn't. Instead, the "timescale" option can be passed to the model. The simulation results are independent of this (new test added for this).

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/tests/integration/test_models/test_full_battery_models/test_lead_acid/test_compare_outputs.py
+++ b/tests/integration/test_models/test_full_battery_models/test_lead_acid/test_compare_outputs.py
@@ -96,20 +96,25 @@ class TestCompareOutputs(unittest.TestCase):
         """
         Changing the timescale (via options) should not change the result.
         """
-        model1 = pybamm.lithium_ion.SPM()
-        parameter_values = model1.default_parameter_values
-        tau = parameter_values.evaluate(model1.param.timescale)
-        model2 = pybamm.lithium_ion.SPM({"timescale": 2 * tau})
+        for model_class in [pybamm.lithium_ion.SPM, pybamm.lithium_ion.DFN]:
+            model1 = model_class()
+            parameter_values = model1.default_parameter_values
+            tau = parameter_values.evaluate(model1.param.timescale)
+            model2 = model_class({"timescale": 2 * tau})
 
-        sim1 = pybamm.Simulation(model1)
-        sim2 = pybamm.Simulation(model2)
-        sol1 = sim1.solve([0, 3600], solver=pybamm.CasadiSolver(rtol=1e-8, atol=1e-8))
-        sol2 = sim2.solve([0, 3600], solver=pybamm.CasadiSolver(rtol=1e-8, atol=1e-8))
+            sim1 = pybamm.Simulation(model1)
+            sim2 = pybamm.Simulation(model2)
+            sol1 = sim1.solve(
+                [0, 3600], solver=pybamm.CasadiSolver(rtol=1e-8, atol=1e-8)
+            )
+            sol2 = sim2.solve(
+                [0, 3600], solver=pybamm.CasadiSolver(rtol=1e-8, atol=1e-8)
+            )
 
-        # sol.t should be different (different internal scalings for the solver)
-        np.testing.assert_array_equal(sol1.t, 2 * sol2.t)
-        # sol.y should be identical (within solver tolerance)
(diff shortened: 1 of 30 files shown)


tests/integration/test_models/test_full_battery_models/test_lead_acid/test_compare_outputs.py
```diff
@@ -92,6 +92,25 @@ def test_compare_outputs_surface_form(self):
             comparison = StandardOutputComparison(solutions)
             comparison.test_all(skip_first_timestep=True)
 
+    def test_compare_outputs_timescale(self):
```
TomTranter, line 95: This works well for SPM but if you run DFN you see differences in solutions. Is that just because of solver?

TomTranter (commented):
Thanks a lot Tino, I ran the following and solutions are quite different:
```
import pybamm

parameter_values.update({'Current function [A]': "[input]"})

model1 = pybamm.lithium_ion.DFN()

symbolic_tau = model1.param.tau_discharge
tau = parameter_values.process_symbol(symbolic_tau)
tau_eval = tau.evaluate()

model2 = pybamm.lithium_ion.DFN({"timescale":2*tau_eval})
sim1 = pybamm.Simulation(model=model1, parameter_values=parameter_values,
                         solver=pybamm.CasadiSolver(rtol=1e-8, atol=1e-8))
sim2 = pybamm.Simulation(model=model2, parameter_values=parameter_values,
                         solver=pybamm.CasadiSolver(rtol=1e-8, atol=1e-8))

sim1.build()
sim2.build()

sol1 = sim1.solve([0, 1800], inputs={'Current function [A]': 5.0})
sol2 = sim2.solve([0, 1800], inputs={'Current function [A]': 5.0})

plot = pybamm.QuickPlot(solutions=[sol1, sol2])
plot.dynamic_plot()
```

tinosulzer:
Ah yeah. It looks more like a bug in one the DFN equations, since it's only off by a little bit, I'll take a look

tinosulzer:
Ok fixed. Will try SEI models too just to check.

Since we now have this feature, do you think we should enforce that there can't be any input parameters in `model.timescale`? The solver currently accepts this, but errors if the inputs change.

TomTranter:
> Ok fixed. Will try SEI models too just to check.
> 
> Since we now have this feature, do you think we should enforce that there can't be any input parameters in `model.timescale`? The solver currently accepts this, but errors if the inputs change.

Would be good If you want to use a parameter that forms part of the default timescale as an input then you get an error and told to use a constant one with the new functionality.

rtimms:
> Would be good If you want to use a parameter that forms part of the default timescale as an input then you get an error and told to use a constant one with the new functionality.

I agree. We should also raise an error up front if you make a parameter that appears in the geometry an input. Currently we throw an error if the lengthscales change, but it might be easier for the user to understand a message that says something like "you can't make the electrode thickness an input as it changes the geometry"

TomTranter (commented):
Still seems to be failing a test

tinosulzer:
@martinjrobins could you see what's going on with the failing test? I can't install IDA KLU on mac M1, but it passes if I switch to the casadi solver

rtimms (approved):
thanks @tinosulzer !

PR #1894, opened 2022-01-10
#1886 added fix within processed variable

### LINKED ISSUE #1886: Order of sensitivites
- PyBaMM version - 21.12:
- Python version - 3.8.10:

**Describe the bug**
The order of sensitivities seems to not align correctly when there are multiple sensitivities. The example below shows the sensitivity of terminal voltage SPM to D_n in two cases. Case 1 where only D_n is an input and Case 2 where D_n and b_n (negative electrolyte bruggeman coefficient) are inputs. In Case 2, it appears that the sensitivites of the terminal voltage w.r.t D_n and b_n have been switched. Seems likely a switching of order in either the building of the sensitivity rhs or when extracting from casadi. 

**To Reproduce**
```
import pybamm
import numpy as np

# Case 1
chemistry = pybamm.parameter_sets.Ecker2015
parameter_values["Negative electrode diffusivity [m2.s-1]"] = "[input]"

sim = pybamm.Simulation(model, parameter_values=parameter_values)

inputs = {
    "Negative electrode diffusivity [m2.s-1]": 1e-13,
}

t_eval = np.linspace(0, 30, 1000)
sim.solve(t_eval=t_eval, inputs=inputs, calculate_sensitivities=True)
sol1 = sim.solution


# Case 2
chemistry = pybamm.parameter_sets.Ecker2015
parameter_values["Negative electrode diffusivity [m2.s-1]"] = "[input]"
parameter_values["Negative electrode Bruggeman coefficient (electrolyte)"] = "[input]"

sim = pybamm.Simulation(model, parameter_values=parameter_values)

inputs = {
    "Negative electrode diffusivity [m2.s-1]": 1e-13,
    "Negative electrode Bruggeman coefficient (electrolyte)": 1.5,
}

sim.solve(t_eval=t_eval, inputs=inputs, calculate_sensitivities=True)
sol2 = sim.solution

# Plot
fig, ax = plt.subplots(2, 1, figsize=(16, 9), sharex=True)
ax[0].plot(
    t_eval,
    sol1["Terminal voltage [V]"].sensitivities[
        "Negative electrode diffusivity [m2.s-1]"
    ],
    label="sol1 D_n",
)
ax[0].plot(
    t_eval,
    sol2["Terminal voltage [V]"].sensitivities[
        "Negative electrode diffusivity [m2.s-1]"
    ],
    linestyle="--",
    label="sol2 D_n",
)

ax[1].plot(
    t_eval,
    sol1["Terminal voltage [V]"].sensitivities[
        "Negative electrode diffusivity [m2.s-1]"
    ],
    label="sol1 D_n",
)
ax[1].plot(
    t_eval,
    sol2["Terminal voltage [V]"].sensitivities[
        "Negative electrode Bruggeman coefficient (electrolyte)"
    ],
    linestyle="--",
    label="sol2 b",
)

ax[0].grid()
ax[1].grid()

ax[0].set_xlabel("Time [s]")
ax[1].set_xlabel("Time [s]")

ax[0].legend()
ax[1].legend()

plt.show()
```

**Expected behaviour**
The sensitivity of terminal voltage w.r.t D_n should be independent of the number of other inputs. 

**Screenshots**
Plot generated by above example. 

image

Scottmar93:
# Description

During the solver setup, the list of inputs/sensitivities is sorted by string and then used to assign elements of the y-array to each sensitivity. This applies the same sorting of the list before extracting the sensitivities during the process variables stage. 

I haven't added a test because I couldn't think of a good way without having a full integration test (passing a full model through the pipeline and then checking that the sensitivities are independent of the order of the inputs dictionary). However, I can add it if you think this is worth it for this. 

Fixes #1886 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# unreleased
+
+## Bug fixes
+-   Fixed issue in extraction of sensitivites ([#1894](https://github.com/pybamm-team/PyBaMM/pull/1894))
+
 # [v21.12](https://github.com/pybamm-team/PyBaMM/tree/v21.11) - 2021-12-29
 
 ## Features
@@ -7,8 +12,6 @@
 -   Added cylindrical geometry and finite volume method ([#1824](https://github.com/pybamm-team/PyBaMM/pull/1824))
 
 ## Bug fixes
-
--   Fixed issue in extraction of sensitivites ([#1894](https://github.com/pybamm-team/PyBaMM/pull/1894))
 -   `PyBaMM` is now importable in `Linux` systems where `jax` is already installed ([#1874](https://github.com/pybamm-team/PyBaMM/pull/1874))
 -   Simulations with drive cycles now support `initial_soc` ([#1842](https://github.com/pybamm-team/PyBaMM/pull/1842))
 -   Fixed bug in expression tree simplification ([#1831](https://github.com/pybamm-team/PyBaMM/pull/1831))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ## Bug fixes
 
--   Fixed issue in extraction of sensitivites ([#1886](https://github.com/pybamm-team/PyBaMM/issues/1886))
+-   Fixed issue in extraction of sensitivites ([#1886](https://github.com/pybamm-team/PyBaMM/pull/1894/files))
 -   `PyBaMM` is now importable in `Linux` systems where `jax` is already installed ([#1874](https://github.com/pybamm-team/PyBaMM/pull/1874))
 -   Simulations with drive cycles now support `initial_soc` ([#1842](https://github.com/pybamm-team/PyBaMM/pull/1842))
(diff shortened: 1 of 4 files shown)


tinosulzer (approved):
Thanks @Scottmar93 . I checked changing the order in [this test](https://github.com/pybamm-team/PyBaMM/blob/c8a01ba72065a81e3b1c95586361cd76d326adc4/tests/unit/test_solvers/test_casadi_solver.py#L573-L643) and indeed it fails, so it only used to work because p,q,r,s is already sorted. Could you change the order of the sensitivities in the test to make sure it still works with your fix?

Scottmar93:
@tinosulzer ah nice!! Yeah will do!

Scottmar93:
Okay, the original fix passed for some variables but not all so I have just changed it so that the inputs dictionary is sorted straight away upon being passed to the solver. This now passes the test with the dictionary keys in a different order.

martinjrobins (approved):
thanks @Scottmar93 , this looks good

Scottmar93:
I've tried merging from develop so hopefully that solves the testing issue. It looked like something that was failing on other branches as well though.

PR #1897, opened 2022-01-12
fix for when only a subset of input params are used for sensitivities

### LINKED ISSUE #1887: DAE Multiple inputs with fewer sensitivites
- PyBaMM version - 21.12:
- Python version - 3.8.10:

**Describe the bug**
An error arises within the algebraic solver if the number of sensitivities is different to the number of inputs. 

Error is: 
``Slice (start=0, stop=3783, step=1) out of bounds with supplied length of 2722``

I believe as a result of line 86 of casadi_algebraic_solver.py:
                 ```len_rhs = model.len_rhs * (inputs.shape[0] + 1)```
which calls inputs.shape[0]. Since there are more inputs than sensitivities, this results in trying to extract too many elements of rhs because less sensitivity equations have been added.

**To Reproduce**
```
import pybamm
import numpy as np

chemistry = pybamm.parameter_sets.Marquis2019
parameter_values["Negative electrode diffusivity [m2.s-1]"] = "[input]"
parameter_values["Negative electrode Bruggeman coefficient (electrolyte)"] = "[input]"

sim = pybamm.Simulation(model, parameter_values=parameter_values)

inputs = {
    "Negative electrode diffusivity [m2.s-1]": 1e-13,
    "Negative electrode Bruggeman coefficient (electrolyte)": 1.5,
}

sensitivities = ["Negative electrode diffusivity [m2.s-1]"]

t_eval = np.linspace(0, 30, 1000)
sim.solve(t_eval=t_eval, inputs=inputs, calculate_sensitivities=sensitivities)
```

**Expected behaviour**
Solving should not through an error when the sensitivities are fewer than the number of inputs. The length of the extracted rhs in casadi_algebraic_solver.py should depend upon the number of sensitivities and not the number of inputs.

martinjrobins:
Fixes #1887 

## Type of change

# Key checklist:

## Further checks:


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 ## Bug fixes
 
+-   Fix bug where sensitivity calculation failed if len of `calculate_sensitivities` was less than `inputs` ([#1897](https://github.com/pybamm-team/PyBaMM/pull/1897))
```
tinosulzer, line 11: needs to be in new section (unreleased)

tests/unit/test_solvers/test_casadi_solver.py
```diff
@@ -864,6 +864,46 @@ def test_solve_sensitivity_then_no_sensitivity(self):
...
+        model.initial_conditions = {var: r}
+
+        # only calculate the sensitivities of a subset of parameters
+        solver = pybamm.CasadiSolver(rtol=1e-10, atol=1e-10)
+        t_eval = np.linspace(0, 1, 80)
+        solution = solver.solve(
+            model,
+            t_eval,
+            inputs={"p": 0.1, "q": 2, "r": -1, "s": 0.5},
+            calculate_sensitivities=["p", "q"],
```
tinosulzer, line 884: Can we also test non-alphabetical order, and calculating with "r" and "s" only?

tinosulzer (approved):
Looks good, see suggestion for additional test

Scottmar93 (approved):

PR #1908, opened 2022-01-21
Fix failing build

priyanshuone6:
# Description

- The documentation build is still failing (https://readthedocs.org/projects/pybamm/builds/15835390/) as pybamm is just cloned and not installed to build docs
- Fix version and automate conf.py

### DIFF
--- a/.github/release_checklist.md
+++ b/.github/release_checklist.md
@@ -1,6 +1,7 @@
 - Run `scripts/update_version.py` to
   - Increment version number in
     - `pybamm/version.py`
+    - `docs/conf.py`
     - `CITATION.cff`
     - `vcpkg.json`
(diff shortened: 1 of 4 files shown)


Saransh-cpp (approved):
> as pybamm is just cloned and not installed to build docs

Oh yes, completely missed that!

Saransh-cpp:
@priyanshuone6 the URL checker is failing on the line -
https://github.com/pybamm-team/PyBaMM/blob/838422665fb4d8424abcb441f6123ab354cc1a63/scripts/update_version.py#L67

for the link (what it is taking as a link) -
```
🤔 Uh oh... The following urls did not pass:
https://github.com/pybamm-team/PyBaMM/tree/v
```

Can this URL be ignored in this PR?

priyanshuone6:
The current url checker fails a lot and catches valid links every once in a while. Will look into alternatives to replace it with

tinosulzer:
But in this case `https://github.com/pybamm-team/PyBaMM/tree/v` isn't actually a link, so we should ignore it

PR #1909, opened 2022-01-21
I1906 solution json

### LINKED ISSUE #1906: save a pybamm solution object as json
**Summary**

 Extend the `pybamm.Solution.save` function to be able to export a solution in json format

 **Motivation**
 
Like csv, json is a standard recognisable format for serialising data. Unlike csv, it is more suitable for serializing 1D or 2D variables

**Additional context**
 
The save function can already output in pickle, matlab or csv format

martinjrobins:
# Description

- Allow `Solution.save_data()` to return a string if filename is None
- add extra `to_format` of json

Fixes #1906 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/tests/unit/test_solvers/test_solution.py
+++ b/tests/unit/test_solvers/test_solution.py
@@ -242,7 +242,7 @@ class TestSolution(unittest.TestCase):
         # set variables first then save
         solution.update(["c", "d"])
         with self.assertRaisesRegex(ValueError, "pickle"):
-            solution.save_data(None, to_format="pickle")
+            solution.save_data(to_format="pickle")
         solution.save_data("test.pickle")
 
         data_load = pybamm.load("test.pickle")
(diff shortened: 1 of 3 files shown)


tests/unit/test_solvers/test_solution.py
```diff
@@ -268,11 +276,36 @@ def test_save(self):
...
+        # check string is the same as the file
+        with open('test.json') as f:
+            # need to strip \r chars for windows
+            self.assertEqual(
+                json_str.replace('\r', ''), f.read()
+            )
+
+        # check if string has the right values
+        json_data = json.loads(json_str)
+        np.testing.assert_array_almost_equal(json_data["c"], solution.data["c"])
```
tinosulzer, line 306: why is it only almost equal?

tinosulzer (approved):
Thanks, looks good to me!

tests/unit/test_solvers/test_solution.py
martinjrobins: I was only copying the other lines of this test. But it would need to be almost equal, as the numbers (e.g. 1/3) will be rounded off when they are written to json

PR #1912, opened 2022-01-21
Remove anytree

tinosulzer:
# Description

It seems like we can get away with not using `anytree` when creating expression trees, which saves a fair amount of overhead. Functionality like `render` still works fine, as long as the symbol has a `children` attribute.
`Symbol.new_copy` no longer needs to be used, but I have left the functions there in case we need to go back.

Let's give it a couple of releases to be sure, but if things are working well without `anytree` we can also:
- remove `Symbol.id` and implement `__eq__` and `__hash__` instead (previously, these were not implemented to avoid loops in anytree)
- Simplify `model.new_copy` to just make shallow copies

I'll open an issue for that if this PR is approved.

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 -   Added an option to force install compatible versions of jax and jaxlib if already installed using CLI ([#1881](https://github.com/pybamm-team/PyBaMM/pull/1881))
 
+## Optimizations
+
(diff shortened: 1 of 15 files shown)


tinosulzer:
🚀 
Screen Shot 2022-01-21 at 2 08 05 PM

martinjrobins (approved):
this looks great, glad to be rid of `anytree`!

PR #1917, opened 2022-01-24
Issue 1903 power

### LINKED ISSUE #1691: Simulate external resistance mode
**Summary**
Add the ability to simulate the battery connected to a constant or time-varying external load which is characterized by resistance.

 **Motivation**
 To simulate an external short circuit

**Additional context**
Use existing class PowerFunctionControl(FunctionControl):

### LINKED ISSUE #1903: Different constant-power approach
Instead of an extra algebraic equation, constant power could be achieved by specifying the current to be `P_applied / voltage`, where `voltage` is the boundary value of the potential.
This only works for the DFN, where potential is a variable. For SPM(e), potential is defined in terms of the current, so it would be circular, and we still need to include the algebraic equation.

tinosulzer:
# Description

Adds more operating modes for power and resistance control.

The explicit models can only be used with DFN without surface form, so it's not possible to completely eliminate algebraic equations using that (but it can be done using the differential models)

Fixes #1903 
Fixes #1691 

## Type of change

# Key checklist:

## Further checks:


examples/scripts/DFN.py
```diff
@@ -5,46 +5,41 @@
 import pybamm
```
rtimms, line 5: need to change the example script back

pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -186,7 +193,17 @@ def __init__(self, extra_options):
...
+            "operating mode": [
+                "current",
+                "voltage",
+                "power",
+                "differential power",
+                "explicit power",
+                "resistance",
+                "differential resistance",
+                "explicit resistance",
+                "CCCV",
```
rtimms, line 205: the CCCV option isn't listed in the docs

pybamm/models/submodels/external_circuit/explicit_control_external_circuit.py
```diff
@@ -0,0 +1,98 @@
...
+class ExplicitResistanceControl(BaseModel):
+    """External circuit with current set explicitly to hit target resistance."""
+
+    def __init__(self, param):
+        super().__init__(param)
+
+    def get_coupled_variables(self, variables):
+        param = self.param
+
+        # Current is given as applied power divided by voltage
```
rtimms, line 73: voltage divided by resistance

rtimms (approved):
thanks @tinosulzer looks good!

PR #1929, opened 2022-02-01
#1927 half-cell thermal models

### LINKED ISSUE #1927: lumped thermal model with half cell
Using a lumped thermal model with half cell gives an error because the options aren't passed to thermal models (so they don't know it's a half cell)

```python3
import pybamm

model = pybamm.lithium_ion.DFN(
    options={
        "working electrode": "positive",
        "thermal": "lumped",
    }
)
```

gives 

```
DomainError: Functions can only be applied to variables on the same domain
```

rtimms:
# Description
Modifies thermal submodels so they work with half-cell models

Fixes #1927 

## Type of change

# Key checklist:

## Further checks:


tinosulzer (commented):
There is still an anode in the half-cell, so it will contribute to the thermal problem in the same way as the current collectors do

rtimms:
> There is still an anode in the half-cell, so it will contribute to the thermal problem in the same way as the current collectors do

duh, sorry rushed into this. will fix it properly

tinosulzer:
No worries, would have done the same, only realized it as I was reviewing the PR

rtimms:
> No worries, would have done the same, only realized it as I was reviewing the PR. I guess "cell temperature" and "temperature that goes into the other equations (concentration/potential)" need to be separate variables

Yeah, you get away with it if you're happy with the interpretation that "Cell temperature" is either neg.elec./sep/pos.elec. or sep/work.elec. depending on if you are half-cell or not. But it means you can't use "x-full" with half-cells the way it is currently implemented.

tinosulzer (approved):
Thanks! We can just figure out x-full later if someone needs it

PR #1958, opened 2022-03-02
#1957 return heat source in isothermal models

### LINKED ISSUE #1957: [Bug]: isothermal model doesn't get heat source terms
### Description

I changed it so that the isothermal models compute the heat source terms using the fixed temperature rather than returning zero, but I forgot to add a `get_coupled_variables` method to `Isothermal` so it currently doesn't calculate anything! Whoops

### Motivation

_No response_

### Possible Implementation

_No response_

### Additional context

_No response_

rtimms:
# Description
Fixes a bug where the isothermal submodel doesn't calculate any heat source terms. There is a new option "calculate heat source for isothermal models" that is "false" by default and "true" otherwise. Default is "false" since this option may require additional parameters not needed by the electrochemical model (specifically current collector parameters).

Fixes #1957 

## Type of change

# Key checklist:

## Further checks:


brosaplanella (approved):
Looks good! I didn't review the last PR, so is there any tests that checks that the heat source terms are calculated correctly?

rtimms:
Do you mean to test if the variables have the correct expression (i.e. zero or not)? Or some numerical check on the actual values?

brosaplanella:
> Do you mean to test if the variables have the correct expression (i.e. zero or not)? Or some numerical check on the actual values?

I guess both, though just realised I don't know if we are testing this for the other variables anyway haha

rtimms:
> I guess both, though just realised I don't know if we are testing this for the other variables anyway haha

We compare outputs across different models that should be the same and have some tests on variables for things like conservation or checking that things that should be increasing are increasing etc., but we don't test for correct values. 

I'll merge this for now, but maybe we can revisit if there are any useful extra tests on variables.

PR #1959, opened 2022-03-02
I1898 make evaluator functions consistent across casadi, python and jax

### LINKED ISSUE #1898: refactor `process` function in BaseSolver to produce "raw" functions with a common signature
**Summary**
The evaluate signature for expression trees is:
`evaluate(self, t=None, y=None, y_dot=None, inputs=None, known_evals=None) -> result, known_evals`

The call signature used for calls by the solvers are:
`__call__(t, y, inputs) -> result`

in `BaseSolver`, the `process` function is used to (optionally) convert an pybamm expression tree into one of pybamm.EvaluatorJax, pybamm.EvaluatorPython or Casadi functions, and then wrap them in a `SolverCallable` so that the call signature for the solvers above are satisfied. It also generates "raw" functions (the 1st return value), which can be:
1. the `evaluate` function of a pybamm expression tree
2. the `evaluate` function of a EvaluatorJax or EvaluatorPython class
3. a casadi function

The main issue here is that the signature of 3 is the same as the call signature above, while 1 and 2 use the `evaluate` signature. This is mainly why `SolverCallable` is needed to wrap them in a common interface.

I propose that we:
1. remove the 1st "raw" option above, so that these "raw" functions are either one of the evaluator functions, or a casadi function
2. ensure that the `evaluate` function of the Evaluator classes use the call signature used by the solvers
3. remove the `SolverCallable` class entirely, and have the solvers all use the raw functions directly (which are either the evaluate methods of one of the Evaluator classes, or a casadi function)

 **Motivation**

The motivation is to:
- speed up the calls by the solvers. The `SolverCallable` needs to be generic across all formats, but the raw casadi and evaluator functions have the potential to be faster. Note that this is currently done by the casadi solver anyway, which uses raw casadi functions rather than the `SolverCallable`
- simplify the BaseSolver code: it currently needs to keep track of raw functions (generally casadi function) and `SolverCallable` functions, this way it only keeps track of the raw functions
- allow all solvers access to the raw functions. At the moment only the casadi solver uses the raw casadi functions, but other solvers might need these as well (see #1863)

martinjrobins:
# Description

- changes the `process` function in `pybamm.BaseSolver` to return "raw" evaluator functions
- all evalators now use the api expected by a solver, rather than the `evaluate` function used by an expression tree.The evaluators return different types (e.g casadi returns DMs, python returns ndarrays), but this is expected to be handled in each solver.
- this brings in other evaluator functions in line with the casadi evaluator

Fixes #1898 

## Type of change

- [x] refactor

# Key checklist:

## Further checks:


pybamm/expression_tree/operations/evaluate_python.py
```diff
@@ -491,21 +491,17 @@ def __init__(self, symbol):
         compiled_function = compile(python_str, result_var, "exec")
         exec(compiled_function)
 
-    def evaluate(self, t=None, y=None, y_dot=None, inputs=None, known_evals=None):
+    def __call__(self, t=None, y=None, inputs=None):
```
tinosulzer, line 494: why no more `y_dot`?

tinosulzer (approved):
Looks good - can you open an issue to remove `known_evals` in other places, where it is no longer needed either?

pybamm/expression_tree/operations/evaluate_python.py
martinjrobins: I removed it because I was focused on making this consistent with the casadi evaluator. I remember adding it in because I was planning for support for full dae equations rather than just semi-explicit form, but this never really went forward. Happy to add it back if we still do want to go in that direction?

tinosulzer: Ok, sticking to semi-explicit is fine by me

tinosulzer:
There are some coverage issues to fix before merging

martinjrobins:
> There are some coverage issues to fix before merging

Yup, I'll fix that up. Do you know where the Modulo pybamm op is currently being tested? looks like a branch of that code is not being covered.

tinosulzer:
https://github.com/pybamm-team/PyBaMM/blob/fb76610f4a4cbf7305d54bdbfa81eaac8377b392/tests/unit/test_solvers/test_scikits_solvers.py#L345

tests/unit/test_solvers/test_scikits_solvers.py
```diff
@@ -782,13 +782,13 @@ def test_model_step_events(self):
...
         model = pybamm.BaseModel()
-        model.timescale = pybamm.Scalar(1)
+        model.timescale_eval = pybamm.Scalar(1)
         var1 = pybamm.Variable("var1")
         var2 = pybamm.Variable("var2")
         a = 0.6
         discontinuities = (np.arange(3) + 1) * a
 
-        model.rhs = {var1: pybamm.Modulo(pybamm.t * model.timescale, a)}
+        model.rhs = {var1: pybamm.Modulo(pybamm.t * model.timescale_eval, a)}
```
tinosulzer, line 791: note that `x * 1` automatically gets simplified to `x`

pybamm/solvers/base_solver.py
```diff
@@ -397,287 +449,225 @@ def jacp(*args, **kwargs):
...
+                            )
+                        )
+                elif isinstance(symbol, pybamm.Modulo):
+                    found_t = False
+                    # Dimensionless
+                    if symbol.left.id == pybamm.t.id:
+                        expr = symbol.right
+                        found_t = True
+                    # Dimensional
+                    elif symbol.left.id == (pybamm.t * model.timescale_eval).id:
```
martinjrobins, line 543: @tinosulzer: I'm not understanding this line. In line 150-153 we ensure that model.timescale_eval is a scalar, so when you do `pybamm.t * model.timescale_eval` wont this be simplified to `pybamm.t` so this elif is the same as the if above it?

tinosulzer: `pybamm.t` is just an object that returns `t` when `.evaluate(t=t)` is called (not a constant). So when multiplied by a scalar it doesn't get simplified

tinosulzer: But I'm happy to merge this as is. Those lines that aren't being covered weren't covered before already

martinjrobins: no worries, I want to cover those lines now!

martinjrobins: think I know what I was doing wrong now.....

PR #1964, opened 2022-03-04
Remove 32 bits support

### LINKED ISSUE #1955: Look into removing `PyPy` 32-bit windows wheels
### PyBaMM Version

develeop

### Python Version

--

### Describe the bug

See https://github.com/pypa/cibuildwheel/issues/1041

We should not be building wheels for `PyPy` on `Windows` 32-bit as they are now not supported.

### Steps to Reproduce

_No response_

### Expected behaviour

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

Saransh-cpp:
# Description

Removed support for Windows 32-bits from workflows. Additionally, all the `CMakeLists.txt` files had some conditions for 32-bit windows, and I am not sure if they should be removed. The tests are passing when these lines are removed but I'll revert the deletions if they are supposed to stay.
https://github.com/pybamm-team/PyBaMM/blob/9f8f2c3fe769c15b9c4f1dca98eb6dbe07a587d1/scripts/replace-cmake/sundials-4.1.0/CMakeLists.txt#L322-L331 

Fixes #1955 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -62,7 +62,10 @@ jobs:
 
       - name: Build 64 bits wheels on Windows
         if: matrix.os == 'windows-latest'
-        run: python -m cibuildwheel --output-dir wheelhouse
+        run: |
+          choco install --yes visualstudio2019community
+          choco install --yes visualstudio2019-workload-nativedesktop
+          python -m cibuildwheel --output-dir wheelhouse
         env:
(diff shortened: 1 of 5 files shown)


tinosulzer (approved):
Thanks! I think you can delete those lines. Is there a way to raise an error if the system is 32-bit windows, saying we no longer support it?

Saransh-cpp:
Installing the latest version of `PyBaMM` using `pip` on Windows 32-bit will automatically raise an error. Older versions would work with the 32-bit system. I'm not sure if we should add a warning for them, or if we can even do so?

tinosulzer:
Yes, older versions should continue to work

PR #1969, opened 2022-03-08
Issue 1575 discharge energy

### LINKED ISSUE #1575: Add Discharged energy [W.h] as variable
**Summary**
Add Discharged energy [W.h] as variable

 **Motivation**
Like discharged capacity but using current * voltage. Useful for performance metrics and minimal overhead

tinosulzer:
# Description

Add "Discharge energy [W.h]" as an output variable (integral of power).
Set option "calculate discharge energy" to "true" to turn on. It's false by default as calculating it can be slow.

Fixes #1575 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -60,19 +60,12 @@ jobs:
           path: C:\Users\runneradmin\AppData\Local\vcpkg\archives
           key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('vcpkg.json') }}
 
-      - name: Build 32 bits wheels on Windows
+      - name: Build 64 bits wheels on Windows
         if: matrix.os == 'windows-latest'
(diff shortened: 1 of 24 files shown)


TomTranter:
@tinosulzer this is a lot of code changes just for one calculation. Did it really slow things down so much to warrant making it an option?

tinosulzer:
Here are the benchmarks with the discharge energy calculation: https://github.com/pybamm-team/PyBaMM/runs/5458023045?check_suite_focus=true

It could probably be done as a post-processing step, with low overhead, but that would be a bigger change

tinosulzer:
It's not that many changes? Most of them are just passing `options` to the external circuit submodels, which we might want to do at some point later anyway

TomTranter:
ok if you think might be useful anyway. I just prefer not having too many options

PR #1977, opened 2022-03-10
simplify model copy

tinosulzer:
# Description

Remove `model.new_empty_copy()` and simplify `model.new_copy()`.
Not sure why I made it so complicated in the first place, but this seems to work fine and be faster

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/tests/unit/test_models/test_full_battery_models/test_lithium_ion/test_basic_models.py
+++ b/tests/unit/test_models/test_full_battery_models/test_lithium_ion/test_basic_models.py
@@ -10,30 +10,18 @@ class TestBasicModels(unittest.TestCase):
         model = pybamm.lithium_ion.BasicDFN()
         model.check_well_posedness()
 
-        copy = model.new_copy()
-        copy.check_well_posedness()
-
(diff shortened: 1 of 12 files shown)


tests/unit/test_models/test_full_battery_models/test_lithium_ion/test_basic_models.py
```diff
@@ -10,31 +10,19 @@ def test_dfn_well_posed(self):
         model = pybamm.lithium_ion.BasicDFN()
         model.check_well_posedness()
 
-        copy = model.new_copy()
```
brosaplanella, line 13: What was the point of these tests, originally?

brosaplanella (approved):
Looks great, thanks Tino!

tests/unit/test_models/test_full_battery_models/test_lithium_ion/test_basic_models.py
tinosulzer: just coverage

PR #1979, opened 2022-03-12
Updating the notebook in parameter management

### LINKED ISSUE #1697: Expose parameter management function
The parameter [command line interface](https://pybamm.readthedocs.io/en/latest/source/parameters_cli.html?highlight=cli#pybamm.parameters_cli.add_parameter) requires users to use bash. However, there are some python functions defined in `parameters_cli.py` that we could expose so that e.g. instead of 
```
%%bash
pybamm_edit_parameter lithium_ion
```
you can just do 
```python3
pybamm.edit_parameter(["lithium_ion"])
```
directly in python.

soma2000-lang:
# Description

 instead of
%%bash
pybamm_edit_parameter lithium_ion

pybamm.edit_parameter(["lithium_ion"])
Fixes  #1697

# Key checklist:

### DIFF
--- a/examples/notebooks/parameterization/parameter-management.ipynb
+++ b/examples/notebooks/parameterization/parameter-management.ipynb
@@ -54,6 +54,13 @@
    "metadata": {},
    "outputs": [
     {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "ERROR: Invalid requirement: '#'\n"
+     ]
+    },
+    {
      "name": "stdout",
      "output_type": "stream",
      "text": [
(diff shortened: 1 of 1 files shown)


soma2000-lang:
@rtimms  Please look into this pr

Saransh-cpp (changes requested):
@soma2000-lang, the issue requires converting the `CLI` to user-facing functions (setting them up in a way that they can be used both through `CLI` and `Python` script). For instance, a user right now can do -
```
pybamm_edit_parameter lithium_ion
```
from the terminal to create a new folder named `lithium_ion`, which would have all the default parameters.

We want to make this more flexible, such that the user can do the above task from their `Python` script -
```py
import pybamm

# create lithium_ion directory
pybamm.edit_parameter(["lithium_ion"])
```

This would require editing the `CLI` code present in [parameters_cli.py](https://github.com/pybamm-team/PyBaMM/blob/develop/pybamm/parameters_cli.py) file. You will have to update the functions first, or if you just want to update the notebooks, you can wait for someone else to make these changes! :)

soma2000-lang:
Thank you for the clarification.I would like to work changing the cli
paramaters myself first.

soma2000-lang:
@Saransh-cpp @tinosulzer I need to change the edit parameters portion right??

Saransh-cpp:
```
pybamm.parameters_cli:edit_parameter
pybamm.parameters_cli:add_parameter
pybamm.parameters_cli:remove_parameter
```

These are the primary functions that would require changes

PR #1990, opened 2022-03-21
Add Jax solver example for SPMe model

wigging:
# Description

There are no examples of using the Jax solver in the PyBaMM repository or in the documentation. This request adds an example script that uses the Jax solver with the SPMe model.

## Type of change

# Key checklist:

## Further checks:


tinosulzer:
Looks like just a dependency issue. More worryingly, it's now taking an hour to build linux, but that doesn't look like something you changed.

I'm not sure what this example really adds? There's nothing special about using `JaxSolver` instead of something else?

wigging:
Closing this because request because no one is interested in a Jax example.

PR #1995, opened 2022-03-24
automate generate debug files

### LINKED ISSUE #1690: Automatically generate debugging file
When running long experiments, it can be hard to debug if the simulation fails a long way in.

If the simulation fails, automatically generate a debugging file with all the information necessary to recreate the failing simulation in just one cycle:
- model including options
- experiment
- parameter values
- discretization
- solver
- starting solution

Could be saved to `pybamm.settings.debug_filename` which defaults to `debug.txt` (and add `debug.txt` to the gitignore)

arunavabasu-03:
### Description
Fixes #1690 
Hey @rtimms @tinosulzer, I implement the `__repr__` function please check this approach, give me some feedback.
so, I implement other classes.

### DIFF
--- a/tests/integration/test_models/test_full_battery_models/test_lithium_ion/test_external/test_external_temperature.py
+++ b/tests/integration/test_models/test_full_battery_models/test_lithium_ion/test_external/test_external_temperature.py
@@ -7,6 +7,9 @@ import numpy as np
 
 
 class TestExternalThermalModels(unittest.TestCase):
+    def __repr__(self):
+        return super().__repr__()
+       
     def test_external_lumped_temperature(self):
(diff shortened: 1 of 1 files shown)


tinosulzer:
The `__repr__` function needs to be added to the `BaseModel` class, not the test class. As a first step, make it return the model class as a string (e.g. `"pybamm.lithium_ion.DFN()"`). Then we will need to make it return which non-default options and keyword arguments were passed in. See `Symbol.__repr__` for an example of a repr class

arunavabasu-03:
Oh! that's my fault , thanks @tinosulzer

tinosulzer:
Closing, reopen if you're still working on it

PR #2022, opened 2022-04-19
#2021 remove deprecation errors

### LINKED ISSUE #2021: Deprecation message for `reaction rate`
At the moment, if the parameter dictionary contains a parameter containing `reaction rate` it throws an error as this was deprecated. However, if you manually introduce this variable again in the definition (to use it as an input, for example) the error comes up again even if the code works fine. Given that this was introduced about two years ago, it can probably be safely removed. However, this is not the only deprecation error so might be worth reconsidering them all.

The deprecation message is in [l374 of parameter_values.py](https://github.com/pybamm-team/PyBaMM/blame/develop/pybamm/parameters/parameter_values.py).

brosaplanella:
# Description

Removed all the deprecation errors that I could find that were one year or older. Not sure what's the best way to automate removing deprecation errors in the future. Maybe we can just add a comment when we introduce one of the errors with the key work "deprecation" (or anything starting with deprec) and the date, so it is easy to search them and check how old they are.

Fixes #2021 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/expression_tree/symbol.py
+++ b/pybamm/expression_tree/symbol.py
@@ -878,10 +878,6 @@ class Symbol:
         """
         return any(isinstance(symbol, symbol_classes) for symbol in self.pre_order())
 
-    def simplify(self, simplified_symbols=None, clear_domains=True):
-        """`simplify()` has now been removed."""
-        raise pybamm.ModelError("simplify is deprecated as it now has no effect")
-
     def to_casadi(self, t=None, y=None, y_dot=None, inputs=None, casadi_symbols=None):
         """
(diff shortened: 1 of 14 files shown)


tinosulzer (approved):
Thanks, looks good. We could start using https://pypi.org/project/deprecation/#description, which specifies by which version a warning should be removed and fails tests if not removed by then

brosaplanella:
> Thanks, looks good. We could start using https://pypi.org/project/deprecation/#description, which specifies by which version a warning should be removed and fails tests if not removed by then

Shall I replace some of the remaining deprecation errors using the package or should I merge the PR now?

PR #2033, opened 2022-04-23
Update using-model-options_thermal-example.ipynb

tinosulzer:
# Description

Fixes broken link (I think?)


brosaplanella (approved):
The link works for me. Weirdly enough, the links that fail the test work for me when I click on them so not sure where the issue is.

tinosulzer:
yeah it's strange

PR #2035, opened 2022-04-24
creating output instead of throwing solverError

### LINKED ISSUE #1992: Creating output instead of throwing SolverError
### Description

The feature consist of small changes in casadi_solver.py which involve creating output from a simuation if the maximum number of decreased steps is reached. A warning will be created instead of an error and the simulation will terminate normally.

### Motivation

Especially when setting up a parameter study and simulations being solved for a lot of timesteps it always bugged me when i got the pybamm.SolverError `For more timesteps try solving the model up to this time only or reducing dt_max (currently, dt_max={}).`
This meant that thousands of timesteps would have to be simulated again because no output was created. The issue is even more severe when you try to do a large parallelised parameter study where one terminated simulation could terminate a whole branch of simulations which were meant to be solved in sequence.

### Possible Implementation

the basic idea involves changes in the casadi_solver. This is merely an example of how to handle the error and could be extended to other solvers. I attached a screenshot where you can see which file i added as well as next to the line numbers where i changed something in the code.
casadi_solver_changes

### Additional context

This issue is related to issue #1986 . After i removed the "electrolyte zero-cut off" in the DFN as mentioned in #1982 two things happened:

- for high currents: when before the simulation terminated due to the electrolyte zero cut-off, it now terminated due to a maximum/minimum voltage event. So everythings fine. Physically speaking the voltage loss in the electrolyte was high enough to meet the final cut-off voltage.
-  for low currents: the simulations has a problem to reach the maximum/minimum voltage altogether. Even when increasing the max_step_decrease_count in the Casadi solver, the solver came to a point where it would do timestep sizes of miliseconds until it finally broke but saved no output. For this case i came up with the idea to change the pybamm.SolverError to a warning and create the output thas was acquired until the respective timestep.

dion-w:
# Description

Fixes #1992 

## Type of change

# Key checklist:

## Further checks:


dion-w:
checks are failing at  `test_casadi_solver.TestCasadiSolver.test_model_solver_failure` because the tests that expect a SolverError are now getting a SolverWarning. Also the tests all fail in the first timestep without creating a solution.  This wasn't a problem since a SolverError was raised and the test terminated as expected. Now though the following functions can't run, since the solution object was only created as "None" and therefore has no properties. 

Now a test is needed that creates a solution in the first timestep but runs into a problem in a later timestep, so that there already is a solution object that can be processed by the follow up functions. @tinosulzer could you help out with creating such a test?

tinosulzer:
See extra comments added with `####`
```python
def test_model_solver_failure(self):
        # Create model
        model = pybamm.BaseModel()
        var = pybamm.Variable("var")
        model.rhs = {var: -pybamm.sqrt(var)}
        model.initial_conditions = {var: 1}
        # add events so that safe mode is used (won't be triggered)
        model.events = [pybamm.Event("10", var - 10)]
        # No need to set parameters; can use base discretisation (no spatial operators)

        # create discretisation
        disc = pybamm.Discretisation()
        model_disc = disc.process_model(model, inplace=False)

        solver = pybamm.CasadiSolver(
            extra_options_call={"regularity_check": False}, dt_max=1e-3
        )
        #### By reducing dt_max, this one should pass. You need to add a check that the final solution 
        #### does indeed stop before t=20
        # Solve with failure at t=2
        t_eval = np.linspace(0, 20, 100)
        with self.assertWarns(pybamm.SolverWarning):
            solver.solve(model_disc, t_eval)
        # Solve with failure at t=0
        model.initial_conditions = {var: 0}
        model_disc = disc.process_model(model, inplace=False)
        t_eval = np.linspace(0, 20, 100)
        #### This one should fail immediately and throw a `SolverError`
        #### since no progress can be made from the first timestep
        with self.assertWarns(pybamm.SolverWarning):
            solver.solve(model_disc, t_eval)
```

tinosulzer:
Continued in #2350

PR #2051, opened 2022-05-04
Update parameters

DrSOKane:
# Description

Updated some parameters:
-Negative electrode OCP functions for Chen2020, Chen2020_plating and O'Regan2021 parameter sets now have an addiional term to capture the behaviour at high stoichiometry
-Chen2020_plating no longer uses a lookup table, so is now much faster
-Changed one parameter in the Ecker2015 positive electrode OCP function to better fit the data (R^2 is now 0.999)
-Corrected exchange current density in both electrodes for Ecker2015

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:


pybamm/input/parameters/lithium_ion/negative_electrodes/graphite_Chen2020/graphite_LGM50_ocp_Chen2020.py
```diff
@@ -26,10 +27,11 @@ def graphite_LGM50_ocp_Chen2020(sto):
 
     u_eq = (
         1.9793 * exp(-39.3631 * sto)
-        + 0.2482
+        + 0.15561
```
tinosulzer, line 30: We need to be careful that this is still representative of the results in the paper, tagging @brosaplanella

pybamm/input/parameters/lithium_ion/negative_electrodes/graphite_Ecker2015/graphite_electrolyte_exchange_current_density_Ecker2015.py
```diff
@@ -33,7 +33,7 @@ def graphite_electrolyte_exchange_current_density_Ecker2015(c_e, c_s_surf, T):
         Exchange-current density [A.m-2]
     """
 
-    k_ref = 1.995 * 1e-10
+    k_ref = 1.11 * 1e-10
```
tinosulzer, line 36: is this explicitly in the paper, or calculated by you?

DrSOKane: Calculated by me. What's in the paper is the exchange curent density at 50% lithiation.

The reason the value is out is it was changed to 1.995E-10 to account for not all of the solid phase being active. But when the provision for inactive material was added, the rate constants weren't changed back. @rtimms argued the model isn't sensitive enough to this parameter to justify changing, and for a discharge curve it's not, but we have GITT results for this cell that are massively improved by changing the rate constants.

DrSOKane:
I merged these changes into the branch used for partially reversible plating and for some reason this fixes the specific error that prevents this pull request being merged. I'm therefore closing this one.

pybamm/input/parameters/lithium_ion/negative_electrodes/graphite_Ecker2015/graphite_electrolyte_exchange_current_density_Ecker2015.py
rtimms: Not sure I remember this. Would be keen to see the GITT results if you’re able to share. Even though changes like this make a better fit I do worry a bit about making them in this way. People may rightly expect the params from the Ecker2015 set to be exactly those in the paper, not a tweaked version. This is something we should bear in mind. 

I am aware I added thermal params to this set from a different source, so I’m guilty of making additions/tweaks too.

DrSOKane: I'm not tweaking though. The rate constant can be calculated directly from the exchange current density at any given lithiation, and that is included in the paper.

rtimms: But doesn’t this change the fit given in the paper (fig 11 of part 1)? Or am I misunderstanding something?

PR #2053, opened 2022-05-05
#2027 refactor submodels for more flexible domains

### LINKED ISSUE #2027: More flexible domains
We currently have "full cell" and "half-cell" (the latter being hardcoded separator - positive electrode). Look into doing this more flexibly so we can have symmetric cell (separator only), dual electrodes, dead lithium layers, etc

tinosulzer:
# Description

Refactors submodels so that domains of variables defined over multiple domains are more flexible (previously, only cases ["negative electrode", "separator", "positive electrode"] and ["separator", "positive electrode"] were possible).

This PR does not introduce any new models (e.g. negative electrode half-cell or symmetric cell), but adding those should be easy once these changes have been made

Fixes #2027 

## Type of change

# Key checklist:

## Further checks:


tinosulzer:
How does this look? It reduces the amount of code, but is `f"{Dom} porosity times electrolyte concentration"` too hard to read?
We should also add a way to see which submodel added which variables

brosaplanella:
I like it. Thinking of more general models (beyond batteries) I think that attaching a variable to a domain in this generic way is more useful (and similarly for the parameters). For example, if you have a heat transfer with phase change problem, you will again have all the parameters & variables twice (one for each domain).

pybamm/models/submodels/electrolyte_conductivity/surface_potential_form/full_surface_form_conductivity.py
```diff
@@ -136,26 +130,30 @@ def _get_conductivities(self, variables):
         return conductivity, sigma_eff
 
     def set_initial_conditions(self, variables):
-        if self.domain == "Separator":
+        Domain = self.domain.capitalize()
+
+        if self.domain == "separator":
             return
```
brosaplanella, line 136: Is this equivalent to `return None` (used below)? If so, should we keep it consistent (even if the output is the same)?

pybamm/models/submodels/transport_efficiency/base_transport_efficiency.py
```diff
@@ -23,38 +23,26 @@ def __init__(self, param, component, options=None):
...
-                    f"Negative {self.component.lower()} transport efficiency": tor_n,
-                    f"X-averaged negative {self.component.lower()} "
-                    "transport efficiency": pybamm.x_average(tor_n),
-                }
-            )
+        variables = {f"{self.component} transport efficiency": tor}
+
+        for domain, tor_k in tor_dict.items():
+            domain = domain.split()[0]
+            Domain = domain.capitalize()
```
brosaplanella, line 35: Can't you use the `domain_Domain` here?

brosaplanella (approved):
Looks amazing, thanks @tinosulzer! Just a couple of very minor questions.

pybamm/models/submodels/base_submodel.py
```diff
@@ -129,33 +116,32 @@ def domain(self):
...
+            "negative",
+            "separator",
+            "positive",
+            "negative electrode",
             "Negative electrolyte",
             "Separator electrolyte",
-            "Positive electrode",
+            "positive electrode",
             "Positive electrolyte",
             None,
```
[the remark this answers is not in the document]
tinosulzer: I think we can get rid of "Positive electrolyte" etc

pybamm/models/submodels/electrolyte_conductivity/surface_potential_form/full_surface_form_conductivity.py
tinosulzer: I looked it up and there is a stylistic difference. I will try to keep it consistent https://stackoverflow.com/a/15300671/10816761

pybamm/models/submodels/transport_efficiency/base_transport_efficiency.py
tinosulzer: In this case we want `domain = "negative"` and `Domain = "Negative"`. `domain_Domain` would return `"negative electrode", "Negative electrode"` and then we would need to split them both.

PR #2057, opened 2022-05-11
Merge fix for wheels

brosaplanella:
Merge fix for wheels to main in order to complete the release


martinjrobins (approved):
happy to merge this, but isn't the release for this month already on pypi? should we wait until next month to merge (or we can just do a 22.4.post1 version I guess)?

priyanshuone6:
On pypi we just have a sdist file and no wheels. That too is outdated. We can still publish wheels in the same pypi release.

I think we should yank that release on pypi as the sdist file is already published, and we can not delete or replace it. I'm not sure, but it might cause issues while compiling the conda release as a few things were fixed in cmakebuild.py after the pypi sdist release.

Can we skip v22.4 and release v22.5 instead?

martinjrobins:
> Can we skip v22.4 and release v22.5 instead?

Yea, I think we should just wait until june and do v22.5

brosaplanella:
Closing this PR as we will merge the fix when releasing 22.5

Issue #1695, opened 2021-09-28
Temperature dependence in SEI models


tinosulzer:
- For the EC reaction limited, reaction limited, and interstitial diffusion limited, multiply the exponent by the following factor (dimensionless 1/T):
```python
T = variables["Negative electrode temperature"]
prefactor = 1 / (1 + self.param.Theta * T)
```
- Add arrhenius dependency to diffusivity in ec reaction limited and solvent diffusion limited (I don't know what the arrhenius factor should be)
- Arrhenius dependence for the conductivity of the SEI in electron migration limited

soma2000-lang:
Working on this issue@tinosulzer

soma2000-lang:
@tinosulzer  I will be submitting a proposal for the idea "Documentation" ,for GSOC22 under pybamm.
I have a doubt about how to know what the arrhenius dependency should be.
It would be great to be mentored by you and  other mentors

soma2000-lang:
@tinosulzer Can you please tell me where SEI model file is, I am having difficulty in finding them.

tinosulzer:
https://github.com/pybamm-team/PyBaMM/blob/develop/pybamm/models/submodels/interface/sei/sei_growth.py

soma2000-lang:
@tinosulzer  thanks

PR #2080, opened 2022-05-28
add temperature dependense in SEI model

Vaibhav-Chopra-GT:
# Description
For issue #1695 
Multiplied the exponent in EC reaction limited, reaction limited, and interstitial diffusion-limited in SEI models with 
`1 / (1 + self.param.Theta * T)`, where `T = variables["Negative electrode temperature"]`

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/models/submodels/interface/sei/sei_growth.py
+++ b/pybamm/models/submodels/interface/sei/sei_growth.py
@@ -84,9 +84,15 @@ class SEIGrowth(BaseModel):
         if self.options["SEI"] == "reaction limited":
             # alpha = param.alpha
             C_sei = param.C_sei_reaction
+            T = variables["Negative electrode temperature"]
 
             # need to revise for thermal case
-            j_sei = -(1 / C_sei) * pybamm.exp(-0.5 * (delta_phi - j * L_sei * R_sei))
+            j_sei = (
+                -(1 / C_sei)
+                * pybamm.exp(-0.5 * (delta_phi - j * L_sei * R_sei))
+                * 1
+                / (1 + self.param.Theta * T)
+            )
 
         elif self.options["SEI"] == "electron-migration limited":
             U_inner = self.param.U_inner_electron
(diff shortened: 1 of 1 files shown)


tinosulzer (changes requested):
Thanks @Vaibhav-Chopra-GT , see comments

pybamm/models/submodels/interface/sei/sei_growth.py
```diff
@@ -84,9 +84,15 @@ def get_coupled_variables(self, variables):
         if self.options["SEI"] == "reaction limited":
             # alpha = param.alpha
             C_sei = param.C_sei_reaction
+            T = variables["Negative electrode temperature"]
 
             # need to revise for thermal case
-            j_sei = -(1 / C_sei) * pybamm.exp(-0.5 * (delta_phi - j * L_sei * R_sei))
+            j_sei = (
+                -(1 / C_sei)
+                * pybamm.exp(-0.5 * (delta_phi - j * L_sei * R_sei))
```
brosaplanella, line 92: Speaking out of memory, should a similar factor be in the exponential to account for the temperature in the `R T / F`?

tinosulzer: Oh yeah, it should only be inside the exponential right?

brosaplanella: I am not sure about the outside as I don't remember the details of the model, but certainly inside the exponential. Are these from Scott's thesis?

tinosulzer: The ones outside could be specified by making SEI reaction rate a function of temperature (e.g. arrhenius) I think?

brosaplanella: Actually I don't know where do the ones outside come from exactly...

tinosulzer:
will be fixed by #2104

PR #2085, opened 2022-05-31
Merge sdist workflow into wheels

### LINKED ISSUE #2084: sdist unintentional workflow triggers
The sdist workflow is sometimes triggering without merging PRs to main, which is creating a new release on pypi

https://github.com/pybamm-team/PyBaMM/actions/runs/2410208047    (v22.5)
https://github.com/pybamm-team/PyBaMM/actions/runs/2268973466    (v22.4) (ran even after build_wheels failed)

Thinking of merging sdist into the wheels workflow

priyanshuone6:
# Description

Merge the sdist workflow into the wheels workflow

Fixes #2084 

Test run:
https://github.com/priyanshuone6/PyBaMM/actions/runs/2412252029


tinosulzer (approved):
thanks, should this be merged?

priyanshuone6:
Having two workflows is complicating the release process. The sdist workflow is dependent on the wheels workflow, and for some reason, it triggers even before the wheels workflow, which creates a new release on pypi (even when wheels fail to build).

Now with one workflow, if wheels fail, the entire workflow will fail. The sdist process will not interfere with wheels because it will be created in a new environment in a different job and then uploaded with the wheels at the same time.

PR #2086, opened 2022-05-31
Benchmarks for model X parameter set combinations

Vaibhav-Chopra-GT:
# Description

Added benchmarks for `SPM`, `SPMe`, and `DFN` with all the parameter sets (except Sulzer2019 and Xu2019).

Some concerns -
- I could not solve the `DFN` model with `Mohtat2020` parameter sets for the benchmarks (commented code). I tried varying `dt_max` and used different solvers but nothing worked. Here is a reproducible example for the same-
```python
import numpy as np
import pybamm

def prepare_model_Mohtat2020(model):

    param = pybamm.ParameterValues("Mohtat2020")
    param.process_model(model)

    var_pts = {"x_n": 20, "x_s": 20, "x_p": 20, "r_n": 30, "r_p": 30, "y": 10, "z": 10}
    mesh = pybamm.Mesh(geometry, model.default_submesh_types, var_pts)

    disc.process_model(model)

solver = pybamm.CasadiSolver()
c_rate = 1
tmax = 4000 / c_rate
nb_points = 500
t_eval = np.linspace(0, tmax, nb_points)
prepare_model_Mohtat2020(model)
solver.solve(model, t_eval=t_eval)
solution = solver.solve(model, t_eval=t_eval)
voltage = solution['Terminal voltage [V]']
c_s_n_surf = solution['Negative particle surface concentration']
c_s_p_surf = solution['Positive particle surface concentration']
t = solution["Time [s]"].entries
x = solution["x [m]"].entries[:, 0]
f, (ax1, ax2, ax3) = plt.subplots(1, 3, figsize=(13,4))

ax1.plot(t, voltage(t))
ax1.set_xlabel(r'$Time [s]$')
ax1.set_ylabel('Terminal voltage [V]')

ax2.plot(t, c_s_n_surf(t=t, x=x[0]))  
ax2.set_xlabel(r'$Time [s]$')
ax2.set_ylabel('Negative particle surface concentration')

ax3.plot(t, c_s_p_surf(t=t, x=x[-1])) 
ax3.set_xlabel(r'$Time [s]$')
ax3.set_ylabel('Positive particle surface concentration')

plt.tight_layout()
plt.show()
```
The error-
```
The linesearch algorithm failed with too small a step.
At t = 3.7305e-005 and h = 3.3406e-019, the corrector convergence failed repeatedly or with |h| = hmin.
At t = 3.73046e-005 and h = 3.12482e-019, the corrector convergence failed repeatedly or with |h| = hmin.
At t = 3.73044e-005 and h = 3.16429e-020, the corrector convergence failed repeatedly or with |h| = hmin.
At t = 3.7305e-005 and h = 3.3406e-019, the corrector convergence failed repeatedly or with |h| = hmin.
  File "d:\LearningProjects\python\second.py", line 25, in <module>
    solver.solve(model, t_eval=t_eval)
  File "C:\Users\Vaibhav Chopra\AppData\Local\Programs\Python\Python39\lib\site-packages\pybamm\solvers\base_solver.py", line 1018, in solve
  File "C:\Users\Vaibhav Chopra\AppData\Local\Programs\Python\Python39\lib\site-packages\pybamm\solvers\casadi_solver.py", line 263, in _integrate
    raise pybamm.SolverError(
pybamm.expression_tree.exceptions.SolverError: Maximum number of decreased steps occurred at t=0.0. Try solving the model up to this time only or reducing dt_max (currently, dt_max=5.010622778599846).
PS D:\LearningProjects> python -u "d:\LearningProjects\python\second.py"
The linesearch algorithm failed with too small a step.
At t = 3.7305e-005 and h = 3.3406e-019, the corrector convergence failed repeatedly or with |h| = hmin.
At t = 3.73046e-005 and h = 3.12482e-019, the corrector convergence failed repeatedly or with |h| = hmin.
At t = 3.73044e-005 and h = 3.16429e-020, the corrector convergence failed repeatedly or with |h| = hmin.
At t = 3.7305e-005 and h = 3.3406e-019, the corrector convergence failed repeatedly or with |h| = hmin.
  File "d:\LearningProjects\python\second.py", line 25, in <module>
    solver.solve(model, t_eval=t_eval)
  File "C:\Users\Vaibhav Chopra\AppData\Local\Programs\Python\Python39\lib\site-packages\pybamm\solvers\base_solver.py", line 1018, in solve
  File "C:\Users\Vaibhav Chopra\AppData\Local\Programs\Python\Python39\lib\site-packages\pybamm\solvers\casadi_solver.py", line 263, in _integrate
    raise pybamm.SolverError(
pybamm.expression_tree.exceptions.SolverError: Maximum number of decreased steps occurred at t=0.0. Try solving the model up to this time only or reducing dt_max (currently, dt_max=5.010622778599846).
```
- Should I group these benchmarks in some way. For example- multiple files with each file containing benchmarks of a particular parameter or model.
- As discussed with @brosaplanella on slack I have used `ScikitsDaeSolver` for some benchmarks, will this create a problem or is it safe to use different solvers for similar benchmarks?

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/benchmarks/time_setup_models_and_sims.py
+++ b/benchmarks/time_setup_models_and_sims.py
@@ -26,6 +26,7 @@ class TimeBuildSPMMarquis2019:
         self.param.process_model(self.model)
         compute_discretisation(self.model, self.param).process_model(self.model)
 
+
 class TimeBuildSPMORegan2019:
     def __init__(self):
         self.param = pybamm.ParameterValues("ORegan2021")
@@ -34,6 +35,7 @@ class TimeBuildSPMORegan2019:
         self.model = pybamm.lithium_ion.SPM()
         self.param.process_model(self.model)
         compute_discretisation(self.model, self.param).process_model(self.model)
+
 
 class TimeBuildSPMNCA_Kim2011:
     def __init__(self):
@@ -44,6 +46,7 @@ class TimeBuildSPMNCA_Kim2011:
         self.param.process_model(self.model)
         compute_discretisation(self.model, self.param).process_model(self.model)
 
+
 class TimeBuildSPMPrada2013:
     def __init__(self):
         self.param = pybamm.ParameterValues("Prada2013")
@@ -52,6 +55,7 @@ class TimeBuildSPMPrada2013:
         self.model = pybamm.lithium_ion.SPM()
         self.param.process_model(self.model)
         compute_discretisation(self.model, self.param).process_model(self.model)
+
 
 class TimeBuildSPMAi2020:
     def __init__(self):
@@ -62,6 +66,7 @@ class TimeBuildSPMAi2020:
         self.param.process_model(self.model)
         compute_discretisation(self.model, self.param).process_model(self.model)
 
+
 class TimeBuildSPMRamadass2004:
     def __init__(self):
         self.param = pybamm.ParameterValues("Ramadass2004")
@@ -70,6 +75,7 @@ class TimeBuildSPMRamadass2004:
         self.model = pybamm.lithium_ion.SPM()
         self.param.process_model(self.model)
         compute_discretisation(self.model, self.param).process_model(self.model)
+
 
 class TimeBuildSPMMohtat2020:
     def __init__(self):
@@ -80,6 +86,7 @@ class TimeBuildSPMMohtat2020:
         self.param.process_model(self.model)
         compute_discretisation(self.model, self.param).process_model(self.model)
 
+
 class TimeBuildSPMChen2020:
     def __init__(self):
         self.param = pybamm.ParameterValues("Chen2020")
@@ -89,6 +96,7 @@ class TimeBuildSPMChen2020:
         self.param.process_model(self.model)
         compute_discretisation(self.model, self.param).process_model(self.model)
 
+
 class TimeBuildSPMChen2020_plating:
     def __init__(self):
         self.param = pybamm.ParameterValues("Chen2020_plating")
@@ -97,6 +105,7 @@ class TimeBuildSPMChen2020_plating:
         self.model = pybamm.lithium_ion.SPM()
         self.param.process_model(self.model)
         compute_discretisation(self.model, self.param).process_model(self.model)
+
 
 class TimeBuildSPMEcker2015:
     def __init__(self):
@@ -117,6 +126,7 @@ class TimeBuildSPMeMarquis2019:
         self.param.process_model(self.model)
         compute_discretisation(self.model, self.param).process_model(self.model)
 
+
 class TimeBuildSPMeORegan2021:
     def __init__(self):
         self.param = pybamm.ParameterValues("ORegan2021")
@@ -125,6 +135,7 @@ class TimeBuildSPMeORegan2021:
         self.model = pybamm.lithium_ion.SPMe()
         self.param.process_model(self.model)
         compute_discretisation(self.model, self.param).process_model(self.model)
+
 
 class TimeBuildSPMeNCA_Kim2011:
     def __init__(self):
@@ -135,6 +146,7 @@ class TimeBuildSPMeNCA_Kim2011:
         self.param.process_model(self.model)
         compute_discretisation(self.model, self.param).process_model(self.model)
 
+
 class TimeBuildSPMePrada2013:
     def __init__(self):
         self.param = pybamm.ParameterValues("Prada2013")
@@ -143,6 +155,7 @@ class TimeBuildSPMePrada2013:
         self.model = pybamm.lithium_ion.SPMe()
         self.param.process_model(self.model)
         compute_discretisation(self.model, self.param).process_model(self.model)
+
 
 class TimeBuildSPMeAi2020:
     def __init__(self):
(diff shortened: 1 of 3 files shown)


benchmarks/time_solve_models.py
```diff
@@ -22,11 +22,328 @@ def prepare_model(model):
     disc.process_model(model)
 
 
+def prepare_model_ORegan2021(model):
+    geometry = model.default_geometry
+
+    # load parameter values and process model and geometry
+    param = pybamm.ParameterValues("ORegan2021")
```
brosaplanella, line 29: If you use parameters (as in my previous comment) you can have a single `prepare_model(model, parameters)` function.

benchmarks/time_solve_models.py
```diff
@@ -22,11 +22,328 @@ def prepare_model(model):
...
+    disc = pybamm.Discretisation(mesh, model.default_spatial_methods)
+    disc.process_model(model)
+
+
 def solve_model_once(model, solver, t_eval):
     solver.solve(model, t_eval=t_eval)
 
 
-class TimeSolveSPM:
+class TimeSolveSPMMarquis2019:
```
brosaplanella, line 182: Same as before, but need to be careful as you will have two `parameter` sets for the tests: the battery parameters and the solve (True, False). In the [docs](https://asv.readthedocs.io/en/stable/writing_benchmarks.html#parameterized-benchmarks) there is an example on how to do it.

benchmarks/time_solve_models.py
```diff
@@ -22,11 +22,328 @@ def prepare_model(model):
...
+        c_rate = 1
+        tmax = 4000 / c_rate
+        nb_points = 500
+        self.t_eval = np.linspace(0, tmax, nb_points)
+        prepare_model_Ramadass2004(self.model)
+        if solve_first:
+            solve_model_once(self.model, TimeSolveSPMRamadass2004.solver, self.t_eval)
+
+    def time_solve_model(self, solve_first):
+        TimeSolveSPMRamadass2004.solver.solve(self.model, t_eval=self.t_eval)
```
brosaplanella, line 287: This applies everywhere (but just spotted it here). You can instead type `self` to refer to the own class, so it is more compact.

benchmarks/time_solve_models.py
```diff
@@ -36,15 +353,181 @@ def setup(self, solve_first):
...
+        if solve_first:
+            solve_model_once(self.model, TimeSolveSPMePrada2013.solver, self.t_eval)
+
+    def time_solve_model(self, solve_first):
+        TimeSolveSPMePrada2013.solver.solve(self.model, t_eval=self.t_eval)
+
+
+class TimeSolveSPMeAi2020:
+    params = [True, False]
+    solver = pybamm.ScikitsDaeSolver()
```
brosaplanella, line 438: Sorry, I didn't explain myself very well on Slack. In the benchmarks we should use the same solver (with same settings, like dt_max) for all tests, so results are comparable. I suggested trying Scikits to check if the model solved fine. If it does, then you can go and fiddle with CasADI. We still don't know why this happens, but sometimes changing the tmax (like to `3800 / c_rate` or `4200 / c_rate`) makes the solver work. If you can't get it to work straight away, just comment the benchmark out and we can skip it for the time being.

brosaplanella (changes requested):
Hi Vaibhav! It looks pretty good, but there are some changes I suggested (mostly to make use of parameters in benchmarks so you can simplify your code a lot). 

I couldn't attend last meeting, so if any of my comments contradicts anything you agreed previously do let me know (and apologies in advanced).

pybamm/models/submodels/interface/sei/sei_growth.py
```diff
@@ -84,9 +84,15 @@ def get_coupled_variables(self, variables):
         if self.options["SEI"] == "reaction limited":
             # alpha = param.alpha
             C_sei = param.C_sei_reaction
+            T = variables["Negative electrode temperature"]
```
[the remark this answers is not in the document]
Vaibhav-Chopra-GT: Yes, done!

Vaibhav-Chopra-GT:
Thank you for the review @tinosulzer, @brosaplanella! I have parameterized the benchmarks and I am still looking into the solvers. Could you please review it again and give your feedback. Also, keeping the voltage same for every parameter did not throw any error, should I keep it the same or change it to match every parameter's lower voltage.

brosaplanella (approved):
Nice, it looks much cleaner now!

brosaplanella:
> Thank you for the review @tinosulzer, @brosaplanella! I have parameterized the benchmarks and I am still looking into the solvers. Could you please review it again and give your feedback. Also, keeping the voltage same for every parameter did not throw any error, should I keep it the same or change it to match every parameter's lower voltage.

For the voltage, it would be better to test down to the right voltage specified in the parameters so we ensure we have a full discharge. You can just get the value for `Lower voltage cut-off [V]' and format the string of the experiment.

benchmarks/time_setup_models_and_sims.py
```diff
@@ -89,12 +149,24 @@ def time_setup_SPMe_simulation(self, with_experiment):
...
+            "ORegan2021",
+            "NCA_Kim2011",
+            "Prada2013",
+            "Ai2020",
+            "Ramadass2004",
+            "Mohtat2020",
+            "Chen2020",
+            "Chen2020_plating",
+            "Ecker2015",
+        ],
```
priyanshuone6, line 165: Could you make a global variable with a list of these parameter sets outside the classes and use that in all the functions to reduce code duplication

tinosulzer (approved):
Thanks, looks good!

priyanshuone6:
Let’s merge this

PR #2092, opened 2022-06-07
add unit benchmarks

Vaibhav-Chopra-GT:
# Description

Created unit benchmarks (create an expression, parameterise a model, discretise a model, and solve a model) using this [example](https://github.com/Vaibhav-Chopra-GT/PyBaMM/blob/develop/examples/notebooks/Creating%20Models/3-negative-particle-problem.ipynb).

I tried developing these benchmarks using classes and inheritance, but `asv` runs `time_*` functions of the parent class in the child classes too. This results in benchmarks running multiple times.

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/benchmarks/unit_benchmarks.py
+++ b/benchmarks/unit_benchmarks.py
@@ -1,13 +1,11 @@
 import pybamm
 import numpy as np
 
-R = 1e-5
-
 
 def time_create_expression():
     global model
     model = pybamm.BaseModel()
-
+    global R
     R = pybamm.Parameter("Particle radius [m]")
     D = pybamm.Parameter("Diffusion coefficient [m2.s-1]")
     j = pybamm.Parameter("Interfacial current density [A.m-2]")
--- a/benchmarks/unit_benchmarks.py
+++ b/benchmarks/unit_benchmarks.py
@@ -1,0 +1,98 @@
+import pybamm
+import numpy as np
+
+R = 1e-5
+
+
+def time_create_expression():
+    global model
+    model = pybamm.BaseModel()
+
+    R = pybamm.Parameter("Particle radius [m]")
+    D = pybamm.Parameter("Diffusion coefficient [m2.s-1]")
+    j = pybamm.Parameter("Interfacial current density [A.m-2]")
+    F = pybamm.Parameter("Faraday constant [C.mol-1]")
+    c0 = pybamm.Parameter("Initial concentration [mol.m-3]")
+
+    c = pybamm.Variable("Concentration [mol.m-3]", domain="negative particle")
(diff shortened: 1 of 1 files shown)


benchmarks/unit_benchmarks.py
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+
+def setup_parameterise():
+    time_create_expression()
+
+
+def time_parameterise():
+
+    global param
```
tinosulzer, line 42: you should be able to do this without using `global`. I have never seen a good reason to use it

benchmarks/unit_benchmarks.py
```diff
@@ -0,0 +1,98 @@
+import pybamm
+import numpy as np
+
+R = 1e-5
+
+
+def time_create_expression():
```
brosaplanella, line 7: You should define a class like with the [other benchmarks](https://github.com/pybamm-team/PyBaMM/blob/develop/benchmarks/time_solve_models.py). The first bit can be pushed under the `setup` method, and then have the different benchmarks.

benchmarks/unit_benchmarks.py
```diff
@@ -0,0 +1,98 @@
...
+
+    spatial_methods = {"negative particle": pybamm.FiniteVolume()}
+    disc = pybamm.Discretisation(mesh, spatial_methods)
+    disc.process_model(model)
+
+
+time_discretise.setup = setup_discretise
+
+
+def setup_solve():
```
brosaplanella, line 86: The discretisation and parameterisation methods modify the model, so if you define `model` to be a variable of the class then hopefully you do not need to initialise every single time.

brosaplanella (changes requested):

benchmarks/unit_benchmarks.py
Vaibhav-Chopra-GT: The benchmarks don't work without these global variables even after adding the setup functions. I don't know why this is the case, but when I run the benchmarks with the global variables they work

benchmarks/unit_benchmarks.py
Vaibhav-Chopra-GT: I tried this but got an error because `asv` runs every method separately, so the changes made in one method don't carry on to others. For example-
for the `time_solve` benchmark to work we will need to run previous methods in its setup, and a class cannot have more than 1 setup; hence, I cannot put this under the class containing other `time_*` functions as they require a different setup

tinosulzer (approved):
Thanks, this looks much cleaner!

brosaplanella (approved):
Looks great, thanks Vaibhav!

PR #2073, opened 2022-05-23
Composite release

tinosulzer:
# Description

- Integrates changes for a composite electrode into the submodel structure.
- Adds "Chen2020_composite" parameter set for a composite graphite/silicon electrode. Silicon parameters are added as a standard negative electrode parameter set, specified as the "negative electrode secondary" set, which automatically adds "Secondary:" to the start of each parameter name. Primary (graphite) parameter names are unchanged.
- Adds example for a Si/Gr negative electrode
- Composite positive electrode is also possible
- Tertiary, quaternary, ... particles are not implemented but would not require a big change
- Many parameters have been renamed from `param.n.c_max` to `param.n.prim.c_max` (to allow for the equivalent `param.n.sec.c_max`)
- Combining with other submodels (polynomial profiles, degradation, etc) is not done here but should be quite easy now.

Fixes #2052 
Fixes #1433 
Fixes #2026

## Type of change

Please add a line in the relevant section of [CHANGELOG.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CHANGELOG.md) to document the change (include PR #) - note reverse order of PR #s. If necessary, also add to the list of breaking changes.

- [x] New feature (non-breaking change which adds functionality)
- [ ] Optimization (back-end change that speeds up the code)
- [ ] Bug fix (non-breaking change which fixes an issue)

# Key checklist:

- [x] No style issues: `$ flake8`
- [ ] All tests pass: `$ python run-tests.py --unit`
- [ ] The documentation builds: `$ cd docs` and then `$ make clean; make html`

You can run all three at once, using `$ python run-tests.py --quick`.

## Further checks:

- [ ] Code is commented, particularly in hard-to-understand areas
- [ ] Tests added that prove fix is effective or that feature works

PR #2094, opened 2022-06-08
add ocp as a separate submodel

tinosulzer:
# Description

Adds OCV as a submodel. Currently there is only one kind of submodel, but #2073 will be adding the current sigmoid.

## Type of change

# Key checklist:

## Further checks:


pybamm/models/full_battery_models/lithium_ion/base_lithium_ion_model.py
```diff
@@ -44,6 +44,32 @@ def __init__(self, options=None, name="Unnamed lithium-ion model", build=False):
             )
         self.set_standard_output_variables()
 
+        self.set_external_circuit_submodel()
+        self.set_porosity_submodel()
+        self.set_interface_utilisation_submodel()
+        self.set_crack_submodel()
+        self.set_active_material_submodel()
+        self.set_transport_efficiency_submodels()
+        self.set_convection_submodel()
```
rtimms, line 53: think you need to move the defn of this into this file from SPM, SPMe, DFN

rtimms (commented):
oh, actually it looks like this had broken the half-cell models?

PR #2114, opened 2022-06-20
Solvent consumption

RuiheLi:
# Description

This script is an example of the following work:
R. Li, S. O’Kane, M. Marinescu, G.J. Offer, Modelling solvent consumption from SEI layer growth in lithium-ion batteries, J. Electrochem. Soc. 169 (2022) 060516. https://doi.org/10.1149/ 1945-7111/ac6f84.

A solvent consumption model is introduced, which is seperated from the PyBaMM core. 
That means, you can simply run this script on any version of PyBaMM without changing any .py files, which is super easy!

## New feature
-   Add a new solvent consumption submodel as an independent notebook in examples\notebooks\Solvent consumption model - RHL - Imperial.ipynb, based on this paper:  R. Li, S. O’Kane, M. Marinescu, G.J. Offer, Modelling solvent consumption from SEI layer growth in lithium-ion batteries, J. Electrochem. Soc. 169 (2022) 060516. https://doi.org/10.1149/ 1945-7111/ac6f84.

# Key checklist:

## Further checks:


tinosulzer:
I think this would be best as a separate repository at this stage, like for example https://github.com/brosaplanella/TEC-reduced-model or https://github.com/brosaplanella/SPMe_SR or https://github.com/rtimms/asymptotic-pouch-cell

PR #2132, opened 2022-06-28
add degradation models benchmarks

Vaibhav-Chopra-GT:
# Description

Added benchmarks for degradation models for SPM and DFN (build, simulate, solve) for each model option by adding model values and model as params.
Also, what should I name them?

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:


tinosulzer:
You should name them based on the option e.g. `TimeBuildModelLossActiveMaterial`

benchmarks/degredation_models.py
```diff
@@ -0,0 +1,549 @@
+import pybamm
+import numpy as np
+
+
+def compute_discretisation(model, param):
+    var_pts = {
+        pybamm.standard_spatial_vars.x_n: 20,
+        pybamm.standard_spatial_vars.x_s: 20,
+        pybamm.standard_spatial_vars.x_p: 20,
+        pybamm.standard_spatial_vars.r_n: 30,
```
brosaplanella, line 10: Are you using 30 in the particles for some reason (e.g. convergence)? Otherwise might be better to keep the default (20).

benchmarks/degredation_models.py
```diff
@@ -0,0 +1,549 @@
+import pybamm
```
brosaplanella, line 1: Small typo in the filename (`degredation` instead of `degradation`).

benchmarks/degredation_models.py
```diff
@@ -0,0 +1,549 @@
...
+        nb_points = 500
+        self.t_eval = np.linspace(0, tmax, nb_points)
+        geometry = self.model.default_geometry
+
+        # load parameter values and process model and geometry
+        param = pybamm.ParameterValues("Ai2020")
+        param.process_model(self.model)
+        param.process_geometry(geometry)
+
+        # set mesh
```
brosaplanella, line 84: Can these lines be replaced by `compute_discretisation`? (It appears multiple times below, so not repeating the comment)

Vaibhav-Chopra-GT:
> how long does it take to run now?

15 minutes

tinosulzer (approved):

PR #2134, opened 2022-06-29
make c_s_max an input to functions

tinosulzer:
# Description

Adds `c_s_max` as an input to some functions, instead of defining it as a parameter inside the function.
This is an attempt to split up #2073 

## Type of change

# Key checklist:

## Further checks:


pybamm/input/parameters/lithium_ion/negative_electrodes/graphite_ORegan2021/graphite_LGM50_electrolyte_exchange_current_density_ORegan2021.py
```diff
@@ -1,7 +1,9 @@
 from pybamm import exp, constants, Parameter
 
 
-def graphite_LGM50_electrolyte_exchange_current_density_ORegan2021(c_e, c_s_surf, T):
+def graphite_LGM50_electrolyte_exchange_current_density_ORegan2021(
+    c_e, c_s_surf, T, c_s_max
```
brosaplanella, line 5: Very picky comment (and maybe quite annoying to implement). Do we want to flip `T` and `c_s_max`? This way we have electrochemical variables first, and temperature last.

PR #2157, opened 2022-07-06
Add work precision sets and an automated workflow for the same  

Vaibhav-Chopra-GT:
# Description
Created work precision sets for time vs abstols, and added a workflow to create a pr with new benchmarks on every release.
- The script `work_precision_sets_abstols.py` creates and saves the plot in `benchmark_images` folder, and prepends the `release_work_precision_sets.md`  file with the plot.
- The workflow runs the script on every release and creates a pull request.

I tested the workflow and it works. I couldn't trigger workflow without a release so I added the workflow dispatch option and manually triggered it on my fork. It created this [PR](https://github.com/Vaibhav-Chopra-GT/PyBaMM/pull/7) which looks good.

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:


benchmarks/work_precision_sets_abstols.py
```diff
@@ -0,0 +1,92 @@
...
+
+        # discretise model
+        disc = pybamm.Discretisation(mesh, model.default_spatial_methods)
+        disc.process_model(model)
+        time = 0
+        for k in range(0, 20):
+
+            solution = solver.solve(model, t_eval=t_eval)
+            time += solution.solve_time.value
+        time = time / 20
```
brosaplanella, line 69: Might be good to define the number of runs as a variable so users can easily change it locally.

brosaplanella (approved):
Very nice! Just a small comment

benchmarks/work_precision_sets/time_vs_abstols.py
```diff
@@ -0,0 +1,126 @@
+import pybamm
```
brosaplanella, line 1: Will make some general comments on this script on how to simplify the code, but they apply to all the scripts I believe.

benchmarks/work_precision_sets/time_vs_abstols.py
```diff
@@ -0,0 +1,126 @@
...
+import matplotlib.pyplot as plt
+
+
+parameters = [
+    "Marquis2019",
+    "Prada2013",
+    "Ramadass2004",
+    "Chen2020",
+]
+models = ["SPM", "DFN"]
```
brosaplanella, line 12: You can actually create a list of PyBaMM models. You can also pass their name (to abbreviate it). This way you can get rid of lines 34-37.

```python3
models = [pybamm.lithium_ion.SPM(name="SPM"), pybamm.lithium_ion.DFN(name="DFN")]
```

You can then access the name as `models[i].name`.

benchmarks/work_precision_sets/time_vs_abstols.py
```diff
@@ -0,0 +1,126 @@
...
+    1.0e-7,
+    1.0e-8,
+    1.0e-9,
+    1.0e-10,
+    1.0e-11,
+    1.0e-12,
+    1.0e-13,
+]
+solvers = ["IDAKLU", "Casadi - safe", "Casadi - fast"]
+
```
brosaplanella, line 27: You can define the figure outside the `for` loop and then loop over the axes (see https://stackoverflow.com/questions/20288842/matplotlib-iterate-subplot-axis-array-through-single-list).

You can also get all the possible combinations of solvers and models by using `itertools.product` (see https://stackoverflow.com/questions/533905/get-the-cartesian-product-of-a-series-of-lists).

If you zip the `axes` and the output of `itertools.product`, you can get rid of lines 68-85. Hopefully, this way it will also be more generalisable (i.e. if we want to add a model).

benchmarks/work_precision_sets/time_vs_abstols.py
```diff
@@ -0,0 +1,126 @@
...
+    1.0e-6,
+    1.0e-7,
+    1.0e-8,
+    1.0e-9,
+    1.0e-10,
+    1.0e-11,
+    1.0e-12,
+    1.0e-13,
+]
+solvers = ["IDAKLU", "Casadi - safe", "Casadi - fast"]
```
brosaplanella, line 26: You can do the same as with the models and define a list of PyBaMM solvers. They also have a name that you can get, but in this case comes by default and you can't set it when defining the solver. But I think it will be fine.

benchmarks/work_precision_sets/time_vs_abstols.py
```diff
@@ -0,0 +1,126 @@
...
+                time = 0
+                runs = 20
+                for k in range(0, runs):
+
+                    solution = solver.solve(model, t_eval=t_eval)
+                    time += solution.solve_time.value
+                time = time / runs
+
+                time_points.append(time)
+            plt.subplot(3, 2, x)
```
brosaplanella, line 97: If you iterate over the axes (assume the iterating variable is `ax`) you can just do

```python3
ax.plot(...)
ax.set_xlabel(...)
ax.set_ylabel(...)
```
and so on (not sure if the ticks and scale options also use the `set_`, check `matplotlib` documentation).

benchmarks/work_precision_sets/time_vs_abstols.py
```diff
@@ -0,0 +1,126 @@
...
+            plt.xscale("log")
+            plt.yscale("log")
+            plt.ylabel("time(s)")
+
+plt.gca().legend(
+    parameters,
+    loc="upper right",
+)
+
+plt.tight_layout()
```
brosaplanella, line 112: I think you have to move the `tight_layout` after you change the plot size.

benchmarks/work_precision_sets/time_vs_abstols.py
```diff
@@ -0,0 +1,126 @@
...
+
+                    solution = solver.solve(model, t_eval=t_eval)
+                    time += solution.solve_time.value
+                time = time / runs
+
+                time_points.append(time)
+            plt.subplot(3, 2, x)
+
+            plt.plot(abstols, time_points)
+            plt.title(f"Work Precision Sets for {model_} with {solver_} solver")
```
brosaplanella, line 100: Remove `"Work Precision Sets for "` and leave only the model and solver. Hopefully this way the titles fit in the plot.

benchmarks/work_precision_sets/time_vs_dt_max.py
```diff
@@ -0,0 +1,119 @@
...
+        plt.plot(dt_max, time_points)
+        plt.title(f"Work Precision Sets for {model_}")
+        plt.xlabel("dt_max")
+        plt.xticks(dt_max)
+        plt.xscale("log")
+        plt.yscale("log")
+        plt.ylabel("time(s)")
+
+
+plt.gca().legend(
```
brosaplanella, line 104: The legend only show the DFN parameters and not the SPM. You can either have two legends (but it will be a bit too packed) or, instead, you put the legend on SPM and you change the order of the SPM parameters so `Ramadass2004` is second. This way, the legend will also work for DFN.

Related to that, why is DFN only tested for 2 parameter sets?

benchmarks/work_precision_sets/time_vs_no_of_states.py
```diff
@@ -0,0 +1,80 @@
...
+
+            plt.plot(ns, solutions)
+            plt.title(f"Work Precision Sets for {model_}")
+            plt.xlabel("number of states")
+            # plt.xticks(ns)
+            plt.xscale("log")
+            plt.yscale("log")
+            plt.ylabel("time(s)")
+
+plt.gca().legend(parameters, loc="upper right", fancybox=True, framealpha=0.5)
```
brosaplanella, line 66: Not sure what is going on here but I can't comment the `png` so I comment the script. The colours look weird, like some are repeated and some don't appear.

brosaplanella (changes requested):

brosaplanella:
Another picky general comment about the plots. The aspect ratio of the plots look a bit strange (e.g. the ones with 2 columns and 1 row are a bit too tall). I found this [blog](https://jwalton.info/Embed-Publication-Matplotlib-Latex/) that gives some hints on how to automatically scale them so they look nice.

benchmarks/work_precision_sets/time_vs_reltols.py
```diff
@@ -0,0 +1,115 @@
...
+]
+
+solvers = {
+    "IDAKLUSolver": pybamm.IDAKLUSolver(),
+    "Casadi - safe": pybamm.CasadiSolver(),
+    "Casadi - fast": pybamm.CasadiSolver(mode="fast"),
+}
+
+
+fig, axs = plt.subplots(len(solvers), len(models), figsize=(8, 5))
```
brosaplanella, line 37: Make taller, e.g. `(8, 10)`

brosaplanella (commented):
Looks much better, good job! Just some minor comments

brosaplanella (approved):
(Changing to approved to previous request changes goes away, but the previous set of minor comments should be implemented before merging).

benchmarks/work_precision_sets/time_vs_no_of_states.py
```diff
@@ -0,0 +1,94 @@
...
+            }
+            sim = pybamm.Simulation(
+                model, solver=solver, parameter_values=param, var_pts=var_pts
+            )
+
+            time = 0
+            runs = 20
+            for k in range(0, runs):
+
+                solution = sim.solve([0, 3500])
```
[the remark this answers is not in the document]
tinosulzer: It's difficult because we don't want to trigger the event when testing casadi fast, but we do want to trigger it when testing casadi safe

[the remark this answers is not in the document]
brosaplanella: Good point. Then I think this is fine as it is.

benchmarks/work_precision_sets/time_vs_abstols.py
```diff
@@ -0,0 +1,115 @@
...
+    itertools.product(models.values(), solvers.values()),
+    itertools.product(models, solvers),
+):
+
+    for params in parameters:
+        time_points = []
+        solver = i[1]
+
+        model = i[0].new_copy()
+        c_rate = 10
```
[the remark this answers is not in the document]
Vaibhav-Chopra-GT: When I use any `c_rate` = 1, I get solver errors for a lot of combinations -
```
traceback (most recent call last):
  File "d:\code\PyBaMM\benchmarks\work_precision_sets\time_vs_abstols.py", line 86, in <module>
    solver.solve(model, t_eval=t_eval)
  File "C:\Users\Vaibhav Chopra\AppData\Local\Programs\Python\Python39\lib\site-packages\pybamm\solvers\base_solver.py", line 1083, in solve        
  File "C:\Users\Vaibhav Chopra\AppData\Local\Programs\Python\Python39\lib\site-packages\pybamm\solvers\casadi_solver.py", line 177, in _integrate  
    solution = self._run_integrator(
  File "C:\Users\Vaibhav Chopra\AppData\Local\Programs\Python\Python39\lib\site-packages\pybamm\solvers\casadi_solver.py", line 741, in _run_integrator
    raise pybamm.SolverError(e.args[0])
pybamm.expression_tree.exceptions.SolverError: .../casadi/interfaces/sundials/cvodes_interface.cpp:392: CVode returned "CV_RHSFUNC_FAIL". Consult CVODES documentation.
```
If I use a try-except block to skip through the error, I get different lengths of x and y coordinate lists which cannot be plotted together.
Should I manually create x and y coordinates while iterating through the combinations rather than using the list defined at the top directly?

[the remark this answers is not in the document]
brosaplanella: That's weird. Do you know exactly which combinations are actually failing?

Can you try setting `tmax = 3500` (like the example below) and see if that works better?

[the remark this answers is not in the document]
Vaibhav-Chopra-GT: All work now except `Prada2013` with Casadi - fast, so I replaced it with Ecker2015 which was previously giving error, but now works!

brosaplanella:
Looks great, thanks! Are there any changes you need to do or shall I merge?

Vaibhav-Chopra-GT:
> Looks great, thanks! Are there any changes you need to do or shall I merge?

Codacy is erroring out on the '_' in the file name. Should I rename the file or should I leave it as it is?
No other changes from my side!

brosaplanella:
I believe it is because for Markdown files it is recommended to use `-` instead of `_`. Maybe try renaming `release_work_precision_sets.md` to `release-work-precision-sets.md` and see if it works.

Issue #2168, opened 2022-07-11
[Bug]: Using `build=False` when creating model doesn't work


brosaplanella:
### PyBaMM Version

22.6

### Python Version

3.7

### Describe the bug

In version 22.6 if you create a model with `build=False` the model is built anyway. That didn't happen in 22.5

### Steps to Reproduce

1. Run
```python3
import pybamm

model = pybamm.lithium_ion.SPMe(build=False)
print(model._built)

model.build_model()
```
2. See error
3. Switch to v22.5 and observe there's no error

rtimms:
Is it just for SPMe? I think we need to pass `build` [here](https://github.com/pybamm-team/PyBaMM/blob/develop/pybamm/models/full_battery_models/lithium_ion/spme.py#L47).

brosaplanella:
Just checked and it doesn't happen with DFN, good catch! Is there an open PR for it or shall I open it?

rtimms:
Great! I am working on #2056 so can fix it as part of that. Hopefully will make a PR today or tomorrow.

brosaplanella:
Cool! I can work with v22.5 for the moment so no rush :)

rtimms:
Fixed by #2169

PR #2169, opened 2022-07-11
Move submodels and build model

### LINKED ISSUE #2056: Move submodels and `build_model` to `BaseModel`
Refactor to move general code about submodels to `BaseModel` instead of `BaseBatteryModel`. This would allow people to write their own models that are composed of their own submodels. This would be a step to allowing more general models in pybamm and would also be super useful for training. Being able to write your own submodels and see how they interact would help with learning how all the battery submodels fit together. 

I think this would mainly be moving `build_model` and the methods it calls from `BaseBatteryModel` to `BaseModel`. Things like `set_voltage_variables()` would obviously need to stay in `BaseBatteryModel`. 

The only thing I haven't considered is if this is compatible with just defining `rhs` etc. explicitly? I think it is.

Thoughts??

rtimms:
# Description
Move general code about submodels to `BaseModel` instead of `BaseBatteryModel`. This makes it easier to build custom models from submodels.

Also fixes a bug where SPMe always build even when `build=False`.

Fixes #2056 #2168 

## Type of change

# Key checklist:

## Further checks:


brosaplanella (approved):
Looks good, thanks Rob!

PR #2175, opened 2022-07-16
bugfix(double-layer cap scaling): rescaling overpotential RHS by 1/a

bardsleypt:
# Description
Rescaling RHS of overpotential differential equation by 1/a to correct for proper non-dimensionalization for non-constant specific-surface areas (i.e., non-constant active material volume fractions and/or particle radii). 

Changes include:
  - Updating full_surface_form_conductivity RHS to rescale RHS of
    negative-electrode overpotential DE by a factor of 1/a, which
    remains present upon non-dimensionalization
  - 1/a scaling only affects DFN models with 'surface form' =
    'differential' and any of the following:
    - non-constant particle radii across electrodes
    - non-constant active material volume fractions across electrodes
    - loss of active material enabled
    - other forms of non-constant specific surface areas 'a'

See Discussion #2142 for further details

## Type of change

# Key checklist:

- [x] No style issues: $ flake8
- [ ] All tests pass: $ python run-tests.py --unit
- [x] The documentation builds: $ cd docs and then $ make clean; make html

You can run all three at once, using $ python run-tests.py --quick.

## Further checks:

### DIFF
--- a/pybamm/models/submodels/electrolyte_conductivity/surface_potential_form/full_surface_form_conductivity.py
+++ b/pybamm/models/submodels/electrolyte_conductivity/surface_potential_form/full_surface_form_conductivity.py
@@ -283,4 +283,4 @@ class FullDifferential(BaseModel):
             "Sum of " + self.domain.lower() + " electrode interfacial current densities"
         ]
 
-        self.rhs[delta_phi] = 1 / C_dl * (pybamm.div(i_e) - a * sum_j)
+        self.rhs[delta_phi] = 1 / (a * C_dl) * (pybamm.div(i_e) - a * sum_j)


bardsleypt:
I ran the tests, but a lot of the casadi interface tests seemt to fail - likely because of my install. I have had a hard time getting the dev install to run successfully for an M1 mac - pip doesn't want to install scikits.odes so manual installs and tox installs all seem to fail. Any advice on this is appreciated, otherwise maybe someone can check the tests for me on a proper dev-installed machine?

Figured for this PR its a small enough change that I shouldn't be causing significant issues, at least not that can't be obviously tracked (failing surface-form unit test for example).

tinosulzer (approved):
The CI will catch any failing tests, I would be very very surprised if these changes caused tests to fail

tinosulzer:
@all-contributors add @bardsleypt for bugs, code

PR #2207, opened 2022-08-01
#1986 remove zero electrolyte concentration cut-off

### LINKED ISSUE #1986: Remove zero-concentration cut-off event from DFN
- add `cutoff_event` kwarg (boolean, default False) to the full electrolyte model
- in `set_events` in the full electrolyte model, don't add the event if `cutoff_event` is `False`
- make `True` for the SPMe (for cut-off, see discussion in #1982 )
- make the cut-off tolerance smaller when `cutoff_event` is True

Easy issue, assuming the models still work without the event

tinosulzer:
# Description

Removes the "Zero electrolyte concentration cut-off" event.
Adds some bounds to functions so that the solver still works even if we step into electrolyte depletion

Fixes #1986 

## Type of change

# Key checklist:

## Further checks:


pybamm/parameters/lithium_ion_parameters.py
```diff
@@ -683,6 +693,9 @@ def D_dimensional(self, sto, T):
 
     def j0_dimensional(self, c_e, c_s_surf, T):
         """Dimensional exchange-current density [A.m-2]"""
+        # tol = 1e-3
+        # c_e = c_e  # pybamm.maximum(c_e, tol)
+        # c_s_surf = pybamm.maximum(pybamm.minimum(c_s_surf, 1 - tol), tol)
```
brosaplanella, line 698: Should be deleted

pybamm/parameters/lithium_ion_parameters.py
```diff
@@ -199,11 +199,13 @@ def _set_dimensional_parameters(self):
 
     def D_e_dimensional(self, c_e, T):
         """Dimensional diffusivity in electrolyte"""
+        c_e = pybamm.maximum(c_e, 10)
```
brosaplanella, line 202: Out of curiosity, which kind of error do you get if you make this smaller?

Related to that, do we want to hard-code this or set some tolerances that the user can change?

brosaplanella (approved):
Looks good, thanks!

pybamm/parameters/lithium_ion_parameters.py
tinosulzer: Making this smaller gives more casadi jacobian warnings. Yes, I will make it into a setting

PR #2211, opened 2022-08-02
add a few more expression simplifications

tinosulzer:
# Description

A few more simplifications of the expression tree. Spotted these while debugging but I doubt they will be very common

## Type of change

# Key checklist:

## Further checks:


pybamm/expression_tree/binary_operators.py
```diff
@@ -1309,18 +1315,35 @@ def simplified_matrix_multiplication(left, right):
...
-    elif isinstance(right, Addition):
-        if (right.left.is_constant() or right.right.is_constant()) and not (
+    elif isinstance(right, (Addition, Subtraction)):
+        # Simplify A @ (b +- c) to (A @ b) +- (A @ c) if (A @ b) or (A @ c) is constant
+        # This is a common construction that appears from discretisation of spatial
+        # operators
+        # Or simplify A @ (B @ b +- C @ c) to (A @ B @ b) +- (A @ C @ c) if (A @ B)
+        # and (A @ C) are constant
+        # Don't do this if either b or c is a number as this will lead to matmul errors
+        if (
```
rtimms, line 1325: Do we need A to be constant here too? I.e. `left.is_constant()`?

rtimms (approved):
thanks, looks good!

PR #2212, opened 2022-08-03
Issue 2040 events

tinosulzer:
# Description

Raises a SolverError if the initial conditions trigger the events. This requires the events to always be written in a way that they are positive initially, and should be triggered when they become negative.

In an experiment, if an "[experiment]" event is triggered, then the step is just skipped (returns `EmptySolution`) without raising an error. This is to deal with cases such as "Charge at 1C until 4.2V" when we start from full SOC.

Fixes #2040 

## Type of change

# Key checklist:

## Further checks:


brosaplanella (approved):
Looks good, thanks!

PR #2215, opened 2022-08-04
add method to calculate condition number

Vaibhav-Chopra-GT:
# Description

Created a method `condition_number` in `Solutions` class that calculates and returns the condition number
Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/solvers/solution.py
+++ b/pybamm/solvers/solution.py
@@ -780,12 +780,20 @@ class Solution(object):
 
         return new_sol
 
-    def condition_number(self, model):
+    def condition_number(self):
         # returns condition number for the solution
-        f = model.rhs_algebraic_eval(self.t[0], self.y[:, 0], [])
-        J = model.jac_rhs_algebraic_eval(self.t[0], self.y[:, 0], [])
 
-        return np.linalg.norm(J) / (np.linalg.norm(f) / np.linalg.norm(self.y[:, 0]))
+        entries = np.empty(len(self.t))
+        for ts, ys, model in zip(self.all_ts, self.all_ys, self.all_models):
+            for inner_idx, t in enumerate(ts):
+                t = ts[inner_idx]
+                y = ys[:, inner_idx]
+                f = model.rhs_algebraic_eval(t, y, [])
+                J = model.jac_rhs_algebraic_eval(t, y, [])
+                cond = np.linalg.norm(J) / (np.linalg.norm(f) / np.linalg.norm(y))
+                np.insert(entries, -1, cond)
+
+        return entries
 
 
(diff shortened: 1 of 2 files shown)


pybamm/solvers/solution.py
```diff
@@ -780,6 +780,13 @@ def copy(self):
 
         return new_sol
 
+    def condition_number(self, model):
+        # returns condition number for the solution
+        f = model.rhs_algebraic_eval(self.t[0], self.y[:, 0], [])
```
tinosulzer, line 785: you can use `self.all_models[0]` here instead of `model`

pybamm/solvers/solution.py
```diff
@@ -780,6 +780,13 @@ def copy(self):
...
+        J = model.jac_rhs_algebraic_eval(self.t[0], self.y[:, 0], [])
```
tinosulzer, line 786: loop over all indices of `self.t` and `self.y` and return a vector of length `len(sol.t)`

tinosulzer (changes requested):

tinosulzer:
See https://github.com/pybamm-team/PyBaMM/blob/3c7ad7c8f9d093bab2860bb2664986007a07c26c/pybamm/solvers/processed_variable.py#L109-L119 for constructing the condition number array

tinosulzer:
```python3
for ts, ys, inputs, model in zip( 
     self.all_ts, self.all_ys, self.all_inputs_casadi, self.all_models
 ): 
```

Vaibhav-Chopra-GT:
Thank you for the review, @tinosulzer! I have made some changes could you please check if they are correct? I will add tests once the function is ready.

tinosulzer (approved):
Looks good, just needs a test

tests/unit/test_solvers/test_solution.py
```diff
@@ -346,6 +346,23 @@ def test_solution_evals_with_inputs(self):
         time = sim.solution["Time [h]"](sim.solution.t)
         self.assertEqual(len(time), 10)
 
+    def test_condition_number(self):
```
tinosulzer, line 349: This is not a great test, since it just copies the code in the function. If the code gives the wrong result then the test won't catch it. It's better to test that the code outputs the expected result in some specific cases. These can be very simple cases, e.g. `dx/dt = 1` and `dx/dt = x`, in which case you can calculate the jacobian (and hence the condition number) explicitly

tinosulzer: Another test could be to check that the condition number gets larger as the problem gets stiffer

rtimms:
Closing this as it’s been hanging around a while and not sure it’s needed.

PR #2222, opened 2022-08-09
Added new parameter z_sei to allow the user to set SEI valancy

### LINKED ISSUE #2065: [Bug]: Scaling by 2x in SEI growth?
### PyBaMM Version

22.3

### Python Version

3.9.12

### Describe the bug

I believe there may be a scaling bug present in the SEI growth model by a factor of 2x, at least when considering the 'solvent-diffusion limited' growth model for SEI. Following is my reasoning (apologies for clunky equations as it doesn't appear TeX is supported here).

As determined from PyBaMM source code, the ODE for the *dimensional* SEI growth takes the form

dL/dt = (1/2) (c_sol * D_sol * V_sei) / L

where, by default there is an L_inner, and L_outer which satisfy the same growth law. I am fine with this growth law, crucially the scaling by (1/2), **when there is only a single SEI layer considered**. My current understanding of the (1/2) factor is from the stoichiometry of the SEI reaction requiring 2 moles of lithium. 

However, **when there is a double-layer SEI considered** (which is the behavior in models/submodels/interface/sei/sei_growth.py)., it seems the law should read

dL/dt = (1/4) (c_sol * D_sol * V_se) / L

My understanding of this scaling by (1/4) is due to the same stoichiometric scaling of (1/2), and an additional scaling of (1/2) coming from assuming the growth of the SEI layers is the same for both inner/outer. Thus half of the SEI growth comes from the inner layer growth, and the other half from the outer layer growth.

I have verified that the scaling in both the L_inner and L_outer growth laws does use the 1/2 scaling, as opposed to the 1/4 scaling. Thus, it would seem the default PyBaMM SEI growth model implements a 2x speed up of the SEI layer growth. My reference for this scaling comes directly from one of the papers using the Chen2020_plating parameter set:

O’Kane, Simon E. J., et al. “Lithium-Ion Battery Degradation: How to Model It.” Physical Chemistry Chemical Physics, vol. 24, no. 13, 2022, pp. 7909–22. DOI.org (Crossref), https://doi.org/10.1039/D2CP00417H.

As well as the supplemental information for this article.

### Steps to Reproduce

[pybamm_sei_scaling.pdf](https://github.com/pybamm-team/PyBaMM/files/8720394/pybamm_sei_scaling.pdf)

PDF  is an exported jupyter notebook (with execution displays) with the following commands:

```python
import pybamm
import numpy as np

model = pybamm.lithium_ion.DFN({'SEI': 'solvent-diffusion limited', 
                                'SEI porosity change': 'true'})
params = pybamm.ParameterValues('Chen2020_plating')

dLdt = list(model.submodels['sei'].rhs.values())[0]
dLdt = params.process_symbol(dLdt)
dLdt.render()

def rhs_from_model(L_outer):
    A = dLdt.left.evaluate()
    B = dLdt.right.left.evaluate()
    nondim_rhs = A*B/L_outer
    return nondim_rhs

def rhs_from_okane(L_outer, scl=0.5, tau_s=1, L0=1):
    dim_rhs = params['Bulk solvent concentration [mol.m-3]']*params['Outer SEI solvent diffusivity [m2.s-1]']*\
        params['Outer SEI partial molar volume [m3.mol-1]'] / L_outer
    nondim_scl = tau_s / L0 / L0 # tau_s for time scaling, one copy L0 for LHS ODE, one copy for RHS ODE
    return scl*nondim_scl*dim_rhs

tau_s = params.evaluate(model.timescale)
L0 = params.evaluate(model.param.L_sei_0_dim)

model_rhs = rhs_from_model(np.logspace(-9, -8, 10))
model_okane2 = rhs_from_okane(np.logspace(-9, -8, 10), scl=0.5, tau_s=tau_s, L0=L0)
model_okane4 = rhs_from_okane(np.logspace(-9, -8, 10), scl=0.25, tau_s=tau_s, L0=L0)

relerr2 = np.linalg.norm(model_rhs - model_okane2) / np.linalg.norm(model_rhs)
relerr4 = np.linalg.norm(model_rhs - model_okane4) / np.linalg.norm(model_rhs)
print('Rel. Error (1/2 scaling) = {:.4g}'.format(relerr2))
print('Rel. Error (1/4 scaling) = {:.4g}'.format(relerr4))
```

### Relevant log output

_No response_

DrSOKane:
# Description

A new parameter "Ratio of lithium moles to SEI moles" (short name z_sei) has been added. This variable was previously hardcoded to 1 for all SEI growth options except "ec reaction limited". As pointed out in issue #2065, z_sei=1 is incorrect for the two most common SEI compoints - LEDC and Li2CO3, both of which require z=2.

The SEI parameter sets have been updated accordingly. The new parameter z_sei is set to 2 by default for all parameter sets except OKane2022, which used the old code for which it was set to 1.

Fixes #2065 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/models/submodels/interface/sei/base_sei.py
+++ b/pybamm/models/submodels/interface/sei/base_sei.py
@@ -167,7 +167,7 @@ class BaseModel(BaseInterface):
             v_bar = 1
             L_inner_0 = 0
             L_outer_0 = 0
-            li_mols_per_sei_mols = 1
+            z_sei = 1
         else:
             if self.reaction_loc == "interface":
                 # scales in mol/m2 (n is an interfacial quantity)
@@ -186,11 +186,11 @@ class BaseModel(BaseInterface):
             if self.options["SEI"] == "ec reaction limited":
                 L_inner_0 = 0
                 L_outer_0 = 1
-                li_mols_per_sei_mols = 2
+                z_sei = 2
             else:
                 L_inner_0 = param.L_inner_0
                 L_outer_0 = param.L_outer_0
-                li_mols_per_sei_mols = 1
+                z_sei = param.z_sei
 
         if self.reaction == "SEI":
             L_inner = variables["Inner SEI thickness"]
@@ -215,7 +215,7 @@ class BaseModel(BaseInterface):
                 L_n = self.param.n.L
 
             Q_sei = (
-                li_mols_per_sei_mols
+                z_sei
                 * delta_n_SEI
                 * n_scale
                 * L_n
@@ -259,7 +259,7 @@ class BaseModel(BaseInterface):
 
             # Q_sei_cr in mol
             Q_sei_cr = (
-                li_mols_per_sei_mols
+                z_sei
                 * delta_n_SEI_cr
                 * n_scale
                 * self.param.n.L
--- a/pybamm/models/submodels/interface/sei/base_sei.py
+++ b/pybamm/models/submodels/interface/sei/base_sei.py
@@ -182,15 +182,14 @@ class BaseModel(BaseInterface):
                     param.L_sei_0_dim * param.n.a_typ / param.V_bar_outer_dimensional
                 )
             v_bar = param.v_bar
+            z_sei = param.z_sei
             # Set scales for the "EC Reaction Limited" model
             if self.options["SEI"] == "ec reaction limited":
                 L_inner_0 = 0
                 L_outer_0 = 1
-                z_sei = 2
             else:
                 L_inner_0 = param.L_inner_0
                 L_outer_0 = param.L_outer_0
-                z_sei = param.z_sei
 
         if self.reaction == "SEI":
             L_inner = variables["Inner SEI thickness"]
--- a/pybamm/parameters/lithium_ion_parameters.py
+++ b/pybamm/parameters/lithium_ion_parameters.py
@@ -412,9 +412,6 @@ class LithiumIonParameters(BaseParameters):
 
         self.c_sei_init = self.c_ec_0_dim / self.c_sei_outer_scale
 
-        # I don't think this gets used anywhere?
-        self.beta_sei = self.n.a_typ * self.L_sei_0_dim * self.Gamma_SEI
-
         # lithium plating parameters
         self.c_plated_Li_0 = self.c_plated_Li_0_dim / self.c_Li_typ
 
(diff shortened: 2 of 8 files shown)


pybamm/models/submodels/interface/sei/base_sei.py
```diff
@@ -186,11 +186,11 @@ def _get_standard_concentration_variables(self, variables):
             if self.options["SEI"] == "ec reaction limited":
                 L_inner_0 = 0
                 L_outer_0 = 1
-                li_mols_per_sei_mols = 2
+                z_sei = 2
```
tinosulzer, line 189: I would make this `param.z_sei` now too

pybamm/parameters/lithium_ion_parameters.py
```diff
@@ -332,7 +332,9 @@ def _set_dimensionless_parameters(self):
         )
 
         # SEI parameters
-        self.alpha_sei = pybamm.Parameter("Inner SEI reaction proportion")  # was 0.5
+        self.inner_prop = pybamm.Parameter("Inner SEI reaction proportion")  # was 0.5
```
tinosulzer, line 335: would be good to have `_sei` in this parameter name

pybamm/parameters/lithium_ion_parameters.py
```diff
@@ -407,9 +409,17 @@ def _set_dimensionless_parameters(self):
                 )
             )
         )
-        self.beta_sei = self.n.a_typ * self.L_sei_0_dim * self.Gamma_SEI
+
         self.c_sei_init = self.c_ec_0_dim / self.c_sei_outer_scale
 
+        # This ensures z_sei is hardcoded to 2 if ec reaction limited is selected
```
tinosulzer, line 415: I don't this needs to be hardcoded

tinosulzer (commented):
I don't see `beta_sei` and `beta_plating` being used anywhere

pybamm/models/submodels/interface/sei/base_sei.py
DrSOKane: Done

pybamm/parameters/lithium_ion_parameters.py
DrSOKane: Changed to inner_sei_proportion

pybamm/parameters/lithium_ion_parameters.py
DrSOKane: Removed

DrSOKane:
> I don't see `beta_sei` and `beta_plating` being used anywhere

I remember beta_sei being a legacy parameter from when the porosity submodel was an ODE, because I created beta_plating using the same logic. The porosity submodel seems to be ever-changing so is it worth keeping these?

tinosulzer:
Let's remove them, it's easy to add back in if needed, and confusing otherwise

tinosulzer (approved):

PR #2223, opened 2022-08-10
Increase coverage

brosaplanella:
Various fixes to increase coverage and improve docstrings.


tinosulzer (approved):
Thanks!

brosaplanella:
There are still some additional changes I want to do before merging

tests/unit/test_expression_tree/test_symbol.py
```diff
@@ -244,6 +249,13 @@ def test_evaluate_ignoring_errors(self):
         self.assertEqual(pybamm.t.evaluate_ignoring_errors(t=0), 0)
         self.assertIsNone(pybamm.Parameter("a").evaluate_ignoring_errors())
         self.assertIsNone(pybamm.StateVector(slice(0, 1)).evaluate_ignoring_errors())
+        self.assertIsNone(pybamm.StateVectorDot(slice(0, 1)).evaluate_ignoring_errors())
+
+        # with self.assertRaisesRegex(pybamm.ShapeError, "Cannot find shape"):
+        #     pybamm.inner(
+        #         pybamm.Matrix(np.zeros((2, 3))), pybamm.Matrix(np.ones((4, 5)))
+        #     ).evaluate_ignoring_errors()
```
brosaplanella, line 257: For some reason this test works on my computer but fails on GitHub actions. Locally it catches the `pybamm.ShapeError` but on GitHub it triggers the previous `ValueError` (e.g. see https://github.com/pybamm-team/PyBaMM/runs/7972858999?check_suite_focus=true). Any ideas?

tinosulzer: It could be something to do with whether `debug_mode` is True or False, it sometimes behaves differently if you run an individual test script vs the whole test suite

brosaplanella: Nvm. It did something very weird: when debugging it didn't access that line, but if I change it I get an error because `pybamm.ShapeError` is not raised.

brosaplanella:
I don't think I will have time to work on this in the next few weeks so I think it is better to merge. If you think otherwise let me know, otherwise I will merge tomorrow.

tests/unit/test_expression_tree/test_binary_operators.py
```diff
@@ -265,6 +267,19 @@ def test_inner(self):
         # check doesn't evaluate on edges anymore
         self.assertEqual(model.variables["inner"].evaluates_on_edges("primary"), False)
 
+        # check _binary_jac
```
tinosulzer, line 270: jacobian should be tested in `test_jac.py`

tests/unit/test_expression_tree/test_concatenations.py
```diff
@@ -46,6 +46,13 @@ def test_base_concatenation(self):
         with self.assertRaisesRegex(TypeError, "ConcatenationVariable"):
             pybamm.Concatenation(a, b)
 
+        # base concatenation jacobian
+        a = pybamm.Symbol("a", domain="test a")
+        b = pybamm.Symbol("b", domain="test b")
+        conc3 = pybamm.Concatenation(a, b)
+        with self.assertRaises(NotImplementedError):
+            conc3._concatenation_jac(None)
```
tinosulzer, line 54: this is fine here

tinosulzer (commented):
thanks, just one minor comment!

tests/unit/test_expression_tree/test_binary_operators.py
tinosulzer: yes

PR #2231, opened 2022-08-15
unpin numpy since scikits is fixed

tinosulzer:
# Description

Revert numpy to >= 1.16 since scikits.odes has released fixed version


Saransh-cpp (approved):
Thanks, @tinosulzer!

PR #2236, opened 2022-08-16
simplify nondimensionalisation of kinetics

tinosulzer:
# Description

Simplifies non-dimensionalisation of the kinetics, which I think was making things overly complicated.

Previously, `j0` was calculated as =
```python
j0 = gamma * j0(...) / C_r
```
but
Screen Shot 2022-08-16 at 5 24 10 PM

and we are currently scaling `j0(...)` as `j0(...) = j0_dim(...) / j0_ref`, so 

```python
j0 = gamma * j0_dim(...) / j0_ref / C_r = j0_dim(...) / j_scale
```

and if instead we scale `j0(...) = j0_dim(...) / j_scale` then

```python
j0 = j0(...)
```

## Type of change

# Key checklist:

## Further checks:


tinosulzer:
just sanity checking

Screen Shot 2022-08-17 at 5 40 21 PM

brosaplanella (approved):
Looks good, thanks @tinosulzer!

brosaplanella:
BTW, should we set a deprecation warning for the unused constants with the deprecated package?

tinosulzer:
> BTW, should we set a deprecation warning for the unused constants with the deprecated package?

the parameters are still there, so where were you thinking of adding the warning?

Also I have no idea why that test is failing, and only on python 3.7

brosaplanella:
Yes, not too fussed about the test.

I mean that, if the dimensionless numbers are not used, we might want to get rid of them eventually. If so, we can use deprecated to fail at a given release so we remove it.

Issue #2017, opened 2022-04-13
Allow cubic interpolant for 2D data


TomTranter:
### Description

Currently the scipy interpolate.interp2d function is called if there are 2 x arguments and y is 2d but we only allow "linear" interpolants. The function will accept "cubic" and "quintic" also. It would be useful to have these options accessible. Were they prevented for any good reason? Convergence issues maybe?

### Motivation

Functions that interpolate 2 variables will be more accurately represented.

tinosulzer:
Don't think they were prevented for any particular reason, there just wasn't a good reason to have them.
We should definitely add them. But we should be wary of the fact that they don't preserve monotonicity

rtimms:
Fixed in #2229

PR #2258, opened 2022-08-24
#2229 default linear interp

### LINKED ISSUE #2229: Make default interpolator linear
I think we should change the default interpolant to be linear. This is normally the default for all other interpolators and is the "safest" option in the sense that you don't get oscillations or introduce maxima.

rtimms:
# Description

Make the default interpolator linear. 

Fixes #2229 #2017 

## Type of change

# Key checklist:

## Further checks:


tinosulzer (approved):
why is "cubic" better than "cubic spline"? Cubic spline is used by scipy https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.CubicSpline.html

rtimms:
I originally changed it as I was going to just pass `kind=interpolator` to `interp1d`, but you couldn't take the derivative of that so changed back to using `CubicSpline`. Then I left it as "cubic" for no real reason. You can just pass  `kind=interpolator` to `interp2d`.

pybamm/parameters/parameter_values.py
```diff
@@ -617,8 +617,14 @@ def _process_symbol(self, symbol):
...
                         input_data = data
 
+                    # For parameters provided as data we use a cubic interpolant
+                    # Note: the cubic interpolant can be differentiated
                     function = pybamm.Interpolant(
-                        input_data[0], input_data[-1], new_children, name=name
+                        input_data[0],
+                        input_data[-1],
+                        new_children,
+                        interpolator="cubic",
```
tinosulzer, line 626: so the default when creating an interpolant is "linear", but the default when using data is "cubic"?

rtimms: Yeah, I didn't like doing this. I think it would be better to use "linear" all the time, but you can't take the derivative. We could switch to using [InterpolatedUnivariateSpline](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.InterpolatedUnivariateSpline.html) for the 1D interpolation and just change the degree, then you can, but it's not smooth anyway.

PR #2260, opened 2022-08-25
Calculate total current for each phase separately

tinosulzer:
# Description

Calculate total current ("sum of ...") separately for each phase, which is needed for some of the degradation models (e.g. SEI)

The variables like "Sum of negative electrode interfacial current densities" have been removed, since when summing interfacial current densities we should always sum over the _volumetric_ ones ("Sum of negative electrode volumetric interfacial current densities") to account for differences in surface area to volume ratio between different reactions or phases

## Type of change

# Key checklist:

## Further checks:


brosaplanella (approved):
Looks good, thanks @tinosulzer

PR #2262, opened 2022-08-25
Bug fix: added SEI on cracks to total_kinetics.py

DrSOKane:
# Description

Previously, the loop in total_kinetics.py did not include the SEI on cracks reaction, so the SEI on cracks did not actually remove any lithium from the system. This has now been fixed.

## Type of change

# Key checklist:

## Further checks:


tinosulzer:
> The space in reaction_name moved, because reasons

It's to keep you on your toes

tinosulzer:
Wouldn't it be easier to put the roughness variable into the SEI current? I'm just thinking about conservation of lithium with this approach (we need to check conservation of lithium under side reactions more carefully)

DrSOKane:
> Wouldn't it be easier to put the roughness variable into the SEI current? I'm just thinking about conservation of lithium with this approach (we need to check conservation of lithium under side reactions more carefully)

You would think so, but the fundamental variable for SEI - including on cracks - is thickness, not concentration, so multiplying by roughness would give the wrong SEI thickness, which matters a lot in the diffusion-limited cases.

There are three places where I've deemed it necessary to multiply by roughness:
1) When calculating the SEI concentration, which also determines the "Loss of lithium to SEI" variables.
2) When calculating the volumetric current density, because that determines all the source terms
3) When incrementing the "sum of interfacial current densities" variables, to ensure current conservation.

The only way to make it less confusing that I can think of is to change the fundamental variable to concenration instead of thickness.

tinosulzer:
So is concentration not always directly proportional to thickness? If not, what is the equation that relates thickness and concentration in the general case?

DrSOKane:
> So is concentration not always directly proportional to thickness? If not, what is the equation that relates thickness and concentration in the general case?

n_inner_cr = L_inner_cr * (roughness - 1)  # inner SEI cracks concentration
n_outer_cr = L_outer_cr * (roughness - 1)  # outer SEI cracks concentration

tinosulzer:
In that case it would be cleaner to have the fundamental variable be concentration (and for lithium plating too). Could you try this? It shouldn't be too difficult.

DrSOKane:
> In that case it would be cleaner to have the fundamental variable be concentration (and for lithium plating too). Could you try this? It shouldn't be too difficult.

I started to code this before realizing that the conversion between the two depends on the surface area to volume ratio. This matters because the initial condition is (a) given in metres, not mol.m-3 and (b) not zero. Whereas for plating there is no distiction between concentration and thickness, because the initial condition tends to be zero.

tinosulzer:
Could you calculate the initial concentration based on the initial thickness? We know the initial surface area to volume ratio, even as a function of x.

DrSOKane:
> Could you calculate the initial concentration based on the initial thickness? We know the initial surface area to volume ratio, even as a function of x.

I could do that next week.

To be honest, the initial SEI concentration should be part of the cell balancing process. That's beyond the scope of this PR, but making concentration the fundamental variable is a stepping stone that will make it possible in the future.

CHANGELOG.md
```diff
@@ -20,13 +20,17 @@
 
 ## Bug fixes
 
+-   Added `SEI on cracks` to loop over all interfacial reactions ([#2262](https://github.com/pybamm-team/PyBaMM/pull/2262))
+-   Fixed `X-averaged SEI on cracks concentration` so it's an average over x only, not y and z ([#2262](https://github.com/pybamm-team/PyBaMM/pull/2262))
+-   Corrected initial state for SEI on cracks ([#2262](https://github.com/pybamm-team/PyBaMM/pull/2262))
```
tinosulzer, line 25: these need to be moved to the newer section of the changelog

tinosulzer (approved):
Thanks, will merge once tests pass

PR #2272, opened 2022-08-29
Add action to format code with black

priyanshuone6:
# Description

Add an github action to format code with black and push it to the existing branch.

## Type of change

# Key checklist:

## Further checks:


Saransh-cpp:
I think it might be better to use [`pre-commit`](https://pre-commit.com/) instead of manually writing down the workflow. The [`pre-commit bot`](https://pre-commit.ci/) will then automatically commit on the branch/PR making the required changes.

A `.pre-commit-config.yaml` file like this should work -
```yml
ci:
  autoupdate_commit_msg: "chore: update pre-commit hooks"
  autofix_commit_msg: "style: pre-commit fixes"

repos:
  - repo: https://github.com/asottile/pyupgrade
    rev: v2.37.3
    hooks:
      - id: pyupgrade
        args: [--py37-plus]

  - repo: https://github.com/psf/black
    rev: 22.8.0
    hooks:
      - id: black-jupyter

  - repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v4.3.0
    hooks:
      - id: check-added-large-files
      - id: check-case-conflict
      - id: check-merge-conflict
      - id: check-symlinks
      - id: check-yaml
      - id: debug-statements
      - id: end-of-file-fixer
        exclude: ^docs
      - id: mixed-line-ending
      - id: requirements-txt-fixer
      - id: trailing-whitespace

  - repo: https://github.com/PyCQA/isort
    rev: 5.10.1
    hooks:
      - id: isort

  - repo: https://github.com/PyCQA/flake8
    rev: 5.0.4
    hooks:
      - id: flake8
        additional_dependencies:
          - flake8-bugbear
          - flake8-docstrings
          - flake8-print

  - repo: https://github.com/codespell-project/codespell
    rev: v2.2.1
    hooks:
      - id: codespell
        args: ["-L", "hist,nd,circularly,ba"]
        exclude: ^(notebooks/xarray.ipynb|notebooks/BoostHistogramHandsOn.ipynb)$

  - repo: https://github.com/pre-commit/mirrors-prettier
    rev: "v3.0.0-alpha.0"
    hooks:
      - id: prettier
        types_or: [yaml, markdown, html, css, scss, javascript, json]
        exclude: assets/js/webapp\.js

  - repo: https://github.com/asottile/blacken-docs
    rev: v1.12.1
    hooks:
      - id: blacken-docs
        args: ["-E"]
        additional_dependencies: [black==22.8.0]

  - repo: https://github.com/pre-commit/pygrep-hooks
    rev: v1.9.0
    hooks:
      - id: python-check-blanket-noqa
      - id: python-check-blanket-type-ignore
        exclude: ^src/vector/backends/_numba_object.py$
      - id: python-no-log-warn
      - id: python-no-eval
      - id: python-use-type-annotations
      - id: rst-backticks
      - id: rst-directive-colons
      - id: rst-inline-touching-normal
```

Saransh-cpp:
We can further filter out the `pre-commit` hooks as per our needs!

priyanshuone6:
Thanks, I was having trouble pushing black changes to the pull request or fork. Pre commit should make it so much more convenient to carry out these.

priyanshuone6:
I've put up a request to install the pre-commit ci bot on pybamm-team/PyBaMM to automatically commit on the PR, making the required changes.

Saransh-cpp (changes requested):
Thanks, @priyanshuone6!

Some comments -
- We now won't be needing - 

https://github.com/pybamm-team/PyBaMM/blob/8c4b3d60d5980748ba8a673be50aa140eb57cb63/.github/workflows/test_on_push.yml#L27-L41

---

- I think there should be a note in `CONTRIBUTING.md` directing users how to use `pre-commit`. The [Pre commit checks](https://github.com/pybamm-team/PyBaMM/blob/develop/CONTRIBUTING.md#pre-commit-checks) section can also be modified to use `pre-commit` instead of `flake8` directly! An extra section can be also be added, saying something like this -

## Installing and using pre-commit

`PyBaMM` uses a set of `pre-commit` hooks and the `pre-commit` bot to format and prettify the codebase. The hooks can be installed locally using -

```bash
pip install pre-commit
pre-commit install
```

This would run the checks every time a commit is created locally. The checks will only run on the files modified by that commit, but the checks can be triggered for all the files using -

```bash
pre-commit run --all-files
```

If you would like to skip the failing checks and push the code for further discussion, use the `--no-verify` option with `git commit`.

---

- Some more useful `pre-commit`s that can be added -
```yaml
  - repo: https://github.com/asottile/pyupgrade
    rev: v2.37.3
    hooks:
      - id: pyupgrade
        args: [--py37-plus]

  - repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v4.3.0
    hooks:
      - id: check-added-large-files
      - id: check-case-conflict
      - id: check-merge-conflict
      - id: check-symlinks
      - id: check-yaml
      - id: debug-statements
      - id: end-of-file-fixer
        exclude: ^docs
      - id: mixed-line-ending
      - id: requirements-txt-fixer
      - id: trailing-whitespace

  - repo: https://github.com/PyCQA/isort
    rev: 5.10.1
    hooks:
      - id: isort

  - repo: https://github.com/pre-commit/mirrors-prettier
    rev: "v3.0.0-alpha.0"
    hooks:
      - id: prettier
        types_or: [yaml, markdown, html, css, scss, javascript, json]
        exclude: assets/js/webapp\.js
```

---

- Lastly, I think `pre-commit run --all-files` should be run once on this PR. This would bring the codebase in sync with the `pre-commit` hooks!

.github/workflows/test_on_push.yml
```diff
@@ -24,22 +24,6 @@ jobs:
...
-      with:
-        python-version: 3.7
-
-    - name: Check style
-      run: |
-        python -m pip install tox
-        tox -e flake8
-
   build:
     needs: style
```
tinosulzer, line 28: This needs to be changed now that `style` is deleted

tinosulzer (commented):
- Can you explain the order in which the workflows are run (and depend on each other)?
- What happens if someone doesn't have pre-commit installed locally?
- What happens to errors triggered by `flake8` that can't be fixed automatically by `black` (e.g. undefined variable error)?

priyanshuone6:
Both pre-commit ci and test_on_push will run individually on each PR. The pre-commit CI will take care of formatting the code and will push the formatted code directly to the PR regardless of whether pre-commit is installed locally (it will skip pushing if the code is already formatted and the pre-commit CI tests pass). For some reason, if black can't fix any errors as you mentioned, the flake8 test in pre-commit CI, which is put at the last in the config, will fail, which will fail the whole pre-commit CI run.

Saransh-cpp:
For some reason the pre-commit bot is not working. Should I give it the required access? The access can be provided from - https://results.pre-commit.ci/

tinosulzer:
> Both pre-commit ci and test_on_push will run individually on each PR. The pre-commit CI will take care of formatting the code and will push the formatted code directly to the PR regardless of whether pre-commit is installed locally (it will skip pushing if the code is already formatted and the pre-commit CI tests pass). 

Sounds good

> For some reason, if black can't fix any errors as you mentioned, the flake8 test in pre-commit CI, which is put at the last in the config, will fail, which will fail the whole pre-commit CI run.

Ok, let's make sure that the other tests don't get run unnecessarily if flake8 fails

priyanshuone6:
pre-commit.ci run

priyanshuone6:
pre-commit.ci run

.pre-commit-config.yaml
```diff
@@ -0,0 +1,19 @@
+ci:
+  autoupdate_commit_msg: "chore: update pre-commit hooks"
+  autofix_commit_msg: "style: pre-commit fixes"
+
+repos:
+  - repo: https://github.com/PyCQA/isort
+    rev: 5.10.1
+    hooks:
+      - id: isort
```
Saransh-cpp, line 9:
```suggestion
      - id: isort
        args: [".", "--skip", "pybamm/__init__.py"]
```

Turns out, the order of imports in `pybamm`'s `__init__.py` file matters (not sure why). Also the changed made to `__init__.py` should be reverted.

Saransh-cpp (changes requested):
The bot is working now, nice! One final comment -

.pre-commit-config.yaml
tinosulzer: Order matters because the expression tree classes are defined in a hierarchy, so we have to define `pybamm.Symbol` before any others. Is there a better way to do this?

Saransh-cpp: There is a way to ignore specific import sections of a file -
```py
import a
import b
# isort: off
import x
import y
# isort: on
import f
import g
```
(ignores `x` and `y` imports)

I tried this locally, but the order of imports is also required for battery models and geometry (maybe even for other imports). We can get rid of `args` here by adding `# isort: skip_file` to the top of `__init__.py`. Not sure what is the best thing to here.

tinosulzer: Can we just skip isort altogether? Why is it needed?

Saransh-cpp: I think `isort` was added through my suggestions, as it is a standard pre-commit hook 😅. Can be removed if not needed!

priyanshuone6:
> Ok, let's make sure that the other tests don't get run unnecessarily if flake8 fails 

I found no way to link a GitHub app with a workflow and run workflows only if the app's CI passes; something like `needs` in the workflow but for apps.

tinosulzer:
Will the pre-commit app always run before the workflows? i.e. does the test_on_push.yml workflow get run on the code after the pre-commit run has been done?

priyanshuone6:
They both run at the same time. Both will function separately and independently of one another. The pre-commit and PyBaMM's tests will be rerun for the new commit if the pre-commit ci pushes any additional modifications.

Saransh-cpp:
We can do something like this in the workflows - 

```yaml
jobs:
  pre-commit:
    runs-on: ubuntu-latest
    name: Pre-commit
    steps:
      - uses: actions/checkout@v3
      - uses: actions/setup-python@v4
        with:
          python-version: 3.x
      - uses: pre-commit/action@v3.0.0
```

This would run pre-commit twice, once in the workflows (which cannot trigger the bot, but can be used as a dependency for the main tests) and once in the `pre-commit` CI (which can trigger the bot and automatically fix the PRs; fixing the PRs will trigger the workflow again, making sure the tests run).

Both of these pre-commit jobs will be running on different CI providers; hence, we won't be blocked as they will run parallely.

tinosulzer:
> This would run pre-commit twice, once in the workflows (which cannot trigger the bot, but can be used as a dependency for the main tests) and once in the pre-commit CI (which can trigger the bot and automatically fix the PRs; fixing the PRs will trigger the workflow again, making sure the tests run).

Why not just run it once in the workflow in that case? Sorry if I am being stupid

priyanshuone6:
Running pre-commit in just the workflow will only pass or fail the style tests and not push the formatting changes to the PR. Should I go ahead with this?

tinosulzer:
Ok, go for it

tinosulzer (approved):

tinosulzer:
Ready to merge once conflicts are resolved

tinosulzer:
Will also need approval from @Saransh-cpp

tinosulzer:
Pre-commit checks are failiing but the other actions still run: https://github.com/pybamm-team/PyBaMM/pull/2342

Can we check that pre-commit was successful before we run the other actions?

Saransh-cpp:
We can use this - https://github.com/pybamm-team/PyBaMM/pull/2272#issuecomment-1253927026. Running the `pre-commit` hooks in `pre-commit.ci` enables the bot to fix black related issues, but we cannot link `pre-commit.ci` with GitHub Actions.

tinosulzer:
Could we just put the flake8 style check back into the test_on_push action?

PR #2278, opened 2022-09-04
remove old scripts

tinosulzer:
# Description

These scripts have not been updated in 3 years and don't seem to be used anywhere. I think they can safely be removed?


Saransh-cpp (approved):
Thanks, @tinosulzer!

PR #2348, opened 2022-10-06
Explicitly integrate variables which appear nowhere else on the rhs

### LINKED ISSUE #2240: Separate differential states that can be explicitly integrated
In all of the battery models, there is the state `Q=Variable("Discharge capacity [A.h]")`, which is just the integral of the current. This variable does not appear in any of the rhs or algebraic equations. 

This might be making the problem unnecessarily stiff. In particular, this is causing problems for the "Rest" steps, where dQ/dt = 0 which makes the first line of the Jacobian all zero and hence the Jacobian singular.

We should not be solving for Q at the same time as the rest of the model, since we can just do a post-processing step where we integrate the current. However, we still want to be able to access Q just like the other variables with `solution["Discharge capacity [A.h]"]`. I can think of two approaches:

1. Do this implicitly behind the scenes. When we discretize, notice that `Q` does not appear anywhere in the rhs or algebraic equations and moving it to a different dictionary `rhs_explicit_integration`
2. Do this explicitly by putting `Q` in a different dictionary `rhs_explicit_integration` when creating the model

Then the `rhs_explicit_integration` dictionary can be handled by the solver after the other variables have been solved for, e.g. using `cumtrapz`. Ideally, this wouldn't even be done until the user calls `solution["Discharge capacity [A.h]"]` (i.e. the `Solution` object handles it, maybe by adding some functionality to `ProcessedVariable`)

abillscmu:
# Description

This change moves variables from the model which do not appear on the right hand side to the variables dictionary with a `pybamm.ExplicitTimeIntegral` wrapper. When called by the solution object, the `pybamm.ExplicitTimeIntegral` integrates the variable (which can be easily explicitly calculated from the solution object).

Variables are **only** moved if **all** of the following is true

- the variable in question does not appear in the calculation of any other variable in the algebraic equations, the rhs, or the boundary conditions
- the variable does not appear in `y_slices` (needed for experiments in the case that the variable is used in one experiment step and not others).
- the variable is 0 dimensional (`len(var.domains)==0`)
- the variable has not already been discretized.
- there is at least one more variable on the rhs

Issue #2240 provides a brief overview of the impetus for the PR, but the main point is that in the case where the model has a rest step, not doing this results in a singular Jacobian matrix (due to `Discharge capacity [A.h]`, which can cause slowdowns and failures in some solvers.

Fixes #2240 

## Type of change

# Key checklist:

## Further checks:


tinosulzer (changes requested):
Code changes look good, some style comments. Let's see what the tests say. Could maybe do with a couple more tests (e.g. testing the output of an explicit integration - unless that is now done somewhere else by one of the existing tests?)

pybamm/solvers/solution.py
```diff
@@ -471,16 +471,43 @@ def update(self, variables):
...
+                    # Convert variable to casadi
+                    # Make all inputs symbolic first for converting to casadi
+                    var_sym = var_pybamm.to_casadi(t_MX, y_MX, inputs=inputs_MX_dict)
+
+                    var_casadi = casadi.Function(
+                        "variable", [t_MX, y_MX, inputs_MX], [var_sym]
+                    )
+                    model._variables_casadi[key] = var_casadi
+                    vars_pybamm[i] = var_pybamm
+                elif key in model._variables_casadi:
```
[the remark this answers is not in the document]
abillscmu: Problem is that if that if statement is first, the thing returned will be incorrect, since the `cumtrapz` is located inside `ProcessedVariable`, the thing stored in `model._variables_casadi` is the integrand. Shouldn't `__getitem__` prevent this from being processed every time? 

But can fix duplication issue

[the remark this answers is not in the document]
tinosulzer: Oh yeah you're right, `__getitem__` will take care of it

tinosulzer (approved):
Thanks, looks good (once tests pass)

tinosulzer:
Just had a thought, it might be worth special-casing the case where the rhs is `Scalar(0)`, and in that case simply replacing the variable with its initial condition (instead of the `ExplicitTimeIntegral`)

pybamm/discretisations/discretisation.py
```diff
@@ -1197,3 +1200,85 @@ def check_variables(self, model):
...
+        for tree in rhs_variables:
+            pybamm.tree_search(model.rhs[tree], var, this_var_list)
+        for tree in algebraic_variables:
+            pybamm.tree_search(model.algebraic[tree], var, this_var_list)
+        for (keys, tree) in zip(boundary_variable_keys, boundary_variables):
+            for key in keys:
+                pybamm.tree_search(
+                    model.boundary_conditions[tree][key][0], var, this_var_list
+                )
+        for tree in model.variables.keys():
```
tinosulzer, line 1238: minor comment, `tree` here will be a string, not an expression tree, so `name` would be more appropriate

pybamm/discretisations/discretisation.py
```diff
@@ -1197,3 +1200,85 @@ def check_variables(self, model):
...
+                if len(model.algebraic) != 1:
+                    pybamm.logger.info(
+                        "removing variable {} from algebraic".format(var)
+                    )
+                    my_initial_condition = model.initial_conditions[var]
+                    model.variables[var.name] = model.algebraic[var]
+                    del model.algebraic[var]
+                    del model.initial_conditions[var]
+                else:
+                    break
```
tinosulzer, line 1283: We shouldn't do this for algebraic variables. The algebraic dictionary is `0=...`, *not* `key=value` like it looks like you are assuming here. For example, `0=div(sigma*grad(phi_s)) + a*j` for Ohm's law. The key in the dictionary defines which variable that equation "solves" for, not equality between the variable and the equation.

pybamm/discretisations/discretisation.py
```diff
@@ -1197,3 +1200,85 @@ def check_variables(self, model):
...
+            model, this_var_is_independent = self.search_for_independent_var(model, var)
+            if this_var_is_independent:
+                if len(model.rhs) != 1:
+                    pybamm.logger.info("removing variable {} from rhs".format(var))
+                    my_initial_condition = model.initial_conditions[var]
+                    model.variables[var.name] = pybamm.ExplicitTimeIntegral(
+                        model.rhs[var], my_initial_condition
+                    )
+                    # edge case where a variable appears
+                    # in the rhs twice under different names
```
tinosulzer, line 1262: typo? `in the rhs` -> `in the variables`

pybamm/discretisations/discretisation.py
```diff
@@ -1197,3 +1200,85 @@ def check_variables(self, model):
                             var.shape, model.rhs[rhs_var].shape, var
                         )
                     )
+
+    def search_for_independent_var(self, model, var):
+        pybamm.logger.info("Removing independent blocks.")
```
tinosulzer, line 1205: make this a verbose, not an info

tinosulzer (commented):
A few more comments from looking more closely ...

tinosulzer:
@all-contributors add @abillscmu for code

PR #2350, opened 2022-10-06
Return solution if solver fails early

tinosulzer:
# Description

Add option to return the solution if the solver manages to take some steps but fails early. This is only applicable for the casadi solver with safe mode.

Fixes #2035 

## Type of change

# Key checklist:

## Further checks:


pybamm/solvers/casadi_solver.py
```diff
@@ -222,21 +230,41 @@ def _integrate(self, model, t_eval, inputs_dict=None):
...
                         solved = True
-                    except pybamm.SolverError:
+                    except pybamm.SolverError as error:
                         dt /= 2
-                        # also reduce maximum step size for future global steps
-                        dt_max = dt
+                        # also reduce maximum step size for future global steps,
+                        # but skip them in the beginning
+                        # sometimes, for the first integrator smaller timesteps are
+                        # neeeded, but this won't affect the global timesteps. The
```
rtimms, line 240: neeeeeeeeeeeeeeeded

rtimms (approved):
thanks, looks good!

pybamm/solvers/casadi_solver.py
tinosulzer, line 240:
```suggestion
                        # needed, but this won't affect the global timesteps. The
```

PR #2356, opened 2022-10-11
add perturbation to initial conditions in casadi solver

tinosulzer:
# Description

For some reason this fixed a bug in the casadi solver for one particular case I was looking at. Opening PR to check tests and benchmarks.

## Type of change

# Key checklist:

## Further checks:


pybamm/solvers/casadi_solver.py
```diff
@@ -658,14 +663,22 @@ def _run_integrator(
...
+            len_alg = len_alg * (num_parameters + 1)
 
         y0_diff = y0[:len_rhs]
         y0_alg = y0[len_rhs:]
+        if self.perturb_algebraic_initial_conditions and len_alg > 0:
+            # Add a tiny perturbation to the algebraic initial conditions
+            # For some reason this helps with convergence
+            # The actual value of the initial conditions for the algebraic variables
+            # doesn't matter
+            y0_alg = y0_alg * (1 + 1e-6 * casadi.DM(np.random.rand(len_alg)))
```
brosaplanella, line 681: Do we want to allow users to fiddle with this value or is it very sensitive?

brosaplanella (approved):
Looks good, thanks!

pybamm/solvers/casadi_solver.py
tinosulzer: It's not that sensitive. It's just the first thing I tried and it worked for the specific case I was testing. We can see if other cases come up whether changing that number makes a difference

PR #2375, opened 2022-10-17
Add option for initial soc in build

### LINKED ISSUE #2372: Move initial soc out of solve
### Description

It would be nice if it was in its own method, which was then called by either solve or build.

### Motivation

Necessary to get initial soc for Julia solvers

### Possible Implementation

just move it to its own method, then call that method in build/solve

### Additional context

_No response_

abillscmu:
- add option in build to initialize soc
- remove added dependency
- add test

# Description

move initial soc setting to its own function, which can be called from build
Fixes #2372 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/simulation.py
+++ b/pybamm/simulation.py
@@ -1,7 +1,6 @@
 #
 # Simulation class
 #
-from calendar import c
 import pickle
 import pybamm
 import numpy as np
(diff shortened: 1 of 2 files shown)


tinosulzer (approved):
Looks good to me

PR #2381, opened 2022-10-19
Fix typo

agriyakhetarpal:
# Description

Fixes a minor typo in the [docs](https://pybamm.readthedocs.io/en/latest/install/install-from-source.html#running-the-tests)

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/docs/install/install-from-source.rst
+++ b/docs/install/install-from-source.rst
@@ -177,7 +177,7 @@ The default command
 will run the full test suite (integration and unit tests).
 This can take several minutes.
 
-Is is often sufficient to run the unit tests only. To do so, use
+It is often sufficient to run the unit tests only. To do so, use
 
    .. code:: bash
 


Saransh-cpp (commented):
Maybe we can add a pre-commit to detect typos? Something like this -
```yaml
  - repo: https://github.com/codespell-project/codespell
    rev: v2.2.2
    hooks:
      - id: codespell
        args: ["-L", "{add false positives here separated by comma - pybamm,dfn,spm,...}"]
```

agriyakhetarpal:
I'll be glad to add it, @Saransh-cpp! Would you prefer it as a separate PR?

tinosulzer:
Would the spell check have caught this case ("is" is a valid word)? I think it might be overkill with way too many exceptions

agriyakhetarpal:
That's right, it's more of a grammatical error rather than a typo – `codespell` has the potential to cause some problems with triggering certain function names in [some](https://github.com/matplotlib/matplotlib/issues/22740/) repositories

Saransh-cpp:
> codespell has the potential to cause https://github.com/matplotlib/matplotlib/issues/22740/ problems with triggering certain function names in some repositories

I don't understand the reference here (the issue talks about adding `codespell` to `matplotlib`?), but yes, as @tinosulzer mentioned, `codespell` might be overkill (and we might end up maintaining a long list of false positives).

Thanks for fixing this! 😄

tinosulzer:
@all-contributors add @agriyakhetarpal for infra

PR #2402, opened 2022-10-25
fix: pybamm.have_julia is True when PyCall.jl uninstalled

awadell1:
Currently `pybamm.have_julia()` returns True even if PyCall.jl is not configured this results in test failures as the tests assume they can call:

```python
from julia.api import Julia
Julia()
```

Which fails if Julia has not been configured. This commit fixes that.

## Type of change

# Key checklist:

- [] No style issues: `$ flake8`

## Further checks:


tinosulzer (approved):
Thanks, will probably need to update this soon / coordinate with @abillscmu as we switch to pythoncall

abillscmu:
I'll add a test but this should be obsolete when #2363 is merged.

PR #2425, opened 2022-11-03
Asymmetric SEI reaction

### LINKED ISSUE #2338: SEI transfer coefficient
The SEI transfer coefficient for `reaction limited` and `EC reaction limited` have the transfer coefficient hardcoded to 0.5. This value is a modelling assumption (see [Safari et al. 2009](https://iopscience.iop.org/article/10.1149/1.3043429/pdf)), so I think we should allow to define it in the parameter values so it can be changed.

brosaplanella:
# Description

Allow for asymmetric SEI reactions, in the cases of `"reaction limited"` and `"ec reaction limited"`. The option is set to `"reaction limited (symmetric)"` and `"reaction limited (asymmetric)"` (same for EC).

Fixes #2338

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ## Features
 
+- SEI reactions can now be asymmetric ([#2425](https://github.com/pybamm-team/PyBaMM/pull/2425))
 - Third-party parameter sets can be added by registering entry points to `pybamm_parameter_set` ([#2396](https://github.com/pybamm-team/PyBaMM/pull/2396))
 - Added three-dimensional interpolation ([#2380](https://github.com/pybamm-team/PyBaMM/pull/2380))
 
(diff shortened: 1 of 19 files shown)


tinosulzer (commented):
Changes look good, I think the symmetric option name shouldn't change though, since it's a breaking change

brosaplanella:
> Changes look good, I think the symmetric option name shouldn't change though, since it's a breaking change

Which should it be?
1. `"reaction limited"` and `"reaction limited (asymmetric)"`
2. `"reaction limited"`, `"reaction limited (symmetric)"` and `"reaction limited (asymmetric)"`, where the first two are equivalent

tinosulzer:
Option 1

tinosulzer (approved):
Thanks, looks good!

Issue #2346, opened 2022-10-05
[Bug]: Installing `PyBaMM` on `M1` chip fails (both `pip` and install from source)


Saransh-cpp:
### PyBaMM Version

all

### Python Version

3.9

### Describe the bug

`PyBaMM` cannot be installed on an `M1` mac from `pip` as well as from source. I think `casadi` [does not support](https://github.com/casadi/casadi/issues/2829) `M1` chips?

I was able to set everything up by installing `casadi` from source, and removing `casadi` as a dependency when installing `PyBaMM` from source. Should this be documented in the installation page?

Also, should we be specifically building `intel` wheels till `casadi` supports `M1` wheels? This way users will get a "No matching distribution found for pybamm" error when installing through `pip` on `M1`, rather than than the installation failing midway.

Building casadi locally -
```
brew install cmake swig
git clone https://github.com/casadi/casadi.git casadi
cd casadi
git checkout 3.5.5
mkdir build
cd build
cmake -DWITH_PYTHON=ON \
-DWITH_PYTHON3=ON \
-DPYTHON_INCLUDE_DIR=$(python3 -c "from distutils.sysconfig import get_python_inc; print(get_python_inc())") \
-DPYTHON_LIBRARY=$(python3 -c "import distutils.sysconfig as sysconfig; print(sysconfig.get_config_var('LIBDIR'))") \
..
make
sudo make install
```

### Steps to Reproduce

```
pip3 install pybamm
```

```
pip3 install -e .
```

### Relevant log output

```shell
ERROR: Could not find a version that satisfies the requirement casadi>=3.5.0 (from pybamm) (from versions: none)
ERROR: No matching distribution found for casadi>=3.5.0
```

tinosulzer:
I'm able to install pybamm on M1 from source, but maybe it's using Rosetta?

Saransh-cpp:
I think so. #2072 looks related (`casadi` not found), the only difference is that I did not try building up the `IDAKLUSolver`.

abillscmu:
You can force it to use rosetta if you're using conda by setting the environment variable `CONDA_SUBDIR` to `osx-x64`.

savvn001:
I managed to get it to install and sort of work.
Using an M1 MBA (2020):

- Download and install Python 3.8.10 **universal** installer (maybe 3.9 works too, I haven't tried)
[...]

PR #2435, opened 2022-11-07
Support Python 3.10 and 3.11

Saransh-cpp:
# Description

Casadi has published a couple of pre-releases with Python 3.10 support - https://pypi.org/project/casadi/3.5.6.pre2/#files!!

>**Warning**
> This PR is meant to test PyBaMM's support on Python 3.10. This should not be merged until casadi releases a "non-pre" release.

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:


tox.ini
```diff
@@ -2,6 +2,7 @@
 envlist = {windows}-{tests,unit,dev},tests,unit,dev,julia
 
 [testenv]
+pip_pre = true
```
Saransh-cpp, line 5: More specifically, before merging this PR we would need to remove this, and this can be removed once casadi publishes a "non-pre" release.

.github/workflows/test_on_push.yml
```diff
@@ -46,7 +46,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: [3.8, 3.9]
+        python-version: ["3.8", "3.9", "3.10"]
```
brosaplanella, line 49: Shall we also try Python 3.11?

Saransh-cpp:
Also, should there be a CHANGELOG entry for #2346? Saying something like - "PyBaMM is now natively supported on Apple silicon chips"?

brosaplanella (approved):
Looks great, thanks! Yes, I think we should add a line in CHANGELOG about Mac.

tinosulzer (approved):
Thanks! merging

qnovo-tissa:
Hi there, Considering that the support for Python 3.10 is listed here as "unreleased", would that mean that I cannot currently use PyBamm with Python 3.10? If I can, please help explain how to do so. If I cannot, could you please let me know approximately when the support for 3.10 would be released?
Thanks,

brosaplanella:
At the moment it is only available if you install from source, but we aim to make a release in the next few days.

Issue #2418, opened 2022-11-01
Dimensional models


tinosulzer:
Some basic profiling reveals that implementing the models in a dimensional way provides no slow-down compared to dimensionless. Since dimensionless models are a big barrier to entry, we should reformat the models to be entirely dimensional. Further benchmarking and testing will be needed.

Note: scaling variables is still a good idea for the condition number of the solver, but this can be done directly in the solver

tinosulzer:
We currently have a lot of duplicate variables (dimensionless/dimensional).
Q: Should we just get rid of the dimensionless ones?

tinosulzer:
Would be good to revive #406 when implementing this to add an additional layer of testing

brosaplanella:
> We currently have a lot of duplicate variables (dimensionless/dimensional). Q: Should we just get rid of the dimensionless ones?

I think so, I would just keep those that have actually some sort of physical meaning beyond being the dimensionless form of something (e.g. dimensionless particle concentrations / stoichiometry).

PR #2440, opened 2022-11-09
Variable scale and reference

tinosulzer:
# Description

Adds `scale` and `reference` attributes to a variable, which are used to multiply and/or add a value to a variable to make the resulting numerics better conditioned.
Progress towards #2418 

## Type of change

# Key checklist:

## Further checks:


pybamm/discretisations/discretisation.py
```diff
@@ -739,6 +741,9 @@ def process_dict(self, var_eqn_dict):
             Equations ({variable: equation} dict) to dicretise
             (can be model.rhs, model.algebraic, model.initial_conditions or
             model.variables)
+        ics : bool, optional
+            Whether the equations are initial conditions. If True, the equations are
```
brosaplanella, line 745:
```suggestion
            Whether the equations have initial conditions. If True, the equations are
```

pybamm/parameters/lead_acid_parameters.py
```diff
@@ -422,7 +422,7 @@ def kappa_e(self, c_e, T):
         kappa_scale = self.F**2 * self.D_e_typ * self.c_e_typ / (self.R * self.T_ref)
         return self.kappa_e_dimensional(c_e_dimensional, self.T_ref) / kappa_scale
 
-    def chiT_over_c(self, c_e, T):
+    def chiRT_over_Fc(self, c_e, T):
```
brosaplanella, line 425: What's the meaning of Fc?

brosaplanella (approved):
Looks good, but the dimensional DFN is not there yet, is it?

pybamm/parameters/lead_acid_parameters.py
tinosulzer: Faraday's constant times c

tinosulzer:
I removed the dimensional DFN since when we eventually dimensionalize all the models it will not be there, so I don't want people adding it to their code. This PR now only adds the scaling and reference

pybamm/discretisations/discretisation.py
tinosulzer: This is meant to be `are`, this is the function that discretises the equations (rhs, algebraic, initial conditions) and there is a special case for initial conditions (reference only gets taken away from initial conditions)

Issue #2217, opened 2022-08-05
Improvements to IDAKLU solver


martinjrobins:
### Description

Many things still left to do on the IDAKLU solver, here is the copied list from #1863 

- [ ] we're allocating memory for return vectors in the solve, this could be done in solver setup
- [x] we initialise ida_mem and everything every time, could do this once when solver is created, or only on the 1st solve etc
- [x] create an options dict to pass to the solver, e.g. option to print solver stats,  don't/do use jacobian
- [x] add option to use iterative solvers
- [x] add option to use other linear solvers in ida
- [x] add option to use dense jacobian
- [ ] add option to use banded solvers
- [ ] benchmarks for all of the above solver options
- [ ] investigate if we need to densify the rhs_algebraic casadi evaluation, could it return a sparse vector?
- [ ] investigate if constructing a new casadi function for the residual (rather than doing it manually in c++) is faster
- [ ] investigate if constructing a new casadi function for the sensitivity calculations (rather than doing it manually in c++) is faster
- [ ] we calculate consistent initial conditions at the start of the idklu solve, so we might not need to do it in the BaseSolver setup
- [x] Sundials v6 has a slightly different API using a context object, need to update code to support both this and version 5

### Motivation

1. The IDAKLU solver can be faster than casadi, so we want to expose more functionality from sundials
2. Make the solver more efficient

martinjrobins:
"we're allocating memory for return vectors in the solve, this could be done in solver setup"
 Need to think more on this one, if we re-use this memory for subsequent solves will overwrite solution returned to user :(

tinosulzer:
There are still scenarios where that's ok - e.g. parameter estimation - so it could be an optional feature?

tinosulzer:
Been reading a bit more about Sundials. Super excited to see these changes when they're ready! For pack modeling it could be extremely useful

tinosulzer:
Are we now in a place where we can make the IDAKLU solver the default if it's installed (and fall back to casadi otherwise)?
Would be interesting to test/benchmark

martinjrobins:
Re benchmarks, it looks like the main advantage of the idaklu solver is for running experiments:

```
 ============ ============= ======================================================= =========================================== ===========================================
              --                                                                                                                       solver_class                                     
              ---------------------------------------------------------------------------------- ---------------------------------------------------------------------------------------
[...]

PR #2444, opened 2022-11-09
I2217 idaklu

martinjrobins:
# Description

Work related to #2217. Added an `options` argument to the idaklu solver:

```
    options: dict, optional
        Addititional options to pass to the solver, by default:
        {
            print_stats: False, # print statistics of the solver after every solve
            jacobian: "sparse", # jacobian form, can be "none", "dense", "sparse",
                                # "matrix-free"
            linear_solver: "SUNLinSol_KLU", # name of sundials linear solver to use
                                  # options are: "SUNLinSol_KLU",
                                  # "SUNLinSol_Dense", "SUNLinSol_LapackDense"
                                  # "SUNLinSol_SPBCGS", "SUNLinSol_SPFGMR",
                                  # "SUNLinSol_SPGMR", "SUNLinSol_SPTFQMR",
            preconditioner: "BBDP", # preconditioner for iterative solvers,
                                    # can be "none", "BBDP"
            linsol_max_iterations: 5, # for iterative linear solvers, max number of
                                      # iterations
            precon_half_bandwidth: 5, # for iterative linear solver preconditioner,
                                      # bandwidth of approximate jacobian
            precon_half_bandwidth_keep: 5 #for iterative linear solver preconditioner,
                                      # bandwidth of approximate jacobian that is kept

        }
        Note: These options only have an effect if model.convert_to_format == 'casadi'
```

The iterative solvers only seem to work well for SPM and SPMe models, my best guess atm is that the preconditioner I'm using is not suitable for DFN. Sundials unfortunately only has a single preconditioner (BBDP), but we could manually implement a preconditioner if anyone knows a good one for DFN

I've also built in support for sundials v5 and v6, as the api changed in v6

I've also generally restructured the idaklu C++ code (the casadi stuff, we are going to drop the python idaklu solver...) to be a bit nicer

## Type of change

# Key checklist:

## Further checks:


tinosulzer:
> we are going to drop the python idaklu solver

What does this mean? How will we interface to the C++ code?

tests/unit/test_solvers/test_idaklu_solver.py
```diff
@@ -54,7 +58,7 @@ def test_ida_roberts_klu(self):
             np.testing.assert_array_almost_equal(solution.y[0, :], true_solution)
 
     def test_model_events(self):
-        for form in ["python", "casadi", "jax"]:
+        for form in ["casadi", "python", "casadi", "jax"]:
```
rtimms, line 61: casadi twice here

tests/unit/test_solvers/test_idaklu_solver.py
```diff
@@ -443,6 +447,57 @@ def test_dae_solver_algebraic_model(self):
...
+        self.assertIn("Solver Stats", s)
+
+        solver = pybamm.IDAKLUSolver(options={"print_stats": False})
+        f = io.StringIO()
+        with redirect_stdout(f):
+            solver.solve(model, t_eval)
+        s = f.getvalue()
+        self.assertEqual(len(s), 0)
+
+        # test everything else
```
rtimms, line 479: should "garbage" throw an error?

rtimms (approved):
thanks @martinjrobins looks good! should we do some checking of the options since it seems sundials doesn't raise and error if you pass a bad option for the jacobian or lin solver?

tests/unit/test_solvers/test_idaklu_solver.py
martinjrobins: perhaps...? I went with the option of just using a sensible default if a bad option was passed in, while printing a warning (e.g. if you choose jacobian=sparse and linear_solver=dense it just changes the linear_solver to KLU and prints a warning). Happy to reconsider tho....

tinosulzer: I prefer erroring if inputs aren't compatible. In most cases it will probably be a typo, which is better caught by an error than a warning (e.g. if they have a low level logger on the warning might get lost). If a user has some weird workflow that means they have to pass in incompatible inputs, they can write their own input preprocessing functions

tests/unit/test_solvers/test_idaklu_solver.py
martinjrobins: fixed

martinjrobins:
Running the failed benchmark (time to setup DFN for ORegan2022 param) on my machine shows no difference between `develop` and this branch. @rtimms @tinosulzer are you happy if I treat this as a false positive and merge this PR?

PR #2465, opened 2022-11-17
Use `@cache` and `@cached_property` for memoization

### LINKED ISSUE #2288: Use `@cache` and `@cache_property` from `functools` for memoization
See https://docs.python.org/3/library/functools.html

We are currently doing this manually using try-except blocks with `AttributeError` ([example](https://github.com/pybamm-team/PyBaMM/blob/8c4b3d60d5980748ba8a673be50aa140eb57cb63/pybamm/expression_tree/symbol.py#L907-L916)). Using the decorators from functools would be cleaner

This is different functionality from using `@property` with a setter and getter - those should not be changed.

agriyakhetarpal:
# Description

Uses `@cache` and `@cached_property` via [functools](https://docs.python.org/3/library/functools.html) to implement memoization.  I shall add more commits to this PR for more changes in the remaining files.

Fixes #2288 

## Type of change

# Key checklist:

## Further checks:


agriyakhetarpal:
I assume flake8 seems to be failing because it does not recognise `@cache` since it was introduced in Python 3.9, though the `@cached_property` feature existed in Python 3.8 – [`functools`](https://docs.python.org/3/library/functools.html)

tinosulzer:
I think you have to import cache and cached property from functools. Did this work locally on your machine?

agriyakhetarpal:
Oh, that makes more sense. I had run `python run-tests.py --unit` and assumed the problem is with `flake8` – (my fault!), though importing `functools` within the changed files gives

> Ran 1193 tests in 3.718s
FAILED (failures=34, errors=811, skipped=59)

I'll push the changes in a moment

tinosulzer:
The `cached_property` methods should not define `._name`. They should just return the output of the constructor. i.e.
```python3
class DataSet:
    def __init__(self, sequence_of_numbers):
        self._data = sequence_of_numbers

    @property
    @cache
    def stdev(self):
        return statistics.stdev(self._data)
```
instead of 
```python3
class DataSet:
    def __init__(self, sequence_of_numbers):
        self._data = sequence_of_numbers

    @property
    @cache
    def stdev(self):
        self._stdev = statistics.stdev(self._data)
        return self._stdev
```
example is from https://docs.python.org/3/library/functools.html#functools.cached_property

agriyakhetarpal:
I happened to memoize functions one-by-one and have been able to identify specific functions that cause errors upon memoization (the others are working as necessary):

>Ran 1195 tests in 301.155s
FAILED (failures=1, skipped=60)

Would it be recommended to open a separate PR to clear out the mess or should I try and fix individual commits in this one?

pybamm/expression_tree/symbol.py
```diff
@@ -417,14 +419,12 @@ def scale(self):
...
-            return self._id == other._id
-        except AttributeError:
-            if isinstance(other, numbers.Number):
-                return self._id == pybamm.Scalar(other)._id
-            else:
-                return False
+        if isinstance(other, numbers.Number):
+            return self._id == pybamm.Scalar(other)._id
+        else:
+            return False
```
tinosulzer, line 427: this one should not be edited, the `AttributeError` is doing something different here (you can tell because there is no assignment happening)

pybamm/expression_tree/symbol.py
```diff
@@ -894,46 +892,40 @@ def new_copy(self):
...
         """
         Size of an object, found by evaluating it with appropriate t and y
         """
-        try:
-            return self._saved_size
-        except AttributeError:
-            self._saved_size = np.prod(self.shape)
-            return self._saved_size
+        self._saved_size = np.prod(self.shape)
+        return self._saved_size
```
tinosulzer, line 901: should just be `return np.prod(self.shape)`, see https://github.com/pybamm-team/PyBaMM/pull/2465#issuecomment-1319085195

pybamm/expression_tree/symbol.py
```diff
@@ -894,46 +892,40 @@ def new_copy(self):
...
+            evaluated_self = self.evaluate(0, y, y, inputs="shape test")
 
-            # Return shape of evaluated object
-            if isinstance(evaluated_self, numbers.Number):
-                self._saved_shape = ()
-            else:
-                self._saved_shape = evaluated_self.shape
+        # Return shape of evaluated object
+        if isinstance(evaluated_self, numbers.Number):
+            self._saved_shape = ()
```
tinosulzer, line 926: should just be `return ()`

pybamm/expression_tree/symbol.py
```diff
@@ -894,46 +892,40 @@ def new_copy(self):
...
+        else:
+            self._saved_shape = evaluated_self.shape
```
tinosulzer, line 928: should just be `return evaluated_self.shape`

tinosulzer (commented):
> I happened to memoize functions one-by-one and have been able to identify specific functions that cause errors upon memoization (the others are working as necessary):

I see 800 failing tests in this PR

> Would it be recommended to open a separate PR to clear out the mess or should I try and fix individual commits in this one?

Keep in same PR

> P.S. Can [this](https://github.com/pybamm-team/PyBaMM/blob/develop/pybamm/expression_tree/symbol.py#L830-L853) be memoized as well? I do not see any specific test failing upon doing so

Yes, it can

agriyakhetarpal:
Apologies, I closed the PR by error (😅) – I removed a commit too many

I'll reopen this with the new changes in a single commit

pybamm/models/full_battery_models/lithium_ion/electrode_soh.py
```diff
@@ -246,28 +247,22 @@ def __init__(self, parameter_values, param=None):
...
-        try:
-            return self._full_sim
-        except AttributeError:
-            full_model = ElectrodeSOH(param=self.param)
-            self._full_sim = pybamm.Simulation(
-                full_model, parameter_values=self.parameter_values
-            )
-            return self._full_sim
+        full_model = ElectrodeSOH(param=self.param)
+        self._full_sim = pybamm.Simulation(
```
tinosulzer, line 253: return here (don't assign full_sim)

pybamm/models/full_battery_models/lithium_ion/electrode_soh.py
```diff
@@ -246,28 +247,22 @@ def __init__(self, parameter_values, param=None):
...
-            )
-            x0_model = ElectrodeSOHx0(param=self.param)
-            x0_sim = pybamm.Simulation(x0_model, parameter_values=self.parameter_values)
-            self._split_sims = [x100_sim, x0_sim]
-            return self._split_sims
+        x100_model = ElectrodeSOHx100(param=self.param)
+        x100_sim = pybamm.Simulation(x100_model, parameter_values=self.parameter_values)
+        x0_model = ElectrodeSOHx0(param=self.param)
+        x0_sim = pybamm.Simulation(x0_model, parameter_values=self.parameter_values)
+        self._split_sims = [x100_sim, x0_sim]
```
tinosulzer, line 264: return here

tinosulzer (commented):
Looks good, try using `lru_cache` instead of cache for compatibility with python 3.8

agriyakhetarpal:
I saw that [functools](https://docs.python.org/3/library/functools.html) also describes using `@property` and `@cache` together to implement similar functionality, which can be used in cases wherein using `@cached_property` failed tests upon implementation; for example in this function [variables_and_events](https://github.com/pybamm-team/PyBaMM/blob/develop/pybamm/models/base_model.py#L194-L206)

Is this an approach that can be taken forward? This would again not be compatible with Python 3.8 however

tinosulzer:
It's not obvious to me why using `@cache` and `@property` would work when `@cached_property` doesn't.
I think `variables_and_events` is failing for a different reason (probably that the cache needs to be cleared at some point and this isn't happening if using `@cached_property`).

Happy to merge this as-is if it's ready

tinosulzer (approved):

agriyakhetarpal:
Great! I am not aware of any ways (yet) to measure if there is any performance increase by using `@cache` and `@property`, I suggested it because no tests failed 😃 I'm happy as it is – I'll push another commit to update the CHANGELOG

tinosulzer:
Some of the solvers fail inconsistently, we haven't yet been able to figure out why

tinosulzer:
@all-contributors add @agriyakhetarpal for code

PR #2475, opened 2022-11-18
Rename to entrypoint to `pybamm_parameter_sets`

awadell1:
Fixed a typo in docs on adding a new parameter set and added emphasis on installing/reinstalling the parameter set package.

Typo found via: https://github.com/pybamm-team/PyBaMM/discussions/2452


tinosulzer (approved):
Thanks!

PR #2486, opened 2022-11-25
Delete postBuild

tinosulzer:
File is not used

### DIFF
--- a/postBuild
+++ b/postBuild
@@ -1,1 +1,0 @@
-pip install .


Saransh-cpp (approved):
Thanks, @tinosulzer! I guess `build-wheel.sh` and `build-pure-wheel.sh` are also not being used anywhere?

tinosulzer:
Yes, I think we can remove those too

PR #2510, opened 2022-11-30
explicitly create interpolants from functions

### LINKED ISSUE #2410: Control automatic function parameter interpolation method
### Description

Allow users to specify which interpolation method they would like to be applied to function parameters. 

### Motivation

When using three-dimensional lookup tables I have found better model build times when moving from "cubic" to "linear" interpolation (in particular convert_to_casadi is slow). At the moment I can only control this by modifying the process symbol method of parameter values. 

The speed up might be due to differences in calculating the Jacobian (or perhaps not calculating it if the Jacobian at all is not defined everywhere for linear interpolation) but I am not sure of the exact reason. 

### Possible Implementation

I can think of a couple of ways of doing this but I'm not sure what is preferable for everyone. Both should be non-breaking. 

**1. Pass optional dictionary to parameter_values class** 
This would just involve adding an optional argument to the parameter_values class and then adding the input as an attribute of the class. The user would just pass an options dictionary in the following way: 
```python
options = {"interpolation methods": {"name of function parameter to target": "linear"} 
```
if not specified then "cubic" interpolation will be used so that this change is nonbreaking. 

**2. Allow users to add a third entry to the tuple within the parameter values dictionary** 
i.e. 
```python
options = {"interpolation method": "linear"}
parameter_values_dictionary = {
   "negative electrode open circuit voltage [V]": ("negative electrode open circuit voltage [V]", ( (x_data, ), y_data), options), 
   "positive electrode open circuit voltage [V]": ("positive electrode open circuit voltage [V]", ( (x_data, ), y_data)), 
}
```
This could either be a dictionary like I suggest here (so that additional control can be easily added later) or it could just be "linear" or "cubic". This third argument should be optional with "cubic" used by default so that we don't break all previous code. 

### Additional context

I am happy to implement once there is agreement on how to do so. 

Could perhaps be done in conjunction with modifications to function parameter in #2403

tinosulzer:
# Description

In the standard parameter sets, create the interpolants from functions explicitly (instead of relying on the `ParameterValues` class to do it). This allows finer control of hyperparameters (e.g. interpolator).

Using "cubic" interpolator for OCV gives much faster performance than linear interpolator (e.g. 1s -> 0.5s solve time for Ai2020 on my laptop)

Fixes #2410 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/examples/scripts/compare_lithium_ion.py
+++ b/examples/scripts/compare_lithium_ion.py
@@ -7,16 +7,18 @@ pybamm.set_logging_level("INFO")
 
 # load models
 models = [
-    pybamm.lithium_ion.SPM(),
-    pybamm.lithium_ion.SPMe(),
+    # pybamm.lithium_ion.SPM(),
+    # pybamm.lithium_ion.SPMe(),
     pybamm.lithium_ion.DFN(),
-    pybamm.lithium_ion.NewmanTobias(),
+    # pybamm.lithium_ion.NewmanTobias(),
 ]
 
 # create and run simulations
 sims = []
 for model in models:
-    sim = pybamm.Simulation(model)
+    sim = pybamm.Simulation(
+        model, parameter_values=pybamm.ParameterValues("NCA_Kim2011")
+    )
     sim.solve([0, 3600])
     sims.append(sim)
 
--- a/examples/scripts/compare_lithium_ion.py
+++ b/examples/scripts/compare_lithium_ion.py
@@ -7,18 +7,16 @@ pybamm.set_logging_level("INFO")
 
 # load models
 models = [
-    # pybamm.lithium_ion.SPM(),
-    # pybamm.lithium_ion.SPMe(),
+    pybamm.lithium_ion.SPM(),
+    pybamm.lithium_ion.SPMe(),
     pybamm.lithium_ion.DFN(),
-    # pybamm.lithium_ion.NewmanTobias(),
+    pybamm.lithium_ion.NewmanTobias(),
 ]
 
 # create and run simulations
 sims = []
 for model in models:
-    sim = pybamm.Simulation(
-        model, parameter_values=pybamm.ParameterValues("NCA_Kim2011")
-    )
+    sim = pybamm.Simulation(model)
     sim.solve([0, 3600])
     sims.append(sim)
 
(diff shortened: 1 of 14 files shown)


examples/scripts/compare_lithium_ion.py
```diff
@@ -7,16 +7,17 @@
 
 # load models
 models = [
-    pybamm.lithium_ion.SPM(),
-    pybamm.lithium_ion.SPMe(),
+    # pybamm.lithium_ion.SPM(),
+    # pybamm.lithium_ion.SPMe(),
```
Scottmar93, line 11: should these be uncommented?

pybamm/input/parameters/ecm/example_set.py
```diff
@@ -5,13 +5,38 @@
...
-dUdT = pybamm.parameters.process_2D_data_csv("ecm_example_dudt.csv", path=path)
+dUdT_data = pybamm.parameters.process_2D_data_csv("ecm_example_dudt.csv", path=path)
+
+
+def ocv(sto):
+    name, (x, y) = ocv_data
+    return pybamm.Interpolant(x, y, sto, name)
+
+
+def r0(T_cell, current, soc):
```
Scottmar93, line 22: A nice, this is better

Scottmar93 (approved):
Looks great cheers

tinosulzer:
ready for review (changed a bit since @Scottmar93 reviewed)

examples/scripts/compare_lithium_ion.py
brosaplanella: These are still commented

examples/scripts/compare_lithium_ion.py
```diff
@@ -7,16 +7,18 @@
...
-    pybamm.lithium_ion.NewmanTobias(),
+    # pybamm.lithium_ion.NewmanTobias(),
 ]
 
 # create and run simulations
 sims = []
 for model in models:
-    sim = pybamm.Simulation(model)
+    sim = pybamm.Simulation(
+        model, parameter_values=pybamm.ParameterValues("NCA_Kim2011")
```
brosaplanella, line 20: Do we want to go back to the default parameters (i.e. not prescribed)?

brosaplanella (approved):
Looks great, thanks Valentin! Just a couple of very minor comments.

PR #2514, opened 2022-12-02
Fix conflict main-develop

brosaplanella:
# Description

Fix conflict between `main` and `develop` to allow merging #2497


Saransh-cpp:
Re-running the failing `pre-job`. See https://github.com/fkirc/skip-duplicate-actions/issues/301#issuecomment-1334988536.

Saransh-cpp (approved):
Thanks, @brosaplanella! (`pre-job` looks fixed too!)

PR #2518, opened 2022-12-04
Issue 2358 add temperature to experiment

### LINKED ISSUE #2358: Add "temperature" argument to experiment
Simulation should then update the parameter values ("Ambient temperature [K]" and "Initial temperature [K]", not "Reference temperature [K]"), with the given temperature

Scottmar93:
# Description

Extension of the experiment class to allow control of the temperature. 

Input form is now: 
```python
     "Charge at C/3 for 1 hour at 25oC (1 second period)"
```

Open to suggestions on where best to have the temperature.

Also added a global temperature for the experiment and made some minor modifications to the ECM to ensure it is compatible with the temperature control.

Getting a minor error in the docs when building which appears to also be in develop: 
image

Fixes #2358 

## Type of change

# Key checklist:

## Further checks:


pybamm/experiments/experiment.py
```diff
@@ -221,23 +251,47 @@ def read_string(self, cond, drive_cycles):
...
                 time = self.convert_time_to_seconds(cond_list[idx_for + 1 : idx_until])
-                events = self.convert_electric(cond_list[idx_until + 2 :])
+
+                # remove temperture part of string
+                reduced_cond_list = cond_list[idx_until + 2 :]
+                if "at" in reduced_cond_list:
+                    at_idx = reduced_cond_list.index("at")
+                    reduced_cond_list = reduced_cond_list[:at_idx]
+
+                events = self.convert_electric(reduced_cond_list)
```
tinosulzer, line 265: this block of code looks like it's repeated 3 times, is this really the best way of doing this? If so, could it be a function?

I think you should be able to regex match `"at XoC"` where `X` is any number, record temperature, and remove that from the string for all cases

pybamm/simulation.py
```diff
@@ -358,6 +364,23 @@ def get_experiment_parameter_values(self, op):
...
+            if op_number == 0:
+                experiment_parameter_values.update(
+                    {
+                        "Initial temperature [K]": ambient_temperature,
+                    }
+                )
+        else:
+            experiment_parameter_values.update(
+                {"Ambient temperature [K]": self._original_temperature}
+            )
```
tinosulzer, line 382: add comment explaining why this needs to be done like this

tinosulzer (changes requested):
Looks good, thanks! Just some implementation comments

pybamm/experiments/experiment.py
tinosulzer: i.e. find it and then do `.replace("at 25oC", "")`

Scottmar93:
Making the changes for this today. Apologies for the delays

pybamm/simulation.py
Scottmar93: I've added: 

```
            # If at the first operation, then the initial temperature
            # should be the ambient temperature.
```

I thought this was the most intuitive behavior but I guess we don't absolutely have to have the cell start at the ambient temperature. Can either always start at 25oC or take whatever is in the parameter_values? What's your thoughts?

Scottmar93:
I reformatted the checking for the temperature condition a bit so that we just match a regex. If we don't find the regex for temperature then we raise a `pybamm.logger.warning` and use the globally set temperature or the temperature from the parameter values. 

Unfortunately, this does mean that cases typos in the temperature part of the string will only be raised as a warning but still pass through. I think this is the awkward side effect of not requiring the temperature to be in the string (for backwards compatibility). 
Will raise an error in a few cases though such as if "oC" is in the string but no match has been found.

tinosulzer (approved):
Thanks, looks good, I think this is the correct level of error-checking while still allowing back-compatibility. Can you update changelog?

Scottmar93:
Nice thanks, have updated

PR #2526, opened 2022-12-05
Fix colab and pybtex issue

### LINKED ISSUE #2524: Google Colab fails to install PyBaMM
### Discussed in https://github.com/pybamm-team/PyBaMM/discussions/2521
Google Colab has updated to Python 3.8 but now fails to install PyBaMM due to pybtex.

<sup>Originally posted by **reenubosepybamm** December  5, 2022</sup>
%pip install pybamm -q    # install PyBaMM if it is not installed
import pybamm
import numpy as np
GIVING AN ERROR
PluginNotFound                            Traceback (most recent call last)
[<ipython-input-5-103384696aab>](https://localhost:8080/#) in <module>
      2 get_ipython().run_line_magic('pip', 'install pybamm --no-cache -q')
      3 get_ipython().run_line_magic('pip', 'install pybamm -q    # install PyBaMM if it is not installed')
----> 4 import pybamm
      5 import numpy as np
      6 import matplotlib.pyplot as plt

6 frames
[/usr/local/lib/python3.8/dist-packages/pybtex/plugin/__init__.py](https://localhost:8080/#) in _load_entry_point(group, name, use_aliases)
     79         for entry_point in pkg_resources.iter_entry_points(search_group, name):
     80             return entry_point.load()
---> 81     raise PluginNotFound(group, name)
     82 
     83 

PluginNotFound: plugin pybtex.database.input.bibtex not found</div>

Saransh-cpp:
# Description

This looks like this right now -

image

Now that I think of it, it might be better to not show a warning on importing pybamm (would be too noisy). Rather, we could do -

```py
    def __init__(self):
        # Set of citation keys that have been registered
        self._papers_to_cite = set()

        # Dict mapping citations keys to BibTex entries
        self._all_citations: dict[str, str] = dict()

        if "google.colab" not in sys.modules:
            self.read_citations()
            self._reset()
```
and then -
```py
def print_citations(filename=None, output_format="text"):
    if "google.colab" in sys.modules:
        warnings.warn(
            """pybtex does not work with Google Colab due to a known bug -
            https://bitbucket.org/pybtex-devs/pybtex/issues/148/.
            Please manually cite all the references.""".replace(
                "\n                ", " "
            ),
        )
    else:
        pybamm.citations.print(filename, output_format)
```

Fixes #2524 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/citations.py
+++ b/pybamm/citations.py
@@ -5,6 +5,7 @@
 #
 import pybamm
 import os
+import sys
 import warnings
 import pybtex
 from pybtex.database import parse_file, parse_string, Entry
@@ -35,8 +36,18 @@ class Citations:
         # Dict mapping citations keys to BibTex entries
         self._all_citations: dict[str, str] = dict()
 
-        self.read_citations()
-        self._reset()
+        if "google.colab" in sys.modules:
+            warnings.warn(
+                """pybtex does not work with Google Colab due to a known bug -
+                https://bitbucket.org/pybtex-devs/pybtex/issues/148/.
+                Please manually cite all the references.""".replace(
+                    "\n                ", " "
+                ),
+                ImportWarning,
+            )
+        else:
+            self.read_citations()
+            self._reset()
 
     def _reset(self):
         """Reset citations to default only (only for testing purposes)"""
--- a/pybamm/citations.py
+++ b/pybamm/citations.py
@@ -147,11 +147,9 @@ def print_citations(filename=None, output_format="text"):
     """See :meth:`Citations.print`"""
     if "google.colab" in sys.modules:
         raise ImportWarning(
-            """pybtex does not work with Google Colab due to a known bug -
-            https://bitbucket.org/pybtex-devs/pybtex/issues/148/.
-            Please manually cite all the references.""".replace(
-                "\n            ", " "
-            ),
+            "pybtex does not work with Google Colab due to a known bug -"
+            "https://bitbucket.org/pybtex-devs/pybtex/issues/148/."
+            "Please manually cite all the references."
         )
     else:
         pybamm.citations.print(filename, output_format)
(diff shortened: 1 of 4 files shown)


pybamm/citations.py
```diff
@@ -143,7 +145,16 @@ def print(self, filename=None, output_format="text"):
...
 def print_citations(filename=None, output_format="text"):
     """See :meth:`Citations.print`"""
-    pybamm.citations.print(filename, output_format)
+    if "google.colab" in sys.modules:
+        raise ImportWarning(
+            """pybtex does not work with Google Colab due to a known bug -
+            https://bitbucket.org/pybtex-devs/pybtex/issues/148/.
+            Please manually cite all the references.""".replace(
+                "\n                ", " "
+            ),
```
tinosulzer, line 154:
```suggestion
            "pybtex does not work with Google Colab due to a known bug -"
            "https://bitbucket.org/pybtex-devs/pybtex/issues/148/."
            "Please manually cite all the references."
```

tests/unit/test_citations.py
```diff
@@ -64,6 +66,12 @@ def test_print_citations(self):
         with self.assertRaisesRegex(pybamm.OptionError, "'text' or 'bibtex'"):
             pybamm.print_citations("test_citations.txt", "bad format")
 
+        # google colab issue - https://github.com/pybamm-team/PyBaMM/issues/2524
+        sys.modules["google.colab"] = "mock"
```
tinosulzer, line 70: cool test!

tinosulzer (approved):
Looks good thanks! Will merge when tests pass

tests/unit/test_citations.py
Saransh-cpp: 😄

PR #2587, opened 2022-12-28
Fix publish workflow

brosaplanella:
# Description

Publish worfklow threw a similar error to the test workflow (in macos).

Test publish running [here](https://github.com/pybamm-team/PyBaMM/actions/runs/3795132883)


tinosulzer (approved):
Thanks!

Issue #2322, opened 2022-09-27
Make "experiment" simulations more customizable


tinosulzer:
Add an "intermediate" representation of model+experiment that consists of a list of specialized models to solve, e.g.
```python3
pybamm.Experiment(
    [
        (
            "Discharge at C/5 for 10 hours or until 3.3 V",
            "Rest for 1 hour",
            "Charge at 1 A until 4.1 V",
            "Hold at 4.1 V until 10 mA",
            "Rest for 1 hour",
        ),
    ]
    * 3
)
```
becomes
```python3
models_with_experiment = [
discharge_model,
rest_model,
charge_model,
hold_model,
rest_model
]
```
This is already what `Simulation` does when processing models, but we should make it more accessible so that users can run different types of experiment even if they can't be defined by the `Experiment` class

tinosulzer:
@Scottmar93 doing this would make both step-wise temperature control and CC-CT-CV (as we were discussing on slack) very easy, since the hard part is actually processing the strings from experiments

tinosulzer:
easy if you know the codebase

tinosulzer:
I like the idea and the suggested syntax, a couple of suggestions for syntax changes:
- temperature should be a separate input. Maybe in Kelvin to keep consistent with parameter values.
- drive cycles should be a 2-column numpy array containing the data, rather than a string-reference. It's not too much to ask of the user to load that data from the appropriate file, and saves us having to deal with all the edge cases from different OS
- Cycle ref should be via a `Cycle` object, e.g. `Cycle((Run(...), Rest(...), ref="..."))`
- `until` could be a `StopCondition` class e.g. `StopCondition(V=3.2, T=26)`
- Existing string functionality should still work (e.g. convert to this form automatically) - it's fine if it only works for a subset of cases

See also discussion in #1839 as the start time could be another argument to the experiment step classes

Scottmar93:
Really like the ideas in #1839 to help replicate long term usage patterns. Would work well with this context. 

Agree with the drive cycle being passed by the user, makes a lot of sense. Also agree that keeping existing experiment class functionality is a good idea to avoid breaking a bunch of people’s code. Perhaps just doing this in a separate class makes sense? Maybe l call it VirtualCycler or something? 

[...]

PR #2593, opened 2023-01-05
Add tags to `Experiment`

### LINKED ISSUE #2538: Tag experiments [former Reference Performance Test (RPT) experiment]
### Description

The capacity of the battery (e.g. when doing ageing tests) is typically done through RPTs, an experimental step consisting of a full charge and a CC discharge at a given C-rate until the cut-off voltage is hit. The capacity in that discharge is the "measured capacity".

These tests can be easily run with the current features (though the experiment needs to be split and solution reinitialised which can be cumbersome). However, this is a fairly standard test which is intercalated between many other steps, and often we only care about the measurements during the RPT. Possible ways around that would be to define a separate RPT class and then:
* Store RPT subsolutions (or maybe even just the capacity measure) as a separate variable in the solution.
* Alternatively, just store a list with the indices corresponding the RPT so they can be easily accessed.

Anyway, what are people thoughts on that?

### Motivation

Fairly standard experiment type used in ageing experiments.

### Possible Implementation

Define a separate RPT experimental step (ties with #2322)

### Additional context

This might make #2219 no longer needed, as the main motivation was doing RPT.

brosaplanella:
# Description

Tags can be added to the end of an experiment as `[tag1,tag2]`. I have added the method `search_tag` which returns the list of cycles with a certain tag, which should help saving only certain cycles.

Fixes #2538 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/experiments/experiment.py
+++ b/pybamm/experiments/experiment.py
@@ -464,5 +464,6 @@ class Experiment:
                     tags = tag_str[0:-1].split(",")
                     if tag in tags:
                         cycles.append(i)
+                        break
 
         return cycles
--- a/pybamm/experiments/experiment.py
+++ b/pybamm/experiments/experiment.py
@@ -145,7 +145,7 @@ class Experiment:
 
     def read_string(self, cond, drive_cycles):
         """
-        Convert a string to a tuple of the right format
+        Convert a string to a dictionary of the right format
 
         Parameters
         ----------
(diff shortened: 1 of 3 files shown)


brosaplanella:
At the moment we are not using the `"tags"` entry of the dictionary of processed experiments (I realised I didn't need it after writing that part). Should I remove them or keep it (could be useful in the future).

pybamm/experiments/experiment.py
```diff
@@ -440,3 +454,16 @@ def is_cccv(self, step, next_step):
             if op["events"] == {k: v for k, v in next_op.items() if k in op["events"]}:
                 return True
         return False
+
+    def search_tag(self, tag):
```
tinosulzer, line 458: seems like this should use the "tags" entry of the dictionary?

tinosulzer (approved):
Looks good to me, thanks!

pybamm/experiments/experiment.py
brosaplanella: That was my original idea, but then I couldn't find a clear mapping between `operating_conditions` and the actual cycle number. Going forward, when we implement #2322, that will have to be addressed and will make more sense to use the dictionary. But for the time being, processing the strings is fast enough.

Issue #2439, opened 2022-11-09
Autogenerate docstrings


tinosulzer:
There's a lot of repetition in the docstrings for expression tree classes and submodel classes, which leads to errors.
Would be cool to automatically generate the docstrings based on the arguments and the parent function, i.e.

- Detect what the arguments of `__init__` are and add the relevant parameter docstring for each argument
- Detect what the parent class is and add `**Extends:**`

Some ideas for how to do this from stackoverflow https://stackoverflow.com/questions/4056983/how-do-i-programmatically-set-the-docstring and https://stackoverflow.com/questions/13603088/python-dynamic-help-and-autocomplete-generation

tinosulzer:
Some other potentially useful references for this
- https://stackoverflow.com/questions/67205506/how-to-avoid-repeating-docstring-and-parameters-for-python-subclasses
- https://stackoverflow.com/questions/2025562/inherit-docstrings-in-python-class-inheritance?noredirect=1&lq=1
- https://stackoverflow.com/questions/5516032/docstring-inheritance-for-properties-using-sphinxs-autodoc
- https://stackoverflow.com/questions/8100166/inheriting-methods-docstrings-in-python?noredirect=1&lq=1

agriyakhetarpal:
I could give this one a try, though I would need a bit of help

tinosulzer:
great, happy to help

agriyakhetarpal:
> Detect what the parent class is and add **Extends:**

So I went through all the links that were given, but I think found a simple fix in the [Sphinx AutoAPI documentation](https://sphinx-autoapi.readthedocs.io/en/latest/reference/config.html#confval-autoapi_options), via the `show-inheritance:` directive which automatically picks up the base class for a derived class. By adding it, for example, in the `binary_operator.rst` file 
```
.. autoclass:: pybamm.BinaryOperator
  :members:
  :show-inheritance:
```
and then building the documentation will add the following line just below the name of the associated class like this
```
class pybamm.BinaryOperator(name, left, right)
Bases: [Symbol](http://127.0.0.1:8000/source/expression_tree/symbol.html#pybamm.Symbol)
```

So it writes the word **Bases** rather than **Extends**, (I don't know how to change that though) and links the base class for the derived class as intended. This will have to be manually added for all the autoclass directives similarly wherever it is applicable.
[...]

PR #2597, opened 2023-01-09
Issue 2439 autogenerate docstrings

agriyakhetarpal:
# Description

Adds `autodoc_default_options` to the default Sphinx configuration file so that based classes can be programmatically generated inside a docstring for a derived subclass, this will remove the need for the **Extends:** keyword used for manually doing so

Fixes https://github.com/pybamm-team/PyBaMM/issues/2439

## Type of change

# Key checklist:

## Further checks:


agriyakhetarpal:
Hi @tinosulzer, continuing our previous discussion on the issue page here: the other requirement is to 
> Detect what the arguments of __init__ are and add the relevant parameter docstring for each argument

After going through the links mentioned, I think I still don't understand this fully. Could you give me an example of what this would mean for, say, the `Function` class, that inherits from `pybamm.Symbol` (or any other example from the expression tree subclasses)? What should the outputs and/or documentation look like upon usage of said class?

tinosulzer:
For example the parameters in `FullOhm` https://github.com/pybamm-team/PyBaMM/blob/fa5ec1b58e93de2f087d9cd495a241af04d5200a/pybamm/models/submodels/electrode/ohm/full_ohm.py#L13-L18 are identical to the ones in `BaseOhm` so the docstring should be automatically copied from there https://github.com/pybamm-team/PyBaMM/blob/fa5ec1b58e93de2f087d9cd495a241af04d5200a/pybamm/models/submodels/electrode/ohm/base_ohm.py#L14-L19

agriyakhetarpal:
I was able to try three potential fixes:

- The `@wraps` decorator from [functools ](https://docs.python.org/3/library/functools.html) can help with copying the docstring from a base class to that of a derived class, though that would have to be repeated manually over every class (or inspect can be used to append it semi-automatically).

This would not be ideal if the docstring for the derived class is only a subset of the same for the base class, i.e., if the derived class has fewer parameters.

- Or we can use [inspect](https://docs.python.org/3/library/inspect.html) and do  `self.__class__.__init__.__doc__ = inspect.getdoc(BaseOhm.__init__)` inside FullOhm since Sphinx assumes that the docstring for the class is the same as that of `__init__`

- I also wrote a Sphinx extension that can be included in `conf.py`
```
from sphinx.ext import autodoc
from inspect import getmembers, isclass, signature

def generate_docstrings(app, what, name, obj, options, lines):
    if not isclass(obj):
        return
    for base in obj.__bases__:
        if isclass(base):
            init_method = None
            for name, member in getmembers(base):
                if name == '__init__':
                    init_method = member
            if init_method is not None:
                lines.clear()
                sig = signature(init_method)
                lines.append("""
                Parameters
                ----------
                """)
                for param in sig.parameters:
                    lines.append("    {} : {}".format(param, sig.parameters[param].annotation.__name__))
                lines.append("""
                """)

def setup(app):
    app.connect('autodoc-process-docstring', generate_docstrings)
```
to add parameters from a base class, but since many classes (like BaseModel in this case) are also deriving from another class, this messes up the hierarchy and does not carry over the documentation (manually) written for the parameter itself. 

Are there any other ways? I am somewhat stuck otherwise

tinosulzer:
Thanks for digging so deeply into this.

It's perfectly ok for this to be manual, i.e. just create a custom decorator based on `@wraps` so that
```python3
@extends(pybamm.electrode.ohm.BaseModel)
class Full(BaseModel):
    """
    Full model of electrode employing Ohm's law.   
    """
```
is equivalent to (has the same docstring as)
```python3
class Full(BaseModel):
    """
    Full model of electrode employing Ohm's law.
    
    Parameters
    ----------
    param : parameter class
        The parameters to use for this submodel
    domain : str
        Either 'negative' or 'positive'
    options : dict, optional
        A dictionary of options to be passed to the model.
    **Extends:** :class:`pybamm.electrode.ohm.BaseModel`
    """
```
The decorator should copy the `Parameters` section (which may themselves be copied from a parent class), and add the `Extends` line. If it's difficult to deal with multiple levels of copying, we could have two decorators, e.g. `@extends(ohm.BaseModel)` and `@parameters_from(BaseSubModel)`.

Adding a single line decorator to every class is fine. The key is that it will allow us to update the documentation in only one place (the base submodel class) if we ever want to change it, instead of having to go through every single file to make the same change.

agriyakhetarpal:
This should be fairly easy to do: since functools `@wraps` works only for methods and not classes themselves, I did not have to use it – plus the fact Sphinx is picking up documentation from the class itself rather than its `__init__` method, so we can define a decorator like this
```python
def parameters_from(cls):
    from inspect import getmro
    base_cls = getmro(cls)[1]
    cls.__doc__ += '\n' + ''.join(base_cls.__doc__.partition("Parameters")[1:])
    return cls

@parameters_from
class Full(BaseModel):
    """Full model of electrode employing Ohm's law."""
```

and then we can do `help(pybamm.models.submodels.electrode.ohm.Full)` which gives us

```
Help on class Full in module pybamm.models.submodels.electrode.ohm.full_ohm:

class Full(pybamm.models.submodels.electrode.ohm.base_ohm.BaseModel)
 |  Full(param, domain, options=None)
 |
 |  Full model of electrode employing Ohm's law.
 |
 |  Parameters
 |      ----------
 |      param : parameter class
 |          The parameters to use for this submodel
 |      domain : str
 |          Either 'negative' or 'positive'
 |      options : dict, optional
 |          A dictionary of options to be passed to the model.
 ```
 as required. However, the formatting is slightly messed up. Attaching a screenshot of the docs below
 
Formatting for Full(BaseModel)

Is this the intended way?

>  and add the Extends line

This was done in the commit on this PR via Sphinx's autodoc options based on our discussion in the issue thread. Though it writes the word **Bases** rather than **Extends**, but I think doing this with a decorator should not be difficult either

**Comment by tinosulzer** [comment]
Thanks @agriyakhetarpal , this is exactly what I was intending! Just need to fix the formatting. Could it say `Bases: pybamm.electrode.ohm.BaseModel` instead of just `BaseModel`?

**Comment by agriyakhetarpal** [comment]
Thanks, @tinosulzer! I was able to fix the formatting by modifying the decorator just slightly
```python
def parameters_from(cls):
    from inspect import getmro
    base_cls = getmro(cls)[1]
    cls.__doc__ += '\n\n    ' + ''.join(base_cls.__doc__.partition("Parameters")[1:])
    return cls
```
> Could it say `Bases: pybamm.electrode.ohm.BaseModel` instead of just `BaseModel`?

I looked through the Sphinx documentation hoping to find a solution, but it turns out that it cannot show the entire inheritance hierarchy. So the commit in this PR can be reverted, and we add another decorator in the same way as before such that
```python
def parameters_from(cls):
    from inspect import getmro
    base_cls = getmro(cls)[1]
    cls.__doc__ += '\n\n    ' + ''.join(base_cls.__doc__.partition("Parameters")[1:])
    return cls

def extends(cls):
    from inspect import getmro
    base_cls_name = getmro(cls)[1].__module__ + getmro(cls)[1].__name__
    cls.__doc__ += "\n\n    " + "**Extends:** :class:`" + base_cls_name + "`\n    "
    return cls 

@extends
@parameters_from
class Full(BaseModel):
```
which will add the **Extends** keyword.

Using both decorators builds the documentation exactly as required, albeit with a caveat: the latter decorator writes the full name of the class, i.e., `pybamm.models.submodels.electrode.ohm.base_ohmBaseModel`, rather than `pybamm.electrode.ohm.BaseModel`, due to which Sphinx cannot automatically link it – here is a screenshot of the docs. Is there a way you could suggest on how this could be fixed?

image

tinosulzer:
There's a typo in the class name, does it work if you change `pybamm.models.submodels.electrode.ohm.base_ohmBaseModel` to `pybamm.models.submodels.electrode.ohm.base_ohm.BaseModel` (adding a dot before `BaseModel`)?

Also, instead of `Extends` being in the `Parameters` list, it should be separate. And it should be `**Extends**` not `**Extends(**`. I know that is how it is in the current version of the docs but it's a mistake

Thanks for your work and good ideas on this, I didn't realize it would be this difficult

agriyakhetarpal:
Thank you for pointing that out – I had missed it. Building the docs (with the new theme as well) after fixing the typo makes it work, and now Sphinx can indeed link through the MRO as intended: 

tooltip


> Also, instead of Extends being in the Parameters list, it should be separate. And it should be **Extends** not **Extends(**. I know that is how it is in the current version of the docs but it's a mistake

Maybe we can add the Extends section above the Parameters, so that they are not clubbed together, like this:

new-pybamm-docs

tinosulzer:
> Maybe we can add the Extends section above the Parameters, so that they are not clubbed together, like this:

Yes, let's do that

agriyakhetarpal:
Okay, I can proceed with pushing changes for now – where should I define the decorators (and the documentation for them)? Also, the "Extends" line will have to be removed from every class docstring otherwise that too gets copied over for derived classes where these decorators are to be used

tinosulzer:
Yes, remove all "Extends", and adding decorators manually is fine (actually, preferred).

You can define them in a new file `pybamm/docs_util.py`

agriyakhetarpal:
Sure, which other functions need to be decorated? 

Additionally, I may write unit tests for these decorators too if you could help with some of those. My notion is that to test them, it can be checked if the decorated classes' docstrings include the parameters and the extends section – but that means we will need to rely on Sphinx doing the formatting correctly, plus the docstrings are being updated at build time; so I am not sure about how it goes

tinosulzer:
> My notion is that to test them, it can be checked if the decorated classes' docstrings include the parameters and the extends section – but that means we will need to rely on Sphinx doing the formatting correctly, plus the docstrings are being updated at build time; so I am not sure about how it goes

You should be able to just test that `.__doc__` includes `Extends` and `Parameters`

pybamm/models/submodels/electrode/ohm/full_ohm.py
```diff
@@ -3,22 +3,13 @@
 #
 import pybamm
 from .base_ohm import BaseModel
+from pybamm.doc_utils import parameters_from, extends
 
 
+@parameters_from
+@extends
```
tinosulzer, line 10: can you rename to `copy_parameter_doc_from_parent` and `doc_extend_parent`

tinosulzer:
Looking at it again, `parameters_from` should be manual but ideally `extends` would be automatic for all classes, is there a way to do that?

agriyakhetarpal:
> Looking at it again, `parameters_from` should be manual but ideally `extends` would be automatic for all classes, is there a way to do that?

Maybe I can write a Sphinx extension to do this like the one that we had discussed previously in this thread. Doing that would be the best way in my opinion since the decorator is only for stylistic purposes here, i.e., building docs

pybamm/models/submodels/electrode/ohm/full_ohm.py
agriyakhetarpal: Yes

agriyakhetarpal:
> Looking at it again, parameters_from should be manual but ideally `extends` would be automatic for all classes, is there a way to do that?

@tinosulzer on another attempt, I managed to get the first one out of the three ideas above to work: here is a Sphinx extension I was able to come up with to automatically add the extends label to all PyBaMM classes that derive from a base class. This offers the same functionality as `doc_extend_parent` in `pybamm/docs_util.py`

In `docs/sphinx_extend_parent.py`
```python
from inspect import getmro, isclass

def process_docstring(app, what, name, obj, options, lines):
    # if it is not a class, do nothing
    if not isclass(obj):
        return
    # else if it is a class
    else:
        # check if the class derives from another class
        if not len(getmro(obj)) > 2:
            # do nothing if it is not a derived class
            return 
        # check if the class has a docstring
        elif lines:
            # get the base class name
            base_cls_name = f"{getmro(obj)[1].__module__}.{getmro(obj)[1].__name__}"
            # add the extends keyword to the docstring
            lines.append(f"**Extends:** :class:`{base_cls_name}`")

def setup(app):
    app.connect("autodoc-process-docstring", process_docstring)
```

and in `docs/conf.py`
```python
extensions = [
    "sphinx.ext.autodoc",
    "sphinx.ext.doctest",
    "sphinx.ext.intersphinx",
    "sphinx.ext.mathjax",
    "sphinx.ext.viewcode",
    "sphinx.ext.napoleon",
    "sphinx_design",
    "sphinx_copybutton",
    "myst_parser",
    "sphinx_extend_parent",
]
```

This also fixes the formatting of the original **Extends:** section (which was earlier being included inside the Parameters instead of being shown separately), e.g., in the docs for `Base Electrolyte Diffusion Submodel`:

image

The `@copy_parameter_doc_from_parent` decorator is unchanged, it remains in `pybamm/doc_utils.py`

tinosulzer:
Thanks, this looks great. Did this add the first or second `Extends` in the screenshot you shared?

agriyakhetarpal:
The second one

agriyakhetarpal:
I removed the `Extends:` labels from the docstrings since the extension is now doing the job. Would a separate PR be preferred for adding the other decorator `@copy_parameter_doc_to_parent` on any of the needed functions?

tinosulzer:
> I removed the Extends: labels from the docstrings since the extension is now doing the job.

Perfect, thanks!

> Would a separate PR be preferred for adding the other decorator @copy_parameter_doc_to_parent on any of the needed functions?

No preference

agriyakhetarpal:
Great, I think a separate one would be much cleaner. This looks ready to merge?

tinosulzer:
Yes, once tests pass

agriyakhetarpal:
Fixed! I noticed that the `.ruff_cache` folder was not being ignored so I added it as well.

tinosulzer (approved):
Merging, thanks!

PR #2598, opened 2023-01-09
Added contact resistance

DrSOKane:
# Description

A new parameter "Contact resistance [Ohm]" has been added. This allows for the creation of a new variable "Contact voltage [V]", which is the same as "Terminal voltage [V]" but corrects for said contact resistance.

## Type of change

# Key checklist:

## Further checks:


DrSOKane:
The BPX test is failing because contact resistance is not included in the BPX schema. If everything else about this PR is fine, I can raise an issue on the BPX Github. But I can see why you might be concerned about my having created separate variables for voltage with/without contact resistance.

pybamm/models/submodels/electrode/base_electrode.py
```diff
@@ -196,5 +197,32 @@ def _get_standard_whole_cell_variables(self, variables):
...
                 )
             )
+            # Contact resistance alters the voltage. To make sure it works for all
+            # operating modes, a new variable "Contact voltage [V]" is created instead
+            # of changing the existing "Terminal voltage [V]".
+            V_dim = variables["Terminal voltage [V]"]
+            if "Current [A]" in variables:
+                I = variables["Current [A]"]
+                V_contact_dim = V_dim - I * param.R_contact
+            elif self.options["operating mode"] == "explicit power":
```
rtimms, line 207: In this case, wouldn't P=I*V_contact? And similar for explicit resistance? As in, the measured power would include the contact resistance? Or am I wrong?

It seems to me the cleanest way to do this would be to have a "contact resistance" option (so that the parameter is optional as @tinosulzer suggests) and then not allow "contact resistance" and "explicit power/resistance"?

tests/unit/test_parameters/test_bpx.py
```diff
@@ -34,6 +34,7 @@ def setUp(self):
                     "in parallel to make a cell": 34,
                     "External surface area [m2]": 3.79e-2,
                     "Volume [m3]": 1.28e-4,
+                    "Contact resistance [Ohm]": 0,
```
rtimms, line 37: Remove this from here and the test will pass. BPX doesn't let you add extra fields that aren't in the schema, so it throws an error. You already hard-code zero contact resistance in `bpx.py`. If users want to read in BPX and add the contact parameter resistance after, they can do that in the usual way.

DrSOKane:
All of the above issues have now been resolved but it's now failing to build scikits-odes...

tinosulzer:
Yes, @martinjrobins is working on fixing the scikits issue (not related to your changes)

pybamm/models/submodels/electrode/base_electrode.py
```diff
@@ -151,8 +156,8 @@ def _get_standard_current_collector_potential_variables(self, phi_s_cn, phi_s_cp
             "Positive current collector potential [V]": phi_s_cp_dim,
             "Local voltage": V_cc,
             "Local voltage [V]": self.param.ocv_ref + V_cc * pot_scale,
-            "Terminal voltage": V,
-            "Terminal voltage [V]": V_dim,
+            "Terminal voltage": V - delta_phi_contact,
```
tinosulzer, line 159: why is this minus and not plus?

tinosulzer: also would be good to add contact resistance as its own variable

DrSOKane: "Contact resistance [Ohm]" is a parameter. Did you mean make "Contact overpotential [V]" a variable? I can do that if you think it's appropriate.

tinosulzer: Yes I meant contact overpotential as a variable. Also smart to ignore my dumb minus vs plus comment, brain fart on my part

tinosulzer (approved):

PR #2596, opened 2023-01-09
I2550 fix bench

martinjrobins:
# Description

Fixes #2550 

## Type of change

Please add a line in the relevant section of [CHANGELOG.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CHANGELOG.md) to document the change (include PR #) - note reverse order of PR #s. If necessary, also add to the list of breaking changes.

- [x] Bug fix (non-breaking change which fixes an issue)

# Key checklist:

- [x] No style issues: `$ flake8`
- [x] All tests pass: `$ python run-tests.py --unit`
- [x] The documentation builds: `$ cd docs` and then `$ make clean; make html`

You can run all three at once, using `$ python run-tests.py --quick`.

## Further checks:

- [x] Code is commented, particularly in hard-to-understand areas
- [x] Tests added that prove fix is effective or that feature works

PR #2600, opened 2023-01-10
I2599 bench idaklu

### LINKED ISSUE #2599: benchmark IDAKLU solver alongside casadi
### Description

the IDAKLU solver seems generally faster than casadi in safe mode, it would be good to benchmark this properly, with the aim to make IDAKLU solver the default solver (if installed)

### Motivation

so provide evidence to justify making the IDAKLU solver the default

### Possible Implementation

_No response_

### Additional context

_No response_

martinjrobins:
# Description

Adds the idaklu solver to the existing benchmarks. Also refactors the simulation benchmarks into a single test suite

Fixes #2599 

Note: this follows on from PR #2596

## Type of change

- [x] updates to benchmarking

# Key checklist:

## Further checks:


tinosulzer (approved):
Thanks, it looks like this also includes changes from #2596 so we can just merge this one and close the other one right?

tinosulzer:
Can you update changelog with new benchmarks?

martinjrobins:
> Thanks, it looks like this also includes changes from #2596 so we can just merge this one and close the other one right?

Yes, I think that is the way to go.

This PR isn't quite finished, I can't seem to get asv to install pybamm with the idaklu solver, it works fine on my computer!

martinjrobins:
ok, looks like benchmarks are building ok now. I had to hardcode in the LD_LIBRARY_PATH in asv.conf.yml for the runners, because asv does not support env variables in the conf file. I'm not happy with this, but I can't see an alternative....

PR #2615, opened 2023-01-24
Support for Python 3.10

Dapid:
# Description

Now it works under Python 3.10

# Key checklist:

- [ x] No style issues: `$ flake8`
- [ x] All tests pass: `$ python run-tests.py --unit`
- [ x] The documentation builds: `$ cd docs` and then `$ make clean; make html`

## Further checks:
N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Features
 
 - Steps in `Experiment` can now be tagged and cycle numbers be searched based on those tags ([#2593](https://github.com/pybamm-team/PyBaMM/pull/2593)).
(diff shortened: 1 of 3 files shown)


brosaplanella:
Hi! Thanks for the contribution, unfortunately we can't support Python 3.10 yet as a main dependency (CasADI) does not support it. We already have #2435 in progress for when CasADI updates.

If you want to contribute to PyBaMM a good place to start are the [issues labelled as easy](https://github.com/pybamm-team/PyBaMM/issues?q=is%3Aopen+is%3Aissue+label%3A%22difficulty%3A+easy%22).

PR #2616, opened 2023-01-25
Timestamped experiments

### LINKED ISSUE #1839: Time-based events in experiments
**Summary**
Allow the definition of experiments based on actions happening at a given time which would be useful to simulate real usage scenarios. For example:
```python3
"On 2021-01-01 at 08:00 discharge at 1C until 2.5 V",
"On 2021-01-01 at 17:00 charge at C/3 until 4.2 V",
"Hold at 4.2 V until C/20",
...
```
The idea is that the experiment would add the necessary resting periods before and after the first action (triggered by the "On ..." statement) but would run the third one immediately after the second one (no "On ..." statement).

What are your thoughts? Still not clear to me what would be the best way to pass dates and times (e.g do we want to work with absolute dates, or can we also set up "On day 1..."?

 **Motivation**
This would allow the simulation of real usage scenarios for electric vehicles.

brosaplanella:
# Description

Enable functionality to "timestamp" experiment steps so they only get triggered at that date & time. At the moment the timestamps can be defined as:

- `"Day D HH:MM:SS"`
- `"YYYY-MM-DD HH:MM:SS"`

I am using the `datetime` package so the flexibility with the format (e.g. padding zeros) is that of `datetime`.

Fixes #1839 

## Type of change

# Key checklist:

## Further checks:


pybamm/simulation.py
```diff
@@ -775,6 +810,52 @@ def solve(
...
+                            kwargs["inputs"] = {
+                                **user_inputs,
+                                **op_conds,
+                                "Ambient temperature [K]": ambient_temp,
+                                "start time": start_time,
+                            }
+                            # Make sure we take at least 2 timesteps
+                            # The period is hardcoded to 10 minutes, the user can
+                            # always override it by adding a rest step
+                            npts = max(int(round(rest_time / 600)) + 1, 2)
```
brosaplanella, line 847: Not sure what's the better approach here. Another option would be to take 10 (or 20) steps or 10 min steps, whatever results in fewer steps.

brosaplanella:
@tinosulzer, following #2960, do we want the tags to be in the string or just as an option in `_Step`? I guess the latter, but just wanted to double check before undoing some of the changes.

brosaplanella:
I think this is ready to review

pybamm/simulation.py
```diff
@@ -256,6 +257,23 @@ def set_up_and_parameterise_model_for_experiment(self):
             )
             self.experiment_unique_steps_to_model[repr(op)] = parameterised_model
 
+        # Set up rest model if experiment has timestamps
```
rtimms, line 260: How come we need a separate model for this? Can't we use the current model and set the current to be zero like we do for a rest?

brosaplanella: I thought it would be more robust to have a separate model for rest, in case the user only has, for example, power driven steps (unlikely, I know).

rtimms: ah ok, yeah, makes sense

rtimms (approved):
looks good, thanks @brosaplanella !

tinosulzer (approved):
Functionality looks good. In terms of API I am wondering whether `timestamp` is the best name (what about `start_time`?). Also wondering whether we should just always require a datetime object, instead of providing support for some (but not all) datetime formats. It's not too much work for someone to write their own function to create a datetime object from a string.

PR #2618, opened 2023-01-26
Change docs to use Pydata theme

tinosulzer:
# Description

This is the theme used by numpy, scipy, pandas, etc and looks much better than our current one.
Also did a little bit of reformatting, started a user guide, etc, but there is more to be done

## Type of change

# Key checklist:

## Further checks:


tinosulzer:
Docs look like this on latest build: https://pybamm--2618.org.readthedocs.build/en/2618/
Don't know why the pydata theme isn't being used

tinosulzer:
Almost ready to go, final problem is the code is not displaying on this page https://pybamm--2618.org.readthedocs.build/en/2618/source/user_guide/fundamentals/index.html (it's not displaying locally either)

tinosulzer:
Found the problem, I was doing `.. code-block::python` instead of `.. code-block:: python` 🤦

docs/source/api/plotting/index.rst
```diff
@@ -4,7 +4,6 @@ Plotting
 .. toctree::
 
   quick_plot
-  dynamic_plot
```
Saransh-cpp, line 7: Was this intentional?

Saransh-cpp (approved):
The documentation looks amazing! Thanks, @tinosulzer!

brosaplanella (approved):
Looks great, thanks! We should aim to merge this by Tuesday so we can have the new docs with the new release.

docs/source/api/plotting/index.rst
tinosulzer: yes, the contents of that file are now in `quick_plot` since they go together

PR #2640, opened 2023-01-30
Fix broken links in README.md

iatzak:
The links to the docs in the [Installing PyBaMM section](https://github.com/pybamm-team/PyBaMM/blob/develop/README.md#-installing-pybamm) of README.md are broken. Also, even though not broken, in the [Using PyBaMM section](https://github.com/pybamm-team/PyBaMM/blob/develop/README.md#-using-pybamm) the "full API documentation" link brings the reader to the main page of the docs, instead of the [API page](https://pybamm.readthedocs.io/en/latest/source/api/index.html).


tinosulzer (approved):
Thanks for catching this and fixing so quickly! We usually have a check for this but it's not working right now :( #2641

PR #2672, opened 2023-02-08
Allow particle size to be a tuple and fix MPM half-cell

### LINKED ISSUE #2669: Allow "particle size" option to be a tuple
This allows us to have a single particle size in one electrode and a distribution in the other.

rtimms:
# Description
"Particle size" can now be a tuple to allow different behaviour in each electrode. `MPM` now works as a half-cell model, and no longer requires the user to provide distribution parameters for both electrodes. 

Also allows SPM(e) to have size distributions so that users can create more customised models (e.g. SPMe with a size distribution in one electrode only)

Fixes #2669 

## Type of change

# Key checklist:

- [ ] No style issues: `$ pre-commit run`

## Further checks:


tinosulzer (approved):
Looks good to me. How much effort is fixing the coverage?

PR #2679, opened 2023-02-11
Running integration tests locally #2674

### LINKED ISSUE #2674: Running integration tests locally
To run integration tests locally using `run-tests.py` I have to do `python run-tests.py --unit --folder integration` but would expect to be able to just do `python run-tests.py --folder integration` (i.e. I shouldn't need unit **and** folder). Is this the intended behaviour? 

It would also be handy to be able to just do `python run-tests.py --integration`

ayeankit:
I have modified **python run-tests.py --integration**  instead of using **python run-tests.py --unit --folder integration**

Fixes #2674

Added the integration instead of using unit and folder.

### DIFF
--- a/run-tests.py
+++ b/run-tests.py
@@ -387,8 +387,8 @@ if __name__ == "__main__":
     interpreter = args.interpreter
     # Unit tests
     if args.integration:
-        has_run= True
-        folder=args.folder[1]
+        has_run = True
+        folder = args.folder[1]
         run_code_tests(True, folder, interpreter)
     if args.unit:
         has_run = True
--- a/run-tests.py
+++ b/run-tests.py
@@ -315,7 +315,13 @@ if __name__ == "__main__":
         description="Run unit tests for PyBaMM.",
         epilog="To run individual unit tests, use e.g. '$ tests/unit/test_timer.py'",
     )
+
     # Unit tests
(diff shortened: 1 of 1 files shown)


run-tests.py
```diff
@@ -315,7 +315,13 @@ def export_notebook(ipath, opath):
         description="Run unit tests for PyBaMM.",
         epilog="To run individual unit tests, use e.g. '$ tests/unit/test_timer.py'",
     )
+
     # Unit tests
+    parser.add_argument(
+        "--integration",
+        action="store_true",
+        help="Which folder to run the tests from.",
```
rtimms, line 323: Can you update help to say "Run integration tests using the `python` interpreter."

rtimms (changes requested):
Thanks! Do you still need to use `--unit` and `--folder` to specify a folder? I think you should be able to type e.g. `python run-tests.py --folder all` instead of `python run-tests.py --unit --folder all`.

run-tests.py
ayeankit: We can do both either` python run-tests.py --unit --folder all` or `python run-tests.py --folder all`. It will work fine for both cases.

ayeankit:
Hello @rtimms, I have updated the changes . Please review it.

PR #2717, opened 2023-02-21
Issue #2333 Adding Single-Layer SEI

### LINKED ISSUE #2333: Add single-layer SEI

jeromtom:
# Description
This PR edits the SEI defaults to single-layer SEI and allows changing to the original with inner and outer layer SEI.
A "number of SEI layers" option was added to `BatteryModelOptions`, if it's "2", use the original code. If it's "1", only create a single variable L_sei (instead of L_inner and L_outer), and then only write equations for that variable.
Then references to two SEI thicknesses are edited reading `self.options["number of SEI layers"]`

Fixes #2333 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/models/full_battery_models/base_battery_model.py
+++ b/pybamm/models/full_battery_models/base_battery_model.py
@@ -122,6 +122,11 @@ class BatteryModelOptions(pybamm.FuzzyDict):
                     "solvent-diffusion limited", "electron-migration limited", \
                     "interstitial-diffusion limited", "ec reaction limited" \
                     or "ec reaction limited (asymmetric)": :class:`pybamm.sei.SEIGrowth`
+            * "number of SEI layers": int
+                Number of SEI layers to include in the model. Can be:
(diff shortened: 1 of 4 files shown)


Saransh-cpp:
@jeromtom any updates on this?

jeromtom:
@Saransh-cpp, I will be busy until 15th April since my team and I are participating in SAE India E-Baja. Will be available to work on it after that.

brosaplanella:
@jeromtom any updates on this?

jeromtom:
> @jeromtom any updates on this?

Will get back to this.

jeromtom:
@brosaplanella 
I tried to start fresh after closing this branch. But after multiple attempts, I am unable to do it. I suggest assigning the issue to someone else, and I will learn from it. Thanks.

brosaplanella:
No worries, will do! If you want to contribute some other issue, feel free to check those that are unassigned in the issue list (there are quite a few of easy ones that need addressing).

PR #2719, opened 2023-02-22
Changed current_sigmoid_ocp to be valid for both electrodes

DrSOKane:
# Description

The existing version of current_sigmoid_ocp is only valid for the negative electrode and gives the wrong result for hysteresis in the positive electrode. Fortunately, adding an extra four lines of code fixes this.

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ## Bug fixes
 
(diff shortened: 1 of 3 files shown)


tinosulzer:
Would it be clearer to define `lithiation_current` with opposite signs for positive and negative electrodes?

tinosulzer (approved):

PR #2724, opened 2023-02-23
Divided by 3600 to replace t0_cr

### LINKED ISSUE #2418: Dimensional models

DrSOKane:
# Description

The old non-dimensionalized PyBaMM had a constant t0_cr used for the crack propagation submodel. This was deleted, so I divided dl_cr/dt by 3600 (standard value for t0_cr) to compensate and retreive the same results as existing PyBaMM.

Fixes #2418 

## Type of change

# Key checklist:

## Further checks:


tinosulzer (approved):
Great! Thanks very much

PR #2732, opened 2023-02-25
Issue 2265 solved

### LINKED ISSUE #2265: Raise error if updating a parameter that already exists with `check_already_exists=False`
This will force people to use `check_already_exists=False` sparingly, and help to catch bugs related to misnaming parameters (this is caught if `check_already_exists=True`, but not if it's `False`

ayeankit:
Add a new parameter which is not in the dictionary while having `check_already_exists=True` and error that throws an error if while having `check_already_exists=False` the user tries to change an existing parameter.

Fixes #2265


ayeankit:
Hello @brosaplanella , I have raise a PULL request. Can you review it?

PR #2734, opened 2023-02-26
Migrate to Lychee to check URLs

agriyakhetarpal:
# Description

Implements a [GitHub Actions](https://github.com/lycheeverse/lychee-action) workflow offered by [Lychee](https://github.com/lycheeverse/lychee) (Rust-based URL checker) to potentially speed up URL checking

## Type of change

# Key checklist:

## Further checks:


agriyakhetarpal:
TODO: (I shall do this soon)
- Add more links across files
- Explicitly mention file types as `.rst` / `.md` / `.yml` because it currently reads all files as plaintext
- Set up a cache in accordance with cron (if this may be required?)
- Set up `lychee.toml` file for config
- Pin version for enhanced security (avoid tag spoofing)
- Perhaps add a repo badge to README.md

agriyakhetarpal:
@tinosulzer @Saransh-cpp I managed to get it to work, but it is currently failing with too many requests after a point, for which an API token may be required (I have added the config in the workflow file with `GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}` but I am not sure and cannot verify if it is using any; I do not have permissions to add secrets to the PyBaMM repository)

P.S. Please check the todo I added above as well

.github/workflows/lychee_url_checker.yml
```diff
@@ -0,0 +1,27 @@
...
+jobs:
+  linkChecker:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Lychee URL checker
+      # use stable version for now to avoid breaking changes
+      # sha-256 hash of latest stable version for security reasons
+        uses: lycheeverse/lychee-action@d322e38893906cef8788eefe9c63694c562c7dfa
```
Saransh-cpp, line 19: Also, I think it should be okay to leave it as `@v1`. These actions are usually safe. (Given that `@v1` works, I think it should.)
```suggestion
        uses: lycheeverse/lychee-action@v1
```

.github/workflows/lychee_url_checker.yml
```diff
@@ -0,0 +1,27 @@
...
+        with:
+          # fail the action on broken links
+          fail: true
+        env:
+        # to be used in case rate limits are surpassed
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+          # arguments with file types to check
+          args: --no-progress  './**/*.rst' './**/*.md' './**/*.py' './**/*.ipynb' --timeout 10 --max-retries 5
```
Saransh-cpp, line 27: It is usually a good practice ([1](https://stackoverflow.com/questions/2287967/why-is-it-recommended-to-have-empty-line-in-the-end-of-a-source-file), [2](https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file)) to leave an empty line at the end of code files.

.github/workflows/lychee_url_checker.yml
agriyakhetarpal: Maybe I can pin it to v1.6.1 instead because it has been updated to use the latest `lycheeVersion` by default, i.e., v0.11.1

.github/workflows/lychee_url_checker.yml
agriyakhetarpal: Noted!

.github/workflows/lychee_url_checker.yml
```diff
@@ -0,0 +1,27 @@
+name: Check URLs with Lychee
+
+on:
+  push:
```
Saransh-cpp, line 4: I think we can get rid of the `push` trigger. This would run the Action twice on a PR that is created from a branch of the main PyBaMM repository (once when the commit is "pushed" to the branch on PyBaMM, and the second time when a "pull request" is created from this branch). 

Or, maybe even better, we can run it on `push`, but only when the `push` is made to either `main` or `develop` (that is when a PR is merged, as pushing directly to `main` and `develop` is not allowed) -

```suggestion
  push:
    branches:
      - main
      - develop
```

agriyakhetarpal: Good idea, though any failing links at PRs will then be caught only when the PR is being merged

Saransh-cpp: Yes, both options sound good.

agriyakhetarpal:
Should be ready to merge now despite the fact that some links are failing. Some other files and folders may also be excluded as well

tinosulzer:
Why are the links failing? They're all passing the old links checker

agriyakhetarpal:
> Why are the links failing? They're all passing the old links checker

The links are valid but the response code is 429 so we're getting rate limited, which should hopefully be fixed by the cache and the token (the cache is not being generated for some reason). Investigating this right now

Saransh-cpp (changes requested):
I found some resources for the failing `github` links with `429` error code -

- https://github.com/lycheeverse/lychee-action/issues/138
- https://github.com/lycheeverse/lychee/issues/634
- https://lychee.cli.rs/#/troubleshooting/rate-limits

This is a GitHub issue, not a `lychee` issue (though they are trying to fix it from their end). The PR looks perfect but would require minor hit and try modifications to see what satisfies GitHub's requirements. (As one of the comments says, GitHub could be thinking that we are trying to DDoS their servers haha)

agriyakhetarpal:
I went through the links, I think it is also partly because Lychee needs the links to pass the checks before it caches them, which is why the GitHub links have not been cached yet – sort of a deadlock situation. 

I doubt if changing the number of max retries can fix it (I have kept it to 5, in line with that of the old links checker) or if decreasing the number of concurrent requests would help (will cause a slowdown albeit still faster than the old one – though I will try this). I can accept `429` status codes with `--accept 429` for now

Saransh-cpp:
I think maybe it would be good to open an issue in `lychee` (or a discussion) to check if we are doing something wrong. I was hoping accepting the `429` error code would work.

agriyakhetarpal:
We could also check the GitHub links with less concurrency in another similar action as someone had mentioned, but that would be a bit of a hack. I'll open up a discussion there and follow up on it.

mre:
Hey y'all! Thanks for using lychee.

The pipeline isn't working because `args` is in the wrong position.
Here's a fix: https://github.com/agriyakhetarpal/PyBaMM/pull/1

mre:
Two more hints, just because I took a look at the remaining errors:

- You can exclude loopback links (your http://127.0.0.1:8000) with ` --exclude-loopback`
- https://twitter.com/pybamm_ is timing out. I would exclude that for the time being. More info here: https://github.com/mermaid-js/mermaid/pull/3973#issuecomment-1458054029

.github/workflows/lychee_url_checker.yml
```diff
@@ -0,0 +1,51 @@
...
+        with:
+          # arguments with file types to check
+          args: >-
+            --cache
+            --no-progress
+            --max-cache-age 2d
+            --timeout 10
+            --max-retries 5
+            --skip-missing
+            --accept 200,429
```
Saransh-cpp, line 41:
```suggestion
            --skip-missing
            --exclude-loopback
            --exclude https://twitter.com/pybamm_
            --accept 200,429
```

Saransh-cpp (approved):
Thank you so much for the help, @mre!

@agriyakhetarpal, this looks good to me now with 1 final change (suggested by @mre above).

The failing URLs (true positives) can be fixed in this PR or in a separate PR, your call. (It would be nice to fix them in this PR itself.)

agriyakhetarpal:
Happy to fix the links in this PR itself

brosaplanella:
Cool, let us know if there is any reference you can't trace back.

agriyakhetarpal:
The links in the markdown files are failing since they are empty (Sphinx is still able to build the docs without them by referencing the links internally). Should I change them to include the readthedocs links for the stable release?

tinosulzer:
It's nice to be able to reference links internally in the docs without having to give the full path. This also makes it more robust to url changes. Is there a way around this without having to give the full path?

agriyakhetarpal:
Though it does not sound desirable, maybe we can exclude the entirety of these folders or the files therein, in particular (`docs/source/user_guide/` and in `examples/notebooks/`) for now?

agriyakhetarpal:
I think the best way is to explicitly (manually) exclude all the links to the files since there are only eight of them failing anyway. We can set up a `.lycheeignore` file and add them there, so that the workflow file remains relatively cleaner

mre:
Alternatively you could consider checking the links on the rendered HTML, which should contain the full link paths. I don't know how to build the HTML files for the project, so I could not test it myself.
If you have relative links (e.g. `../foo/bar.html`), that you'd like to check as well, you can also set `--base /path/to/root/folder` so that all the links will be checked relative to this base path.

But yeah, excluding the links certainly is an option as well. 👍

agriyakhetarpal:
I doubt if the link checks could be performed in conjunction with building the docs using a composite action – if I am correct, readthedocs has its own workflow file that is separate from GitHub Actions. Maybe this can be integrated within `tox`? I'll exclude the links for the time being

Saransh-cpp:
Excluding them sounds good to me

tinosulzer:
It's not ideal to have to exclude everything manually. Could we exclude the whole path `file:///home/runner/work/PyBaMM/PyBaMM/docs/source/` or would that then miss some errors?

agriyakhetarpal:
Yes, the whole folder `PyBaMM/docs/source/` can be excluded as well – though it would miss errors for external hyperlinks lest they are added to the API docs or the user guide in the future

tinosulzer:
Ok, yeah we don't want to miss those, let's leave like this then. Ready to merge?

agriyakhetarpal:
Yes sure — that failing integration test probably was a one-off one, shouldn't come up on a re-run

PR #2740, opened 2023-03-02
Other useful variables

tinosulzer:
# Description

Add some useful variables, fix plot_voltage_components and add example

## Type of change

# Key checklist:

## Further checks:


tinosulzer:
This PR looks nasty but it's mainly a lot of renaming variables, as detailed in CHANGELOG

CHANGELOG.md
```diff
@@ -1,7 +1,20 @@
 # [Unreleased](https://github.com/pybamm-team/PyBaMM/)
 
+## Features
+
+- Renamed "Terminal voltage [V]" to just "Voltage [V]". "Terminal voltage [V]" can still be used and will return the same value as "Voltage [V]".
+- Added "Anode potential [V]", which is the value of the surface potential difference (`phi_s - phi_e`) at the anode/separator interface, commonly controlled in fast-charging algorithms to avoid plating. Also added "Cathode potential [V]", which is the value of the surface potential difference at the cathode/separator interface. ([#2740](https://github.com/pybamm-team/PyBaMM/pull/2740))
```
brosaplanella, line 6: Shouldn't it be "Positive electrode" and "Negative electrode" so it is consistent with the other variables?

examples/scripts/compare_lithium_ion.py
```diff
@@ -7,9 +7,9 @@
 
 # load models
 models = [
-    pybamm.lithium_ion.SPM(),
-    pybamm.lithium_ion.SPMe(),
-    pybamm.lithium_ion.DFN(),
+    # pybamm.lithium_ion.SPM(),
+    # pybamm.lithium_ion.SPMe(),
+    # pybamm.lithium_ion.DFN(),
```
brosaplanella, line 12:
```suggestion
    pybamm.lithium_ion.SPM(),
    pybamm.lithium_ion.SPMe(),
    pybamm.lithium_ion.DFN(),
```

brosaplanella (approved):
Looks good, thanks! Just a couple of small comments

CHANGELOG.md
tinosulzer: The full name for this would be "Negative electrode surface potential difference at the separator interface". But I think in the fast-charging literature it's just called "Anode potential". Which do you think is better here?

TomTranter:
This caused liionpack to break as I was using the Measured battery ocv. Also when I changed to use the new OCV results looked strange and so I'm now using surface ocv.

TomTranter:
I think it might be nice to have some example of what they all mean and where they are used

rtimms:
What was "Measured open circuit voltage [V]" has been renamed to "Surface open-circuit voltage [V]". This is the OCV calculated from the surface concentration. "Open-circuit voltage [V]" is calculated from the bulk. 

Since the thing you would actually measure as (pseudo) OCV in an experiment is the surface OCV, I wonder if "Open-circuit voltage [V]" and "Surface open-circuit voltage [V]" should be the same, and we should also have "Bulk open-circuit voltage [V]".

We could explain this more thoroughly in the [plot voltage components notebook](https://github.com/pybamm-team/PyBaMM/blob/develop/examples/notebooks/plotting/plot_voltage_components.ipynb)

TomTranter:
I agree if the more observable one is Surface that should be the "Open-circuit voltage [V]", what is the bulk one used for?

rtimms:
So you can separate out the effect of concentration gradients within the particle on the voltage

tinosulzer:
I think the "Open-circuit voltage" should be the one that you would relax to if you turned the current off, i.e. the bulk one. What makes the surface one more observable?

rtimms:
It's just a point of notation where we should be consistent. If you asked people "what is the OCV?" in their model they would probably say `U_p(c_p_surf) - U_n(c_n_surf)` which is what we now call "Surface open-circuit voltage [V]". I agree that it makes to have bulk open-circuit voltage (the one you relax to), but is this consistent with the defn most people have in their head when thinking about continuum models?

tinosulzer:
Got it. There's also the option of just not defining "Open-circuit voltage" so you have to specify bulk or surface explicitly each time. We should update this either way before releasing

Issue #2744, opened 2023-03-06
[Bug]: The NCA_Kim2011 OCP curve CSV value is inconsistent with the original Article Function


1836005678:
### PyBaMM Version

22.11.1

### Python Version

3.9.13

### Describe the bug

I am using the parameter set NCA_Kim2011. When I set the initial soc to 1 and begin my simulation just like
`solver_used = pybamm.CasadiSolver(mode="safe")`
`sim = pybamm.Simulation(model_dfn, parameter_values=params_values_used, experiment = experiment_used, solver = solver_used)`
`sim.build_for_experiment(initial_soc=1)'`
`sim.solve()`
then I come to the bug 'interpolation bounds were exceeded' "Interpolant 'nca_ocp_Kim2011_data' lower bound".
![image](https://user-images.githubusercontent.com/57339960/223064276-59ed6f66-d307-4885-aea4-113c364a0e24.png)
So I go to the nca_ocp_Kim2011_data.csv and find that the maxnimum sto provided is 0.993382838635443. This will arise problems not only when we set initial soc to 1, but also when we charge the battery to the Upper Voltage Cut-off. Therefore, I wanted to replace CSV Data with Function. However, when I compare the OCP curve provided from the CSV and the Kim2011 Article, I find that there are difference between them. The CSV Data Curve is higher than the Function Curve. The difference is much more obvious when sto is close to 1 or 0.37.
![image](https://user-images.githubusercontent.com/57339960/223063298-a0ad9788-2010-4a88-82e8-1c30e46a9952.png)
My question is:
1.Which curve is reliable.
2.If the CSV Data Curve is correct, to provide additional interpolation points, what the Voltage is when sto equals to 1?

### Steps to Reproduce

1.CSV
Open pybamm\input\parameters\lithium_ion\data\nca_ocp_Kim2011_data.csv
Draw picture using scatter diagram.
2.Original Article
[https://iopscience.iop.org/article/10.1149/1.3597614](url)
Open page 8 of the article, see Table III, Positive electrode U+(V) item.
Draw picture using your code, or just like
`x = 0.37:0.01:0.99`
[...]

PR #2751, opened 2023-03-07
Update NCA_Kim2011.py

1836005678:
Fixing #2744 to make sure we hit the voltage cut-off, instead of the particle stoichiometry going above 1.
Using the function in Kim2011 article to take the place of pybamm\input\parameters\lithium_ion\data\nca_ocp_Kim2011_data.csv for parameter 'Positive electrode OCP[V]' in Kim2011 parameter set.
The function 'nca_ocp_Kim2011(sto)' in pybamm\input\parameters\lithium_ion\NCA_Kim2011.py is edited.


tinosulzer (changes requested):
Looks good, you just need to remove `import os` in this file since it is no longer used

tinosulzer:
Hmm, tests are failing because the old NCA Kim error was being used to test for the error reporting, if you feel confident to fix it I would suggest adding a different example where the interpolant extrapolates instead. I think something like this should work
```python3
x = np.linspace(0,2)
var = pybamm.Variable("var")
rhs = pybamm.Interpolant(x, x, var, name=name, interpolator="linear")
model = pybamm.BaseModel()
model.rhs[var] = rhs
model.initial_conditions[var] = 1
solver = pybamm.ScipySolver()
t_eval = [0,5]

# solution should grow exponentially and go above var=2, then the interpolant should error
solver.solve(model, t_eval)
```

1836005678:
Sorry, I don't know what 'old NCA Kim error was being used to test for the error reporting' means. Could you explain it in detail?

tinosulzer:
This test is failing since line 508 expects an error from the parameter set before you fixed it. https://github.com/pybamm-team/PyBaMM/blob/f45f4a5988da047ee356a29fc570475b1a110421/tests/unit/test_solvers/test_casadi_solver.py#L488-L509

Instead, we should test using a standalone piece of code like the one I shared above, and still make sure the error is raised

tinosulzer (approved):

1836005678:
Although I still don't quite understand the reason and purpose of this test, I am very grateful for this update and your work.

PR #2769, opened 2023-03-12
Fix version switcher on readthedocs

### LINKED ISSUE #2766: [Bug]: readthedocs version switcher does not style properly for older PyBaMM versions
### PyBaMM Version

23.2

### Python Version

3.9.13

### Describe the bug

1. The [version switcher dropdown](https://pydata-sphinx-theme.readthedocs.io/en/v0.8.1/user_guide/configuring.html?highlight=orange#style-the-switcher-buttons) does not style properly on [readthedocs]( https://pybamm.readthedocs.io): version text is displayed in a green coloured font regardless of the active version (which should ideally be `orange` for dev, `green` for the latest version, and `red` for older versions of PyBaMM)

Some other related issues: 

2. Links with the version switcher to the API documentation for past versions are broken (though this is because the API documentation index page was earlier not hosted separately)
3. The documentation does not use the PyData Theme for PyBaMM versions 22.12 and older, and uses the old theme instead

The ideal version switcher should look like this: https://numpy.org/doc/1.23/reference/index.html

### Steps to Reproduce

Hosted on readthedocs and with the [instructions to build docs](https://github.com/pybamm-team/PyBaMM/blob/develop/CONTRIBUTING.md#building-the-documentation)

### Relevant log output

_No response_

agriyakhetarpal:
# Description

This updates the version switcher styling on readthedocs and configures a `.json` file for the version switcher dropdown

Fixes #2766 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -138,7 +138,7 @@ html_theme_options = {
     ],
     "switcher": {
         "version_match": release,
-        "json_url": "source/_static/versions.json",
+        "json_url": "https://pybamm.readthedocs.io/en/latest/_static/versions.json",  # noqa: E501
     },
     # turn to False to not fail build if json_url is not found
     "check_switcher": True,
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -136,12 +136,14 @@ html_theme_options = {
             "url": "https://github.com/pybamm-team/PyBaMM/tree/develop/CONTRIBUTING.md",
         },
     ],
-    # Add light/dark mode and documentation version switcher:
-    # "navbar_end": ["theme-switcher", "version-switcher", "navbar-icon-links"],
-    # "switcher": {
-    #     "version_match": switcher_version,
-    #     "json_url": "https://numpy.org/doc/_static/versions.json",
-    # },
+    "switcher": {
+        "version_match": release,
+        # make sure to update this file when releasing a new version
+        "json_url": "source/_static/versions.json",
+    },
+    # turn to False to not fail build if json_url is not found
+    "check_switcher": True,
+    "navbar_end": ["version-switcher"],
     "use_edit_page_button": True,
 }
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -142,7 +142,8 @@ html_theme_options = {
     },
     # turn to False to not fail build if json_url is not found
     "check_switcher": True,
-    "navbar_end": ["version-switcher"],
+    # for dark mode toggle, version switcher, and social media links
+    "navbar_end": ["theme-switcher", "version-switcher", "navbar-icon-links"],
     "use_edit_page_button": True,
 }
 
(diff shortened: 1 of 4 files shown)


agriyakhetarpal:
The reason why all the versions are being shown in green is because `json_url` is not configured in `html_theme_options` yet, I shall do that in this PR

agriyakhetarpal:
@tinosulzer the readthedocs build is passing but I can not test how it would look like (if it will show versions with the correct colour scheme when hosted) – I added the version switcher dropdown to the header like you requested but it is currently not including past versions on my local build and shows up blank: 

image

Though this configuration matches that of the [numpy docs](https://github.com/numpy/doc/blob/main/_static/versions.json), which is why I suspect it should work as intended.

Also, the new file `docs/_static/versions.json` will have to be updated with the latest version on the creation of each release.  Right now there's the option `"check_switcher": True` in `html_theme_options` but that can only test whether `versions.json` can be found/read, and not if the file includes the latest version. Happy to write some sort of check or a warning for this if needed in the existing workflows if you could guide me with it

tinosulzer:
To keep the version up to date, add the appropriate code to https://github.com/pybamm-team/PyBaMM/blob/develop/scripts/update_version.py. This file gets run automatically at the end of every month to update the version in the various places it needs to be updated.

Also, make sure your changes don't kill the existing day/night switch and social links

agriyakhetarpal:
I added the code to update the json and tested it, though it messes up the formatting after doing so. Should still be readable

> Also, make sure your changes don't kill the existing day/night switch and social links

I am assuming that shouldn't happen because the code for them was already commented out, and also because the links do not show up on building the docs locally, but do so when hosted on readthedocs

tinosulzer:
You can see the built docs here, it's missing the links https://pybamm--2769.org.readthedocs.build/en/2769/
I think when "navbar_end" is not specified it defaults to the day/night and link checker, which is why they were removed by you adding a "navbar_end" without them included

agriyakhetarpal:
That makes more sense, I updated it to include the links and the day/night button. Also TIL one can see the docs built from a PR this way

agriyakhetarpal:
I was able to show the versions in the dropdown locally (the colours appear in the readthedocs build). This should be ready to merge now

image

This was done by adding the `.json` file at a static URL (tried it with a link to the raw file in this PR) – I was not able to get it to work with a local `file:///` URL despite multiple tries (according to the second point in the [docs](https://pydata-sphinx-theme.readthedocs.io/en/latest/user_guide/version-dropdown.html#configure-switcher-json-url)), but remote URLs do work. So I put a link to `https://pybamm.readthedocs.io/en/latest/_static/versions.json` as the first point suggests; this link currently returns a 404 (fails Lychee as well), but that would be fixed after this PR gets merged and `versions.json` gets added to the codebase.

tinosulzer (approved):

tinosulzer:
Looks good, thanks!

PR #2773, opened 2023-03-13
Issue 2760 step

### LINKED ISSUE #2760: Offset start time when stepping
Currently, when solving a simulation by stepping, the final time step of the old solution (e.g. `[0,0.01 ..., 1]`) is the same as the first time step of the next solution (e.g. `[1,1.01,...]`). We get around this by excluding the first time step of the next solution when constructing the full solution (`[0,0.01,...,1,1.01,...]`). However, this means that the solution naively constructed by concatenating together all the steps is different from the solution as constructed by PyBaMM. Instead, we should offset the start of each step by a relevant small number (e.g. `1e-6`) and then keep the first step.

tinosulzer:
# Description

Adds offset to the start of steps which follow from previous solution. This avoid duplicate time in solution and ensures that concatenations of steps (/cycles) have the same length as the whole cycle (/solution)

Fixes #2760

## Type of change

# Key checklist:

## Further checks:


brosaplanella (approved):
Looks good, thanks! Happy to merge once tests pass

PR #2795, opened 2023-03-21
`InputParameter` in variable bounds

### LINKED ISSUE #2793: [Bug]:  Error if variable bounds contain an `InputParameter`
Now that we switched to using dimensional variables, if you have an `InputParameter` in the bounds of a variable you can't discretise it. We should skip the bounds in this case IMO, otherwise, you would need to re-discretize each time you changed the input.

Code to reproduce this:

```python3
import pybamm
param = pybamm.ParameterValues("Chen2020")
param["Maximum concentration in positive electrode [mol.m-3]"] = "[input]"
sim = pybamm.Simulation(model, parameter_values=param)
sim.solve([0, 3600], inputs = {"Maximum concentration in positive electrode [mol.m-3]": 65000})
```

rtimms:
# Description
If variable bounds contain an `InputParameter` set the bound to +/- inf.

Fixes #2793

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ## Bug fixes
 
+- Fixed a bug where variable bounds could not contain `InputParameters` ([#2795](https://github.com/pybamm-team/PyBaMM/pull/2795))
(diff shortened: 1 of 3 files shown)


brosaplanella (approved):
Looks good, thanks!

PR #2846, opened 2023-04-01
Issue #2457 make jupyter a dev only dependency

### LINKED ISSUE #2457: Make `jupyter` a dev-only dependency
There are some issues with dependencies on arm64 (`qtconsole`)

arjxn-py:
# Description

Made jupyter a development only dependency instead of required dependency.

Fixes #2457 

## Type of change

Jupyter has been moved from required dependency to development only dependency in `requirements.txt` & `setup.py` and also been added as a testing dependency in `tox.ini`.

# Key checklist:

## Further checks:


arjxn-py:
@tinosulzer I have listed the packages that were dependent on Jupyter in required dependencies but now the tests are failing while running example notebooks most likely on this line - `%pip install pybamm -q` giving syntax error.
I would be grateful If I could get some suggestion on it.

tinosulzer:
you should install jupyter in `tox.ini` before running the example tests

arjxn-py:
@tinosulzer Thankfully, almost all the tests are passing now. Though i'm not sure about the Lynchee one.

setup.py
```diff
@@ -197,7 +197,8 @@ def compile_KLU():
         "scikit-fem>=0.2.0",
         "casadi>=3.5.0",
         "imageio>=2.9.0",
-        "jupyter",  # For example notebooks
+        "importlib-metadata",
+        "nbconvert",
```
Saransh-cpp, line 201: IMO `nbconvert` should also be a `dev` dependency, given that it is only used for testing.

arjxn-py: Before making `jupyter` a `dev` dependency `nbconvert` was being installed with that only but after moving `jupyter` to dev dependency it was giving same error as `importlib-metadata` so I had to explicitly list it as required dependency.
That being said, i may try moving it to dev dependency as well if you suggest. I'll wait for your response to do that after this comment.

Saransh-cpp: Yes, `nbconvert` should be made a `dev` dependency.

tinosulzer:
won't jupyter install nbconvert (i.e. no need to make explicit)

arjxn-py:
> won't jupyter install nbconvert (i.e. no need to make explicit)

I just listed it to avoid the potential risk to examples tests failing, but yes `jupyter` should install `nbconvert` with it so I am looking into it.
Thanks for suggesting.

setup.py
arjxn-py: Did it, therefore marking it as resolved.

tox.ini
```diff
@@ -18,6 +18,7 @@ deps =
      doctests: sphinx-copybutton
      doctests: myst-parser
      !windows-!mac: scikits.odes
+     !windows-!mac: jupyter   # For example notebooks
```
Saransh-cpp, line 21: Also, shouldn't this be just `examples`? Given that `jupyter` is only used to test example notebooks? Or are we using it somewhere else too?

https://github.com/pybamm-team/PyBaMM/blob/08dc8f7b0de950e109981fd81a6bb78866548067/tox.ini#L27

```suggestion
     examples: jupyter   # For example notebooks
```

setup.py
```diff
@@ -218,6 +218,8 @@ def compile_KLU():
         "dev": [
             "pre-commit",  # For code style checking
             "black",  # For code style auto-formatting
+            "jupyter",  # For example notebooks
+            "nbconvert",  # For example notebooks
```
Saransh-cpp, line 222:
```suggestion
```

Saransh-cpp (changes requested):

tox.ini
arjxn-py: Yes, we are only using `jupyter` to run `examples`. 
I want to ask one thing in this regard out of curiosity that  `!windows-!mac` would already avoid specified dependency from getting installed on windows & mac, what would `examples` do?

arjxn-py:
One of the integration test is failing but I don't think I've made any changes in that regard.

Saransh-cpp (approved):
Thanks, @arjxn-py! Looks good to me!

tinosulzer (approved):
Can you update changelog (add this as a breaking change)?

arjxn-py:
> Can you update changelog (add this as a breaking change)?

Yes i'll do that.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # [Unreleased](https://github.com/pybamm-team/PyBaMM/)
 
+# Breaking changes
+
+- Made `Jupyter` a development only dependency. Now `Jupyter` would not be a required dependency for users while installing `PyBaMM`. ([#2457](https://github.com/pybamm-team/PyBaMM/pull/2846))
```
tinosulzer, line 5:
```suggestion
- Made `Jupyter` a development only dependency. Now `Jupyter` would not be a required dependency for users while installing `PyBaMM`. ([#2457](https://github.com/pybamm-team/PyBaMM/pull/2846))
```

tinosulzer (approved):

PR #2862, opened 2023-04-06
Issue 1182 verbose outputs and tagging PyBaMM citations

### LINKED ISSUE #1182: Tag PyBaMM citations
**Summary**
The way citations are printed now (all together) can make it quite hard for an inexperienced user to know why the need to cite each reference. It would be good to add a tag when registering PyBaMM citations and also some extra functionality that, when printing the citations, reads the tags and groups them in a more meaningful way. For example, we could print some sort of list or table that lists:

- Model: Fuller1994
- Parameters: Chen2020
- Solver: Andersson 2019
and so on...

Please add your comments and suggestions

 **Motivation**
 Make it easier for new or inexperienced users to know what's the purpose of each reference, which might lead in them just dumping the citations in their papers all together. Adding this functionality should lead to higher quality citations in the papers.

agriyakhetarpal:
# Description

Adds a caller method for `citations.py` to tag PyBaMM citations to their use of particular models or solvers and adds verbose output while printing citations

Fixes #1182 

## Type of change

# Key checklist:

## Further checks:


agriyakhetarpal:
@rtimms this implementation differs from what I had thought about doing initially with a metaclass and `@register_citation`, but it does seem to work as intended – now that I see it, @brosaplanella's idea of adding a table for even simpler output (https://github.com/pybamm-team/PyBaMM/issues/1182#issue-711744118) can be done too (though probably in a different PR?)

tinosulzer (approved):
Looks good to me, thanks @agriyakhetarpal !

agriyakhetarpal:
Ready to merge once tests pass

tinosulzer:
why did you skip ci?

agriyakhetarpal:
In the last commit I only added comments to improve coverage, I did not know that skipping CI this way would not run the coverage workflow

agriyakhetarpal:
I'll do a force push to re-run them if that would be okay

tinosulzer:
Yeah you have to run all the tests to get CI, it's usually worth doing anyway unless you really did only add extra tests. Still need to fix one line of coverage

agriyakhetarpal:
Thanks, I fixed the coverage – `tag_citations()` is available as a standalone method as of now but can be set to private later since it is wrapped within the current `print` methods too. 

Though I don't know how to fix `no-method-argument (E0211)` on Codacy, because changing it from `_caller()` to `_caller(self)` will not run it on the same code object included in the outer frame but execute it on itself, which breaks the intended functionality with `inspect.stack` (I tested on ipython and I was having issues, it was not tagging the correct class). Otherwise this should be fine to merge now

PR #2867, opened 2023-04-07
Issue 2725 reduce Windows and macOS tests and run all tests nightly

### LINKED ISSUE #2725: Reduce testing on Windows and Mac
I'm wondering if there's any reason to run the integration tests on Windows and Mac. The unit tests should be sufficient (and maybe even for Linux python 3.8). If we then moved the integration and examples testing on 3.9 only to new processes in parallel, the whole test suite would be much faster.

agriyakhetarpal:
# Description

- `test_on_push.yml` will now run integration tests only on GNU/Linux with Python 3.11; unit tests will run on all platforms – with 3.11 on GNU/Linux and with Python 3.8, 3.9, and 3.10 on Windows and macOS
- Adds a workflow `run_periodic_tests.yml` to run all unit tests and integration tests on all platforms and Python versions, scheduled at 3 am UTC and on PRs to the `main` branch 
- coverage gets generated with Python 3.11 instead of Python 3.9 (on both workflows)

Fixes #2725 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/.github/workflows/run_periodic_tests.yml
+++ b/.github/workflows/run_periodic_tests.yml
@@ -1,0 +1,128 @@
+# Run all unit tests and integration tests for all Python versions 
+# and platforms at 3am UTC every day and on PRs to the main branch
+name: PyBaMM
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches: 
+    - main
+
+  # Run everyday at 3 am UTC
+  schedule:
+    - cron: "0 3 * * *"
+
+jobs:
+  pre_job:
+    runs-on: ubuntu-latest
+    # Map a step output to a job output
+    outputs:
+      should_skip: ${{ steps.skip_check.outputs.should_skip }}
+    steps:
+      - id: skip_check
+        uses: fkirc/skip-duplicate-actions@master
+        with:
+          # All of these options are optional, so you can remove them if you are happy with the defaults
+          concurrent_skipping: "never"
+          cancel_others: "true"
+          paths_ignore: '["**/README.md"]'
+
+  style:
+    needs: pre_job
+    if: ${{ needs.pre_job.outputs.should_skip != 'true' }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Setup python
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.9
+
+      - name: Check style
+        run: |
+          python -m pip install pre-commit
+          pre-commit run ruff
+
+  build:
+    needs: style
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: ["3.8", "3.9", "3.10", "3.11"]
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install Linux system dependencies
+        if: matrix.os == 'ubuntu-latest'
+        run: |
+          sudo apt-get update
+          sudo apt install gfortran gcc libopenblas-dev graphviz
+          sudo apt install texlive-full
+
+      # Added fixes to homebrew installs:
+      # rm -f /usr/local/bin/2to3
+      # (see https://github.com/actions/virtual-environments/issues/2322)
+      - name: Install MacOS system dependencies
+        if: matrix.os == 'macos-latest'
+        run: |
+          rm -f /usr/local/bin/2to3*
+          rm -f /usr/local/bin/idle3*
+          rm -f /usr/local/bin/pydoc3*
+          rm -f /usr/local/bin/python3*
+          brew update
+          brew install graphviz
+          brew install openblas
+
+      - name: Install Windows system dependencies
+        if: matrix.os == 'windows-latest'
+        run: choco install graphviz --version=2.38.0.20190211
+
+      - name: Install standard python dependencies
+        run: |
+          python -m pip install --upgrade pip wheel setuptools
+          python -m pip install "tox<4"
+
+      - name: Install SuiteSparse and Sundials
+        if: matrix.os == 'ubuntu-latest'
+        run: tox -e pybamm-requires
+
+      - name: Run unit tests for GNU/Linux with Python 3.8, 3.9, and 3.10
+        if: matrix.os == 'ubuntu-latest' && matrix.python-version != 3.11
+        run: python -m tox -e unit
+
+      - name: Run unit tests for GNU/Linux with Python 3.11 and generate coverage report
+        if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.11
+        run: | 
+          python -m tox -e unit
+          python -m tox -e coverage
+
+      - name: Upload coverage report
+        if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.11
+        uses: codecov/codecov-action@v2.1.0
+
+      - name: Run integration tests for GNU/Linux
+        if: matrix.os == 'ubuntu-latest'
+        run: python -m tox -e integration
+
+      - name: Run unit tests for Windows and MacOS
+        if: matrix.os != 'ubuntu-latest'
+        run: python -m tox -e mac-windows-unit
+
+      - name: Run integration tests for Windows and MacOS
+        if: matrix.os != 'ubuntu-latest'
+        run: python -m tox -e mac-windows-integration
+
+      - name: Install docs dependencies and run doctests
+        if: matrix.os == 'ubuntu-latest'
+        run: tox -e doctests
+
+      - name: Install dev dependencies and run example tests
+        if: matrix.os == 'ubuntu-latest'
+        run: tox -e examples
--- a/.github/workflows/test_on_push.yml
+++ b/.github/workflows/test_on_push.yml
@@ -4,10 +4,6 @@ on:
   workflow_dispatch:
   pull_request:
 
-  # everyday at 3 am UTC
-  schedule:
-    - cron: "0 3 * * *"
-
 jobs:
   pre_job:
     runs-on: ubuntu-latest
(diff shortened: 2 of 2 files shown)


agriyakhetarpal:
Maybe it would also make sense to have separate workflow files for different platforms with all Python versions? Because in this case, it currently shows builds being completed for a particular platform and Python version, say, macOS/Windows with Python 3.10 – it is only on a closer look via the Actions page that one finds the unit tests were not executed at all for that platform or version; while Ubuntu will run the tests on Python 3.11. Likewise, separate workflows for different Python versions with all platforms can also be considered

.github/workflows/run_periodic_tests.yml
```diff
@@ -0,0 +1,128 @@
+# Run all unit tests and integration tests for all Python versions 
+# and platforms at 3am UTC every day and on PRs to the main branch
+name: PyBaMM
```
Saransh-cpp, line 3: Maybe something like -
```suggestion
name: Scheduled tests
```

.github/workflows/run_periodic_tests.yml
```diff
@@ -0,0 +1,128 @@
...
+  style:
+    needs: pre_job
+    if: ${{ needs.pre_job.outputs.should_skip != 'true' }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Setup python
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.9
```
Saransh-cpp, line 39: `3.11` is faster. I am not sure if we will get speed up here, but we can try.
```suggestion
          python-version: 3.11
```

.github/workflows/run_periodic_tests.yml
```diff
@@ -0,0 +1,128 @@
...
+
+      - name: Run unit tests for GNU/Linux with Python 3.8, 3.10, and 3.11
+        if: matrix.os == 'ubuntu-latest' && matrix.python-version != 3.9
+        run: python -m tox -e unit
+
+      - name: Run unit tests for GNU/Linux with Python 3.9 and generate coverage report
+        if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.9
+        run: | 
+          python -m tox -e unit
+          python -m tox -e coverage
```
Saransh-cpp, line 104: `tox -e coverage` triggers `coverage run -m unittest` behind the scenes. `coverage.py` runs the unit tests with its own interpreter and calculates the coverage value. No need to run unit tests before this :)
```suggestion
          python -m tox -e coverage
```

.github/workflows/test_on_push.yml
```diff
@@ -95,28 +91,26 @@ jobs:
 
       - name: Run unit tests for GNU/Linux with Python 3.9 and generate coverage report
         if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.9
-        run: tox -e coverage
+        run: |
+          python -m tox -e unit
+          python -m tox -e coverage
```
Saransh-cpp, line 96: Ditto!
```suggestion
          python -m tox -e coverage
```

.github/workflows/test_on_push.yml
```diff
@@ -95,28 +91,26 @@ jobs:
...
-      - name: Run integration tests for GNU/Linux
-        if: matrix.os == 'ubuntu-latest'
+      - name: Run integration tests for GNU/Linux with Python 3.11
+        if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.11
         run: python -m tox -e integration
 
-      - name: Run unit tests for Windows and MacOS
-        if: matrix.os != 'ubuntu-latest'
+      - name: Run unit tests for Windows and MacOS with Python 3.8, 3.9, and 3.10
+        if: matrix.os != 'ubuntu-latest' && matrix.python-version != 3.11
```
Saransh-cpp, line 107: Why are we not running unit tests for python 3.11 on Windows and Mac?

.github/workflows/run_periodic_tests.yml
```diff
@@ -0,0 +1,128 @@
...
+
+      - name: Upload coverage report
+        if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.9
```
Saransh-cpp, line 107: Any specific reasons for using 3.9 to generate and upload coverage reports? We can try 3.11 to check if there are any speed-ups.

Saransh-cpp (changes requested):
Thanks, @agriyakhetarpal! Looks great overall! Some minor comments below -

.github/workflows/run_periodic_tests.yml
brosaplanella: I agree this would be more informative, but I think it might also change the name in the badge we have in the README. Not sure if there is a way to have another name in the badge.

.github/workflows/run_periodic_tests.yml
agriyakhetarpal: Thanks!

.github/workflows/run_periodic_tests.yml
agriyakhetarpal: I used 3.9 because generating coverage with 3.11 failed on a previous run  – you may see the logs here: https://github.com/pybamm-team/PyBaMM/actions/runs/4640011359. I think it is an issue with `pybamm_install_jax`, though I haven't looked at why it failed in detail

.github/workflows/test_on_push.yml
agriyakhetarpal: According to https://github.com/pybamm-team/PyBaMM/issues/2725#issuecomment-1498202987, I ran unit + integration tests on Linux for 3.11, and unit tests only for 3.8 – 3.10 for macOS and Windows; so yes, maybe I should include 3.11 too since it is not being tested at all

.github/workflows/run_periodic_tests.yml
agriyakhetarpal: Maybe we can rename it to just `CI`

.github/workflows/test_on_push.yml
Saransh-cpp: Yes, I think it should be included too.

Saransh-cpp (commented):
Looks great, thanks for this @agriyakhetarpal! Minor comments below (maybe I am nitpicking too much).

Besides the comments below, we might have to think about a short but descriptive name for the workflow (as it will appear on the badge). We should also add a badge for the scheduled CI in README, but I think the markdown code for the badge will only appear once we merge this in.

.github/workflows/test_on_push.yml
```diff
@@ -95,28 +91,25 @@ jobs:
...
+          python -m tox -e coverage
 
       - name: Upload coverage report
         if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.9
         uses: codecov/codecov-action@v2.1.0
 
-      - name: Run integration tests for GNU/Linux
-        if: matrix.os == 'ubuntu-latest'
+      - name: Run integration tests for GNU/Linux with Python 3.11
+        if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.11
```
Saransh-cpp, line 102: This is right now failing while installing SUNDIALS and SuiteSparse. But, I think this will fail even if the SUNDIALS step passes because of `jax` and `jaxlib`. We might have to make some changes to `tox.ini` to ensure that `pybamm_install_jax` runs only on `3.8` and `3.9`.

`JaxSolver` is currently not being tested on Python `3.8` as well, but it should be. If you look closely -
- `[testenv:coverage]` installs `jax`, making it available while executing `tox -e coverage`
- The base `[testenv]` only installs `jax` on `tests-!windows-!mac`, which isn't used on CI anywhere, rather it should install `jax` in `integration` and `unit` but only on Python `3.8` and `3.9`.

I'll look into the `jax` and `jaxlib` thing.

.github/workflows/run_periodic_tests.yml
agriyakhetarpal: I may mention that in this PR if it is required since it is only a minor change

.github/workflows/test_on_push.yml
```diff
@@ -95,28 +91,25 @@ jobs:
...
         run: python -m tox -e integration
 
-      - name: Run unit tests for Windows and MacOS
+      - name: Run unit tests for Windows and MacOS with Python 3.8, 3.9, 3.10, and 3.11
```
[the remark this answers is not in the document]
agriyakhetarpal: Makes sense

.github/workflows/test_on_push.yml
agriyakhetarpal: That sounds right, we are indeed missing 3.8 with `JaxSolver` – though I'll leave this over to you like you ask :)

agriyakhetarpal: I think the previous run wasn't failing due to jaxlib but due to some error with `wget` internally, SUNDIALS and SuiteSparse were installed successfully on this one

.github/workflows/run_periodic_tests.yml
Saransh-cpp: We can maybe name it just "Scheduled"?

Saransh-cpp (commented):
Thanks, @agriyakhetarpal! Looks great, sorry for the delay, was caught up with college. Very minor comments -

.github/workflows/run_periodic_tests.yml
agriyakhetarpal: Yes, that works

.github/workflows/test_on_push.yml
agriyakhetarpal: Sure, I'll add a note about `JaxSolver` in the docs for installing PyBaMM from source. Please let me know if I need to add it anywhere else too

Saransh-cpp (approved):

docs/source/user_guide/installation/GNU-linux.rst
```diff
@@ -125,6 +125,9 @@ Optional - JaxSolver
 Users can install ``jax`` and ``jaxlib`` to use the Jax solver.
 Currently, only GNU/Linux and macOS are supported.
 
+.. note::
+   The Jax solver is currently available only on Python 3.8 and 3.9.
+
```
Saransh-cpp, line 130: Not required anymore, thanks to #2927!
```suggestion
```

Saransh-cpp (changes requested):

Saransh-cpp (approved):
Thanks, @agriyakhetarpal!

Also, this might create some conflicts (both literally and thought process wise) with #2929, just a heads-up @BradyPlanden @martinjrobins. (I'm not sure which one should be merged first 😅)

Saransh-cpp:
Here is the new badge - [Scheduled](https://github.com/pybamm-team/PyBaMM/actions/workflows/run_periodic_tests.yml)

We might want to replace the original badge with this badge in the README. @agriyakhetarpal could you please create a quick PR for this too? Thanks!

agriyakhetarpal:
Sure, on it!

Issue #2858, opened 2023-04-05
Updates for casadi 3.6


tinosulzer:
Release notes: https://github.com/casadi/casadi/releases/tag/3.6.0

Will add as found

- time grid points should be passed when creating interpolator instead of when solving
- rootfinder errors

tinosulzer:
I don't know how to fix the failing sensitivity test so skipping for now, @martinjrobins @jsbrittain could you take a look at this?

martinjrobins:
sure, this is also reported in https://github.com/pybamm-team/PyBaMM/issues/2789#issuecomment-1482739316.

tinosulzer:
I think it might be a different error here

martinjrobins:
Did PR #2883 fix the problem? I just tried `test_sensitivities` on spm, spme and dfn with idaklu solver and it seemed to work ok.

martinjrobins:
ahh ok, I wasn't aware you'd added that

martinjrobins:
near as I can figure this is the problem. Say you have a model like this

$$\frac{du}{dt} = - a * u$$

$$ v = u $$

$$ u(0) = 1$$

$$ v(0) = 1$$

For $a=1$, sundials calculates the initial condition for $s = \frac{dv}{da}$ as $s(0) = [0, 0]$, presumably because v(0) does not depend on $a$, but if you correctly write the initial condition for $v$ as:

$$\frac{du}{dt} = - a * u$$

$$ v = u $$

$$ u(0) = 1$$

$$ v(0) = a$$

Then sundials finds the initial condition should be $s(0) = [0, 1]$. Not exactly sure what the expected behaviour is really, or why sundials isn't calculating the correct ic for $s$ even when you don't give the correct symbolic form for $v(0)$

tinosulzer:
$s(0) = [0, 1]$ is the correct initial condition in the second case, right?

martinjrobins:
> s(0)=[0,1] is the correct initial condition in the second case, right?

yup, and I believe it should be for the 1st as well, so that is why I'm a bit confused that the `calcic` function in sundials (ie. calculate consistent initial conditions) isn't finding it, it just returns `[0, 0]` for the 1st and `[0, 1]` for the 2nd.

martinjrobins:
ahh ok. You have to call `IDAGetSens` after `IDACalcIC` in order for the solution to actually show up in your sensitivity vector..... fixed now

PR #2883, opened 2023-04-19
#2789 add "all" to sensitivities for idaklu solver

### LINKED ISSUE #2789: [Bug]: Short discharge pulse in drive cycle partially ignored
### PyBaMM Version

v23.2

### Python Version

3.8.10

### Describe the bug

I am simulating a drive cycle with a 1C discharge for a couple of minutes, followed by a rest phase, followed by a 10 second 1C discharge (beginning of a typical HPPC test). I am using a drive cycle rather than an Experiment as I am fitting to experimental data.

I noticed that, during the short pulse, the predicted voltage is not as expected and traced it back to the absence of any diffusive contribution; in fact the concentration in the electrodes does not change during the pulse. So it looks like part of the simulation (diffusion equation) ignores the presence of the pulse.

### Steps to Reproduce

(Edit: modified the MWE to make it occur even with safe Casadi solver)

MWE:

```python
import numpy as np
import pybamm

Q = model.default_parameter_values["Nominal cell capacity [A.h]"]

time = np.arange(0, 1100, 2, float)
current = 0 * time
current[(time <= 500)] = Q
current[(time >= 1000) & (time <= 1010)] = Q

param = model.default_parameter_values.copy()
param["Current function [A]"] = pybamm.Interpolant(
    time, current, model.timescale * pybamm.t
)

sim = pybamm.Simulation(
    model,
    parameter_values=param,
    solver=pybamm.CasadiSolver(atol=1e-3, rtol=1e-3, mode="safe"),
)
sol = sim.solve()
```

Plot:

```python
plt.plot(time, current)
plt.plot(sol["Time [s]"].entries, sol["Current [A]"].entries, ls="--")
plt.show()

for domain in ["positive", "negative"]:
    plt.plot(
        sol["Time [s]"].entries, sol[f"Average {domain} particle concentration"].entries
    )
plt.show()
```

Output: 

image

So while the current correctly shows the short peak, it has no impact on concentration.

Note that by playing with the length of the discharge and rest phases, one can make the issue appear and disappear.

### Relevant log output

_No response_

martinjrobins:
# Description

Partially fixes #2789 and unblocks #2858 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Bug fixes
 
 - Fixed a bug in the discretisation of initial conditions of a scaled variable ([#2856](https://github.com/pybamm-team/PyBaMM/pull/2856))
+- Fixed keyerror on "all" when getting sensitivities from IDAKLU solver([#2883](https://github.com/pybamm-team/PyBaMM/pull/2883))
 
 # Breaking changes
 
--- a/pybamm/solvers/idaklu_solver.py
+++ b/pybamm/solvers/idaklu_solver.py
@@ -532,7 +532,8 @@ class IDAKLUSolver(pybamm.BaseSolver):
         y_out = sol.y.reshape((number_of_timesteps, number_of_states))
 
         # return sensitivity solution, we need to flatten yS to
-        # (#timesteps * #states (where t is changing the quickest),) to match format used by Solution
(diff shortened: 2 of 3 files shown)


tinosulzer (approved):
Thanks!

PR #2923, opened 2023-05-03
2447 cache Linux (`apt`) dependencies in CI

### LINKED ISSUE #2447: cache github actions
Installing linux dependencies takes >10min and mac takes >5min. This is done every single time the github actions are run. Can this be cached somehow? @priyanshuone6

### LINKED ISSUE #2632: cache apt packages in ci
### Description

@tinosulzer reports that a large part of the testing ci is installing the apt package dependencies, perhaps we can cache these to speed things up? e.g. use this action:

https://github.com/marketplace/actions/cache-apt-packages

### Motivation

speed up the ci

### Possible Implementation

_No response_

### Additional context

_No response_

agriyakhetarpal:
# Description

Caches dependencies for faster installation and setup of PyBaMM's dependencies via the `apt` package manager on Linux

Fixes #2447 
Fixes #2632

### Tasks

- ~Cache `pip` dependencies~ (`setup-python` and `actions/cache` do not provide any benefits since most of our dependencies are installed within `Tox` environments)
- [x] Setup a cache for `apt` dependencies (partially done)
- [x] Investigate caching for `brew` dependencies (unhelpful)

## Type of change

# Key checklist:

## Further checks:


agriyakhetarpal:
We could also look into caching entire [pip installations](https://blog.allenai.org/python-caching-in-github-actions-e9452698e98d) as well and their related directories, but I am not doing this for now may cause issues when updating Python versions or with dependencies that need post-install configuration steps (relevant discussion on https://github.com/actions/setup-python/issues/330)

tinosulzer:
How do we know if this is working?

agriyakhetarpal:
We will need to compare the runtime for the step `Setup Python ${{ matrix.python-version }}` in the jobs with that of previous runs, and similarly for other package managers. It did not find any cache in most of the runs right now but I guess it should do so after a few iterations

agriyakhetarpal:
The `awalsh/cache-apt-pkgs-action` took just short of 8 minutes to install dependencies, and 5 minutes to cache them – with a cache size of 3 GB. The installation time was the same in previous runs, but I expect it to improve in further pushes

agriyakhetarpal:
The `awalsh/cache-apt-pkgs-action` took just short of 8 minutes to install dependencies, and 5 minutes to cache them – with a cache size of 3 GB. The installation time was the same in previous runs, but I expect it to improve in further pushes with the cache now generated

agriyakhetarpal:
@Saransh-cpp installing `OpenBLAS` explicitly and without cache helps pass the `pybamm-requires` step, but it fails the `latexify` and `visualise.png` unit tests – is there a way to make them optional?
 
Using the cache now manages to shave off 7-8 minutes when installing Linux dependencies, which is a significant boost (8-9 minutes comes down to 1.5 minutes).  Assuming that the error is related to caching `texlive-full`, I tried installing `OpenBLAS` and `texlive-full` explicitly but it happens to reverse the benefits we gain completely. I will look into how we can cache `texlive-full` properly (it seems it takes the longest time to install amongst all the dependencies).

Anyhow; as reported by most `Homebrew` caching techniques I could find, it is unlikely to speed up our CI. Also, it is to be noted that the Homebrew cache size is exceeding `actions/cache@v3`'s 5GB size limit every time – this might be the reason why the cache is being built at each workflow run (cache beyond the limit probably gets evicted).

agriyakhetarpal:
I started working on this once more today since the issues with `pybamm-requires` and `OpenMP` have now been resolved upstream, and I think this PR is ready for review now and I shall mark it so. The caching here does have a few caveats and inconsistencies though, which I learned through my experience with several attempts and experiments with action runs to test it. Here is a summary:

1. I tried an experimental solution for caching `Homebrew` packages by taking hints from workflows in other repositories, but the general accord is that it is unlikely to speed up our CI – this is because Homebrew formulae are small in size, so any caching solution with fast downloads from a CDN plus longer installation times would making caching redundant (the installation time will remain mostly the same), which is the case here. It is now taking extra time in the installation steps to cache `Homebrew` installation locations, for which I am unable to receive a proper cache hit.

    - The `Homebrew` environment variables are helpful and are nice-to-haves; they mostly speed up the command line by disabling analytics and cleanup processes (the latter happens every time an installation command succeeds).
    
    - We can probably remove the Homebrew caching altogether since it does not offer any real benefits?

2. Caching with `pip` using the `setup-python` GitHub Action currently does not install the packages but only caches downloads (wheels), which is a known issue – https://github.com/actions/setup-python/issues/330. Also, most of our development workflow is dependent on `tox` (and will change soon to `nox` in #3005), which means we would need to look for caching `nox` environments since almost all of our standard project-specific dependencies are installed within them. 

    - Using `actions/cache@v3` is not a solution either because the aforementioned action uses it under the hood. In previous runs with the cache enabled, setting Python up took ~5–10 seconds. 

    - On the other hand, setting Python up without a cache does not take more than 3 seconds on any run, so `pip` caching is disadvantageous in its current state

3. Caching `apt` packages with [`awalsh128/cache-apt-pkgs-action`](https://github.com/awalsh128/cache-apt-pkgs-action) also causes problems with certain packages, especially ones that have pre-installation and post-installation steps/scripts (there is a parameter to execute_install_scripts but it is not foolproof): 

    - I tried caching all packages first (`gcc`, `gfortran`, `libopenblas-dev`, and `texlive-full`) but it was quickly found that the `pybamm-requires` step failed because the runner had trouble finding `BLAS` libraries (I might need to set environment variables for its install location, but I'm not sure where exactly `apt` packages get installed on Linux). We might need to use [`ccache`](https://ccache.dev/) but I am unaware of an approach yet since we don't use `CMake` directly, we just use it through the `install_KLU_sundials` script that calls it.
    
    - It is to be noted that `graphviz` also has a post-installation step to set its fonts and plugins with `dot -c` after it is loaded from a previous cache.
    
    - I identified that the `texlive-full` package is one of the largest packages that get installed with more than 4 GB of occupied space (mostly because of its huge collection of extra fonts and documentation) which is not needed, so I replaced it with a medium-sized build of `TeXLive`, i.e., `texlive-latex-extra` and a few other packages such as `divpng` from [CTAN](https://ctan.org/). These packages are not included in this much smaller build (450 MB) but can be installed externally to account for missing features. `texlive-full` took ~8–10 minutes to collect its sub-dependencies and install every time, which now gets truncated to ~1 minute on average with `texlive-latex-extra`. Here is a comparison of different `TeXLive` builds and some dependency graphs for more explanation: https://tex.stackexchange.com/q/245982. I tried to cache `TeXLive` too, unsuccessfully (there were post-installation scripts for both `texlive-full` and `texlive-latex-extra`).
    
    - Nevertheless, the cache created without `texlive-full` (i.e., consisting of `gcc`, `gfortran`, and `graphviz`, since `libopenblas-dev` is installed separately) is barely 50 MB in size and does not offer any benefits either; there is a reduction of half a minute at most.

So this PR adds (i) a very minute speedup for macOS (`brew`) dependencies and (ii) a substantial speedup for Linux (`apt`) dependencies. I am glad to implement a cache for `choco` dependencies if required, though the only dependency on Windows is `graphviz` which gets installed fairly quickly.

.github/workflows/periodic_benchmarks.yml
```diff
@@ -27,12 +27,14 @@ jobs:
...
           sudo apt-get update
-          sudo apt install gfortran gcc libopenblas-dev
+          sudo apt-get install gfortran gcc libopenblas-dev
       - name: Install python dependencies
         run: |
           python -m pip install --upgrade pip wheel setuptools virtualenv asv wget cmake casadi numpy
       - name: Install SuiteSparse and Sundials
         run: python scripts/install_KLU_Sundials.py
+      - name: Install tox and asv
+        run: pip install -U pip "tox<4" asv
```
Saransh-cpp, line 37: I think this should not be here? We are installing `asv` in one of the steps above and I don't think this workflow uses `tox`.

.github/workflows/test_on_push.yml
```diff
@@ -71,12 +88,26 @@ jobs:
...
+        uses: actions/cache@v3
+        if: matrix.os == 'macos-latest'
+        with:
+          # paths for Homebrew packages, formulae, casks, and binaries
+          path: |
+            /usr/local/Homebrew/*
+            /usr/local/Cellar/*
+            /usr/local/etc/*
+            /usr/local/opt/*
+          key: brew-cache-${{ runner.os }}-${{ hashFiles('**/Brewfile.lock.json') }}
```
Saransh-cpp, line 104: > We can probably remove the Homebrew caching altogether since it does not offer any real benefits?

Yep, we can get rid of this then. I am guessing the one minute speedup is due to the homebrew environment flags and not this step?

Saransh-cpp (approved):
This looks awesome, thanks for the detailed research on this, @agriyakhetarpal!

I went through the explanation above and it looks like you've sorted the Linux dependencies. We can skip the MacOS and Windows caching part for now (until there is a native GH Actions solution or a working third-party solution). We can also skip caching Python packages as the workflow currently only caches the wheel file.

Some very minor comments below -

.github/workflows/periodic_benchmarks.yml
agriyakhetarpal: Apologies, I don't know where and when I added this, or how, even. I have been working on this PR for too long haha

agriyakhetarpal:
Thanks for the review! For Pythonic dependencies, we do have another solution via caching the entire Python virtual environment that I had shared earlier in the discussion ([this blog](https://blog.allenai.org/python-caching-in-github-actions-e9452698e98d)) and since `tox` (or `nox`) also creates environments for installing PyBaMM and running the test suite, I will open another PR to try and cache the environments themselves after we migrate to `nox` in #3005. This way we can cache the `lib/site-packages` subfolders too, which should work in theory, and might save us some time since the `dev` and `docs` environments install a lot of packages.

.github/workflows/test_on_push.yml
agriyakhetarpal: Though, by "a very minute speedup", I was referring to "minute" as in _small_, not that it caused a one-minute speedup 😆

Saransh-cpp: Ah, that makes it clear, thanks!

PR #2943, opened 2023-05-10
Make the "partial molar volume" a function of stoichiometry

abillscmu:
# Description
For many chemistries, notably high-nickel NMCs and graphite, the volume change associated with intercalation is a strong function of stoichiometry. Here, I have taken the r-averaged value for Ω and used it to calculate total stress. This accounts for the change without violating the initial model assumptions.

## Type of change

# Key checklist:

## Further checks:


tinosulzer (commented):
Thanks! 

There is already a parameter, `t_change`, for the electrode thickness change as a function of particle concentration. Should that be the same parameter?

abillscmu:
They are closely related, but not the same. In most cases, `t_change` is a function of `Omega`, but `t_change` is more of a summary variable to monitor volume changes of the cell while `Omega` is used to calculate the stress within a particle. According to Appendix B of Ai 2020 (which based on comments is the paper around which pybamm's stress models are implemented), this is to avoid numerical difficulties.

abillscmu:
I think we could get rid of t_change with this, but it would require us to change the parameter sets to be different from the papers from which they were created.

abillscmu:
Closed in favor of #3576

PR #2959, opened 2023-05-16
Issue 2901 parameters csv

### LINKED ISSUE #2901: Deprecate old ways of creating ParameterValues
I think the transition to using python files or json file (BPX) for parameter values has gone fairly well so we can start to deprecate some of the old methods for importing values from csv files. These are no longer recommended but still possible from the codebase. We should also update the training videos on this very soon.
- [ ] remove functionality to update from chemistry (old approach with separate files for anode/cathode/electrolyte etc)
- [ ] remove functionality to load from csv
- [ ] remove functionality to load a function from a string "[function]function_name"
- [ ] remove functionality to load data from a string "[data]data_name"

Basically, users will be responsible for passing in a python dictionary where functions are python functions. Interpolants can be created as functions [like this](https://github.com/pybamm-team/PyBaMM/blob/18d4e84d07a8afb1d8a71081c2912f486b11283d/pybamm/input/parameters/lithium_ion/Ai2020.py#L501-L517)

This is a breaking change so let's discuss before committing

tinosulzer:
# Description

In recent versions of PyBaMM, we recommend using python files to specify parameter sets (as opposed to csv files in early versions). This PR removes all the functionality that was included to make loading from csv files possible. Loading from csv files is no longer possible, and Python files should now always be used instead.

Fixes #2901 

## Type of change

# Key checklist:

## Further checks:


tests/unit/test_parameters/test_parameter_values.py
```diff
@@ -143,6 +89,10 @@ def test_update(self):
         with self.assertRaisesRegex(KeyError, "Cannot update parameter"):
             param.update({"b": 1})
 
+        # udpate with a ParameterValues object
```
brosaplanella, line 92:
```suggestion
        # update with a ParameterValues object
```

brosaplanella (approved):
Looks good, thanks! You need to add to the CHANGELOG before merging.

Issue #2908, opened 2023-04-30
`inspect` causing big performance regression


tinosulzer:
<img width="1008" alt="Screenshot 2023-04-30 at 1 17 59 PM" src="https://user-images.githubusercontent.com/20817509/235366967-24401e33-17b2-47af-9b71-767c4afccae7.png">
<img width="1246" alt="Screenshot 2023-04-30 at 1 18 42 PM" src="https://user-images.githubusercontent.com/20817509/235366978-602e5d28-1aa7-4d53-b68c-2b99e13494ba.png">

This is due to #2862 

@agriyakhetarpal can you look into a way to make this faster? Otherwise we should revert, it's not worth the functionality gain

tinosulzer:
This is a huge performance hit in some cases (e.g. repeated solving of the SPM is not 10x slower) so we should fix immediately and make a 23.4.1 release
<img width="1251" alt="Screenshot 2023-04-30 at 1 23 01 PM" src="https://user-images.githubusercontent.com/20817509/235367174-c1bed33d-b8c9-48a3-bcdf-9da7ce7c8fbd.png">

Maybe this helps https://stackoverflow.com/questions/17407119/python-inspect-stack-is-slow

tinosulzer:
Also weird that the benchmarks never caught this, are they working now @Saransh-cpp ?

agriyakhetarpal:
Since this is immediate I am happy to revert — though I will get back on this and try some other faster implementation in the next week, wasn't aware of such a potential slowdown 😅 Where can I run the above profiler?

tinosulzer:
It's snakeviz, you can pip install then run in `ipython` terminal 
```
> %load_ext snakeviz
> %snakeviz function_to_profile()
```

tinosulzer:
Potential solutions are:
- find a faster implementation of doing the same thing
- use some form of caching so this is only calculated once per `register` call
- add tagging manually as an input to the `register` function
- remove altogether

agriyakhetarpal:
I agree with removing altogether and releasing a patch for now.

- We can indeed explore a caching option, albeit will some overhead still remain when a model is called and its citation is registered for the first time when running an experiment? In that case a faster `inspect.stack()` implementation altogether would be better or by modifying all `__call__` methods with a metaclass

- Preserving a `dict` of models mapped to their citations hardcoded manually is also a solution 

[...]

PR #2961, opened 2023-05-16
Issue 1182 faster tagging for PyBaMM citations 

### LINKED ISSUE #1182: Tag PyBaMM citations

agriyakhetarpal:
# Description

Enables a faster method to tag citations based on where they come up in the code during a simulation (models, solvers, etc.) – via a `caller_name` method that calls `sys._getframe()` to access current frames for the `pybamm.citations.register()` function. 

This PR continues the discussion on #2862 and #2908, with some results of the speedup available at https://github.com/pybamm-team/PyBaMM/issues/2908#issuecomment-1537077212. Here, `sys._getframe` is preferred to `inspect._getframe()` for avoiding the internal `hasattr()`, in order to reduce any potential performance impact(s) of attribute checks at runtime.

Fixes #1182 

## Type of change

# Key checklist:

## Further checks:


agriyakhetarpal:
@tinosulzer this is a discussion of somewhat different scope, so I am commenting here instead of on the previous thread – I notice that the `pybamm.citations.register()` function is calling itself recursively when it tries to parse the citation with `pybtex` and add it to the `_all_citations` dictionary. This adds to the overhead if a particular model (or solver) registers multiple citations, either explicitly or through its base class due to the use of `super().__init__`. In the earlier profiling results on #2908, it looks like getting the current frame with `_caller_name()` amounts to only a fraction of the total time for `register()` (which just becomes `_caller_name()` + parsing the citations with `pybtex`) – about 30-50% without caching and around 10% if `functools` `@cache` is used with it (if I am thinking about this correctly – I guess this will require some more profiling, with a larger experiment perhaps).

We can move the citation parsing functionality to a new, separate function that only gets called inside `print_citations()`, and keep only the tagging (that cannot be moved outside `register()` since getting the current frame will depend on initialising a model or solver with `register()` at runtime). This way, registering citations would affect performance slightly lesser when solving a model, which is a critical feature. On the other hand, the `print_citations` function will compensate for it and get slightly slower, which should be fine. I am happy to make these changes in this PR itself after further discussion

agriyakhetarpal:
New profiling results: now `_caller_name()` takes more time within `register()` since parsing the citation with `pybtex` has been moved to a separate function `parse_citation()`, which gets called inside `print_citations()` instead

```python
# For the SPM model
%load_ext snakeviz
%snakeviz sim.solve([0, 3600])
```

Solving for the first time

image

and solving for the second time (most things get cached)

image

I am unsure why the ncalls are the same here as the previous https://github.com/pybamm-team/PyBaMM/issues/2908#issuecomment-1550213890 though. I am looking for ways to cache `_caller_name()` at the moment, we should not use `@lru_cache` on functions though, so something else must be needed to make this faster.

agriyakhetarpal:
Linking the citation tags hardcoded to their respective models or solvers remains the fastest solution because 1. It does not need to be executed at runtime, and 2. read/write operations to dictionaries or lists within the same file should be very fast

The only disadvantage this has is that it is neither programmatic nor Pythonic

agriyakhetarpal:
Lol the solution to not calling `_caller_name()` was so obvious, but it didn't strike me until now. Just a `try-except` block to not call it if a citation tag already exists 😆

Newer profiling results are much better (the above + parsing registered citations only when printing them): 

<details><summary>DFN</summary>
<p>

```python
# For the DFN model
%load_ext snakeviz
%snakeviz sim.solve([0, 3600])
%snakeviz sim.solve([0, 3600])
```

`sim.solve([0, 3600])` first iteration

image

`sim.solve([0, 3600])` second iteration

image

Impact on the DFN model is negligible and it was not impacted in earlier results either, so testing SPM is more important

</p>
</details>

<details><summary>SPM</summary>
<p>

```python
# For the SPM model
%load_ext snakeviz
%snakeviz sim.solve([0, 3600])
%snakeviz sim.solve([0, 3600])
```

`sim.solve([0, 3600])` first iteration

image

`sim.solve([0, 3600])` second iteration

image

i.e., never going past ~0.5-0.6% performance impact on any further iterations, was tested by repeating the same code 10 times. 

</p>
</details> 

Do these look good?

tinosulzer (approved):

PR #2970, opened 2023-05-19
save variables for individual lithium plating terms

tinosulzer:
# Description

Add some variables for each term in the lithium plating equation to see different contributions

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/models/submodels/interface/lithium_plating/base_plating.py
+++ b/pybamm/models/submodels/interface/lithium_plating/base_plating.py
@@ -61,6 +61,21 @@ class BasePlating(BaseInterface):
                 }
             )
 
+            for term in ["stripping", "plating", "dead lithium"]:
+                j = variables[
+                    f"Lithium plating interfacial current density ({term} term) [A.m-2]"
+                ]
+                a_j = a * j
+                a_j_av = pybamm.x_average(a_j)
+                variables.update(
+                    {
+                        "Lithium plating volumetric interfacial current density "
+                        f"({term} term) [A.m-3]": a_j,
+                        "X-averaged lithium plating volumetric interfacial current "
+                        f"density ({term} term) [A.m-3]": a_j_av,
+                    }
+                )
+
         zero_av = pybamm.PrimaryBroadcast(0, "current collector")
         zero = pybamm.FullBroadcast(0, "positive electrode", "current collector")
         variables.update(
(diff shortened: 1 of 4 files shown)


DrSOKane (commented):
Hi Tino, this is a good idea and I appreciate you requesting my feedback. 

However, it's important to explain at this point that my equations for partially reversible plating differ from most others in the literature in that no lithium is plated irreversibly. Instead, all lithium is plated reversibly but gradually becomes irreversible over time. So there are two reactions: the plating/stripping reaction and the reversible -> irreversible reaction. The second reaction carries no current. 

So in theory, there is no dead lithium current. In practice there is, because the plated lithium -> dead lithium reaction alters the thermodynamic equilibrium of the plating/stripping reaction in favour of plating, so there is a current that results from the dead lithium. Therefore, the "dead lithium current" is a useful variable, in spite of the fact it doesn't exist.

The challenge is how to explain all this when a user plots the current components and discovers they don't add up!

pybamm/models/submodels/interface/lithium_plating/base_plating.py
```diff
@@ -132,27 +147,48 @@ def _get_standard_concentration_variables(self, c_plated_Li, c_dead_Li):
...
         variables : dict
             The variables which can be derived from the plated lithium thickness.
         """
-        # Set scales to one for the "no plating" model so that they are not required
-        # by parameter values in general
+        j_plating_av = pybamm.x_average(j_plating)
         j_stripping_av = pybamm.x_average(j_stripping)
+        j_dead_lithium_av = pybamm.x_average(j_dead_lithium)
+
+        j_plating_net = j_plating + j_stripping + j_dead_lithium
```
tinosulzer, line 173: @DrSOKane if I understand correctly this is incorrect and the plating current (the one that goes into the total current conservation equation) should just be plating + stripping?

DrSOKane:
@tinosulzer Correct.

Note that the code is agnostic as to the mechanism by which reversible lithium become dead lithium, and I intend to keep it that way for now. If the mechanism is mechanical isolation of Li metal, than no electrons are transferred at all. If it's a chemical mechanism (e.g. 2Li + C3H4O3 -> Li2CO3 + C2H4), two electrons are transferred but they come from the Li itself and not the external circuit. So either way there is no current.

Do you think it's also worth changing the irreversible model to make dead Li instead of plated Li? I'm working on an example notebook for coupled degradation and it may be helpful to define separate variable for irreversible and reversibel capacity loss.

tinosulzer:
I think we should keep the irreversible notebook as is, since it might confuse people why the plated lithium concentration is not changing

DrSOKane:
One way around it could be to define a dead lithium flux [mol.m-3] instead:

`dead_lithium_flux = -dead_lithium_decay_rate * c_plated_Li / a`

i.e. the same as `j_dead_lithium` but without multiplying by F

kratman:
Is this still relevant or can this PR be closed?

Issue #2962, opened 2023-05-16
`pybamm_install_odes` is not tested anywhere


Saransh-cpp:
The script gets outdated very often because it is not being tested anywhere. The script can be tested nightly on Ubuntu once the infrastructure proposed in https://github.com/pybamm-team/PyBaMM/pull/2867 is merged.

arjxn-py:
We want to test `pybamm_install_odes` i.e. run tests associated with `scikit.odes` to avoid it from getting outdated, i believe `jax` & `jaxlib` could also be involved in here. Currently If i would want to test these, i'd have to make changes to the existing infrastructure but we want to test these in the proposed infrastructure (#2867), should I wait for it to be merged or any suggestion @Saransh-cpp, Thanks.

Saransh-cpp:
Maybe not even run tests, just check if `scikits.odes` is getting installed, as the tests are already running using `tox`. `jax` related things are already tested. We don't want to slow down the CI so we can create a new step just for testing `pybamm_install_odes` in the nightly run. I think yes, you should wait for it to be merged.

Saransh-cpp:
We can add a minimal test for this now

arjxn-py:
> We can add a minimal test for this now

Sure, on it.

PR #2537, opened 2022-12-09
Migrate to tox 4.0

Saransh-cpp:
# Description

- A list of breaking changes - https://tox--2601.org.readthedocs.build/en/2601/faq.html#breaking-changes-in-tox-4.
- The relevant issue about env variables not being visible - https://github.com/tox-dev/tox/issues/2659 (macOS tests are probably failing because of this).

Fixes # (issue)

## Type of change

Please add a line in the relevant section of [CHANGELOG.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CHANGELOG.md) to document the change (include PR #) - note reverse order of PR #s. If necessary, also add to the list of breaking changes.

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (back-end change that speeds up the code)
- [ ] Bug fix (non-breaking change which fixes an issue)

# Key checklist:

- [ ] No style issues: `$ flake8`
- [ ] All tests pass: `$ python run-tests.py --unit`
- [ ] The documentation builds: `$ cd docs` and then `$ make clean; make html`

You can run all three at once, using `$ python run-tests.py --quick`.

## Further checks:

- [ ] Code is commented, particularly in hard-to-understand areas
- [ ] Tests added that prove fix is effective or that feature works

PR #3005, opened 2023-05-31
Migrate from `tox 3.28` to `nox`

arjxn-py:
# Description

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:


agriyakhetarpal:
There was also a discussion on switching from `Tox` to `Nox` in https://github.com/pybamm-team/PyBaMM/pull/2537#issuecomment-1435794609, something that I mentioned in my proposal. Though we can also take a look at https://github.com/wntrblm/nox/pull/687 after it gets merged to do that

agriyakhetarpal:
For the test suite, I was also looking at this library called [`test-stages`](https://gitlab.com/ppentchev/test-stages) – it allows us to group particular sections of `Tox` and `Nox` tests using [`parse-stages`](https://devel.ringlet.net/devel/parse-stages) and run those groups in parallel, and seems to be maintained too.

I was trying to get this set up locally some time ago but I could not get anything concrete so as to push changes. Maybe you can look into this, we might be able to bring CI time down on Linux as well with this by running `tox -e docs` in conjunction with `tox -e unit` or, say, the trio of `examples` + `unit` + `integration` together since the example tests take the longest amount of time

arjxn-py:
> For the test suite, I was also looking at this library called [`test-stages`](https://gitlab.com/ppentchev/test-stages) – it allows us to group particular sections of `Tox` and `Nox` tests using [`parse-stages`](https://devel.ringlet.net/devel/parse-stages) and run those groups in parallel, and seems to be maintained too.


Apologies for the delay in response @agriyakhetarpal, will surely give it a try. Thankfully, for now almost everything has been migrated to `nox` with trivial differences in the completion time. I am also attaching [logs](https://github.com/arjxn-py/PyBaMM/actions/runs/5167708708) for further reference.

agriyakhetarpal:
Though I cannot review or approve the PR, looks good to me! You also probably need to rewrite the documentation to change all instances of `tox` to `nox` and update the instructions to account for differences in how to run the tests and for installation from source

Saransh-cpp (commented):
Taking a quick look, this looks amazing, and the CI is passing too. I'll review this in a few days!

noxfile.py
```diff
@@ -0,0 +1,76 @@
...
+@nox.session(name="doctests", reuse_venv=True)
+def run_doctests(session):
+    session.run("pip", "install", "-e", ".[docs]")
+    session.run("python", "run-tests.py", "--doctest")
+
+
+@nox.session(name="doctests", reuse_venv=True)
+def run_doctests(session):
+    session.run("pip", "install", "-e", ".[docs]")
+    session.run("python", "run-tests.py", "--doctest")
```
Saransh-cpp, line 51: I think this was repeated twice.

noxfile.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+@nox.session(name="unit", reuse_venv=True)
+def run_doctests(session):
```
Saransh-cpp, line 55:
```suggestion
def run_unit(session):
```

.github/workflows/run_periodic_tests.yml
```diff
@@ -87,40 +87,41 @@ jobs:
...
+  
       - name: Install SuiteSparse and Sundials
         if: matrix.os == 'ubuntu-latest'
-        run: tox -e pybamm-requires
+        run: nox -s pybamm-requires
 
       - name: Run unit tests for GNU/Linux with Python 3.8, 3.10, and 3.11
         if: matrix.os == 'ubuntu-latest' && matrix.python-version != 3.9
-        run: python -m tox -e unit
+        run: python -m nox -s unit
```
Saransh-cpp, line 99:
```suggestion
        run: nox -s unit
```

.github/workflows/run_periodic_tests.yml
```diff
@@ -87,40 +87,41 @@ jobs:
...
 
       - name: Run unit tests for GNU/Linux with Python 3.9 and generate coverage report
         if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.9
-        run: python -m tox -e coverage
+        run: python -m nox -s coverage
```
Saransh-cpp, line 103:
```suggestion
        run: nox -s coverage
```

.github/workflows/run_periodic_tests.yml
```diff
@@ -87,40 +87,41 @@ jobs:
...
 
       - name: Run integration tests for GNU/Linux
         if: matrix.os == 'ubuntu-latest'
-        run: python -m tox -e integration
+        run: python -m nox -s integration
 
       - name: Run unit tests for Windows and MacOS
         if: matrix.os != 'ubuntu-latest'
-        run: python -m tox -e mac-windows-unit
+        run: python -m nox -s mac-windows-unit
```
Saransh-cpp, line 115:
```suggestion
        run: nox -s mac-windows-unit
```

.github/workflows/run_periodic_tests.yml
```diff
@@ -87,40 +87,41 @@ jobs:
...
 
       - name: Run integration tests for Windows and MacOS
         if: matrix.os != 'ubuntu-latest'
-        run: python -m tox -e mac-windows-integration
+        run: python -m nox -s mac-windows-integration
```
Saransh-cpp, line 119:
```suggestion
        run: nox -s mac-windows-integration
```

.github/workflows/run_periodic_tests.yml
```diff
@@ -87,40 +87,41 @@ jobs:
       - name: Install standard python dependencies
         run: |
           python -m pip install --upgrade pip wheel setuptools
-          python -m pip install "tox<4"
-
+          python -m pip install "tox"
+          python -m pip install "nox"
```
Saransh-cpp, line 91:
```suggestion
```

.github/workflows/run_periodic_tests.yml
```diff
@@ -87,40 +87,41 @@ jobs:
       - name: Install standard python dependencies
         run: |
           python -m pip install --upgrade pip wheel setuptools
```
Saransh-cpp, line 89:
```suggestion
          python -m pip install --upgrade pip wheel setuptools nox
```

tox.ini
Saransh-cpp, line 1: We can remove this file now?

Saransh-cpp (changes requested):
Amazing work, thanks, @arjxn-py! Some quick comments below. Could you also completely remove the `tox.ini` file and check if the workflows still pass? I'll have another look on this in a while and try running the testing suite locally with `nox`.

arjxn-py:
@Saransh-cpp here are the scheduled [test logs](https://github.com/arjxn-py/PyBaMM/actions/runs/5207483350) after the latest changes.
`Coverage` session is failing while executing `coverage combine` with the same response - `No data to combine` locally as well.

agriyakhetarpal (commented):
`coverage.py` reads these parameters from configuration files; such as `setup.cfg`, or `tox.ini` through the `[coverage:run]` section header ([coverage docs](https://coverage.readthedocs.io/en/stable/config.html)). But it doesn't read from `noxfile.py` yet (I guess). So we need to create a `.coveragerc` configuration file in the root folder with these options:

```
[run]
source = pybamm
concurrency = multiprocessing
```

or one may pass it as a command-line argument via `--concurrency=multiprocessing`. Apparently, we can't pass `--source` as an argument, so a new configuration file should be the way to go. These parameters were not found when `tox.ini` was deleted

noxfile.py
```diff
@@ -0,0 +1,70 @@
...
+    homedir = os.getenv("HOME")
+    session.env["SUNDIALS_INST"] = session.env.get("SUNDIALS_INST", f"{homedir}/.local")
+    session.env[
+        "LD_LIBRARY_PATH"
+    ] = f"{homedir}/.local/lib:{session.env.get('LD_LIBRARY_PATH')}"
+    session.install("coverage", "scikits.odes")
+    session.run("pip", "install", "-e", ".[dev]")
+    if sys.platform != "win32" or sys.platform != "darwin":
+        session.run("pybamm_install_jax")
+    session.run("coverage", "run", "run-tests.py", "--nosub")
```
agriyakhetarpal, line 31:
```suggestion
    session.run("coverage", "run", "--rcfile=.coveragerc", "run-tests.py", "--nosub")
```

Just to ensure that it reads from the file, not needed though

agriyakhetarpal (changes requested):

noxfile.py
agriyakhetarpal: Also, don't mind me barging in with a review lol—I suggested these changes because I received a review request notification on mobile, but I don't see it on web

arjxn-py:
Thanks @agriyakhetarpal, really so nice of you.

noxfile.py
arjxn-py: It's totally cool, you can always suggest changes if you feel like, i'm glad you took a look into it.

noxfile.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+@nox.session(name="coverage")
+def run_coverage(session):
+    homedir = os.getenv("HOME")
+    session.env["SUNDIALS_INST"] = session.env.get("SUNDIALS_INST", f"{homedir}/.local")
+    session.env[
+        "LD_LIBRARY_PATH"
+    ] = f"{homedir}/.local/lib:{session.env.get('LD_LIBRARY_PATH')}"
+    session.install("coverage", "scikits.odes")
```
Saransh-cpp, line 27: This should ideally go in the if statement below. We don't really want to install `scikits.odes` on every OS. Installing `coverage` here should be fine.

noxfile.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+@nox.session(name="mac-windows-unit", reuse_venv=True)
+def run_mac_windows_unit(session):
+    session.run("pip", "install", "-e", ".[dev]")
+    session.run("python", "run-tests.py", "--unit")
+
+
+@nox.session(name="mac-windows-integration", reuse_venv=True)
+def run_mac_windows_integration(session):
```
Saransh-cpp, line 62: `run_integration` and `run_mac_windows_integration` are same. Can we just use a single function for every OS? Given that the dependencies are being installed using a different function.

noxfile.py
```diff
@@ -0,0 +1,70 @@
...
+
+@nox.session(name="unit", reuse_venv=True)
+def run_unit(session):
+    session.run("pip", "install", "-e", ".[dev]")
+    session.run("pybamm_install_jax")
+    session.run("python", "run-tests.py", "--unit")
+
+
+@nox.session(name="mac-windows-unit", reuse_venv=True)
+def run_mac_windows_unit(session):
```
Saransh-cpp, line 56: See the comment on `run_mac_windows_unit`

noxfile.py
```diff
@@ -0,0 +1,70 @@
...
+def run_coverage(session):
+    homedir = os.getenv("HOME")
+    session.env["SUNDIALS_INST"] = session.env.get("SUNDIALS_INST", f"{homedir}/.local")
+    session.env[
+        "LD_LIBRARY_PATH"
+    ] = f"{homedir}/.local/lib:{session.env.get('LD_LIBRARY_PATH')}"
+    session.install("coverage", "scikits.odes")
+    session.run("pip", "install", "-e", ".[dev]")
+    if sys.platform != "win32" or sys.platform != "darwin":
+        session.run("pybamm_install_jax")
```
Saransh-cpp, line 30: Can you refactor this step out as a new function? Maybe something like this -

```py
@nox.session(name="install-jax")
    def run_install_jax(session):
        if sys.platform != "win32":
             session.run("pybamm_install_jax")
```

and then execute it only once in the workflows?

noxfile.py
```diff
@@ -0,0 +1,70 @@
...
+    session.run("coverage", "run", "--rcfile=.coveragerc", "run-tests.py", "--nosub")
+    session.run("coverage", "combine")
+    session.run("coverage", "xml")
+
+
+@nox.session(name="integration", reuse_venv=True)
+def run_integration(session):
+    session.run("pip", "install", "-e", ".[dev]")
```
Saransh-cpp, line 38: It would also be good to refactor the `[dev]` and `[doc]` installations to avoid installing stuff multiple times.

.github/workflows/test_on_push.yml
```diff
@@ -97,35 +97,36 @@ jobs:
...
+        run: python -m nox -s coverage
 
       - name: Upload coverage report
         if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.11
         uses: codecov/codecov-action@v2.1.0
 
       - name: Run integration tests for GNU/Linux with Python 3.11
         if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.11
-        run: python -m tox -e integration
+        run: python -m nox -s integration
```
Saransh-cpp, line 120:
```suggestion
        run: nox -s integration
```

.github/workflows/test_on_push.yml
```diff
@@ -97,35 +97,36 @@ jobs:
...
+      - name: Run unit tests for GNU/Linux with Python 3.8, 3.10, and 3.11
+        if: matrix.os == 'ubuntu-latest' && matrix.python-version != 3.9
+        run: python -m nox -s unit
 
-      - name: Run unit tests for GNU/Linux with Python 3.11 and generate coverage report
-        if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.11
-        run: python -m tox -e coverage
+      - name: Run unit tests for GNU/Linux with Python 3.9 and generate coverage report
+        if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.9
+        run: python -m nox -s coverage
```
Saransh-cpp, line 112:
```suggestion
        run: nox -s coverage
```

.github/workflows/test_on_push.yml
```diff
@@ -97,35 +97,36 @@ jobs:
...
         if: matrix.os == 'ubuntu-latest'
-        run: tox -e pybamm-requires
+        run: nox -s pybamm-requires
 
-      - name: Run unit tests for GNU/Linux with Python 3.8, 3.9, and 3.10
-        if: matrix.os == 'ubuntu-latest' && matrix.python-version != 3.11
-        run: python -m tox -e unit
+      - name: Run unit tests for GNU/Linux with Python 3.8, 3.10, and 3.11
+        if: matrix.os == 'ubuntu-latest' && matrix.python-version != 3.9
+        run: python -m nox -s unit
```
Saransh-cpp, line 108:
```suggestion
        run: nox -s unit
```

.github/workflows/test_on_push.yml
```diff
@@ -97,35 +97,36 @@ jobs:
...
 
       - name: Run integration tests for GNU/Linux with Python 3.11
         if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.11
-        run: python -m tox -e integration
+        run: python -m nox -s integration
 
       - name: Run unit tests for Windows and MacOS with all Python versions
         if: matrix.os != 'ubuntu-latest'
-        run: python -m tox -e mac-windows-unit
+        run: python -m nox -s mac-windows-unit
```
Saransh-cpp, line 124:
```suggestion
        run: nox -s mac-windows-unit
```

.coveragerc
```diff
@@ -0,0 +1,3 @@
+[run]
+source = pybamm
+concurrency = multiprocessing
```
Saransh-cpp, line 3:
```suggestion
concurrency = multiprocessing

```

Saransh-cpp (changes requested):
Thanks, @arjxn-py and @agriyakhetarpal! Looks much better! Some comments below. Also, @arjxn-py, you might have to merge the develop branch to get all the changes Agriya has recently made into your PR.

noxfile.py
arjxn-py: Now `scikits.odes` would only be installed in GNU/Linux.

agriyakhetarpal: I don't think this should be the case, since `scikits.odes` does get installed on macOS when one runs `tox -e coverage`. The difference is that on Linux, we should use the `pybamm_install_odes` command, and on macOS, we should use `pip install scikits.odes`

Saransh-cpp: `tox -e coverage` never runs on MacOS, so we currently only test the solver on Linux. `pybamm_install_odes` is not really used anywhere in the CI, the script is only meant for users (see #2962).

`pybamm_install_odes` calls `pip install scikits.odes` internally, and acts almost identical to `tox -e coverage`, the only difference being that it does not build `SUNDIALS` with `KLU`. We can use `pip install scikits.odes` on any OS, given that we have `SUNDIALS` installed correctly.

agriyakhetarpal: Thanks, I get it now – we shouldn't use `pybamm_install_odes` in the CI that way. However, though `tox -e coverage` does not run on macOS in the CI (it runs only for Linux), it is intended to be run locally too on all three platforms. So I suggest we install `scikits.odes` on macOS too with `nox -s coverage`, just to replicate the functionality of the local macOS `tox -e coverage`. 

Otherwise, we should have different `coverage` environments for CI and for running locally (which does not seem to be a good idea, obviously)

.github/workflows/test_on_push.yml
```diff
@@ -97,35 +97,36 @@ jobs:
...
+      - name: Run unit tests for GNU/Linux with Python 3.8, 3.10, and 3.11
+        if: matrix.os == 'ubuntu-latest' && matrix.python-version != 3.9
+        run: nox -s unit
 
-      - name: Run unit tests for GNU/Linux with Python 3.11 and generate coverage report
-        if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.11
-        run: python -m tox -e coverage
+      - name: Run unit tests for GNU/Linux with Python 3.9 and generate coverage report
+        if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.9
+        run: nox -s coverage
```
agriyakhetarpal, line 112:
```suggestion
      - name: Run unit tests for GNU/Linux with Python 3.11 and generate coverage report
        if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.11
        run: nox -s coverage
```

.github/workflows/test_on_push.yml
```diff
@@ -97,35 +97,36 @@ jobs:
...
         if: matrix.os == 'ubuntu-latest'
-        run: tox -e pybamm-requires
+        run: nox -s pybamm-requires
 
-      - name: Run unit tests for GNU/Linux with Python 3.8, 3.9, and 3.10
-        if: matrix.os == 'ubuntu-latest' && matrix.python-version != 3.11
-        run: python -m tox -e unit
+      - name: Run unit tests for GNU/Linux with Python 3.8, 3.10, and 3.11
+        if: matrix.os == 'ubuntu-latest' && matrix.python-version != 3.9
+        run: nox -s unit
```
agriyakhetarpal, line 108:
```suggestion
      - name: Run unit tests for GNU/Linux with Python 3.8, 3.9, and 3.10
        if: matrix.os == 'ubuntu-latest' && matrix.python-version != 3.11
        run: nox -s unit
```

Maybe this was reversed when merging changes from #3016 into your branch

.github/workflows/run_periodic_tests.yml
```diff
@@ -86,41 +86,35 @@ jobs:
...
         if: matrix.os == 'ubuntu-latest'
-        run: tox -e pybamm-requires
+        run: nox -s pybamm-requires
 
-      - name: Run unit tests for GNU/Linux with Python 3.8, 3.9, and 3.10
-        if: matrix.os == 'ubuntu-latest' && matrix.python-version != 3.11
-        run: python -m tox -e unit
+      - name: Run unit tests for GNU/Linux with Python 3.8, 3.10, and 3.11
+        if: matrix.os == 'ubuntu-latest' && matrix.python-version != 3.9
+        run: nox -s unit
```
agriyakhetarpal, line 97:
```suggestion
      - name: Run unit tests for GNU/Linux with Python 3.8, 3.9, and 3.10
        if: matrix.os == 'ubuntu-latest' && matrix.python-version != 3.11
        run: nox -s unit
```

.github/workflows/run_periodic_tests.yml
```diff
@@ -86,41 +86,35 @@ jobs:
...
 
-      - name: Run unit tests for GNU/Linux with Python 3.11 and generate coverage report
-        if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.11
-        run: python -m tox -e coverage
+      - name: Run unit tests for GNU/Linux with Python 3.9 and generate coverage report
+        if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.9
+        run: nox -s coverage
```
agriyakhetarpal, line 101:
```suggestion
      - name: Run unit tests for GNU/Linux with Python 3.11 and generate coverage report
        if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.11
        run: nox -s coverage
```

noxfile.py
arjxn-py: Now we've single function for `integration`

noxfile.py
arjxn-py: Yes, actually `[pybamm_requires]` is enough for most of the sessions although `examples` do require `[dev]` installation.

noxfile.py
arjxn-py: Now we have single `unit` test for all OS in which `jax` is only being installed on linux.

noxfile.py
arjxn-py: I can do this but I would still have to install `jax` explicitely in requiring sessions due to `isolated_build`.
That being said we are executing `pybamm_install_jax` two times (only on linux) which barely takes 10-15 seconds on CI.

agriyakhetarpal:
Also, we still need some more `nox` sessions in `noxfile.py` for usage in developer workflows locally:

1. `dev` / `nox -s dev` (to replicate `tox -e dev`, which installs PyBaMM and all of its extra dependencies locally for development purposes)
2. `docs` / `nox -s docs` (to replicate `tox -e docs` which serves the documentation locally with `sphinx-autobuild`)
3. `tests` / nox -s tests` (to replicate `tox -e tests`, which wraps unit tests and integration tests with one command)

arjxn-py:
> Also, we still need some more `nox` sessions in `noxfile.py` for usage in developer workflows locally.

Yes @agriyakhetarpal I am still making those sessions.

noxfile.py
```diff
@@ -0,0 +1,60 @@
...
+
+@nox.session(name="coverage")
+def run_coverage(session):
+    homedir = os.getenv("HOME")
+    session.env["SUNDIALS_INST"] = session.env.get("SUNDIALS_INST", f"{homedir}/.local")
+    session.env[
+        "LD_LIBRARY_PATH"
+    ] = f"{homedir}/.local/lib:{session.env.get('LD_LIBRARY_PATH')}"
+    session.install("coverage")
+    session.run("pip", "install", "-e", ".[pybamm_requires]")
```
agriyakhetarpal, line 28: `setup.py` does not define `[pybamm_requires]` as an extra (the only extras are `dev` and `docs`)—it also shows a warning in the logs. This should be just `pip install -e .` (and similarly in other places too)

arjxn-py: Sure, will make the changes.

noxfile.py
```diff
@@ -0,0 +1,105 @@
...
+        )
+
+
+@nox.session(name="coverage")
+def run_coverage(session):
+    homedir = os.getenv("HOME")
+    session.env["SUNDIALS_INST"] = session.env.get("SUNDIALS_INST", f"{homedir}/.local")
+    session.env[
+        "LD_LIBRARY_PATH"
+    ] = f"{homedir}/.local/lib:{session.env.get('LD_LIBRARY_PATH')}"
```
Saransh-cpp, line 26: Hmm, why aren't we setting these variables in `run_pybamm_requires`?

noxfile.py
```diff
@@ -0,0 +1,105 @@
...
+    envbindir = session.bin
+    session.run("pip", "install", "-e", ".[dev]")
+    session.install("cmake")
+    session.run(
+        "echo",
+        "export",
+        f"LD_LIBRARY_PATH={LD_LIBRARY_PATH}",
+        ">>",
+        f"{envbindir}/activate",
+    )
```
Saransh-cpp, line 76: Do we require setting these variables again? What I meant with dev -

We're installing the Pythonic dependencies in every session, which will slow down the CI. We should refactor that installation process in a single function and call that function in the CI before any tests (`nox -s dev`).

.coveragerc
Saransh-cpp, line 3: We can use the `--concurrency` flag while executing coverage. Something like `--concurrency=multiprocessing`. We can then get rid of this file.

Saransh-cpp (changes requested):
Nice work, @arjxn-py! I think you've added back the `tox.ini` file? That should be removed. Also, it might be a good idea to squash your commits - the count is already above 60 right now haha

.github/workflows/run_periodic_tests.yml
Saransh-cpp: Note: this will change `git blame`. It would be better to fix the merge instead of manually committing the upstream changes.

.github/workflows/run_periodic_tests.yml
Saransh-cpp: Ditto

.github/workflows/test_on_push.yml
Saransh-cpp: Ditto

.github/workflows/test_on_push.yml
Saransh-cpp: Ditto

.coveragerc
agriyakhetarpal: Actually, no we can't—we can specify `--concurrency=multiprocessing` as a command-line argument but not `--source = pybamm`. The `coverage` docs say it needs this in a config file which can be either `tox.ini` or `setup.cfg` or `.coveragerc` (as discussed above). So it would be better to have both of these options in one place/file, in that sense

Saransh-cpp: Ah, we can do `--source` but we can't do `--source` and `--concurrency` together.

agriyakhetarpal: Yes. I suppose an extra file should not be a problem until `coverage` adds support for `noxfile.py`

noxfile.py
arjxn-py: Since these were not also declared in `tox : [pybamm-requires]` I'm setting up these variables wherever required.

noxfile.py
arjxn-py: @Saransh-cpp I had tried doing that more than a couple of times before & just now too but since `isolated_build` is turned on every session is creating seperate `.nox/env/lib/python/site-packages` which means even if I ran a session to install all the deps say `dev` I'd still have to install the deps as the session wouldn't look for libraries in `.nox/dev/lib/python/site-packages` but in `.nox/session_name/lib/python/site-packages`.

arjxn-py: About setting those variables again, I can globally declare the variables above but global declaration is not recommended.

noxfile.py
Saransh-cpp: I think it would be better to just add them in `run_pybamm_requires`.

arjxn-py: I've added them in `run_pybamm_requires` as well. 
You don't want me to set them globally, do you 👀.

noxfile.py
agriyakhetarpal: Something left to be addressed in this PR is to update the CHANGELOG, and the documentation: to account for the change to `nox` wherever required and breaking changes in the installation and developer workflow processes (if any)

arjxn-py: > Something left to be addressed in this PR is to update the CHANGELOG, and the documentation: to account for the change to `nox` wherever required and breaking changes in the installation and developer workflow processes (if any)

On it.

Saransh-cpp: Almost forgot the docs! Thanks for the reminder!

.github/workflows/run_benchmarks_over_history.yml
```diff
@@ -27,8 +27,8 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: 3.8
-      - name: Install tox and asv
-        run: pip install -U pip "tox<4" asv
+      - name: Install nox and asv
+        run: pip install -U pip "nox" asv
```
agriyakhetarpal, line 31:
```suggestion
      - name: Install nox and asv
        run: pip install -U pip nox asv
```

We can remove the double quotes since we don't need to install a specific version of `nox`

docs/source/user_guide/installation/install-from-source.rst
```diff
@@ -47,12 +47,12 @@ You can install the above with
...
 		brew install python openblas gcc gfortran libomp
 
-Finally, we recommend using `Tox <https://tox.readthedocs.io/en/latest/>`_.
+Finally, we recommend using `Nox <https://nox.thea.codes/en/stable/>`_.
 You can install it with
 
 .. code:: bash
 
-	  python3.X -m pip install --user "tox<4"
+	  python3.X -m pip install --user "nox"
```
agriyakhetarpal, line 55:
```suggestion
	  python3.X -m pip install --user nox
```

noxfile.py
```diff
@@ -0,0 +1,110 @@
...
+        "pydata-sphinx-theme",
+        "sphinx-autobuild",
+        "sphinx_design",
+        "sphinx-copybutton",
+        "myst-parser",
+        "sphinx-inline-tabs",
+    )
+    session.run(
+        "sphinx-autobuild", "--open-browser", "-qT", ".", f"{envbindir}/../tmp/html"
+    )
```
agriyakhetarpal, line 110: This failed for me locally because changing to the `docs/` directory was trying to install `pybamm` there, rather than the root directory – it could not find `setup.py` inside `docs/`. Also, `sphinx-autobuild` wasn't able to find `conf.py` for building the documentation (it should run from within the `docs/` folder).

We probably don't need to make a `docs/` directory since it's not git-ignored and should never get deleted

```suggestion
@nox.session(name="docs", reuse_venv=True)
def build_docs(session):
    envbindir = session.bin
    session.run("pip", "install", "-e", ".")
    session.install(
        "sphinx>=1.5",
        "pydata-sphinx-theme",
        "sphinx-autobuild",
        "sphinx_design",
        "sphinx-copybutton",
        "myst-parser",
        "sphinx-inline-tabs",
    )
    session.chdir("docs/")
    session.run(
        "sphinx-autobuild", "--open-browser", "-qT", ".", f"{envbindir}/../tmp/html"
    )
```

agriyakhetarpal (changes requested):
Looks good to me overall! I tested this locally and all but the `docs` environment worked for me as expected, so I have some minor comments (waiting on @Saransh-cpp to test it too)

noxfile.py
```diff
@@ -0,0 +1,108 @@
...
+@nox.session(name="doctests", reuse_venv=True)
+def run_doctests(session):
+    session.run("pip", "install", "-e", ".[docs]")
+    session.run("python", "run-tests.py", "--doctest")
+
+
+@nox.session(name="unit", reuse_venv=True)
+def run_unit(session):
+    session.run("pip", "install", "-e", ".")
+    if sys.platform == "linux":
```
Saransh-cpp, line 57: This should also install `scikits.odes`.

noxfile.py
```diff
@@ -0,0 +1,108 @@
...
+        session.install("scikits.odes")
+        session.run("pybamm_install_jax")
+    session.run("coverage", "run", "--rcfile=.coveragerc", "run-tests.py", "--nosub")
+    session.run("coverage", "combine")
+    session.run("coverage", "xml")
+
+
+@nox.session(name="integration", reuse_venv=True)
+def run_integration(session):
+    session.run("pip", "install", "-e", ".")
```
Saransh-cpp, line 44: `scikits.odes` should be installed here.

Saransh-cpp (commented):
We'll have to look into the skipped tests. Maybe try setting env variables and then refactor the code.

arjxn-py:
@Saransh-cpp no tests are getting skipped now in any python version expect 1 in integration that's being skipped everywhere (in develop branch as well).
Time difference is still around 20 minutes [here](https://github.com/arjxn-py/PyBaMM/actions/runs/5271710145)

.github/workflows/test_on_push.yml
```diff
@@ -96,36 +96,36 @@ jobs:
       - name: Install standard Python dependencies
         run: |
           python -m pip install --upgrade pip wheel setuptools
-          python -m pip install "tox<4"
+          python -m pip install "nox"
```
Saransh-cpp, line 99:
```suggestion
          python -m pip install nox
```

CONTRIBUTING.md
```diff
@@ -54,7 +54,7 @@ You now have everything you need to start making changes!
 10. [Test your code!](#testing)
 11. PyBaMM has online documentation at http://pybamm.readthedocs.io/. To make sure any new methods or classes you added show up there, please read the [documentation](#documentation) section.
 12. If you added a major new feature, perhaps it should be showcased in an [example notebook](#example-notebooks).
-13. When you feel your code is finished, or at least warrants serious discussion, run the [pre-commit checks](#pre-commit-checks) and then create a [pull request](https://help.github.com/articles/about-pull-requests/) (PR) on [PyBaMM's GitHub page](https://github.com/pybamm-team/PyBaMM).
+13. When you feel your code is finished, or at least warrants serious discussion, run the [pre-commit checks](#pre-commit-checks) and then create a [pull request](https://help.github.com/articles/about-pull-requests/) (PR) on [PyBaMM&#39;s GitHub page](https://github.com/pybamm-team/PyBaMM).
```
Saransh-cpp, line 57:
```suggestion
13. When you feel your code is finished, or at least warrants serious discussion, run the [pre-commit checks](#pre-commit-checks) and then create a [pull request](https://help.github.com/articles/about-pull-requests/) (PR) on [PyBaMM's GitHub page](https://github.com/pybamm-team/PyBaMM).
```

noxfile.py
```diff
@@ -0,0 +1,121 @@
...
+        "echo",
+        "export",
+        f"LD_LIBRARY_PATH={LD_LIBRARY_PATH}",
+        ">>",
+        f"{envbindir}/activate",
+    )
+
+
+@nox.session(name="tests", reuse_venv=True)
+def run_tests(session):
```
Saransh-cpp, line 98: We'll need to install `scikits.odes` here and set the env variables.

CONTRIBUTING.md
```diff
@@ -384,7 +372,7 @@ Code coverage (how much of our code is actually seen by the (linux) unit tests)
 Configuration files:
 
 ```
-tox.ini
+noxfile.py
```
Saransh-cpp, line 375:
```suggestion
.coveragerc
```

Saransh-cpp (changes requested):
Thanks for the great work, @arjxn-py! Some more minor changes below 🙂

noxfile.py
arjxn-py: Thanks for the review @Saransh-cpp , I have made the suggested changes.

Saransh-cpp:
Thanks, @arjxn-py! You'll need to merge in develop so that the tests pass. I'd recommend squashing your commits locally and force-pushing to reduce the number of commits in this PR (71 looks like a lot 😅).

arjxn-py:
@Saransh-cpp I've squashed all the commits & also merged with develop

noxfile.py
```diff
@@ -0,0 +1,127 @@
...
+    session.run("pip", "install", "-e", ".")
+    session.install(
+        "sphinx>=1.5",
+        "pydata-sphinx-theme",
+        "sphinx-autobuild",
+        "sphinx_design",
+        "sphinx-copybutton",
+        "myst-parser",
+        "sphinx-inline-tabs",
+    )
```
Saransh-cpp, line 123: I think this should work?
```suggestion
    session.run("pip", "install", "-e", ".[docs]")
```

noxfile.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+@nox.session(name="integration", reuse_venv=True)
+def run_integration(session):
+    homedir = os.getenv("HOME")
+    session.env["SUNDIALS_INST"] = session.env.get("SUNDIALS_INST", f"{homedir}/.local")
+    session.env[
+        "LD_LIBRARY_PATH"
+    ] = f"{homedir}/.local/lib:{session.env.get('LD_LIBRARY_PATH')}"
+    session.run("pip", "install", "-e", ".")
```
Saransh-cpp, line 49:
```suggestion
    session.run("pip", "install", "-e", ".[dev]")
```

noxfile.py
```diff
@@ -0,0 +1,127 @@
...
+        session.run(
+            "git",
+            "clone",
+            "https://github.com/pybind/pybind11.git",
+            "pybind11/",
+            external=True,
+        )
+
+
+@nox.session(name="coverage")
```
Saransh-cpp, line 25: Should we also reuse this environment? Should make things a bit faster.
```suggestion
@nox.session(name="coverage", reuse_venv=True)
```

noxfile.py
```diff
@@ -0,0 +1,127 @@
+import nox
+import os
+import sys
+
+
+@nox.session(name="pybamm-requires")
```
Saransh-cpp, line 6: Ditto!
```suggestion
@nox.session(name="pybamm-requires", reuse_venv=True)
```

Saransh-cpp (commented):
Amazing work, @arjxn-py! Last set of comments 🙂 

Thanks for working on this patiently!

noxfile.py
Saransh-cpp: Also, looking at this again, it would be better to just replace all the `session.run("pip", "install", "-e", "something")` occurrences with `session.install("-e", "something")`.

noxfile.py
agriyakhetarpal: Note: when I had re-run `pybamm-requires` locally, it failed because `pybind11` was already a directory in the source tree. We might need to wrap something like

```python
if os.path.exists('pybind11_path):
    os.rmdir(pybind11_path)
```

or with 'pathlib.Path.rmdir' or 'shutil.rmtree' as necessary and catch related errors.

agriyakhetarpal: Note: when I had re-run `pybamm-requires` locally, it failed because `pybind11` was already a directory in the source tree. We might need to wrap something like

```python
if os.path.exists(pybind11_path):
    os.rmdir(pybind11_path)
```

or with 'pathlib.Path.rmdir' or 'shutil.rmtree' as necessary and catch related errors.

agriyakhetarpal: Note: when I had re-run `pybamm-requires` locally, it failed because `pybind11` was already a directory in the source tree. We might need to wrap something like

```python
if os.path.exists(pybind11_path):
    os.rmdir(pybind11_path)
```

or with `pathlib.Path.rmdir` or `shutil.rmtree` as necessary and catch related errors.

noxfile.py
arjxn-py: It would make no effect on time but yeah no harm to use it.

noxfile.py
arjxn-py: Sure, doing it this way.

noxfile.py
arjxn-py: Thanks @agriyakhetarpal, looking into it.

Saransh-cpp (approved):
Thanks for working on this patiently, @arjxn-py! Looks awesome! Given that the CI time has not increased, we can safely say that installing dependencies in different environments, again and again, is okay.

I'll request reviews from a few more people just to be sure that I'm not missing anything 🙂

Saransh-cpp:
Once this PR is merged, we can improve the `noxfile` a bit more in the next PRs, maybe by -

- [ ] refactoring the env variables
- [ ] refactoring the installation step in a separate Python function and calling it in every session
- [ ] adding default sessions
- [ ] adding more informative commands in docs, like `nox -l`, `nox -p`
- [ ] ...

agriyakhetarpal:
I have plans to open a PR to cache these environments, you may add that too to the list

agriyakhetarpal (approved):
I ran this locally on macOS; all commands worked as expected, except for a few:

1. It would be great if there were a way to make `nox -s dev` run like `nox -s dev --verbose` by default, otherwise, just the message `nox > Session dev was successful.` is displayed and no package installation logs are shown in the command line

2. Running the unit tests with `nox -s unit` skipped 50 tests: 27 because `scikits.odes` was not installed and 23 because `jax` was not installed. Should we install them on macOS too? (looks like they were [skipped in CI](https://github.com/pybamm-team/PyBaMM/actions/runs/5302287481/jobs/9597114458?pr=3005#step:14:1609) as well)

3. The `nox -e docs` command was not successful because `sphinx-autobuild` was not installed (which is weird since it worked fine when I tested it earlier). We should add it in the `docs` extras in `setup.py`

4. If `nox -s pybamm-requires` is executed _again_, it gives another error: `OSError: [Errno 66] Directory not empty: './pybind11'`. We can do this, rather:

```python
if not os.path.exists("./pybind11"):
    session.run("git", "clone", "https://github.com/pybind/pybind11.git", "pybind11/", external=True,)
```

6. L13 in `noxfile.py`
```python
if sys.platform != "win32" or sys.platform != "darwin":
```
Here, `or sys.platform != "darwin"` is redundant—because we want `nox -s pybamm-requires` to not run on just Windows, and to run on both other platforms. We can just use either `if sys.platform != "win32"` or use `if sys.platform == "darwin" or sys.platform == "linux"` (both are the same). Another reason why this is currently redundant is that it has a logical error; an `or` Boolean operator rather than an `and` operator – if in any case we had to exclude macOS here, we would have had to use `and` (De Morgan's laws 😄). For Windows users, we could throw an error too instead, notifying them that the command is available on Linux and macOS only. This way `nox` will avert creating a futile environment altogether

arjxn-py:
Thanks @Saransh-cpp @agriyakhetarpal for the detailed review, I'll look into the suggested things.

arjxn-py:
> Running the unit tests with nox -s unit skipped 50 tests: 27 because scikits.odes was not installed and 23 because jax was not installed. Should we install them on macOS too? (looks like they were [skipped in CI](https://github.com/pybamm-team/PyBaMM/actions/runs/5302287481/jobs/9597114458?pr=3005#step:14:1609) as well)

As far as I remember we were doing the same with `tox` as well but if you people would like `scikits.odes` & `jax` on macos too, I can look into it.

agriyakhetarpal:
I see, on a fresh PyBaMM installation on macOS with `tox` it skips 1 test for me though (it does install both `scikits.odes` and `jax` using `pybamm_install_jax`)

arjxn-py:
> I see, on a fresh PyBaMM installation on macOS with `tox` it skips 1 test for me though (it does install both `scikits.odes` and `jax` using `pybamm_install_jax`)

I'm not sure how it is installing `scikits.odes` with  `pybamm_install_jax` can you once try testing again after removing `unit` folder from `.tox/` directory.

agriyakhetarpal:
I double-checked to be sure, yes... and I always delete the `.tox/` directory. Will test once again and edit this comment.

arjxn-py:
My bad, sorry both I fixed it for nightly tests (i.e. Ubuntu) please give me a couple of hours to look into this too.

arjxn-py:
Now that I am installing `scikits.odes` on macos, the [CI](https://github.com/pybamm-team/PyBaMM/actions/runs/5313576345/jobs/9619940050?pr=3005) is failing.
I still am wondering that we are doing the same with `tox.ini` as well (not installing `scikits.odes` on windows & macos) & these tests are getting skipped with `tox` as well on macos [here](https://github.com/pybamm-team/PyBaMM/actions/runs/5311543441/jobs/9614910547)

Saransh-cpp:
I think you misread my comment @arjxn-py. We can skip installing `scikits.odes` and `jax` on macos for now.

noxfile.py
```diff
@@ -0,0 +1,119 @@
...
+
+@nox.session(name="tests", reuse_venv=True)
+def run_tests(session):
+    homedir = os.getenv("HOME")
+    session.env["SUNDIALS_INST"] = session.env.get("SUNDIALS_INST", f"{homedir}/.local")
+    session.env[
+        "LD_LIBRARY_PATH"
+    ] = f"{homedir}/.local/lib:{session.env.get('LD_LIBRARY_PATH')}"
+    session.install("-e", ".")
+    if sys.platform == "linux" or sys.platform == "darwin":
```
Saransh-cpp, line 106:
```suggestion
    if sys.platform == "linux":
```

Saransh-cpp:
> The nox -e docs command was not successful because sphinx-autobuild was not installed (which is weird since it worked fine when I tested it earlier). We should add it in the docs extras in setup.py

> For Windows users, we could throw an [error](https://nox.thea.codes/en/stable/config.html#nox.sessions.Session.error) too instead, notifying them that the command is available on Linux and macOS only. This way nox will avert creating a futile environment altogether

@arjxn-py were these review points resolved?

arjxn-py:
Yes @Saransh-cpp resolved.

arjxn-py:
@tinosulzer I'd be glad If I could get your review on this too 🤞🏻

noxfile.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+@nox.session(name="tests", reuse_venv=True)
+def run_tests(session):
+    homedir = os.getenv("HOME")
+    session.env["SUNDIALS_INST"] = session.env.get("SUNDIALS_INST", f"{homedir}/.local")
+    session.env[
+        "LD_LIBRARY_PATH"
+    ] = f"{homedir}/.local/lib:{session.env.get('LD_LIBRARY_PATH')}"
+    session.install("-e", ".")
```
Saransh-cpp, line 107:
```suggestion
    session.install("-e", ".[dev]")
```

Saransh-cpp (approved):
Looks perfect now 🙂 

Will merge this tonight if there are no blockers

PR #3031, opened 2023-06-12
#2954 turn off benchmark: ORegan2022 DFN solver with the casadi model

### LINKED ISSUE #2954: Track down and exclude non-deterministic benchmarks
Currently, several benchmarks pass or fail randomly on different runs. These benchmarks should be tracked and excluded to ensure the credibility of the remaining benchmarking suite.

Note: random regressions are okay because GH Actions can be a bit noisy sometimes, but solvers failing in a particular benchmark randomly should not be okay.

martinjrobins:
# Description

Fixes #2954

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/benchmarks/time_sims_experiments.py
+++ b/benchmarks/time_sims_experiments.py
@@ -17,7 +17,7 @@ class TimeSimulation:
             "Hold at 4.1 V until 10 mA",
             "Rest for 1 hour",
         ],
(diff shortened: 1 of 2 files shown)


martinjrobins:
benchmarks still failling due to regressions in timings, but I think this PR as is will fix a lot of the failing simulations we've been getting

brosaplanella (approved):
Looks good, thanks!

Saransh-cpp (approved):

PR #3034, opened 2023-06-13
Use `sphinxcontrib-bibtex` for citing references in docs

agriyakhetarpal:
# Description

Removes hard-coded references to relevant papers within docstrings and other places in the documentation using `sphinxcontrib-bibtex` to parse citation keys. This also replaces `CITATIONS.txt` with `CITATIONS.bib` 

Closes #2649 

## Type of change

Please add a line in the relevant section of [CHANGELOG.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CHANGELOG.md) to document the change (include PR #) - note reverse order of PR #s. If necessary, also add to the list of breaking changes.

- [x] New feature (non-breaking change which adds functionality)

# Key checklist:

- [x] No style issues: `$ pre-commit run` (see [CONTRIBUTING.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CONTRIBUTING.md#installing-and-using-pre-commit) for how to set this up to run automatically when committing locally, in just two lines of code)
- [ ] All tests pass: `$ python run-tests.py --all`
- [x] The documentation builds: `$ python run-tests.py --doctest`

You can run unit and doctests together at once, using `$ python run-tests.py --quick`.

## Further checks:

- [x] Code is commented, particularly in hard-to-understand areas
- [x] Tests added that prove fix is effective or that feature works

PR #3043, opened 2023-06-16
Use `nbsphinx` to embed Jupyter notebooks

agriyakhetarpal:
# Description

This PR adds support for `nbsphinx` in order to render Jupyter notebooks in the documentation. All the notebooks will subsequently be moved from the `examples/` folder to `docs/source/examples/` and will be configured as adequate. Furthermore, `nb-gallery` will be added to display thumbnails derived from the outputs of the examples.

Tasks

- [x] Embed all Jupyter notebooks (Tutorial 1 was modified slightly due to issues with printing citations with BibTeX formatting).
- [x] Display thumbnail galleries
- [x] Fix related warnings that come up when building the documentation
- [x] Prologs and epilogs for buttons
- [x] Explore other, advanced configuration options

## Type of change

# Key checklist:

## Further checks:


agriyakhetarpal:
@tinosulzer @rtimms @Saransh-cpp I am currently facing these roadblocks:

1. It looks like there aren't any reliable ways to include files from outside the `docs/` folder (in this case, the example notebooks inside the `examples/notebooks` folder) in the Sphinx documentation (related issue: https://github.com/sphinx-doc/sphinx/issues/701). Some extensions can be used to add external notebooks (`nbsphinx-link`, `nbsphinx-multilink`) but they do not seem to be maintained. Symbolic links cannot be used because they won't work on Windows. 
I have moved the notebooks to `docs/source/examples/notebooks` for this reason. However, this means that `tox -e doctests` (i.e., `python run-tests.py --doctest`) checks the notebooks too now, which it shouldn't – because we have different tests for the notebooks plus `.ipynb` files are not formatted like `.rst` files, so some false warnings are getting flagged. `sphinx-build` converts the warnings to errors, failing the build. I have disabled those temporarily in commit [1198259](https://github.com/pybamm-team/PyBaMM/pull/3043/commits/1198259e218130c7cc8ea9098ed46220d2089d6c).
    -  in the CI a potential workaround is easy: just `rm -rf`ing the `docs/source/examples` folder before the doctests are run should make them run as they are meant to be run originally. However,
    - for running locally, I checked the `sphinx-build` [command-line configuration](https://www.sphinx-doc.org/en/master/man/sphinx-build.html) which says that we can use `exclude_patterns` to not run them on the notebooks, but this is the case only when they are not included in the root toctree. We could move the examples outside of the `docs/` folder before running the doctests with `shutil.move('docs/source/examples', '../../../examples_temp')` or similar and then copy them back afterwards in order to exclude the examples, but this sounds more like a hack rather than an elegant method.

    Is there another way through this that I may be missing?

2. There are some missing thumbnails for the galleries generated by `nbsphinx` and `sphinx-gallery` because some of the notebooks do not contain any images to render. We can edit the cell metadata in this fashion: `{"tags": ["nbsphinx-gallery"]}` to embed the image output from any cell or include a link to an external image (in `.png` format).

3. Having added `sphinx-gallery` in this PR, I also tried adding the example scripts from `examples/scripts/` too. They throw an `ExtensionError` currently – since they don't contain any `.rst`-like documentation inside them to build the notebooks (as we were discussing this week). Should I do this in another, later PR after the documentation gets updated?

P.S. We can now use the `.. nblinkgallery::` directive in the API docs or elsewhere to link notebooks for displaying examples for specific functionality. We can also remove hard-coded citations in notebooks after #3034 gets merged.

.gitignore
```diff
@@ -118,4 +118,11 @@ vcpkg_installed/
...
-results/
\ No newline at end of file
+results/
+
+# expression tree images for notebooks in docs
+!docs/source/examples/notebooks/expression_tree/expression_tree1.png
+!docs/source/examples/notebooks/expression_tree/expression_tree2.png
+!docs/source/examples/notebooks/expression_tree/expression_tree3.png
+!docs/source/examples/notebooks/expression_tree/expression_tree4.png
+!docs/source/examples/notebooks/expression_tree/expression_tree5.png
```
Saransh-cpp, line 128:
```suggestion
!docs/source/examples/notebooks/expression_tree/expression_tree5.png

```

run-tests.py
```diff
@@ -64,20 +65,20 @@ def run_doc_tests():
     used).
     """
     print("Checking if docs can be built.")
-    p = subprocess.Popen(
-        ["sphinx-build", "-b", "doctest", "docs", "docs/build/html", "-W"]
-    )
+    p = subprocess.Popen(["sphinx-build", "-b", "doctest", "docs", "docs/build/html"])
```
Saransh-cpp, line 68: Can't we pass an --ignore equivalent flag here to ignore testing the notebooks? I'm not sure if something like this exists

Saransh-cpp: Ah, just read your comment, will check if there exists a solution.

Saransh-cpp (commented):
Great work, thanks @agriyakhetarpal! The examples look amazing on the website!

I think we can now delete all notebooks from the `examples` directory in PyBaMM. It seems natural for a user to look for examples on PyBaMM's website and not the GitHub source.

Yes, we can move the scripts in another PR, that would be better in my opinion.

For the thumbnails, I think @brosaplanella, @tinosulzer, and @rtimms would be able to help better.

.gitignore
agriyakhetarpal: Did not notice that, thanks!

run-tests.py
agriyakhetarpal: Looks like we have another issue: the doctests now create a `docs/build/` folder containing static files that are not deleted upon completion; re-running the doctests builds a duplicate of the docs with that folder included. This causes an array of several other errors 🙁

Saransh-cpp (approved):
Great work, thanks @agriyakhetarpal! I am guessing that these are the 2 things that we'll have to deal with once this PR is merged -
- [ ] thumbnails
- [ ] check if lychee reports any broken links

agriyakhetarpal:
Should be good to go now, fixing the merge conflicts for the notebooks was cumbersome haha

Issue #2881, opened 2023-04-17
Reduce release frequency


brosaplanella:
After today's developer meeting we agreed to move to less frequent releases. The plan is to move to 3 releases a year (Jan, May, Sep) keeping calendar versioning (i.e. 23.1, 23.5 and 23.9). The release schedule from now should be 23.4 (end Apr), 23.5 (end May) and then 23.9 (end Sep).

Tasks:
- [ ] Fix release workflow
- [x] Announce new release schedule

agriyakhetarpal:
That is a good idea. Glad to take it up

agriyakhetarpal:
That is a good idea. Glad to take it up in multiple PRs or in a single one

brosaplanella:
Copying here the main points from the discussion we had in the infrastructure Slack channel about this topic.
- Automated publication of a release candidate into PyPI one month before the actual release (i.e. end of Dec, Apr and Aug). This should allow packages depending on PyBaMM to test and fix the new release.
- If there are issues building the wheel or issues are flagged downstream, we will make further release candidates manually.
- Once we are happy with the release candidate, we merge to main which would trigger the actual release.

brosaplanella:
I was thinking about the version update of `develop`. I am a bit worried that merging the release branch back into develop will result into CHANGELOG conflicts to fix every time. Should we update the version to the new one (e.g. v23.9) in develop and then branch out to branch `v23.9`, where we set the `rc0` tag manually and eventually make the release from there?

Then we can also modify the old automatic PR to trigger at the end of the previous month for the feature freeze.

agriyakhetarpal:
A note that we have an issue open for using [`towncrier`](https://towncrier.readthedocs.io/en/stable/tutorial.html) (#2885) which could potentially resolve the problem of CHANGELOG conflicts

PR #3045, opened 2023-06-16
Stop automatic version updating

brosaplanella:
Quick fix to stop the scheduled dispatch of the version update workflow (the one that opens the release PR). I have not deleted the workflow as the plan for #2881 is to update it, but this quick fix should prevent the PRs from being open in the meantime.


Saransh-cpp (approved):
Thanks, @brosaplanella!

priyanshuone6 (approved):

PR #3046, opened 2023-06-17
Fix failing notebook

brosaplanella:
The last changes I made to #2616 made the example notebook fail, sorry! I have now fixed it (and added a bit more detail).


Saransh-cpp (approved):
Looks good, thanks @brosaplanella!

PR #3047, opened 2023-06-18
Make release v23.5

brosaplanella:
**WAIT TO MERGE UNTIL WE ARE HAPPY WITH THE RELEASE CANDIDATE**

I wasn't sure whether to have v23.5 or v23.5rc0 in the CHANGELOG and the CITATION. Everywhere else it is `23.5rc0` which should make a pre-release in PyPI.


tinosulzer:
23.5 in the changelog is good, but we need to change the date

brosaplanella:
> 23.5 in the changelog is good, but we need to change the date

Fair point. I have updated the date now, I will try to make the release and publish to pypi, but I am not sure if I have the permissions to trigger the action and actually push to pypi. I will let you know if not and you can trigger it.

CHANGELOG.md
```diff
@@ -1,5 +1,33 @@
 # [Unreleased](https://github.com/pybamm-team/PyBaMM/)
 
+# [v23.5](https://github.com/pybamm-team/PyBaMM/tree/v23.5) - 2023-06-18
```
Saransh-cpp, line 3: Will the date here be updated?

Saransh-cpp (approved):

CHANGELOG.md
brosaplanella: I was thinking of keeping it as it is the date of the feature freeze

PR #3044, opened 2023-06-16
Create Optional Dependencies

arjxn-py:
# Description

The goal of this PR is to make most of the dependencies optional & split them in specific group

Fixes #3035 

## Type of change

Please add a line in the relevant section of [CHANGELOG.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CHANGELOG.md) to document the change (include PR #) - note reverse order of PR #s. If necessary, also add to the list of breaking changes.

- [x] New feature (non-breaking change which adds functionality)
- [x] Optimization (back-end change that speeds up the code)
- [ ] Bug fix (non-breaking change which fixes an issue)

# Key checklist:

- [x] No style issues: `$ pre-commit run` (see [CONTRIBUTING.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CONTRIBUTING.md#installing-and-using-pre-commit) for how to set this up to run automatically when committing locally, in just two lines of code)
- [x] All tests pass: `$ python run-tests.py --all`
- [x] The documentation builds: `$ python run-tests.py --doctest`

You can run unit and doctests together at once, using `$ python run-tests.py --quick`.

## Further checks:

- [x] Code is commented, particularly in hard-to-understand areas
- [x] Tests added that prove fix is effective or that feature works

PR #3062, opened 2023-06-22
Temporarily pin `autograd` < 1.6, fix builds

agriyakhetarpal:
# Description

Adds an upper-bound constraint to `autograd`'s version, in order to fix builds for Windows and ARM M-series macOS, for more details see https://github.com/HIPS/autograd/issues/601

**Update**: fixed after a `1.6.1` patch release
Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:


tinosulzer:
We can probably just remove autograd as a dependency, we don't use it anywhere (only in one test specifically just for autograd)

agriyakhetarpal:
Good idea, since we can fall back to using `derivative` instead of `autograd` when differentiating. Do you want me to try doing this in a further PR?

agriyakhetarpal:
The only place where we use it is here:

tinosulzer:
Yes please. I don't think the fall back to derivative will work, that is specifically for interpolants. But we never need the fallback in practice

agriyakhetarpal:
The only place where we use it is here:

https://github.com/pybamm-team/PyBaMM/blob/bf8581174460005f4d4ecb1a0b56cbf665cd59ce/pybamm/expression_tree/functions.py#L94-L122

agriyakhetarpal:
@tinosulzer I thought of some more discussion before I start doing it concretely. Using "derivative" instead of `"autograd"` did not work, as you said above. Five tests from `test_expression_tree.test_functions` fail locally when I remove `autograd` as an option, and it is the error that comes up when `len(children) > 1`: 
```
ValueError: differentiation using '.derivative()' not implemented for functions with more than one child
```
should I remove those tests? Debugging this could be slightly outside my forte

tinosulzer:
I see, let's keep this as is then, and just make autograd an optional dependency as is already being done in #3044

agriyakhetarpal:
Sure, closing this since the build issue was resolved with autograd `1.6.2`

PR #3079, opened 2023-06-29
Cache `nox` environments and refactor unit tests in CI

### LINKED ISSUE #3095: Refactor test related yaml files
Leveraging the fact that we don't require a separate environment for macos and windows tests now, I think it should be possible to get rid of these steps with minor modifications to the yaml files -

https://github.com/pybamm-team/PyBaMM/blob/a28a0ebb7f8666cd849a334c09bef02501f7ba7b/.github/workflows/test_on_push.yml#L121-L123

https://github.com/pybamm-team/PyBaMM/blob/a28a0ebb7f8666cd849a334c09bef02501f7ba7b/.github/workflows/run_periodic_tests.yml#L110-L112

agriyakhetarpal:
# Description

This PR caches `nox` test environments in order to speed up the CI. Environments are matched with a key that corresponds to relevant files that are subject to change, and the Python version (including the subversion) to avoid mismatch constraints. It also refactors the unit tests to run in a single step for Linux, macOS, and Windows; closes #3095 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/.github/workflows/test_on_push.yml
+++ b/.github/workflows/test_on_push.yml
@@ -111,7 +111,7 @@ jobs:
         if: matrix.os == 'ubuntu-latest' && matrix.python-version != 3.11
         with:
           path: ${{ github.workspace }}/.nox/unit/
-          key: ${{ runner.os }}-nox-unit-${{ matrix.python-version }}-${{ hashFiles('**/noxfile.py') }}-${{ hashFiles('**/setup.py') }}
+          key: ${{ runner.os }}-nox-unit-${{ matrix.python-version }}-${{ hashFiles('**/noxfile.py', '**/setup.py') }}
 
       - name: Run unit tests for GNU/Linux with Python 3.11 and generate coverage report
         if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.11
--- a/.github/workflows/test_on_push.yml
+++ b/.github/workflows/test_on_push.yml
@@ -60,7 +60,7 @@ jobs:
           execute_install_scripts: true
 
       # dot -c is for registering graphviz fonts and plugins
-      - name: Install OpenBLAS and TexLive for Linux
+      - name: Install OpenBLAS and TeXLive for Linux
         if: matrix.os == 'ubuntu-latest'
         run: |
           sudo apt-get update
@@ -105,6 +105,13 @@ jobs:
       - name: Run unit tests for GNU/Linux with Python 3.8, 3.9, and 3.10
         if: matrix.os == 'ubuntu-latest' && matrix.python-version != 3.11
         run: nox -s unit
+
+      - name: Cache unit tests nox environment for GNU/Linux with Python 3.8, 3.9, and 3.10
+        uses: actions/cache@v3
+        if: matrix.os == 'ubuntu-latest' && matrix.python-version != 3.11
+        with:
+          path: ${{ github.workspace }}/.nox/unit/
+          key: ${{ runner.os }}-nox-unit-${{ matrix.python-version }}-${{ hashFiles('**/noxfile.py') }}-${{ hashFiles('**/setup.py') }}
 
       - name: Run unit tests for GNU/Linux with Python 3.11 and generate coverage report
         if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.11
--- a/.github/workflows/test_on_push.yml
+++ b/.github/workflows/test_on_push.yml
@@ -90,7 +90,7 @@ jobs:
 
       - name: Install Windows system dependencies
         if: matrix.os == 'windows-latest'
-        run: choco install graphviz --version=2.38.0.20190211
+        run: choco install graphviz --version=8.0.5
 
       - name: Install standard Python dependencies
         run: |
--- a/.github/workflows/test_on_push.yml
+++ b/.github/workflows/test_on_push.yml
@@ -70,7 +70,7 @@ jobs:
       # Added fixes to homebrew installs:
       # rm -f /usr/local/bin/2to3
       # (see https://github.com/actions/virtual-environments/issues/2322)
-      - name: Install MacOS system dependencies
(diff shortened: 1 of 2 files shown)


agriyakhetarpal:
I will run all the checks again when the Homebrew issue with macOS builds gets fixed, and report the improvements I have noticed. On a side note, compiling SuiteSparse and SUNDIALS is a lot faster (~1.5 minutes ➡️ ~20 seconds).

agriyakhetarpal:
I am marking this PR ready for review now, this is what it does as far as the caching is concerned:

### Approach

We cache all the virtual environment folders in `.nox` which contain either PyBaMM's developer installation dependencies, documentation extras, or both. This way we already have local installations of core dependencies, but even if a dependency has a newer release, the `pip install` command ensures we have the newer version for it. The files for creating a hash digest (for the cache keys) are `noxfile.py`, `setup.py`, and `docs/requirements.txt`. Though `noxfile.py` is not needed since it only configures how the tests are run and does not contain a list of requirements like `tox.ini` used to, so we can remove it. Any changes to these hashed files will invalidate the previously built caches.

Some macOS tests have been failing across Python versions randomly due to a Homebrew issue; we might need to re-run some of the checks in this PR, so that older caches can be evicted. I have been able to successfully generate a cache hit for all `nox` environment folders across all Python versions and operating systems.

### Results

Inference on the improvements is indirect because GitHub Actions runners are not deterministic. The `nox -s pybamm-requires` command times have a significant improvement because we cache both the SuiteSparse and SUNDIALS installation folders, and the local folders where the header files and examples for CVODES, IDA, KINSOL, and the other SUNDIALS components are stored. The `pybind11/` folder is also cached, and we could also induce a minor improvement by not downloading the SuiteSparse and SUNDIALS archives again if they already exist, since we already skip `pybind11` via `nox` (and possibly also keep an MD5 hash of their contents hard-coded in the `install_KLU_sundials.py` script to ensure that the downloaded files are not corrupted, which we can get using `wget`)

For measuring improvements on the other commands, we instead look at the time taken between when the `nox -s <session>` command is run and when the tests start, which should be expected to decrease with caching since all dependencies will already be installed. There is some overhead (8–13 seconds) to copy over previously built caches to the root folder in the runners.

The tables below enlist some approximate average times across five runs, grouped for related commands, agnostic of the Python version and operating systems, except where not applicable.

| Compiling the build-time requirements | Time taken before caching | Time taken after caching |
|--------|--------|--------|
| `pybamm-requires`, to install SuiteSparse and SUNDIALS) | 1 minute 33 seconds |  18 seconds |

and while running the tests

| The `nox` test environments | The time difference between invoking the `nox` command and when the first test is successful, before caching | Time taken after storing the cache |
|--------|--------|--------|
| `unit` (no extra dependencies are installed, ~143MB)  | 1 minute 4 seconds | 13 seconds |
|  `doctests` (installs the `docs` dependencies, ~228MB) | 53 seconds | 20 seconds[^1] |
| `integration`[^2], `examples`, (the `dev` dependencies are installed, ~230MB) | 57 seconds | 19 seconds |

Overall, the caching provides only minor, _kaizen_-like benefits[^3], as was expected, but it should be nice to keep these changes.

[^1]: Highly dependent upon the time taken for parsing the `CITATIONS.bib` file and loading the `intersphinx` inventory
[^2]: The `pybamm_install_jax` command can take a variable amount of time
[^3]: A speedup may not be as evident because some sets of `nox` environments are run sequentially on the same runner, which eliminates the need for `pip` to download extra packages (requirements are already satisfied since `pip` caches them on the same build). We can probably cache the pip cache directory too in this case.

.github/workflows/test_on_push.yml
```diff
@@ -114,18 +142,46 @@ jobs:
...
 
-      - name: Run unit tests for Windows and MacOS with all Python versions
+      - name: Cache unit tests nox environment for Windows and macOS with all Python versions
+        uses: actions/cache@v3
+        if: matrix.os != 'ubuntu-latest'
+        with:
+          path: ${{ github.workspace }}/.nox/unit/
+          key: ${{ runner.os }}-nox-unit-${{ matrix.python-version }}-${{ hashFiles('**/noxfile.py', '**/setup.py') }}
+
+      - name: Run unit tests for Windows and macOS with all Python versions
```
Saransh-cpp, line 163: Is it required to cache ubuntu and (mac + windows) dependencies separately?

Unrelated to this PR, but we thanks to sing nox sessions for multiple OS, we might be able to refactor this workflow file a bit. Could we then cache everything in a single step?

agriyakhetarpal: Yes, we can definitely do it in a single step—it might even be faster because it would then perform the copying operation for the `nox` environments just once.  Should we cache the `pybamm-requires` + related files in one step, and the `unit`, `integration`, `examples`, and `doctests` environments in another step? Or should we group caching by OS?

Saransh-cpp: > Should we cache the pybamm-requires environment + related files in one step, and the unit, integration, examples, and doctests environments in another step.

I think this sounds good, given that I am trying to get rid of OS separation in the workflows - #3095

agriyakhetarpal: I was going to try this but I realised that we have different files to hash for different environments, so we will have to keep three separate cache steps: one that caches `unit` + `integration` + `examples`, one for `pybamm-requires`, and one for `doctests` (with different sets of `extras` installations)

I think it would be better to keep all environments' caches separate (i.e., as they currently are) since they contain the name of the environment too, which will make debugging easy—unless you have a better idea

agriyakhetarpal:
Should I do #3095 in this PR itself (it looks like a small one to fix)?

agriyakhetarpal:
The job step names are slightly longer now but not too long. We can remove the info about the Python version and OS if we want to keep them shorter

tinosulzer (approved):

Issue #3048, opened 2023-06-18
Improve the `nox` CI


Saransh-cpp:
Once #3005 is merged, we can improve the `nox` CI a bit more -

- [x] refactoring the env variables (will have to check, but the current version has a lot of repetitions)
- [x] refactoring the installation step in a separate Python function and calling it in every session 
- [x] [adding default sessions](https://github.com/scikit-hep/vector/blob/12dc350c28b1bfc4d1fac3e008b8b082101ca9a5/noxfile.py#L9)
- [x] [adding a `pre-commit` session](https://github.com/scikit-hep/vector/blob/12dc350c28b1bfc4d1fac3e008b8b082101ca9a5/noxfile.py#L15-L19)
- [x] adding more informative commands in docs, like `nox -l`, `nox -p`
- [x] Cache nox environment for speedup
- [x] make `nox -s dev` run like `nox -s dev --verbose`
- [ ] add [docstrings for nox sessions](https://nox.thea.codes/en/stable/config.html#session-description)

cc : @arjxn-py @agriyakhetarpal

agriyakhetarpal:
See also: [The Nox Cookbook](https://nox.thea.codes/en/stable/cookbook.html)

PR #3082, opened 2023-06-29
Refactor env variables in a single function

arjxn-py:
# Description

Currently environment variables in `noxfile.py` are being set in each session explicitely, This PR aims to refactor & reduce the repetations of setting those variables again & again.

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:


arjxn-py:
Related to #3048

arjxn-py:
MacOS builds are failing randomly, might because of `brew` or `virtualbox`. I do not have much idea of it but may I do something in that regard?

agriyakhetarpal:
GitHub Actions and Homebrew have been currently facing downtimes intermittently (https://github.com/Homebrew/brew/issues/15610) since the past day, I guess it will be fixed soon

arjxn-py:
> GitHub Actions and Homebrew have been currently facing downtimes intermittently ([Homebrew/brew#15610](https://github.com/Homebrew/brew/issues/15610)) since the past day, I guess it will be fixed soon

I see, thanks for the update. We're cool then.

Saransh-cpp:
pre-commit.ci run

noxfile.py
```diff
@@ -42,51 +44,42 @@ def run_coverage(session):
...
+@set_env.register(complex)
 @nox.session(name="dev", reuse_venv=True)
 def set_dev(session):
-    homedir = os.getenv("HOME")
-    LD_LIBRARY_PATH = f"{homedir}/.local/lib:{session.env.get('LD_LIBRARY_PATH')}"
-    envbindir = session.bin
-    session.install("-e", ".[dev]")
     session.install("cmake")
     session.run(
         "echo",
```
Saransh-cpp, line 85: pre-commit CI is erroring out here (correctly) -
```
noxfile.py:87:28: F821 Undefined name `LD_LIBRARY_PATH`
noxfile.py:89:12: F821 Undefined name `envbindir`
```

arjxn-py:
pre-commit.ci run

agriyakhetarpal:
Note that we can do something like this too, since [session.env](https://nox.thea.codes/en/stable/config.html#nox.sessions.Session.env) accepts a dictionary of strings, but we pass them one by one as dictionary objects are not callable:

```python
# set constants for any and all required environment variables
HOMEDIR = os.getenv("HOME")
LD_LIBRARY_PATH = os.getenv("LD_LIBRARY_PATH")

ENV_VARS = {
    "SUNDIALS_INST": f"{HOMEDIR}/.local",
    "LD_LIBRARY_PATH": f"{HOMEDIR}/.local/lib:{LD_LIBRARY_PATH}",
}

# to be called inside relevant functions decorated by @nox.session
def set_environment_variables(session, env_vars=ENV_VARS):
    for key, value in env_vars.items():
        session.env[key] = value
```

This code snippet works but it needs a bit of refactoring, this was only my initial attempt 😅

arjxn-py:
> This code snippet works but it needs a bit of refactoring, this was only my initial attempt 😅

Thanks so much @agriyakhetarpal , I think I tried this in this [commit](https://github.com/pybamm-team/PyBaMM/pull/3082/commits/64fbdde1b59bd07ddaf69180c1a50f00652498dc) & ENV_VARS were setting as expected. 
Linux tests are failing while building `scikits.odes`, currently don't know why.

arjxn-py:
Marking this PR ready for review as CI is passing now. @Saransh-cpp @agriyakhetarpal is there any way I can pass pre-commit check?

agriyakhetarpal:
We should fix the error if possible. Otherwise, silence it with `# noqa: F821`

noxfile.py
```diff
@@ -57,62 +51,55 @@ def run_integration(session):
 
 @nox.session(name="doctests", reuse_venv=True)
 def run_doctests(session):
+    set_environment_variables(session=session)
```
Saransh-cpp, line 54:
```suggestion
```

noxfile.py
```diff
@@ -57,62 +51,55 @@ def run_integration(session):
...
     if sys.platform == "linux":
-        session.run("pybamm_install_jax")
         session.install("scikits.odes")
+        session.run("pybamm_install_jax")
     session.run("python", "run-tests.py", "--unit")
 
 
 @nox.session(name="examples", reuse_venv=True)
 def run_examples(session):
+    set_environment_variables(session=session)
```
Saransh-cpp, line 71:
```suggestion
```

noxfile.py
```diff
@@ -57,62 +51,55 @@ def run_integration(session):
...
     session.run("python", "run-tests.py", "--examples")
 
 
 @nox.session(name="dev", reuse_venv=True)
 def set_dev(session):
-    homedir = os.getenv("HOME")
-    LD_LIBRARY_PATH = f"{homedir}/.local/lib:{session.env.get('LD_LIBRARY_PATH')}"
+    set_environment_variables(session=session)
     envbindir = session.bin
-    session.install("-e", ".[dev]")
```
Saransh-cpp, line 89: I think this should not have been removed?

noxfile.py
```diff
@@ -57,62 +51,55 @@ def run_integration(session):
...
     if sys.platform == "linux" or sys.platform == "darwin":
-        session.run("pybamm_install_jax")
         session.install("scikits.odes")
+        session.run("pybamm_install_jax")
     session.run("python", "run-tests.py", "--all")
 
 
 @nox.session(name="docs", reuse_venv=True)
 def build_docs(session):
+    set_environment_variables(session=session)
```
Saransh-cpp, line 102:
```suggestion
```

Saransh-cpp (changes requested):
Looks great, thanks @arjxn-py! Some comments below -

noxfile.py
arjxn-py:
```python 
    homedir = os.getenv("HOME")
    LD_LIBRARY_PATH = f"{homedir}/.local/lib:{session.env.get('LD_LIBRARY_PATH')}"
    envbindir = session.bin
    session.install("-e", ".[dev]")
```
Added some extra as locally it is breaking with : 
```python 
NameError: name 'LD_LIBRARY_PATH' is not defined
```

noxfile.py
```diff
@@ -73,8 +73,10 @@ def run_examples(session):
 
 @nox.session(name="dev", reuse_venv=True)
 def set_dev(session):
-    set_environment_variables(session=session)
+    homedir = os.getenv("HOME")
+    LD_LIBRARY_PATH = f"{homedir}/.local/lib:{session.env.get('LD_LIBRARY_PATH')}"
     envbindir = session.bin
+    session.install("-e", ".[dev]")
     session.install("cmake")
     session.run(
```
Saransh-cpp, line 81: Does this session really require the `run` line? I think the env variables can just be set using `set_environment_variables(session=session)`?

arjxn-py: I believe it is required as we are using `bash` to export variables in the shell.

Saransh-cpp: But we aren't doing this in other sessions?

agriyakhetarpal:
I think we should modify `set_environment_variables(session)` to take an arbitrary environment variable key-value pair as an argument:

```python
def set_environment_variables(session, env_dict):
    """
    Sets environment variables for a nox session object.

    Args:
        session (nox.session): The session to set the environment
            variables for.
        env_dict (dict): A dictionary of environment variable names and values.

    """
    for key, value in env_dict.items():
        session.env[key] = value
```

and to keep it DRY, use a constant `PYBAMM_ENV` as a dictionary at the top of the file that gets passed on in the sessions we want with `set_environment_variables(session=session, PYBAMM_ENV)`

noxfile.py
arjxn-py: We are't doing it other sessions because we don't have to do `source .nox/session/bin/activate` with any other session.
But in case to dev it is required to activate the dev env further.

Saransh-cpp: Makes sense, but we should not redefine the variables in `set_dev`, instead, we can read them from the environment and pass those values in the `run` command

noxfile.py
```diff
@@ -8,13 +8,30 @@
...
+
+
+def set_environment_variables(env_dict, session):
+    """
+    Sets environment variables for a nox session object.
+
+    Args:
+        session (nox.session): The session to set the environment
+            variables for.
+        env_dict (dict): A dictionary of environment variable names and values.
```
Saransh-cpp, line 25:
```suggestion
    Parameters
    -----------
        session : nox.Session
            The session to set the environment variables for.
        env_dict : dict
            A dictionary of environment variable names and values.
```

Saransh-cpp (changes requested):
The `noxfile` looks so clean now! Thanks, @arjxn-py! Some comments below.

PR #3091, opened 2023-07-03
Fix scheduled tests

agriyakhetarpal:
# Description

A quick fix to quick the failing doctests step in the scheduled nightly tests, after #3043 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/.github/workflows/run_periodic_tests.yml
+++ b/.github/workflows/run_periodic_tests.yml
@@ -63,7 +63,7 @@ jobs:
         if: matrix.os == 'ubuntu-latest'
         run: |
           sudo apt-get update
-          sudo apt install gfortran gcc libopenblas-dev graphviz
+          sudo apt install gfortran gcc libopenblas-dev graphviz pandoc
           sudo apt install texlive-full
 
       # Added fixes to homebrew installs:


arjxn-py:
@agriyakhetarpal macOS tests are also failing randomly, I suggest we also try to fix those in this PR only

agriyakhetarpal:
I don't think they are failing on our end

brosaplanella:
The mac tests are failing randomly due to some Github issue. It has been solved, but it is being rolled out so still randomly fails (if I understood correctly). I think it is safe to merge.

More info: https://github.com/Homebrew/homebrew-core/issues/135315

brosaplanella (approved):

brosaplanella:
Looks good to me, but didn't quite understand why pandoc is needed.

agriyakhetarpal:
It is needed for converting the notebooks when building the docs (https://github.com/spatialaudio/nbsphinx/issues/36)

.github/workflows/run_periodic_tests.yml
```diff
@@ -63,7 +63,7 @@ jobs:
         if: matrix.os == 'ubuntu-latest'
         run: |
           sudo apt-get update
-          sudo apt install gfortran gcc libopenblas-dev graphviz
+          sudo apt install gfortran gcc libopenblas-dev graphviz pandoc
```
Saransh-cpp, line 66: Will this be required while building the docs locally, or is it only required for deployment? We might want to add this to the dev docs if it is an absolute requirement.

Saransh-cpp: Ah, great!

Saransh-cpp:
I missed the discussion on this PR in the meet. Is this ready for merging?

agriyakhetarpal:
We missed discussing this one apparently, should be ready to go

PR #3096, opened 2023-07-04
fix Prada parameter set

### LINKED ISSUE #1811: Prada2013 LFP Cell data
I am a little confused if the active material fraction is correct in the CSV file of LFP positive electrode (Prada 2013). For example, for Prada2013 (A123 M1A LFP Cell) positive active material fraction is 0.28 and porosity fraction is 0.12 in the csv file. Lain2019 in his paper says that the cathode composition in M1A cell is active:binder:carbon=79:11:10. Can you comment on this mismatch

tinosulzer:
# Description

The Prada2013 parameter set was a mix of different papers that gave inconsistent results.
It now matches the parameters from the Prada2013 paper, except OCPs which are not given in the paper.
The exchange-current densities are also from a different paper since the values they give in the paper are strange.
Removed thermal parameters since they don't have a thermal model. We should make it clear when thermal parameters just come from the default set we use everywhere

Fixes #1811 

## Type of change

# Key checklist:

## Further checks:


brosaplanella (approved):
Looks good, thanks!

PR #3104, opened 2023-07-04
Document other useful `nox` commands

arjxn-py:
# Description

Make nox -s dev run like nox -s dev --verbose to so that all commands run is shown.

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/docs/source/user_guide/installation/install-from-source.rst
+++ b/docs/source/user_guide/installation/install-from-source.rst
@@ -228,3 +228,18 @@ Doctests, examples, style and coverage
 
 
 Use ``nox -s coverage`` to measure current test coverage on all platforms.
+
+Extra tips while using Nox
+--------------------------
+Here are some additional useful commands you can run with Nox:
+
+- ``nox -v``: Enables verbose mode, providing more detailed output during the execution of Nox sessions.
+- ``nox -l``: Lists all available Nox sessions and their descriptions.
+- ``nox -p <python.version>``: Runs only sessions with the specified Python version. Replace ``python.version`` with the desired Python version.
+- ``nox -k "<keyword_expression>"``: Runs test sessions that match the specified pytest-style keyword expression.
+- ``nox -t "<tag1>" "<tag2>"``: Runs test sessions that have the specified tags.
+- ``nox --stop-on-first-error``: Stops the execution of Nox sessions immediately after the first error or failure occurs.
+- ``nox --envdir <path>``: Specifies the directory where Nox creates and manages the virtual environments used by the sessions. In this case, the directory is set to ``<path>``.
+- ``nox --install-only``: Skips the test execution and only performs the installation step defined in the Nox sessions.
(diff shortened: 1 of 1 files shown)


agriyakhetarpal:
The --verbose flag allows developers to see what packages are installed by `pip` in the environments, which is especially useful for the `unit`, `examples`, and the `docs` environments. Is there a way to increase the verbosity for all `nox` environments by default?

arjxn-py:
> The --verbose flag allows developers to see what packages are installed by `pip` in the environments, which is especially useful for the `unit`, `examples`, and the `docs` environments. Is there a way to increase the verbosity for all `nox` environments by default?

There doesn't seem to be having much info available about increasing the verbosity level by default atm. 

I could find this much according to the docs : 
>> ## Controlling commands verbosity
>> By default, Nox will only show output of commands that fail, or, when the commands get passed silent=False. By passing --verbose to Nox, all output of all commands run is shown, regardless of the silent argument.

arjxn-py:
Apparantly I tried doing this & some other things as well
```Python 
nox.options.verbose = True
```
But it did't make any effect on verbosity.

docs/source/user_guide/installation/install-from-source.rst
```diff
@@ -109,6 +109,8 @@ Using Nox (recommended)
 		# in the PyBaMM/ directory
 		nox -s dev
 
+**Note** : It is recommended to use ``--verbose`` or ``-v`` to see outputs of all commands run.
```
agriyakhetarpal, line 112:
```suggestion
.. note::
    It is recommended to use ``--verbose`` or ``-v`` to see outputs of all commands run.
```

agriyakhetarpal: Just some better formatting

agriyakhetarpal (changes requested):

docs/source/user_guide/installation/install-from-source.rst
arjxn-py: Thanks @agriyakhetarpal

Saransh-cpp (approved):
Looks great, thanks @arjxn-py!

agriyakhetarpal (approved):
I checked the other command-line arguments if we can add any more, I think we should also document the `--install-only` command since it allows the creation of an environment but only installs the dependencies, to keep it ready for later use—this could be useful for the `unit` and `docs` environments

arjxn-py:
> I checked the other command-line arguments if we can add any more, I think we should also document the `--install-only` command since it allows the creation of an environment but only installs the dependencies, to keep it ready for later use—this could be useful for the `unit` and `docs` environments

Yes, I had added it [here](https://github.com/pybamm-team/PyBaMM/pull/3104/files#diff-ceedebcbf80d97f94255635c9cf4d2b5105c89a1e099f1e090127ae35d3f08b2R241)

PR #3108, opened 2023-07-05
Configure `versions.json` and remove the version switcher dropdown

agriyakhetarpal:
# Description

Adds a version banner at the top of the page for old and unstable versions, and configures the version switcher dropdown to work again

## Type of change

# Key checklist:

## Further checks:


agriyakhetarpal:
For some reason, the switcher dropdown on readthedocs has disappeared for all our versions and does not work locally either. I double-checked the PyData theme documentation and our configuration, and compared it with that of NumPy and SciPy too just in case I may have missed something. 

I can confirm that the switcher was working earlier, however: https://github.com/pybamm-team/PyBaMM/pull/2769#issuecomment-1466007620. It could be due to a related JS file not loading itself when it should (i.e., upon clicking the switcher). I will keep this a draft for now until I am able to dig further 😕

tinosulzer (approved):

agriyakhetarpal:
I see that you approved this @tinosulzer, but it is not ready to merge yet, let's keep this open

agriyakhetarpal:
Marking this as ready for review since I think it would make sense to wait for https://github.com/pydata/pydata-sphinx-theme/pull/1354 to get the warning bar from the theme itself; we may discuss this in this week's meeting

We should add the warning bar when the theme has a new release and configure `versions.json` before our `23.9` release in a separate PR later, so that the currently stable build (`23.5`) will start displaying the version warning banner when it becomes an old release. Meanwhile, I would suggest merging this PR so that it does not become outdated afterwards lest it is kept open

Saransh-cpp (approved):
Looks good, thanks @agriyakhetarpal! I am guessing that there are no blockers for this PR now? You might also want to merge in develop.

agriyakhetarpal:
No blockers, but I am yet to be convinced whether the logic I wrote for setting the switcher is correct. I would say we should wait for a while

docs/conf.py
```diff
@@ -168,6 +168,25 @@
...
+# the version type is "external"), and we set the versions accordingly
+
+if version_match is None or version_type == "external":
+    # Use local versions.json if not building on readthedocs
+    # or if building on readthedocs but on PRs
+    html_theme_options["switcher"]["json_url"] = "_static/versions.json"
+
+if version_type == "external":
+    # Set the version to latest on readthedocs PRs
+    html_theme_options["switcher"]["version_match"] = "latest"
```
OriolAbril, line 188: This will only set the `html_theme_options["switcher"]["version_match"]`, it won't work for `stable` nor `latest`. Currently, they'll get `version = pybamm.__version__` (defined in line 33) which is not the version you want to have matched to them IUUC. In addition to these checks to cover local and PR builds, I think you should also have specific checks for stable and latest. They do have independent builds and they appear on the versions.json file, so their builds should have the right value of the version match. Otherwise, users will click on latest and go to `.../en/latest/...` docs, but the "folded" version switcher won't show latest while browsing the docs.

agriyakhetarpal: Thanks for the review! As noted previously in our conversation—you mentioned that the version switcher is still not going to work for previous documentation releases, so we are deciding to remove the dropdown altogether in order to rely on the server-side readthedocs one. That is, unless there exists a workaround for not having to deal with patch releases with the latest `versions.json` set?

agriyakhetarpal:
@Saransh-cpp let's get this in; I don't think we can get the switcher on older versions anyway (https://github.com/pydata/pydata-sphinx-theme/issues/1389#issuecomment-1652120798). The next step after this is merged will be to add the theme's version warning bar that will come with their next release, before our release

PR #3151, opened 2023-07-14
set T_init as a function in the x domain

tommaull:
# Description

Changed T_init to functional parameter in the x domain

Fixes # 3131 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/parameters/thermal_parameters.py
+++ b/pybamm/parameters/thermal_parameters.py
@@ -10,10 +10,13 @@ class ThermalParameters(BaseParameters):
     Standard thermal parameters
     """
 
-    def __init__(self):
+    def __init__(self, options=None):
+        self.options = options
+
         # Get geometric parameters
         self.geo = pybamm.geometric_parameters
 
+        # Set domain attribute
         self.n = DomainThermalParameters("negative", self)
         self.s = DomainThermalParameters("separator", self)
         self.p = DomainThermalParameters("positive", self)
@@ -28,8 +31,9 @@ class ThermalParameters(BaseParameters):
 
     def _set_parameters(self):
         """Defines the dimensional parameters"""
-        for domain in self.domain_params.values():
-            domain._set_parameters()
+        for domain, params in self.domain_params.items():
+            params._domain = domain  # Set _domain attribute
+            params._set_parameters()
 
         # Reference temperature
         self.T_ref = pybamm.Parameter("Reference temperature [K]")
@@ -38,8 +42,25 @@ class ThermalParameters(BaseParameters):
         self.h_edge = pybamm.Parameter("Edge heat transfer coefficient [W.m-2.K-1]")
         self.h_total = pybamm.Parameter("Total heat transfer coefficient [W.m-2.K-1]")
 
+        x = pybamm.SpatialVariable(
+            f"x_{domain[0]}",
+            domain=[f"{domain} electrode"],
+            auxiliary_domains={"secondary": "current collector"},
+            coord_sys="cartesian",
+        )
+        # Domain parameters
+        for domain in self.domain_params.values():
+            domain._set_parameters()
+
         # Initial temperature
-        self.T_init = pybamm.Parameter("Initial temperature [K]")
+        self.T_init = pybamm.FunctionParameter(
+            f"{domain}Initial temperature in {domain} electrode [mol.m-3]",
+            {
+                "Through-cell distance (x) [m]": pybamm.PrimaryBroadcast(
+                    x, f"{domain} particle"
+                ),
+            },
+        )
 
     def T_amb(self, t):
         """Dimensional ambient temperature"""


tommaull:
Lots of test failures, opening PR for visibility only.

tommaull:
So I'm getting an error that I don't fully understand within the symbol.py, called through the equation tree. There appears to be math operation where two different domains are being subtracted. (Current collector and particle).

brosaplanella:
My guess is that the errors are caused because now `T_init` is a function of `x` but this has not been changed in the other scripts, so you need to pass the additional argument to all instances of `T_init`.

tommaull:
OK, so got some more stuff working, but now it seems like there's some confusion around the domain variable between the submodels and the base model. 

Error trace:

File [c:\users\tom.maull\desktop\pybammdev\pybamm\pybamm\models\submodels\thermal\lumped.py:24](file:///C:/users/tom.maull/desktop/pybammdev/pybamm/pybamm/models/submodels/thermal/lumped.py:24), in Lumped.__init__(self, param, domain, options)
     23 def __init__(self, param, domain, options=None):
---> 24     super().__init__(param, domain, options=options)
     25     pybamm.citations.register("Timms2021")

File [c:\users\tom.maull\desktop\pybammdev\pybamm\pybamm\models\submodels\thermal\base_thermal.py:23](file:///C:/users/tom.maull/desktop/pybammdev/pybamm/pybamm/models/submodels/thermal/base_thermal.py:23), in BaseThermal.__init__(self, param, domain, options)
     21 def __init__(self, param, domain, options=None):
     22     super().__init__(param, options=options)
---> 23     self.domain = domain

File [c:\users\tom.maull\desktop\pybammdev\pybamm\pybamm\models\submodels\base_submodel.py:123](file:///C:/users/tom.maull/desktop/pybammdev/pybamm/pybamm/models/submodels/base_submodel.py:123), in BaseSubModel.domain(self, domain)
    120 @domain.setter
    121 def domain(self, domain):
    122     if domain is not None:
--> 123         domain = domain.lower()
    124     ok_domain_list = ["negative", "separator", "positive", None]
    125     if domain in ok_domain_list:

tommaull:
image

tommaull:
Hey is anyone able to take a look at this one?

pybamm/parameters/thermal_parameters.py
```diff
@@ -28,17 +31,40 @@ def __init__(self):
...
 
         # Cooling coefficient
         self.h_total = pybamm.Parameter("Total heat transfer coefficient [W.m-2.K-1]")
 
+        x = pybamm.SpatialVariable(
+            f"x_{domain[0]}",
+            domain=[f"{domain} electrode"],
+            auxiliary_domains={"secondary": "current collector"},
+            coord_sys="cartesian",
+        )
```
brosaplanella, line 49: I think the issue with the tests is here as the domain is inherited from the for loop and it is not a string.

brosaplanella:
Hi @tommaull! Let me know if you have any questions about the comments above (not sure I was very clear haha).

tommaull:
Hi @brosaplanella and @rtimms, here's the latest. I've definitely bitten off more than I should have here. Appreciate your help.

pybamm/parameters/thermal_parameters.py
```diff
@@ -28,17 +31,40 @@ def __init__(self):
 
     def _set_parameters(self):
         """Defines the dimensional parameters"""
-        for domain in self.domain_params.values():
-            domain._set_parameters()
+        for domain, params in self.domain_params.items():
+            params._domain = domain  # Set _domain attribute
```
rtimms, line 35: I don't think this is needed, since `T` is defined across all domains, not by component

pybamm/parameters/thermal_parameters.py
rtimms: this should be the "whole cell" x, i.e.
```python3
whole_cell = ["negative electrode", "separator", "positive electrode"]
x = pybamm.SpatialVariable(
    "x",
    domain=whole_cell,
    coord_sys="cartesian",
)
```
or, even better, just do `pybamm.standard_spatial_vars.x` which will fetch the same defn from `standard_spatial_vars.py`

pybamm/parameters/thermal_parameters.py
```diff
@@ -28,17 +31,40 @@ def __init__(self):
...
         # Cooling coefficient
         self.h_total = pybamm.Parameter("Total heat transfer coefficient [W.m-2.K-1]")
 
+        x = pybamm.SpatialVariable(
+            f"x_{domain[0]}",
+            domain=[f"{domain} electrode"],
+            auxiliary_domains={"secondary": "current collector"},
+            coord_sys="cartesian",
+        )
+        # Domain parameters
```
rtimms, line 50: now this can go, for the same reasons as above

pybamm/models/submodels/thermal/base_thermal.py
```diff
@@ -12,6 +12,8 @@ class BaseThermal(pybamm.BaseSubModel):
     ----------
     param : parameter class
         The parameters to use for this submodel
+    domain : str
```
rtimms, line 15: not needed, thermal submodel is set across whole cell, not by domain

pybamm/models/submodels/thermal/lumped.py
```diff
@@ -67,4 +67,4 @@ def set_rhs(self, variables):
 
     def set_initial_conditions(self, variables):
         T_vol_av = variables["Volume-averaged cell temperature [K]"]
-        self.initial_conditions = {T_vol_av: self.param.T_init}
+        self.initial_conditions = {T_vol_av: pybamm.x_average(self.param.T_init)}
```
rtimms, line 70: will also need to do this in the "x-lumped" models for pouch cell (in the folder `submodels/thermal/pouch_cell/`)

rtimms:
thanks @tommaull it's getting there - just some confusion around the cell temperature being set on the "whole cell" domain instead of separately for negative electrode/separator/positive electrode

tommaull:
Failing tests where non-lumped models are used due to mismatch between eqn and var domains. 
image

Seems to be because the variable is set for Negative electrode temperature [K], Separator temperature [K], Positive electrode temperature [K] but the equation just has initial temp. 
image

pybamm/parameters/thermal_parameters.py
```diff
@@ -28,17 +31,39 @@ def __init__(self):
...
+            #params._domain = domain  # Set _domain attribute
+            params._set_parameters()
 
         # Reference temperature
         self.T_ref = pybamm.Parameter("Reference temperature [K]")
 
         # Cooling coefficient
         self.h_total = pybamm.Parameter("Total heat transfer coefficient [W.m-2.K-1]")
 
+        whole_cell = ["negative electrode", "separator", "positive electrode"]
```
tommaull, line 44: does this need to be a string, not a list?

tommaull:
image
So I've been trying to suss out the reason for the domain mismatch, and it seems to be, at least partially, because the x_average or Xaverage functions only return a value in the secondary domain. I'm not sure I understand why this occurs?

tinosulzer:
`domain` can be thought of as "the primary spatial dimension along which this variable changes" which is why `XAverage` has this behavior. `T(x,y,z,t)` changes primarily along the through-cell (x) dimension, then secondarily along the current collector (y,z) dimension(s), its x-average is `T_av(y,z,t)` which changes primarily along the current collector dimension. Depending on the geometry the "current collector" dimension might just be 0D. Variables that depend only on time have no spatial dimension hence no domain.

If you want a variable with primary domains matching the original variable, you can re-broadcast 
```python
pybamm.PrimaryBroadcast(T_av, ["negative electrode", "separator", "positive electrode"])
```

PR #3157, opened 2023-07-18
Ensure `setup-python` outputs are always included in `nox` cache keys

agriyakhetarpal:
# Description

A follow-up to #3079 and fixes the failing Windows tests with Python 3.11 on #3108. This PR ensures that the Python version (including sub-versions and patch releases) from the outputs of the `setup-python` action is used to build the cache keys and invalidates previously built caches when there is a version change (i.e., from Python 3.11.3 to Python 3.11.4), making the caching more robust. See https://github.com/actions/setup-python/issues/330#issuecomment-1126638275 for more details.

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/.github/workflows/test_on_push.yml
+++ b/.github/workflows/test_on_push.yml
@@ -46,7 +46,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
-      - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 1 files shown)


Saransh-cpp (approved):
Thanks, @agriyakhetarpal!

PR #3110, opened 2023-07-05
Add `nbqa` support support for example notebooks

arjxn-py:
# Description

Adding nbqa as pre-commit hook for jupyter notebooks.

Fixes #2813 

## Type of change

Please add a line in the relevant section of [CHANGELOG.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CHANGELOG.md) to document the change (include PR #) - note reverse order of PR #s. If necessary, also add to the list of breaking changes.

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (back-end change that speeds up the code)
- [ ] Bug fix (non-breaking change which fixes an issue)

# Key checklist:

- [ ] No style issues: `$ pre-commit run` (see [CONTRIBUTING.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CONTRIBUTING.md#installing-and-using-pre-commit) for how to set this up to run automatically when committing locally, in just two lines of code)
- [ ] All tests pass: `$ python run-tests.py --all`
- [ ] The documentation builds: `$ python run-tests.py --doctest`

You can run unit and doctests together at once, using `$ python run-tests.py --quick`.

## Further checks:

- [ ] Code is commented, particularly in hard-to-understand areas
- [ ] Tests added that prove fix is effective or that feature works

PR #3166, opened 2023-07-19
Fix SEI example notebook

brosaplanella:
Fix the failing notebook, as identified in https://github.com/pybamm-team/PyBaMM/pull/3110#discussion_r1267382517. I also wrote some extra details about how to fix the moving boundary, which were part of the non-dimensionalisation and thus were lost with the shift to dimensional models.


brosaplanella:
@arjxn-py this should fix the problems with #3110, so once we merge this into develop you can merge develop back into your branch.

arjxn-py:
> @arjxn-py this should fix the problems with #3110, so once we merge this into develop you can merge develop back into your branch.

Thanks so much @brosaplanella 😄

rtimms (approved):
thanks @brosaplanella looks good to me!

PR #3171, opened 2023-07-20
Changed fundamental SEI variable from thickness to concentration

### LINKED ISSUE #3006: [Bug]: LLI due to SEI incorrect if loss of active material is enabled
### PyBaMM Version

23.4.1

### Python Version

3.8.10

### Describe the bug

In order to convert SEI thickness into SEI concentration, the surface area to volume ratio a is required. However, if any of the loss of active material models are enabled, a changes over time.

Prior to 15th November 2022, a_typ is used to convert from thickness to concentration. This results in SEI growth being overestimated, with SEI growing on particles that have already been lost.

On 15th November 2022, a_typ was replaced with the time-dependent a(t). However, this is also flawed. If a particle that already had SEI on it is lost, PyBaMM considers that SEI to also be lost, which results in "Loss of lithium due to SEI [mol]" decreasing over time in some cases.

One solution could be to have separate ODEs for the various LLI variables so they are incremented at the point the lithium is lost, but that would be computationally expensive and surely there must be a better way?

### Steps to Reproduce

```
import pybamm
model = pybamm.lithium_ion.DFN({
    "SEI": "interstitial-diffusion limited",
    "loss of active material": ("stress-driven", "none")
})
param.update({
    "Inner SEI reaction proportion": 0.5,
    "Initial inner SEI thickness [m]": 2.5e-09,
    "Initial outer SEI thickness [m]": 2.5e-09,
    "Ratio of lithium moles to SEI moles": 2.0,
})
var_pts = {"x_n": 20, "x_s": 20, "x_p": 20, "r_n": 30, "r_p": 30}
exp = pybamm.Experiment(["Discharge at 1C until 2.5 V", "Charge at 0.3C until 4.2 V", "Hold at 4.2 V until C/100"])
sim = pybamm.Simulation(model, parameter_values=param, experiment=exp)
sol = sim.solve()
t = sol["Time [s]"].entries
Q_SEI = sol["Loss of capacity to SEI [A.h]"].entries
plt.figure()
plt.plot(t,Q_SEI)
plt.show()
```

### Relevant log output

_No response_

DrSOKane:
# Description

Previously, lithium was not conserved if both SEI growth and loss of active material were enabled. To fix this, the fundamental variable to be solved for is now SEI concentration, not thickness. This has the added bonus of simplifying the SEI on cracks model.

Fixes #3006

## Type of change

# Key checklist:

## Further checks:


pybamm/models/submodels/interface/sei/sei_growth.py
```diff
@@ -198,73 +221,76 @@ def set_rhs(self, variables):
...
-            phase_param.V_bar_inner * j_inner / (param.F * phase_param.z_sei)
-        )
-        dLdt_SEI_outer = (
-            phase_param.V_bar_outer * j_outer / (param.F * phase_param.z_sei)
-        )
-
-        # we have to add the spreading rate to account for cracking
+        dLdt_SEI_inner = a * j_inner / (param.F * phase_param.z_sei)
+        dLdt_SEI_outer = a * j_outer / (param.F * phase_param.z_sei)
+
```
tinosulzer, line 255: should these be `dLdt` or `dcdt` ?

tinosulzer (commented):
Thanks, this is great! Exactly what I was envisioning.
Could you post some "before and after" plots of SEI thickness and concentration vs time just to confirm that the results are unchanged (except where the lithium conservation bugs are fixed)? 
Are there any tests you can think of adding to verify that lithium is now conserved (these tests should fail on develop but pass on this branch)?

valentinsulzer (dismissed):
Thanks, test looks good, need to fix the failing example though

DrSOKane:
Hi Tino, the test is NOT good as it passed even on develop. I changed it so it now uses the `loss of lithium due to loss of active material` variable, but after much experimenting I found this variable was calculated using the incorrect assumption that the average of a product is the same as the product of averages. I managed to fix this, but the test still fails!

Here are three plots, all of which were calculated using the following code:
```
model = pybamm.lithium_ion.DFN({
    "SEI": "reaction limited",
    "particle mechanics": ("swelling and cracking", "swelling only"),
    "SEI on cracks": "true",
    "loss of active material": "stress and reaction-driven",
})
var_pts = {"x_n": 5, "x_s": 5, "x_p": 5, "r_n": 30, "r_p": 20}
sim = pybamm.Simulation(
    model, parameter_values=param, var_pts=var_pts, solver=solver
)
solution = sim.solve([0, 3500])
t = solution["Time [s]"].entries
particles = solution["Total lithium in particles [mol]"].entries
neg = solution["Loss of lithium due to loss of active material in negative electrode [mol]"].entries
pos = solution["Loss of lithium due to loss of active material in positive electrode [mol]"].entries
side = solution["Total lithium lost to side reactions [mol]"].entries
plt.figure()
plt.plot(t,particles+side+neg+pos)
plt.xlabel("Time [s]")
plt.ylabel("Lithium in particles + lithium lost [mol]")
plt.show()
```
coupled_degradation_test_before
coupled_degradation_test_after_SEI
coupled_degradation_test_after_both
The third plot is a massive improvement on the first two but still doesn't meet the requirement of 12 decimal places.

Regarding the failing example, I have no idea how to fix it! It's saying `Loss of lithium to SEI [mol]` has a domain `[`current collector`]`. I ran it in debug mode, set a checkpoint in `_get_standard_total_concentration_variables` and every time the checkpoint was triggered, the variable in question had a domain of `[]`, as expected for a yz_averaged variable.

Give yourself a gold star for reading this essay!

RuiheLi:
A bit more information on the glithces Simon found regarding `Loss of lithium to SEI [mol]` has a domain `[current collector]`.
I find that: (1) this only occurs in SPM model, for SPMe or DFN it is fine; (2) this may relate to how PyBaMM handle solution. 
If we print the type of `Loss of lithium to SEI [mol]`, which is Q_sei in `def _get_standard_total_concentration_variables` under pybamm/models/submodels/interface/sei/base_sei.py, then it shows that it is `'pybamm.expression_tree.scalar.Scalar'>` or 'pybamm.expression_tree.binary_operators.Multiplication'>

however, when I do `type(sol["Loss of lithium to SEI [mol]"])`, it shows that it is `pybamm.solvers.processed_variable.ProcessedVariable`

rtimms:
@DrSOKane, thanks for this! have you had any luck in debugging the notebook? would be good to get this in for the next release

brosaplanella:
Might be worth to merge develop into this branch first, as we have improved how notebooks are tested, which might help debug.

agriyakhetarpal:
Note: that also means they need to be written in a slightly different way, i.e. in a literate programming sense rather than as a Python script which they were earlier converted to at the time of testing

rtimms:
@DrSOKane can you update this branch and then point me to where things break and I can take a look? This is what we worked on debugging when I visited Imperial earlier this year, right?

DrSOKane:
Superseded by #4869

PR #3175, opened 2023-07-21
#3160 update versioning in README

### LINKED ISSUE #3160: Update Versioning text in README
The [versioning section](https://github.com/pybamm-team/PyBaMM#versioning) needs updating to reflect the new release schedule (3/year).

brosaplanella:
# Description
Fixed the README to explain the most recent versioning.

Fixes #3160

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ Note that the examples on the default `develop` branch are tested on the latest 
 
 ## Versioning
 
-PyBaMM uses [CalVer](https://calver.org/), which means that we make new releases every month with the version number `YY.MM`. There is no difference between releases that increment the year and releases that increment the month; in particular, releases that increment the month may introduce breaking changes. Breaking changes for each release are communicated via the [CHANGELOG](CHANGELOG.md), and come with deprecation warnings or errors that are kept for at least one year (12 releases). If you find a breaking change that is not documented, or think it should be undone, please open an issue on [GitHub](https://github.com/pybamm-team/pybamm).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -94,7 +94,7 @@ Note that the examples on the default `develop` branch are tested on the latest
...
+PyBaMM makes releases every four months and we use [CalVer](https://calver.org/), which means that the version number is `YY.MM`. The releases happen, approximately, at the end of January, May and September. There is no difference between releases that increment the year and releases that increment the month; in particular, releases that increment the month may introduce breaking changes. Breaking changes for each release are communicated via the [CHANGELOG](CHANGELOG.md), and come with deprecation warnings or errors that are kept for at least one year (12 releases). If you find a breaking change that is not documented, or think it should be undone, please open an issue on [GitHub](https://github.com/pybamm-team/pybamm).
```
Saransh-cpp, line 97: Should this be 3 now?
```suggestion
PyBaMM makes releases every four months and we use [CalVer](https://calver.org/), which means that the version number is `YY.MM`. The releases happen, approximately, at the end of January, May and September. There is no difference between releases that increment the year and releases that increment the month; in particular, releases that increment the month may introduce breaking changes. Breaking changes for each release are communicated via the [CHANGELOG](CHANGELOG.md), and come with deprecation warnings or errors that are kept for at least one year (3 releases). If you find a breaking change that is not documented, or think it should be undone, please open an issue on [GitHub](https://github.com/pybamm-team/pybamm).
```

brosaplanella: Yes, thanks! Good thing I am not a lecturer in mathematics 😆

Saransh-cpp (approved):
Thanks!

PR #3182, opened 2023-07-23
Enable downloadable PDF, EPUB, and zipped HTML formats on Read the Docs

agriyakhetarpal:
# Description

According to the [documentation](https://docs.readthedocs.io/en/stable/guides/enable-offline-formats.html). This PR will build PyBaMM's documentation in more formats and make it available for download for subsequent versions.

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -11,6 +11,11 @@ sphinx:
   configuration: docs/conf.py
   fail_on_warning: false
 
+formats:
+  - pdf
+  - epub
+  - htmlzip
+
 # Set the version of Python and other tools you might need
 build:
   # Install Graphviz to build SVG files


agriyakhetarpal:
@tinosulzer my guess is that somewhere between our 0.2.3 and 0.3.0 release the automatic builds for the documentation downloads were disabled by default by the Read the Docs configuration, which explains why never got these EPUB and PDF formats afterwards. I do not think we can build downloads for previous releases, but this will ensure that versions 23.5, 23.9, and further will have their downloads on https://readthedocs.org/projects/pybamm/downloads/. You, @martinjrobins, and I are project administrators on Read the Docs so we have the permissions to modify the downloads if needed.

tinosulzer (approved):

PR #3205, opened 2023-07-30
#2368 overload numpy functions

### LINKED ISSUE #2368: Overload numpy functions
There is a way to write `np.exp(symbol)` (and other functions) to return what is currently returned by `pybamm.exp(symbol)`. This would be useful to implement so that functions can be defined using np instead of pybamm

https://towardsdatascience.com/wrapping-numpys-arrays-971e015e14bb

tinosulzer:
# Description

- Numpy functions can now be called on pybamm symbolic objects. This will return a pybamm symbolic object applying the same function (e.g. `np.exp(x)` returns `pybamm.Exp(x)`). Parameter sets in the library have been changed to use numpy functions, but they do not need to be (the old style of calling pybamm functions will still work)
- Binary operators applied to a pybamm object and numpy array return a pybamm binary operator, where previously it raised an error. The numpy array gets converted to a pybamm Vector. This only works with 1D arrays (it could work with other objects, but restricting it to 1D arrays makes things easier for now

Fixes #2368 

## Type of change

# Key checklist:

## Further checks:


examples/scripts/compare_lithium_ion.py
```diff
@@ -16,7 +16,7 @@
 # create and run simulations
 sims = []
 for model in models:
-    sim = pybamm.Simulation(model)
+    sim = pybamm.Simulation(model, parameter_values=pybamm.ParameterValues("Chen2020"))
```
brosaplanella, line 19: Do we need to use Chen2020 for a specific reason or was this for developing reasons only? If the latter, I would vote for using default parameters.

brosaplanella (approved):
Looks good, thanks! Just a very small comment (you will have to update the CHANGELOG too).

PR #3217, opened 2023-08-01
Idaklu solver can be given a list of variables to calculate during the solve

### LINKED ISSUE #2796: New solver mode: only including required variables (don't store entire `y`)
### Description

Provide a mode for the solver so that only a user given list of variables are calculated and stored as part of the solution

### Motivation

Remove the need to store the entire `y` solution and store only the variables given by the user

### Possible Implementation

Interface might look like the following:

```python
sol = solver.solve(model, t_eval, solution_include=['Voltage [V]'])
print(sol['Voltage [V]'].entries)

sol['Negative particle surface concentration']  # gives an exception
```

### Additional context

_No response_

jsbrittain:
# Description

Provides a mode for the (Idaklu) solver so that only a user given list of variables are calculated and stored as part of the solution. This functionality extends to sensitivity calculations. The PR also includes a restructuring of the idaklu c++ codebase to support further extensions and abstractions in the future.

Fixes #2796 

## Type of change

# Key checklist:

- [X] No style issues: `$ pre-commit run` (see [CONTRIBUTING.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CONTRIBUTING.md#installing-and-using-pre-commit) for how to set this up to run automatically when committing locally, in just two lines of code)
- [X] All tests pass: `$ python run-tests.py --all`
- [X] The documentation builds: `$ python run-tests.py --doctest`

## Further checks:


jsbrittain:
Interface is as suggested in #2796 , namely:

```python
sol = solver.solve(model, t_eval, output_variables=['Voltage [V]'])
print(sol['Voltage [V]'].entries)
```

Note that only variables listed in `output_variables` can be queried since the full model state vector is no longer propagated and returned.

jsbrittain:
@martinjrobins Ready for review. Some brief comments on the checks:
1) Benchmarks don't appear to be running, and this seems to be widespread across other PRs. I recommend waiting until these work before merging, but I think the code can still be reviewed in the meantime.
2) Codacy - seems to be reporting a false-positive that I can't get rid of, despite trying to initalise the variable in 3 different ways(!)
3) Doctests - I don't know why these are failing as I haven't touched any code related to them.

pybamm/solvers/c_solvers/idaklu.cpp
```diff
@@ -17,6 +18,7 @@ Function generate_function(const std::string &data)
 namespace py = pybind11;
 
 PYBIND11_MAKE_OPAQUE(std::vector<np_array>);
+//PYBIND11_MAKE_OPAQUE(std::vector<Function>);
```
martinjrobins, line 21: do we need this commented out code anymore?

pybamm/solvers/c_solvers/idaklu/CasadiSolverOpenMP.cpp
```diff
@@ -0,0 +1,550 @@
+#include "CasadiSolverOpenMP.hpp"
+#include "casadi_sundials_functions.hpp"
+#include <casadi/casadi.hpp>
+#include <casadi/core/function.hpp>
+#include <casadi/core/sparsity.hpp>
+
+/*
+ * This is an abstract class that implements an OpenMP solution but
```
martinjrobins, line 8: can we please move the documentation into the header files please? Ideally in doxygen format, but that is not neccessary at this point since we don't generate any api docs

pybamm/solvers/c_solvers/idaklu/CasadiSolverOpenMP_solvers.hpp
```diff
@@ -0,0 +1,231 @@
+#ifndef PYBAMM_IDAKLU_CASADI_SOLVER_OPENMP_HPP
+#define PYBAMM_IDAKLU_CASADI_SOLVER_OPENMP_HPP
+
+#include "CasadiSolverOpenMP.hpp"
+#include "casadi_solver.hpp"
+
+class CasadiSolverOpenMP_Dense : public CasadiSolverOpenMP {
```
martinjrobins, line 7: Looks good, this looks a lot better than my mess of if-statements :)

martinjrobins: could probably generate all these classes with macros, but the effort might might not the worth the reduced # lines.

pybamm/solvers/c_solvers/idaklu/casadi_functions.hpp
```diff
@@ -3,56 +3,106 @@
...
+  void operator()(std::vector<realtype*> inputs,
+                  std::vector<realtype*> results);
 
 public:
   std::vector<const double *> m_arg;
   std::vector<double *> m_res;
-  void operator()();
 
-private:
+//private:
```
martinjrobins, line 52: should this be uncommented?

pybamm/solvers/idaklu_solver.py
```diff
@@ -251,6 +275,49 @@ def resfn(t, y, inputs, ydot):
                 "mass_action", [v_casadi], [casadi.densify(mass_matrix @ v_casadi)]
             )
 
+            # if output_variables specified then convert functions to casadi
+            # expressions for evaluation within the idaklu solver
+            self.var_casadi_fcns = {}
```
martinjrobins, line 280: all the expressions are typically converted to casadi function within the base solver (in `set_up`). Probably best to do this there so that other solvers can use them

pybamm/solvers/idaklu_solver.py
```diff
@@ -251,6 +275,49 @@ def resfn(t, y, inputs, ydot):
...
+            self.var_casadi_idaklu_fcns = []
+            self.dvar_dy_fcns = []
+            self.dvar_dp_fcns = []
+            for key in self.output_variables:
+                # variable functions
+                if isinstance(
+                    model.variables_and_events[key], pybamm.ExplicitTimeIntegral
+                ):
+                    continue
```
martinjrobins, line 289: is ExplicitTimeIntegral not supported. Should we throw an error here?

pybamm/solvers/idaklu_solver.py
```diff
@@ -589,7 +668,35 @@ def _integrate(self, model, t_eval, inputs_dict=None):
...
-            return sol
+            newsol.integration_time = integration_time
+            if self.output_variables:
+                # Populate variables and sensititivies dictionaries directly
+                number_of_samples = sol.y.shape[0] // number_of_timesteps
+                sol.y = sol.y.reshape((number_of_timesteps, number_of_samples))
+                startk = 0
+                for vark, var in enumerate(self.output_variables):
+                    if isinstance(
+                        model.variables_and_events[var], pybamm.ExplicitTimeIntegral
```
martinjrobins, line 679: and/or here...

pybamm/solvers/processed_variable_var.py
```diff
@@ -0,0 +1,437 @@
...
+# Processed Variable class
+#
+import casadi
+import numpy as np
+import pybamm
+from scipy.integrate import cumulative_trapezoid
+import xarray as xr
+
+
+class ProcessedVariableVar(object):
```
martinjrobins, line 11: I think we need to explain the differences to `ProcessedVariable` in the docstring. Also, I'm unsure what the last `Var` in `ProcessedVariableVar` means? `Variant`?

martinjrobins (changes requested):
Thanks @jsbrittain, this looks great. Few suggestions below. I think the main one that will take work is to move the generation of the variable casadi functions into the base solver (set_up) so that other solvers can take advantage of this as well (e.g. we want to do the same thing for the jax solver)

pybamm/solvers/c_solvers/idaklu.cpp
jsbrittain: removed

pybamm/solvers/c_solvers/idaklu/casadi_functions.hpp
jsbrittain: m_func is accessed for nnz and sparsity values outside of the class. i have added the private declaration back in for the other variables but made m_func public rather than provide dedicated getters (unless that approach is strongly preferred?).

pybamm/solvers/c_solvers/idaklu/CasadiSolverOpenMP.cpp
jsbrittain: 👍

pybamm/solvers/c_solvers/idaklu/CasadiSolverOpenMP_solvers.hpp
jsbrittain: I wasn't entirely happy with the repetition here either. The macro approach worked nicely and has been implemented now, thanks.

pybamm/solvers/idaklu_solver.py
jsbrittain: ExplicitTimeIntegral receives special treatment throughout the codebase. There is no action required here to convert this type of variable. This is confirmed by comparing the standard and new-solver outputs for these variables, which form part of the unit test cases in test_solvers/test_idaklu_solver.py#584-585.

pybamm/solvers/idaklu_solver.py
jsbrittain: (as above)

pybamm/solvers/processed_variable_var.py
jsbrittain: The class has now been renamed to `ProcessedVariableComputed` and the docstring updated to specify the purpose of the variant class.

pybamm/solvers/idaklu_solver.py
jsbrittain: I've now moved the generation of casadi variable (and sensitivity) functions to within the base_solver (with some renaming for clarity), while retaining the idaklu conversion code within the idaklu_solver file.

pybamm/solvers/base_solver.py
```diff
@@ -1585,6 +1648,14 @@ def jacp(*args, **kwargs):
                 [t_casadi, y_and_S, p_casadi_stacked, v],
                 [jac_action_casadi],
             )
+            # Compute derivate wrt p-stacked (can be passed to solver to
+            # compute sensitivities online)
+            if return_jacp_stacked:
```
martinjrobins, line 1653: should this go under the `elif model.calculate_sensitivities:` (line 1621)?

jsbrittain: Ah, yes. I've moved the code to a more appropriate place.

martinjrobins:
Thanks @jsbrittain. The doctests and notebooks, and the example scripts don't seem anything to do with this PR, but there are a few issues to fix in the Cadacy Static Code Analysis

jsbrittain:
> Thanks @jsbrittain. The doctests and notebooks, and the example scripts don't seem anything to do with this PR, but there are a few issues to fix in the Cadacy Static Code Analysis

@martinjrobins yes, I believe the doctest and notebooks are due to matplotlib deprecating a feature over the weekend - there are other PRs in progress to resolve this. As for codacy, I was looking at the report and the (big!) jump in codacy issues appears to be due to the macro function that was introduced in CasadiSolverOpenMP_solvers.hpp to help automate creation of classes for the various solvers. Codacy doesn't appear to expand the expressions when evaluating them and so believes that there are now many dangling properties. The only way I can think of adequately resolving this would be to revert the change and remove the macro. Do you have a strong preference here?

brosaplanella:
> @martinjrobins yes, I believe the doctest and notebooks are due to matplotlib deprecating a feature over the weekend - there are other PRs in progress to resolve this.

The fixes are now merged in the develop branch

martinjrobins:
> . The only way I can think of adequately resolving this would be to revert the change and remove the macro. Do you have a strong preference here?

I would be in favor of keeping the macro, can you turn off the error in some way?

brosaplanella:
Seems like Codacy got stuck. Is it an issue at their end or at our end

jsbrittain:
> Seems like Codacy got stuck. Is it an issue at their end or at our end?

@brosaplanella Not sure why codacy stalled. I've been making some changes and commited again so will keep an eye on it.

jsbrittain:
@martinjrobins Two main changes since the last review:
1. I've replaced the macro with a variadic arguments approach. This keeps the boilerplate down, but should also make the code easier for linters to follow and to debug going forwards.
2. Despite the above, codacy still has trouble keeping track of some of the parent class variables passed in this way, so I've added `cppcheck-suppress` statements on these lines (since this is what codacy [currently] uses) and provided comments in the code to justify them.

martinjrobins (approved):
fantastic, thanks @jsbrittain. All looks good to merge

PR #3222, opened 2023-08-02
Update publish workflow

### LINKED ISSUE #2881: Reduce release frequency

brosaplanella:
# Description

Publish workflow is now triggered by Github release (and pre-releases). 

Fixes #2881 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/.github/workflows/publish_pypi.yml
+++ b/.github/workflows/publish_pypi.yml
@@ -1,8 +1,8 @@
 name: Build and publish package to PyPI
 
 on:
-  push:
-    branches: main
+  release:
+    types: [published]
   workflow_dispatch:
     inputs:
       target:


Saransh-cpp (commented):
For your question, we should add the following below every `actions/checkout@v3` line -

```yaml
  with:
    # The branch, tag or SHA to checkout. When checking out the repository that
    # triggered a workflow, this defaults to the reference or SHA for that event.
    # Otherwise, uses the default branch.
    ref: 'main'
```

This would require merging develop into main before making a release, similar to how it is done right now.

I might have also stumbled upon a bug in our release workflow. See how none of the `actions/checkout` calls specify the `main` branch right now -

https://github.com/pybamm-team/PyBaMM/blob/d3c6bc6eb3ab5f0ee8dfa34ad910810079568380/.github/workflows/publish_pypi.yml#L22

which means that we have been building releases on the `develop` branch and publishing them, whereas we intended to make the releases on the `main` branch (after merging in `develop`).

Tags are actually independent of branches, but they point to the latest commit of the branch that they were created on, which means when a tag is created in `pybamm`, it points to the latest commit of the `develop` branch, rather than the latest commit of the `main` branch.

Finally, the actual releases are also being made through the `develop` branch - 

https://github.com/pybamm-team/PyBaMM/blob/d3c6bc6eb3ab5f0ee8dfa34ad910810079568380/.github/workflows/create_release.yml#L46-L50

Note how `main` isn't specified anywhere, and this action picks up the default branch if no branch is specified.

For instance, see how the [main branch](https://github.com/pybamm-team/PyBaMM/commits/main) says that `v23.5` should have the time based experiment change as one of the last changes. But, the actual [`v23.5` tag](https://github.com/pybamm-team/PyBaMM/commits/v23.5?before=5e28bb9a316f394282b40c8a56b285e85d987b77+35&branch=v23.5&qualified_name=refs%2Ftags%2Fv23.5) has a ton of other changes after the time based experiment one.

This went unnoticed because usually `develop` and `main` were perfectly synced together before the releases, but introducing `rc` versions allowed the develop branch to move ahead of the main branch. It looks like this is a serious and a big issue, but given that no user has pointed it out, we should be safe 🙂

brosaplanella:
I see. So how should we proceed going forward when making a release? If I understand you correctly, tags do not point to specific commit, so not sure how can we resolve that when making a release.

The publishing workflow should work for pre-releases too, so it should take the `vYY.MM` branch. For the final release we could either use `main` or `vYY.MM`. 

BTW, you know workflows much better than I do, so feel free to continue this if you want/have time. Otherwise I will continue working on it once we clarify this.

Saransh-cpp:
On my radar for next week!

Saransh-cpp:
I wanted to test some things on my fork so I ended up doing a lot of work there 🙂

I have created #3249 for this. The PR seems to work, but I will still test some edge cases and then mark it ready for review.

brosaplanella:
Awesome, will close this PR then. Thanks @Saransh-cpp!

PR #3231, opened 2023-08-03
Add `pip` caches, remove `nox` caches, and remove `requirements.txt`

### LINKED ISSUE #3184: CI cache improvements (`nox`, `pip` and more)
## Description

After or alongside #3048, we can add some improvements to the `nox` caching to make it more reliable—I noticed that there is a lot of bloat with the [generated caches](https://github.com/pybamm-team/PyBaMM/actions/caches) sometimes which leads to the weekly cache limit (10 GB) getting exceeded on many occasions. I am choosing to open this as a separate issue for the sake of clarity. The caching mostly works on consequent commits on a PR branch, but new caches continue to be built across different PR branches, which should not be the case (Lychee's caches from even days ago are restored in almost all workflow runs).

Here are some ideas I had:

## Tasks

- ~[ ] Use some general `restore-keys` in case cache hits do not occur on keys. We already use the output of the setup-python step to ensure that there are no broken symlinks and that the Python version in the environments matches that of the GitHub Actions runners. This way we can stay beneath the size limit offered by GitHub runners.~
- [x] `noxfile.py` is used as a hash file for most environments' caches, but it is not required as one: it is not included in any of the cache locations nor is it its purpose to track dependencies and their versions (we use `setup.py` or `requirements.txt`).
- [x] We know [where `pip` stores its caches](https://pip.pypa.io/en/stable/topics/caching/) on various platforms. Although downloaded wheels are reused between `nox` environments since they all run after one another, we can cache these locations between builds too for a minor speedup.

Please feel free to add more to this list.

agriyakhetarpal:
# Description

This PR adds restore keys for `nox` environments' caches. Also removes `requirements.txt` since it serves no purpose now.
The `pip` caches were added and removed because the cache size is small and would not have any substantial effect on reducing CI times.

Fixes #3184

## Type of change

# Key checklist:

## Further checks:


agriyakhetarpal:
The only thing that stumps me is: now that we have `restore-keys` with this PR—shall some dependency have a newer release in some build in the future, will `session.install` fetch the latest version for it? If a cache hit is not found, it would fall back to using one of the restore keys and thereby use the older version of the dependency and return `Requirement already satisfied` for said dependency. This could be _not_ what we want in some scenarios (say, we want a newer version for the dependency due to deprecations in the earlier version that are incompatible with other versions). Otherwise, this is ready for review

Saransh-cpp:
> shall some dependency have a newer release in some build in the future, will session.install fetch the latest version for it?

If in the future we require a newer version of a dependency, we will most likely create a lower bound pin for that (given that the older version would be breaking pybamm), and then the workflows won't be able to find any sort of cache for that. So, I think we should be safe here.

.github/workflows/test_on_push.yml
```diff
@@ -122,7 +121,8 @@ jobs:
         if: (matrix.os == 'ubuntu-latest' && matrix.python-version != 3.11) || (matrix.os != 'ubuntu-latest')
         with:
           path: ${{ github.workspace }}/.nox/unit/
-          key: ${{ runner.os }}-nox-unit-${{ steps.setup-python.outputs.python-version }}-${{ hashFiles('**/noxfile.py', '**/setup.py') }}
+          key: ${{ runner.os }}-nox-unit-${{ steps.setup-python.outputs.python-version }}-${{ hashFiles('**/setup.py') }}
+          restore-keys: ${{ runner.os }}-nox-unit-${{ steps.setup-python.outputs.python-version }}-
```
Saransh-cpp, line 125: How are the restore-keys different from keys? Or how is this improving/helping the cache system?

agriyakhetarpal: In cases wherein no cache hits occur on a specific key hash (which has been the case a lot lately), it falls back to using of the listed `restore-keys` sequentially to restore files from. Sort of a _partial_ match. Here we only use one restore key because the Python version should always be in it

agriyakhetarpal: Also, would the `pip` cache work if we were to install "-e .[all]" in the `setup-python` step? (I think it would if we install standard Python dependencies and PyBaMM-specific onew in that step itself, see the end of [this section](https://github.com/actions/setup-python/blob/v4/docs/advanced-usage.md#caching-packages). That way all `nox` environments can benefit from it as well

Saransh-cpp: I think it should work.

Saransh-cpp (approved):
Thanks, @agriyakhetarpal! Is this ready to merge?

agriyakhetarpal:
Almost, I shall make the changes for and try out `pip` caching. If we install all dependencies in the CI globally instead of in `nox` environments, we might be able to skip the installation step in all of them (it will work only in the CI though) through a combination of `--no-venv` and `--no-install`. Their equivalent is `nox.options.default_venv_backend = None` but there isn't any for the latter so `noxfile.py` may have some annoying if-elses if we do this. Even if we don't skip the installation step, it should still make things fast due to `pip`'s caching (which will always work)

agriyakhetarpal:
This should be ready to merge now. The [pip caches](https://github.com/pybamm-team/PyBaMM/actions/runs/5769514914/job/15641624070?pr=3231#step:7:21) provided by `setup-python` are working and are more reliable. All Pythonic dependencies will now be installed globally (in the CI) instead of in `nox` environments, which makes the `nox` caches redundant—since `nox` can use the cached wheel files by `pip` and execute the installation steps in a matter of seconds whenever a new session is created. We can skip installations for `nox` sessions using the `--no-install` flag as well.

Note: the `session.install` commands do not work outside of virtual environments, so I switched to using `session.run` and `session.run_always` instead. I think this is a good move because they will display a list of installed dependencies and their versions, which also fixes the issue of `nox` not displaying verbose outputs by default that we had encountered earlier. The difference between `session.run` and `session.run_always` is that the latter runs even in the presence of the `--install-only` flag.

agriyakhetarpal:
With a global installation in place, we might be in a position to try out [this blog](https://blog.allenai.org/python-caching-in-github-actions-e9452698e98d) again to cache entire Python environments, including the installed packages

agriyakhetarpal:
Okay, so it fails to find IDAKLU components when installed in a global installation. I'm looking into how this can be fixed

.github/workflows/test_on_push.yml
```diff
@@ -45,12 +45,10 @@ jobs:
...
     steps:
-      - uses: actions/checkout@v3
-      - id: setup-python
-        name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v4
+      - name: Check out PyBaMM repository
+        uses: actions/checkout@v3
         with:
-          python-version: ${{ matrix.python-version }}
+          fetch-depth: 0
```
Saransh-cpp, line 51: Nitpicking here, but is this required? This would fetch the entire git history, slowing down this step maybe by a few seconds. Is the cache stored using git?

Saransh-cpp (approved):
Thia looks amazing! Thanks for picking this up, @agriyakhetarpal! :)

.github/workflows/test_on_push.yml
agriyakhetarpal: Not really, was just experimenting – I forgot to remove this 😄

Saransh-cpp: We can try getting rid of it!

agriyakhetarpal: Done!

jeromtom:
Hey @agriyakhetarpal and @Saransh-cpp, I think this merged PR is causing [a build error](https://github.com/pybamm-team/PyBaMM/actions/runs/5765926288/job/15634909588?pr=3226#step:12:462:~:text=ERROR%3A%20THESE%20PACKAGES,exit%20code%201.) and a possible solution shown is updating the hashes.

agriyakhetarpal:
That's not possible because you haven't merged the latest changes from the `develop` branch yet in your PR branch, so these changes are not included. Also, we may safely ignore that error

jeromtom:
Okay, thank you.

PR #3234, opened 2023-08-04
Processed variables now get spatial variables automatically

### LINKED ISSUE #3200: Enable 2D processed variables for arbitrary domains
### Description

At the moment, 2D processed variables use a series of `if` statements to check the domains on which to plot things, which end throwing an error if none of the cases is recognised.

https://github.com/pybamm-team/PyBaMM/blob/695917e345dc28af65d7654e66d07aedb9494d93/pybamm/solvers/processed_variable.py#L288-L329

We should enable a generic case where, if none of the above, the model is plotted taking the secondary domain as horizontal axis and the primary domain as vertical axis (to match the batteries case particle-electrode). Potentially, we could let the user modify this.

### Motivation

This would enable running non-battery models in PyBaMM and plotting them.

### Possible Implementation

Change current `else` statement to assign `self.first_dimension` and `self.second_dimension` from primary and secondary domain variables. It seems `r_sol`, `x_sol` et al are only used in testing.

### Additional context

_No response_

brosaplanella:
# Description

Processed variables now automatically take the names of the domains to plot automatically from spatial variables. This was to enable the plotting of 2D variables in arbitrary domains. Note that it required adding a `geometry` variable to the model as at the moment there is no way to get the geometry from a processed model, even though geometry is a key part of the model. Probably I didn't do it in the most efficient way so suggestions are welcome.

Fixes #3200 

## Type of change

# Key checklist:

## Further checks:


pybamm/models/base_model.py
```diff
@@ -312,6 +313,14 @@ def length_scales(self, values):
             "length_scales has been removed since models are now dimensional"
         )
 
+    @property
+    def geometry(self):
+        return self._geometry
+
+    @geometry.setter
+    def geometry(self, geometry):
+        self._geometry = geometry
```
tinosulzer, line 322: I don't think this should have a public setter class

tinosulzer (changes requested):
Looks good, just one comment about the public/private API

tinosulzer (approved):

PR #3237, opened 2023-08-04
Allow for cylindrical and spherical shell domains

### LINKED ISSUE #3115: Allow for cylindrical & spherical shell domains
### Description

At the moment we cannot simulate spherical shells with a boundary condition at the inner boundary other than no flux, as PyBaMM assumes the domain is always `[0, R]` and thus throws an error, see
https://github.com/pybamm-team/PyBaMM/blob/fffe650253596f14b6918d9e57fdd912bc3cfbdc/pybamm/discretisations/discretisation.py#L457-L464

Apparently it works for cylindrical polar (the error does not catch that case) but it makes me suspect that this feature is not well tested, so it would be good to also add some easy tests against analytical solutions to ensure this functionality works correctly.

### Motivation

Current models do not make use of this feature, but one can imagine how it could be relevant (e.g. core shell models).

In any case, for completeness of the functionality it would be good to enable this anyway.

### Possible Implementation

_No response_

### Additional context

_No response_

brosaplanella:
# Description

Now we only throw and error for non zero-Neumann bcs if the boundary is at `r = 0`, which means cylindrical and spherical shells can be solved with any boundary condition. Also added some integration tests to check it works.

Fixes #3115

## Type of change

# Key checklist:

## Further checks:


brosaplanella:
This is now ready to review

tests/integration/test_spatial_methods/test_finite_volume.py
```diff
@@ -294,6 +294,51 @@ def get_error(m):
...
+    model.initial_conditions = {u: pybamm.Scalar(0)}
+    model.variables = {"u": u, "r": r}
+    geometry = {"domain": {r: {"min": pybamm.Scalar(1), "max": pybamm.Scalar(2)}}}
+    submesh_types = {"domain": pybamm.Uniform1DSubMesh}
+    var_pts = {r: 500}
+    mesh = pybamm.Mesh(geometry, submesh_types, var_pts)
+    spatial_methods = {"domain": pybamm.FiniteVolume()}
+    disc = pybamm.Discretisation(mesh, spatial_methods)
+    disc.process_model(model)
+    solver = pybamm.CasadiAlgebraicSolver()
```
tinosulzer, line 318: It's a bit strange to have a solver inside this test. Not the end of the world but would be better without, so that this is really a unit test (and not dependent on the solver). Is there another way to test the same thing?

brosaplanella:
It's an integration test though, isn't it fine to have a solver there?

tinosulzer:
Ah yes true, then it's fine

tinosulzer (approved):

Issue #3225, opened 2023-08-02
[Bug]: create_from_bpx(): reaction rate activation energies are always zero


ejfdickinson:
### PyBaMM Version

23.4

### Python Version

3.9.16

### Issue

The effective values of reaction rate constant activation energies specified in a BPX JSON file are always zero after processing by `create_from_bpx()`.

See also #3051 which reports other out-of-specification behaviour of the same function.

### Cause:

This is because BPX specifies that these have the name `Reaction rate constant activation energy [J.mol-1]` which is rendered by `_bpx_to_domain_param_dict()` into

```
`Negative electrode reaction rate constant activation energy [J.mol-1]`
`Positive electrode reaction rate constant activation energy [J.mol-1]`
```

However, ` _bpx_to_param_dict()` checks as follows when recreating exchange current density functions:

```
E_a_n = pybamm_dict.get(
    negative_electrode.pre_name + "reaction rate activation energy [J.mol-1]", 0.0
)
[...]
E_a_p = pybamm_dict.get(
    positive_electrode.pre_name + "reaction rate activation energy [J.mol-1]", 0.0
)
```

Because the word `constant` is missing from the lookup, imported reaction rate constant activation energies are always replaced with zero in the resulting exchange current density function in the `ParameterValues` object.

### Proposed fix:

Either:

- Correct the name lookups in ` _bpx_to_param_dict()`, **or**
- Delete the substring `"constant "` when processing these fields in `_bpx_to_domain_param_dict()`.

### Steps to Reproduce

Using `nmc_pouch_cell_BPX.json` from the BPX example download:

```
import pybamm

from pybamm import constants, exp, log

# From BPX example download
params = pybamm.ParameterValues.create_from_bpx("nmc_pouch_cell_BPX.json")

cs_max = params["Maximum concentration in positive electrode [mol.m-3]"]
cl_ref = params["Initial concentration in electrolyte [mol.m-3]"]

Ea_kin_pos = params["Positive electrode reaction rate constant activation energy [J.mol-1]"]

# Evaluate exchange current density at Tref
T_ref  = params["Reference temperature [K]"]
j0_pos_ref = params["Positive electrode exchange-current density [A.m-2]"](cl_ref, cs_max/2, cs_max, T_ref)

# Evaluate exchange current density at Ttest =/= Tref
T_test       = 290
j0_pos_test = params["Positive electrode exchange-current density [A.m-2]"](cl_ref, cs_max/2, cs_max, T_test)

[...]

Issue #3051, opened 2023-06-19
[Bug]: create_from_bpx(): effective values of diffusivity activation energies are overwritten by "electrolyte conductivity activation energy"


ejfdickinson:
### PyBaMM Version

23.4

### Python Version

3.9.16

### Describe the bug

When temperature-dependent functions are created in `create_from_bpx()`, the apparent activation energy for the following callable parameter definitions in the generated `parameter_values` instance is replaced by the value of `"Electrolyte conductivity activation energy [J.mol-1]"`.

This appears to be a late-binding bug due to the repeat use of non-local variable `E_a` in the declaration these functions, which is repeatedly redefined, the electrolyte conductivity activation energy being the last such redefinition.

Affected functions:

- "Negative electrode diffusivity [m2.s-1]"
- "Positive electrode diffusivity [m2.s-1]"
- "Electrolyte diffusivity [m2.s-1]"

Note: identified in 23.4 but no relevant code changes in latest version

### Steps to Reproduce

Using `nmc_pouch_cell_BPX.json` from the BPX example download:

```
import pybamm
from pybamm import constants, exp, log

# From BPX example download
params = pybamm.ParameterValues.create_from_bpx("nmc_pouch_cell_BPX.json")

Ea_cond_el   = params["Electrolyte conductivity activation energy [J.mol-1]"]
Ea_diff_neg  = params["Negative electrode diffusivity activation energy [J.mol-1]"]

# Evaluate negative electrode diffusivity at Tref
T_ref        = params["Reference temperature [K]"]
Ds_neg_ref   = params["Negative electrode diffusivity [m2.s-1]"](0.5, T_ref)

# Evaluate negative electrode diffusivity at Ttest =/= Tref
T_test       = 290
Ds_neg_test  = params["Negative electrode diffusivity [m2.s-1]"](0.5, T_test)

# Evaluate expected negative electrode diffusivity at Ttest
D_neg_test_expected = Ds_neg_ref * exp(-Ea_diff_neg / constants.R * (1/T_test - 1/T_ref))

# Evaluate apparent activation energy
Ea_diff_neg_apparent = -constants.R / (1/T_test - 1/T_ref) * log(Ds_neg_test/Ds_neg_ref)

print("Negative electrode diffusivity (test temperature, expected):", D_neg_test_expected)
print("Negative electrode diffusivity (test temperature, actual):", Ds_neg_test)

[...]

PR #3242, opened 2023-08-07
Issue 3051 - bpx activation energies

darryl-ad:
# Description

Fixes a bug where `E_a` was used as the alias for multiple different activation energies in `bpx`, causing them to all take the final value. Also fixes bug where `reaction rate constant activation energy` was always set as 0.
- Assigned unique variable names
- Edited naming convention for better clarity
- Updated variable name `reaction rate activation energy` -> `reaction rate constant activation energy`  in `bpx` parameter search.

Fixes # [3051](https://github.com/pybamm-team/PyBaMM/issues/3051) and # [3225](https://github.com/pybamm-team/PyBaMM/issues/3225)

## Type of change

# Key checklist:

## Further checks:


darryl-ad:
Added an additional commit to fix https://github.com/pybamm-team/PyBaMM/issues/3225 here. I was concerned separate PRs would cause a merge conflict.

tinosulzer (changes requested):
Looks good, can you add tests for coverage? 

You could also reduce code duplication by defining the arrhenius function up front and just calling it everywhere e.g.
```python
def arrhenius(Ea, T):
     return exp(Ea / constants.R * (1 / T_ref - 1 / T))
```
and later 
```python
return arrhenius(Ea_D_e, T) * ...
```

darryl-ad:
- refactored code as suggested
- added a simple unit test for coverage, and to validate arrhenius calculations

I've not made a test before, so please double check it's ok!

See separate discussion here https://github.com/pybamm-team/PyBaMM/issues/3225 about full testing requirements for BPX.

brosaplanella:
The test looks good to me. The issues with the coverage seem to be that for various parameters it never reaches the `else` statement. The two previous `if` statements check for callables and data, so I guess the `else` should be for constant values, but I am not sure why `test_constant_functions` is not covering those lines.

darryl-ad:
I agree, I'm not sure why `test_constant_functions` no longer covers those lines

brosaplanella:
Can you run it in debug mode and check why the constant case does not reach the else?

darryl-ad:
Sorry not sure what you mean by debug mode. However I tested it with print statements and the constant case does reach the else

brosaplanella:
I think the issue is that in `test_constant_functions` the parameter values are created from BPX but then not used. Therefore, the header of the function is covered (as the tests goes inside the if statement) but the return line doesn't as it is not get called. I think that if you add some additional tests after line 163 which check that the values are as expected it should be fine.

darryl-ad:
@brosaplanella I've added a test that I think addresses your previous comment, thanks!

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 - Error generated when invalid parameter values are passed. ([#3132](https://github.com/pybamm-team/PyBaMM/pull/3132))
 - Thevenin() model is now constructed with standard variables: `Time [s], Time [min], Time [h]` ([#3143](https://github.com/pybamm-team/PyBaMM/pull/3143))
 - Fix SEI Example Notebook ([#3166](https://github.com/pybamm-team/PyBaMM/pull/3166))
+- Fixed bug causing incorrect activation energies using `create_from_bpx()` ([#3242](https://github.com/pybamm-team/PyBaMM/pull/3242))
```
brosaplanella, line 16: Can you move this up so they PRs are in decreasing order?

brosaplanella (approved):
Looks good to me! I couldn't tell why the doc tests are failing though, @agriyakhetarpal do you have any insight?

CHANGELOG.md
darryl-ad: done!

agriyakhetarpal:
There is a reST syntax warning/error in the installation instructions for WSL. I am adding a link to the erroneous lines because I cannot suggest changes directly here:

https://github.com/pybamm-team/PyBaMM/blob/develop/docs/source/user_guide/installation/windows-wsl.rst?plain=1#L24-L36

Removing the extra colons in points 3 and 4 will fix it

brosaplanella:
Thanks Agriya! @darryl-ad, can you resolve the conflict with CHANGELOG so tests run, and do the small fix Agriya suggested?

darryl-ad:
@brosaplanella I've resolved the CHANGELOG conflict. All tests are passing on my end now. The erroneous lines in the docs that Agriya suggested aren't present in this branch - do I need to create an issue and branch for it?

tinosulzer (approved):

CHANGELOG.md
```diff
@@ -8,8 +8,8 @@
 - Implement the MSMR model ([#3116](https://github.com/pybamm-team/PyBaMM/pull/3116))
 
 ## Bug fixes
-
 - Fixed bug causing incorrect activation energies using `create_from_bpx()` ([#3242](https://github.com/pybamm-team/PyBaMM/pull/3242))
```
brosaplanella, line 11: Can you move this line down so the PRs are sorted in decreasing order? Happy to merge after that

darryl-ad: Done!

PR #3253, opened 2023-08-08
Current-driven LAM

### LINKED ISSUE #3252: Implement current depedent LAM
### Description

Add a current dependent LAM model, like the one in eq [26] of [Reniers et al (2019)](https://iopscience.iop.org/article/10.1149/2.0281914jes/pdf)
image

### Motivation

This is another fairly standard LAM model which would be very easy to implement in the framework.

### Possible Implementation

I think for this one it would be good to keep it general and define the RHS of the equation from a parameter function depending on current density and temperature. The user could define the function similarly to what we do with exchange current density so other models can be easily tested.

### Additional context

_No response_

brosaplanella:
# Description

Implement current-driven LAM submodel. At the moment it cannot be combined with the other ones, but that's because I wasn't sure how we should set the options (i.e. single string, list of strings...). Happy to leave it as is for the time being or to implement a fix if we agree on a format.

Fixes #3252 

## Type of change

# Key checklist:

## Further checks:


tinosulzer:
Can you add an integration test as well?

brosaplanella:
> Can you add an integration test as well?

For the integration tests I need parameter values, but I don't have any (need to check the Safari & Delacourt paper, but not sure they will translate). Shall I just some dummy parameters (i.e. made up) and define them in place for the integration tests? I want to update the notebook anyway, so I will come up with some parameters for that too.

tinosulzer:
Yes dummy parameters are fine

brosaplanella:
Requesting also @DrSOKane review as I made some changes to the LAM submodels notebook and I believe he was the one who wrote it.

DrSOKane (changes requested):
This new LAM option implements the empirical model of Delacourt and Safari (2012) in PyBaMM. As it is a purely empirical model, I don't see much point in making it compatible with the other LAM models. However, in both the example and the tests, only the linear term of their equation ([26] in Reniers, Mulder and Howey 2019) is included and the nonlinear term is left out. Can you try including the nonlinear term and seeing what happens?

DrSOKane (approved):

tinosulzer:
Question for @brosaplanella and @DrSOKane : should the minus sign for the current-driven LAM term be in the model or in the parameter?

brosaplanella:
I agree, but thought it would be better to stick to the notation in Reniers et al

PR #3261, opened 2023-08-10
Improve memory use in experiments

### LINKED ISSUE #3081: Improve memory usage for experiments with `start_time`
At the moment, two steps with the same operating conditions but different `start_time` are treated as different. This means that the model needs to be processed twice, which can end up with a big memory consumption (and longer running times) for very long experiments.

A way to improve this would be to read the `start_time` from the unprocessed step, this way we can have a single processed step for all of them.

brosaplanella:
# Description

Improve memory use in experiments by redefining what steps are unique. Two steps are considered the same if `type`, `value`, `termination` and `temperature` are the same, as these are the variables used when processing the model. The remaining variables are only used when running the model and can be passed as inputs. I have also defined a `basic_repr` which returns a representation with only those variables, while `__repr__` returns a representation with all the variables.

Fixes #3081 

## Type of change

# Key checklist:

## Further checks:


pybamm/solvers/solution.py
```diff
@@ -421,6 +424,15 @@ def cycles(self, cycles):
     def summary_variables(self):
         return self._summary_variables
 
+    @property
+    def initial_start_time(self):
+        return self._initial_start_time
+
+    @initial_start_time.setter
+    def initial_start_time(self, value):
+        """Updates the reason for termination"""
```
rtimms, line 433: update docstring

rtimms (approved):
thanks @brosaplanella looks good to me!

PR #3264, opened 2023-08-10
Split CI test groups into concurrent jobs

agriyakhetarpal:
# Description

> [!Important]  
> This PR also contains the changes made as a part of #3279, so both PRs should be merged together, with that one being merged before this one.

Improves CI time by creating concurrent jobs for separate `nox` sessions for running tests, also adds the concurrency setting to cancel in-progress jobs.

## Type of change

# Key checklist:

## Further checks:


.github/workflows/test_on_push.yml
```diff
@@ -19,6 +19,8 @@ jobs:
           cancel_others: "true"
           paths_ignore: '["**/README.md"]'
 
+# TODO: add concurrency syntax for groups and cancelling in-progress jobs
+
```
Saransh-cpp, line 23: See https://github.com/scikit-hep/vector/blob/349aca6123873bcbaa1470d411304dab4009327a/.github/workflows/ci.yml#L11-L15 for inspiration

agriyakhetarpal:
Failure for a job cancels other jobs in the matrix and the further jobs that depend on it, as intended: https://github.com/pybamm-team/PyBaMM/actions/runs/5823292074. Doctests on other platforms were cancelled because the ones on Windows and Python 3.11 failed, and subsequently the examples tests did not run.

.github/workflows/test_on_push.yml
agriyakhetarpal: FIxed in https://github.com/pybamm-team/PyBaMM/pull/3264/commits/ddc0fc587bbf95000ac2d4344f83b80d5ff0f94f, I improved the `vector` one to work on pushes outside of PRs as well

agriyakhetarpal:
I turned on all tests (unit, coverage, integration, doctests, examples) for all platforms temporarily. Just wanting to check runtimes for the test suite and those for individual jobs

agriyakhetarpal:
Weird that the coverage tests on macOS are [failing](https://github.com/pybamm-team/PyBaMM/actions/runs/5824687678/job/15794395295) in the workflows because of installing `[odes]`. It is unable to create wheels for `scikits.odes` (which is a common error), but this session passes for me locally (it iooked to me like a problem with the `set_environment_variables` function in `noxfile.py`, but it wasn't—the `SUNDIALS_INST` and `LD_LIBRARY_PATH` environment variables are indeed being set correctly). @Saransh-cpp we could skip it like we were doing before, but in theory this shouldn't be the case: let me know

Saransh-cpp:
> @Saransh-cpp we could skip it like we were doing before, but in theory this shouldn't be the case: let me know

Yes, we can skip them for now

agriyakhetarpal:
> Yes, we can skip them for now

Even more weird that `scikits.odes` now fails on Ubuntu with Python 3.11 too, which definitely should not happen

Saransh-cpp (commented):
We could also run integration tests in parallel to unit tests. This will save us a lot of CI time, and we can just let them fail if the unit tests are failing.

agriyakhetarpal:
Done, although it might not save us CI time because the examples tests will take ~35-40 minutes anyway

.github/workflows/run_periodic_tests.yml
```diff
@@ -102,7 +102,7 @@ jobs:
 
       - name: Upload coverage report
         if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.11
-        uses: codecov/codecov-action@v2.1.0
+        uses: codecov/codecov-action@v3.1.4
```
Saransh-cpp, line 105: These versions go stale very often in PyBaMM. I have opened #3276 for this.

Saransh-cpp (approved):
Looks great, thanks @agriyakhetarpal! I will let this sit for a while for others to take a look at 🙂

.github/workflows/run_periodic_tests.yml
agriyakhetarpal: Awesome!

agriyakhetarpal:
Also, do look at the document too that I had shared on Slack, when you are able to: it contains some more ideas that I can implement in this PR (or possibly in another PR that would get merged before this one does). Hopefully @tinosulzer has further suggestions on those points

agriyakhetarpal:
The CI runtime has indeed halved from the earlier ~65 minutes—the new examples job ran in 34 minutes. This means we can get sub-30-minute times on a good day 🎉

Saransh-cpp:
@agriyakhetarpal you can rebase this now!

agriyakhetarpal:
I know that we need eyes from others on this, a note will need to merge this soon—otherwise the example scripts will not be tested on further PRs (see #3279). We were ideally supposed to keep both of them open and merge them one after the other

PR #3266, opened 2023-08-10
[WIP] Refactor `Simulation.solve`

### LINKED ISSUE #3176: Refactor `Simulation.solve`
As part of #3101 I was looking at the `Simulation` class and noticed that the messiest part of it is basically the `solve` method (and in particular, the experiment part). In my mind, the bigger goal here would be to make the `Simulation` class (without experiments) to be compatible with any model (even models beyond battery).

I think that just simply refactoring the whole experiment part as a separate function would help a lot (maybe with an extra check of whether a given model supports experiments or not). The other main point I would like to discuss here is whether we deprecate the `"drive cycle"` mode, given that now this can easily be done with the experiment steps.

brosaplanella:
# Description

Refactor `Simulation.solve` to make it more readable and easy to maintain. Tasks:

- [x] Remove `drive cycle` mode.
- [x] Add integration test to check that simulation works with non-battery model (NOTE: added it in unit tests as there was a similar one for ODEs, let me know if I should move it to integration).
- [ ] Refactor `solve` method `with experiment` to reuse code. 

Fixes #3176 

## Type of change

# Key checklist:

## Further checks:


kratman:
@brosaplanella This is a bit old and has a lot of conflicts, can it be closed?

brosaplanella:
Yes please

PR #3292, opened 2023-08-23
Do not generate notebooks in PDF builds on Read the Docs

### LINKED ISSUE #3240: Enable pdf build in readthedocs
The [build is failing](https://readthedocs.org/projects/pybamm/builds/21517100/) with the following error -
```
Build output directory for format "pdf" contains multiple files and it is not currently supported.
Please, remove all the files but the "pdf" you want to upload.
```
Related to https://github.com/pybamm-team/PyBaMM/pull/3182?

cc: @agriyakhetarpal

agriyakhetarpal:
# Description

Please take a look at the discussion on Slack for more details. This PR removes the Jupyter notebooks when the Sphinx LaTeXBuilder is invoked and generates a singular PDF file. (Probably) fixes #3240

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -203,6 +203,15 @@ htmlhelp_basename = "PyBaMMdoc"
 
 # -- Options for LaTeX output ------------------------------------------------
 
+# Note: we exclude the examples directory from the LaTeX build because it has
+# problems with the creation of PDFs on Read the Docs
+# https://github.com/readthedocs/readthedocs.org/issues/2045
+
+# Detect if we are building LaTeX output through the invocation of the build commands
+if any("latex" in arg for arg in sys.argv) or any("latexmk" in arg for arg in sys.argv):
+    exclude_patterns.append("source/examples/*")
+    print("Skipping compilation of .ipynb files for LaTeX build.")
+
 latex_elements = {
     # The paper size ('letterpaper' or 'a4paper').
     #
--- a/docs/source/user_guide/index.md
+++ b/docs/source/user_guide/index.md
@@ -25,9 +25,13 @@ fundamentals/index
 
 # Example notebooks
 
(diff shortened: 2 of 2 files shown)


agriyakhetarpal:
The PDF can be built now and there is no `Fatal error` message like before, but it still returns an error code 1. I hope that it does not fail on RTD due to this reason

agriyakhetarpal:
@Saransh-cpp this is ready to merge now. I was able to generate a PDF using a custom build script that copies the PDF generated while ignoring LaTeX errors. Check it out here from the outputs of the temporary job I had added in the CI: https://github.com/pybamm-team/PyBaMM/suites/15568063870/artifacts/888533643

This will hopefully fix the PDF builds on Read the Docs for a long time (famous last words 😄).

agriyakhetarpal:
In case this does not fix it, we will have the following options:

1. Exclude PDF builds entirely
2. Look into alternative PDF solutions like `rinohtype` and provide PDF builds for new versions and onwards
3. Consider migration from Read the Docs to GitHub Pages and use a few more Sphinx extensions to provide versioned documentation with GitHub Actions (we might not get PR builds that way but we can always include another workflow to check if the builds pass).

agriyakhetarpal:
In any case, our PDFs can be generated correctly because we force a return code of 0 on `latexmk` commands, but we should move to the standard PDF builds at some point later on. See https://github.com/readthedocs/readthedocs.org/issues/10549

agriyakhetarpal:
It passed and I can see both the newest changes and the PDF download in the flyout menu: https://docs.pybamm.org/_/downloads/en/latest/pdf/

agriyakhetarpal:
It passed and I can see both the newest changes and the PDF download in the flyout menu: https://docs.pybamm.org/_/downloads/en/latest/pdf/

Issue #3312, opened 2023-09-05
Improvements to the Docker installation and usage procedures


agriyakhetarpal:
A placeholder issue for marking a list of tasks to be done for the current Docker images established in #3223.

- [ ] Ensure that macOS M-series users can pull the latest PyBaMM image from DockerHub without the use of the `--platform linux/x86_64` command-line flag
- [ ] Add relevant instructions for developers in order to contribute to PyBaMM (setting up `git` configuration, setting a `git` remote, document the addition or configuration of SSH and GPG keys, and other credentials)
- [ ] Enable the display of `matplotlib` plots for examples that use QuickPlot and similar by either switching to a TKAgg backend in the code, via a `matplotlibrc` file, or otherwise document this change in an FAQ or Troubleshooting section
- [ ] Enable root permissions in the Dockerfile by setting the user ID to 1000, so that users can install `apt` dependencies and do other tasks inside the container
- [ ] Add CI to build and push the Docker images either nightly or on commits to the `develop` branch, such that those pulling the image from DockerHub will always have the latest changes and would not have to do `git pull` every time
- [ ] Add a parallel job to run the entire test suite inside the Docker container in the scheduled tests workflow
- [ ] Push DockerHub images on releases, starting from either 23.9 or 24.1? Note: this would be more helpful to users rather than developers.
- [ ] Expose ports for smooth integration of Jupyter notebooks and briefly document how to run them inside a container
- [ ] Expose ports for building the documentation with `nox -s docs` / `sphinx-autobuild`
- [ ] Install all extras dependencies by default except optional solvers, which are provided through command-line arguments, and install `pandoc` and `graphviz` (required for the rendering of Jupyter notebooks and inheritance diagrams in the built documentation). With `apt`, users might get an older version of `pandoc` so we can try getting it from `conda-forge`.
- [ ] Add functionality for GitHub codespaces in order to enable an online development environment and add associated documentation about their usage and billing.

brosaplanella:
I tried it at my end with the current instructions and Git was setup for me automatically. It only prompted me to log in with my Github user, but I could pull and push easily (note that I have push access to the PyBaMM repo, so people who don't will have to set the remote).

agriyakhetarpal:
I tried it on my end as well and I faced some issues too. Here are some more things that we will have to take care of:

[...]

PR #3316, opened 2023-09-09
Create CI to build & push PyBaMM's Docker Images

arjxn-py:
# Description

This PR aims to create a CI that builds all PyBaMM's docker images & then pushes them to DockerHub.

## Type of change

# Key checklist:

## Further checks:


arjxn-py:
Related to #3312 
cc: @agriyakhetarpal @brosaplanella @Saransh-cpp

arjxn-py:
I'd also need help with setting up the docker secrets for this one.

agriyakhetarpal (commented):
Would building and pushing images in their separate jobs be a better way, rather than in a series of steps?

.github/workflows/docker.yml
```diff
@@ -0,0 +1,66 @@
+name: Build & Push Docker Images
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches:
+    - main
```
agriyakhetarpal, line 7: This should point to the `develop` branch.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,66 @@
...
+
+      - name: List built images
+        run: docker images
+
+      - name: Build and Push Docker Image (Without Solvers)
+        env:
+          IMAGE_NAME: pybamm/pybamm:latest
+        run: |
+          docker build -t $IMAGE_NAME -f scripts/Dockerfile .
+          docker push $IMAGE_NAME
```
agriyakhetarpal, line 34: We would need to use the official Docker action to push images (`docker/build-push-action`) and point it to a path where the built images eill be stored

arjxn-py:
> Would building and pushing images in their separate jobs be a better way, rather than in a series of steps?

Can they run parallelly in case of separate jobs?

arjxn-py:
> > Would building and pushing images in their separate jobs be a better way, rather than in a series of steps?
> 
Can they run parallelly in case of separate jobs? I'd be happy to switch to separate jobs either way though 🙂 
Thanks!

.github/workflows/docker.yml
arjxn-py: Thanks, will look into it.

agriyakhetarpal:
> Can they run parallelly in case of separate jobs?

Yes, they will run and upload images in parallel. We would need to check out the repository in each job, however. Since it would be developers who would be using the image, could you remove the sparse checkout and let it fetch the entire commit depth of the repository?

arjxn-py:
> could you remove the sparse checkout and let it fetch the entire commit depth of the repository?

Yes, sure.

agriyakhetarpal:
Actually, multiple jobs might not be needed in this case, a matrix for the solver arguments should suffice after the check out step

Saransh-cpp:
> I'd also need help with setting up the docker secrets for this one.

Do you want us to add them in the GitHub repo or do you want us to generate it from Dockerhub?

arjxn-py:
> Do you want us to add them in the GitHub repo or do you want us to generate it from Dockerhub?

I'd need help with adding them to github repo, i can generate & send secrets to you on slack.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,64 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Login to DockerHub
+        if: github.event_name != 'pull_request'
+        uses: docker/login-action@v2
+        with:
+            username: ${{ secrets.DOCKERHUB_USERNAME }}
+            password: ${{ secrets.DOCKERHUB_TOKEN }}
```
Saransh-cpp, line 22: Added these secrets in the PyBaMM repo. Let me know if everything works.

arjxn-py: This CI is working on my fork [here](https://github.com/arjxn-py/PyBaMM/actions/runs/6266390975/job/17017264690?pr=6).
Can you please once confirm if secrets are set how we want them to?

Saransh-cpp: Yes! We'll only be able to test it once this PR is merged. The secrets are not passed down to PRs from forks for security reasons.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,63 @@
+name: Build & Push Docker Images
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches:
+    - develop
```
Saransh-cpp, line 7: We should not run this on PRs as secrets are not passed down to PRs for security reasons. This should only run on a push to develop branch.

agriyakhetarpal:
Note: to push to Docker Hub we should use https://github.com/docker/build-push-action.

I noticed that they have some examples about setting up QEMU, we should be able to build both `AMD64` and `ARM64` with that (I may take a look in another PR).

.github/workflows/docker.yml
arjxn-py: > Yes I was just running it on this PR in order to test, this workflow can be triggered manually or run periodically afterwards, whichever preferred.

Saransh-cpp: Should be ready to merge once `pull_request` is removed. Now that I think of it, we should also add a cron job that builds and pushes docker images nightly? Pushing on every commit might be a bit excessive.

Saransh-cpp (approved):
Thanks, @arjxn-py! Looks great! Should be ready to merge once the comment below is resolved -

.github/workflows/docker.yml
agriyakhetarpal: I think we should also check out other repositories and projects as to how they do it (most related workflows push to the development branch and overwrite the previous image). Upon multiple pushes to `develop` we can cancel intermediate builds here as we do for the unit tests and benchmarks workflows

Saransh-cpp: > I think we should also check out other repositories and projects as to how they do it

+1!

agriyakhetarpal: The first workflow I looked at was `PyHF`'s, of course. They build docker images on PRs to their development branch (`main` in their case), pushes to `main`, scheduled nightly, on version tags, for releases, and also have a workflow dispatch trigger.

Building and pushing images on release tags _might_ require configuring the naming scheme of the images to include the version since we would not want to overwrite any existing images. We should look into that before `23.9` comes out (additionally `23.9rcX`). For now, I think just the two options: pushes to `develop` and a workflow dispatch trigger should be fine, with concurrency checks to cancel the build in case two PRs are merged one after the other. The primary disadvantage of the nightlies is that it could cause some hinderances for developers working within their containers in case a large PR or one with breaking changes gets merged and the Dockerfile is not updated, though this would be a rare occurrence—developers have to deal with merge conflicts sometimes anyway, and this is the case without Docker as well.

It is possible to exercise more fine-grained control over how we can push to Docker Hub. Some data science projects use a workflow call trigger with particular event webhooks to run a workflow from other workflows. To say it is complicated is an understatement but this way, in theory, one can build the Docker images only at the completion of a particular workflow by calling another workflow through it. In this case, it would be when the tests pass on merging a PR to `develop`, which, in-turn, will be cancelled on subsequent pushes due to the concurrency setting (or do we not run CI on merge commits anymore?) We should explore this option only if the solution above seems to be excessive or just switch to the nightlies.

Saransh-cpp: `push` and `workflow_dispatch` sounds good to me as well!

arjxn-py:
> storage limits on Docker Hub

As of my knowledge & research there's no [storage limit for images on DockerHub](https://forums.docker.com/t/does-docker-hub-have-a-size-limitation-on-repos-or-images/10154) (i.e. we can have as many tags we want) however there's an individual image size limit which we don't have to bother about as our image size is smaller.

agriyakhetarpal:
That's great, thanks! Pushing all images should be fine then

Saransh-cpp (approved):

agriyakhetarpal (approved):

.github/workflows/docker.yml
```diff
@@ -0,0 +1,75 @@
...
+          push: true
+          build-args: |
+            IDAKLU=true
+
+      - name: Build and Push Docker Image (With All Solvers)
+        uses: docker/build-push-action@v5
+        with:
+          context: .
+          file: scripts/Dockerfile
+          tags: pybamm/pybamm:latest
```
agriyakhetarpal, line 72: I think we did not tag the images properly. On https://hub.docker.com/r/pybamm/pybamm/tags I can see that this image was pushed with the `latest` tag, and

.github/workflows/docker.yml
```diff
@@ -0,0 +1,75 @@
...
+      - name: Login to DockerHub
+        uses: docker/login-action@v3
+        with:
+            username: ${{ secrets.DOCKERHUB_USERNAME }}
+            password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: List built images
+        run: docker images
+
+      - name: Build and Push Docker Image (Without Solvers)
```
agriyakhetarpal, line 29: this one **also** has the `latest` tag

agriyakhetarpal: the earlier one (with all solvers) should have had the `all` tag. I do have a local branch which fixes things up a bit for this workflow (including logging in and uploading these images in parallel jobs) and will write a PR soon, I might look into adding the multi-platform images there itself.

PR #3317, opened 2023-09-10
Add and update dependencies for Docker images

agriyakhetarpal:
# Description

See #3312

1. Bumps Python version to 3.11
2. Removes dependence on CMake==3.22 and the latest stable CMake version is used
3. `nox` is installed by default in both the `base` and the `pybamm` environments
4. Includes all extras dependencies and `[docs]` dependencies, adds `pandoc`, and and a `TeXLive` distribution (for the latexify tests, rendering notebooks in the built documentation, et cetera).

> [!NOTE]  
> This PR should be merged **after** #3316 

## Type of change

# Key checklist:

## Further checks:


scripts/Dockerfile
```diff
@@ -21,45 +21,50 @@ ENV CMAKE_C_COMPILER=/usr/bin/gcc
...
+    pip install --upgrade --user pip setuptools wheel wget nox && \
+    pip install cmake && \
     python scripts/install_KLU_Sundials.py && \
     git clone https://github.com/pybind/pybind11.git && \
-    pip install --user -e ".[all,dev,jax,odes]"; \
+    pip install --user -e ".[all,dev,docs,jax,odes]"; \
     fi
 
-RUN pip install --user -e ".[all,dev]"
+RUN pip install --user -e ".[all,dev,docs]"
```
Saransh-cpp, line 68: Should we install `setuptools wheel wget nox cmake` here instead of every `if` condition?

agriyakhetarpal: Done

scripts/Dockerfile
```diff
@@ -21,45 +21,44 @@ ENV CMAKE_C_COMPILER=/usr/bin/gcc
...
 ARG IDAKLU
 ARG ODES
 ARG JAX
 ARG ALL
 
-RUN conda create -n pybamm python=3.9
-RUN conda init --all
-SHELL ["conda", "run", "-n", "pybamm", "/bin/bash", "-c"]
-RUN conda install -y pip
+RUN pip install --upgrade --user pip setuptools wheel wget nox cmake
```
Saransh-cpp, line 36: Hmm, `nox` should be installed as a dev dependency -

https://github.com/pybamm-team/PyBaMM/blob/ec5383c7538ed54d41ea1afbbf61f076992535e1/setup.py#L256

```suggestion
RUN pip install --upgrade --user pip setuptools wheel wget cmake
```

scripts/Dockerfile
```diff
@@ -21,45 +21,44 @@ ENV CMAKE_C_COMPILER=/usr/bin/gcc
...
     python scripts/install_KLU_Sundials.py && \
-    pip install --user -e ".[all,odes,dev]"; \
+    git clone https://github.com/pybind/pybind11.git && \
+    pip install --user -e ".[all,dev,docs,odes]"; \
     fi
 
 RUN if [ "$JAX" = "true" ]; then \
-    pip install --user -e ".[jax,all,dev]";\
+    python scripts/install_KLU_Sundials.py && \
+    git clone https://github.com/pybind/pybind11.git && \
```
Saransh-cpp, line 52: This isn't required for JaxSolver
```suggestion
```

scripts/Dockerfile
```diff
@@ -21,45 +21,44 @@ ENV CMAKE_C_COMPILER=/usr/bin/gcc
...
-    pip install cmake==3.22 && \
-    pip install --upgrade --user pip setuptools wheel wget && \
     python scripts/install_KLU_Sundials.py && \
     git clone https://github.com/pybind/pybind11.git && \
-    pip install --user -e ".[all,dev,jax,odes]"; \
+    pip install --user -e ".[all,dev,docs,jax,odes]"; \
     fi
 
-RUN pip install --user -e ".[all,dev]"
+RUN pip install --user -e ".[all,dev,docs]"
```
Saransh-cpp, line 62: Can we check if all the optional args are false and only then execute this?

Saransh-cpp (changes requested):
Thanks, @agriyakhetarpal! A few comments below -

scripts/Dockerfile
agriyakhetarpal: I agree, I added it because Jax and scikits.odes are optional, but it would be good to have IDAKLU by default so that one would get IDAKLU + Jax. Do we want the optional solvers to be exclusive?

scripts/Dockerfile
agriyakhetarpal: Keeping it as is should be fine, since this line will run only if no optional args are provided or are false (they are false by default)

scripts/Dockerfile
```diff
@@ -21,45 +21,44 @@ ENV CMAKE_C_COMPILER=/usr/bin/gcc
...
-    pip install --user -e ".[all,dev]"; \
+    pip install --user -e ".[all,dev,docs]"; \
     fi
 
 RUN if [ "$ODES" = "true" ]; then \
-    pip install cmake==3.22 && \
-    pip install --upgrade --user pip wget && \
     python scripts/install_KLU_Sundials.py && \
-    pip install --user -e ".[all,odes,dev]"; \
+    git clone https://github.com/pybind/pybind11.git && \
```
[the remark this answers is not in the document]
agriyakhetarpal: Good catch, thanks. However, this would be valid only if both solvers are going to be exclusive of each other (see the other comment). Happy to remove otherwise

scripts/Dockerfile
Saransh-cpp: Hmm yes, but building pybamm with KLU locally would require extra time, should be okay if a developer fetches it from Dockerhub. We could make the IDAKLU solver as a default solver here.

agriyakhetarpal: The only issue would be an increase in the build time, for us (which is okay because images won't be built on pull requests), but we would have to update the docs about this change. At this time, your idea sounds more reasonable because the use of `IDAKLU=true` _and_ `JAX=true` is more explicit about what solver components are being installed, compared to the case if someone were to get `IDAKLU` by default with the Jax solver (one build argument is installing both, which is slightly misleading to the user).

agriyakhetarpal:
@Saransh-cpp and I both have ARM machines so debugging might be a bit difficult, @arjxn-py could you test the new changes locally?

arjxn-py:
> @Saransh-cpp and I both have ARM machines so debugging might be a bit difficult, @arjxn-py could you test the new changes locally?

Hey, so sorry for the delay in response, I'll test this and confirm.

agriyakhetarpal:
> Hey, so sorry for the delay in response, I'll test this and confirm.

No worries, I tested the images locally and fixed some errors with CMake. It requires a non-user installation if one installs it with `pip` otherwise the system CMake is used for compilation (which could be outdated). We still need the Jax versions to be bumped for those sets of images (JAX and ALL) to work natively on aarch64 (#3443), but the other images (IDAKLU and ODES) seem to be working well since all solver-related tests pass.

agriyakhetarpal:
This should be ready to merge but only after #3430

agriyakhetarpal:
A gentle bump here, @arjxn-py and @Saransh-cpp! The Docker images here were working for me locally – now that we have multi-architecture images being pushed, I think it should be fine to go ahead with this.

arjxn-py (approved):
Changes looks nice, thanks @agriyakhetarpal. 
I'll also confirm once after testing Dockerfile locally.

Saransh-cpp (approved):
Looks good, sorry for the delay. Thanks, @agriyakhetarpal!

PR #3318, opened 2023-09-10
Documentation suggestions for the installation and tutorials

kratman:
# Description

Minor updates to the documentation pages.

Changes:
 - Fixes a broken link between the windows installation page and the windows WSL installation page. The broken link was caused by having the incorrect name for the windows WSL page.
 - Removes a note about using chocolatey from the MacOS install from source tab. I left the note in the Linux install from source tab for the use-case where a windows installation is done with Cygwin. I can remove the note from both MacOS and Linux if desired.
 - Tutorial 6 now uses a temporary folder to write the pickle files. I feel it is cleaner to write output to a specific folder rather than the current working directory. Since this is my opinion and it does add a small amount of complexity, I can remove this change if it is unwanted.

## Type of change

Documentation and notebook changes only. No additional features or updates to the library.

# Key checklist:

## Further checks:

### DIFF
--- a/docs/source/examples/notebooks/getting_started/tutorial-6-managing-simulation-outputs.ipynb
+++ b/docs/source/examples/notebooks/getting_started/tutorial-6-managing-simulation-outputs.ipynb
@@ -24,7 +24,7 @@
     {
      "data": {
       "text/plain": [
-       "<pybamm.solvers.solution.Solution at 0x136d2fac0>"
+       "<pybamm.solvers.solution.Solution at 0x14c144a90>"
       ]
      },
      "execution_count": 1,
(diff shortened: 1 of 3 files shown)


Saransh-cpp (changes requested):
Thanks for working on this, @kratman! Looks good overall.

> Tutorial 6 now uses a temporary folder to write the pickle files. I feel it is cleaner to write output to a specific folder rather than the current working directory

I think this should be reverted, given that the change makes it difficult for new Python/PyBaMM users to follow up with the tutorial. Experienced users will not face any difficulties, but this might push away newcomers.

kratman:
No problem. I will revert that change in an hour or so. Thanks for taking a look.

Saransh-cpp (approved):

Saransh-cpp:
@all-contributors please add @kratman for docs and infra

PR #3333, opened 2023-09-13
Issue 3328 correctly point notebook download URLs according to versions

### LINKED ISSUE #3328: Released version of PyBaMM does not match tutorials
While working on the tutorials, I noticed that tutorial 4 was failing when I downloaded it from the user guide.

Cell:
```python
parameter_values["Current function [A]"] = 10
parameter_values["Open-circuit voltage at 100% SOC [V]"]=3.4
parameter_values["Open-circuit voltage at 0% SOC [V]"]=3.0
```
Result:
```bash
KeyError: "'Open-circuit voltage at 100% SOC [V]' not found. Best matches are ['SEI open-circuit potential [V]', 'Outer SEI open-circuit potential [V]', 'Upper voltage cut-off [V]']"
```

After digging into it a bit I noticed that the released version from both pip and conda lack the recently renamed "Open-circuit voltage at XX% SOC [V]" parameters from Chen2020.
```bash
(envPyBaMM) kratz@Kratzs-MacBook-Pro ~ % python
Python 3.9.18 | packaged by conda-forge | (main, Aug 30 2023, 03:53:08) 
[Clang 15.0.7 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import pybamm
>>> pybamm.__version__
'23.5'
>>> params = pybamm.ParameterValues("Chen2020")
>>> params.search("Open-circuit")
Inner SEI open-circuit potential [V]	0.1
Outer SEI open-circuit potential [V]	0.8
SEI open-circuit potential [V]	0.4
```
So the tutorials appear to have been updated before the new parameters were added to the release.

Is a new release of PyBaMM coming so that the tutorials will work with the pip/conda releases? The tutorials work fine when running in my IDE using PyBaMM installed from source.

agriyakhetarpal:
# Description

Closes #3328; ensures that users of the example notebooks can download the correct notebook from GitHub according to the latest version, stable versions starting from 23.5, and on PR builds. The Google Colab badge now points to the correct renditions of the notebooks, rather than the `develop` branch.

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -303,6 +303,41 @@ bibtex_tooltips = True
 # a conflict with the sphinx-docsearch extension for Algolia search
 
 nbsphinx_requirejs_path = ""
+
+# For notebook downloads (23.5 onwards), we get the version from the environment
+# variable READTHEDOCS_VERSION and set it accordingly.
+
+# If the version is set to "latest", then we are on the develop branch, and we
+# point to the notebook in the develop blob
+# If we are on "stable", we point to the notebook in the relevant release tree
+# for the PyBaMM version
+# On a PR build, we use READTHEDOCS_GIT_COMMIT_HASH which will always point to changes
+# made to a notebook, if any.
+# On local builds, the version is not set, so we use "latest".
+
+if (os.environ.get("READTHEDOCS_VERSION") == "latest") or (os.environ.get("READTHEDOCS_VERSION") is None): # noqa: E501
+    notebooks_version = "develop"
+    github_download_url = (
+        "https://github.com/pybamm-team/PyBaMM/blob/" + notebooks_version
+    )
+if os.environ.get("READTHEDOCS_VERSION") == "stable":
+    notebooks_version = version
+    github_download_url = (
+        "https://github.com/pybamm-team/PyBaMM/tree/v" + notebooks_version
+    )
+if os.environ.get("READTHEDOCS_VERSION_TYPE") == "external":
+    notebooks_version = os.environ.get("READTHEDOCS_GIT_COMMIT_HASH")
+    github_download_url = (
+        "https://github.com/pybamm-team/PyBaMM/blob/" + notebooks_version
+    )
+
+html_context.update(
+    {
+        "notebooks_version": notebooks_version,
+        "github_download_url": github_download_url,
+    }
+)
+
 nbsphinx_prolog = r"""
 
 {% set github_docname =
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -320,18 +320,17 @@ if (os.environ.get("READTHEDOCS_VERSION") == "latest") or (
 ):
     notebooks_version = "develop"
     append_to_url = f"blob/{notebooks_version}"
-    )
+
 if os.environ.get("READTHEDOCS_VERSION") == "stable":
     notebooks_version = version
     append_to_url = f"tree/v{notebooks_version}"
-    )
+
 if os.environ.get("READTHEDOCS_VERSION_TYPE") == "external":
     notebooks_version = os.environ.get("READTHEDOCS_GIT_COMMIT_HASH")
     append_to_url = f"blob/{notebooks_version}"
-    )
 
-google_colab_url = github_download_url.replace("github.com", "githubtocolab.com")
 github_download_url = f"https://github.com/pybamm-team/PyBaMM/{append_to_url}"
+google_colab_url = github_download_url.replace("github.com", "githubtocolab.com")
 
 html_context.update(
     {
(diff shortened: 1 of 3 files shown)


kratman:
Switching the download URL based on the current version of the docs page sounds good to me.

Will https://docs.pybamm.org default to stable after this change? I know it the previous version the pybamm docs defaulted to latest, which could make the problem repeat.

agriyakhetarpal:
> Will https://docs.pybamm.org/ default to stable after this change

I just checked the Read the Docs configuration options; I can't find a way to set the `stable` version as the default one instead of `latest` when pointing to https://docs.pybamm.org. However, we have been wanting to correct this by being more explicit about it. We initially wanted to implement a version warning banner/bar on the latest version to point users to the stable documentation, but we decided against it because we would not get it for older versions in that case which would be pointless (it cannot be done unless there is a server-side implementation for it). Read the Docs has disabled their version warning bar lately in favour of some work going on at https://github.com/readthedocs/addons that I will look to add as soon as it is available for users. 

>  I know it the previous version the pybamm docs defaulted to latest, which could make the problem repeat.

I get that – the versioning is deemed incorrect for notebooks on `23.5`, sadly there is not much that can be done to fix it as of now because the tag already exists. I am hoping this PR can fix it for further versions/releases at least

kratman:
Ok thank you for taking a look and this is definitely an improvement. I only asked about stable/latest because I had assumed I installed my PyBaMM environment incorrectly when I first saw the errors. However, a release would also prevent that since I assume major changes to parameter and external function names are rare. So this change extra on top of what I was hoping to see.

docs/conf.py
```diff
@@ -303,14 +303,53 @@
...
+
+# If the version is set to "latest", then we are on the develop branch, and we
+# point to the notebook in the develop blob
+# If we are on "stable", we point to the notebook in the relevant release tree
+# for the PyBaMM version
+# On a PR build, we use READTHEDOCS_GIT_COMMIT_HASH which will always point to changes
+# made to a notebook, if any.
+# On local builds, the version is not set, so we use "latest".
+
+if (os.environ.get("READTHEDOCS_VERSION") == "latest") or (os.environ.get("READTHEDOCS_VERSION") is None): # noqa: E501
```
Saransh-cpp, line 318: You should not `noqa` this line. `ruff` should be able to fix this? If not then you might have to run `black` once.

docs/conf.py
```diff
@@ -303,14 +303,53 @@
...
+    notebooks_version = "develop"
+    github_download_url = (
+        "https://github.com/pybamm-team/PyBaMM/blob/" + notebooks_version
```
Saransh-cpp, line 321: 1. f-strings should be preferred for string concatenation as they are more efficient
2. It would be better to just refactor out the github URL to avoid repetition
```suggestion
    append_to_url = f"blob/{notebooks_version}"
```

docs/conf.py
```diff
@@ -303,14 +303,53 @@
...
+    )
+if os.environ.get("READTHEDOCS_VERSION") == "stable":
+    notebooks_version = version
+    github_download_url = (
+        "https://github.com/pybamm-team/PyBaMM/tree/v" + notebooks_version
```
Saransh-cpp, line 326:
```suggestion
    append_to_url = f"tree/v{notebooks_version}"
```

docs/conf.py
```diff
@@ -303,14 +303,53 @@
...
+    )
+if os.environ.get("READTHEDOCS_VERSION_TYPE") == "external":
+    notebooks_version = os.environ.get("READTHEDOCS_GIT_COMMIT_HASH")
+    github_download_url = (
+        "https://github.com/pybamm-team/PyBaMM/blob/" + notebooks_version
```
Saransh-cpp, line 331:
```suggestion
    append_to_url = f"blob/{notebooks_version}"
```

docs/conf.py
```diff
@@ -303,14 +303,53 @@
...
+    )
+
+google_colab_url = github_download_url.replace("github.com", "githubtocolab.com")
```
Saransh-cpp, line 334:
```suggestion
google_colab_url = github_download_url.replace("github.com", "githubtocolab.com")
github_download_url = f"https://github.com/pybamm-team/PyBaMM/{append_to_url}"
```

Saransh-cpp (changes requested):
Thanks, @agriyakhetarpal! Looks good overall, some comments below.

docs/conf.py
agriyakhetarpal: Yeah I was a bit fuddled about how to refactor them. It looks much neater now, thanks!

Saransh-cpp (approved):
Thanks for fixing the mess 🙂

agriyakhetarpal:
Yes, please go ahead!

PR #3335, opened 2023-09-14
Replace deprecated pkg_resources

kratman:
# Description

While looking through the code, I noticed a deprecation warning for pkg_resources. I replaced the two usages I found in the code with importlib.metadata.

## Type of change

Only changes an imported library, there should be no real changes to the functionality.

# Key checklist:

## Further checks:


kratman:
The drop in coverage appears to be from the code in #3116, not the changes from this PR

agriyakhetarpal:
Thanks for this @kratman. I have been doing the same thing in #3301 😄

kratman:
@agriyakhetarpal I did not see the changes there when I went to do this. This one is small and ready to go though.

I rebased yesterday to see if it would retrigger the coverage report which was showing a drop due to a different merged PR. It passed everything else before I did the rebase.

agriyakhetarpal:
Yes, that PR is a bit stalled though, for other reasons. The changes there are similar to yours and everything passes, I expect them to pass here too

kratman:
The failing tests (```nox -s examples```) for this PR are failing on develop as well.

tests/unit/test_parameters/test_parameter_sets_class.py
```diff
@@ -25,7 +25,7 @@ def test_all_registered(self):
         """Check that all parameter sets have been registered with the
         ``pybamm_parameter_sets`` entry point"""
         known_entry_points = set(
-            ep.name for ep in pkg_resources.iter_entry_points("pybamm_parameter_sets")
+            ep.name for ep in il.entry_points()["pybamm_parameter_sets"]
```
agriyakhetarpal, line 28: Looks like `importlib.metadata` has had some API changes in the past: this part gives a deprecation warning on newer Python versions (3.10 and above). Here are the logs:

1. https://github.com/pybamm-team/PyBaMM/actions/runs/6224348861/job/16919897538?pr=3335#step:11:1335
2. https://github.com/pybamm-team/PyBaMM/actions/runs/6224348861/job/16919897976?pr=3335#step:11:1333

could you change it and add a conditional? Something like

```python
if sys.version_info < (3, 10)
    # something
else:
```

should do it

kratman: Ok I will take a look at fixing it

Saransh-cpp: Also, importing `importlib.metadata` as `il` is not very conventional, you could do `from importlib.metadata import ...`

kratman: I imported importlib as "il" due to the generic naming of "entry_points". I figured an object with that name could easily be created in a project this size. I tend to be cautious about potential name clashes or confusing names.

I can change that as well when I push the fix above.

pybamm/parameters/parameter_sets.py
```diff
@@ -37,9 +38,17 @@ class ParameterSets(Mapping):
...
             self.__all_parameter_sets[entry_point.name] = entry_point
 
+    @staticmethod
+    def get_entries(group_name):
+        # Wrapper for the importlib version logic
+        if sys.version_info < (3, 10):
+            return importlib.metadata.entry_points()[group_name]
+        else:
+            return importlib.metadata.entry_points(group=group_name)
+
```
kratman, line 51: I put the version logic in a function so it is in a single location, and can be removed later.

pybamm/parameters/parameter_sets.py
```diff
@@ -37,9 +38,17 @@ class ParameterSets(Mapping):
     def __init__(self):
         # Dict of entry points for parameter sets, lazily load entry points as
         self.__all_parameter_sets = dict()
-        for entry_point in importlib.metadata.entry_points()["pybamm_parameter_sets"]:
```
kratman, line 40: This line would also cause a warning, so I fixed it when I did the others.

kratman:
@Saransh-cpp, @agriyakhetarpal It looks like coverage is dropping due to the version logic. Is there something else you want me to change, or is this OK?

agriyakhetarpal:
Perhaps adding `# pragma: no cover` on the conditional should be okay in this case

Saransh-cpp:
> Perhaps adding # pragma: no cover on the conditional should be okay in this case

Yes, should be fine in this case.

pybamm/parameters/parameter_sets.py
```diff
@@ -37,9 +38,17 @@ class ParameterSets(Mapping):
...
+        for entry_point in self.get_entries("pybamm_parameter_sets"):
             self.__all_parameter_sets[entry_point.name] = entry_point
 
+    @staticmethod
+    def get_entries(group_name):
+        # Wrapper for the importlib version logic
+        if sys.version_info < (3, 10):
+            return importlib.metadata.entry_points()[group_name]
```
kratman, line 48:
```suggestion
            return importlib.metadata.entry_points()[group_name] # pragma: no cover
```

pybamm/parameters/parameter_sets.py
```diff
@@ -37,9 +38,17 @@ class ParameterSets(Mapping):
...
         # Dict of entry points for parameter sets, lazily load entry points as
         self.__all_parameter_sets = dict()
-        for entry_point in importlib.metadata.entry_points()["pybamm_parameter_sets"]:
+        for entry_point in self.get_entries("pybamm_parameter_sets"):
             self.__all_parameter_sets[entry_point.name] = entry_point
 
+    @staticmethod
+    def get_entries(group_name):
+        # Wrapper for the importlib version logic
+        if sys.version_info < (3, 10):
```
kratman, line 47:
```suggestion
        if sys.version_info < (3, 10): # pragma: no cover
```

kratman:
Changed thanks

Saransh-cpp (approved):
Thanks, @kratman!

PR #3342, opened 2023-09-16
Fix #3303 Added Better Option Checking

### LINKED ISSUE #3303: Better option checking

```python3
                raise NotImplementedError(
                    "distributions."
                )
```

Saswatsusmoy:
Fixes #3303 

The key changes:

- Use .get() instead of direct indexing to handle cases where the key is missing
- Check if the value is a tuple using isinstance()
- If it's a tuple, use the first index
- Handle lithium plating the same way

# Description

This allows the code to work regardless of whether the options are single values or tuples.

### DIFF
--- a/pybamm/models/full_battery_models/base_battery_model.py
+++ b/pybamm/models/full_battery_models/base_battery_model.py
@@ -481,12 +481,28 @@ class BatteryModelOptions(pybamm.FuzzyDict):
                 )
 
         # Options not yet compatible with particle-size distributions
-        if options["particle size"] == "distribution":
-            if options["lithium plating"] != "none":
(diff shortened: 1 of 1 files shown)


brosaplanella:
Before we can review this PR you have to fix the failing tests (e.g. https://github.com/pybamm-team/PyBaMM/actions/runs/6206766116/job/16852066323?pr=3342) but the issue seems to be with the particle-size distribution models.

rtimms:
@Saswatsusmoy thanks for working on this. Sorry if it wasn't clear in the issue, but the "particle size" is just one example of such an option. It would be good to implement better option checking that handles checking whether an option is a string or a tuple (for all options).

PR #3343, opened 2023-09-16
Updates for PyData theme 0.14.0

agriyakhetarpal:
# Description

Updates for the new version of the PyData Sphinx Theme, CSS improvements, and some cleanup with the documentation builds (dependencies, external links, etc).

## Type of change

# Key checklist:

## Further checks:


agriyakhetarpal:
There are some errors upon adding the version warning bar (the theme switcher button stops working and our versioning for dev branches is a bit different than what the configuration requires for the announcement banner to be active). As previously discussed, I think the one from Read the Docs addons would be better.

noxfile.py
```diff
@@ -16,13 +16,6 @@
     "SUNDIALS_INST": f"{homedir}/.local",
     "LD_LIBRARY_PATH": f"{homedir}/.local/lib:",
 }
-# Do not stdout ANSI colours on GitHub Actions
-if os.getenv("CI") == "true":
-    os.environ["NO_COLOR"] = "1"
-    # The setup-python action installs and caches dependencies by default, so we skip
```
agriyakhetarpal, line 22: Not needed anymore after we use separate jobs instead of running tests sequentially (#3264). I added coloured outputs to GitHub Actions too

noxfile.py
```diff
@@ -46,7 +39,7 @@ def run_pybamm_requires(session):
     """Download, compile, and install the build-time requirements for Linux and macOS: the SuiteSparse and SUNDIALS libraries."""  # noqa: E501
     set_environment_variables(PYBAMM_ENV, session=session)
     if sys.platform != "win32":
-        session.run_always("pip", "install", "wget", "cmake")
```
agriyakhetarpal, line 49: Using `session.run_always()` as an alias to `session.install` was only needed for local development for cases where `pip` did not display an output. It was and is fine to use it in CI

agriyakhetarpal, line 49: Using `session.run_always()` as an alias to `session.install` was only needed for local development for cases where `pip` did not display a list of dependencies installed in its output. It was and is fine to use it in CI

agriyakhetarpal:
Upon making changes to the noxfile to bring it back to its previous state, some of the older `nox` caches have started interfering with the current workflows. @Saransh-cpp, is there a chance you could delete the [current cache entries](https://github.com/pybamm-team/PyBaMM/actions/caches) and trigger another build for this branch?

.github/workflows/run_periodic_tests.yml
```diff
@@ -111,6 +111,10 @@ jobs:
         if: matrix.os == 'ubuntu-latest'
         run: nox -s doctests
 
+      - name: Check if the documentation can be built
+        if: matrix.os == 'ubuntu-latest'
+        run: nox -s docs
+
```
agriyakhetarpal, line 117: Because I noticed some errors and warnings are not caught by the doctests, this can help ensure we are free of them. We could move the examples job above in the code too so that it can run in case the docs and/or the doctests are going to fail.

agriyakhetarpal:
Oops, didn't realise I replicated the changes I made here in #3301 😄 @Saransh-cpp this is ready for merging now, the version warning bar on the PyData Theme seems to have some issues with detecting the preferred version, we should look into that some other time

Saransh-cpp (approved):
This looks amazing, thanks @agriyakhetarpal!

agriyakhetarpal:
Should I try to un-suppress the `myst.xref_warnings` here or would keeping them be okay?

Saransh-cpp:
Keeping them suppressed should be okay if they are unfixable in our case.

agriyakhetarpal:
I can fix them here, I opened a new issue for checking other broken links and references (#3387) if someone would want to pick it up. Also, I am considering moving the doctests steps to another job instead of running them before the notebooks as I mentioned above. Doing that will simplify the workflow (allows examples to run) and reduce CI runtime by a bit (~5-6 minutes), because this is still the longest job

agriyakhetarpal:
Okay, should be good to go now once the tests pass. After this, it looks like we will be able to achieve sub-30-minute CI times due to the changes in #3301 (will merge `develop` there after this is merged)

PR #3344, opened 2023-09-16
Revamped notebook testing infrastructure

### LINKED ISSUE #3165: Some notebook lines are not being tested
As spotted in https://github.com/pybamm-team/PyBaMM/pull/3110#discussion_r1267382517, currently some lines in the notebooks are not tested. I don't know the reason, but I suspect that it is because in [this example](https://github.com/pybamm-team/PyBaMM/blob/v23.5/docs/source/examples/notebooks/creating_models/6-a-simple-SEI-model.ipynb) the following lines are not being tested
```python3
import ipywidgets as widgets
widgets.interact(plot, t=widgets.FloatSlider(min=0,max=solution.t[-1]*1e6,step=0.1,value=0));
```
so the error is not detected.

I am not sure if we were aware of this or not. I will work with @arjxn-py to fix that particular example, but this issue is to discuss if we need to fix the notebook testing.

agriyakhetarpal:
# Description

Adds `pytest` plus the `pytest-xdist` and `nbmake` plugins to execute notebooks in parallel, along with associated and ancillary changes to the documentation and infrastructure

Fixes #3165 

## Type of change

# Key checklist:

## Further checks:


agriyakhetarpal:
Even if it is not a huge improvement shown in the CI with `pytest-xdist` because Linux GitHub Actions runners have [two cores](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) (it is still faster by ~5-7 minutes than the current job though), it is great to have for local development (on my machine with 8 physical cores, I can execute all the notebooks in 11 minutes, which is much faster than the previous 30 minutes). The times without `nbmake` and with non-parallel `nbmake` are nearly identical, with `nbmake` being slightly faster. We should be able to achieve sub-30 minute CI times by splitting the doctests into another job and reorganising some of the redundant job steps like the installation of PyBaMM dependencies (a job which can get repeated often).

Another thing to be noted is that `nbmake` is a notebook executor and runs notebooks from the directory they are located in. Our previous method for testing notebooks was to convert them to Python scripts with `nbconvert` and run them with Python or the ipython kernel, which has accounted for some errors. We did not go forward with alternatives like `nbval` because it is more of a tool to test notebooks by comparing their outputs to the stored ones, which would not be apt because many PyBaMM features can be stochastic in their nature. Reducing CI time for notebook execution was also a factor in consideration, which meant we did not opt for other tools like `papermill` and `testbook`, both of them seem to be less actively maintained than before. We looked at another notebook testing toolkit called `ploomber-engine` for programmatic access to testing notebooks and some advanced features such as parameterising them; however, `nbmake` as a command-line application seems easier to use for our limited use-case and will also provide `pytest` as a developer dependency for the possible migration of the test suite from `unittest` to `pytest` some time in the future

agriyakhetarpal:
Also, I'm not sure about this but `nbmake` has an Apache license, do we need to include its license file in our source code?

agriyakhetarpal:
Update: after the failures were fixed, the notebooks step in particular completed in 12 minutes, which looks like it is half of the previous times! Now our notebook tests will take roughly the same time as the unit tests or integration tests jobs.

setup.py
```diff
@@ -251,9 +251,16 @@ def compile_KLU():
...
-            "ruff", # For code style auto-formatting
-            "nox", # For running testing sessions
+            # For working with pre-commit hooks
+            "pre-commit",
+             # For code style checks: linting and auto-formatting
+            "ruff",
+            # For running testing sessions
+            "nox",
+            # For testing Jupyter notebooks
+            "pytest",
```
Saransh-cpp, line 261:
```suggestion
            "pytest>=6",
```

run-tests.py
```diff
@@ -440,13 +262,6 @@ def export_notebook(ipath, opath):
     if args.doctest:
         has_run = True
         run_doc_tests()
-    # Notebook tests
-    elif args.examples:
-        has_run = True
-        run_notebooks(interpreter)
-    if args.debook:
-        has_run = True
-        export_notebook(*args.debook)
```
Saransh-cpp, line 449: Should we keep this and run `pytest` within a `subprocess`? Would keep the testing suite a bit uniform??

Saransh-cpp (commented):
Thanks, @agriyakhetarpal! A fantastic step towards migrating to `pytest`!

run-tests.py
agriyakhetarpal: We could use a subprocess, but it would come with the hassle of capturing stdout and stderr according to `pytest`'s logging functionality. I am not sure how to do it because we noted that in case a worker keeps running a notebook cell that takes a bit longer to execute, it tends to duplicate the progress bar for that notebook in its output stream. Considering we would migrate to `pytest` and eventually use it directly instead of invoking tests through `run-tests.py`, we could deprecate this command-line option instead and/or perhaps add a comment here about how the new notebook tests can be run?

setup.py
Saransh-cpp: Scientific Python guide says `>=6`

I'm not sure why, never looked at it in detail, but we can keep it `>=6`

run-tests.py
Saransh-cpp: Deprecating sounds good. We should add a `ValueError` here.

agriyakhetarpal: Done

run-tests.py
```diff
@@ -380,10 +214,10 @@ def export_notebook(ipath, opath):
     parser.add_argument(
         "--examples",
         action="store_true",
-        help="Test all Jupyter notebooks in `docs/source/examples/`.",
+        help="Test all Jupyter notebooks in `docs/source/examples/` (deprecated, use nox or pytest instead).",  # noqa: E501
     )
     parser.add_argument(
-        "-debook",
+        "--debook",
```
agriyakhetarpal, line 220: Apparently, this was an error that wasn't caught. I updated the argument to match what the documentation says (or said, rather).

Saransh-cpp (approved):
Looks good, thanks @agriyakhetarpal!

PR #3358, opened 2023-09-19
Use trusted publisher deployment

Saransh-cpp:
# Description

PyPI now allows and encourages uploading files without using any API TOKENS. The trusted publisher deployment can be configured from PyBaMM's PyPI dashboard -

Your account > pybamm > publishing

The tab will look something like this -

image

Environment name should be set to `pypi` while filling in the details, and the workflow should work!

> [!NOTE]  
> The owner account will have to enable 2FA (which should not be an extra overhead, given that 2FA will be mandatory on all PyPI accounts by the end of this year).


brosaplanella:
I don't have access to it, @tinosulzer do you have access? If not, I can reset the password and set this up.

brosaplanella:
Any update on this? Should I chase for the access details?

Saransh-cpp:
> Should I chase for the access details?

Yes, that would be great. To be on the safer side, we can enable this for `v24.1`.

kratman:
@Saransh-cpp, @brosaplanella Is this still being done for v24.1

brosaplanella:
This has now been approved, so happy to go with it. @Saransh-cpp let me know if you need the credentials to log in.

agriyakhetarpal (commented):
Looks good, thanks, @Saransh-cpp and @brosaplanella – this should modify the TestPyPI job too with an accompanying addition of a `testpypi` environment in the repository, in case we need that on some occasion – I have found that pretty useful for some personal projects

Saransh-cpp:
> @Saransh-cpp let me know if you need the credentials to log in.

That would be great, @brosaplanella. I'll be able to update everything in PyPI.

> this should modify the TestPyPI job too with an accompanying addition of a testpypi environment in the repository

maybe sometime after the v24.1 release (or if someone else could take a look at it before the release)

agriyakhetarpal:
Now should be a good time to get this in. Also, to note: the API credentials should be removed from the TestPyPI step as well.

kratman:
@Saransh-cpp, @agriyakhetarpal, @brosaplanella What still needs to be done for this one?

agriyakhetarpal:
The changes in the workflow file are correct, we just need the `pypi` environment set up in the repository settings and the OIDC credentials set up on PyPI

Saransh-cpp:
Ahh, I always forget about this PR. Not this time.

Saransh-cpp:
Okay this should be good to go now (thanks for updating the settings!)

agriyakhetarpal:
> the API credentials should be removed from the TestPyPI step as well

Do we need this, @Saransh-cpp? Both of these uploads are in steps that use the same environment/job, so removing the API keys will break the TestPyPI uploads. We never know when we might require them.

kratman:
> > the API credentials should be removed from the TestPyPI step as well
> 
> Do we need this, @Saransh-cpp? Both of these uploads are in steps that use the same environment/job, so removing the API keys will break the TestPyPI uploads. We never know when we might require them.

I would say we should merge this and then open a new PR to also fix the TestPyPi stuff. That way we have this in place and if we forget the next step the failing test will remind us

kratman:
@agriyakhetarpal I created a ticket for the TestPyPi job

PR #3373, opened 2023-09-27
Bug fix: Make examples more compatible with Mac

kratman:
# Description

Fixing some issues in the notebooks that caused failures on MacOS and in Nox sessions.

Changes:
- Changed ```%pip install pybamm[plot,cite] -q``` to ```%pip install "pybamm[plot,cite]" -q``` in the notebooks. This prevents failures on MacOS.
- Fixed the temporary file name in the quick plot function create_gif() function. Running ```nox -s examples``` had failures due to "file not found" errors. The function would use the same file names on different threads creating a race condition.
- Fixed the link to keras' callbacks.py file. The files moved and the link no longer functioned.

## Type of change

Minor bug fixes in the notebooks.

- [x] Bug fix (non-breaking change which fixes an issue)

# Key checklist:

- [x] No style issues: `$ pre-commit run` (or `$ nox -s pre-commit`) (see [CONTRIBUTING.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CONTRIBUTING.md#installing-and-using-pre-commit) for how to set this up to run automatically when committing locally, in just two lines of code)
- [x] All tests pass: `$ python run-tests.py --all` (or `$ nox -s tests`)
- [x] The documentation builds: `$ python run-tests.py --doctest` (or `$ nox -s doctests`)

You can run integration tests, unit tests, and doctests together at once, using `$ python run-tests.py --quick` (or `$ nox -s quick`).

## Further checks:

- [x] Code is commented, particularly in hard-to-understand areas
- [x] Tests added that prove fix is effective or that feature works

PR #3377, opened 2023-09-28
tests: Use temporary directories in unit tests

kratman:
# Description

The unit tests leave a large number of files behind in the git repo. This change will make the tests use temporary directories when creating files.

I used the tempfile module in python. There are a few other ways to do this, and I believe the preferred method is to use pytest. However, since pytest does not appear to be used, I went with tempfile as the easy solution.

## Type of change

Minor fix in the tests. 

# Key checklist:

## Further checks:

### DIFF
--- a/tests/unit/test_batch_study.py
+++ b/tests/unit/test_batch_study.py
@@ -5,6 +5,7 @@ from tests import TestCase
 import os
 import pybamm
 import unittest
+from tempfile import TemporaryDirectory
 
 spm = pybamm.lithium_ion.SPM()
 spm_uniform = pybamm.lithium_ion.SPM({"particle": "uniform profile"})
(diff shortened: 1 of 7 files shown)


kratman:
This will have conflicts with #3373, so it should be reviewed/merged after that one is complete.

kratman:
Conflicts resolved, this should be ready.

kratman:
Lychee error is unrelated to this change

kratman:
@Saransh-cpp Can this be removed so that these files do not keep getting created when I test?

Saransh-cpp:
Sorry for the delay, @kratman, I'll review this before the end of this week. The extra files should not be a problem given that they should be ignored by git? Is there a specific redundant file that is not being ignored in `.gitignore`?

kratman:
None are missed by gitignore, just they exist until I run a cleanup command

Saransh-cpp (approved):
Thanks for working on this, looks amazing, @kratman!

PR #3379, opened 2023-09-29
Fix periodic benchmarks and remove version cap on `asv`

agriyakhetarpal:
# Description

The scheduled benchmarks suite was failing to run because the version was not pinned to 0.5.1. This PR resolves both benchmarks workflows and allows the benchmarks to run

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/.github/workflows/benchmark_on_push.yml
+++ b/.github/workflows/benchmark_on_push.yml
@@ -26,7 +26,7 @@ jobs:
 
       - name: Install python dependencies
         run: |
-          python -m pip install --upgrade pip wheel setuptools virtualenv wget cmake casadi numpy virtualenv packaging
+          python -m pip install --upgrade pip wheel setuptools wget cmake casadi numpy packaging
           python -m pip install asv[virtualenv]
 
       - name: Install SuiteSparse and SUNDIALS
--- a/.github/workflows/benchmark_on_push.yml
+++ b/.github/workflows/benchmark_on_push.yml
@@ -18,16 +18,20 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: 3.8
+
       - name: Install Linux system dependencies
         run: |
           sudo apt-get update
           sudo apt install gfortran gcc libopenblas-dev
+
       - name: Install python dependencies
-      # Pin asv==0.5.1 to fix failing benchmarks. Related to https://github.com/airspeed-velocity/asv/issues/1323
         run: |
-          python -m pip install --upgrade pip wheel setuptools virtualenv asv==0.5.1 wget cmake casadi numpy
-      - name: Install SuiteSparse and Sundials
+          python -m pip install --upgrade pip wheel setuptools virtualenv wget cmake casadi numpy virtualenv packaging
+          python -m pip install asv
+
+      - name: Install SuiteSparse and SUNDIALS
         run: python scripts/install_KLU_Sundials.py
+
       - name: Fetch base branch
         run: |
           # This workflow also runs for merge commits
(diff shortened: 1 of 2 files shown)


.github/workflows/benchmark_on_push.yml
```diff
@@ -49,7 +53,8 @@ jobs:
           HEAD_COMMIT=$(git rev-parse HEAD)
           echo $BASE_COMMIT | tee commits_to_compare.txt
           echo $HEAD_COMMIT | tee -a commits_to_compare.txt
-          asv run HASHFILE:commits_to_compare.txt --m "GitHubRunner" --show-stderr --strict -v
+          asv run HASHFILE:commits_to_compare.txt --m "GitHubRunner" --show-stderr -v
```
Saransh-cpp, line 56: Re: removal of `--strict`

Do we want this workflow to pass even if the benchmarks fail? `asv` exits with a non-zero return code only if `--strict` is present. If we remove this, the workflow will always pass.

.github/workflows/periodic_benchmarks.yml
```diff
@@ -20,26 +20,33 @@ jobs:
...
+          python -m pip install asv
+
+      - name: Install SuiteSparse and SUNDIALS
         run: python scripts/install_KLU_Sundials.py
+
       - name: Run benchmarks
         run: |
           asv machine --machine "GitHubRunner"
-          asv run --machine "GitHubRunner" NEW --show-stderr --strict -v
+          asv run --machine "GitHubRunner" NEW --show-stderr -v
```
Saransh-cpp, line 45: Ditto

.github/workflows/benchmark_on_push.yml
```diff
@@ -18,16 +18,20 @@ jobs:
...
           sudo apt-get update
           sudo apt install gfortran gcc libopenblas-dev
+
       - name: Install python dependencies
-      # Pin asv==0.5.1 to fix failing benchmarks. Related to https://github.com/airspeed-velocity/asv/issues/1323
         run: |
-          python -m pip install --upgrade pip wheel setuptools virtualenv asv==0.5.1 wget cmake casadi numpy
-      - name: Install SuiteSparse and Sundials
+          python -m pip install --upgrade pip wheel setuptools virtualenv wget cmake casadi numpy virtualenv packaging
+          python -m pip install asv[virtualenv]
```
Saransh-cpp, line 30: We can get rid of `virtualenv` from one of these lines.

.github/workflows/periodic_benchmarks.yml
```diff
@@ -20,26 +20,33 @@ jobs:
...
         run: |
           sudo apt-get update
           sudo apt-get install gfortran gcc libopenblas-dev
+
       - name: Install python dependencies
         run: |
-          python -m pip install --upgrade pip wheel setuptools virtualenv asv wget cmake casadi numpy
-      - name: Install SuiteSparse and Sundials
+          python -m pip install --upgrade pip wheel setuptools virtualenv wget cmake casadi numpy virtualenv packaging
+          python -m pip install asv[virtualenv]
```
Saransh-cpp, line 37: Ditto!

Saransh-cpp (changes requested):
Thanks, @agriyakhetarpal! Some comments below -

.github/workflows/benchmark_on_push.yml
agriyakhetarpal: They deprecated the `--strict` command-line flag in `0.6` so the benchmarks will always fail if they are destined to ([changelog](https://asv.readthedocs.io/en/v0.6.1/changelog.html#id5))

.github/workflows/benchmark_on_push.yml
agriyakhetarpal: Done

.github/workflows/benchmark_on_push.yml
Saransh-cpp: Oh nice! Thanks for the clarification!

Saransh-cpp (approved):

PR #3381, opened 2023-09-29
Added support for tuples in option checking. Fixes #3303

### LINKED ISSUE #3303: Better option checking

aryan1165:
# Description

Earlier:
Screenshot 2023-09-29 at 9 24 25 PM

Earlier "particle size" was not treated as a tuple and this caused problem as explained in Issue #3303.

Now:

Screenshot 2023-09-29 at 9 33 27 PM

Now, there is added compatibility for "particle size" to be either a tuple or a single value.

Fixes #3303 

# Key checklist:


kratman:
This seems to be the same as PR #3342

PR #3378, opened 2023-09-29
resolve [Bug]: Coordinate systems of variables do not get checked against known #3114

RuiheLi:
# Description

Fixes # 3114

## Type of change

Please add a line in the relevant section of [CHANGELOG.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CHANGELOG.md) to document the change (include PR #) - note reverse order of PR #s. If necessary, also add to the list of breaking changes.

- [Yes] Bug fix (non-breaking change which fixes an issue)

# Key checklist:

- [ ] No style issues: `$ pre-commit run` (or `$ nox -s pre-commit`) (see [CONTRIBUTING.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CONTRIBUTING.md#installing-and-using-pre-commit) for how to set this up to run automatically when committing locally, in just two lines of code)
- [ ] All tests pass: `$ python run-tests.py --all` (or `$ nox -s tests`)
- [ ] The documentation builds: `$ python run-tests.py --doctest` (or `$ nox -s doctests`)

You can run integration tests, unit tests, and doctests together at once, using `$ python run-tests.py --quick` (or `$ nox -s quick`).

## Further checks:

- [ ] Code is commented, particularly in hard-to-understand areas
- [ ] Tests added that prove fix is effective or that feature works

PR #3394, opened 2023-10-02
Issue 3114 check coord

### LINKED ISSUE #3114: [Bug]: Coordinate systems of variables do not get checked against known
### PyBaMM Version

23.5

### Python Version

3.9.16

### Describe the bug

Currently the coordinate system of a variable is not checked against the known system. This means that if the wrong argument is passed (e.g. due to a typo) it will default to `"cartesian"` without throwing an error.

### Steps to Reproduce

The MWE would be very lengthy, but basically it can be checked that:

1. The only place where `KNOWN_COORD_SYS` is checked is to test meshes for the battery model (i.e. it does not actually check for wrong coordinate systems).
2. In `finite_volume.py` there is an else statement that means that anything not being cylindrical polar or spherical polar will default to cartesian, see

https://github.com/pybamm-team/PyBaMM/blob/fffe650253596f14b6918d9e57fdd912bc3cfbdc/pybamm/spatial_methods/finite_volume.py#L139-L150

### Relevant log output

_No response_

RuiheLi:
# Description

Fixes #3114

## Type of change

# Key checklist:

## Further checks:


RuiheLi:
sorry I accidentally create another one

agriyakhetarpal:
I think this was the correct way. @RuiheLi, you may reopen this pill request and cherry-pick your commits from the develop branch (#3378) to this branch so that you don't lose progress

RuiheLi:
> I think this was the correct way, because it is from a feature branch to the `develop` branch. @RuiheLi, you may reopen this pull request and cherry-pick your commits from the `develop` branch (#3378, which can be closed afterwards) to this branch so that you don't lose progress

Thank you so much for the kind suggestion. Reopen now. 
All the commits before should be alright because I cereated the branch issue_3114_check_coord based on the develop branch.

pybamm/spatial_methods/finite_volume.py
```diff
@@ -146,11 +146,29 @@ def divergence(self, symbol, discretised_symbol, boundary_conditions):
...
             elif submesh.coord_sys == "cylindrical polar":
                 out = divergence_matrix @ (r_edges * discretised_symbol)
-        else:
+        elif submesh.coord_sys == "cartesian":
             out = divergence_matrix @ discretised_symbol
-
+        elif submesh.coord_sys == "none":
+            # what should we do now
+            print("It is none!")
+            out = divergence_matrix @ discretised_symbol
```
RuiheLi, line 154: I have a question here actually. if the coord_sys is none, what do we do for `out = divergence_matrix @ discretised_symbol`? @agriyakhetarpal

agriyakhetarpal: I haven't worked with the spatial methods before haha, perhaps @brosaplanella can answer this better

However, here, the coordinate system of the submesh should not be compared with `"none"` (which is a `str`) but rather with `None` (a null object in Python).

kratman: I took his comment to mean that None was already an acceptable option since it would default to Cartesian. I have not worked on it to double check though.

So something sort of like this instead of adding to "KNOWN_COORD_SYS".
```python
if coord_sys is not None and coord_sys not in KNOWN_COORD_SYS:
            raise ValueError(
                f"Coordinate system is {coord_sys}, not in cartesian, cylindrical polar, spherical polar, None")
```

brosaplanella: I think `None` does not need to be caught here (as if there is no `coord_sys` we do not need to discretise). You just need to catch it as an exception at `independent variable.py`. I think that should make the tests pass.

pybamm/expression_tree/independent_variable.py
```diff
@@ -131,6 +131,10 @@ def __init__(
             raise pybamm.DomainError(
                 "domain cannot be particle if name is '{}'".format(name)
             )
+        # check coord_sys against KNOWN_COORD_SYS
+        if coord_sys not in KNOWN_COORD_SYS:
+            raise ValueError(
+                f"Coordinate system is {coord_sys}, not in cartesian, cylindrical polar, spherical polar, none")
```
kratman, line 137: I moved my comment to a suggestion.
```suggestion
        # check coord_sys against KNOWN_COORD_SYS
        if coord_sys is not None and coord_sys not in KNOWN_COORD_SYS:
            raise ValueError(
                f"Coordinate system is {coord_sys}, not in cartesian, cylindrical polar, spherical polar, None")
```

pybamm/expression_tree/independent_variable.py
```diff
@@ -5,7 +5,7 @@
 
 import pybamm
 
-KNOWN_COORD_SYS = ["cartesian", "cylindrical polar", "spherical polar"]
+KNOWN_COORD_SYS = ["cartesian", "cylindrical polar", "spherical polar","none"]
```
kratman, line 8: None can be removed from here

pybamm/expression_tree/independent_variable.py
```diff
@@ -131,6 +131,10 @@ def __init__(
             raise pybamm.DomainError(
                 "domain cannot be particle if name is '{}'".format(name)
             )
+        # check coord_sys against KNOWN_COORD_SYS
+        if coord_sys is not None and coord_sys not in KNOWN_COORD_SYS:
+            raise ValueError(
+                f"Coordinate system is {coord_sys}, not in cartesian, cylindrical polar, spherical polar, None")
```
brosaplanella, line 137: This line is too long and it crashes the precommit, please reduce so it is maximum 88 characters.

pybamm/spatial_methods/finite_volume.py
```diff
@@ -146,11 +146,25 @@ def divergence(self, symbol, discretised_symbol, boundary_conditions):
                 out = divergence_matrix @ ((r_edges**2) * discretised_symbol)
             elif submesh.coord_sys == "cylindrical polar":
                 out = divergence_matrix @ (r_edges * discretised_symbol)
-        else:
+        elif submesh.coord_sys == "cartesian":
             out = divergence_matrix @ discretised_symbol
-
+        else:
+            raise ValueError(
+                f"Coordinate system is {submesh.coord_sys}, not in cartesian, cylindrical polar, spherical polar")
```
brosaplanella, line 153: This line is too long and it crashes the precommit, please reduce so it is maximum 88 characters.

pybamm/spatial_methods/finite_volume.py
```diff
@@ -146,11 +146,25 @@ def divergence(self, symbol, discretised_symbol, boundary_conditions):
...
         return out
 
+    def check_coordinate_system(self, symbol):
+        """
+        Double check coordinate system
+        """
+        submesh = self.mesh[symbol.domain]
+        if submesh.coord_sys not in ["cylindrical polar", "spherical polar","cartesian"]:
```
brosaplanella, line 161: This line is too long and it crashes the precommit, please reduce so it is maximum 88 characters.

pybamm/spatial_methods/finite_volume.py
```diff
@@ -146,11 +146,25 @@ def divergence(self, symbol, discretised_symbol, boundary_conditions):
...
+            raise ValueError(
+                f"Coordinate system is {submesh.coord_sys}, not in cartesian, cylindrical polar, spherical polar")
```
brosaplanella, line 163: This line is too long and it crashes the precommit, please reduce so it is maximum 88 characters.

pybamm/spatial_methods/finite_volume.py
```diff
@@ -146,11 +146,25 @@ def divergence(self, symbol, discretised_symbol, boundary_conditions):
                 out = divergence_matrix @ ((r_edges**2) * discretised_symbol)
             elif submesh.coord_sys == "cylindrical polar":
                 out = divergence_matrix @ (r_edges * discretised_symbol)
-        else:
+        elif submesh.coord_sys == "cartesian":
             out = divergence_matrix @ discretised_symbol
-
+        else:
```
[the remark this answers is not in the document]
tinosulzer: Better to only check in `independent_variable.py` in order to reduce code duplication

brosaplanella (commented):
Looks good, thanks! Before I can approve, can you add a line to the CHANGELOG?

You can copy the format of one of the other lines, but you will need to add it under the "Unreleased" section rather than 23.9.

brosaplanella:
@all-contributors please add @RuiheLi for code and tests

brosaplanella:
@all-contributors please add @RuiheLi for code and tests

pybamm/spatial_methods/finite_volume.py
```diff
@@ -146,11 +146,12 @@ def divergence(self, symbol, discretised_symbol, boundary_conditions):
                 out = divergence_matrix @ ((r_edges**2) * discretised_symbol)
             elif submesh.coord_sys == "cylindrical polar":
                 out = divergence_matrix @ (r_edges * discretised_symbol)
-        else:
+        elif submesh.coord_sys == "cartesian":
             out = divergence_matrix @ discretised_symbol
-
         return out
```
kratman, line 151: I think this should have an else to catch potential issues in the future. The else could raise an error for an unknown type. I know this should be caught by
```python
        # check coord_sys against KNOWN_COORD_SYS
        if coord_sys is not None and coord_sys not in KNOWN_COORD_SYS:
            raise ValueError(
                f"Coordinate system is {coord_sys}, not in"
                " cartesian, cylindrical polar, spherical polar, None")
```
but it is good to write code as safely as possible.

kratman (changes requested):

pybamm/spatial_methods/finite_volume.py
RuiheLi: Sorry for the disappear. Done now.

kratman:
Please update the PR description so that 3114 gets closed by this change (use the fixes tag in the template)

pybamm/spatial_methods/finite_volume.py
```diff
@@ -135,22 +135,28 @@ def divergence(self, symbol, discretised_symbol, boundary_conditions):
...
+            elif coord_sys == "cylindrical polar":
                 out = divergence_matrix @ (r_edges * discretised_symbol)
-        else:
+        elif coord_sys == "cartesian":
             out = divergence_matrix @ discretised_symbol
-
+        elif coord_sys is not None:
+            raise ValueError(
+                f"Coordinate system is {coord_sys}, not in"
+                " cartesian, cylindrical polar, spherical polar, None")
```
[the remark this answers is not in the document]
brosaplanella: Also note that this line is not covered so an additional test is needed.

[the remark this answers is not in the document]
RuiheLi: ok, that comes something I am very weak now. Just reading the test_finite_volume.py file and only come up with:

`    def test_divergence(self):
        whole_cell = ["negative electrode", "separator", "positive electrode"]
        var = pybamm.Variable("var", domain=whole_cell)
        x = pybamm.SpatialVariableEdge(
                "x_s",
                domain=["separator"],
                coord_sys="cartesian_2",
            )
        with self.assertRaises(ValueError):
            pybamm.SpatialMethod.divergence(var,x)`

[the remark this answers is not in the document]
RuiheLi: but I am pretty sure this won't work, so just to raise a question.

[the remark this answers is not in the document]
brosaplanella: This test is similar to the one you have, which is necessary because it checks the error is raised in `independent_variable.py`. You need another test that calls directly the `divergence` method in `finite_volume.py`. However, I am not sure we can do that test without triggering the other error before. That's because in order to define the symbol to pass to `divergence` we need to give it a domain with its own coordinate system. @kratman, any thoughts?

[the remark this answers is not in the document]
RuiheLi: Thank you for the kind reply. @brosaplanella That's something I am a bit struggled as well. Though I notice that I can get some mesh by  `mesh = get_mesh_for_testing()`. If I look closer at `get_mesh_for_testing`, they already assign coordinate system, and I don't know how to change that.

kratman:
@RuiheLi Are you still working on this?

RuiheLi:
> @RuiheLi Are you still working on this?

I have added the final else as suggested. But I had a try and still cannot write a proper test for the `divergence` as it will require to define a spatial variable with incorrect coordinate system first which will trigger an early error.

brosaplanella:
> I have added the final else as suggested. But I had a try and still cannot write a proper test for the `divergence` as it will require to define a spatial variable with incorrect coordinate system first which will trigger an early error.

@kratman what's the best way to approach that? Should we add a `# pragma: no cover` tag?

kratman:
I will take a look at this closer in the afternoon and give my opinion on the tests

pybamm/spatial_methods/finite_volume.py
```diff
@@ -135,22 +135,30 @@ def divergence(self, symbol, discretised_symbol, boundary_conditions):
...
             r_edges_numpy = np.kron(np.ones(second_dim_repeats), submesh.edges)
             r_edges = pybamm.Vector(r_edges_numpy)
-            if submesh.coord_sys == "spherical polar":
+            if coord_sys == "spherical polar":
                 out = divergence_matrix @ ((r_edges**2) * discretised_symbol)
-            elif submesh.coord_sys == "cylindrical polar":
+            elif coord_sys == "cylindrical polar":
                 out = divergence_matrix @ (r_edges * discretised_symbol)
-        else:
+        elif coord_sys == "cartesian":
```
rtimms, line 150: there are a few other places where there is similar logic that should be updated (do a search for `submesh.coord_sys in ["cylindrical polar", "spherical polar"]`)

rtimms:
> I have added the final else as suggested. But I had a try and still cannot write a proper test for the divergence as it will require to define a spatial variable with incorrect coordinate system first which will trigger an early error.

>@kratman what's the best way to approach that? Should we add a # pragma: no cover tag?

now that the coordinate system is already checked should the logic in discretisation be changed? i.e. we don't need to throw that error and the final else can just be the Cartesian case? alternatively, if we want to keep the error and have a test you could create a spatial variable with the correct domain, then manually overwrite `cood_sys` for the submesh later and then create the divergence object. seems clunky though

kratman:
@RuiheLi Sorry it took so long for me to respond, but I think there is a nice workaround to test this change without ignoring anything lines.

If you are having trouble testing that function, then it might get easier if you create a function for the block in question. If you create a function called something along the lines of "check_coordinate_system()" for the if-elif-else block, then it can take all things it is supposed to check as arguments. That might make it easier to write a test without adding making the test too complicated. I did something similar in this commit: https://github.com/pybamm-team/PyBaMM/pull/3682/commits/fa6bdc22eac87e63a3a200bc826ed1eb97b2ebea

In general short functions are easy to test. So breaking up complicated functions is probably a good thing anyway. Whenever it is difficult to test a function and we consider adding a pragma to ignore it. It is often a sign that we need to break up the function into smaller testable methods.

rtimms:
thanks @kratman, i like this approach as we check coordinates a few times in that file, so will reduce duplication too

kratman:
@RuiheLi Are you still working on this?

PR #3397, opened 2023-10-02
Serialisation of models

pipliggins:
# Description

Creates a serialisation method for writing PyBaMM models to JSON. Models can be written to/read-in from JSON by saving the discretised model properties. Variables, meshes and geometry can be optionally saved if users wish to use PyBaMM's plotting functionality when a model is read back in and solved.

A Serialise class is created to contain the methods, including custom JSON encoders for PyBaMM objects which can iterate through expression trees.
Symbols and meshes have to_json and from_json functions added to enable PyBaMM objects to be serialised using JSON.

Fixes # 2787

## Type of change

# Key checklist:

## Further checks:


pybamm/expression_tree/array.py
```diff
@@ -57,6 +57,30 @@ def __init__(
...
+                ),
+                shape=snippet["entries"]["shape"],
+            )
+        else:
+            matrix = snippet["entries"]
+
+        instance.__init__(
+            matrix,
+            name=snippet["name"],
+            domains=snippet["domains"],
```
martinjrobins, line 79: question: do we need to serialise domains (sorry if I've asked this before, feels like we might have)? Since we are targeting already discretised models, perhaps we don't need domains? Or is this for plotting etc.

pybamm/expression_tree/operations/serialise.py
```diff
@@ -0,0 +1,384 @@
...
+            The desired name of the JSON file. If no name is provided, one will be
+            created based on the model name, and the current datetime.
+        """
+        if model.is_discretised is False:
+            raise NotImplementedError(
+                "PyBaMM can only serialise a discretised, ready-to-solve model."
+            )
+
+        model_json = {
+            "py/object": str(type(model))[8:-2],
```
martinjrobins, line 110: perhaps we should include the version of pybamm used to serialize the model? might need this when handling loading legacy models in later versions of pybamm

pybamm/expression_tree/unary_operators.py
```diff
@@ -34,6 +34,21 @@ def __init__(self, name, child, domains=None):
...
         self.child = self.children[0]
 
+    @classmethod
+    def _from_json(cls, name, snippet: dict):
+        """Use to instantiate when deserialising"""
+
+        instance = cls.__new__(cls)
+
+        super(UnaryOperator, instance).__init__(
+            name,
```
martinjrobins, line 44: is `name` here going to be the same as `self.name`?

pybamm/meshes/one_dimensional_submeshes.py
```diff
@@ -34,7 +34,7 @@ def __init__(self, edges, coord_sys, tabs=None):
         self.internal_boundaries = []
 
         # Add tab locations in terms of "left" and "right"
-        if tabs:
+        if tabs and "negative tab" not in tabs.keys():
             self.tabs = {}
```
martinjrobins, line 38: I'm not sure I understand what is going on here. Why are we checking that `negative tab` isn't in the dict?

pybamm/meshes/one_dimensional_submeshes.py
```diff
@@ -52,6 +52,9 @@ def near(x, point, tol=3e-16):
                         f"{tab} tab located at {tab_location}, "
                         f"but must be at either 0 or {l_z}"
                     )
+        elif tabs:
```
martinjrobins, line 55: so presumably "negative tab" will only be in `tabs` if the model has been serialised. I still don't understand why tho?

martinjrobins (commented):
Thanks @pipliggins, this is great work to serialise/deserialise such a large and complicated class(es)! I've only got up to `pybamm/models/event.py` in the review and need to stop for now, so I'll just add this part as a comment then finish the review later. I notice that there is a failing test to fix as well, looks like a small thing though

pybamm/plotting/quick_plot.py
```diff
@@ -154,6 +154,10 @@ def __init__(
                     f"No default output variables provided for {models[0].name}"
                 )
 
+        # check variables have been provided after any serialisation
+        if any(len(m.variables) == 0 for m in models):
+            raise AttributeError("Variables not provided by the serialised model")
```
martinjrobins, line 159: could just be a model without any variables? Perhaps change error to "No variables to plot"

tests/unit/test_expression_tree/test_interpolant.py
```diff
@@ -325,6 +326,52 @@ def test_processing(self):
 
         self.assertEqual(interp, interp.new_copy())
 
+    def test_to_json_error(self):
+        x = np.linspace(0, 1, 10)
+        y = pybamm.StateVector(slice(0, 2))
+        interp = pybamm.Interpolant(x, 2 * x, y)
+
+        print(interp.children)
```
martinjrobins, line 334: remove print

tests/unit/test_expression_tree/test_interpolant.py
```diff
@@ -325,6 +326,52 @@ def test_processing(self):
 
         self.assertEqual(interp, interp.new_copy())
 
+    def test_to_json_error(self):
```
martinjrobins, line 329:
```suggestion
    def test_to_json(self):
```

tests/unit/test_expression_tree/test_unary_operators.py
```diff
@@ -52,6 +53,20 @@ def test_negation(self):
             pybamm.PrimaryBroadcast(pybamm.PrimaryBroadcast(nega, "test"), "test2"),
         )
 
+        # Test from_json
+        input_json = {
+            "name": "-",
+            "id": -2659857727954094888,
```
martinjrobins, line 59:
```suggestion
            "id": mock.ANY,
```

martinjrobins (changes requested):
finished going through this now, sorry for the delay! Looks excellent @pipliggins, happy to merge once you've dealt with the minor comments below

pybamm/expression_tree/array.py
pipliggins: We don't need the domains for running the model alone, but when plotting is required and variables/meshes have been saved, the variable domains are used to match to the correct mesh when deserialising.

pybamm/expression_tree/operations/serialise.py
pipliggins: Good idea, I'll add this.

pybamm/expression_tree/unary_operators.py
pipliggins: It should be, yes - I'll edit this to be more in line with other similarly structured subclasses (e.g. in binary_operators.py), where the superclass _from_json() function just takes the dictionary snippet, taking the name from that dictionary.

pybamm/meshes/one_dimensional_submeshes.py
pipliggins: At the point of mesh initialisation, if tabs have been provided, they look something like
`{"negative": {"z_centre": 0}, "positive": {"z_centre": 1}}`.
During the init function they are converted so that self.tabs = e.g. 
`{"negative tab": "left, "positive tab": "right"}`,
so an initialised mesh['tabs'] will have `negative tab` in the dict.

The former format is never saved, and the serialiser only ever a fully initialised mesh, i.e. the point where the tab locations have been fully determined, so that it what's written out to the json file. When reading back in and hitting the __init__ function to re-create the mesh, we search for whether the lab location has already been determined (by the presence of `negative tab` vs `negative` in the tab dict) and proceed accordingly.

pybamm/meshes/one_dimensional_submeshes.py
pipliggins: See above - "negative tab" is present in `tabs` once a mesh has been initialised, but __init__ will only come across this if a mesh has been serialised and is now being re-created upon deserialisation.

martinjrobins (changes requested):
Thanks @pipliggins, all looks good, except for a few added lines that arn't covered by tests, see codecov report.

martinjrobins (approved):
looks great, thanks @pipliggins. Happy to merge once changelog is added to

TomTranter:
@pipliggins @martinjrobins thanks very much for working on this. I think it will make PyBaMM much more transferable. I checked out the branch and ran into a couple of issues though when trying to couple a loaded model with an experiment. All the examples just solve with a time assuming fixed current.

This code fails

```
import pybamm
experiment = pybamm.Experiment(["Discharge at 1C for 1 hour"])
sim.solve()
sim.save_model("spm_experiment", mesh=False, variables=False)
```
with

NotImplementedError: 
                PyBaMM can only serialise a discretised model.
                Ensure the model has been built (e.g. run `solve()`) before saving.

and this code fails 

```
# create the model
spm_model = pybamm.lithium_ion.SPM()
# set up and discretise ready to solve
geometry = spm_model.default_geometry
param = spm_model.default_parameter_values
param.process_model(spm_model)
mesh = pybamm.Mesh(geometry, spm_model.default_submesh_types, spm_model.default_var_pts)
disc = pybamm.Discretisation(mesh, spm_model.default_spatial_methods)
disc.process_model(spm_model)
# Serialise the spm_model, providing the varaibles and the mesh
spm_model.save_model("example_model", variables=spm_model.variables, mesh=mesh)
# read back in
new_spm_model = pybamm.load_model("example_model.json")
experiment = pybamm.Experiment(["Discharge at 1C for 1 hour"])
new_spm_sim = pybamm.Simulation(new_spm_model, experiment=experiment)
# plot the solution
new_spm_sim.solve()
new_spm_sim.plot()
```
with
AttributeError: 'DomainConcatenation' object has no attribute '_full_mesh'

Is there a plan to implement this or does it require experiments to be serialized. Would be a shame if you couldn't re-use the model with different simulations.

pipliggins:
@TomTranter thanks for raising this! Models coupled to experiments isn't a workflow we'd considered yet, and given the capacity for a single experiment to hold multiple models this will require some work/thinking around how best to structure the serialised file(s), for import particularly.

Having spoken to @martinjrobins, the plan is to merge the current branch (I've added an explicit warning that serialising models coupled to experiments is not yet supported), close this pull request and create a new one focused on expanding the basic model serialisation established here to support the Experiment class.

TomTranter:
I can see why saving a model once coupled to a simulation doesn't work without extra development but the second example I posted where just saving and loading a model before putting into a sim with an experiment feels like it should work. It's a bit confusing having the sim save the model actually. I appreciate why it's been done that way because a sim is changing it's associated objects and Simulation probably contains too much logic rather than simply being a way to group together those components.

pybamm/expression_tree/array.py
```diff
@@ -57,6 +57,30 @@ def __init__(
             name, domain=domain, auxiliary_domains=auxiliary_domains, domains=domains
         )
 
+    @classmethod
+    def _from_json(cls, snippet: dict):
```
Saransh-cpp, line 61: Given that PyBaMM supports Python 3.8+, each file using the newer type hints should import -
```py
from __future__ import annotations
```
at the top. This can also be automated using the `isort` rules in `ruff`.

pybamm/expression_tree/functions.py
```diff
@@ -244,6 +256,25 @@ class SpecificFunction(Function):
     def __init__(self, function, child):
         super().__init__(function, child)
 
+    @classmethod
+    def _from_json(cls, function: Callable, snippet: dict):
```
Saransh-cpp, line 260: The type should be narrowed down if possible - `Callable[[..., ..., ...], ...].` Similarly, the `dict` type should also be narrowed down - `dict[..., ...]`.

Saransh-cpp (commented):
Chiming in to review the type hints 🙂

rtimms:
Thanks @pipliggins, this is a great addition!

My thoughts on serialising models for experiments...

The issue with trying to use a serialised model with an experiment is that the model changes depending on what the operating mode is -- you need to add an extra algebraic constraint to specify power or voltage (at least in the way we have formulated the model). This currently happens in `Simulation.build_for_experiment`, which creates the dict `op_conds_to_built_models`. For a serialised model to be used with _any_ future experiment we would need to create the current/voltage/power control versions of the model at the point of serialisation, so they could all be accessed in the future. Not sure what the API for this would look like - `Simulation` expects to receive a single model, so there'd have to be some new function that sets up a simulation using e.g. a dict mapping the operating mode to the correct model.

An alternative approach would be to create a version of the model where the control is always implemented as an extra algebraic constraint, and then have an option in the simulation that uses this single model instead of creating a new model for each distinct operating mode. The control function would be something like `(I_app - I_var)*I_switch + ...` with similar switches for voltage and power, and everything would be controlled via input parameters. The drawback of this option is that it would make _all_ models DAEs, but maybe that's not a huge deal in an experiment anyway. Would need to do some solver testing to see how performance is affected. 

I don't think either of these options are ideal.

As mentioned in #3530, the logic for setting up experiments in the `Simulation` class is very complex, so perhaps a rehaul of that is in order before we decide how to handle serialising models for use with experiments.

And, of course, this will all break if we allow more generic experiment control in the future (#3530).

docs/source/examples/notebooks/models/saving_models.ipynb
```diff
@@ -0,0 +1,274 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%pip install pybamm -q    # install PyBaMM if it is not installed\n",
+    "import pybamm\n",
+    "\n",
+    "# do the example\n",
+    "dfn_model = pybamm.lithium_ion.DFN()\n",
+    "dfn_sim = pybamm.Simulation(dfn_model)\n",
+    "dfn_sim.solve([0, 3600])\n",
```
rtimms, line 31: This is a very minor comment, but I wonder if the example (and the warning) should encourage the use of `sim.build` rather than `sim.solve`? I have a feeling many users aren't aware you can even call `sim.build`, but it seems like you might want to create a model and save it without actually running a simulation. 

I can imagine a workflow where people end up just doing a dummy solve for a short time before saving a model as they think they need to solve it first.

pipliggins:
> Chiming in to review the type hints 🙂
> 
> Edit: this can be ignored, given that the main motive of this PR was not to add type hints

Thanks @Saransh-cpp - I'll leave these for this PR but as I'm doing type hints elsewhere in #3497 this is useful!

TomTranter:
> Thanks @pipliggins, this is a great addition!
> 
> My thoughts on serialising models for experiments...
> 
> The issue with trying to use a serialised model with an experiment is that the model changes depending on what the operating mode is -- you need to add an extra algebraic constraint to specify power or voltage (at least in the way we have formulated the model). This currently happens in `Simulation.build_for_experiment`, which creates the dict `op_conds_to_built_models`. For a serialised model to be used with _any_ future experiment we would need to create the current/voltage/power control versions of the model at the point of serialisation, so they could all be accessed in the future. Not sure what the API for this would look like - `Simulation` expects to receive a single model, so there'd have to be some new function that sets up a simulation using e.g. a dict mapping the operating mode to the correct model.
> 
> An alternative approach would be to create a version of the model where the control is always implemented as an extra algebraic constraint, and then have an option in the simulation that uses this single model instead of creating a new model for each distinct operating mode. The control function would be something like `(I_app - I_var)*I_switch + ...` with similar switches for voltage and power, and everything would be controlled via input parameters. The drawback of this option is that it would make _all_ models DAEs, but maybe that's not a huge deal in an experiment anyway. Would need to do some solver testing to see how performance is affected.
> 
> I don't think either of these options are ideal.
> 
> As mentioned in #3530, the logic for setting up experiments in the `Simulation` class is very complex, so perhaps a rehaul of that is in order before we decide how to handle serialising models for use with experiments.
> 
> And, of course, this will all break if we allow more generic experiment control in the future (#3530).

I like option 2 better. The file size for a DFN with variables is already 13MB. Having lots of these will quickly add up

pipliggins:
Thanks for your thoughts @TomTranter @rtimms - I would tend to agree that something like option 2 would be better. The size/number of files that could be written out if we ended up trying to serialise every model permutation would get very unwieldy pretty fast. The downside is that it probably has a higher overhead to implement.

As already discussed, the issue with going between single model <-> experiment is that the structure of the built model changes. The reason for forcing the model to be discretised before serialising is ultimately to reduce the complexity of what is being written out, and make it more portable to other solvers; the downside is that a once-serialised model cannot be edited (for example to add in experimental constraints) once it's read back in, as the original RHS/algebraic... etc are lost. The with/without experiment options also follow pretty different flows through the Simulation class to set up for solving. If there are already plans afoot to reorganize this somewhat, possibly associated with #3530, my suggestion would be to put a pin in the integration of the Experiment options with serialisation until those changes have been made - hopefully with an eye on making the with/without experiment workflows more integrated - and in the meantime merging the current PR as a serialisation V1.

@martinjrobins what do you think?

martinjrobins:
I don't think we should do serialisation of model + experiment in this PR, the scope here was to serialize a single already built model, which is already complicated enough, and clearly there needs to be more discussion on how exactly we are going to support the serialisation of model + experiment.

It would be great if we could reformulate the models such that we could have a single model for all experiment steps! That would certainly make it simpler if the solve time wasn't affected too much. Definitely something to try out.

I take @TomTranter point about the sim being the object to save the model. There is a `BaseModel.save_model` function which is the natural place to put it, but since users generally interact with pybamm models via `Simulation` we wanted to add one there as well (which just calls the `save_model` function on the build model)

@rtimms @TomTranter, should we move the discussion on serialisation of model + experiment to a separate issue? We can then merge this in and use it as a basis for future work.

PR #3406, opened 2023-10-04
tests: Consistent random seed in benchmarks

### LINKED ISSUE #3119: fix initial seeds for benchmarks
### Description

unit tests fix the initial seeds for reliability, this feature request proposes doing the same for the benchmarks

### Motivation

reliability for the benchmarks an their timing

### Possible Implementation

_No response_

### Additional context

_No response_

kratman:
# Description

A set_random_seed() function was added and used inside the benchmarks. This will make them run with the same seed every time.

Additional work was done to cleanup some of the files used in the benchmarks.

Fixes #3119

## Type of change

Adds a random seed function.

# Key checklist:

## Further checks:


kratman:
Since np.random.seed() is not thread safe, I am going to leave this as a draft until I do more testing. I want to check for any parallelization of the benchmarks/tests to make sure that the seeding works as expected.

kratman:
It looks like asv only runs in parallel for the build step. So I think this change is safe. Failures in on windows appear to be unrelated. The windows tests failed because of an issue with the nox installation.

benchmarks/memory_unit_benchmarks.py
```diff
@@ -1,8 +1,15 @@
 import pybamm
+from benchmarks.benchmark_utils import set_random_seed
 import numpy as np
 
 
 class MemCreateExpression:
+    R: pybamm.Parameter
+    model: pybamm.BaseModel
```
brosaplanella, line 8: Out of curiosity, what does this syntax do? Never seen it before

brosaplanella (approved):
Looks good, thanks!

benchmarks/memory_unit_benchmarks.py
kratman: It is a way to declare member variables without an init function.
```
variableName: type
```
It also enforces the typing of the variable.

I used it here because the benchmark classes were adding member variables inside other functions, which gives a static analysis warning. It is generally best practice to declare all member variables either in the init() method or in the class like I did here. The main reasoning is to avoid the variables being undefined or assuming that member functions are called in a certain order. Not a major safety concern in these small classes with a very specific purpose, but I still prefer to clean it up if I see it.

kratman: I should probably add that this syntax is just general typing of python variables:
```python
def some_function(a: str, b: int) -> float:
```
It lets you declare the type of variables with ":" after the variable name and return types with "->" after the function. Typing support is a bit of a pain in Python, but it is useful sometimes.

Saransh-cpp: We do need type hints across PyBaMM, maybe a small GSoC project 🙂

PR #3408, opened 2023-10-04
fix: Implement Automatic Activation of Virtual Environment in the dev nox session

### LINKED ISSUE #3388: Make it possible to activate the venv in the `dev` `nox` session
In `noxfile.py` we define a `dev` session that installs PyBaMM's dependencies for use by developers and contributors. It is possible to go one step further: automatically activate the virtual environment it creates with some logic based on the OS or the platform. See https://nox.thea.codes/en/stable/cookbook.html#instant-dev-environment for more information.

akhilender-bongirwar:
# Description

- Implement automatic activation of the virtual environment in the 'dev' Nox session
- Enhance the 'dev' Nox session defined in `noxfile.py`
- Simplify the development setup process for contributors and developers

This commit enhances the 'dev' Nox session to automatically activate the virtual environment when running 'nox -s dev'.

Fixes #3388

## Type of change

# Key checklist:

## Further checks:


noxfile.py
```diff
@@ -117,7 +118,11 @@ def set_dev(session):
     """Install PyBaMM in editable mode."""
     set_environment_variables(PYBAMM_ENV, session=session)
     envbindir = session.bin
-    session.install("-e", ".[all]", silent=False)
+    venv_dir = pathlib.Path('./venv').resolve()
+    session.install("virtualenv")
+    session.run("virtualenv", os.fsdecode(venv_dir), silent=True)
+    python = os.fsdecode(venv_dir.joinpath("bin/python"))
+    session.run(python, "-m", "pip", "install", "-e", ".[all]", silent=False)
```
agriyakhetarpal, line 125: The purpose of the `dev` session is also to set up an instant developer environment besides getting an editable installation ready. This session should also install the `dev` dependencies. Also, we should use `external=True` to silence the warning about the Python interpreter being outside `nox`'s control.

agriyakhetarpal: Another thing that can be done here is to install the `[jax,odes]` sets of extras for macOS and Linux

noxfile.py
```diff
@@ -117,7 +118,11 @@ def set_dev(session):
...
     session.install("cmake", silent=False)
     if sys.platform == "linux" or sys.platform == "darwin":
         session.run(
```
agriyakhetarpal, line 128: @arjxn-py, do you remember why we added this command? It doesn't seem to make any changes for me

agriyakhetarpal (commented):
Thanks, @akhilender-bongirwar, looks off to a good start! I tested this locally and it doesn't look like there is a method to activate the virtual environment from _inside_ a `nox` session (we should ask users to do so like they currently do). We can target some other relevant improvements, some comments below

noxfile.py
arjxn-py: This command is setting up the environment variable `LD_LIBRARY_PATH` which is then appended to a file named `activate` located in the `.nox/dev/bin`.
Basically the significance this command holds is to activate the `venv` with `source .nox/dev/bin/activate` in GNU/Linux & MacOS & with `.nox\dev\Scripts\activate.bat` in Windows.

akhilender-bongirwar: @agriyakhetarpal @arjxn-py,  Should I change anything over here ?

noxfile.py
```diff
@@ -117,8 +119,14 @@ def set_dev(session):
     """Install PyBaMM in editable mode."""
     set_environment_variables(PYBAMM_ENV, session=session)
     envbindir = session.bin
-    session.install("-e", ".[all]", silent=False)
+    session.install("virtualenv")
+    session.run("virtualenv", os.fsdecode(VENV_DIR), silent=True)
+    python = os.fsdecode(VENV_DIR.joinpath("bin/python"))
+    session.run(python, "-m", "pip", "install", "-e", ".[all]", silent=False)
     session.install("cmake", silent=False)
+    session.install("-e", ".[dev]", external=True)
```
agriyakhetarpal, line 127: We should install `cmake` before we install the other dependencies, otherwise PyBaMM will build itself without its compiled components. Also, the extra dependencies can be installed at once (i.e., `[all,dev]`) to not build the wheels more than once

noxfile.py
```diff
@@ -117,8 +119,14 @@ def set_dev(session):
...
+    if session.platform.startswith("macos") or session.platform.startswith("linux"):
```
agriyakhetarpal, line 128: Could you use `sys.platform` here instead, just to be consistent with how the operating system is detected at other locations in the file?

noxfile.py
agriyakhetarpal: Given that the `pybamm-requires` session would have been run before `dev`, you can remove this command. This is because SUNDIALS would have already been set up then (`scikits.odes` would now be installed within the session because we have environment variables that point to the installation of SUNDIALS)

agriyakhetarpal (commented):
I think this is almost ready. Could you update the documentation for the source installation [in this file](https://github.com/pybamm-team/PyBaMM/blob/develop/docs/source/user_guide/installation/install-from-source.rst), highlighting the change and updating the new location of the virtual environment directory (plus how to activate it)? I left some more comments below

noxfile.py
```diff
@@ -116,18 +118,13 @@ def run_scripts(session):
...
-    if sys.platform == "linux" or sys.platform == "darwin":
-        session.run(
-            "echo",
-            "export",
-            f"LD_LIBRARY_PATH={PYBAMM_ENV['LD_LIBRARY_PATH']}",
-            ">>",
-            f"{envbindir}/activate",
-            external=True,  # silence warning about echo being an external command
-        )
+    session.install("virtualenv","cmake")
```
agriyakhetarpal, line 121:
```suggestion
    session.install("virtualenv", "cmake")
```

noxfile.py
```diff
@@ -116,18 +118,13 @@ def run_scripts(session):
...
+    session.run("virtualenv", os.fsdecode(VENV_DIR), silent=True)
+    python = os.fsdecode(VENV_DIR.joinpath("bin/python"))
+    session.run(python, "-m", "pip", "install", "-e", ".[all,dev]", silent=False)
+    if sys.platform == "linux" or sys.platform == "macos":
```
agriyakhetarpal, line 125:
```suggestion
    if sys.platform == "linux" or sys.platform == "darwin":
```

agriyakhetarpal (changes requested):
You also need to update the documentation for the source installation page as highlighted above

akhilender-bongirwar:
> You also need to update the documentation for the source installation page as highlighted above

Hey @agriyakhetarpal, could you please suggest what exactly should I add in the documentation

agriyakhetarpal:
You need to update [this section](https://docs.pybamm.org/en/latest/source/user_guide/installation/install-from-source.html#using-nox-recommended). We need to 1. update the location of the virtual environment—should now be `venv/` instead of `.nox/dev/` wherever it is referenced, and 2. update the command used to activate the virtual environment—this should be `source venv/bin/activate` instead of `.nox/dev/bin/activate`, and so on

akhilender-bongirwar:
Thank you @agriyakhetarpal : ).

akhilender-bongirwar:
@agriyakhetarpal, please review the changes : ).

agriyakhetarpal:
The Files interface currently shows that that the entirety of the lines in the two files were modified. Did you change the Git settings about your line endings on your platform recently, or otherwise push the last commit from another platform that might have a different line ending scheme compared to the commits before thag? We would have to normalise it to properly display the modified lines. Could you drop your most recent commit and force push? This answer might help: https://stackoverflow.com/a/30757723

akhilender-bongirwar:
@agriyakhetarpal , I actually pushed the commit from WSL instead of using Bash. I reverted the last the last commit.

noxfile.py
```diff
@@ -116,18 +118,13 @@ def run_scripts(session):
 def set_dev(session):
     """Install PyBaMM in editable mode."""
     set_environment_variables(PYBAMM_ENV, session=session)
-    envbindir = session.bin
-    session.install("-e", ".[all]", silent=False)
-    session.install("cmake", silent=False)
+    session.install("virtualenv", "cmake")
+    session.run("virtualenv", os.fsdecode(VENV_DIR), silent=True)
+    python = os.fsdecode(VENV_DIR.joinpath("bin/python"))
+    session.run(python, "-m", "pip", "install", "-e", ".[all,dev]", silent=False)
```
agriyakhetarpal, line 124:
```suggestion
    session.run(python, "-m", "pip", "install", "-e", ".[all,dev]", external=True)
```

We don't require `silent=False` for `session.run()` because it is the default behaviour, could you add `external=True` instead?

noxfile.py
```diff
@@ -116,18 +118,13 @@ def run_scripts(session):
...
     if sys.platform == "linux" or sys.platform == "darwin":
-        session.run(
-            "echo",
-            "export",
-            f"LD_LIBRARY_PATH={PYBAMM_ENV['LD_LIBRARY_PATH']}",
-            ">>",
-            f"{envbindir}/activate",
-            external=True,  # silence warning about echo being an external command
-        )
+        session.run(python, "-m", "pip", "install", ".[jax,odes]", silent=False)
```
agriyakhetarpal, line 126: Same as above

agriyakhetarpal (approved):
Looks good to me, thanks for working on this, @akhilender-bongirwar! Tested this locally and it works without issues. I will request a review from @Saransh-cpp in case there are any additional changes to be made and also to get a workflow run triggered here

akhilender-bongirwar:
> Looks good to me, thanks for working on this, @akhilender-bongirwar! Tested this locally and it works without issues. I will request a review from @Saransh-cpp in case there are any additional changes to be made and also to get a workflow run triggered here

Thank you, @agriyakhetarpal, for allowing me to work under your guidance for this issue. Although I made a few mistakes in the beginning, it has been great working together.

noxfile.py
```diff
@@ -116,18 +118,13 @@ def run_scripts(session):
...
-        session.run(
-            "echo",
-            "export",
-            f"LD_LIBRARY_PATH={PYBAMM_ENV['LD_LIBRARY_PATH']}",
-            ">>",
-            f"{envbindir}/activate",
-            external=True,  # silence warning about echo being an external command
-        )
+        session.run(python, "-m", "pip", "install", ".[jax,odes]", external=True)
+
```
Saransh-cpp, line 127:
```suggestion
    else:
        session.run(python, "-m", "pip", "install", "-e", ".[all,dev]", external=True)
```

Saransh-cpp (changes requested):
Thanks for working on this, @akhilender-bongirwar! I think we'll still need to set the env variables before installing `scikits.odes` (the `dev` session fails for me locally) - @arjxn-py @agriyakhetarpal.

Also, we should use an `if-else` block to save time by installing pybamm only once -

agriyakhetarpal:
Okay, that's strange because it works well on my end (I checked and purged the cached wheels so that it builds a wheel for `scikits.odes` again). The `dev` session is meant to be run after `pybamm-requires` so that the SUNDIALS installation can be found (the `set_environment_variables` function is taking care of the SUNDIALS_INST environment variable)

agriyakhetarpal:
Note: the earlier export command was incorrect anyway because it didn't change anything in the activation script, so if we want to keep that we should modify it to use a command similar to what `pybamm_install_odes` does

It might also be a discrepancy in the Python version. This newer dev session will create a virtualenv with the version of Python with which it was installed

Saransh-cpp:
Weird. I'll suggest that we install `scikits.odes` only on Linux in this PR, given that we install `scikits.odes` only on Linux in the other `nox` sessions (except for the `coverage` session, don't really know why). #3417 can then modify the `noxfile` appropriately to support MacOS.

agriyakhetarpal:
> Weird. I'll suggest that we install scikits.odes only on Linux in this PR

No worries, happy to help debug your installation offline 😄

> except for the coverage session, don't really know why

This is because we don't compile the IDAKLU component for macOS in CI, because we are running `pybamm-requires` only on Linux. I have fixed that in #3301, but as discussed in the developer meeting we'll be merging that one after 23.9 non-RC is out, and that could take some time (the wheels have to be fixed too). @arjxn-py, please feel free to copy the relevant changes from there if you want to or create a new PR because that one is slightly big (but contains some needed changes).

akhilender-bongirwar:
@agriyakhetarpal, should I still make any new changes or is the code working fine ?

agriyakhetarpal:
For now, you may install `[odes]` only on Linux and also resolve Saransh's comments. Happy to approve then

noxfile.py
```diff
@@ -116,18 +118,13 @@ def run_scripts(session):
 def set_dev(session):
     """Install PyBaMM in editable mode."""
     set_environment_variables(PYBAMM_ENV, session=session)
-    envbindir = session.bin
-    session.install("-e", ".[all]", silent=False)
-    session.install("cmake", silent=False)
+    session.install("virtualenv", "cmake")
+    session.run("virtualenv", os.fsdecode(VENV_DIR), silent=True)
+    python = os.fsdecode(VENV_DIR.joinpath("bin/python"))
     if sys.platform == "linux" or sys.platform == "darwin":
```
agriyakhetarpal, line 124: This should be just "linux"

agriyakhetarpal (changes requested):

agriyakhetarpal (approved):
Everything looks good to me now. Thanks for working on this, @akhilender-bongirwar!

Saransh-cpp (approved):

PR #2973, opened 2023-05-20
Test scikit.odes installation nightly

arjxn-py:
# Description

- `run_periodic_tests.yml` will now run test for `pybamm_install_odes` only on GNU/Linux

Fixes # (issue)

## Type of change

- [ ] New feature (non-breaking change which adds functionality)
- [x] Optimization (back-end change that speeds up the code)
- [ ] Bug fix (non-breaking change which fixes an issue)

# Key checklist:

- [ ] No style issues: `$ pre-commit run` (see [CONTRIBUTING.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CONTRIBUTING.md#installing-and-using-pre-commit) for how to set this up to run automatically when committing locally, in just two lines of code)
- [ ] All tests pass: `$ python run-tests.py --all`
- [ ] The documentation builds: `$ python run-tests.py --doctest`

You can run unit and doctests together at once, using `$ python run-tests.py --quick`.

## Further checks:

- [ ] Code is commented, particularly in hard-to-understand areas
- [ ] Tests added that prove fix is effective or that feature works

PR #3417, opened 2023-10-06
Extend `pybamm_install_odes` to include support for macOS systems

### LINKED ISSUE #3404: Extend `pybamm_install_odes` to include support for macOS systems
The `pybamm/install_odes.py` script, i.e., the `pybamm_install_odes` entry point currently installs the build-time requirements (SUNDIALS), sets relevant environment variables, and `scikits.odes` for use by the optional solver. Considering that users can also use `scikits.odes` on macOS by pointing to a SUNDIALS installation using the `SUNDIALS_INST` environment variable, we can add suitable modifications to this script to make it work on all UNIX-based systems (macOS and Linux).

Some reference can be used from the `install_KLU_Sundials.py` which currently works for both macOS and Linux. However, a difference to be noted is that `install_odes.py` does not compile the KLU module, whereas the aforementioned script does.

This will help simplify the installation of `scikits.odes` and enable doing so without the usage of a `nox` session, and can then be tested in CI.

arjxn-py:
# Description

Fixes #3404 
## Type of change

# Key checklist:

## Further checks:


.github/workflows/test_on_push.yml
```diff
@@ -299,6 +299,10 @@ jobs:
           pip install --upgrade pip wheel setuptools nox
           pip install -e .[all,docs]
 
+      - name: Test pybamm_install_odes on MacOS (for only this PR)
+        if: matrix.os == 'macos-latest'
+        run: pybamm_install_odes
```
agriyakhetarpal, line 304: Note: this step won't run at all currently because the job that it is in runs only on Ubuntu. Adding a parallel job (in a matrix for GNU/Linux and macOS) that needs the style tests would be better because we would want to use a fresh runner

agriyakhetarpal (changes requested):

.github/workflows/test_on_push.yml
arjxn-py: My bad, I added it to wrong place thanks.

agriyakhetarpal: It is still a step in another job, could you create a different job with steps that check out the repository but don't run `pybamm-requires` beforehand?

agriyakhetarpal:
Seems like we were testing `pybamm_install_odes` earlier (#2973) but that was back when we were using `tox`, do you remember why we removed it? I can't seem to find it anywhere. Maybe it was because we decided not to have a `nox` equivalent of `tox -e odes` or we just forgot to add one

arjxn-py:
> Seems like we were testing `pybamm_install_odes` earlier (#2973) but that was back when we were using `tox`, do you remember why we removed it? I can't seem to find it anywhere. Maybe it was because we decided not to have a `nox` equivalent of `tox -e odes` or we just forgot to add one

I don't remember exactly but it might be because of we wanted to cut time in CI or just might have forgotten to have a session for odes.

.github/workflows/test_on_push.yml
```diff
@@ -177,6 +177,49 @@ jobs:
       - name: Upload coverage report
         uses: codecov/codecov-action@v3.1.4
 
+  test_install_odes:
+    needs: style
+    runs-on: macos-latest
+    strategy:
```
agriyakhetarpal, line 183:
```suggestion
    strategy:
        matrix:
            os: [ubuntu-latest, macos-latest]
            python-version: ["3.8", "3.9", "3.10", "3.11"]
```

We can keep it to just Python 3.11 for GNU/Linux and macOS later after we add this to the scheduled tests workflow at the time of completion

.github/workflows/test_on_push.yml
```diff
@@ -177,6 +177,49 @@ jobs:
...
+          # Homebrew environment variables
+          HOMEBREW_NO_INSTALL_CLEANUP: 1
+          HOMEBREW_NO_AUTO_UPDATE: 1
+          HOMEBREW_NO_COLOR: 1
+          # Speed up CI
+          NONINTERACTIVE: 1
+        run: |
+          brew analytics off
+          brew update
+          brew install graphviz openblas
```
agriyakhetarpal, line 202:
```suggestion
          brew install openblas
```

.github/workflows/test_on_push.yml
```diff
@@ -177,6 +177,49 @@ jobs:
...
+
+      - name: Set up Python ${{ matrix.python-version }}
+        id: setup-python
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
```
agriyakhetarpal, line 208: It should pick up the Python version correctly after adding the changes requested above. The matrix did not have a Python version array earlier so it picked up the system Python 2 iiuc, which did not have `pathlib`

.github/workflows/test_on_push.yml
```diff
@@ -177,6 +177,49 @@ jobs:
...
+          cache: 'pip'
+          cache-dependency-path: setup.py
+
+      - name: Install PyBaMM dependencies
+        run: |
+          pip install --upgrade pip wheel setuptools nox
+          pip install -e .[all,docs]
```
agriyakhetarpal, line 215:
```suggestion
      - name: Install PyBaMM dependencies
        run: |
          pip install --upgrade pip wheel setuptools nox
          pip install -e .[all]
```

can be sped up a bit, no `docs` dependencies required

.github/workflows/test_on_push.yml
```diff
@@ -177,6 +177,49 @@ jobs:
...
+
+      - name: Test pybamm_install_odes on MacOS
+        if: matrix.os == 'macos-latest'
+        run: |
+          pip install wget
```
agriyakhetarpal, line 220:
```suggestion
          pip install wget cmake
```

agriyakhetarpal (changes requested):

agriyakhetarpal:
Feel free to copy changes from #3301 to fix the inability to find a Fortran compiler (probably just needs a `brew reinstall gcc` because it comes bundled with `gcc` in most cases)

.github/workflows/test_on_push.yml
```diff
@@ -177,6 +177,52 @@ jobs:
...
 
+  test_install_odes:
+    needs: style
+    runs-on: macos-latest
+    strategy:
+        matrix:
+            os: [ubuntu-latest, macos-latest]
+            python-version: ["3.8", "3.9", "3.10", "3.11"]
+        fail-fast: false
+    name: Test pybamm_install_odes on MacOS
```
agriyakhetarpal, line 188:
```suggestion
    name: Test pybamm_install_odes on ${{ matrix.os }}
```

.github/workflows/test_on_push.yml
```diff
@@ -177,6 +177,52 @@ jobs:
...
+          cache: 'pip'
+          cache-dependency-path: setup.py
+
+      - name: Install PyBaMM dependencies
+        run: |
+          pip install --upgrade pip wheel setuptools nox
+          pip install -e .[all]
+
+      - name: Test pybamm_install_odes on MacOS
+        if: matrix.os == 'macos-latest'
```
agriyakhetarpal, line 221:
```suggestion
```

.github/workflows/test_on_push.yml
```diff
@@ -177,6 +177,52 @@ jobs:
...
+          python-version: ${{ matrix.python-version }}
+          cache: 'pip'
+          cache-dependency-path: setup.py
+
+      - name: Install PyBaMM dependencies
+        run: |
+          pip install --upgrade pip wheel setuptools nox
+          pip install -e .[all]
+
+      - name: Test pybamm_install_odes on MacOS
```
agriyakhetarpal, line 220:
```suggestion
      - name: Test pybamm_install_odes on ${{ matrix.os }}
```

.github/workflows/test_on_push.yml
```diff
@@ -177,6 +177,52 @@ jobs:
...
+          # Homebrew environment variables
+          HOMEBREW_NO_INSTALL_CLEANUP: 1
+          HOMEBREW_NO_AUTO_UPDATE: 1
+          HOMEBREW_NO_COLOR: 1
+          # Speed up CI
+          NONINTERACTIVE: 1
+        run: |
+          brew analytics off
+          brew update
+          brew install openblas
```
agriyakhetarpal, line 205:
```suggestion
          brew install openblas
          brew reinstall gcc gfortran
```

`install` instead of `reinstall` should work too, but just in case

agriyakhetarpal (changes requested):

arjxn-py:
Thanks @agriyakhetarpal, waiting on tests to pass the initial steps and then fail while installing odes to observe logs.

agriyakhetarpal:
It looks like it worked without any additional configuration (just like it did for me locally). Is there an error in specific that we should reproduce to see if it is robust enough?

We could refactor the script a bit and add some additional configuration (since this entry point is for user installations, it would be wise to add `.zshrc` alongside `.bashrc` in the `update_LD_LIBRARY_PATH` function because it is enabled by default on newer macOS systems). Some other stuff I identified that we can do:

1. We ask users to run the script and that it works out of the box, but it would fail on not finding `wget` and `cmake`. We could install both of them with `pip` inside a subprocess or instead throw more helpful errors
2. Use f-strings in the script since they are more efficient
3. `scikits_odes_setup.log` is not git-ignored currently
4. add a `nox` session that installs `cmake` and `wget` and wraps `pybamm_install_odes` (actually, this might not be needed because users don't get or need `nox` installed as a dependency unless they install the `dev` dependencies). Developers can already install SUNDIALS but with KLU enabled using `pybamm-requires`.
5. Update the documentation about macOS support for this entry point in the optional solvers sectionn and other areas (and maybe make the script exit early with a helpful error if on a Windows system because we haven't done so yet). We can update the CHANGELOG about this too since this is rather a user-facing change.

agriyakhetarpal:
We should also run `pip cache purge` in the installation step so that the runners don't end up using the cached wheels between runs and/or remove the cache configuration from the `setup-python` step

docs/source/user_guide/installation/GNU-linux.rst
```diff
@@ -125,9 +125,11 @@ Currently, only GNU/Linux and macOS are supported.
 
    .. code:: bash
 
-	  pip install scikits.odes
+      brew install openblas
+      pybamm_install_odes
 
-   Assuming that SUNDIALS was installed as described :ref:`above<user-install-label>`.
+   The ``pybamm_install_odes`` command is installed with PyBaMM. It automatically downloads and installs the SUNDIALS library on your
+   system (under ``~/.local``), before installing ``scikits.odes``. (Alternatively, one can install SUNDIALS without this script and run ``pip install pybamm[odes]`` to install ``pybamm`` with ``scikits.odes``.)
```
agriyakhetarpal, line 132: We should also instruct users on how to avoid failures while installing `scikits.odes` without this script. We were mentioning "Assuming that SUNDIALS was installed as described above", but if someone runs `pip install pybamm[odes]` now, it would still fail even if SUNDIALS is installed on the system because the SUNDIALS_INST environment variable would not be set. Above in the same file, we recommend running `brew install sundials` before `pip install pybamm`, so we can use the location Homebrew installs it to and ask users to set SUNDIALS_INST to that location. On an ARM-based machine for my case it is installed under `/opt/homebrew/Cellar/sundials/6.6.0`, but I think it would be under `usr/local/` on an Intel-based macOS machine. We could either check that and provide that ourselves or we could ask them to run `brew info sundials` to find the location for their case (which is how I myself found it). Note: this also shows that Homebrew and `pybamm_install_odes` install SUNDIALS under different locations (which is good because we don't want to interfere with Homebrew's files).

Perhaps adding a link to https://scikits-odes.readthedocs.io/en/stable/installation.html would also be helpful. Otherwise, we should _recommend_ the use of `pybamm_install_odes`. Please feel free to resolve this comment as necessary, though I will tag @Saransh-cpp for his input on this too.

pybamm/install_odes.py
```diff
@@ -8,28 +8,32 @@
 
 from pybamm.util import root_dir
 
-try:
-    # wget module is required to download SUNDIALS or SuiteSparse.
-    import wget
+def check_platform():
+    if sys.platform == "win32":
+        raise Exception("pybamm_install_odes is not supported on Windows.")
```
agriyakhetarpal, line 13: Do we really need a separate function to do this considering that this file is a script meant to be executed rather than an importable file, and will therefore run from top to bottom?

pybamm/install_odes.py
```diff
@@ -8,28 +8,32 @@
...
-except ModuleNotFoundError:
-    NO_WGET = True
+check_platform()
 
+def install_required_module(module):
+    try:
+        __import__(module)
+    except ModuleNotFoundError:
+        print(f"{module} module not found. Installing {module}...")
+        subprocess.run(["pip", "install", module], check=True)
```
agriyakhetarpal, line 22: Curious about the `__import__` syntax here since I have seen that this is offered and used more for programmatic ways of importing modules as objects. Could you try using the `importlib` module for this instead? I found this thread insightful: https://stackoverflow.com/questions/28231738/import-vs-import-vs-importlib-import-module

pybamm/install_odes.py
```diff
@@ -41,10 +45,7 @@ def install_sundials(download_dir, install_dir):
         raise RuntimeError("CMake must be installed to build SUNDIALS.")
 
     url = (
-        "https://github.com/LLNL/"
-        + "sundials/releases/download/v{}/sundials-{}.tar.gz".format(
-            sundials_version, sundials_version
-        )
+        f"https://github.com/LLNL/sundials/releases/download/v{sundials_version}/sundials-{sundials_version}.tar.gz"
```
agriyakhetarpal, line 48: An improvement could be that instead of using `sundials_version`, we could provide it as a constant `SUNDIALS_VERSION` at the top of the file (would make it easier to access and edit)

pybamm/install_odes.py
```diff
@@ -75,40 +76,36 @@ def install_sundials(download_dir, install_dir):
...
-    if os.getenv("LD_LIBRARY_PATH") and "{}/lib".format(install_dir) in os.getenv(
-        "LD_LIBRARY_PATH"
-    ):
-        print("{}/lib was found in LD_LIBRARY_PATH.".format(install_dir))
-        print("--> Not updating venv activate or .bashrc scripts")
+        if sys.platform == "linux":
+            script_path = os.path.join(os.environ.get("HOME"), ".bashrc")
+        if sys.platform == "darwin":
+            script_path = os.path.join(os.environ.get("HOME"), ".zshrc")
+
```
agriyakhetarpal, line 94: Not sure if this will factor in the cases for those who don't use `zsh` as the default shell. Is there an elegant way to detect the default terminal shell programmatically and append the installation directory to LD_LIBRARY_PATH in a more Pythonic way (which is also permanent)? I haven't dived into the internals of the `os` module that deep but I think it should be possible. 

That level of abstraction will clean up some bits of the script which are opening and writing to these shell configuration files. Users will then not be restricted to using `bash` or `zsh`.

agriyakhetarpal (commented):
Thanks, @arjxn-py, looks very good and the tests are all passing too, though I do have some suggestions and questions below

.github/workflows/test_on_push.yml
```diff
@@ -177,6 +177,53 @@ jobs:
...
+          brew reinstall gcc gfortran
+
+      - name: Set up Python ${{ matrix.python-version }}
+        id: setup-python
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: 'pip'
+          cache-dependency-path: setup.py
```
agriyakhetarpal, line 214: We don't cache anything in the scheduled tests jobs so we can remove this too
```suggestion
```

pybamm/install_odes.py
arjxn-py: I've unwrapped it and eliminated the function. Thanks.

pybamm/install_odes.py
arjxn-py: Done 👍🏻

pybamm/install_odes.py
arjxn-py: Done

pybamm/install_odes.py
arjxn-py: Detecting default shell type via `os.environ` now.

pybamm/install_odes.py
```diff
@@ -5,46 +5,46 @@
...
-except ModuleNotFoundError:
-    NO_WGET = True
+SUNDIALS_VERSION = "6.5.0"
 
+def install_required_module(module):
+    try:
+        import_module(module)
+    except ModuleNotFoundError:
+        print(f"{module} module not found. Installing {module}...")
+        subprocess.run(["pip", "install", module], check=True)
```
agriyakhetarpal, line 22: I know I requested adding this earlier, but I think modifying someone's virtual environment in such an impromptu way seems weird. I suggest we should let go of this function and throw an error on `wget` and `CMake` not being installed instead so that users can do that on their own (but add to the documentation that the script requires them). I am working on a branch locally where `urllib` from the standard library will be used to download SUNDIALS (we can let go of `wget` then). Is it fine if I push the same changes to your branch after the work on your end is complete?

P.S. The command would have failed anyway if someone would have not had `pip` but `pip3` as their shell-completion or if they used `python3 -m pip` instead of `pip`.

pybamm/install_odes.py
```diff
@@ -75,38 +75,36 @@ def install_sundials(download_dir, install_dir):
...
 
     venv_path = os.environ.get("VIRTUAL_ENV")
     if venv_path:
         script_path = os.path.join(venv_path, "bin/activate")
     else:
-        script_path = os.path.join(os.environ.get("HOME"), ".bashrc")
+        if 'BASH' in os.environ:
+            script_path = os.path.join(os.environ.get("HOME"), ".bashrc")
+        if 'ZSH' in os.environ:
+            script_path = os.path.join(os.environ.get("HOME"), ".zshrc")
```
agriyakhetarpal, line 92: I think this doesn't work either – I don't have these environment variables set on my machine. We should probably look for both of these rc files in the HOME directory and make changes to whichever file is found (both files if both are found)

agriyakhetarpal (commented):
It has been some time since this PR has been active – I know this is still a draft, but I have some new comments that can be addressed before any further changes are made.

pybamm/install_odes.py
arjxn-py: > Is it fine if I push the same changes to your branch after the work on your end is complete?

Yeah, no worries I'd let you know once i finish up with my changes

pybamm/install_odes.py
arjxn-py: Ahh I see, looking accordingly into this.

arjxn-py: I hope now its alright.

docs/source/user_guide/installation/GNU-linux.rst
```diff
@@ -133,9 +131,29 @@ order to use the wrapped SUNDIALS ODE and DAE
...
+
+Ensure that the path matches the installation location on your system. You can verify the installation location by running:
+
+.. code:: bash
+
+   brew info sundials
+
+Look for the installation path, and use that path to set the ``SUNDIALS_INST`` variable.
+
+Note: The location where Homebrew installs SUNDIALS might vary based on the system architecture (ARM or Intel). Adjust the path in the ``export SUNDIALS_INST`` command accordingly.
```
agriyakhetarpal, line 154: Can we link the `scikits.odes` installation docs here as well (because it requires other things like C compilers and `gfortran` that may not be present on someone's system)?

arjxn-py: Sure, will do that.

pybamm/install_odes.py
agriyakhetarpal: Yes it is, thanks!

agriyakhetarpal:
I will test the changes locally in a moment after removing my local installation of SUNDIALS

.github/workflows/run_periodic_tests.yml
```diff
@@ -153,3 +153,48 @@ jobs:
...
+        env:
+          # Homebrew environment variables
+          HOMEBREW_NO_INSTALL_CLEANUP: 1
+          HOMEBREW_NO_AUTO_UPDATE: 1
+          HOMEBREW_NO_COLOR: 1
+          # Speed up CI
+          NONINTERACTIVE: 1
+        run: |
+          brew analytics off
+          brew update
```
agriyakhetarpal, line 181:
```suggestion
          brew update
```

This was failing on GitHub Actions runners for some reason, better to remove

docs/source/user_guide/installation/GNU-linux.rst
```diff
@@ -102,17 +102,15 @@ For an introduction to virtual environments, see
 Optional - scikits.odes solver
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Users can install `scikits.odes <https://github.com/bmcage/odes>`__ in
-order to use the wrapped SUNDIALS ODE and DAE
-`solvers <https://docs.pybamm.org/en/latest/source/api/solvers/scikits_solvers.html>`__.
+Users can install `scikits.odes <https://github.com/bmcage/odes>`__ to utilize the wrapped SUNDIALS ODE and DAE `solvers <https://docs.pybamm.org/en/latest/source/api/solvers/scikits_solvers.html>`__ in PyBaMM.
```
agriyakhetarpal, line 105:
```suggestion
Users can install `scikits.odes <https://github.com/bmcage/odes>`__ to utilize its interfaced SUNDIALS ODE and DAE `solvers <https://docs.pybamm.org/en/latest/source/api/solvers/scikits_solvers.html>`__ wrapped in PyBaMM.
```

A bit nitpicky; since the most of the work done by the solver is through the Python bindings that `scikits.odes` provides

docs/source/user_guide/installation/GNU-linux.rst
```diff
@@ -133,9 +131,29 @@ order to use the wrapped SUNDIALS ODE and DAE
 
    .. code:: bash
 
-	  pip install scikits.odes
+      brew install openblas
+      pybamm_install_odes
```
agriyakhetarpal, line 135: Should we add `gcc` and `gfortran` here?

```suggestion
      brew install openblas gcc gfortran
      pybamm_install_odes
```

docs/source/user_guide/installation/GNU-linux.rst
```diff
@@ -121,10 +119,10 @@ order to use the wrapped SUNDIALS ODE and DAE
 
    .. code:: bash
 
-	   apt install libopenblas-dev
-	   pybamm_install_odes
+      apt install libopenblas-dev
+      pybamm_install_odes
 
-   The ``pybamm_install_odes`` command is installed with PyBaMM. It automatically downloads and installs the SUNDIALS library on your
+   The ``pybamm_install_odes`` command, installed with PyBaMM, automatically downloads and installs the SUNDIALS library on your
```
agriyakhetarpal, line 125:
```suggestion
```

This line is present below as well, we might as well keep it in one place

pybamm/install_odes.py
agriyakhetarpal, line 1: Tip: set `CMAKE_BUILD_PARALLEL_LEVEL` as an environment variable in this file (like I have done for resolving #3646) to make the installation faster!

agriyakhetarpal (approved):
Tested locally without a pip-cached `scikits.odes` wheel and without SUNDIALS installed via Homebrew nor present in `~/.local/` and it works, thanks, @arjxn-py! I left a few comments, feel free to address. 

@Saransh-cpp, I remember you were having problems with getting `scikits.odes` to work on your machine before, could you test this and see if it works?

docs/source/user_guide/installation/GNU-linux.rst
agriyakhetarpal: It might also make sense to rename this file to `gnu-linux-mac.rst` or something btw and fix any broken links, the heading at the top of the file says

```rst
GNU-Linux & MacOS
=================
```

but the file name is misleading for contributors

docs/source/user_guide/installation/GNU-linux.rst
arjxn-py: Ah yes, i'll also fix it's indentation to appear on both tabs.

docs/source/user_guide/installation/GNU-linux.rst
arjxn-py: Yes, better to add.

docs/source/user_guide/installation/GNU-linux.rst
arjxn-py: Sure, thanks for pointing it out.

agriyakhetarpal:
Feel free to add some of the [un-fixable broken links](https://github.com/pybamm-team/PyBaMM/actions/runs/7309935911?pr=3417#summary-19918060269) to the `.lycheeignore` file

arjxn-py:
> Feel free to add some of the [un-fixable broken links](https://github.com/pybamm-team/PyBaMM/actions/runs/7309935911?pr=3417#summary-19918060269) to the `.lycheeignore` file

Most of them are due to file rename which I hope should be fixed expect [this one](https://github.com/LLNL/sundials/releases/download/v%7BSUNDIALS_VERSION%7D/sundials-%7BSUNDIALS_VERSION%7D.tar.gz), let me know how should i go with this.

agriyakhetarpal:
> Most of them are due to file rename which I hope should be fixed except [this one](https://github.com/LLNL/sundials/releases/download/v%7BSUNDIALS_VERSION%7D/sundials-%7BSUNDIALS_VERSION%7D.tar.gz), let me know how should i go with this.

It can be added to `.lycheeignore` as well, or we can ignore files or entire folders using glob patterns in the workflow file, let me know if neither of these options work

arjxn-py:
Added the non-fixable one in `.lycheeignore`, hopefully others would be fixed upon merging.

docs/source/user_guide/installation/gnu-linux-mac.rst
```diff
@@ -1,5 +1,5 @@
-GNU-Linux & MacOS
-=================
+gnu-linux-mac & MacOS
+=====================
```
agriyakhetarpal, line 2:
```suggestion
GNU/Linux & macOS
=================
```

.github/workflows/run_periodic_tests.yml
```diff
@@ -153,3 +153,47 @@ jobs:
           eval "$(pyenv init -)"
           pyenv activate pybamm-${{ matrix.python-version }}
           pyenv uninstall -f $( python --version )
+
+  test_install_odes:
+    needs: style
+    runs-on: macos-latest
```
agriyakhetarpal, line 159:
```suggestion
    runs-on: ${{ matrix.os }}
```

Doesn't require the `style` job to pass since no extra code will be checked out here

.github/workflows/run_periodic_tests.yml
```diff
@@ -153,3 +153,47 @@ jobs:
...
+        matrix:
+            os: [ubuntu-latest, macos-latest]
+            python-version: ["3.8", "3.9", "3.10", "3.11"]
+        fail-fast: false
+    name: Test pybamm_install_odes on ${{ matrix.os }}
+
+    steps:
+      - name: Check out PyBaMM repository
+        uses: actions/checkout@v4
+
```
agriyakhetarpal, line 170:
```suggestion

      - name: Install Linux system dependencies
        if: matrix.os == 'ubuntu-latest'
        run: |
          sudo apt-get update
          sudo apt-get install gfortran gcc libopenblas-dev
```

.github/workflows/run_periodic_tests.yml
```diff
@@ -153,3 +153,47 @@ jobs:
...
+      - name: Install macOS system dependencies
```
agriyakhetarpal, line 171:
```suggestion
      - name: Install macOS system dependencies
        if: matrix.os == 'macos-latest'
```

.github/workflows/run_periodic_tests.yml
```diff
@@ -153,3 +153,47 @@ jobs:
...
+      - name: Set up Python ${{ matrix.python-version }}
+        id: setup-python
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install PyBaMM dependencies
+        run: |
+          pip install --upgrade pip wheel setuptools nox
+          pip install -e .[all]
```
agriyakhetarpal, line 193:
```suggestion
      - name: Install PyBaMM
        run: pip install -e .
```

I think this workflow was written a long while back so we don't need to install the build-time dependencies now, and the `[all]` extra is not required either

agriyakhetarpal (changes requested):
I realised I missed reading through the workflow so I have a few comments. Could you please run it on your fork once and add a link to the logs here?

I think we can leave `wget` as-is for this PR. We can look into removing it in #3651 together after this gets merged.

arjxn-py:
> Could you please run it on your fork once and add a link to the logs here?

Here's the [link](https://github.com/arjxn-py/PyBaMM/actions/runs/7315438180) to the workflow run on the respective branch, waiting for tests to pass.

agriyakhetarpal (approved):

arjxn-py:
Those failing example tests on linux are unrelated, right?

agriyakhetarpal:
Yes, opening a new issue about them

.github/workflows/run_periodic_tests.yml
```diff
@@ -153,3 +153,50 @@ jobs:
...
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install PyBaMM
+        run: pip install -e .
+
+      - name: Test pybamm_install_odes on ${{ matrix.os }}
+        run: |
+          pip cache purge
+          pip install wget cmake
```
Saransh-cpp, line 201: Always nice to use `python -m` to ensure that we're using the correct pip -
```suggestion
        run: python -m pip install -e .

      - name: Test pybamm_install_odes on ${{ matrix.os }}
        run: |
          python -m pip cache purge
          python -m pip install wget cmake
```

Saransh-cpp (approved):
Looks good, thanks for picking this up, @arjxn-py!

Issue #3415, opened 2023-10-06
Scheduled tests failing while running example tests on python 3.8 (Ubuntu)


arjxn-py:
Example tests failing while testing half-cell.ipynb - https://github.com/pybamm-team/PyBaMM/actions/runs/6427005783/job/17452030411#step:15:682

kratman:
I have noticed this on my PRs as well, but it appears to be intermittent. In my one PR it failed for python 3.11 instead of 3.8 like in your example. I have not been able to reproduce it locally either.

arjxn-py:
> I have noticed this on my PRs as well, but it appears to be intermittent. In my one PR it failed for python 3.11 instead of 3.8 like in your example. I have not been able to reproduce it locally either.

I have also noticed it failing on 3.9 but now it is consistently failing on 3.8. Potentially a resultant of #3198

kratman:
I can confirm it is random. If I run this code 3-4 times in the notebook then I get a failure locally.
```python
model_with_degradation = pybamm.lithium_ion.DFN({
    "working electrode": "positive",
    "SEI": "reaction limited",  # SEI on both electrodes
    "SEI porosity change": "true",
    "particle mechanics": "swelling and cracking",
    "SEI on cracks": "true",
    "lithium plating": "partially reversible",
    "lithium plating porosity change": "true",  # alias for "SEI porosity change"
})
param_GrSi = pybamm.ParameterValues("OKane2022_graphite_SiOx_halfcell")
param_GrSi.update({"SEI reaction exchange current density [A.m-2]": 1.5e-07})
var_pts = {"x_n": 1, "x_s": 5, "x_p": 7, "r_n": 1, "r_p": 30}
exp_degradation = pybamm.Experiment(["Charge at 0.3C until 1.5 V", "Discharge at 0.3C until 0.005 V"])
sim3 = pybamm.Simulation(model_with_degradation, parameter_values=param_GrSi, experiment=exp_degradation, var_pts=var_pts)
sol3 = sim3.solve()
t = sol3["Time [s]"].entries
V = sol3["Voltage [V]"].entries
plt.figure()
plt.plot(t,V)
plt.xlabel("Time [s]")
plt.ylabel("Voltage [V]")
plt.show()
```
![image](https://github.com/pybamm-team/PyBaMM/assets/10170302/85dd3cbe-4f1c-47c1-bc48-78e5006016f7)

I am running python 3.11 on a Mac M2 laptop.

[...]

Issue #3371, opened 2023-09-26
[Bug]: `JaxSolver` tests failure on Windows when running `Jax` with GPU support


agriyakhetarpal:
### PyBaMM Version

develop

### Python Version

3.11.5

### Describe the bug

Note: see relevant (extended) discussion on Slack. GPU support for the Jax solver was added in #3121.

Two test cases in the unit tests currently fail with the `JaxSolver` with `jax` and `jaxlib` installed from [community-maintained wheels on Windows](https://whls.blob.core.windows.net/unstable/index.html), which have support for GPU execution. Other Jax-related tests pass, and all Jax-related tests pass with the recent CPU-only versions of `jaxlib` and `jax` with native support for Windows (`0.4.16`).

On the other hand, the unit tests on CPU-only versions of `jax` and `jaxlib` on macOS machines are passing (with both combinations: `jax==0.4.11` + `jaxlib==0.4.11` and `jax==0.4.16` + `jaxlib==0.4.16`, tested on an M-series).

Also, `jaxlib=0.4.11` is the only `0.4.X` version to support CUDA 11.1 with CuDNN 8.6/8.9 in the aforementioned Windows builds. More details at https://github.com/cloudhan/jax-windows-builder

### Steps to Reproduce

Run the unit tests with `nox -s unit` or `python run-tests.py --unit` with `jax==0.4.11`, `jaxlib=0.4.11`, and CUDA 11.8 + CuDNN 8.6.0 installed on a Windows AMD64 system, in order to enable GPU support for the `Jax` solver tests.

### Relevant hardware specifications

Platform: Windows 11 64-bit (AMD64)
GPU: NVIDIA GTX 1650 Ti
VRAM: 4 GB
CUDA 11.1, CuDNN 8.6

### Relevant log output

```
======================================================================
ERROR: test_model_solver (test_solvers.test_jax_solver.TestJaxSolver.test_model_solver)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\Agriya\Desktop\O.S.S\pybamm\PyBaMM\tests\testcase.py", line 29, in wrapped
    return method(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Agriya\Desktop\O.S.S\pybamm\PyBaMM\tests\unit\test_solvers\test_jax_solver.py", line 35, in test_model_solver
    solution = solver.solve(model, t_eval)
[...]

PR #3198, opened 2023-07-27
Negative half-cell models

DrSOKane:
# Description

This pull request introduced negative half-cell models to PyBaMM, by means of three major changes:
* The option `"working electrode": ["both", "negative", "positive"]` has been replaced with `"half-cell": ["false", "true]`. The positive electrode is always considered to be the working electrode, which has always been the case but is now made explicit.
* Two new parameter sets - `Ecker2015_graphite_halfcell` and `OKane2022_negative_halfcell` - have been added to work with the new option.
* The degradation mechanisms `SEI`, `SEI on cracks` and `lithium plating` can now be used on either electrode by specifying them as a 2-tuple, in the same way as is already done for some other options. Unlike for other options, however, passing only a single string instead of a tuple will put that behaviour on the negative electrode only. 

Fixes #2730 

## Type of change

Please add a line in the relevant section of [CHANGELOG.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CHANGELOG.md) to document the change (include PR #) - note reverse order of PR #s. If necessary, also add to the list of breaking changes.

- [x] New feature (non-breaking change which adds functionality)
- [ ] Optimization (back-end change that speeds up the code)
- [x] Bug fix (non-breaking change which fixes an issue)

# Key checklist:

- [x] No style issues: `$ pre-commit run` (see [CONTRIBUTING.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CONTRIBUTING.md#installing-and-using-pre-commit) for how to set this up to run automatically when committing locally, in just two lines of code)
- [x] All tests pass: `$ python run-tests.py --all`
- [x] The documentation builds: `$ python run-tests.py --doctest`

You can run unit and doctests together at once, using `$ python run-tests.py --quick`.

## Further checks:

- [x] Code is commented, particularly in hard-to-understand areas
- [x] Tests added that prove fix is effective or that feature works

PR #3423, opened 2023-10-09
JaxSolver fails when using GPU support with no input parameters

### LINKED ISSUE #3422: [Bug]: JaxSolver fails when using GPU support with no input parameters
### PyBaMM Version

23.5

### Python Version

3.10.12

### Describe the bug

When specified with input parameters, the JaxSolver will parallelise the solve across parameter sets either using `asyncio` (cpu) or jax-`vmap` (gpu). When specified without input parameters the cpu pathway continues to solve the model correctly; however, when the solver is called without an input parameter list in a gpu-enabled jax environment then the solver fails. This is because the `vmap` function requires at least one input argument to contain a non-empty array.

### Steps to Reproduce

```python
import pybamm
import numpy as np

import jax

platform = jax.lib.xla_bridge.get_backend().platform.casefold()
print("Platform: ", platform)

model = pybamm.BaseModel()

x = pybamm.Variable("x")
y = pybamm.Variable("y")

dxdt = 4 * x - 2 * y
dydt = 3 * x - y

model.convert_to_format = 'jax'
model.rhs = {x: dxdt, y: dydt}
model.initial_conditions = {x: pybamm.Scalar(1), y: pybamm.Scalar(2)}
model.variables = {"x": x, "y": y, "z": x + 4 * y}
disc = pybamm.Discretisation()  # use the default discretisation
disc.process_model(model);

solver = pybamm.JaxSolver()
t = np.linspace(0, 1, 20)
solution = solver.solve(model, t)

t_sol, y_sol = solution.t, solution.y  # get solution times and states
x = solution["x"]  # extract and process x from the solution
y = solution["y"]  # extract and process y from the solution

print(x.data)
print(y.data)
```

### Relevant log output

```shell
Platform:  gpu
jax.errors.SimplifiedTraceback: For simplicity, JAX has removed its internal frames from the traceback of the following exception. Set JAX_TRACEBACK_FILTERING=off to include these.

  File "/home/jsb/repos/jsbrittain/PyBaMM/test.py", line 27, in <module>
    solution = solver.solve(model, t)
  File "/home/jsb/repos/jsbrittain/PyBaMM/pybamm/solvers/base_solver.py", line 853, in solve
  File "/home/jsb/repos/jsbrittain/PyBaMM/pybamm/solvers/jax_solver.py", line 240, in _integrate
    y.extend(jax.vmap(self._cached_solves[model])(inputs_v))
ValueError: vmap wrapped function must be passed at least one argument containing an array, got empty *args=({},) and **kwargs={}
```

jsbrittain:
# Description

When specified with input parameters, the JaxSolver will parallelise the solve across parameter sets either using asyncio (cpu) or jax-vmap (gpu). When specified without input parameters the cpu pathway continues to solve the model correctly; however, when the solver is called without an input parameter list in a gpu-enabled jax environment then the solver fails. This is because the vmap function requires at least one input argument to contain a non-empty array.

A fix is to ensure that input parameter sets of length 0 or 1 are directed towards the cpu pathway, since this parallelisation occurs mostly over parameter sets, rather than within solves.

Fixes #3422 

## Type of change

# Key checklist:

## Further checks:

** Existing tests should cover this scenario, but depend upon GPU runners which are work-in-progress. **


brosaplanella (approved):
Looks good, thanks! Let's hold merging this until the wheel publishing is fixed.

agriyakhetarpal:
Thanks for this @jsbrittain, I tested this on Windows locally and I can confirm that #3371 is fixed through the changes in this PR. I didn't know it would be as easy as this but I get it because of my general inexperience with Jax. Could you please close that issue too here and add it to the CHANGELOG?

I have opened a new issue about bumping the `jax` and `jaxlib` versions so that GPU support can be targeted across platforms and for other reasons, see #3443

Saransh-cpp (commented):
The wheel and rc situation is fixed now. The CHANGELOG line should be moved to the unreleased section and everything should be good to go.

agriyakhetarpal:
The [failing doctests](https://github.com/pybamm-team/PyBaMM/actions/runs/6638247118/job/18034727563#step:8:453) here should not be a worry. I would suggest setting that configuration value to `True`, though.

agriyakhetarpal:
The [failing doctests](https://github.com/pybamm-team/PyBaMM/actions/runs/6638247118/job/18034727563#step:8:453) here should not be a worry. I would suggest setting that configuration value to `False`, though.

agriyakhetarpal:
The example notebooks issue should be this one: #3415, which came after the changes made in #3198.

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # [Unreleased](https://github.com/pybamm-team/PyBaMM/)
 
+- Fixed a bug where the JaxSolver would fails when using GPU support with no input parameters ([#3423](https://github.com/pybamm-team/PyBaMM/pull/3423))
```
Saransh-cpp, line 3:
```suggestion
## Bug fixes

- Fixed a bug where the JaxSolver would fails when using GPU support with no input parameters ([#3423](https://github.com/pybamm-team/PyBaMM/pull/3423))
```

Saransh-cpp (approved):
Thanks, @jsbrittain! Sorry for all the release mess 😬

agriyakhetarpal:
@brosaplanella: I missed talking about it in the meeting—should this PR be included in the release given that this is a bug fix? I think #3443 ought to be taken a look at too otherwise the GPU support mentioned in our CHANGELOG for v23.9rc0 would not be available for users on some platforms (e.g., macOS with Metal requires `v0.4.11`—we are still on `v0.4.8`).

I would be happy to write a PR for that this week and incorporate @jsbrittain's suggestions on it into consideration, please let me know if that would be needed

brosaplanella:
If we can include it that would be great. Tagging @Saransh-cpp so he is aware and can provide some input.

Saransh-cpp:
I'll merge this and add it in the rc1 release.

PR #3424, opened 2023-10-09
Fix typo .all-contributorsrc

brosaplanella:
There's a typo in the .all-contributorsrc which causes the bot to fail. This hopefully fixes it.


Saransh-cpp (approved):
Thanks, @brosaplanella!

PR #3429, opened 2023-10-09
Rename parameter "Exchange-current density for plating" to "Exchange-current density for lithium metal electrode"

### LINKED ISSUE #3428: Add better error about the new "exchange-current density for lithium metal electrode [A.m-2]" parameter
It used to be "Exchange-current density for plating [A.m-2]", now renamed to "Exchange-current density for lithium metal electrode [A.m-2]". I understand why the change was made, but we should catch this in the `ParameterValues` class and throw an informative error

agriyakhetarpal:
# Description

Closes #3428, adds a check for the use of the "Exchange-current density for plating [A.m-2]" parameter because it is to be renamed to "Exchange-current density for lithium-metal electrode [A.m-2]"

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/tests/unit/test_parameters/test_parameter_values.py
+++ b/tests/unit/test_parameters/test_parameter_values.py
@@ -147,6 +147,10 @@ class TestParameterValues(TestCase):
             # since + has other meanings in regex
         with self.assertRaisesRegex(ValueError, "Thermodynamic factor"):
             pybamm.ParameterValues({"1 + dlnf/dlnc": 1})
+        with self.assertRaisesRegex(
+            ValueError, "Exchange-current density for lithium-metal electrode"):
+            pybamm.ParameterValues(
+                {"Exchange-current density for plating": 0.001})
 
     def test_process_symbol(self):
         parameter_values = pybamm.ParameterValues({"a": 4, "b": 2, "c": 3})
--- a/tests/unit/test_parameters/test_parameter_values.py
+++ b/tests/unit/test_parameters/test_parameter_values.py
@@ -148,9 +148,9 @@ class TestParameterValues(TestCase):
         with self.assertRaisesRegex(ValueError, "Thermodynamic factor"):
             pybamm.ParameterValues({"1 + dlnf/dlnc": 1})
         with self.assertRaisesRegex(
-            ValueError, "Exchange-current density for lithium-metal electrode"):
+            ValueError, "Exchange-current density for plating [A.m-2]"):
             pybamm.ParameterValues(
-                {"Exchange-current density for plating": 0.001})
+                {"Exchange-current density for lithium-metal electrode [A.m-2]": 0.001})
 
     def test_process_symbol(self):
         parameter_values = pybamm.ParameterValues({"a": 4, "b": 2, "c": 3})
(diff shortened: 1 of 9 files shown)


tests/unit/test_parameters/test_parameter_values.py
```diff
@@ -147,6 +147,10 @@ def test_check_parameter_values(self):
             # since + has other meanings in regex
         with self.assertRaisesRegex(ValueError, "Thermodynamic factor"):
             pybamm.ParameterValues({"1 + dlnf/dlnc": 1})
+        # renamed to "Exchange-current density for lithium-metal electrode [A.m-2]"
```
agriyakhetarpal, line 150: I used `assertRaisesRegex` here to catch the error but it had problems parsing the minus sign (`-`) in the units, so I ended up switching to the usual `assertRaises` instead. Please let me know if is correct.

brosaplanella:
Tagging @Saransh-cpp as this should be included in the release

tests/unit/test_parameters/test_parameter_values.py
tinosulzer: I usually test a substring that works e.g. "Exchange-current density for plating"

tinosulzer (changes requested):
it should be "lithium metal" not "lithium-metal"

agriyakhetarpal:
There is a reference for "plating" here

https://github.com/pybamm-team/PyBaMM/blob/4f5d8051c2acfa18112a95fc554377605844f1d0/pybamm/input/parameters/lithium_ion/OKane2022_graphite_SiOx_halfcell.py#L402-L403

however, another reference for "lithium metal electrode" exists here

https://github.com/pybamm-team/PyBaMM/blob/4f5d8051c2acfa18112a95fc554377605844f1d0/pybamm/input/parameters/lithium_ion/OKane2022_graphite_SiOx_halfcell.py#L456-L457

which one should I keep or modify?

agriyakhetarpal:
There is a reference for "plating" here

https://github.com/pybamm-team/PyBaMM/blob/4f5d8051c2acfa18112a95fc554377605844f1d0/pybamm/input/parameters/lithium_ion/OKane2022_graphite_SiOx_halfcell.py#L402-L403

however, another reference for "lithium metal electrode" already exists here

https://github.com/pybamm-team/PyBaMM/blob/4f5d8051c2acfa18112a95fc554377605844f1d0/pybamm/input/parameters/lithium_ion/OKane2022_graphite_SiOx_halfcell.py#L456-L457

which one should I keep or modify?

tinosulzer:
Both, one is for the plating on the lithium meta electrode and the other is for plating in a porous electrode

Tagging @DrSOKane to make sure the changes get made correctly

agriyakhetarpal:
Makes sense, we would have to rename one of them to keep both of them though because we can't have duplicate keys

tests/unit/test_parameters/test_parameter_values.py
agriyakhetarpal: Okay, I managed to fix that using raw strings in order to keep `regex` happy.

I do have a separate question about how or if the units should be accounted for: because if the units are to be included in the error, `Exchange-current density for plating [A.m-2]` will raise an error as expected, but `Exchange-current density for plating` (without the units) will not

tinosulzer:
Sorry instructions weren't clear on how to fix this, I'll open a new PR with the fix

PR #3430, opened 2023-10-09
Multi-architecture Docker images and parallel builds

agriyakhetarpal:
# Description

This PR uses QEMU to build images for aarch64/arm64 platforms besides amd64, and fixes the error in https://github.com/pybamm-team/PyBaMM/pull/3316#discussion_r1349778849 about an improper tag for one of the images by using a build-matrix setup to create images. Related to #3312.

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -17,6 +17,7 @@ jobs:
     strategy:
       matrix:
         build-args: ["", "JAX=true", "ODES=true", "IDAKLU=true", "ALL=true"]
+      fail-fast: true
 
     steps:
       - name: Checkout
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -9,7 +9,7 @@ on:
 jobs:
   build_docker_images:
     # This workflow is only of value to PyBaMM and would always be skipped in forks
-    # if: github.repository_owner == 'pybamm-team'
+    if: github.repository_owner == 'pybamm-team'
     name: Image (${{ matrix.build-args }})
     runs-on: ubuntu-latest
     strategy:
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -17,7 +17,6 @@ jobs:
     strategy:
       matrix:
         build-args: ["", "JAX=true", "ODES=true", "IDAKLU=true", "ALL=true"]
-        architectures: [amd64, arm64]
 
     steps:
       - name: Checkout
@@ -58,7 +57,7 @@ jobs:
           tags: pybamm/pybamm:${{ steps.tags.outputs.tag }}
           push: false # temporary
           build-args: ${{ matrix.build-args }}
-          platforms: linux/${{ matrix.architectures }}
+          platforms: linux/amd64, linux/arm64
 
       - name: List built image(s)
         run: docker images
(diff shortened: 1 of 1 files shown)


agriyakhetarpal:
I opened this PR from @arjxn-py's fork because that repository has credentials for Docker Hub. Currently, the ALL and the JAX images are going to fail ([logs](https://github.com/arjxn-py/PyBaMM/actions/runs/6462421386)) because conda fails to find a suitable version of `jax` for which we might need to bump the versions or relax the requirements. I will open a separate issue for it soon because it looks like we might need to update them for GPU support on ARM-based macOS machines and possibly on Windows, see the relevant discussion on #3371 and #3423

agriyakhetarpal:
So I did some digging and found this: https://github.com/google/jax/issues/13608#issuecomment-1598713096 which sheds some info. One needs to build `jaxlib` from source to run it inside a Docker container on M-series macOS systems, which is pretty overkill because arm64 macOS wheels for `jaxlib` have been widely available and PyBaMM can of course use the Jax solver without the existence of the build-time requirements, which are (or rather, were) the main cause of our tricky installation procedure. Should we exclude the JAX and ALL images from `arm64` builds, and document this exclusion with a short note somewhere in the Docker installation guide?

agriyakhetarpal:
I realised that specifying the platforms in a matrix might overwrite images depending on which image and platform build gets completed (and thereby pushed) first. For now, adding conditional steps looks like an apt solution to control what images are built and pushed.

arjxn-py:
> I realised that specifying the platforms in a matrix might overwrite images depending on which image and platform build gets completed 

Are these images compatible with other platforms as well?

.github/workflows/docker.yml
```diff
@@ -7,69 +7,80 @@ on:
     - develop
 
 jobs:
-  pre_job:
+  build_docker_images:
+    # This workflow is only of value to PyBaMM and would always be skipped in forks
+    # if: github.repository_owner == 'pybamm-team'
```
Saransh-cpp, line 12: Should be reverted

Saransh-cpp (approved):
Looks amazing, thanks, @agriyakhetarpal! I'll request @arjxn-py's review before merging this.

.github/workflows/docker.yml
agriyakhetarpal: It does run on forks though, right?

arjxn-py (commented):
Thanks @agriyakhetarpal, looks awesome.
But when i see the images at docker hub, it looks like that the `latest` image is getting updated instead of `all` image.

agriyakhetarpal:
> Thanks @agriyakhetarpal, looks awesome.
> But when i see the images at docker hub, it looks like that the `latest` image is getting updated instead of `all` image.

Yes, this workflow when merged will update the images and fix that issue (already identified in https://github.com/pybamm-team/PyBaMM/pull/3316#discussion_r1349778849)

arjxn-py:
> Yes, this workflow when merged will update the images and fix that issue

Best then, happy to go forward with this 🚀

.github/workflows/docker.yml
agriyakhetarpal: Done

agriyakhetarpal:
Thanks for the review @arjxn-py;  the workflow won't run in forks now, but could you un-set the secrets in your repository settings just in case? It would be good for security reasons, say, lest a password breach occurs. No repository outside this one should have them to minimise chances of unauthorised access.

arjxn-py:
> could you un-set the secrets in your repository settings just in case? 

Sure I'll do that.

arjxn-py (approved):

PR #3433, opened 2023-10-10
Set up `.dockerignore` file

agriyakhetarpal:
# Description

Related to #3312 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,33 @@
+# Byte-compiled / optimized / DLL files
+**/*.py[cod]
+
+# Python cache
+**/__pycache__/
+*.cache
+.ipynb_checkpoints/
+
+# Running files
+.~lock*
+*.swp
+
+# Sphinx build files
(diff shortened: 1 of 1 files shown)


agriyakhetarpal:
Oh okay, I added this because I saw some of these files showing up in the Docker containers earlier, but I think they have been fixed in the latest image because they don't show up now. Cloning from the Git repository looks like a cleaner method, didn't notice that! Happy to close this if you think so

arjxn-py:
> fixed in the latest image because they don't show up now.

If the issue has been fixed in latest image then yeah we can close it and if not we can try fixing it with adding some commands in Dockerfile.

agriyakhetarpal:
> If the issue has been fixed in latest image then yeah we can close it and if not we can try fixing it with adding some commands in Dockerfile.

Just pulled and tested—closing, thanks, @arjxn-py!

PR #3436, opened 2023-10-10
Build wheels on the 1st and 15th of every month

Saransh-cpp:
# Description

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/.github/workflows/publish_pypi.yml
+++ b/.github/workflows/publish_pypi.yml
@@ -183,3 +183,15 @@ jobs:
           password: ${{ secrets.TESTPYPI_TOKEN }}
           packages-dir: files/
           repository-url: https://test.pypi.org/legacy/
+
+  open_failure_issue:
+    needs: [build_windows_wheels, buld_wheels, build_sdist]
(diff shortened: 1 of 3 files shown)


.github/workflows/publish_pypi.yml
```diff
@@ -3,6 +3,9 @@ name: Build and publish package to PyPI
 on:
```
Saransh-cpp, line 3: Should we add a badge for this in the README?

agriyakhetarpal: Should be a good idea. Alternatively or in addition we can use one of those actions to open an issue if one of the jobs fails

Saransh-cpp: Works now - https://github.com/Saransh-cpp/PyBaMM/issues/13!

agriyakhetarpal (approved):
Thanks, @Saransh-cpp!

brosaplanella (approved):
Looks great, thanks!

PR #3437, opened 2023-10-11
Issue 2111 custom transport efficiency

### LINKED ISSUE #2111: Adding customized function for tortuosity
### Description

I was wondering if it is possible to enable adding customized non-Bruggeman functions to define tortuosity or MacMullin number as a function of porosity?

Thanks. 

### Motivation

_No response_

### Possible Implementation

_No response_

### Additional context

_No response_

TomTranter:
# Description

Adding new methods for calculating the transport efficiency beyond Bruggeman coefficient. This includes using the following relations for tortuosity factor ($\tau$) for use in transport efficiency $B=\epsilon/\tau$:

image

- Tortuosity factor: allows direct specification of a tortuosity factor
These were taken from 10.1016/j.ces.2007.03.041 and http://dx.doi.org/10.1016/j.coche.2016.02.006

Fixes #2111 

#

## Type of change

# Key checklist:

## Further checks:


TomTranter:
This line looks to be wrong as active material won't always be the only solid conducting material. Using 1-porosity would be better
https://github.com/pybamm-team/PyBaMM/blob/bd5e240a26927800fd7338da294915be6ca7966f/pybamm/models/submodels/transport_efficiency/bruggeman_transport_efficiency.py#L39

pybamm/models/submodels/transport_efficiency/general_transport_efficiency.py
```diff
@@ -0,0 +1,83 @@
...
+    component : str
+        The material for the model ('electrolyte' or 'electrode').
+    options : dict, optional
+        A dictionary of options to be passed to the model.
+    """
+
+    def __init__(self, param, component, options=None):
+        super().__init__(param, component, options=options)
+
+    def _tortuosity_factor_model(self, eps_k):
```
rtimms, line 24: can Bruggeman and tortuosity factor go in this method too, or are they treated differently for some reason?

TomTranter: I don't think it reduces total lines of code as you would have to do the if statement on the domain inside the sub-function

TomTranter: They are using additional specific parameters whereas the other ones just use porosity

rtimms: Makes sense, got it

TomTranter:
@rtimms @brosaplanella this should be ready to go. Benchmark regressions appear to be random and unrelated to changes.

pybamm/models/submodels/transport_efficiency/general_transport_efficiency.py
```diff
@@ -0,0 +1,83 @@
+#
+# Class for Bruggemantransport_efficiency
+#
+import pybamm
+from .base_transport_efficiency import BaseModel
+
+
+class GeneralTransportEfficiency(BaseModel):
```
brosaplanella, line 8: Do we anticipate transport efficiency models that would not be into `general_transport_efficiency.py`? In that case, it may be better to get rid of the base class and just have a `transport_efficiency.py` file. Alternatively, we could split `general_transport_efficiency.py` into subclasses (one for each model). Not sure what's best practice in these situations honestly, so happy to be convinced either way.

brosaplanella: One advantage of the latter is that when checking the submodel components it will be a lot more explicit.

brosaplanella (approved):
Looks good, just see my comment on the classes. We should wait to merge until we have done the rc.

pybamm/models/submodels/transport_efficiency/general_transport_efficiency.py
rtimms: For kinetics, each different option is its own class that just implements `_get_kinetics`

pybamm/models/submodels/transport_efficiency/bruggeman.py
```diff
@@ -1,12 +1,12 @@
...
-# Class for Bruggemantransport_efficiency
+# Class for Bruggeman transport_efficiency
 #
 import pybamm
 from .base_transport_efficiency import BaseModel
 
 
 class Bruggeman(BaseModel):
-    """Submodel for Bruggeman transport_efficiency
+    """Submodel for transport_efficiency
```
brosaplanella, line 9: Thanks, I like it much better now! Could you add a docstring with the specific model name for each class and, if possible, the reference?

brosaplanella (approved):
Looks great, thanks! Just a small comment on the docstrings. Let's hold this until the rc is published.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 # [v23.9rc0](https://github.com/pybamm-team/PyBaMM/tree/v23.9rc0) - 2023-10-31
 
 ## Features
+- Transport efficiency submodel has new options from the literature relating to different tortuosity factor models and also a new option called "tortuosity factor" for specifying the value or function directly as parameters ([#3437](https://github.com/pybamm-team/PyBaMM/pull/3437))
```
Saransh-cpp, line 6: This should be moved to the unreleased section.

TomTranter: Done

TomTranter:
@brosaplanella macOS builds failing but probably nothing to do with me

agriyakhetarpal (commented):
Hello @TomTranter! Could you add references to the docstrings of the model classes using the BibTeX directly instead, e.g.,

```python
:footcite:t:`bruggeman1935berechnung`
```
and so on?

It's much easier than hard-coding the title of the paper, plus the `sphinxcontrib-bibtex` extension will automatically parse it and add the references to the documentation.

pybamm/models/submodels/transport_efficiency/tortuosity_factor.py
```diff
@@ -0,0 +1,46 @@
+#
+# Class for tortuosity factor transport_efficiency
+#
+import pybamm
+from .base_transport_efficiency import BaseModel
+
+
+class TortuosityFactor(BaseModel):
+    """Submodel for user supplied tortusoity factor transport_efficiency
```
brosaplanella, line 9:
```suggestion
    """Submodel for user supplied tortuosity factor transport_efficiency
```

brosaplanella (approved):
Looks good to me, just a small typo on a docstring. Happy to merge when that's suggestion is merged.

brosaplanella:
Who was in the Hackathon team, so we can add them in the contributors table?

TomTranter:
> Who was in the Hackathon team, so we can add them in the contributors table?

@juliawind @amirDahari1 @tommaull @Ruimin-S @isaacsquires are the ones with code commits

brosaplanella:
The doc tests are failing and I think this blocks merging the PR. @agriyakhetarpal can you help with this? It seems the issue is with the references.

docs/source/api/models/submodels/transport_efficiency/index.rst
```diff
@@ -5,4 +5,11 @@ transport_efficiency
   :maxdepth: 1
 
   base_transport_efficiency
-  bruggeman_transport_efficiency
+  bruggeman
+  cation_exchange_membrane
```
agriyakhetarpal, line 9: Could you rename the heading from `transport_efficiency` to `Transport Efficiency`? This has caused errors previously with duplicate headings in a file and its toctree

.virtual_documents/docs/source/examples/notebooks/models/tortuosity_models.ipynb
agriyakhetarpal, line 1: I think this notebook should be removed, it does not show up in the documentation and is showing up as invalid JSON

agriyakhetarpal (commented):
It is a bit difficult to ascertain as to what the cause of the failure is since the failures are in the PDF build of the documentation (#3292) while the HTML and doctest builds have no issues. The logs appear to be very similar and there is hardly any difference between those for a [passing build](https://readthedocs.org/projects/pybamm/builds/22566391/) and those for [this PR (a failing build)](https://readthedocs.org/projects/pybamm/builds/22563790/), except for the keyword `failed` instead of `done` and the names of the new transport efficiency submodel entries, of course. We currently remove all notebooks-related pages in the PDF builds so I don't think the problem is in either of the two notebooks changed here. However, just to be sure and if it is possible: please add an alt-text for the images in the markdown syntax in the notebooks – not doing that was also a cause of an error I had encountered in #3043 with some of them.

I checked the references in the docstrings by matching them from the `CITATIONS.bib` file and they look fine. I can take a look at debugging the build locally but I will be a bit busy until next week, so I have left a few comments below which could help in the meantime.

.virtual_documents/docs/source/examples/notebooks/models/tortuosity_models.ipynb
TomTranter: Ok, I'll remove the file and try adding a new one. I'd like it to be included in the docs

agriyakhetarpal: I think there are two files of the same name though, this one is in `.virtual_documents/` and another one which is valid JSON and shows up correctly

TomTranter:
@agriyakhetarpal I can't seem to find the offending readthedocs failure. Any suggestions?

agriyakhetarpal:
Sorry, I forgot to give an update here but I did try this locally in the last week of November after I said I will. The error message was painfully difficult to debug, and I tried an array of approaches to fix it but resorted to leaving it be later on: moving the paths for various files, excluding all new files from this PR from the docs generation, experiments with switching to an alternative PDF builder (`rinohtype`), and more – nothing has worked so far.

However, I have been able to isolate that the problem is not with the notebooks (they are not processed during the PDF builds), and not with the directives in the `.rst` files – which suggests that this is coming from either how the file structure is distributed for the submodel or from something we missed in the docstrings for the submodel classes.

@tinosulzer had suggested enabling PDF builds when I was a GSoC participant, and so were they done soon enough, but we can look into removing them if we do not wish to include them again. Please let me know what works, in the meantime I can take another look at this on Monday.

DrSOKane (changes requested):
As someone who tried implementing the user-supplied tortuosity factor myself, this is very good to see!

The problem I quickly ran into with non-Bruggeman tortuosity models is how to tell PyBaMM to disregard the tortuosity for solid-phase conductivity. With Bruggeman models it is done by specifying an exponent of 0, but this doesn't work for any of the other models. The only solutions I can think of are:
* Specify separate options for electrolyte and electrode transport efficiency, which is a lot of work for little gain
* Specify a "tortuosity for electrode conductivity" option which can only be true or false
* Parameter sets corresponding to paper that use non-Bruggeman tortuosity models should have the initial electrode conductivity adjusted accordingly
 
Furthermore, using phi = 1 - eps is incorrect for models that feature degradation. The tortuosity will decrease as the SEI grows, and LAM is not accounted for.

These are minor issues that will be easy to solve _if_ everyone can agree on which approach to use.

TomTranter:
Hi @DrSOKane thanks for your comments and good suggestions. I think the best approach is "Specify separate options for electrolyte and electrode transport efficiency". It is perfectly valid to have different functional relationships for each phase. On the point of LAM I would need to dig into this but we should certainly account for it. We should Ideally have a volume fraction model that can account for active material, binder, SEI, non-active material (broken-off chunks) and porous space

TomTranter:
@agriyakhetarpal thanks for he update - it certainly is a hard problem to find and fix. Maybe @martinjrobins has experience that can help or I could contact the readthedocs maintainers

agriyakhetarpal:
I agree. Though, it doesn't look to me as something related to RTD (fails locally as well), so we would rather be better off contacting the Sphinx developers – though I can't seem to find the related issue at this moment, it had been dormant for years the last time I looked ☹️

TomTranter:
@agriyakhetarpal I found the source of the docs errors finally. It was missing footbibliography commands on the rst files

agriyakhetarpal:
Thanks, @TomTranter! That makes so much more sense now, I imagine it was surely not easy to find out the source of it through the unhelpful error message :)

TomTranter:
@brosaplanella can we merge this in now? @DrSOKane made a suggestion for a code change but I don't have time to implement that right now and would rather get this is now it's passing

brosaplanella:
I can't merge as it needs Simon's approval, given that he requested changes. @DrSOKane can you check if you are happy with it?

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Features
 
+- The `pybamm_install_odes` command now includes support for macOS systems and can be used to set up SUNDIALS and install the `scikits.odes` solver on macOS ([#3417](https://github.com/pybamm-team/PyBaMM/pull/3417), [#3706](https://github.com/pybamm-team/PyBaMM/3706]))
```
agriyakhetarpal, line 5:
```suggestion
```

This entry can be removed, since the `scikits.odes` solver was removed in #3932.

DrSOKane (approved):
I'm still not happy with the assumption that `phi = 1 - eps`. However, this is badly needed for the next release, so I will approve it. The binder can be added explicitly in a future PR.

PR #3455, opened 2023-10-17
Fix #3418 Allow pybamm-requires to return early if an existing SUNDIALS installation is found

### LINKED ISSUE #3418: Allow `pybamm-requires` to return early if an existing SUNDIALS installation is found
The `pybamm-requires` session, activated by the `nox -s pybamm-requires` command, can be sped up by detecting the build-time components in the $HOME directory. Note: this would be similar to what `pybamm_install_odes` already does; the script can find an installation of SUNDIALS and move to the next steps, i.e., installing `scikits.odes`. 

However, in cases of a corrupt installation of SuiteSparse and SUNDIALS, we can force the re-installation by `nox`'s posargs feature: providing the `-f` or the `--force` command-line flag when invoking the session will re-compile them. We already do this for the `pybind11` directory in the session by not cloning the repository if it is found.

This way, users can skip the re-installation of the build-time components if already present but still factor in the cases where it is necessary for them to reinstall for debugging purposes. Another way this will help will be to skip the session in CI if a cache-hit occurs (we don't need to bother about validating the caches because GitHub Actions runners would already do that for us)

Saswatsusmoy:
# Description

Users can skip the re-installation of the build-time components if already present but still factor in the cases where it is necessary for them to reinstall for debugging purposes.

The key aspects are:
-Checking for existing Sundials and KLU installations before installing
-Breaking out the install scripts for each dependency
-Allowing a force rebuild with --force
-Keeping the pybind11 install logic the same

Fixes #3418 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -36,11 +36,19 @@ def set_environment_variables(env_dict, session):
 
 @nox.session(name="pybamm-requires")
 def run_pybamm_requires(session):
-    """Download, compile, and install the build-time requirements for Linux and macOS: the SuiteSparse and SUNDIALS libraries."""  # noqa: E501
-    set_environment_variables(PYBAMM_ENV, session=session)
+    """Download, compile, and install the build-time requirements for Linux and macOS: the SuiteSparse and SUNDIALS libraries."""  
+    force_rebuild = "--force" in session.posargs
+    sundials_path = Path("/path/to/sundials")
+    suitesparse_path = Path("/path/to/suitesparse")
+    if (sundials_path.exists() and suitesparse_path.exists()) and not force_rebuild:
+        session.log("Found existing build dependencies")
+        return
     if sys.platform != "win32":
         session.install("wget", "cmake", silent=False)
-        session.run("python", "scripts/install_KLU_Sundials.py")
+        if not os.path.exists("/path/to/klu"):
+            session.run("python", "scripts/install_klu.py") 
+        if not os.path.exists(sundials_path):
+            session.run("python", "scripts/install_sundials.py")
         if not os.path.exists("./pybind11"):
             session.run(
                 "git",
--- a/noxfile.py
+++ b/noxfile.py
@@ -1,3 +1,4 @@
+from pathlib import Path
 import nox
 import os
 import sys
@@ -36,7 +37,7 @@ def set_environment_variables(env_dict, session):
 
 @nox.session(name="pybamm-requires")
 def run_pybamm_requires(session):
-    """Download, compile, and install the build-time requirements for Linux and macOS: the SuiteSparse and SUNDIALS libraries."""
+    """Download, compile, and install the build-time requirements for Linux and macOS"""
     force_rebuild = "--force" in session.posargs
     sundials_path = Path("/path/to/sundials")
     suitesparse_path = Path("/path/to/suitesparse")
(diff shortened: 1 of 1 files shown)


noxfile.py
```diff
@@ -36,11 +37,19 @@ def set_environment_variables(env_dict, session):
 
 @nox.session(name="pybamm-requires")
 def run_pybamm_requires(session):
-    """Download, compile, and install the build-time requirements for Linux and macOS: the SuiteSparse and SUNDIALS libraries."""  # noqa: E501
-    set_environment_variables(PYBAMM_ENV, session=session)
+    """Download, compile, and install the build-time requirements for Linux and macOS"""
+    force_rebuild = "--force" in session.posargs
+    sundials_path = Path("/path/to/sundials")
+    suitesparse_path = Path("/path/to/suitesparse")
```
agriyakhetarpal, line 43: The relevant files whose existence is to be checked for each of the build-time components are these shared objects, i.e., `.so` files (these will be, similarly, dynamic libraries in `.dylib` file format on macOS):

<details><summary>SUNDIALS</summary>
<p>

```bash
libsundials_idas.so
libsundials_sunlinsolklu.so
libsundials_sunlinsoldense.so
libsundials_sunlinsolspbcgs.so
libsundials_sunlinsollapackdense.so
libsundials_sunmatrixsparse.so
libsundials_nvecserial.so
libsundials_nvecopenmp.so
```

</p>
</details> 

<details><summary>SuiteSparse</summary>
<p>

```
libsuitesparseconfig.so
```

</p>
</details>

<details><summary>For the KLU component</summary>
<p>

```
libsuitesparseconfig.so
libklu.so
libamd.so
libcolamd.so
libbtf.so
```

</p>
</details>

If these files are not found, we should run the `install_KLU_Sundials.py` script

For more information: the SuiteSparse KLU module depends on these four components: suitesparseconfig, AMD, COLAMD, and BTF.

noxfile.py
```diff
@@ -36,11 +37,19 @@ def set_environment_variables(env_dict, session):
...
+    if (sundials_path.exists() and suitesparse_path.exists()) and not force_rebuild:
+        session.log("Found existing build dependencies")
```
agriyakhetarpal, line 45:
```suggestion
        session.warn("Found existing build-time requirements, skipping installation. Note: run with the --force flag (nox -s pybamm-requires -- --force) to invoke re-installation.")
```

It is indeed better to run `session.log` here, but `session.warn` will display explicit warnings in yellow-coloured text which will be more visible to the user

noxfile.py
```diff
@@ -36,11 +37,19 @@ def set_environment_variables(env_dict, session):
...
+        return
     if sys.platform != "win32":
         session.install("wget", "cmake", silent=False)
-        session.run("python", "scripts/install_KLU_Sundials.py")
+        if not os.path.exists("/path/to/klu"):
+            session.run("python", "scripts/install_klu.py")
+        if not os.path.exists(sundials_path):
+            session.run("python", "scripts/install_sundials.py")
```
agriyakhetarpal, line 52: We compile the build-time components in a single script, i.e., `install_KLU_Sundials.py`

noxfile.py
agriyakhetarpal: Also, could you similarly add a `session.log` that displays "Found pybind11" if the `pybind11/` folder exists, below?

agriyakhetarpal (changes requested):

agriyakhetarpal:
Closing this because #3461 has been opened. I would recommend pushing commits to the same branch and therefore in a single PR going forward

PR #3461, opened 2023-10-21
Fix #3418 Allow pybamm-requires to return early if an existing SUNDIAL…

### LINKED ISSUE #3418: Allow `pybamm-requires` to return early if an existing SUNDIALS installation is found

Saswatsusmoy:
Users can skip the re-installation of the build-time components if already present but still factor in the cases where it is necessary for them to reinstall for debugging purposes.

The key aspects are:
-Checking for existing Sundials and KLU installations before installing
-Breaking out the install scripts for each dependency
-Allowing a force rebuild with --force
-Keeping the pybind11 install logic the same

Fixes #3418 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -40,10 +40,59 @@ def set_environment_variables(env_dict, session):
 def run_pybamm_requires(session):
     """Download, compile, and install the build-time requirements for Linux and macOS: the SuiteSparse and SUNDIALS libraries."""  # noqa: E501
     set_environment_variables(PYBAMM_ENV, session=session)
+    force_rebuild = "--force" in session.posargs
+    sundials_so_path = [
+            Path("/path/to/sundials/libsundials_idas.so"),
+            Path("/path/to/sundials/libsundials_idas.dylib"),
+
+            Path("/path/to/sundials/libsundials_sunlinsolklu.so"),
+            Path("/path/to/sundials/libsundials_sunlinsolklu.dylib"),
+
+            Path("/path/to/sundials/libsundials_sunlinsoldense.so"),
+            Path("/path/to/sundials/libsundials_sunlinsoldense.dylib"),
+
+            Path("/path/to/sundials/libsundials_sunlinsolspbcgs.so"),
+            Path("/path/to/sundials/libsundials_sunlinsolspbcgs.dylib"),
+
+            Path("/path/to/sundials/libsundials_sunlinsollapackdense.so"),
+            Path("/path/to/sundials/libsundials_sunlinsollapackdense.dylib"),
+
+            Path("/path/to/sundials/libsundials_sunmatrixsparse.so"),
+            Path("/path/to/sundials/libsundials_sunmatrixsparse.dylib"),
+
+            Path("/path/to/sundials/libsundials_nvecserial.so"),
+            Path("/path/to/sundials/libsundials_nvecserial.dylib"),
+
+            Path("/path/to/sundials/libsundials_nvecopenmp.so"),
+            Path("/path/to/sundials/libsundials_nvecopenmp.dylib")
+        ]
+    klu_so_path = [
+            Path("/path/to/suitesparse/libsuitesparseconfig.so"),
+            Path("/path/to/suitesparse/libsuitesparseconfig.dylib"), # for MacOS
+            
+            Path("/path/to/suitesparse/libklu.so"),
+            Path("/path/to/suitesparse/libklu.dylib"),
+            
+            Path("/path/to/suitesparse/libamd.so"),
+            Path("/path/to/suitesparse/libamd.dylib"),
+            
+            Path("/path/to/suitesparse/libcolamd.so"),
+            Path("/path/to/suitesparse/libcolamd.dylib"),
+            
+            Path("/path/to/suitesparse/libbtf.so"),
+            Path("/path/to/suitesparse/libbtf.dylib"),
+        ]
     if sys.platform != "win32":
         session.install("wget", "cmake", silent=False)
-        session.run("python", "scripts/install_KLU_Sundials.py")
-        if not os.path.exists("./pybind11"):
+        
+        if (sundials_so_path.exists() and 
+            klu_so_path.exists()) and not force_rebuild:
+            session.warn("Found existing build-time requirements, skipping installation. Note: run with the --force flag (nox -s pybamm-requires -- --force) to invoke re-installation.")  # noqa: E501
+            return
+            session.run("python", "scripts/install_Klu_Sundials.py")
+        elif os.path.exists("./pybind11"):
+            session.log("Found pybind11")
+            return
             session.run(
                 "git",
                 "clone",
--- a/noxfile.py
+++ b/noxfile.py
@@ -51,7 +51,7 @@ def run_pybamm_requires(session):
         "libsundials_nvecserial",
         "libsundials_nvecopenmp"
     ]
-    
+
     KLU_LIBS = [
         "libsuitesparseconfig",
         "libklu",
@@ -79,7 +79,7 @@ def run_pybamm_requires(session):
             Path(PYBAMM_ENV["LD_LIBRARY_PATH"], lib + ".dylib")
             for lib in KLU_LIBS
     ]
-    
+
     if sys.platform != "win32":
         session.install("wget", "cmake", silent=False)
 
--- a/noxfile.py
+++ b/noxfile.py
@@ -51,7 +51,7 @@ def run_pybamm_requires(session):
         "libsundials_nvecserial",
         "libsundials_nvecopenmp"
     ]
-    
+
     KLU_LIBS = [
         "libsuitesparseconfig",
         "libklu",
@@ -64,7 +64,7 @@ def run_pybamm_requires(session):
         fileext = ".so"
     elif sys.platform == "darwin":
         fileext = ".dylib"
-    
+
     sundials_so_path = [
         Path(PYBAMM_ENV["LD_LIBRARY_PATH"], lib + fileext)
         for lib in SUNDIALS_LIBS
@@ -81,7 +81,7 @@ def run_pybamm_requires(session):
             session.warn("Found existing build-time requirements, skipping installation. Note: run with the --force flag (nox -s pybamm-requires -- --force) to invoke re-installation.")
         else:
             session.run("python", "scripts/install_KLU_Sundials.py")
-        
+
         if os.path.exists("./pybind11"):
             session.log("Found pybind11")
             return
--- a/noxfile.py
+++ b/noxfile.py
@@ -51,7 +51,7 @@ def run_pybamm_requires(session):
         "libsundials_nvecserial",
         "libsundials_nvecopenmp"
     ]
-    
+
     KLU_LIBS = [
         "libsuitesparseconfig",
         "libklu",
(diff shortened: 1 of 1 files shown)


Saswatsusmoy:
@agriyakhetarpal I have created a new PR with all the requested changes.

noxfile.py
```diff
@@ -40,10 +40,59 @@ def set_environment_variables(env_dict, session):
...
+
+            Path("/path/to/sundials/libsundials_sunmatrixsparse.so"),
+            Path("/path/to/sundials/libsundials_sunmatrixsparse.dylib"),
+
+            Path("/path/to/sundials/libsundials_nvecserial.so"),
+            Path("/path/to/sundials/libsundials_nvecserial.dylib"),
+
+            Path("/path/to/sundials/libsundials_nvecopenmp.so"),
+            Path("/path/to/sundials/libsundials_nvecopenmp.dylib")
+        ]
```
agriyakhetarpal, line 68: The path to these `.so` and `.dylib` files is `$HOME/local/lib`, i.e., `LD_LIBRARY_PATH`  (we already have these variables set above as constants in the file which you can reuse).

Here, `.dylib` files exist only on macOS and similarly `.so` files only on Linux, so you should use `sys.platform` like it has been used in other places in the file to check the target platform. Also, could you write the filenames in the list first and then append the `.dylib` and `.so` extensions and construct `Path` objects later? The code will be much cleaner that way. Something like this:

```python

SUNDIALS_LIBS = [
    ...
    "libsundials_idas",
    "libsundials_nvecopenmp",
    ...
]
if sys.platform == "darwin"
    fileext = ".dylib"
if sys.platform == "linux"
    fileext = ".so"
```

and join filenames with `fileext`, create `Path` objects by looping over them, check if they exist in `LD_LIBRARY_PATH`. Alternatively, the `os` module can be used to check if these files exist (`os.path.isfile`), but this is less preferable because we already have a `pathlib` import.

noxfile.py
```diff
@@ -40,10 +40,59 @@ def set_environment_variables(env_dict, session):
...
+
+            Path("/path/to/suitesparse/libamd.so"),
+            Path("/path/to/suitesparse/libamd.dylib"),
+
+            Path("/path/to/suitesparse/libcolamd.so"),
+            Path("/path/to/suitesparse/libcolamd.dylib"),
+
+            Path("/path/to/suitesparse/libbtf.so"),
+            Path("/path/to/suitesparse/libbtf.dylib"),
+        ]
```
agriyakhetarpal, line 84: Same here

noxfile.py
```diff
@@ -40,10 +40,59 @@ def set_environment_variables(env_dict, session):
...
     if sys.platform != "win32":
         session.install("wget", "cmake", silent=False)
-        session.run("python", "scripts/install_KLU_Sundials.py")
-        if not os.path.exists("./pybind11"):
+
+        if (sundials_so_path.exists() and
+            klu_so_path.exists()) and not force_rebuild:
+            session.warn("Found existing build-time requirements, skipping installation. Note: run with the --force flag (nox -s pybamm-requires -- --force) to invoke re-installation.")  # noqa: E501
+            return
```
agriyakhetarpal, line 91:
```suggestion
```

No need for `return` statements since `nox` runs as a CLI application only. You may use `if-else` clauses with the `session.run` commands to achieve the desired effect

noxfile.py
```diff
@@ -40,10 +40,59 @@ def set_environment_variables(env_dict, session):
...
+            session.run("python", "scripts/install_Klu_Sundials.py")
+        elif os.path.exists("./pybind11"):
+            session.log("Found pybind11")
+            return
```
agriyakhetarpal, line 95:
```suggestion
```

No need for the `elif` keyword either since this is another condition we need to check that does not rely on other conditions (`pybind11` can be found even if the SUNDIALS libraries are not found)

agriyakhetarpal (changes requested):

noxfile.py
```diff
@@ -40,19 +40,66 @@ def set_environment_variables(env_dict, session):
...
+    elif sys.platform == "darwin":
+        sundials_so_path = [
+            Path(PYBAMM_ENV["LD_LIBRARY_PATH"], lib + ".dylib")
+            for lib in SUNDIALS_LIBS
+    ]
+        klu_so_path = [
+            Path(PYBAMM_ENV["LD_LIBRARY_PATH"], lib + ".dylib")
+            for lib in KLU_LIBS
+    ]
+
```
agriyakhetarpal, line 82: I tested this locally: here, `Path()` is just constructing PathLike objects and then you're doing a list comprehension over them, this approach doesn't check if the paths to these files actually exist. The correct flow should be to create paths from SUNDIALS_LIBS and KLU_LIBS and have one single function return a boolean if all of them are found. This value can be checked upon calling the function (whether it's `True` or `False`) and the session.install and session.run statements can be executed or skipped accordingly.

agriyakhetarpal (changes requested):
This is starting to look a bit better. Are you testing these changes locally? I would recommend doing that since it will help debug how you should go forth with these changes.

noxfile.py
Saswatsusmoy:
```
SUNDIALS_LIBS = [
    # libs
    ]

KLU_LIBS = [
    # libs
    ]
    

def check_lib_paths(lib_names, lib_dir):

  if sys.platform == "darwin":
    ext = ".dylib"
  elif sys.platform == "linux":
    ext = ".so"

  paths = [Path(lib_dir, lib + ext) for lib in lib_names]  
  return all(os.path.exists(path) for path in paths)


    # Set Variables
    
    sundials_path = check_lib_paths(SUNDIALS_LIBS, PYBAMM_ENV["LD_LIBRARY_PATH"])
    klu_path = check_lib_paths(KLU_LIBS, PYBAMM_ENV["LD_LIBRARY_PATH"])

    if sys.platform != "win32":
        session.install("wget", "cmake", silent=False)

        if sundials_path and klu_path and not force_rebuild:
           # Rest of the code
```
        
        
Something like this?

Saswatsusmoy: When I test this locally, it doesn't exits early even if the dependencies are still present but if I change 

`if sundials_path and klu_path and not force_rebuild:` to `if sundials_path !=[ ] and klu_path != [ ] and not force_rebuild:`

it is exiting early

noxfile.py
```diff
@@ -40,19 +40,66 @@ def set_environment_variables(env_dict, session):
...
-        if not os.path.exists("./pybind11"):
+
+        if sundials_so_path != [] and klu_so_path != [] and not force_rebuild:
+            session.warn("Found existing build-time requirements, skipping installation. Note: run with the --force flag (nox -s pybamm-requires -- --force) to invoke re-installation.")
+        else:
+            session.run("python", "scripts/install_KLU_Sundials.py")
+
+        if os.path.exists("./pybind11"):
+            session.log("Found pybind11")
+        elif not os.path.exists("./pybind11"):
```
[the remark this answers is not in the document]
agriyakhetarpal: No, move this part to the top of the session, rather

noxfile.py
```diff
@@ -40,19 +40,66 @@ def set_environment_variables(env_dict, session):
...
+        "libsundials_nvecopenmp"
+    ]
+
+    KLU_LIBS = [
+        "libsuitesparseconfig",
+        "libklu",
+        "libamd",
+        "libcolamd",
+        "libbtf"
+    ]
```
Saransh-cpp, line 61: @agriyakhetarpal, I think we should check for the existing installation in `install_KLU_Sundials.py` itself? We recommend developers to use `nox` but we also have documentation for doing stuff without `nox`, so it might be better to use `nox` just as a wrapper for these scripts.

IIRC the checks are also present in `install_odes.py` (and not the `nox` session of `install_odes.py`).

agriyakhetarpal: I would recommend doing these checks with `nox` rather than inside the script because `nox` is meant to ease the installation of the build-time requirements both for newer contributors and for experienced developers. The rationale behind my suggestion is that not only does `nox` provide a powerful command-line interface and can abstract away the extra code required to implement a command-line argument parser in a Python script (not that doing that would take time, though), it is also our current "recommended" method of installing from source in the installation instructions. Shall someone run `install_KLU_Sundials.py` and clone the `pybind11` repository manually, it should be assumed that they know what they are doing and the script should _recompile_ SuiteSparse and SUNDIALS as usual even if both components already exist on the system.

However, `pybamm_install_odes` is provided as an entry point for use in user installations. This is why we should skip adding a `nox` session for it – we don't bundle the noxfile in the source distribution or in the wheels anyway, plus for developers this is redundant (`pip install pybamm[odes]` now does the job once `SUNDIALS_INST` is correctly set). Therefore, `install_odes.py` should keep those checks inside itself since it is a standalone application that is shipped with PyBaMM.

Note: yes, the checks are inside `install_odes.py` but they are not complete; just one SUNDIALS-specific file is checked, i.e., `libsundials_idas.so`, compared to the list of files I have included here. We should update that soon (@arjxn-py, could you do that in #3417?)

<hr>

P.S. going forward in the coming months in the aftermath of adding `pyproject.toml`, the changes here will not matter since steps can be taken to eliminate the dependence on `pybamm-requires` completely wherein a combination of a robust caching mechanism inside the script with hashes and the use of CCache can be devised to speed up and simplify the installation for both users and developers.

arjxn-py: > could you do that in #3417?

Yes sure, hope it's fine to do it after completing #3475

noxfile.py
agriyakhetarpal: You can modify `check_lib_paths` to iterate over the list of file paths and return either `True` or `False`, and change the condition in that case, which can then look like

```python
if check_lib_paths() and not force_rebuild:
    session.run("...")
```

i.e., the boolean `and` operator can then take care of the condition

agriyakhetarpal:
@Saswatsusmoy I would now suggest moving the checks for these dynamic libraries and shared object files to `install_KLU_Sundials.py` itself – I did suggest keeping them in `noxfile.py` initially, but this recommendation comes after the fact that I am changing up some parts of the script myself since it was decided to establish a user-specific SUNDIALS and SuiteSparse installation directory. Doing that would not alter or add any files in `.local/`, and we can use the existing argument parser available in the script and add the "--force" argument to it. `nox -s pybamm-requires` can then use posargs to inherit that `--force` argument.

agriyakhetarpal:
This PR has been inactive for two months now, closing. I can take up this task in conjunction with others listed in #3651.

Issue #3462, opened 2023-10-21
Build wheels for more platforms (Linux aarch64 wheels, MUSL wheels, etc.)


agriyakhetarpal:
### Description

Currently, we build wheels for amd64 (Linux, macOS, Windows) and also pypy wheels on Windows (https://pypi.org/project/pybamm/#files). However, we don't provide wheels for arm64 platforms yet neither for Linux nor M-series macOS.

This means that users installing from PyPI use the source distributions, which do not contain compiled extensions. Their compilation requires `pybind11` and the presence of SUNDIALS, SuiteSparse, and CasADi to link against which is easier done with an installation from source (not available and accessible for all users since this needs elevated or administrator permissions on the host system). Otherwise, `pip` falls back to building and storing a wheel without a platform and an ABI tag (`none-any.whl`) on the machine since the IDAKLU component is optional by default and doesn't compile if CMake or pybind11 are not found. This might be (slightly) easier to do after #3049 is completed.

### Motivation

Easier and streamlined installation on the aforementioned platforms since the compiled extension is bundled into the wheels along with its shared external libraries (they are repaired to recreate static linkage and their runtime search paths altered)

### Possible Implementation

GitHub Actions has some larger-sized M-series macOS runners now with smaller runners coming in the future. Linux arm64 runners are also on their roadmap. We can however

1. use emulation with either QEMU (recommended by the Scientific Python guides) or Dockcross (casadi) to build arm64 wheels
2. the Apple Xcode tools are also capable of cross-compilation to build macOS arm64 wheels in the form of a universal wheel which is double the size

### Additional context

https://cibuildwheel.readthedocs.io/en/stable/faq/

### Task list

- [x] macOS arm64 wheels (#3772, #3789)
- [ ] Linux aarch64 wheels
- [ ] MUSL wheels (if needed – note: might not work due to CasADi)

agriyakhetarpal:
The macOS arm64 builds were successful here: https://github.com/agriyakhetarpal/PyBaMM/actions/runs/6599848898/job/17929588884#step:10:1 through cross-compilation. I am now trying to fix troubles with not being able to find BLAS on Linux arm64... iiuc `openblas-devel` isn't available in the index, so I have tried linking with `blas-devel` currently...

agriyakhetarpal:
[...]

PR #3465, opened 2023-10-23
Build wheels for arm64 macOS

agriyakhetarpal:
# Description

Related to #3462, see there for extended discussion. This PR enables `cibuildwheel` to use Xcode available on GitHub Actions runners which can cross-compile wheels for macOS with arm64 architectures.

> [!Important]  
These wheels should be tested on M-series macOS machines, they may be downloaded from this link: 
https://github.com/agriyakhetarpal/PyBaMM/actions/runs/6667379053 (updated)

## Type of change

# Key checklist:

## Further checks:


agriyakhetarpal:
I tried this on my end and I am receiving this trace when trying to import the extension module (within `pybamm.have_idaklu()`):

```
ImportError: dlopen(/Users/agriyakhetarpal/Desktop/PyBaMM/env/lib/python3.11/site-packages/pybamm/solvers/idaklu.cpython-311-darwin.so, 0x0002): tried: '/Users/agriyakhetarpal/Desktop/PyBaMM/env/lib/python3.11/site-packages/pybamm/solvers/idaklu.cpython-311-darwin.so' (mach-o file, but is an incompatible architecture (have 'x86_64', need 'arm64')), '/System/Volumes/Preboot/Cryptexes/OS/Users/agriyakhetarpal/Desktop/PyBaMM/env/lib/python3.11/site-packages/pybamm/solvers/idaklu.cpython-311-darwin.so' (no such file), '/Users/agriyakhetarpal/Desktop/PyBaMM/env/lib/python3.11/site-packages/pybamm/solvers/idaklu.cpython-311-darwin.so' (mach-o file, but is an incompatible architecture (have 'x86_64', need 'arm64'))
```

which is confusing since the [`cibuildwheel` documentation](https://cibuildwheel.readthedocs.io/en/stable/faq/#how-to-cross-compile) mentions that cross-compiling arm64 wheels with Xcode should be as simple as a drop-in addition to `CIBW_ARCHS_MACOS`. Perhaps this is a bug that I should file on their issue tracker?

agriyakhetarpal:
The newest wheels build on my branch is here: https://github.com/agriyakhetarpal/PyBaMM/actions/runs/6667379053

I tried and tested the import of the extension module on my end, the wheel can be installed correctly and `pybamm.have_idaklu()` returns `True` now.  🎉

#3301 removes the file that contains the CMake args and adds them to `setup.py` directly. I can fix the imminent merge conflict on either branch depending on whichever pull request gets merged earlier.

agriyakhetarpal:
The question would now be about what combination to provide for the wheels: `x86_64` + `arm64` or `x86_64` + `universal2`. The `universal2` wheel supports both platforms but the binary is double the size of that of a specific architecture and therefore requires more bandwidth and storage capacity. For most cases and for ours, this is completely okay, since our wheel size has been around 3 MB which is very small.

We don't need to provide all three—just two of these three choices would be fine, because `pip`  resorts to downloading a universal wheel only when a platform-specific wheel is not available. Also, universal wheels require `pip>=20.3` for installation – the good thing is that Python 3.8+ comes with that version by default. There is some discussion about what version constraints universal wheels can bring up on macOS machines on https://github.com/pypa/cibuildwheel/issues/1333

kratman (commented):
I am a rather delayed on this review, but I did get an error when importing pybamm with the MacOS ARM wheel:
```bash
ModuleNotFoundError: No module named 'pybtex'
```
After installing that I got an error with `anytree` and `sympy`. So I don't think the dependencies are there.

I figure this might change after #3301 is merged.

Saransh-cpp:
That was a bug in `v23.9rc0`, but it was fixed in the `rc1` release. @agriyakhetarpal could you merge develop here?

kratman:
The install and import worked on my M2, but idaklu does not seem to be installed with the wheel.
```bash
pybamm.have_idaklu()
False
```

agriyakhetarpal:
Strange, it works on my end. Could you run the internals of the `have_idaklu()` function step by step and post the output?

kratman:
```
idaklu = importlib.util.module_from_spec(idaklu_spec)
  File "<stdin>", line 1, in <module>
  File "<frozen importlib._bootstrap>", line 565, in module_from_spec
  File "<frozen importlib._bootstrap_external>", line 1173, in create_module
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
ImportError: dlopen(/Users/kratz/anaconda3/envs/testPyBaMM/lib/python3.9/site-packages/pybamm/solvers/idaklu.cpython-39-darwin.so, 0x0002): symbol not found in flat namespace '_IDABBDPrecGetNumGfnEvals'
```

kratman:
I was using python 3.9 if that helps.

agriyakhetarpal:
For very long I tried to reproduce this: across wheels for different Python versions, checked search paths from `otool`, tried different import mechanisms, etc. And I was finally able to. I think the wheel has now been compiled with the correct architecture, but there has been some issue with either the cross-compilation for SuiteSparse/SUNDIALS or there is a bug in the IDAKLU extension module, because the linker is looking at the right paths and there is no version mismatch between the libraries (see below for more). As far as I can understand there is no restriction on not being able to cross-compile either of these libraries with Xcode. Could you help debug what this symbol is supposed to represent? I don't know enough C++ for this :)

I did trace it down to this line here: https://github.com/pybamm-team/PyBaMM/blob/develop/pybamm/solvers/c_solvers/idaklu/CasadiSolverOpenMP.cpp#L498. The symbol is mentioned as a function in the [SUNDIALS documentation for IDA](https://sundials.readthedocs.io/en/latest/ida/Usage/index.html#c.IDABBDPrecGetNumGfnEvals).

Soon after I realised the cause of why it returned `True` for me earlier in the thread: the `.so` file has most probably picked up the natively compiled dynamic libraries I already had present in `.local/` first rather than the cross-compiled ones bundled inside the wheel – it shouldn't do that though because the hierarchy for looking up runtime paths for a repaired wheel is most definitely altered to avoid this `LD_LIBRARY_PATH` problem... deleting those and forcing it to use the right ones brought me to the same error.

I'll try the reverse option now, i.e., cross-compiling for x86_64 on my arm64 machine and then testing whether that wheel works as expected on an x86_64 machine.

agriyakhetarpal:
If cross-compilation doesn't end up working, it might be a good idea to get Cirrus CI for the repository for native compilation. They are freely and readily available and have both macOS arm64 and Linux arm64 runners. SciPy and some other scientific Python libraries are using it: in the case of SciPy it is currently as both a shared resource for PRs' tests and for building wheels for PyPI + nightlies. But if we want to keep its usage to a minimum, our thrice-a-year release cycle we should always be much below the limit for the free plan – tagging @brosaplanella for this. It would be great if this could be added to the agenda for the next developer meeting!  It's also pretty easy to write Cirrus CI pipelines for this task owing to the fact that all CI services have inherently similar YAML syntax, so I can volunteer to pick that up if we do end up adding them.

kratman:
Sure I was planning to work on PyBaMM Friday, so I will take a look and see if I can help out. Debugging linking things is always a pain. Might be easier to build/test in a docker container locally so you are sure nothing exists on your path.

agriyakhetarpal:
Superseded by #3789. We can explore cross-compilation later, possibly during this year's rendition of GSoC or when we have upgraded SUNDIALS and SuiteSparse to newer versions.

PR #3472, opened 2023-10-25
Fixes #3441 call processed variables with custom spatial coordinates

### LINKED ISSUE #3441: Call processed variables with custom spatial coordinates
After https://github.com/pybamm-team/PyBaMM/pull/3234 `ProcessedVariable` objects now automatically take the names of the domains to plot automatically from spatial variables, but `__call__` still expects the spatial variables to be `x`, `y`, `z`, `r`, or `R`, see:
```python3
    def __call__(self, t=None, x=None, r=None, y=None, z=None, R=None, warn=True):
        """
        Evaluate the variable at arbitrary *dimensional* t (and x, r, y, z and/or R),
        using interpolation
        """
        kwargs = {"t": t, "x": x, "r": r, "y": y, "z": z, "R": R}
        # Remove any None arguments
        kwargs = {key: value for key, value in kwargs.items() if value is not None}
        # Use xarray interpolation, return numpy array
        return self._xr_data_array.interp(**kwargs).values
```

We should make it possible to pass extra `kwargs` to `__call__`. 

As a workaround, you can do something like 
```python3
# variable that depends on a  spatial variable named "s" and time "t"
my_var_data_array = sim.solution["My var"]._xr_data_array

def my_var_eval(s, t):
    return my_var_data_array.interp(s=s, t=t).values
```

Saswatsusmoy:
# Description

Modified the __call__ method to accept **kwargs as a parameter. This will allow any additional keyword arguments to be passed to the method. You can then include these extra arguments when calling the interp method of the _xr_data_array object.

Fixes #3441

## Type of change

# Key checklist:

## Further checks:


kratman:
Can you add some tests for this?

Saswatsusmoy:
@rtimms I have tried solving the issue, can you please review the same

tests/unit/test_solvers/test_processed_variable_computed.py
```diff
@@ -65,6 +65,8 @@ def process_and_check_2D_variable(
     #    ::test_output_variables
     return y_sol, first_sol, second_sol, t_sol
 
+def call_func(t=None, x=None, r=None, y=None, z=None, R=None, warn=True, **kwargs):
+    return np.random.rand(5, 5)
```
brosaplanella, line 69: Could you improve the tests so they actually tests that the variables are passed correctly? At the moment it only checks that no error message is thrown.

Saswatsusmoy:
```
def test_call_default(self, mock_call_func):
        call_func()
        mock_call_func.assert_called()

    @patch('path.to.call_func')
    def test_call_with_t(self, mock_call_func):
        t_test = 0.5
        call_func(t=t_test)
        mock_call_func.assert_called_with(t=t_test)
.
.
.
.
```

Something like this? @brosaplanella

brosaplanella: I was thinking something more PyBaMM specific. Maybe something like this example (https://github.com/pybamm-team/PyBaMM/blob/369034fc92a4465ba9d27c290652b69ed0ae78c5/tests/unit/test_solvers/test_processed_variable.py#L139-L163), where a dummy model is created but instead of calling your space variable x, try calling it something that would be a kwarg above to see if it is passed correctly. Maybe call it `a` and try both passing `a` to see it works and passing `b` to check it throws an error.

brosaplanella: I think so. Note that you want `a` to be a `SpatialVariable` rather than an `InputParameter` because the latter needs to be passed at solving stage.

kratman:
@Saswatsusmoy Are you still working on this?

Saswatsusmoy:
> @Saswatsusmoy Are you still working on this?

Yeah, I'll be resuming working on this soon

arjxn-py:
@Saswatsusmoy Feel free to let us know in case you feel stuck with something, Maybe we'd be able to help you out.

agriyakhetarpal:
We have not had a response from the contributor in four months by now. I think we should be fine with closing the PR, we can reopen whenever – especially because the linked issue is labelled as "low priority" at the moment

PR #3475, opened 2023-10-26
Resolve default imports for optional dependencies

### LINKED ISSUE #3473: [Bug]: 23.9rc0 "optional dependencies" imported
### PyBaMM Version

23.9rc0

### Python Version

3.11

### Describe the bug

There are a number of "optional dependencies" that are imported with pybamm on 23.9rc0, causing errors since they are not installed by default. This can obviously be solved by installing `pybamm[all]` but it's a bit weird to me that `pip install pybamm`, then `import pybamm` doesn't just work.

Not sure if this is a "real bug" or if it will be resolved when 23.9 is actually released. Feel free to close it if it will be resolved.

### Steps to Reproduce

shell: `pip install pybamm==23.9rc0`
python: `import pybamm`

### Relevant log output

_No response_

arjxn-py:
# Description

With this PR i'm trying to fix the error getting while importing `pybamm` without any extra dependencies for `pybamm=23.9rc0`

Fixes #3473 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -100,9 +100,9 @@ On the other hand... We _do_ want to compare several tools, to generate document
 
 Only 'core pybamm' is installed by default. The others have to be specified explicitly when running the installation command.
 
-### Matplotlib
+### Managing Optional Dependencies and Their Imports
 
-We use Matplotlib in PyBaMM, but with two caveats:
+PyBaMM utilizes optional dependencies to allow users to choose which additional libraries they want to use. Managing these optional dependencies and their imports is essential to provide flexibility to PyBaMM users.
 
 First, Matplotlib should only be used in plotting methods, and these should _never_ be called by other PyBaMM methods. So users who don't like Matplotlib will not be forced to use it in any way. Use in notebooks is OK and encouraged.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -54,7 +54,7 @@ You now have everything you need to start making changes!
 10. [Test your code!](#testing)
 11. PyBaMM has online documentation at http://docs.pybamm.org/. To make sure any new methods or classes you added show up there, please read the [documentation](#documentation) section.
 12. If you added a major new feature, perhaps it should be showcased in an [example notebook](#example-notebooks).
-13. When you feel your code is finished, or at least warrants serious discussion, run the [pre-commit checks](#pre-commit-checks) and then create a [pull request](https://help.github.com/articles/about-pull-requests/) (PR) on [PyBaMM's GitHub page](https://github.com/pybamm-team/PyBaMM).
+13. When you feel your code is finished, or at least warrants serious discussion, run the [pre-commit checks](#pre-commit-checks) and then create a [pull request](https://help.github.com/articles/about-pull-requests/) (PR) on [PyBaMM&#39;s GitHub page](https://github.com/pybamm-team/PyBaMM).
 14. Once a PR has been created, it will be reviewed by any member of the community. Changes might be suggested which you can make by simply adding new commits to the branch. When everything's finished, someone with the right GitHub permissions will merge your changes into PyBaMM main repository.
 
 Finally, if you really, really, _really_ love developing PyBaMM, have a look at the current [project infrastructure](#infrastructure).
@@ -104,17 +104,25 @@ Only 'core pybamm' is installed by default. The others have to be specified expl
 
 PyBaMM utilizes optional dependencies to allow users to choose which additional libraries they want to use. Managing these optional dependencies and their imports is essential to provide flexibility to PyBaMM users.
 
-First, Matplotlib should only be used in plotting methods, and these should _never_ be called by other PyBaMM methods. So users who don't like Matplotlib will not be forced to use it in any way. Use in notebooks is OK and encouraged.
+PyBaMM provides a utility function `have_optional_dependency`, to check for the availability of optional dependencies within methods. This function can be used to conditionally import optional dependencies only if they are available. Here's how to use it:
 
-Second, Matplotlib should never be imported at the module level, but always inside methods. For example:
+Optional Dependencies should never be imported at the module level, but always inside methods. For example:
 
 ```
-def plot_great_things(self, x, y, z):
-    import matplotlib.pyplot as pl
+def use_pybtex(x,y,z):
+    pybtex = have_optional_dependency("pybtex")
     ...
 ```
 
-This allows people to (1) use PyBaMM without ever importing Matplotlib and (2) configure Matplotlib's back-end in their scripts, which _must_ be done before e.g. `pyplot` is first imported.
+While importing a specific attribute instead of whole module:
+
+```
+def use_parse_file(x,y,z):
+    parse_file = have_optional_dependency("pybtex.database","parse_file")
+    ...
+```
+
+This allows people to (1) use PyBaMM without importing Optional dependency by default and (2) configure module dependent functionality in their scripts, which _must_ be done before e.g. `print_citations` method is first imported.
 
 ## Testing
 
@@ -266,7 +274,6 @@ This also means that, if you can't fix the bug yourself, it will be much easier 
       ```
 
       This will start the debugger at the point where the `ValueError` was raised, and allow you to investigate further. Sometimes, it is more informative to put the try-except block further up the call stack than exactly where the error is raised.
-
    2. Warnings. If functions are raising warnings instead of errors, it can be hard to pinpoint where this is coming from. Here, you can use the `warnings` module to convert warnings to errors:
 
       ```python
@@ -276,7 +283,6 @@ This also means that, if you can't fix the bug yourself, it will be much easier 
       ```
 
       Then you can use a try-except block, as in a., but with, for example, `RuntimeWarning` instead of `ValueError`.
-
    3. Stepping through the expression tree. Most calls in PyBaMM are operations on [expression trees](https://github.com/pybamm-team/PyBaMM/blob/develop/docs/source/examples/notebooks/expression_tree/expression-tree.ipynb). To view an expression tree in ipython, you can use the `render` command:
 
       ```python
@@ -284,11 +290,8 @@ This also means that, if you can't fix the bug yourself, it will be much easier 
       ```
 
       You can then step through the expression tree, using the `children` attribute, to pinpoint exactly where a bug is coming from. For example, if `expression_tree.jac(y)` is failing, you can check `expression_tree.children[0].jac(y)`, then `expression_tree.children[0].children[0].jac(y)`, etc.
-
 3. To isolate whether a bug is in a model, its Jacobian or its simplified version, you can set the `use_jacobian` and/or `use_simplify` attributes of the model to `False` (they are both `True` by default for most models).
-
 4. If a model isn't giving the answer you expect, you can try comparing it to other models. For example, you can investigate parameter limits in which two models should give the same answer by setting some parameters to be small or zero. The `StandardOutputComparison` class can be used to compare some standard outputs from battery models.
-
 5. To get more information about what is going on under the hood, and hence understand what is causing the bug, you can set the [logging](https://realpython.com/python-logging/) level to `DEBUG` by adding the following line to your test or script:
 
    ```python3
--- a/pybamm/__init__.py
+++ b/pybamm/__init__.py
@@ -52,7 +52,13 @@ from .util import (
 )
 from .logger import logger, set_logging_level, get_new_logger
 from .settings import settings
-from .citations import Citations, citations, print_citations
+try:
+    import pybtex
+
+    if pybtex is not None:
+        from .citations import Citations, citations, print_citations
+except ImportError:
+    pass
 
 #
 # Classes for the Expression Tree
--- a/pybamm/__init__.py
+++ b/pybamm/__init__.py
@@ -47,6 +47,7 @@ from .util import (
     get_parameters_filepath,
     have_jax,
     install_jax,
+    have_optional_dependency,
     is_jax_compatible,
     get_git_commit_info,
 )
--- a/pybamm/__init__.py
+++ b/pybamm/__init__.py
@@ -53,14 +53,7 @@ from .util import (
 )
 from .logger import logger, set_logging_level, get_new_logger
 from .settings import settings
-try:
-    import pybtex
-
-    if pybtex is not None:
-        from .citations import Citations, citations, print_citations
-except ImportError:
-    pass
-
+from .citations import Citations, citations, print_citations
 #
 # Classes for the Expression Tree
 #
--- a/pybamm/citations.py
+++ b/pybamm/citations.py
@@ -6,10 +6,10 @@
 import pybamm
 import os
 import warnings
-import pybtex
+# import pybtex
 from sys import _getframe
-from pybtex.database import parse_file, parse_string, Entry
-from pybtex.scanner import PybtexError
+# from pybtex.database import parse_file, parse_string, Entry
+# from pybtex.scanner import PybtexError
 
 
 class Citations:
@@ -76,6 +76,7 @@ class Citations:
         """Reads the citations in `pybamm.CITATIONS.bib`. Other works can be cited
         by passing a BibTeX citation to :meth:`register`.
         """
+        parse_file = pybamm.util.have_optional_dependency("pybtex.database","parse_file")
         citations_file = os.path.join(pybamm.root_dir(), "pybamm", "CITATIONS.bib")
         bib_data = parse_file(citations_file, bib_format="bibtex")
         for key, entry in bib_data.entries.items():
@@ -86,6 +87,7 @@ class Citations:
         previous entry is overwritten
         """
 
+        Entry = pybamm.util.have_optional_dependency("pybtex.database","Entry")
         # Check input types are correct
         if not isinstance(key, str) or not isinstance(entry, Entry):
             raise TypeError()
@@ -151,6 +153,8 @@ class Citations:
         key: str
             A BibTeX formatted citation
         """
+        PybtexError = pybamm.util.have_optional_dependency("pybtex.scanner","PybtexError")
+        parse_string = pybamm.util.have_optional_dependency("pybtex.database","parse_string")
         try:
             # Parse string as a bibtex citation, and check that a citation was found
             bib_data = parse_string(key, bib_format="bibtex")
@@ -177,7 +181,6 @@ class Citations:
             for key, entry in self._citation_tags.items():
                 print(f"{key} was cited due to the use of {entry}")
 
-    @pybamm.util.have_optional_dependency("pybtex")
     def print(self, filename=None, output_format="text", verbose=False):
         """Print all citations that were used for running simulations. The verbose
         option is provided to print tags for citations in the output such that it can
@@ -218,6 +221,7 @@ class Citations:
         """
         # Parse citations that were not known keys at registration, but do not
         # fail if they cannot be parsed
+        pybtex = pybamm.util.have_optional_dependency("pybtex")
         try:
             for key in self._unknown_citations:
                 self._parse_citation(key)
--- a/pybamm/citations.py
+++ b/pybamm/citations.py
@@ -177,6 +177,7 @@ class Citations:
             for key, entry in self._citation_tags.items():
                 print(f"{key} was cited due to the use of {entry}")
 
+    @pybamm.util.have_optional_dependency("pybtex")
     def print(self, filename=None, output_format="text", verbose=False):
         """Print all citations that were used for running simulations. The verbose
         option is provided to print tags for citations in the output such that it can
--- a/pybamm/citations.py
+++ b/pybamm/citations.py
@@ -6,10 +6,8 @@
 import pybamm
 import os
 import warnings
-# import pybtex
 from sys import _getframe
-# from pybtex.database import parse_file, parse_string, Entry
-# from pybtex.scanner import PybtexError
+from pybamm.util import have_optional_dependency
 
 
 class Citations:
@@ -76,7 +74,7 @@ class Citations:
         """Reads the citations in `pybamm.CITATIONS.bib`. Other works can be cited
         by passing a BibTeX citation to :meth:`register`.
         """
-        parse_file = pybamm.util.have_optional_dependency("pybtex.database","parse_file")
+        parse_file = have_optional_dependency("pybtex.database","parse_file")
         citations_file = os.path.join(pybamm.root_dir(), "pybamm", "CITATIONS.bib")
         bib_data = parse_file(citations_file, bib_format="bibtex")
         for key, entry in bib_data.entries.items():
@@ -87,7 +85,7 @@ class Citations:
         previous entry is overwritten
         """
 
-        Entry = pybamm.util.have_optional_dependency("pybtex.database","Entry")
+        Entry = have_optional_dependency("pybtex.database","Entry")
         # Check input types are correct
         if not isinstance(key, str) or not isinstance(entry, Entry):
             raise TypeError()
@@ -153,8 +151,8 @@ class Citations:
         key: str
             A BibTeX formatted citation
         """
-        PybtexError = pybamm.util.have_optional_dependency("pybtex.scanner","PybtexError")
-        parse_string = pybamm.util.have_optional_dependency("pybtex.database","parse_string")
+        PybtexError = have_optional_dependency("pybtex.scanner","PybtexError")
+        parse_string = have_optional_dependency("pybtex.database","parse_string")
         try:
             # Parse string as a bibtex citation, and check that a citation was found
             bib_data = parse_string(key, bib_format="bibtex")
@@ -221,7 +219,7 @@ class Citations:
         """
         # Parse citations that were not known keys at registration, but do not
         # fail if they cannot be parsed
-        pybtex = pybamm.util.have_optional_dependency("pybtex")
+        pybtex = have_optional_dependency("pybtex")
         try:
             for key in self._unknown_citations:
                 self._parse_citation(key)
--- a/pybamm/expression_tree/functions.py
+++ b/pybamm/expression_tree/functions.py
@@ -4,7 +4,6 @@
 import numbers
 
 import numpy as np
-import sympy
 from scipy import special
 
 import pybamm
@@ -202,6 +201,7 @@ class Function(pybamm.Symbol):
 
     def to_equation(self):
         """Convert the node and its subtree into a SymPy equation."""
+        sympy = have_optional_dependency("sympy")
         if self.print_name is not None:
             return sympy.Symbol(self.print_name)
         else:
@@ -250,6 +250,7 @@ class SpecificFunction(Function):
 
     def _sympy_operator(self, child):
         """Apply appropriate SymPy operators."""
+        sympy = have_optional_dependency("sympy")
         class_name = self.__class__.__name__.lower()
         sympy_function = getattr(sympy, class_name)
         return sympy_function(child)
@@ -267,6 +268,7 @@ class Arcsinh(SpecificFunction):
 
     def _sympy_operator(self, child):
         """Override :meth:`pybamm.Function._sympy_operator`"""
+        sympy = have_optional_dependency("sympy")
         return sympy.asinh(child)
 
 
@@ -287,6 +289,7 @@ class Arctan(SpecificFunction):
 
     def _sympy_operator(self, child):
         """Override :meth:`pybamm.Function._sympy_operator`"""
+        sympy = have_optional_dependency("sympy")
         return sympy.atan(child)
 
 
--- a/pybamm/expression_tree/functions.py
+++ b/pybamm/expression_tree/functions.py
@@ -3,7 +3,10 @@
 #
 import numbers
 
-import autograd
+try:
+    import autograd
+except ImportError:
+    pass
 import numpy as np
 import sympy
 from scipy import special
--- a/pybamm/expression_tree/functions.py
+++ b/pybamm/expression_tree/functions.py
@@ -3,16 +3,12 @@
 #
 import numbers
 
-try:
-    import autograd
-except ImportError:
-    pass
 import numpy as np
 import sympy
 from scipy import special
 
 import pybamm
-
+from pybamm.util import have_optional_dependency
 
 class Function(pybamm.Symbol):
     """
@@ -99,6 +95,7 @@ class Function(pybamm.Symbol):
         Derivative with respect to child number 'idx'.
         See :meth:`pybamm.Symbol._diff()`.
         """
+        autograd = have_optional_dependency("autograd")
         # Store differentiated function, needed in case we want to convert to CasADi
         if self.derivative == "autograd":
             return Function(
--- a/pybamm/expression_tree/printing/sympy_overrides.py
+++ b/pybamm/expression_tree/printing/sympy_overrides.py
@@ -3,9 +3,10 @@
 #
 import re
 
-from sympy.printing.latex import LatexPrinter
+from pybamm.util import have_optional_dependency
 
 
+LatexPrinter = have_optional_dependency("sympy.printing.latex","LatexPrinter")
 class CustomPrint(LatexPrinter):
     """Override SymPy methods to match PyBaMM's requirements"""
 
@@ -21,4 +22,5 @@ class CustomPrint(LatexPrinter):
 
 
 def custom_print_func(expr, **settings):
+    have_optional_dependency("sympy.printing.latex","LatexPrinter")
     return CustomPrint().doprint(expr)
--- a/pybamm/expression_tree/symbol.py
+++ b/pybamm/expression_tree/symbol.py
@@ -3,20 +3,13 @@
 #
 import numbers
 
-
-try:
-    import anytree
-    from anytree.exporter import DotExporter
-except ImportError:
-    _has_anytree = False
-else:
-    _has_anytree = True
 import numpy as np
 import sympy
 from scipy.sparse import csr_matrix, issparse
 from functools import lru_cache, cached_property
 
 import pybamm
+from pybamm.util import have_optional_dependency
 from pybamm.expression_tree.printing.print_name import prettify_print_name
 
 DOMAIN_LEVELS = ["primary", "secondary", "tertiary", "quaternary"]
@@ -448,8 +441,7 @@ class Symbol:
         """
         Print out a visual representation of the tree (this node and its children)
         """
-        if not _has_anytree:
-            raise ImportError("Module 'anytree' is required to do this")
+        anytree = have_optional_dependency("anytree")
         for pre, _, node in anytree.RenderTree(self):
             if isinstance(node, pybamm.Scalar) and node.name != str(node.value):
                 print("{}{} = {}".format(pre, node.name, node.value))
@@ -467,9 +459,8 @@ class Symbol:
         filename : str
             filename to output, must end in ".png"
         """
-        if not _has_anytree:
-            raise ImportError("Module 'anytree' is required to do this")
 
+        DotExporter = have_optional_dependency("anytree.exporter","DotExporter")
         # check that filename ends in .png.
         if filename[-4:] != ".png":
             raise ValueError("filename should end in .png")
@@ -489,8 +480,7 @@ class Symbol:
         Finds all children of a symbol and assigns them a new id so that they can be
         visualised properly using the graphviz output
         """
-        if not _has_anytree:
-            raise ImportError("Module 'anytree' is required to do this")
+        anytree = have_optional_dependency("anytree")
         name = symbol.name
         if name == "div":
             name = "&nabla;&sdot;"
@@ -534,8 +524,7 @@ class Symbol:
         a
         b
         """
-        if not _has_anytree:
-            raise ImportError("Module 'anytree' is required to do this")
+        anytree = have_optional_dependency("anytree")
         return anytree.PreOrderIter(self)
 
     def __str__(self):
--- a/pybamm/expression_tree/symbol.py
+++ b/pybamm/expression_tree/symbol.py
@@ -3,10 +3,14 @@
 #
 import numbers
 
-import anytree
+
+try:
+    import anytree
+    from anytree.exporter import DotExporter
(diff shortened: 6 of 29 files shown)


pybamm/expression_tree/functions.py
```diff
@@ -3,7 +3,10 @@
 #
 import numbers
 
-import autograd
+try:
+    import autograd
+except ImportError:
+    pass
```
Saransh-cpp, line 9: It would be better to do something like this (for all such `try-except` blocks -
```py
try:
     import autograd
 except ImportError:
      warnings.warn("Install autograd for ... . See <doc_link> for more details.")
```

pybamm/__init__.py
```diff
@@ -52,7 +52,13 @@
 )
 from .logger import logger, set_logging_level, get_new_logger
 from .settings import settings
-from .citations import Citations, citations, print_citations
+try:
+    import pybtex
+
+    if pybtex is not None:
```
Saransh-cpp, line 58: You should use importlib to check if pybtex is installed -
```suggestion
    if importlib.util.find_spec("pybtex") is not None:
```

Saransh-cpp (changes requested):
Thanks, @arjxn-py! Some comments below -

pybamm/expression_tree/functions.py
agriyakhetarpal: Another option is to import optional dependencies inside methods only and not as top-level imports in files (there is a mention of this for `matplotlib` in the contributing guide). The guide should be updated too. We can add a new section about how to handle optional dependencies (about both of these things: adding new dependencies, and how to use them and existing ones while adding new code, regardless of whichever way this PR ends up implementing this mechanism).

arjxn-py: Thanks both, though I was planning to do it this way - https://stackoverflow.com/questions/27361427 
The imports stay top level but with a handling when specific function is called.
The method @Saransh-cpp suggested should give this warning everytime pybamm is imported but I think it would be more ideal to throw a warning when the function with specific optional dependency is called.

agriyakhetarpal:
https://github.com/pandas-dev/pandas/blob/984d75543fce5c68f0dcc4f0b3500256e4a9c0c9/pandas/io/pytables.py#L566

This is how `pandas` does it, we could add a function with similar functionality in `pybamm.util`. These are design choices imo after all, but a write-once-use-anywhere paradigm would be useful in this scenario. For example, adding a decorator would be great such that it can be used on a class (or at its instantiation, not recommended however) which would automatically check if an optional dependency can be imported or not, and would apply to all subsequent functions defined inside it. I am not sure how one would go about implementing it, but something like

```python
@pybamm._check_and_import_optional_dependency(name="pybtex")
class Citations:
   
   def print_citation(...):
    ...

```

This way, it would raise a warning plus a helpful message if `pybtex` cannot be found, but load it into the namespace if it can be found. We currently import the `print_citations()` method directly, which means the decorator would apply to the method only and not throw duplicate warnings. Python imports are cached very effectively which means there is negligible overhead if we import a dependency multiple times inside the same file or within a module.

Although these function-scoped imports are also a great idea btw. We already do it for `matplotlib` in the QuickPlot classes and methods; the good thing is that we don't add optional dependencies often, which means this is going to be a one-time thing and maintaining such imports would not be difficult (`skfem` for the SpatialMethod module, `bpx` for BPX, etc.)

arjxn-py:
I've finally managed to make a decorator function but there are complications with it : 

- If applied `pybamm.util.have_optional_dependency` will be executed whenever `pybamm` is imported. This behavior is expected since decorators are typically execute when the class is imported.
- I am having trouble to make module available within the local scope of the method & i'm not sure if this is possible with a decorator function or not.

I'll also put in some efforts to do it with simple reusable function first.

agriyakhetarpal:
> I've finally managed to make a decorator function but there are complications with it : 
> 
> - If applied `pybamm.util.have_optional_dependency` will be executed whenever `pybamm` is imported. This behavior is expected since decorators are typically execute when the class is imported.

I think this should be fine, the decorator should just execute "silently" and return nothing if a dependency is found. Is there a way to "lazy-load" an entire module at the time of import, e.g., to receive no warnings when doing `import pybamm` but to receive a warning about `pybtex` only if someone executes any of the functionality under `pybamm.citations.` or in other words tries to instantiate the Citations class? Doing that it would make `import pybamm` faster, even, I would say, since certain modules would be skipped.

agriyakhetarpal:
> I am having trouble to make module available within the local scope of the method & i'm not sure if this is possible with a decorator function or not.

In that case, it would be fine to do this for methods only and not for classes: this practice won't conform to DRY, but atleast it won't apply to private methods available under a class (we want to mark just those pieces of functionality which are available as a part of the public API)

agriyakhetarpal:
See also: https://scientific-python.org/specs/spec-0001/ (this demands a separate issue, I guess)

Saransh-cpp:
> See also: https://scientific-python.org/specs/spec-0001/ (this demands a separate issue, I guess)

Yes, separate and important issue. Importing pybamm usually takes relatively more time than other libraries, and I think this should fix that.

arjxn-py:
> See also: https://scientific-python.org/specs/spec-0001/ (this demands a separate issue, I guess)

Yes, happy to open one issue for it & also look into the same, @agriyakhetarpal do let me know if you want to open with some suggestion or simplifications.

agriyakhetarpal:
I see that there are some issues with building the documentation (there is a possibility it _might_ break some of our internal Sphinx extensions that rely on `autodoc`), though I think we can discuss this further on a new issue thread. Some other simplications could be to reorganise top-level imports without an expense to the current API (i.e., no breaking changes).

pybamm/util.py
```diff
@@ -345,3 +345,21 @@ def install_jax(arguments=None):  # pragma: no cover
...
+            if hasattr(module, attribute):
+                imported_attribute = getattr(module, attribute)
+                return imported_attribute
+            else:
+                raise ImportError(f"Optional dependency {module_name}.{attribute} is not available. See https://docs.pybamm.org/en/latest/source/user_guide/installation/index.html#optional-dependencies for more details.")
+        else:
+            return module
+    except ImportError:
+        if attribute:
+            raise ImportError(f"Optional dependency {module_name}.{attribute} is not available. See https://docs.pybamm.org/en/latest/source/user_guide/installation/index.html#optional-dependencies for more details.")
```
agriyakhetarpal, line 363:
```suggestion
            raise ImportError(f"Optional dependency {module_name} is not available. See https://docs.pybamm.org/en/latest/source/user_guide/installation/index.html#optional-dependencies for more details.")
```

I feel users don't need to know the submodule we are using from a particular optional dependency, I feel that just the name of the dependency in the error message should be enough information.

Also, is it possible to directly embed the installation command in the error message? Like, is it possible to programmatically access the optional dependencies from `setup.py` and soon `pyproject.toml`? We can then ask users to install `pip install pybamm[cite]` if `pybtex` isn't found, and similarly for the other dependencies. Either way, the link will be descriptive enough so we can skip doing this if it isn't feasible without the addition of an extra dictionary (it just becomes another thing we need to maintain otherwise).

pybamm/util.py
```diff
@@ -345,3 +345,21 @@ def install_jax(arguments=None):  # pragma: no cover
...
     )
+
+
+def have_optional_dependency(module_name, attribute=None):
+    try:
+        module = importlib.import_module(module_name)
+        if attribute:
+            if hasattr(module, attribute):
+                imported_attribute = getattr(module, attribute)
+                return imported_attribute
```
agriyakhetarpal, line 356: Just needs some tests in `test_util.py` in order to fix the coverage. A good test could be to not import a specific dependency (or the module contained inside it) and test some specific function or class that uses that dependency and it should raise an ImportError (which can be checked using `assertRaisesRegex`)

brosaplanella:
Are the mac failing tests related to this issue?

agriyakhetarpal:
They seem to be unrelated, some macOS GitHub Actions runners do not have `gfortran` which is required by `scipy` (it should come with `gcc` by default, but the error seems to be sporadic). I will soon open a PR to fix those.

pybamm/util.py
```diff
@@ -345,3 +345,28 @@ def install_jax(arguments=None):  # pragma: no cover
...
+                imported_attribute = getattr(module, attribute)
+                return imported_attribute  # Return the imported attribute
+            else:
+                # Raise an ImportError if the attribute is not available
+                raise ImportError(f"Optional dependency {module_name} is not available. See https://docs.pybamm.org/en/latest/source/user_guide/installation/index.html#optional-dependencies for more details.")
+        else:
+            # Return the entire module if no attribute is specified
+            return module
+
+    except ImportError:
```
agriyakhetarpal, line 367:
```suggestion
    except ModuleNotFoundError:
```

`importlib.import_module` raises `ModuleNotFoundError` instead of `ImportError` (I tested this locally)

pybamm/util.py
```diff
@@ -345,3 +345,28 @@ def install_jax(arguments=None):  # pragma: no cover
...
+        module = importlib.import_module(module_name)
+
+        if attribute:
+            # If an attribute is specified, check if it's available
+            if hasattr(module, attribute):
+                imported_attribute = getattr(module, attribute)
+                return imported_attribute  # Return the imported attribute
+            else:
+                # Raise an ImportError if the attribute is not available
+                raise ImportError(f"Optional dependency {module_name} is not available. See https://docs.pybamm.org/en/latest/source/user_guide/installation/index.html#optional-dependencies for more details.")
```
agriyakhetarpal, line 362:
```suggestion
                raise ModuleNotFoundError(f"Optional dependency {module_name} is not available. See https://docs.pybamm.org/en/latest/source/user_guide/installation/index.html#optional-dependencies for more details.")
```

pybamm/util.py
```diff
@@ -345,3 +345,28 @@ def install_jax(arguments=None):  # pragma: no cover
...
+        else:
+            # Return the entire module if no attribute is specified
+            return module
+
+    except ImportError:
+        # Raise an ImportError if the module or attribute is not available
+        if attribute:
```
agriyakhetarpal, line 369: If the dependency itself is not available, do we even need to have a check for the attribute? I think it can be avoided

pybamm/util.py
agriyakhetarpal: I replied to this here: https://github.com/pybamm-team/PyBaMM/pull/3475#discussion_r1387062979

tests/unit/test_util.py
```diff
@@ -88,6 +88,10 @@ def test_git_commit_info(self):
         self.assertIsInstance(git_commit_info, str)
         self.assertEqual(git_commit_info[:2], "v2")
 
+    def test_have_optional_dependency(self):
+        with self.assertRaisesRegex(ImportError,"Optional dependency pybtex is not available. See https://docs.pybamm.org/en/latest/source/user_guide/installation/index.html#optional-dependencies for more details."):
+            pybamm.print_citations()
+
```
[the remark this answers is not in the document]
arjxn-py: > We should do this in a way that it covers all of the current optional dependencies.

As we are testing `have_optional_dependency`, I think it should be sufficient with a couple of dependencies to test the functionality for now. Plus other dependencies have more complex implementations which is not ideal looking at the time constrains for this PR.

> we should cover a case where a required optional dependency is available and can be imported (where it should not raise an ImportError).

Added one, thanks for suggesting.

pybamm/util.py
```diff
@@ -345,3 +345,25 @@ def install_jax(arguments=None):  # pragma: no cover
...
+        module = importlib.import_module(module_name)
+
+        if attribute:
+            # If an attribute is specified, check if it's available
+            if hasattr(module, attribute):
+                imported_attribute = getattr(module, attribute)
+                return imported_attribute  # Return the imported attribute
+            else:
+                # Raise an ModuleNotFoundError if the attribute is not available
+                raise ModuleNotFoundError(f"Optional dependency {module_name} is not available. See https://docs.pybamm.org/en/latest/source/user_guide/installation/index.html#optional-dependencies for more details.")
```
brosaplanella, line 362: This line is not tested. If it cannot be reached to test, add a `# pragma: no cover`

tests/unit/test_util.py
```diff
@@ -88,6 +93,33 @@ def test_git_commit_info(self):
         self.assertIsInstance(git_commit_info, str)
         self.assertEqual(git_commit_info[:2], "v2")
 
+    def test_have_optional_dependency(self):
```
brosaplanella, line 96: For the coverage missing, you might want to call the `have_optional_dependency` method directly.

brosaplanella (approved):
Looks good to me once coverage is addressed. Thanks Arjun for the great work here! :)

tests/unit/test_util.py
```diff
@@ -10,8 +10,12 @@
 import unittest
 from unittest.mock import patch
 from io import StringIO
+from tempfile import TemporaryDirectory
 
+def test_function(arg):
```
agriyakhetarpal, line 15: I think this is not being used anywhere, or is it?

tests/unit/test_util.py
```diff
@@ -10,8 +10,12 @@
...
+    return arg + arg
 
+anytree = sys.modules['anytree']
```
agriyakhetarpal, line 18: This line is repeated at line 36 for some reason which I don't follow. Do we necessarily require `sys.modules['anytree']` there, or can we import `anytree` with `have_optional_dependency` there?

tests/unit/test_util.py
```diff
@@ -29,6 +33,7 @@ def test_rmse(self):
             pybamm.rmse(np.ones(5), np.zeros(3))
 
     def test_is_constant_and_can_evaluate(self):
+        sys.modules['anytree'] = anytree
```
agriyakhetarpal, line 36: i.e., here

tests/unit/test_util.py
```diff
@@ -88,6 +93,33 @@ def test_git_commit_info(self):
         self.assertIsInstance(git_commit_info, str)
         self.assertEqual(git_commit_info[:2], "v2")
 
+    def test_have_optional_dependency(self):
+        with self.assertRaisesRegex(ModuleNotFoundError,"Optional dependency pybtex is not available. See https://docs.pybamm.org/en/latest/source/user_guide/installation/index.html#optional-dependencies for more details."):
+            pybtex = sys.modules['pybtex']
+            sys.modules['pybtex'] = None
+            pybamm.print_citations()
+        with self.assertRaisesRegex(ModuleNotFoundError,"Optional dependency tqdm is not available. See https://docs.pybamm.org/en/latest/source/user_guide/installation/index.html#optional-dependencies for more details."):
```
agriyakhetarpal, line 101: These tests look good, just missing some coverage as Ferran mentions. Instead of hard-coding the name of the dependency, is it possible to mock the dependencies in one go and use f-strings? It might reduce some of the boilerplate in these assert statements.

Also, could you add a small sentence or two highlighting this to the Contributing guide? i.e., whenever one adds a dependency for optional functionality, they should write a corresponding unit test in `test_utilpy` to ensure that an error is raised upon the absence of said dependency.

pybamm/citations.py
```diff
@@ -76,6 +74,7 @@ def read_citations(self):
         """Reads the citations in `pybamm.CITATIONS.bib`. Other works can be cited
         by passing a BibTeX citation to :meth:`register`.
         """
+        parse_file = have_optional_dependency("pybtex.database","parse_file")
```
Saransh-cpp, line 77:
```suggestion
        parse_file = have_optional_dependency("pybtex.database", "parse_file")
```

pybamm/citations.py
```diff
@@ -86,6 +85,7 @@ def _add_citation(self, key, entry):
         previous entry is overwritten
         """
 
+        Entry = have_optional_dependency("pybtex.database","Entry")
```
Saransh-cpp, line 88:
```suggestion
        Entry = have_optional_dependency("pybtex.database", "Entry")
```

pybamm/citations.py
```diff
@@ -151,6 +151,8 @@ def _parse_citation(self, key):
         key: str
             A BibTeX formatted citation
         """
+        PybtexError = have_optional_dependency("pybtex.scanner","PybtexError")
+        parse_string = have_optional_dependency("pybtex.database","parse_string")
```
Saransh-cpp, line 155:
```suggestion
        PybtexError = have_optional_dependency("pybtex.scanner", "PybtexError")
        parse_string = have_optional_dependency("pybtex.database", "parse_string")
```

pybamm/expression_tree/printing/sympy_overrides.py
```diff
@@ -3,9 +3,10 @@
 #
 import re
 
-from sympy.printing.latex import LatexPrinter
+from pybamm.util import have_optional_dependency
 
 
+LatexPrinter = have_optional_dependency("sympy.printing.latex","LatexPrinter")
```
Saransh-cpp, line 9:
```suggestion
LatexPrinter = have_optional_dependency("sympy.printing.latex", "LatexPrinter")
```

pybamm/expression_tree/printing/sympy_overrides.py
```diff
@@ -21,4 +22,5 @@ def _print_Derivative(self, expr):
 
 
 def custom_print_func(expr, **settings):
+    have_optional_dependency("sympy.printing.latex","LatexPrinter")
```
Saransh-cpp, line 25:
```suggestion
    have_optional_dependency("sympy.printing.latex", "LatexPrinter")
```

pybamm/expression_tree/symbol.py
```diff
@@ -460,6 +459,7 @@ def visualise(self, filename):
             filename to output, must end in ".png"
         """
 
+        DotExporter = have_optional_dependency("anytree.exporter","DotExporter")
```
Saransh-cpp, line 462:
```suggestion
        DotExporter = have_optional_dependency("anytree.exporter", "DotExporter")
```

pybamm/util.py
Saransh-cpp, line 362: The message string should be refactored outside the `try-except` block, something like this -
```py
def have_optional_dependency(module_name, attribute=None):
  err_msg = f"Optional dependency {module_name} ...."
  try:
  ...
  raise ModuleNotFoundError(err_msg)
  ...
  raise ModuleNotFoundError(err_msg)
```

This -
1. Avoids repetition in code
2. For longer error messages, the message string itself should not be passed to the error class, rather a variable should be passed in for better readability.
  For example -
    ```py
    sub = "Some value"
    raise RuntimeError(f"{sub!r} is incorrect")
    ```
    prints
    ```py
      File "tmp.py", line 2, in <module>
        raise RuntimeError(f"{sub!r} is incorrect")
    RuntimeError: 'Some value' is incorrect
    ```
    See how the error message is repeated twice. The readability gets worse as we increase the size of the string.

tests/unit/test_util.py
```diff
@@ -88,6 +93,33 @@ def test_git_commit_info(self):
...
+            sys.modules['tqdm'] = None
+            model = pybamm.BaseModel()
+            v = pybamm.Variable("v")
+            model.rhs = {v: -v}
+            model.initial_conditions = {v: 1}
+            sim = pybamm.Simulation(model)
+            sim.solve([0, 1])
+        with self.assertRaisesRegex(ModuleNotFoundError,"Optional dependency anytree is not available. See https://docs.pybamm.org/en/latest/source/user_guide/installation/index.html#optional-dependencies for more details."):
```
Saransh-cpp, line 109: IIRC writing down the complete string is not required while testing, instead, you can get away with just writing a substring. In this case, just writing `"Optional dependency anytree is not available."` should work (and look much more readable).

Saransh-cpp (changes requested):
This looks great, thanks a ton for working on this @arjxn-py!

The comments below are mostly related to best practices, making the output and the written code a bit cleaner. I will be more than happy to push the suggestions myself if you are not available. Please feel free to ping me anytime!

tests/unit/test_util.py
arjxn-py: Oh yes, now it is not being used, removing it in next commit.

tests/unit/test_util.py
arjxn-py: Yes i felt this is necessary as if we don't do it `test_is_constant_and_can_evaluate` fails with:
```
ERROR: test_is_constant_and_can_evaluate (test_util.TestUtil)
----------------------------------------------------------------------
  File "/home/runner/work/PyBaMM/PyBaMM/pybamm/util.py", line 353, in have_optional_dependency
    module = importlib.import_module(module_name)
  File "/usr/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1032, in _find_and_load
ModuleNotFoundError: import of anytree halted; None in sys.modules

During handling of the above exception, another exception occurred:

  File "/home/runner/work/PyBaMM/PyBaMM/tests/testcase.py", line 29, in wrapped
    return method(*args, **kwargs)
  File "/home/runner/work/PyBaMM/PyBaMM/tests/unit/test_util.py", line 34, in test_is_constant_and_can_evaluate
    symbol = pybamm.PrimaryBroadcast(0, "negative electrode")
  File "/home/runner/work/PyBaMM/PyBaMM/pybamm/expression_tree/broadcasts.py", line 76, in __init__
    child = pybamm.Scalar(child)
  File "/home/runner/work/PyBaMM/PyBaMM/pybamm/expression_tree/scalar.py", line 29, in __init__
    super().__init__(name)
  File "/home/runner/work/PyBaMM/PyBaMM/pybamm/expression_tree/symbol.py", line 231, in __init__
    for x in self.pre_order()
  File "/home/runner/work/PyBaMM/PyBaMM/pybamm/expression_tree/symbol.py", line 526, in pre_order
    anytree = have_optional_dependency("anytree")
  File "/home/runner/work/PyBaMM/PyBaMM/pybamm/util.py", line 369, in have_optional_dependency
    raise ModuleNotFoundError(f"Optional dependency {module_name} is not available. See https://docs.pybamm.org/en/latest/source/user_guide/installation/index.html#optional-dependencies for more details.")
ModuleNotFoundError: Optional dependency anytree is not available. See https://docs.pybamm.org/en/latest/source/user_guide/installation/index.html#optional-dependencies for more details.
``` 

> can we import anytree with have_optional_dependency there?

I tried doing this but it was failing with the same logs.

agriyakhetarpal: Okay, it should be fine to keep this. We can look into neater solutions later

tests/unit/test_util.py
arjxn-py: Done, didn't know substrings could also work. Thanks :)

pybamm/util.py
arjxn-py: Done, Got it 🚀

tests/unit/test_util.py
arjxn-py: Added a case to potentially fix the coverage with direct usage of this in [L-122](https://github.com/pybamm-team/PyBaMM/pull/3475/files#diff-213a36200d41e6897aaba82ed856a037dcf622587acf920011b95c64d3428338R122), thanks.

tests/unit/test_util.py
arjxn-py: > Instead of hard-coding the name of the dependency, is it possible to mock the dependencies in one go and use f-strings?

I also thought about doing the same but because each dependency has distinct implementation in terms of functions & complexity(as we are not just checking the availability of a specific dependency but testing the dependent functions), I found myself unable to do that. We can try doing this afterwards too, happy to have your suggestions.

agriyakhetarpal: Sounds good to me in that case. The error message has been stored as a variable, so checking each dependency is not a big deal, given that we don't have a lot of dependencies.

tests/unit/test_util.py
```diff
@@ -88,6 +91,36 @@ def test_git_commit_info(self):
...
+                c = pybamm.Variable("c", "negative electrode")
+                d = pybamm.Variable("d", "negative electrode")
+                sym = pybamm.div(c * pybamm.grad(c)) + (c / d + c - d) ** 5
+                sym.visualise(test_name)
+
+        sys.modules['pybtex'] = pybtex
+        pybamm.util.have_optional_dependency("pybtex")
+        pybamm.print_citations()
+
+        with self.assertRaisesRegex(ModuleNotFoundError,"Optional dependency flask is not available."):
```
agriyakhetarpal, line 121:
```suggestion
```

I guess you forgot to remove this

tests/unit/test_util.py
agriyakhetarpal, line 1: We should mock and thereby test all optional dependencies in this file if possible

pybamm/expression_tree/printing/sympy_overrides.py
agriyakhetarpal, line 1: I pulled your branch and tested the changes locally, it appears that `import pybamm` is still breaking and it comes from this file. I am assuming that this is because `LatexPrinter` is defined and checked outside of the `CustomPrint` class definition.

I installed `pybamm[latexify]` after that and tested the changes again—the import was successful afterwards, so this should be the last hurdle to pass. We should be able to counter cases like that if all optional dependencies are tested.

agriyakhetarpal (changes requested):
Thanks, @arjxn-py. This PR is almost there!

pybamm/expression_tree/printing/sympy_overrides.py
arjxn-py: I see, `CustomPrint` class is inheriting `LatexPrinter` hence if i define it inside class it gives undefined error.

tests/unit/test_util.py
arjxn-py: Marking it as resolved for now, we can open a different issue or i'll straight up open a PR.

agriyakhetarpal (commented):
Thanks, @arjxn-py! It looks like the `sympy_overrides` import has been resolved and `import pybamm` works, but I am facing other errors, i.e., upon initialising any of the models:

#### To reproduce

```python
import pybamm
```

#### Error log

```ipython
>>> model = pybamm.lithium_ion.DFN()
  File "/Users/agriyakhetarpal/Desktop/PyBaMM/pybamm/util.py", line 354, in have_optional_dependency
    module = importlib.import_module(module_name)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.11/3.11.6_1/Frameworks/Python.framework/Versions/3.11/lib/python3.11/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1140, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'anytree'

During handling of the above exception, another exception occurred:

  File "<stdin>", line 1, in <module>
  File "/Users/agriyakhetarpal/Desktop/PyBaMM/pybamm/models/full_battery_models/lithium_ion/dfn.py", line 28, in __init__
    self.set_submodels(build)
  File "/Users/agriyakhetarpal/Desktop/PyBaMM/pybamm/models/full_battery_models/lithium_ion/base_lithium_ion_model.py", line 59, in set_submodels
    self.build_model()
  File "/Users/agriyakhetarpal/Desktop/PyBaMM/pybamm/models/full_battery_models/base_battery_model.py", line 1033, in build_model
    self._build_model()
  File "/Users/agriyakhetarpal/Desktop/PyBaMM/pybamm/models/base_model.py", line 557, in _build_model
    self.build_model_equations()
  File "/Users/agriyakhetarpal/Desktop/PyBaMM/pybamm/models/full_battery_models/base_battery_model.py", line 1023, in build_model_equations
    submodel.set_initial_conditions(self.variables)
  File "/Users/agriyakhetarpal/Desktop/PyBaMM/pybamm/models/submodels/particle/fickian_diffusion.py", line 326, in set_initial_conditions
    self.initial_conditions = {c_s: c_init}
    ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/agriyakhetarpal/Desktop/PyBaMM/pybamm/models/base_model.py", line 156, in initial_conditions
    self._initial_conditions = EquationDict(
                               ^^^^^^^^^^^^^
  File "/Users/agriyakhetarpal/Desktop/PyBaMM/pybamm/models/base_model.py", line 1143, in __init__
    equations = self.check_and_convert_equations(equations)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/agriyakhetarpal/Desktop/PyBaMM/pybamm/models/base_model.py", line 1177, in check_and_convert_equations
    if eqn.has_symbol_of_classes(pybamm.Variable):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/agriyakhetarpal/Desktop/PyBaMM/pybamm/expression_tree/symbol.py", line 876, in has_symbol_of_classes
    return any(isinstance(symbol, symbol_classes) for symbol in self.pre_order())
                                                                ^^^^^^^^^^^^^^^^
  File "/Users/agriyakhetarpal/Desktop/PyBaMM/pybamm/expression_tree/symbol.py", line 526, in pre_order
    anytree = have_optional_dependency("anytree")
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/agriyakhetarpal/Desktop/PyBaMM/pybamm/util.py", line 370, in have_optional_dependency
    raise ModuleNotFoundError(err_msg)
ModuleNotFoundError: Optional dependency anytree is not available. See https://docs.pybamm.org/en/latest/source/user_guide/installation/index.html#optional-dependencies for more details.
```

I think we should hold off on approving this and keep testing basic functionalities in the meantime with no optional dependencies until everything is fine. It is difficult to catch such errors in the unit tests because we are currently installing all optional dependencies as is expected. Right now, the _truly_ optional dependencies for PyBaMM are the `scikits.odes` and `jax` solver, so (maybe?) we should install a bare-bones rendition of PyBaMM in CI temporarily and use `@unittest.skipIf` decorators for all optional dependencies

It does not count towards great user experience, API-wise, where `import pybamm` works but `pybamm.lithium_ion.SPMe()` etc. breaks (cc: @brosaplanella, @Saransh-cpp). The other ways around this issue overall are to 1. revert the move towards optional dependencies if there is an urgency for a release and bring them back with a patch release, or 2. to recommend users to install `pybamm[all]` for now to ensure everything works; and the latter option defeats the purpose of optional dependencies

arjxn-py:
Looks like `anytree` is essential while initialising any of the models, Can you once try just after installing `anytree` locally? This way we can decide on making it one of the required dependencies.

agriyakhetarpal:
> Looks like `anytree` is essential while initialising any of the models, Can you once try just after installing `anytree` locally? This way we can decide on making it one of the required dependencies.

Initialising a model worked after installing `anytree`. I then initialised a Simulation object and tried to solve and plot for an hour, which reported an error about `tqdm` being missing. I think `tqdm` is just for progress bars so it can be caught with a small `if-else` condition in `Simulation.solve()`

agriyakhetarpal:
A good benchmark to see what can be fixable at this time could be to run all of the example scripts and see what features end up breaking. Similarly, the example notebooks should work with just `pybamm[plot,cite]` in an ideal situation (but probably won't because we have been installing the `[all]` set of extras in the examples `nox` session)

CONTRIBUTING.md
```diff
@@ -54,7 +54,7 @@ You now have everything you need to start making changes!
 10. [Test your code!](#testing)
 11. PyBaMM has online documentation at http://docs.pybamm.org/. To make sure any new methods or classes you added show up there, please read the [documentation](#documentation) section.
 12. If you added a major new feature, perhaps it should be showcased in an [example notebook](#example-notebooks).
-13. When you feel your code is finished, or at least warrants serious discussion, run the [pre-commit checks](#pre-commit-checks) and then create a [pull request](https://help.github.com/articles/about-pull-requests/) (PR) on [PyBaMM's GitHub page](https://github.com/pybamm-team/PyBaMM).
+13. When you feel your code is finished, or at least warrants serious discussion, run the [pre-commit checks](#pre-commit-checks) and then create a [pull request](https://help.github.com/articles/about-pull-requests/) (PR) on [PyBaMM&#39;s GitHub page](https://github.com/pybamm-team/PyBaMM).
```
Saransh-cpp, line 57:
```suggestion
13. When you feel your code is finished, or at least warrants serious discussion, run the [pre-commit checks](#pre-commit-checks) and then create a [pull request](https://help.github.com/articles/about-pull-requests/) (PR) on [PyBaMM's GitHub page](https://github.com/pybamm-team/PyBaMM).
```

CONTRIBUTING.md
```diff
@@ -100,21 +100,52 @@ On the other hand... We _do_ want to compare several tools, to generate document
...
+def use_parse_file(x, y, z):
+    parse_file = have_optional_dependency("pybtex.database", "parse_file")
+    ...
+```
+
+This allows people to (1) use PyBaMM without importing optional dependencies by default and (2) configure module-dependent functionalities in their scripts, which _must_ be done before e.g. `print_citations` method is first imported.
+
+**Writing Tests for Optional Dependencies**
+
+Whenever a new optional dependency is added for optional functionality, it is recommended to write a corresponding unit test in _test_util.py_. This ensures that an error is raised upon the absence of said dependency. Here's an example:
```
Saransh-cpp, line 129:
```suggestion
Whenever a new optional dependency is added for optional functionality, it is recommended to write a corresponding unit test in `test_util.py`. This ensures that an error is raised upon the absence of said dependency. Here's an example:
```

pybamm/expression_tree/printing/sympy_overrides.py
```diff
@@ -3,22 +3,27 @@
 #
 import re
 
-from sympy.printing.latex import LatexPrinter
+from pybamm.util import have_optional_dependency
 
 
-class CustomPrint(LatexPrinter):
+def custom_latex_printer(expr, **settings):
+    latex = have_optional_dependency("sympy","latex")
```
Saransh-cpp, line 10:
```suggestion
    latex = have_optional_dependency("sympy", "latex")
```

pybamm/expression_tree/printing/sympy_overrides.py
```diff
@@ -3,22 +3,27 @@
...
+    Derivative = have_optional_dependency("sympy","Derivative")
```
Saransh-cpp, line 11:
```suggestion
    Derivative = have_optional_dependency("sympy", "Derivative")
```

pybamm/plotting/quick_plot.py
```diff
@@ -46,7 +47,7 @@ def split_long_string(title, max_words=None):
 
 def close_plots():
     """Close all open figures"""
-    import matplotlib.pyplot as plt
+    plt = have_optional_dependency("matplotlib", "pyplot")
```
Saransh-cpp, line 50:
```suggestion
    plt = have_optional_dependency("matplotlib.pyplot")
```

Saransh-cpp (changes requested):

pybamm/expression_tree/printing/sympy_overrides.py
```diff
@@ -3,22 +3,27 @@
...
         """Override :meth:`sympy.printing.latex.LatexPrinter._print_Derivative`"""
-        eqn = super()._print_Derivative(expr)
-
-        if getattr(expr, "force_partial", False) and "partial" not in eqn:
-            var1, var2 = re.findall(r"^\\frac{(\w+)}{(\w+) .+", eqn)[0]
-            eqn = eqn.replace(var1, "\partial").replace(var2, "\partial")
-
-        return eqn
-
+        return custom_latex_printer(expr)
```
Saransh-cpp, line 25: I am not sure if this will work or if this is already working (cc: @priyanshuone6 @brosaplanella).

If this is in working condition  - we should get rid of the `CustomPrint` and `custom_print_func`. Both of them are redundant because everything should work if we pass `custom_latex_printer` directly here (the only place `custom_print_func` is used) -
https://github.com/pybamm-team/PyBaMM/blob/b0db430a65cb0c6a964f139c92a6e1346d8dc36a/pybamm/expression_tree/operations/latexify.py#L227-L233

Saransh-cpp (changes requested):
Took over and made some changes. Probably the last change request (phew)

We should also be testing the "lite" pybamm in CI (as pointed out by @agriyakhetarpal), but I think we can do that after this release. The testing part will also tie up nicely with unittest to pytest migration (`pytest.importorskip` is quite intelligent and can be added module-wide).

pybamm/expression_tree/printing/sympy_overrides.py
brosaplanella: I am not very familiar with this part of the code, @priyanshuone6 will know better. It seems to be related to printing partial derivatives, so I would say that as long as partial derivatives get printed fine  (you can check with SPM or DFN), then I am happy with whatever implementation makes our lives easier.

Saransh-cpp: The current fix does not work :(

image

I think this should be a partial derivative in SPM?

brosaplanella: Yes, this should be partial

Saransh-cpp: Fixed it (reverted back to the original class)!

Saransh-cpp (approved):
Thanks a ton for the amazing work, @arjxn-py! 🙂 

I'll wait for others to review this before merging.

agriyakhetarpal (approved):
Looks good now, thank you @arjxn-py and @Saransh-cpp! Just needs a CHANGELOG entry and a documentation update to mark `anytree` as a required dependency in the installation guide.

I tested all example scripts with `pybamm[plot]` and everything worked as expected. I would also recommend re-running the failed integration tests and notebooks tests just to be on the safer side.

Saransh-cpp:
Not related to this PR, se #3519

abillscmu:
this seems to have resolved it, thanks

brosaplanella (approved):
Looks good to me, thank you very much to all who contributed fixing this issue!

Issue #3477, opened 2023-10-26
Switch to `ruff format`


Saransh-cpp:
Ruff can now format files!

Would mostly require changes in `pyproject.toml` and `.pre-commit-config.yaml`

Rjchauhan18:
Hello @Saransh-cpp 
Can i make PR for this ?

As mention in this issue https://github.com/pybamm-team/PyBaMM/issues/3476 
- Refering from this [repositery](https://github.com/astral-sh/ruff-pre-commit/blob/8a43105c597ef7067b5507451f4f1b07e5f6e26d/README.md?plain=1#L68) if we add  like this
```
repos:
  - repo: https://github.com/astral-sh/ruff-pre-commit
    rev: "v0.1.1"
    hooks:
      - id: ruff
        types_or: [python, pyi, jupyter]
        args: [--fix, --ignore=E741, --exclude=__init__.py]

      - id: ruff-format
        args: [--check]

```

 **Do you want it with `--check` or without?**

Rjchauhan18:
![image](https://github.com/pybamm-team/PyBaMM/assets/113838908/910b80be-f08a-40c6-81a9-68556ea00dd5)
---

Hello @Saransh-cpp 
Can you tell me that is this right or test is failing ?

---
`68 files would be reformatted, 472 files left unchanged` this is coming but nothing is change as i use `--check` in pre-commit-config.toml file. but then also 

>nox > Command pre-commit run --all-files failed with exit code 1
>nox > Session pre-commit failed.

---

![image](https://github.com/pybamm-team/PyBaMM/assets/113838908/139fe13a-4e60-4b47-91c7-836fd1c6c28f)

Saransh-cpp:
Scientific Python's guide now has instructions for ruff format - https://learn.scientific-python.org/development/guides/style/

We should follow this initially.

PR #3492, opened 2023-10-31
Fix switch to ruff format 

Rjchauhan18:
# Description

- Changes made in pre-commit-config.yaml file 
```
      - id: ruff-format
        args: [--check]
```

Fixes # (issue)

This PR will fixes the issue  https://github.com/pybamm-team/PyBaMM/issues/3477

## Type of change

## Further checks:


Rjchauhan18:
Hey @Saransh-cpp 
I am interested to doing this but starting 3rd nov, 2023 my exams are starting so i might not able to spend time solving issue till 9th nov, 2023 . 

so till then if someone solve this with you then fine otherwise i will come back on 9th still you can review it give me the comments what changes you want if i am quickly able to change those thing then i will otherwise i wll come on 9th nov .

Rjchauhan18:
And one more thing do you want it to auto fix everything or just check?

Saransh-cpp:
Sounds good! We should autofix everything.

Rjchauhan18:
Now pre-commit bot will auto commit everything which can be change

.pre-commit-config.yaml
```diff
@@ -10,6 +10,8 @@ repos:
         args: [--fix, --show-fixes]
         types_or: [python, pyi, jupyter]
 
+      - id: ruff-format
```
Saransh-cpp, line 13:
```suggestion
      - id: ruff-format
        types_or: [python, pyi, jupyter]
```

Saransh-cpp (changes requested):
Thanks, @Rjchauhan18! You should also add pre-commit's commit's hash in https://github.com/pybamm-team/PyBaMM/blob/develop/.git-blame-ignore-revs

.pre-commit-config.yaml
Rjchauhan18: If i do something like this 

```
 - repo: https://github.com/astral-sh/ruff-pre-commit
    rev: "v0.1.3"
    hooks:
      - id: ruff
        args: [--fix, --show-fixes]

      - id: ruff-format
        types_or: [python, pyi, jupyter]
```

then run the pre-commit locally then it is changing jupyter notebooks and removing `;`.
```
 - repo: https://github.com/astral-sh/ruff-pre-commit
    rev: "v0.1.3"
    hooks:
      - id: ruff
        args: [--fix, --show-fixes]
        types_or: [python, pyi, jupyter]

      - id: ruff-format
 ```
so better we keep this as it is not changing .ipynb files. I have tested it locally twice.

Comment (Rjchauhan18) [reply .pre-commit-config.yaml:13]:
```
  - repo: https://github.com/astral-sh/ruff-pre-commit
    rev: "v0.1.3"
    hooks:
      - id: ruff
        args: [--fix, --show-fixes]
   

      - id: ruff-format
        types_or: [python, pyi, jupyter]

```
It is removing `;` from .ipynb

Comment (kratman) [reply .pre-commit-config.yaml:13]: I think a lot of the semi-colons just stop output like:
```
<pybamm.solvers.solution.Solution at 0x13cc02f10>
```
This does not seem like a huge problem.

Another way to resolve that printing is to save output objects to a discarded variable:
```python
_ = sim.solve()
```
instead of
```python
sim.solve();
```

Rjchauhan18: yes, `_ = sim.solve()` this is in the jupyter notebook

But for plotting graphs without  line like this `[<matplotlib.lines.Line2D at 0x7f4123ea7ac0>]` we need semi-colons.

---
**One Solution is to put `# fmt: off` and `# fmt: on` where  semi-colons appear**
```
# fmt: off
plt.plot(sol["Time [s]"].data, sol["Voltage [V]"].data);
# fmt: on
``` 
Should I use these comments or try to find something else?

kratman: Personally I would prefer to avoid adding tool specific comments if there are other ways.

In this case:
```python
_ = plt.plot(sol["Time [s]"].data, sol["Voltage [V]"].data)
```
works since plot() also has a return

Rjchauhan18: Hello @Saransh-cpp 
what changes you prefer ?

Saransh-cpp: The semicolon thing was fixed in ruff upstream. You might have to reset your commits and introduce ruff format in an entirely new commit, given that we want all the pre-commit related changes bundled in a single commit.

Saransh-cpp (commented):
@Rjchauhan18 let us know if you need any help on this!

Saransh-cpp:
Closing this PR as it went stale, and taking up the work in https://github.com/pybamm-team/PyBaMM/pull/3656

Issue #3304, opened 2023-08-31
Pin versions


brosaplanella:
At the moment we do not pin to specific versions (in some cases we provide lower bounds) but this has the downside that installing old versions do not work (e.g. version 22.12 requires `casadi >= 3.5.0` which at the time meant 3.5, but now it would install 3.6 which is not compatible).

I don't know what's the right approach here, but just wanted to bring it up to discuss if we are happy with it as is, or if we want to be more specific with the pins (and if so, how).

agriyakhetarpal:
I read this blog recently about the nuances of versioning dependencies and learned a lot, it could offer some insights: https://iscinumpy.dev/post/bound-version-constraints/

Saransh-cpp:
TL;DR:

Yes, we should not add upper-bound version constraints unless they are extremely required (outdated PyBaMM versions should be allowed to break). Adding upper-bound constraints will make PyBaMM fall into anti-library patterns, converting it into an application and making it very hard to use it as a library (or as a dependency for other downstream libraries).

On the other hand, adding lower-bound pins is a good idea. We should definitely do that.

PS: The blog's author was my supervisor last year (learned all the python packaging and infrastructure stuff from him!) and we had a good amount of discussion on this. Upper-bound pins can really be a pain for downstream developers (some Scikit-HEP libraries were indeed facing a lot of difficulties from TensorFlow's upper-bound pinning policies).

brosaplanella:
Cool, then we are doing the right thing, right? Will close this.

PR #3501, opened 2023-11-07
Fix `pipx` + `nox` + `pip` dependency resolution bug in macOS workflows

agriyakhetarpal:
# Description

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/.github/workflows/run_periodic_tests.yml
+++ b/.github/workflows/run_periodic_tests.yml
@@ -66,19 +66,14 @@ jobs:
           sudo apt install gfortran gcc libopenblas-dev graphviz pandoc
           sudo apt install texlive-full
 
-      # Added fixes to homebrew installs:
-      # rm -f /usr/local/bin/2to3
-      # (see https://github.com/actions/virtual-environments/issues/2322)
+      # sometimes gfortran cannot be found, so reinstall gcc just to be sure
       - name: Install MacOS system dependencies
         if: matrix.os == 'macos-latest'
-        run: |
-          rm -f /usr/local/bin/2to3*
-          rm -f /usr/local/bin/idle3*
-          rm -f /usr/local/bin/pydoc3*
-          rm -f /usr/local/bin/python3*
+        run:
           brew update
           brew install graphviz
           brew install openblas
+          brew reinstall gcc
 
       - name: Install Windows system dependencies
         if: matrix.os == 'windows-latest'
--- a/.github/workflows/run_periodic_tests.yml
+++ b/.github/workflows/run_periodic_tests.yml
@@ -78,42 +78,42 @@ jobs:
 
       - name: Install standard Python dependencies
         run: |
-          python -m pip install --upgrade pip wheel setuptools
+          python -m pip install --upgrade pip wheel setuptools nox
 
       - name: Install SuiteSparse and SUNDIALS on GNU/Linux
         if: matrix.os == 'ubuntu-latest'
-        run: pipx run nox -s pybamm-requires
+        run: python -m nox -s pybamm-requires
 
       - name: Run unit tests for GNU/Linux with Python 3.8, 3.9, and 3.10, and for macOS and Windows with all Python versions
         if: (matrix.os == 'ubuntu-latest' && matrix.python-version != 3.11) || (matrix.os != 'ubuntu-latest')
-        run: pipx run nox -s unit
+        run: python -m nox -s unit
 
       - name: Run unit tests for GNU/Linux with Python 3.11 and generate coverage report
         if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.11
-        run: pipx run nox -s coverage
+        run: python -m nox -s coverage
 
       - name: Upload coverage report
         if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.11
         uses: codecov/codecov-action@v3.1.4
 
       - name: Run integration tests
-        run: pipx run nox -s integration
+        run: python -m nox -s integration
 
       - name: Install docs dependencies and run doctests
         if: matrix.os == 'ubuntu-latest'
-        run: pipx run nox -s doctests
+        run: python -m nox -s doctests
 
       - name: Check if the documentation can be built
         if: matrix.os == 'ubuntu-latest'
-        run: pipx run nox -s docs
+        run: python -m nox -s docs
 
       - name: Install dev dependencies and run example tests
         if: matrix.os == 'ubuntu-latest'
-        run: pipx run nox -s examples
+        run: python -m nox -s examples
 
       - name: Run example scripts tests
         if: matrix.os == 'ubuntu-latest'
-        run: pipx run nox -s scripts
+        run: python -m nox -s scripts
 
   #M-series Mac Mini
   build-apple-mseries:
--- a/.github/workflows/run_periodic_tests.yml
+++ b/.github/workflows/run_periodic_tests.yml
@@ -69,7 +69,6 @@ jobs:
       - name: Install macOS system dependencies
         if: matrix.os == 'macos-latest'
         run:
-          brew update
           brew install graphviz openblas
           brew reinstall gcc
 
--- a/.github/workflows/run_periodic_tests.yml
+++ b/.github/workflows/run_periodic_tests.yml
@@ -72,7 +72,7 @@ jobs:
         run:
           brew update
           brew install graphviz
-          brew install openblas
+          brew reinstall openblas
           brew reinstall gcc
 
       - name: Install Windows system dependencies
--- a/.github/workflows/run_periodic_tests.yml
+++ b/.github/workflows/run_periodic_tests.yml
@@ -66,13 +66,11 @@ jobs:
           sudo apt install gfortran gcc libopenblas-dev graphviz pandoc
           sudo apt install texlive-full
 
-      # sometimes gfortran cannot be found, so reinstall gcc just to be sure
-      - name: Install MacOS system dependencies
+      - name: Install macOS system dependencies
         if: matrix.os == 'macos-latest'
         run:
           brew update
-          brew install graphviz
-          brew reinstall openblas
+          brew install graphviz openblas
           brew reinstall gcc
 
       - name: Install Windows system dependencies
--- a/.github/workflows/test_on_push.yml
+++ b/.github/workflows/test_on_push.yml
@@ -64,7 +64,6 @@ jobs:
           sudo dot -c
           sudo apt-get install libopenblas-dev texlive-latex-extra dvipng
 
-      # sometimes gfortran cannot be found, so reinstall gcc just to be sure
       - name: Install macOS system dependencies
         if: matrix.os == 'macos-latest'
         env:
@@ -78,7 +77,6 @@ jobs:
           brew analytics off
           brew update
           brew install graphviz openblas
-          brew reinstall gcc
 
       - name: Install Windows system dependencies
         if: matrix.os == 'windows-latest'
@@ -92,7 +90,7 @@ jobs:
           cache: 'pip'
           cache-dependency-path: setup.py
 
-      - name: Install Python dependencies
+      - name: Install standard Python dependencies
         run: |
           pip install --upgrade pip wheel setuptools
 
@@ -151,7 +149,7 @@ jobs:
           cache: 'pip'
           cache-dependency-path: setup.py
 
-      - name: Install Python dependencies
+      - name: Install standard Python dependencies
         run: |
           pip install --upgrade pip wheel setuptools
 
@@ -207,7 +205,6 @@ jobs:
           sudo dot -c
           sudo apt-get install libopenblas-dev texlive-latex-extra dvipng
 
-      # sometimes gfortran cannot be found, so reinstall gcc just to be sure
       - name: Install macOS system dependencies
         if: matrix.os == 'macos-latest'
         env:
@@ -220,9 +217,7 @@ jobs:
         run: |
           brew analytics off
           brew update
-          brew install graphviz
-          brew reinstall gcc
-          brew reinstall openblas
+          brew install graphviz openblas
 
       - name: Install Windows system dependencies
         if: matrix.os == 'windows-latest'
@@ -236,7 +231,7 @@ jobs:
           cache: 'pip'
           cache-dependency-path: setup.py
 
-      - name: Install Python dependencies
+      - name: Install standard Python dependencies
         run: |
           pip install --upgrade pip wheel setuptools
 
@@ -296,7 +291,7 @@ jobs:
           cache: 'pip'
           cache-dependency-path: setup.py
 
-      - name: Install Python dependencies
+      - name: Install standard Python dependencies
         run: |
           pip install --upgrade pip wheel setuptools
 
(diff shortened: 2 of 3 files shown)


agriyakhetarpal:
Looks like there is a different error now: OpenBLAS is not found

agriyakhetarpal:
I did some more digging, and I realise that the error is most likely because the macOS runners are trying to install `scipy` off the source distribution rather than the wheels, but I don't get why this is the case. I don't think there are any version conflicts we need to take care of, because [whenever the run passes](https://github.com/pybamm-team/PyBaMM/actions/runs/6787993298/job/18452367349?pr=3501) a wheel is being downloaded, and the source distribution installation of `scipy` always [causes a failure](https://github.com/pybamm-team/PyBaMM/actions/runs/6787993298/job/18452366848?pr=3501), as expected.

brosaplanella:
> Edit: I am not sure if it is a pip problem, a GitHub Actions one, or a scipy one so as to file it in an appropriate issue tracker, but the act of simply re-triggering the failed runs has made them work in previous instances, so it should be fine to close this or add PIP_ONLY_BINARY="scipy" to the environment variables. Please let me know which of the two would be better

@Saransh-cpp what do you think?

brosaplanella:
Found this: https://github.com/DeepLabCut/DeepLabCut/issues/2138#issuecomment-1414031412

Could the issue be related to the caching (i.e. we cached a version that fails so we run into the failing tests over and over)?

agriyakhetarpal:
I think that is possible, which expands it to potentially be a `setup-python` cache issue too. It should have a configuration option to cache just wheels and not tarballs, but at least that is fixable by removing the cache entries in the Actions tab.

The issue has been irregular which is suggestive that it has to do with `pip`'s dependency resolver as well (it might not be deterministic?). I would suggest bumping the minimum required versions to `scipy>=1.10.1`and `numpy>=1.19.5`, i.e., the versions that last supported Python 3.8 so as to not put excess strain on the resolution algorithm. We could also look into setting up a constraints file to be tested like `pyhf` does in their scheduled workflows (https://github.com/scikit-hep/pyhf/actions/workflows/lower-bound-requirements.yml).

agriyakhetarpal:
I think that is possible, which expands it to potentially be a `setup-python` cache issue too. It should have a configuration option to cache just wheels and not tarballs, but at least that is fixable by removing the cache entries in the Actions tab.

The issue has been irregular which is suggestive that it has to do with `pip`'s dependency resolver as well (it might not be deterministic?). I would suggest bumping the minimum required versions to `scipy>=1.10.1`and `numpy>=1.19.5`, i.e., the versions that last supported Python 3.8 so as to not put excess strain on the resolution algorithm. We could also look into setting up a constraints file to be tested like `pyhf` does in their scheduled workflows (https://github.com/scikit-hep/pyhf/actions/workflows/lower-bound-requirements.yml) as a continuation of #3304.

Saransh-cpp (commented):
I think this is a caching issue, definitely not a gfortran/SciPy issue. This could be a pip issue but I went through PyBaMM's dependencies and we don't have any conflicts. This is probably not a SciPy issue, but instead a SymPy issue. The CI is unable to find a SymPy version that satisfies our constraints, but that should not be a problem because there is only 1 SymPy constraint in our dependency tree -
```
pybamm                                Python Battery Mathematical Modelling.
├── anytree>=2.4.3                    Powerful and Lightweight Python Tree Data Structure with various plugins
│   └── six                           Python 2 and 3 compatibility utilities
├── autograd>=1.2                     Efficiently computes derivatives of numpy code.
│   ├── future>=0.15.2                Clean single-source support for Python 3 and 2
│   └── numpy>=1.12                   Fundamental package for array computing in Python
├── bpx                               BPX schema and parsers
│   ├── pydantic<2                    Data validation and settings management using python type hints
│   │   └── typing-extensions>=4.2.0  Backported and Experimental Type Hints for Python 3.8+
│   └── pyparsing                     pyparsing module - Classes and methods to define and execute parsing grammars
├── casadi>=3.6.0                     CasADi -- framework for algorithmic differentiation and numeric optimization
│   └── numpy                         Fundamental package for array computing in Python
├── imageio>=2.9.0                    Library for reading and writing a wide range of image, video, scientific, and volumetric data formats.
│   ├── numpy                         Fundamental package for array computing in Python
│   └── pillow<10.1.0,>=8.3.2         Python Imaging Library (Fork)
├── importlib-metadata                Read metadata from Python packages
│   └── zipp>=0.5                     Backport of pathlib-compatible object wrapper for zip files
├── matplotlib>=2.0                   Python plotting package
│   ├── contourpy>=1.0.1              Python library for calculating contours of 2D quadrilateral grids
│   │   └── numpy<2.0,>=1.20          Fundamental package for array computing in Python
│   ├── cycler>=0.10                  Composable style cycles
│   ├── fonttools>=4.22.0             Tools to manipulate font files
│   ├── kiwisolver>=1.3.1             A fast implementation of the Cassowary constraint solver
│   ├── numpy<2,>=1.21                Fundamental package for array computing in Python
│   ├── packaging>=20.0               Core utilities for Python packages
│   ├── pillow>=8                     Python Imaging Library (Fork)
│   ├── pyparsing>=2.3.1              pyparsing module - Classes and methods to define and execute parsing grammars
│   └── python-dateutil>=2.7          Extensions to the standard Python datetime module
│       └── six>=1.5                  Python 2 and 3 compatibility utilities
├── numpy>=1.16                       Fundamental package for array computing in Python
├── pandas>=0.24                      Powerful data structures for data analysis, time series, and statistics
│   ├── numpy<2,>=1.23.2              Fundamental package for array computing in Python
│   ├── python-dateutil>=2.8.2        Extensions to the standard Python datetime module
│   │   └── six>=1.5                  Python 2 and 3 compatibility utilities
│   ├── pytz>=2020.1                  World timezone definitions, modern and historical
│   └── tzdata>=2022.1                Provider of IANA time zone data
├── pybtex>=0.24.0                    A BibTeX-compatible bibliography processor in Python
│   ├── PyYAML>=3.01                  YAML parser and emitter for Python
│   ├── latexcodec>=1.0.4             A lexer and codec to work with LaTeX code in Python.
│   │   └── six>=1.4.1                Python 2 and 3 compatibility utilities
│   └── six                           Python 2 and 3 compatibility utilities
├── scikit-fem>=0.2.0                 Simple finite element assemblers
│   ├── numpy                         Fundamental package for array computing in Python
│   └── scipy                         Fundamental algorithms for scientific computing in Python
│       └── numpy<1.28.0,>=1.21.6     Fundamental package for array computing in Python
├── scipy>=1.3                        Fundamental algorithms for scientific computing in Python
│   └── numpy<1.28.0,>=1.21.6         Fundamental package for array computing in Python
├── sympy>=1.8                        Computer algebra system (CAS) in Python
│   └── mpmath>=0.19                  Python library for arbitrary-precision floating-point arithmetic
├── tqdm                              Fast, Extensible Progress Meter
└── xarray                            N-D labeled arrays and datasets in Python
    ├── numpy>=1.22                   Fundamental package for array computing in Python
    ├── packaging>=21.3               Core utilities for Python packages
    └── pandas>=1.4                   Powerful data structures for data analysis, time series, and statistics
        ├── numpy<2,>=1.23.2          Fundamental package for array computing in Python
        ├── python-dateutil>=2.8.2    Extensions to the standard Python datetime module
        │   └── six>=1.5              Python 2 and 3 compatibility utilities
        ├── pytz>=2020.1              World timezone definitions, modern and historical
        └── tzdata>=2022.1            Provider of IANA time zone data
```

I would also suggest installing PyBaMM without Nox in the CI once to check if this is a Nox issue.

.github/workflows/test_on_push.yml
```diff
@@ -219,7 +222,9 @@ jobs:
         run: |
           brew analytics off
           brew update
-          brew install graphviz openblas
+          brew install graphviz
+          brew reinstall gcc
+          brew reinstall openblas
```
brosaplanella, line 227: Are these changes still needed?

.github/workflows/run_periodic_tests.yml
```diff
@@ -66,19 +66,14 @@ jobs:
...
-          rm -f /usr/local/bin/2to3*
-          rm -f /usr/local/bin/idle3*
-          rm -f /usr/local/bin/pydoc3*
-          rm -f /usr/local/bin/python3*
+        run:
           brew update
           brew install graphviz
-          brew install openblas
+          brew reinstall openblas
+          brew reinstall gcc
```
brosaplanella, line 76: Are these changes still needed?

agriyakhetarpal:
Thanks for the reviews. I temporarily removed the version bounds and I am now trying to exercise some new version constraints to see if they work and fix the `sympy` issue (they worked locally when installing `[all, dev, docs]` together). I will clean up the changes after that.

agriyakhetarpal:
Surprisingly (or rather not), the dependency resolution does pass sometimes: https://github.com/pybamm-team/PyBaMM/actions/runs/6812446181/job/18525103672?pr=3501 when the others fail

agriyakhetarpal:
So far, caching is not the problem here and can be ruled out. [These logs](https://github.com/pybamm-team/PyBaMM/actions/runs/6812840135/job/18526496348?pr=3501#step:11:270) have cached wheels for `scipy` and `sympy`, and therefore all integration tests proceed as normal. The issue comes only when `pip` chooses to install from a source distribution of `scipy`, which is always version 1.11.1 as far as I have noticed.

.github/workflows/run_periodic_tests.yml
agriyakhetarpal: Not needed anymore, removing

brosaplanella:
What's the status of this PR? Should we reviewed or does more stuff need to be done?

agriyakhetarpal:
> What's the status of this PR? Should we reviewed or does more stuff need to be done?

So far, I have not been able to reproduce the problem on macOS arm64 or Windows amd64 locally. I will try bumping up a few of the lower bounds for some more of our dependencies to the minimum version where they started supporting Python 3.8 to see if that solves the issue or try to single out the error in other ways

@Saransh-cpp reports that there are no dependency conflicts and I concur, so even if even if bumping the lower bounds doesn't help, it should be a good thing to do and we can merge those changes.

setup.py
```diff
@@ -263,25 +263,17 @@ def compile_KLU():
...
-            "imageio>=2.9.0",
-            "pybtex>=0.24.0",
-            "sympy>=1.8",
-            "bpx",
-            "tqdm",
-            "matplotlib>=2.0",
-            "jupyter",
+            "autograd>=1.6.2",
+            "scikit-fem>=8.1.0",
+            "pybamm[examples,plot,cite,latexify,bpx,tqdm,pandas]"
```
agriyakhetarpal, line 276: This syntax is a pretty cool feature which I got from [this blog from Hynek Schlawack](https://hynek.me/articles/python-recursive-optional-dependencies/ ). It requires `pip` v21+ though, which should be fine for us because `pip` v23.0.1 has been backported to ship with Python 3.8.

agriyakhetarpal:
# Possible areas where this issue is coming from

## Ruled out

| Likely cause | Reason it was ruled out |
|--------|--------|
| `gfortran` missing | Not required when all dependencies have wheels |
| Homebrew issue | Fixed, and Homebrew is not related to and does not manage Python dependencies |
| Caching issue with `setup-python` | Removed in https://github.com/pybamm-team/PyBaMM/pull/3501/commits/f87b6bf5c8306581cdc4ed1e2f4fda3d61f077e4, but subsequent workflow runs failed as well  |
| `nox` | Runs `pip` in a subprocess with the current Python executable, unlikely to be the cause of the error |
| Version conflicts | Checked, none found. No problems when installing locally | 

## Investigating

| Potential reason | (Incomplete) explanation |
|--------|--------|
| Dependency resolution bug in `pip` | Might not be deterministic; sometimes the tests proceed and pass |
| Discrepancies in GitHub Actions runners' configurations | They keep getting updated all the time and not all of them are updated |
| `scipy` | Version `1.11.1` was yanked on PyPI due to a licensing issue, maybe that is why `pip` keeps installing it with its tarball? | 
| `sympy` | Version constraints mismatch? | 
| Bug in `pipx` and/or its use with `nox` | The "Install PyBaMM dependencies" step that I had removed in https://github.com/pybamm-team/PyBaMM/pull/3501/commits/dc1f6eddd4e8e838628ef496ce3d4d9a2a30ac79 does not issue any errors and installs everything normally, so `pipx` could be the culprit | 
| `sympy` | Version constraint mismatch | 

<hr>

Note to self: try `python -m nox` instead of `pipx run nox`

agriyakhetarpal:
It turns out this is likely an issue when working with `pipx` and `nox` together. The `pip` dependency resolver has been deterministic ever since its origin and post its revamp in 2020 as far as I can tell; changing to `python -m nox` instead of `pipx run nox` helped fix it. The failures on the Linux tests (scripts, unit, integration, coverage) here are unrelated since I turned the `pybamm-requires` caches back on – they can be fixed by either deleting some of the recent cache entries in the Actions tab or safely left as-is because these changes will not affect other PRs. The previous run, where they had been removed, had passed: https://github.com/pybamm-team/PyBaMM/actions/runs/6889728515.

The notebook tests and pre-commit.ci tests will be fixed by #3519 and the failing doctests are unrelated coming from an issue with SciPy's intersphinx inventory. All macOS unit and integration tests proceeded to run here and passed after the installation of dependencies.

I have also bumped up the versions of some of the dependencies keeping compatibility to a minimum of Python 3.8 and introduced recursive optional dependencies in `setup.py`. This PR should now be ready to review and merge.

setup.py
agriyakhetarpal: Also, I removed a duplicate entry of `anytree` from the `all` set of extras dependencies since we now have it in the required dependencies after #3475.

Saransh-cpp:
I guess the CI is failing because of the CMake cache?

agriyakhetarpal:
Yes, as mentioned in https://github.com/pybamm-team/PyBaMM/pull/3501#issuecomment-1814332013

setup.py
```diff
@@ -203,11 +203,11 @@ def compile_KLU():
...
-        "numpy>=1.16",
-        "scipy>=1.3",
-        "casadi>=3.6.0",
-        "xarray",
-        "anytree>=2.4.3",
+        "numpy>=1.24.4",
+        "scipy>=1.10.1",
+        "casadi>=3.6.3",
+        "xarray>=23.1.0",
+        "anytree>=2.12.0",
```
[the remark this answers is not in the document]
agriyakhetarpal: This is a very interesting specification. We could add a workflow to open an issue once every quarter about this (and bump the versions manually)

agriyakhetarpal:
I have now bumped the lower bounds keeping to a minimum of those with more than six months left in the support cycle according to the SPEC and ignored those versions which have less than three months left

setup.py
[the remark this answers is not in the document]
Saransh-cpp: That sounds good!

Saransh-cpp (approved):
Looks good now, thanks!

setup.py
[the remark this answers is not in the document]
agriyakhetarpal: Great, making it a separate issue so that we can track that

brosaplanella (approved):
Looks great, thanks!

arjxn-py (approved):
Thanks @agriyakhetarpal, looks nice.

PR #3516, opened 2023-11-10
Fix bug with identical steps with different end times

brosaplanella:
# Description

Fixes a bug in which identical steps but with different end times could not be distinguished, and thus crashed the simulation by using a wrong end time.

It is a hacky fix, a more robust one should be implemented as part of the refactor in #3176.

## Type of change

# Key checklist:

## Further checks:


tinosulzer (approved):
looks good to me

tinosulzer:
All code that goes into develop now is held off until the next release, except critical bug fixes, at least that is how it should be ...

Issue #3390, opened 2023-10-01
Drop support for Python `3.8`, support Python `3.12`?


agriyakhetarpal:
NumPy>=1.26 and SciPy>=1.11 support Python 3.9+ now and already have wheels for Python 3.12. The blocker should be CasADi, which might have a new release soon because the [pre-release](https://github.com/casadi/casadi/releases/tag/nightly-py312) is available. In any case, we should be able to drop Python `3.8` support at this time and `3.12` can wait

Saransh-cpp:
I don't think we should drop support for Python 3.8 unless PyBaMM breaks on 3.8 or 3.8 reaches `end-of-life`.

We can actually start testing PyBaMM on 3.12 in the CI. IIRC, `pybind11` has already migrated to Python 3.12, and we can fetch CasADi's nightly wheels in the 3.12 job.

agriyakhetarpal:
Yes, it would be good to try supporting Python 3.12, which comes tomorrow. `xarray` doesn't run CI against 3.12 yet, but they list their Python dependency as >=3.9. I think we should be fine with using the nightly wheels for CasADi, should we let the optional dependencies catch up later?

> I don't think we should drop support for Python 3.8 unless PyBaMM breaks on 3.8 or 3.8 reaches end-of-life.

That makes sense, I just checked and Python 3.8 EOL is in October 2024, so we have just about a year to remove it

Saransh-cpp:
We should only build and ship pybamm wheels once every dependency works, but testing pybamm (just with the required dependencies) in the CI should be good.

(release candidates of 3.12 have been available on GH Actions for quite some time now)

Saransh-cpp:
Copying from slack -

> https://scientific-python.org/specs/ - like PEPs but for the Scientific Python ecosystem. I think we should follow them to stay up-to-date with other libraries.
>
> The first SPEC suggests removing Python 3.9 support by the end of this year, but I’m not sure if we should do that :grimacing:

agriyakhetarpal:
There are a lot of interesting things available in the Scientific Python ecosystem, perhaps we should also try using [`sp-repo-review`](https://learn.scientific-python.org/development/guides/repo-review/)? I guess this would be better off as a separate issue, please let me know if I should open one

Saransh-cpp:
[...]

PR #3531, opened 2023-11-16
Support Python 3.12

agriyakhetarpal:
# Description

Related to #3390. Adds support for Python 3.12 across the repository

## Type of change

# Key checklist:

## Further checks:


noxfile.py
```diff
@@ -60,10 +92,10 @@ def run_coverage(session):
     """Run the coverage tests and generate an XML report."""
     set_environment_variables(PYBAMM_ENV, session=session)
     session.install("coverage", silent=False)
-    session.install("-e", ".[all]", silent=False)
     if sys.platform != "win32":
```
Saransh-cpp, line 95: We can add a check here for Python 3.12

noxfile.py
```diff
@@ -73,35 +105,36 @@ def run_coverage(session):
 def run_integration(session):
     """Run the integration tests."""
     set_environment_variables(PYBAMM_ENV, session=session)
-    session.install("-e", ".[all]", silent=False)
     if sys.platform == "linux":
```
Saransh-cpp, line 108: Ditto

noxfile.py
```diff
@@ -73,35 +105,36 @@ def run_coverage(session):
...
+    else:
+        install_PyBaMM(session=session, extras=["all"])
     session.run("python", "run-tests.py", "--integration")
 
 
 @nox.session(name="doctests")
 def run_doctests(session):
     """Run the doctests and generate the output(s) in the docs/build/ directory."""
-    session.install("-e", ".[all,docs]", silent=False)
+    install_PyBaMM(session=session, extras=["all","docs"])
```
Saransh-cpp, line 118: `session.install` should work fine here

noxfile.py
```diff
@@ -73,35 +105,36 @@ def run_coverage(session):
...
     session.run("python", "run-tests.py", "--doctest")
 
 
 @nox.session(name="unit")
 def run_unit(session):
     """Run the unit tests."""
     set_environment_variables(PYBAMM_ENV, session=session)
-    session.install("-e", ".[all]", silent=False)
     if sys.platform == "linux":
```
Saransh-cpp, line 126: Ditto

noxfile.py
```diff
@@ -73,35 +105,36 @@ def run_coverage(session):
...
+        install_PyBaMM(session=session, extras=["all"])
     session.run("python", "run-tests.py", "--unit")
 
 
 @nox.session(name="examples")
 def run_examples(session):
     """Run the examples tests for Jupyter notebooks."""
     set_environment_variables(PYBAMM_ENV, session=session)
-    session.install("-e", ".[all,dev]", silent=False)
+    install_PyBaMM(session=session, extras=["all","dev"])
```
Saransh-cpp, line 137: `session.install` should work fine here

noxfile.py
```diff
@@ -110,7 +143,7 @@ def run_examples(session):
 def run_scripts(session):
     """Run the scripts tests for Python scripts."""
     set_environment_variables(PYBAMM_ENV, session=session)
-    session.install("-e", ".[all]", silent=False)
+    install_PyBaMM(session=session, extras=["all"])
```
Saransh-cpp, line 146: `session.install` should work fine here

noxfile.py
```diff
@@ -137,18 +170,18 @@ def set_dev(session):
 def run_tests(session):
     """Run the unit tests and integration tests sequentially."""
     set_environment_variables(PYBAMM_ENV, session=session)
-    session.install("-e", ".[all]", silent=False)
     if sys.platform == "linux" or sys.platform == "darwin":
```
Saransh-cpp, line 173: Ditto

noxfile.py
```diff
@@ -36,6 +36,38 @@ def set_environment_variables(env_dict, session):
         session.env[key] = value
 
 
+def install_PyBaMM(session, extras):
```
Saransh-cpp, line 39: This looks redundant at the first look (and is maybe overcomplicating things). We can simply add an `and` condition in `if-else` blocks to check for Python 3.12

Saransh-cpp (changes requested):
Thanks @agriyakhetarpal! I think it would be better to remove the `install_PyBaMM` function because I think the function is redundant and is complicating simple things.

noxfile.py
agriyakhetarpal: The issue with not having this function is that there will be too many `if-else` blocks to handle. They would work for tests on PRs, but we are also using `nox` for local development besides CI

kratman: If the issue is just for JAX and ODEs, can we just make them a separate install? For instance, pull them out of dev dependencies and just make an additional extras group for modules with potential install difficulties?

agriyakhetarpal: `jax` and `odes` are separate installs at this moment with their own groups; it wouldn't make sense to add another extra set with both of those and then not document that for users. I don't think there is currently a provision for a "semi-private" set of extras of the type `[_odes]`. The problem here is that we can't tailor the noxfile just for our CI needs because it is also being used locally by developers: a user on, say, macOS should be able to install `jax` through any `nox` session if they want instead of restricting that functionality to a particular platform

I can make the change and add if-else blocks for every session but it won't conform to DRY practices

agriyakhetarpal:
I have now removed the `install_PyBaMM()` function and used some extra conditions instead. Doing this will bring many conflicts with some of my other PRs, but I can rebase any of them depending on whichever ones will get merged first.

Saransh-cpp:
@agriyakhetarpal could you rebase this?

agriyakhetarpal:
> @agriyakhetarpal could you rebase this?

I just noticed this comment—I think there will be many merge conflicts here and there and they will be repeated a lot as the latest changes are added upstream—perhaps I can rebase this branch here on top of the changes in #3550, or just wait for #3550 to be reviewed and get merged first. I haven't worked on the installation guide yet to adjust for both of the PRs since they will require changes to it, but I shall do that as soon as possible next week.

noxfile.py
```diff
@@ -73,8 +75,10 @@ def run_coverage(session):
 def run_integration(session):
     """Run the integration tests."""
     set_environment_variables(PYBAMM_ENV, session=session)
+    # TODO: remove this when PyBaMM moves to pyproject.toml
+    session.install("--upgrade", "pip", "setuptools", "wheel", silent=False)
     session.install("-e", ".[all]", silent=False)
-    if sys.platform == "linux":
+    if sys.platform == "linux" and sys.version_info < (3, 12):
```
arjxn-py, line 81: We would update this too once done with #3550, right?

arjxn-py (approved):
Thanks @agriyakhetarpal, looks good.

noxfile.py
agriyakhetarpal: Yes, will update this branch soon

Saransh-cpp (changes requested):
Thanks, @agriyakhetarpal! This should also remove all the instances of `distutils` from the codebase. We are importing `distutils` in `setup.py` at the moment.

noxfile.py
```diff
@@ -61,7 +61,10 @@ def run_coverage(session):
     set_environment_variables(PYBAMM_ENV, session=session)
     session.install("coverage", silent=False)
     if sys.platform != "win32":
-        session.install("-e", ".[all,jax,odes]", silent=False)
+        if sys.version_info > (3, 12):
```
Saransh-cpp, line 64: Should this be `>=` or will `>` capture `3.12.0`?

agriyakhetarpal: `>` is fine here – I had tested this locally on Python 3.12.0, we should be safe

Saransh-cpp (approved):
Thanks, @agriyakhetarpal! Looks good!

Issue #3480, opened 2023-10-27
Simplify installation by indicating `pybind11` as a build-time dependency


agriyakhetarpal:
We currently git-clone the `pybind11` repository inside the `pybamm-requires` `nox` session when installing from source. Once #3301 is merged, this step can be removed by either including it as a submodule in the repository, or via adding it to the `[build-system]` section in `pyproject.toml` plus suitable modifications to the CMake build process. The latter option would be better

agriyakhetarpal:
Closing, this is now being tracked in one place as a part of the migration to `scikit-build-core by @cringeyburger in #3564

PR #3557, opened 2023-11-25
Move ruff, coverage, and pytest configs to pyproject.toml

### LINKED ISSUE #3478: Move all the extra configurations to `pyproject.toml`
Once #3301 is merged, we can move the configurations of pytest, ruff, and others to `pyproject.toml`

Saransh-cpp:
# Description

I've left cibuildwheel as of now as keeping its config in the CI files seems more accessible.

Fixes #3478 

## Type of change

# Key checklist:

## Further checks:


kratman (approved):

kratman:
Missed this when I reviewed, but I think the errors are because "--rcfile=.coveragerc" needs to be removed from the noxfile for coverage

agriyakhetarpal:
Also, the tests for the example notebooks are failing too. I think the `pytest` meta path finder is looking for all installable packages in the root directory, and the git-cloned `pybind11` is also regarded as one. We should be able to exclude that path for the time being until `pybind11` is added as a build-time dependency (#3480).

pyproject.toml
```diff
@@ -174,6 +174,8 @@ pybamm = [
 
 [tool.setuptools.packages.find]
 include = ["pybamm", "pybamm.*"]
+# TODO: remove once https://github.com/pybamm-team/PyBaMM/issues/3480 is resolved
```
kratman, line 177: Unrelated, but do we have and process for periodically checking TODOs to make sure we are taking care of them over time?

pyproject.toml
```diff
@@ -172,3 +174,78 @@ pybamm = [
...
+
+# Logging configuration
+log_cli = "true"
+log_cli_level = "INFO"
+log_format = "%(asctime)s - %(name)s - %(levelname)s - %(message)s"
+log_date_format = "%Y-%m-%d %H:%M:%S"
+
+[tool.coverage.run]
+source = "pybamm"
+concurrency = "multiprocessing"
```
kratman, line 251: This might be the last of what is missing
```suggestion
concurrency = ["multiprocessing"]
```

pyproject.toml
agriyakhetarpal: It should be a good idea to add a workflow to do that. I don't remember the name exactly, but there is a pre-commit hook too that can check for `# TODO:` and `# FIXME:` comments.

agriyakhetarpal: Also, this `# TODO:` comment in specific should not be added imo, I remember that when I worked on this part of the configuration it was also trying to install the `tests`, `examples`, and the `benchmarks` folders treating them as packages, because they are not namespace packages.

pyproject.toml
```diff
@@ -172,3 +174,78 @@ pybamm = [
...
+]
+addopts = [
+  "-nauto",
+  "-v",
+]
+testpaths = [
+  "docs/source/examples/",
+]
+console_output_style = "progress"
+
```
agriyakhetarpal, line 242: I think the `norecursedirs` configuration value should do it: https://docs.pytest.org/en/7.1.x/reference/reference.html#confval-norecursedirs

Saransh-cpp: This surprisingly did not work (it should have according to the docs). Ignoring pybind11 in `addopts` works for me locally.

pyproject.toml
```diff
@@ -172,3 +174,77 @@ pybamm = [
...
+  # "PIE",         # flake8-pie
+  # "PL",          # pylint
+  # "PT",          # flake8-pytest-style
+  # "PTH",         # flake8-use-pathlib
+  # "RET",         # flake8-return
+  "RUF",         # Ruff-specific
+  # "SIM",         # flake8-simplify
+  # "T20",         # flake8-print
+  # "UP",          # pyupgrade
+  "YTT",         # flake8-2020
```
Saransh-cpp, line 200: I do want to add all these rules but this PR will blow up with formatting changes if I enable all of them at once.

kratman: We can always just review the config stuff then enable the other checks right before a final review. I do that sometimes on other projects.

pyproject.toml
agriyakhetarpal: I think it failed again on the latest commit with `addopts`, up to you if you want to try another option or wait for #3480 to be completed – I am almost done!

pyproject.toml
agriyakhetarpal: Should we leave them for a follow-up PR?

kratman: I was figuring that we should do it all at once since we suppress these changes in git blame.

pyproject.toml
Saransh-cpp: Turns out, this wasn't a `pytest`+`pybind11` issue. I was specifying `pytest`'s configuration under `[tool.pytest]`, but it should have been `[tool.pytest.ini_options]`.

pyproject.toml
Saransh-cpp: I think follow-up PRs would be best here. That would also ensure that we have everyone's consensus on adding thee individual rule. For instance, we used to have isort in pybamm somewhere in the past but then it was removed on everyone's consensus. I'll open an issue for this once the PR is merged.

agriyakhetarpal (approved):
Looks great to me, thanks, @Saransh-cpp!

brosaplanella (approved):
Looks good, thanks!

PR #3560, opened 2023-11-25
Specify `pybind11` as a build-time dependency

### LINKED ISSUE #3480: Simplify installation by indicating `pybind11` as a build-time dependency

agriyakhetarpal:
# Description

This PR adds `pybind11` to the list of build-time dependencies in `pyproject.toml` (#3301); which means it is picked up, downloaded, and installed by `pip` automatically and therefore made available to users. CMake can now find the package instead of using the git-cloned directory. Contributors and developers will no longer have to git-clone the pybind11 repository from GitHub. All instances of `pybind11` have been moved to a unified location for other workflows to use.

Closes #3480 

## Type of change

# Key checklist:

## Further checks:


agriyakhetarpal:
I will run the wheel-building workflow on my fork just to make sure these changes are okay.

agriyakhetarpal:
Passes locally for me, strangely!

setup.py
```diff
@@ -241,39 +241,52 @@ def run(self):
...
     PyBind11Found = True
-    windows = (not system()) or system() == "Windows"
+    WindowsFound = (not system()) or system() == "Windows"
+    ReadTheDocsFound = os.getenv("READTHEDOCS", False) == "True"
+    # If running on Windows, check if building a wheel.
+    if WindowsFound:
+        WindowsWheelFound = os.getenv("CIBUILDWHEEL", 0) == "1"
+
+    windows_msg = "Running on Windows" if WindowsFound else "Not running on Windows"
+    pybind11_msg = "Could not find pybind11. Skipping compilation of KLU module."
```
agriyakhetarpal, line 267: So I am having a bit of trouble making the compilation of the IDAKLU solver _optional_ on macOS and Linux platforms, especially because it breaks the installation on source distributions (CMake is installed by default since it is present in the build-system dependencies in `pyproject.toml`). Doing so on Windows is easy since we mention that the IDAKLU solver is not available on Windows (users can follow along with our complicated development procedure for the Windows wheels with MSVC and so on if they want to, though)

- One option I am considering is looking for the relevant SUNDIALS and SuiteSparse `.dylib` or `.so` files in the `$HOME/.local/lib` folder or a user-specified `SUNDIALS_INST` environment variable on macOS and Linux. The issue with this is that it can cause issues with the `pybamm_install_odes` entry point that is reserved for user installations via PyPI – SUNDIALS installed using that entry point does not come with KLU, which would further break someone's IDAKLU installation if they try to install from source sometime later (SUNDIALS + KLU are not compatible with `scikits.odes` because of double vs extended precision constraints). We could document this in our instructions but it doesn't feel like a permanent solution and it feels like we would be then going the other way since the goal is to simplify the installation.
- Another method is to use some build option that gets passed to `setup.py`—`pip` has a `--global-option` for this purpose as far as I am aware—which turns on the IDAKLU solver compilation procedure, but it is still difficult to ensure that the compilation remains an opt-in feature rather than an opt-out one (see also: https://discuss.python.org/t/optional-c-extension-handling/1595/7)
- We can scan for most environment variables during `setup.py` (it is a Python script after all and is meant for Python trickery like this) and turn compilation on or off based on that, I did that for Read the Docs but the doctests are broken currently and we will soon run out of options like those
- Another way: ensure that we can support more platforms (#3462) either via cross-compilation or with the use of native external CI runners (Cirrus CI has had arm64 runners for both Linux and macOS since some while now) until GitHub Actions gets them – this way we will have wheels for every major platform and architecture and we will not need to bother with `setup.py` (wheels don't contain a `setup.py`).
- The best option I can think of at this moment is to remove CMake from the build-time dependencies and ask users to install it on their own in the docs when doing the source installation (it is pretty easy to do so via `pip` or `brew` so it doesn't defeat the purpose of making `pybind11` a build-time dependency). We check for it in the `compile_KLU()` function and return False if it is not found – the CI can be configured to fix that pretty easily.

I think I will go with the last option for now unless there is something I have been missing or anything has been overlooked (cc: @kratman @Saransh-cpp).

agriyakhetarpal:
I'm closing this PR; it's quite out of date now with what we're planning to do in #3564, and @cringeyburger can take a look at this in the coming weeks.

PR #3573, opened 2023-11-28
#2188 implement evaluate at in 1D

### LINKED ISSUE #2188: Introduction of 3 electrode setup
### Description

The feature involves additional potential variables. The first one equates to a potential difference between the current collector of the anode and the middle of separator called "Negative electrode potential vs separator reference". The second one is the potential difference between the middle of the separator and the current collector of the positive electrode called "Positive electrode potential vs sparator reference".

### Motivation

Motivation for this comes from the 3-electrode setup in the lab. 

In a 2-electrode setup in the lab one can not resolve the individual SOCs of each electrode. You only get the potential difference between both current collectors. In a 3-electrode setup a reference electrode is introduced in the middle of the separator for example a little place of lithium or a whole lithium reference ring around the separator. With the reference electrode, the potentials of both electrodes can be monitored separately. These potentials can for example be used for plating detection on lithium ion graphite anodes.

Modelling this feature directly in PyBaMM is not yet possible. There are already a lot of fancy ways in pybamm to resolve processes on the electrodes directly for example different overpotential contributions. These though can not be compared directly to measured data in the lab. With the mentioned feature there is a way to describe these electrode potentials in PyBaMM and relate them directly to measured potentials in the lab.  The additional potential variables could help a great deal for fitting models better to measurements and increase model validity.

### Possible Implementation

As a rough draft for my own simulations i implemented the following lines in "base_electrolyte_conductivity". It could be placed in the current-collector submodel or maybe also "base electrode" but since the separator electrolyte potential is needed this would involve including a new submodel at these places. For now these lines worked for me:

```
        phi_e_s_mid = pybamm.x_average(phi_e_s)  # CAVEAT
        phi_e_s_mid_dim = -param.U_n_ref + phi_e_s_mid * param.potential_scale
        v_n = pybamm.boundary_value(variables["Negative electrode potential [V]"], 'left') \
              - phi_e_s_mid_dim
        v_p = pybamm.boundary_value(variables["Positive electrode potential [V]"], 'right') \
              - phi_e_s_mid_dim
        variables.update(
            {
                "Negative electrode potential vs reference [V]": v_n,
                "Positive electrode potential vs reference [V]": v_p,
            }
        )
```
There is one caveat though. For a point of reference we need a single value (here termed "phi_e_s_mid"). Typically this would just be the evaluation of the electrolyte in the middle of the separator but until now i did not manage to get it working. I tried to do it with `phi_e_s.evaluate(param.L_n+param.L_s/2)' but it didn't want to be evaluated as it said the method "evaluate is not impleted for "Current function"'. I did not try it with another approach as this wouldn't be working for my system.

### Additional context

Provided there is feedback and discussion and it's not too much work to put the feature inside while making sure that everything else in the code still works i would implement this and create a pull request.

rtimms:
# Description
Implement a new unary operataor `EvaluateAt` that evaluates a spatial variable at a given position. For now, this is only implemented for 1D spatial variables using the Finite Volume method. In this method, `EvalauteAt` returns the variable evaluated at the node closest to the requested value. E.g. `EvaluateAt(3*x, 1.9)` on a mesh with nodes`x=[0.5, 2.5, 4.5]` would return 7.5.

The motivation for this is to be able to place a reference electrode at a given point in the cell, to mimic a 3E setup.

Fixes #2188 

## Type of change

# Key checklist:

## Further checks:


tests/unit/test_expression_tree/test_operations/test_jac.py
```diff
@@ -236,6 +236,12 @@ def test_index(self):
         jac = ind.jac(vec).evaluate(y=np.linspace(0, 2, 5)).toarray()
         np.testing.assert_array_equal(jac, np.array([[0, 0, 0, 0, 0]]))
 
+    def test_evluate_at(self):
```
tinosulzer, line 239:
```suggestion
    def test_evaluate_at(self):
```

examples/scripts/3E_cell.py
```diff
@@ -0,0 +1,53 @@
...
+# evaluate the electrolyte potential at the mid-point of the cell
+phi_e_mid = pybamm.EvaluateAt(model.variables["Electrolyte potential [V]"], L_mid)
+
+# add the new variables to the model
+model.variables.update(
+    {
+        "Negative electrode 3E potential [V]": phi_n - phi_e_mid,
+        "Positive electrode 3E potential [V]": phi_p - phi_e_mid,
+    }
+)
```
tinosulzer, line 33: can you create a helper method that does this?

tinosulzer: ideally in a way that doesn't require the `parameter_values` to be passed in, i.e. the "value" ("position") should be processed by `ParameterValues` when setting parameters

pybamm/expression_tree/unary_operators.py
```diff
@@ -1012,6 +1050,51 @@ def __init__(self, child, side):
...
+class EvaluateAt(SpatialOperator):
+    """
+    A node in the expression tree which evaluates a symbol at a given position. Only
+    implemented for variables that depend on a single spatial dimension.
+
+    Parameters
+    ----------
+    child : :class:`pybamm.Symbol`
+        The variable whose boundary value to take
+    value : float
```
tinosulzer, line 1062: "value" is a bit misleading, change to "position"

tinosulzer (changes requested):
Thanks, suggesting a few changes to improve this.

Is the reason not to make this a standard variable that it doesn't work in >1D ? If so, how much work is it really to make it work in higher dimensions? Should just be some kind of kron of the matrix

examples/scripts/3E_cell.py
rtimms: yep, will do

rtimms:
> Is the reason not to make this a standard variable that it doesn't work in >1D ? If so, how much work is it really to make it work in higher dimensions? Should just be some kind of kron of the matrix

Was trying to avoid the complexity of passing in x,y,z,r etc. for now, but can spend more time on this to make it more general. It's probably straightforward for battery models where we can map the input variable to the dimension, but it's harder in general I think.

tinosulzer:
> Was trying to avoid the complexity of passing in x,y,z,r etc. for now, but can spend more time on this to make it more general. It's probably straightforward for battery models where we can map the input variable to the dimension, but it's harder in general I think.

Ah yeah I see. I think it's still ok if you limit it to primary domain and the domain has to be 1D

pybamm/spatial_methods/finite_volume.py
```diff
@@ -1023,6 +1023,50 @@ def boundary_value_or_flux(self, symbol, discretised_child, bcs=None):
...
+        # Get mesh nodes
+        domain = discretised_child.domain
+        mesh = self.mesh[domain]
+        nodes = mesh.nodes
+
+        # Find the index of the node closest to the value
+        index = np.argmin(np.abs(nodes - value))
+
+        # Create a sparse matrix with a 1 at the index
+        matrix = csr_matrix(([1], ([0], [index])), shape=(1, mesh.npts))
```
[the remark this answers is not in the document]
rtimms: I'm not sure linearly interpolating is the right thing to do. Choosing the nearest value is consistent with FVM where the nodes are cell-centered and the function is piecewise constant. Also, you can construct examples where "closest two" isn't well defined. E.g. if the mesh nodes are [025, 0.5, 0.75] and I evaluate at 0.5, what should I do? This kind of example comes up naturally when you evaluate at the midpoint and you have an odd number of nodes.

Choosing the nearest node is basically like asking "which volume am I in?".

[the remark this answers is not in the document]
tinosulzer: Good point, makes sense

tinosulzer (commented):
Thanks for the updates. Why can't the reference electrode always be inserted into the model, even for dimensions > 0 ?

rtimms:
> Why can't the reference electrode always be inserted into the model, even for dimensions > 0 ?

Since `EvaluateAt` is only implemented for 1D and for the primary dimension only, when dimensionality>0 the placement of the reference electrode is no longer a single point, it's `phi_e(position, y, z)`, so the definition of the 3E potentials is ambiguous.

tinosulzer:
In a real system how does it work? Could we evaluate at the "tab" like we do for the voltage?

brosaplanella:
> In a real system how does it work? Could we evaluate at the "tab" like we do for the voltage?

For the system I am (vaguely) familiar with, the EL cells, there is a lithium ring placed between an electrode and the separator. I think in other setups they just place like a small tab inside, but not sure how much precision they have when placing it.

rtimms:
> Could we evaluate at the "tab" like we do for the voltage?

We can for the solid potentials, but where in y,z do we evaluate the electrolyte potential?

tinosulzer (approved):
Fair enough, let's leave as separate method for now, please update changelog to document the separate method

PR #3578, opened 2023-11-30
Adds typing to expression tree

### LINKED ISSUE #3497: Add type hints to expression tree
### Description

Type hints improve code readability and error detection, and make it easier (particularly for new contributors) to follow the structure & flow of data through large projects like PyBaMM.

Static type checkers such as MyPy/Pyright can be added at a later date (see #3489) and allow for gradual annotation of a codebase, so we don't need to add hints to the entire project at once.

As part of developing a solution to #3397 I added type hints to the bulk of the expression tree; this later turned out to be unneeded for serialisation but as they are useful more generally I figured they could be added via this issue instead.

pipliggins:
# Description

Adds type hints to the majority of files within the expression tree.

Fixes #3497 

## Type of change

# Key checklist:


Saransh-cpp (commented):
@pipliggins you might want to merge develop. We introduced a few changes to the configuration of pre-commit and ruff recently.

pipliggins:
Hi @martinjrobins, I've completed the changes in this PR so it's ready to review. The failing doctests appear to be unrelated to the modifications introduced here.

We could consider the inclusion of a MyPy check in the CI pipeline at this point. However, I propose that we refrain from making it a mandatory requirement for now - we can just use it to issue warnings to users about potential issues.

Also, @Saransh-cpp, you previously suggested using the isort rules in ruff to automatically insert the `from __future__ import annotations` line in files utilising newer type hints. I've explored this but haven't found a way to achieve it without affecting all files in the project (including noxfiles, setup files etc). Am I missing a more targeted approach to implement this?

Saransh-cpp:
> Also, @Saransh-cpp, you previously suggested using the isort rules in ruff to automatically insert the from __future__ import annotations line in files utilising newer type hints. I've explored this but haven't found a way to achieve it without affecting all files in the project (including noxfiles, setup files etc). Am I missing a more targeted approach to implement this?

It should be okay to have this import in all the files, but it can be done in a separate PR. Setup files and noxfile should also ideally have type hints (nox provides type hints for nox sessions).

mypy.ini
```diff
@@ -0,0 +1,10 @@
+[mypy]
```
Saransh-cpp, line 1: This should be moved to `pyproject.toml` under a new section - `[tool.mypy]`.

pyproject.toml
```diff
@@ -218,7 +218,6 @@ ignore = [
     "RET506",      # Unnecessary `elif`
     "B018",        # Found useless expression
     "RUF002",      # Docstring contains ambiguous
-    "UP007",       # For pyupgrade
```
Saransh-cpp, line 221: This should not be removed ideally.

Saransh-cpp:
> We could consider the inclusion of a MyPy check in the CI pipeline at this point.

It can be included as a pre-commit hook -
```yaml
- repo: https://github.com/pre-commit/mirrors-mypy
  rev: v1.8.0
  hooks:
    - id: mypy
      files: pybamm
      args: []
      additional_dependencies:
        - numpy
        - packaging
```

but will this fail for the expressions that don't have a type hint at the moment? I really don't remember if it fails or if it passes (with or without the `strict` mode).

Saransh-cpp (commented):
Thanks for working on this! Minor comments below (will go through the type hints too).

pipliggins:
> > We could consider the inclusion of a MyPy check in the CI pipeline at this point.
> 
> It can be included as a pre-commit hook -
> 
> ```yaml
> - repo: https://github.com/pre-commit/mirrors-mypy
>   rev: v1.8.0
>   hooks:
>     - id: mypy
>       files: pybamm
>       args: []
>       additional_dependencies:
>         - numpy
>         - packaging
> ```
> 
> but will this fail for the expressions that don't have a type hint at the moment? I really don't remember if it fails or if it passes (with or without the `strict` mode).

Anything that doesn't have a type hint mypy will ignore (more accurately, it assumes it has type "Any") without "strict" mode. If "strict" is on it will fail for the untyped expressions.

MyPy themselves acknowledge that the "strict" flag will probably be too aggressive for scenarios like this where we're adding typing to a large existing codebase. They have a number of suggestions around how to approach this [here](https://mypy.readthedocs.io/en/stable/existing_code.html#existing-code)

pybamm/experiment/experiment.py
```diff
@@ -73,7 +73,7 @@ def __init__(
         for cycle in operating_conditions:
             # Check types and convert to list
             if not isinstance(cycle, tuple):
-                cycle = (cycle,)
+                cycle = (cycle,)  # type: ignore[assignment]
```
kratman, line 76: Is it possible to avoid these type of ignore comments for the typing?

pybamm/expression_tree/array.py
```diff
@@ -1,12 +1,17 @@
...
+from __future__ import annotations
 import numpy as np
 from scipy.sparse import csr_matrix, issparse
+from typing import TYPE_CHECKING
 
 import pybamm
 from pybamm.util import have_optional_dependency
 
+if TYPE_CHECKING:  # pragma: no cover
+    import sympy
```
kratman, line 13: Why do we need to enable type checking with a global variable?

kratman:
> We could consider the inclusion of a MyPy check in the CI pipeline at this point. However, I propose that we refrain from making it a mandatory requirement for now - we can just use it to issue warnings to users about potential issues.

Using mypy quickly escalates, we should use caution when considering using it. I tend to consider types as nice-to-haves in python, but I am certain that there are a lot of types that would need to be fixed if we use it

pipliggins:
> > We could consider the inclusion of a MyPy check in the CI pipeline at this point. However, I propose that we refrain from making it a mandatory requirement for now - we can just use it to issue warnings to users about potential issues.
> 
> Using mypy quickly escalates, we should use caution when considering using it. I tend to consider types as nice-to-haves in python, but I am certain that there are a lot of types that would need to be fixed if we use it

I would tend to agree with this, but think there are ways to use it to aid consistent type hints without being bullied into making lots of changes & fixes immediately. I've linked this comment in the issue attached to the PR, I suggest discussions on the pros/cons of adding mypy are carried out there :)

pybamm/experiment/experiment.py
pipliggins: These comments are used by mypy when static type checking. I've done my best to use them as little as possible, but in some cases the amount of rewriting required to make mypy happy would make the code less readable than having the comment (in my opinion). If we decide not to use a type checker, they can be removed.

pybamm/expression_tree/array.py
pipliggins: Related to above - this is a statement used by mypy. In this case sympy is used in the file only as a type hint, not during runtime. `TYPE_CHECKING` is a constant assumed to be True by static type checkers, but is false at runtime; so sympy will only be imported if a type checker like mypy is being run. https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING

kratman: Personally I would prefer to just have sympy always imported. If it is used in the type annotations, then why not just always have it there?

Saransh-cpp: That would make sympy a required dependency. Sympy is heavy and I would avoid making it a required dependency (only a subset of features in PyBaMM require sympy)

kratman: I would say that if expression tree is using sympy types, then it is really a dependency. I have not downloaded this branch locally to double check, but I would guess that my IDE would complain if I did not have sympy installed too.

As we add type checking it is going to expose more things that should probably be real dependencies rather than optional ones.

pybamm/experiment/experiment.py
pipliggins: This instance isn't actually a brilliant example as I've managed to get rid of the type error; however there are ~ 10 ignore statements I've had to leave on a line-by-line basis. Mypy raises an error for each `# type: ignore[]` comment even with strict mode turned off, but if we don't enforce mypy as having to pass in the CI then we could chose to use it as a 'todo' list of sorts and remove the `# type: ignore` comments.

The remaining errors are split about 50/50 between difficulties integrating mypy into heavily dynamically-typed areas (e.g. it won't let you re-define a variable with a different type before you've used it, even with the --allow-redefinition flag https://mypy.readthedocs.io/en/latest/command_line.html#miscellaneous-strictness-flags), and areas where pybamm would require more significant re-writing. For example in `serialise.py`, where mypy complains that `BaseModel` doesn't have a `bounds` attribute. This is true at the point where BaseModel is defined, but the attribute is added once the model is discretised; as that particular function only takes a discretised model, the code works properly but I haven't found a way to flag to mypy that this is the case. There are multiple ways to fix this, including creating a `DiscretisedModel` class or editing the `__init__` of `BaseModel` to include attributes that are added later, but that's a lot of work & discussion that shouldn't be done here.

pybamm/expression_tree/array.py
pipliggins: sympy is only used in the expression tree for `to_equation()` to use with LaTeX I believe, rather than the solvers, hence the optional dependency.

kratman: Yeah my point was more that we might have to rethink our optional dependencies if we are using static types. Expression tree depends on sympy if it is using sympy types, but by putting this check here we are pretending that it doesn't depend on sympy.

It is too much for this task, but if we want a typed library and not have all the dependencies for types then we might need to think about how to better separate our subcomponents. Maybe things like the latexify need to go into a separate module if we want it's dependencies to be truly optional

pybamm/experiment/experiment.py
```diff
@@ -39,7 +39,7 @@ class Experiment:
 
     def __init__(
         self,
-        operating_conditions: list[str],
+        operating_conditions: list[str | tuple[str]],
```
martinjrobins, line 42: should this be an `collections.Iterable` type?

pybamm/expression_tree/binary_operators.py
```diff
@@ -10,8 +11,15 @@
 import pybamm
 from pybamm.util import have_optional_dependency
 
+from typing import Union, Callable
 
-def _preprocess_binary(left, right):
+# create type alias(s)
+ChildValue = Union[float, np.ndarray, pybamm.Symbol]
```
martinjrobins, line 17: This is a bit pedantic, but I would probably call this `ChildSymbol` rather than a value. It only becomes a value when the tree is evaluated, and so you could have both, ie.:

```python
ChildSymbol = Union[float, np.ndarray, pybamm.Symbol]
ChildValue = Union[float, np.ndarray]
```

pybamm/expression_tree/binary_operators.py
```diff
@@ -229,19 +253,23 @@ class Addition(BinaryOperator):
...
         """See :meth:`pybamm.BinaryOperator.__init__()`."""
         super().__init__("+", left, right)
 
-    def _diff(self, variable):
+    def _diff(self, variable: pybamm.Symbol):
         """See :meth:`pybamm.Symbol._diff()`."""
         return self.left.diff(variable) + self.right.diff(variable)
 
-    def _binary_jac(self, left_jac, right_jac):
+    def _binary_jac(self, left_jac: float | np.ndarray, right_jac: float | np.ndarray):
```
martinjrobins, line 268: And if you define `ChildValue` like I suggest above, this could then be

```suggestion
    def _binary_jac(self, left_jac: ChildValue, right_jac: ChildValue):
```

pybamm/expression_tree/binary_operators.py
```diff
@@ -229,19 +253,23 @@ class Addition(BinaryOperator):
...
         """See :meth:`pybamm.BinaryOperator._binary_jac()`."""
         return left_jac + right_jac
 
-    def _binary_evaluate(self, left, right):
+    def _binary_evaluate(self, left: float | np.ndarray, right: float | np.ndarray):
```
martinjrobins, line 272: and here

martinjrobins: others below as well

pybamm/expression_tree/broadcasts.py
```diff
@@ -417,15 +462,17 @@ class FullBroadcast(Broadcast):
 
     def __init__(
         self,
-        child,
-        broadcast_domain=None,
-        auxiliary_domains=None,
-        broadcast_domains=None,
-        name=None,
+        child_input: NumberType | float | pybamm.Symbol,
```
martinjrobins, line 465: sort of overlaps with `ChildValue` / `ChildSymbol` defined in binary ops file (except no ndarray, should this be here as well?). Can we just define this in one place and re-use everywhere?

Also, surely float is a `NumberType`?

pybamm/expression_tree/broadcasts.py
```diff
@@ -489,11 +536,11 @@ class FullBroadcastToEdges(FullBroadcast):
 
     def __init__(
         self,
-        child,
-        broadcast_domain=None,
-        auxiliary_domains=None,
-        broadcast_domains=None,
-        name=None,
+        child: NumberType | float | pybamm.Symbol,
```
martinjrobins, line 539: `ChildSymbol` here as well

pybamm/expression_tree/functions.py
```diff
@@ -87,11 +89,11 @@ def diff(self, variable):
 
             derivative = sum(partial_derivatives)
             if derivative == 0:
-                derivative = pybamm.Scalar(0)
+                return pybamm.Scalar(0)
 
             return derivative
 
-    def _function_diff(self, children, idx):
+    def _function_diff(self, children: Sequence[pybamm.Symbol], idx: float):
```
martinjrobins, line 96: In a previous comment I mentioned `Iterable`, perhaps this should be `Sequence`. idk the difference..?

pybamm/expression_tree/independent_variable.py
```diff
@@ -118,8 +126,13 @@ class SpatialVariable(IndependentVariable):
     """
 
     def __init__(
-        self, name, domain=None, auxiliary_domains=None, domains=None, coord_sys=None
-    ):
+        self,
+        name: str,
+        domain: list[str] | str | None = None,
```
martinjrobins, line 131: The types for the different domains (domain, auxilary_domains and domains) are also repeated a lot. As with `ChildValue`/`ChildSymbol` these should be defined in on place and used throughout.

pybamm/expression_tree/symbol.py
```diff
@@ -562,71 +572,75 @@ def __repr__(self):
             {k: v for k, v in self.domains.items() if v != []},
         )
 
-    def __add__(self, other):
+    def __add__(self, other: Symbol | float | np.ndarray) -> pybamm.Addition:
```
martinjrobins, line 575: `ChildSymbol` here as well, more below

pybamm/expression_tree/symbol.py
```diff
@@ -738,7 +757,13 @@ def _jac(self, variable):
         """
         raise NotImplementedError
 
-    def _base_evaluate(self, t=None, y=None, y_dot=None, inputs=None):
+    def _base_evaluate(
+        self,
+        t: float | None = None,
```
martinjrobins, line 762: derived classes (e.g. scalar) have different types than this base function, should be consistent

pybamm/expression_tree/broadcasts.py
pipliggins: Types from `numbers` aren't supported for static type checking (at least in mypy) - they recommend hinting using `complex/float/int` instead according to https://peps.python.org/pep-0484/#the-numeric-tower.

If we want to keep using `numbers.Numbers` in functions like this one, the suggestion is to use `typing.SupportsFloat` (which is what `NumberType` is defined as at the top of the file), which just checks if the class has a `_float_` function (the accepted answer here was really useful to explain the vagaries https://stackoverflow.com/questions/69334475/how-to-hint-at-number-types-i-e-subclasses-of-number-not-numbers-themselv) - although if we start supporting complex numbers this would break.

What I haven't quite worked out is why mypy complains when type-hinting with `numbers.Number` in `FullBroadcast` & `FullBroadcastToEdges`, but not for `PrimaryBroadcast`, `Scalar`, and a few other functions...

pybamm/expression_tree/functions.py
pipliggins: Any `Sequence` is an `Iterable` - but `Sequence` objects must define `_len_`. As `len(children)` is called in this function, Sequence is appropriate.

pybamm/experiment/experiment.py
pipliggins: I don't think so, 
```
for cycle in operating_conditions:
            if not isinstance(cycle, tuple):
                cycle = (cycle,)
```
suggests that it's looking specifically for a tuple, not just any Sequence?

martinjrobins (approved):
great stuff, happy to merge now @pipliggins

PR #3582, opened 2023-12-01
Diffusion coefficient as 2D interpolant

Scottmar93:
# Description

I was getting errors with the current 2D interpolation when I made the diffusion coefficient in the particle a 2D lookup table (I've added a minimal example of this in the examples of this PR which will fail if run on the main branch). The issue was related to resizing after the interpolating function is called. As a potential fix, I've reshaped the output of the function to match the size of the input. This allow interpolating functions to be evaluated on 2D domains such as the electrode particles and have the result maintain that shape. 

I think the issue was introduced when I added the 3D lookup tables for the ECM model. Tests seems to still work and ECM model still runs after the changes. 

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/examples/scripts/minimal_example_of_lookup_tables.py
+++ b/examples/scripts/minimal_example_of_lookup_tables.py
@@ -1,0 +1,51 @@
+import pybamm
+import pandas as pd
+import numpy as np
+
+
+def process_2D(name, data):
+    data = data.to_numpy()
+    x1 = np.unique(data[:, 0])
+    x2 = np.unique(data[:, 1])
+
+    value = data[:, 2]
+
+    x = (x1, x2)
+
+    value_data = value.reshape(len(x1), len(x2), order="C")
+
+    formatted_data = (name, (x, value_data))
+
+    return formatted_data
+
+
+parameter_values = pybamm.ParameterValues(pybamm.parameter_sets.Chen2020)
+
+# overwrite the diffusion coefficient with a 2D lookup table
+D_s_n = parameter_values["Negative electrode diffusivity [m2.s-1]"]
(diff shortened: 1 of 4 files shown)


examples/scripts/minimal_example_of_lookup_tables.py
```diff
@@ -0,0 +1,51 @@
...
+
+# overwrite the diffusion coefficient with a 2D lookup table
+D_s_n = parameter_values["Negative electrode diffusivity [m2.s-1]"]
+df = pd.DataFrame(
+    {
+        "sto": [0, 1, 0, 1, 0, 1],
+        "T": [0, 0, 25, 25, 45, 45],
+        "D_s_n": [D_s_n, D_s_n, D_s_n, D_s_n, D_s_n, D_s_n],
+    }
+)
```
tinosulzer, line 32: is this the standard way you would provide this data (with repetition of sto and T)?

tinosulzer (approved):
Thanks, looks good to me

examples/scripts/minimal_example_of_lookup_tables.py
Scottmar93: Just noticed that the T column (which is changing slower) should actually be the first column so I will update this

I've seen this format used a bit for 2D lookup tables (also used it for the ECM lookup tables in PyBaMM); is there another format you use instead?

tinosulzer: I didn't have anything else in mind specifically, seems clunky to have to repeat entries but can be done easily enough with code

tinosulzer (approved):

PR #3593, opened 2023-12-05
Plett OCP Hysteresis Model

### LINKED ISSUE #3375: Decay rate for single state hysteresis models
### Description

Single state hysteresis models typically feature a decay rate. So far, this has not been implemented in pybamm as far as I can see. 

Previous hysteresis models implemented in pybamm: 
https://github.com/pybamm-team/PyBaMM/pull/2256 and https://github.com/pybamm-team/PyBaMM/pull/3194 

### Motivation

Voltage due to hysteresis can have a long timescale, which is left unmodeled in pybamm. This prevents non-instantaneous hysteresis dynamics from being modeled in pybamm. 

### Possible Implementation

Decay parameter, dependent on change in lithiation. Personally, I also think it would make more sense to have a single OCP with a hysteresis term than switch between two OCPs. 
[Latz_Plett_hysteresis (1).pdf](https://github.com/pybamm-team/PyBaMM/files/12748922/Latz_Plett_hysteresis.1.pdf)

### Additional context

_No response_

mleot:
# Description

Adds the hysteresis model as outlined in [Wycisk 2022](https://www.sciencedirect.com/science/article/abs/pii/S2352152X22010192?via%3Dihub), and as mentioned in 
#3375 and as discussed in #3376 .

This will allow for hysteresis switching to depend not on the current, but on the differential capacity change.  

Fixes #3375 

## Type of change

# Key checklist:

## Further checks:


pybamm/models/submodels/interface/open_circuit_potential/plett_ocp.py
```diff
@@ -0,0 +1,114 @@
+#
+# OCP with empiral hysteresis function, and hysteresis state function
+#
+import pybamm
+from . import BaseOpenCircuitPotential
+
+
+class PlettOpenCircuitPotential(BaseOpenCircuitPotential):
```
tinosulzer, line 8: can you document the Plett paper this is taken from?

tinosulzer: string in `.register()` needs to match the name of the entry in the .bib file

pybamm/parameters/lithium_ion_parameters.py
```diff
@@ -630,6 +634,27 @@ def U(self, sto, T, lithiation=None):
...
+
+    def Q(self, sto):
+        """Capacity change as a function of stoichiometry"""
+        c_max = self.c_max
+        epsilon_s_av = self.epsilon_s_av
+        V_electrode = self.main_param.A_cc*self.domain_param.L
+        Li_max = c_max * V_electrode * epsilon_s_av
+        Q_max = Li_max * self.main_param.F / 3600
+        return Q_max * sto
+
```
mleot, line 656: I had to add a function to get the capacity state stored in the electrode in order to calculated the dQdV. I wonder if there is a better way to do this, and I would love feedback from anyone more familiar with the repo.

pybamm/models/submodels/interface/open_circuit_potential/plett_ocp.py
```diff
@@ -0,0 +1,135 @@
...
+            variables[f'{Domain} electrode {phase_name}OCP hysteresis [V]'] = H
+            # dQdU = variables[f'{Domain} electrode {phase_name}differential capacity [A.s.V-1]']
+
+            dU = pybamm.source(1,ocp_bulk_eq.diff(sto_bulk))
+            # dU = pybamm.PrimaryBroadcast(dU,[f'{domain} electrode'])
+            # dU = pybamm.PrimaryBroadcast(dU,[f'{domain} particle size'])
+            # dU = pybamm.FullBroadcast(dU,broadcast_domains={'primary':f'{domain} particle size', 'secondary':f'{domain} electrode'})
+            Q = variables[f"Total lithium in {phase} phase in {domain} electrode [mol]"] * pybamm.constants.F / 3600
+            # Q = pybamm.PrimaryBroadcast(Q,f'{domain} electrode')
+            dQ = Q.diff(sto_bulk)
```
mleot, line 93: @tinosulzer I am pretty sure that the correct domains for the dU  and  dQ variables should be  the current collector and the electrode. In this case, I am trying to differentiate Q by sto_bulk, although I often end up with a scalar 0 after. Are there specific rules or best practices for differentiating?

pybamm/models/submodels/interface/open_circuit_potential/plett_ocp.py
```diff
@@ -0,0 +1,127 @@
...
+            # h = pybamm.PrimaryBroadcast(h,[f'{domain} electrode'])
+            H_x_av = pybamm.x_average(H)
+            h_x_av = pybamm.x_average(h)
+            if domain_options["particle size"] == "distribution":
+                if sto_surf.domains['primary'] == f"{domain} electrode":
+                    ocp_surf = ocp_surf_eq + H * h
+                else:
+                    ocp_surf = ocp_surf_eq + H_x_av * h_x_av
+            else:
+                ocp_surf = ocp_surf_eq + H * h
```
tanner-leo, line 93: So I have been able to get the DFN tests to pass and MPM tests to pass, although I realized I needed to x_average the h and H variables for the MPM, but not for the DFN with a psd. Is there any variable where I can have that logic execute automatically?

mleot:
@tinosulzer just bumping this comment again to see if you had any ideas

pybamm/models/submodels/interface/open_circuit_potential/plett_ocp.py
rtimms: There's no automatic way of doing this. If you look, for example, in the interface submodels you'll see there's a lot of logic in there for x-averaged vs not x-averaged and PSD vs no PSD.

rtimms: Something like https://github.com/pybamm-team/PyBaMM/blob/develop/pybamm/models/submodels/interface/open_circuit_potential/current_sigmoid_ocp.py#L26 might be relevant

pybamm/models/submodels/interface/open_circuit_potential/plett_ocp.py
mleot: The citation registration was placed in section which checks the model options

pybamm/parameters/lithium_ion_parameters.py
mleot: Okay I added it as a coupled variable and removed it from the lithium ion parameters.

pybamm/models/submodels/interface/open_circuit_potential/plett_ocp.py
mleot: I finished implementing the dU calculation

pybamm/models/submodels/interface/open_circuit_potential/plett_ocp.py
mleot: I think I managed to get a version working (passing all the tests) though it's not very pretty in it's current state.

pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -263,7 +263,7 @@ def __init__(self, extra_options):
                 "stress and reaction-driven",
             ],
             "number of MSMR reactions": ["none"],
-            "open-circuit potential": ["single", "current sigmoid", "MSMR"],
+            "open-circuit potential": ["single", "current sigmoid", "MSMR", "Plett"],
```
mleot, line 266: Is it okay to call it "Plett" OCP when it's from the Wycisk modified plett paper?

pybamm/parameters/lithium_ion_parameters.py
```diff
@@ -535,6 +535,12 @@ def _set_parameters(self):
         eps_c_init_av = pybamm.xyz_average(
             self.epsilon_s * pybamm.r_average(self.c_init)
         )
+        # if self.options['open-circuit potential'] == 'Plett':
+        self.K = pybamm.Parameter(f"{pref}{Domain} particle hysteresis decay rate")
+        self.K_x = pybamm.Parameter(
+            f"{pref}{Domain} particle hysteresis switching factor"
+        )
+        self.h_init = pybamm.Scalar(0)
```
mleot, line 543: I added these as default parameters. Not sure if I should try to create them only when specific model options have been turned on. What do you reccomend?

mleot:
@rtimms Okay I believe everything here is in working order. Is an example notebook necessary? What else do I need to do before it can be merged into develop?

pybamm/models/full_battery_models/base_battery_model.py
rtimms: Can we give it a more informative name like "Differential capacity varying" (or something more concise!). Then we can point to the paper in the docs.

pybamm/models/submodels/interface/open_circuit_potential/__init__.py
```diff
@@ -2,3 +2,4 @@
 from .single_ocp import SingleOpenCircuitPotential
 from .current_sigmoid_ocp import CurrentSigmoidOpenCircuitPotential
 from .msmr_ocp import MSMROpenCircuitPotential
+from .plett_ocp import PlettOpenCircuitPotential
```
rtimms, line 5: similarly, I think we should try to give the class a name that at least hints at the physics rather than the author

pybamm/models/submodels/interface/open_circuit_potential/plett_ocp.py
```diff
@@ -0,0 +1,151 @@
...
+        domain, Domain = self.domain_Domain
+        phase_name = self.phase_name
+        phase = self.phase
+
+        # get dQ, change in capacity with respect to stoichiometry
+        c_max = self.phase_param.c_max
+        epsilon_s_av = self.phase_param.epsilon_s_av
+        V_electrode = self.param.A_cc * self.domain_param.L
+        Li_max = c_max * V_electrode * epsilon_s_av
+        Q_max = Li_max * self.param.F / 3600
```
rtimms, line 33: There is already a variable `f"{Domain} electrode capacity [A.h]"` defined in the active material model that does this calculation also accounting for any change in active material fraction

pybamm/models/submodels/interface/open_circuit_potential/plett_ocp.py
```diff
@@ -0,0 +1,151 @@
...
+        dQ = Q_max
```
rtimms, line 34: This isn't `dQ`, right, it's just the electrode capacity? I think later on you are writing `dU/dQ = dU/dx * dx/dQ`? Can you point to which equation the paper you've implemented as they seem to express it a few ways in the paper?

pybamm/models/submodels/interface/open_circuit_potential/plett_ocp.py
```diff
@@ -0,0 +1,151 @@
...
+        domain, Domain = self.domain_Domain
+        phase_name = self.phase_name
+
+        current = self.param.current_with_time
+        current = variables[f"{Domain} electrode interfacial current density [A.m-2]"]
+        Q_cell = variables[f"{Domain} electrode capacity [A.h]"]
+        dQdU = variables[
+            f"{Domain} electrode {phase_name}differential capacity [A.s.V-1]"
+        ]
+        dQdU = dQdU.orphans[0]
```
rtimms, line 137: why do you need to do this?

pybamm/parameters/lithium_ion_parameters.py
rtimms: They will only appear in the expression tree if these model options are turned on, so it's fine to add them here.

rtimms (commented):
Thanks @mleot this is mostly looking good! I'm just not sure of the dQ/dU calculation.

pybamm/models/full_battery_models/base_battery_model.py
mleot: How about "Differential capacity hysteresis state" as "DCHS" ?

pybamm/models/submodels/interface/open_circuit_potential/__init__.py
mleot:
```suggestion
from .dchs_ocp import DCHSOpenCircuitPotential
```

pybamm/models/submodels/interface/open_circuit_potential/plett_ocp.py
mleot: Originally I was calculating dQ(sto) by means of a function in the lithium ion parameters (main parameters for a phase I believe), but @valentinsulzer suggested placing it in the submodel class as a coupled variable. 

Since dQ(sto) should be a constant value, I realized this could just be a scalar. Although, I am not sure whether it needs to be an equation as d(sto) may be different depending on the number of steps during solving. I still haven't tested this (looked at actual data or results) with the change from I made, so I am not sure if I will actually get the results I expect. All the more reason to make an example notebook.

I will make the example notebook and assess whether the function form of Q was necessary or not.

pybamm/models/submodels/interface/open_circuit_potential/plett_ocp.py
mleot: Yeah my thinking here is that dQ(sto) would be equal to Q*sto, so maybe I need to multiply it by sto.

```suggestion
        dQ = Q_max * sto_bulk
```

pybamm/models/submodels/interface/open_circuit_potential/plett_ocp.py
mleot: I see. I believe it's because I need the individual phase capacities, so I calculated dQ for each phase individually. dQdU is determined in the coupled variables section, but used in the rhs section. Hopefully the example notebook + the changes in the next commit will be revealing.

pybamm/parameters/lithium_ion_parameters.py
```diff
@@ -630,6 +636,26 @@ def U(self, sto, T, lithiation=None):
...
+
+    def Q(self, sto):
+        """Capacity change as a function of stoichiometry"""
+        c_max = self.c_max
+        epsilon_s_av = self.epsilon_s_av
+        V_electrode = self.main_param.A_cc * self.domain_param.L
+        Li_max = c_max * V_electrode * epsilon_s_av
+        Q_max = Li_max * self.main_param.F / 3600
+        return Q_max * sto
+
```
mleot, line 658: @rtimms This is the function that I added to the parameters, so that it's valid for individual phases when the composite electrode model is used.

pybamm/models/submodels/interface/open_circuit_potential/plett_ocp.py
```diff
@@ -0,0 +1,157 @@
...
+            variables[f"{Domain} electrode {phase_name}equilibrium OCP [V]"] = (
+                ocp_surf_eq
+            )
+            variables[f"{Domain} electrode {phase_name}bulk equilibrium OCP [V]"] = (
+                ocp_bulk_eq
+            )
+            variables[f"{Domain} electrode {phase_name}OCP hysteresis [V]"] = H
+
+            dU = self.phase_param.U(sto_surf, T_bulk).diff(sto_surf)
+            # dQ = variables[f"{Domain} electrode {phase_name}capacity [A.h]"]
```
mleot, line 88: I tried using the variable here, but it looks like it doesn't exist for individual phases.

pybamm/models/submodels/interface/open_circuit_potential/plett_ocp.py
```diff
@@ -0,0 +1,151 @@
...
+            f"{Domain} electrode {phase_name}differential capacity [A.s.V-1]"
+        ]
+        dQdU = dQdU.orphans[0]
+        K = self.phase_param.K
+        K_x = self.phase_param.K_x
+        h = variables[f"{Domain} electrode {phase_name}hysteresis state"]
+
+        dhdt = (
+            K * (current / (Q_cell * (dQdU**K_x))) * (1 - pybamm.sign(current) * h)
+        )  #! current is backwards for a halfcell
```
mleot: The equation for this is equation 9, with the modification suggested by equation 10 

image

However I believe for a composite electrode model to be accurate, the `Q_cell` must be for the specific phase.

pybamm/models/submodels/interface/open_circuit_potential/plett_ocp.py
mleot: Succinctly, it is used in the equation below.

pybamm/models/submodels/interface/open_circuit_potential/plett_ocp.py
mleot: I believe it wasn't dQ. I believe I have since fixed it, and reverted back to the functional form of Q and taking the differential of that. 

Here is the equation:
image

docs/source/examples/index.rst
```diff
@@ -50,6 +50,7 @@ The notebooks are organised into subfolders, and can be viewed in the galleries
     notebooks/models/compare-particle-diffusion-models.ipynb
     notebooks/models/composite_particle.ipynb
     notebooks/models/coupled-degradation.ipynb
+    notebooks/models/DCHS.ipynb
```
valentinsulzer, line 53: this is an uncommon acronym, can you write out the full name?

pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -102,7 +102,7 @@ class BatteryModelOptions(pybamm.FuzzyDict):
                 reactions.
             * "open-circuit potential" : str
                 Sets the model for the open circuit potential. Can be "single"
-                (default), "current sigmoid", or "MSMR". If "MSMR" then the "particle"
+                (default), "current sigmoid", "DCHS", or "MSMR". If "MSMR" then the "particle"
```
valentinsulzer, line 105: We should avoid acronyms that are not very common. Could we call this "Wysick" instead?

pybamm/parameters/lithium_ion_parameters.py
```diff
@@ -535,6 +535,12 @@ def _set_parameters(self):
         eps_c_init_av = pybamm.xyz_average(
             self.epsilon_s * pybamm.r_average(self.c_init)
         )
+        # if self.options['open-circuit potential'] == 'Plett':
+        self.K = pybamm.Parameter(f"{pref}{Domain} particle hysteresis decay rate")
+        self.K_x = pybamm.Parameter(
+            f"{pref}{Domain} particle hysteresis switching factor"
+        )
```
valentinsulzer, line 542: this name should be more informative as well since `K` is not a standard abbreviation in battery modeling

pybamm/parameters/lithium_ion_parameters.py
```diff
@@ -630,6 +636,26 @@ def U(self, sto, T, lithiation=None):
             out.print_name = r"U_\mathrm{p}(c^\mathrm{surf}_\mathrm{s,p}, T)"
         return out
 
+    def H(self, sto):
```
valentinsulzer, line 639: this name should be more informative as well since `H` is not a standard abbreviation in battery modeling

valentinsulzer: Also shouldn't this function be calculated from the user-provided lithiation and delithiation OCPs, rather than requiring an additional user input?

valentinsulzer (changes requested):
Mostly looks good except names and the comment about the "hysteresis" function - I don't think it should be user-provided?

docs/source/examples/index.rst
mleot: Sure, I was calling it differential capacity hysteresis state. I will write this out in full for the notebook

pybamm/models/full_battery_models/base_battery_model.py
mleot: I was just thinking along the lines of MSMR, and trying to find a name that captures the physics but not nessesarily the author. But I suppose I don't have any issues with calling it Wycisk instead.

pybamm/models/submodels/interface/open_circuit_potential/wycisk_ocp.py
```diff
@@ -0,0 +1,154 @@
...
+            Q_cell = variables[f"{Domain} electrode {phase_name}phase capacity [A.h]"]
+        else:
+            Q_cell = variables[f"{Domain} electrode capacity [A.h]"]
+
+        dQdU = variables[
+            f"{Domain} electrode {phase_name}differential capacity [A.s.V-1]"
+        ]
+        dQdU = dQdU.orphans[0]
+        K = self.phase_param.hysteresis_decay
+        K_x = self.phase_param.hysteresis_switch
```
mleot, line 142: @valentinsulzer Do you want me to update the shorthand variable names within the OCP class, or are these fine to leave as is?

pybamm/parameters/lithium_ion_parameters.py
mleot: Done

pybamm/parameters/lithium_ion_parameters.py
mleot: It has since been renamed to "hysteresis".

I also updated it to calculate from user-provided OCPs. The average OCP is still required as a parameter. Is that okay?

pybamm/models/submodels/interface/open_circuit_potential/wycisk_ocp.py
```diff
@@ -5,7 +5,7 @@
 from . import BaseOpenCircuitPotential
 
 
-class DCHSOpenCircuitPotential(BaseOpenCircuitPotential):
+class WyciskOpenCircuitPotential(BaseOpenCircuitPotential):
```
valentinsulzer, line 8: This class needs a docstring explaining what the model is + giving the reference (like you do in the notebook) and it needs to be included in the appropriate `.rst` file in the `docs/source/api` folder

pybamm/models/submodels/interface/open_circuit_potential/wycisk_ocp.py
valentinsulzer: Local within the OCP class is fine. As an aside, there is an argument to be made for defining the `FunctionParameter` and `Parameter` objects inside the specific submodels when they are only relevant for a single submodel. But it's not what we do for other submodels so let's leave it like this for now

valentinsulzer (changes requested):
Thanks for addressing these! Just one remaining comment about documentation

pybamm/models/submodels/interface/open_circuit_potential/wycisk_ocp.py
mleot: Done.

mleot:
> Thanks for addressing these! Just one remaining comment about documentation

I believe I addressed the remaining changes.

One last thing: should I add a line to the changelog?

pybamm/parameters/lithium_ion_parameters.py
```diff
@@ -641,6 +649,33 @@ def U(self, sto, T, lithiation=None):
             out.print_name = r"U_\mathrm{p}(c^\mathrm{surf}_\mathrm{s,p}, T)"
         return out
 
+    def hysteresis(self, sto):
```
rtimms, line 652: Since different hysteresis models handle this differently, I wonder if this should just be calculated from the lith and delith OCPs within the submodel?

pybamm/parameters/lithium_ion_parameters.py
```diff
@@ -641,6 +649,33 @@ def U(self, sto, T, lithiation=None):
...
+        delith_ref = pybamm.FunctionParameter(
+            f"{self.phase_prefactor}{Domain} electrode delithiation OCP [V]", inputs
+        )
+        h_ref = abs(delith_ref - lith_ref) / 2
+        # h_ref = pybamm.FunctionParameter(
+        #     f"{self.phase_prefactor}{Domain} electrode OCP hysteresis [V]", inputs
+        # )
+        return h_ref
+
+    def Q(self, sto):
```
rtimms, line 670: What if the active material volume fraction changes? Maybe this is fine as the definition of capacity within the context of the hysteresis model, but again that might be an argument for it to be defined locally rather than here.

docs/source/examples/notebooks/models/differential-capacity-hysteresis-state.ipynb
```diff
@@ -0,0 +1,327 @@
...
+    "lithiation_ocp = parameters_DCHS[\"Secondary: Negative electrode lithiation OCP [V]\"]\n",
+    "delithiation_ocp = parameters_DCHS[\"Secondary: Negative electrode delithiation OCP [V]\"]\n",
+    "\n",
+    "\n",
+    "# define the new OCP function\n",
+    "def ocp_avg(sto):\n",
+    "    return (lithiation_ocp(sto) + delithiation_ocp(sto)) / 2\n",
+    "\n",
+    "\n",
+    "# define hysteresis function\n",
```
rtimms, line 120: no longer needed

docs/source/examples/notebooks/models/differential-capacity-hysteresis-state.ipynb
```diff
@@ -0,0 +1,327 @@
...
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "output_variables = [\n",
+    "    \"Terminal voltage [V]\",\n",
+    "    \"Current [A]\",\n",
+    "    \"Negative electrode secondary open-circuit potential [V]\",\n",
+    "]\n",
+    "pybamm.QuickPlot(\n",
```
rtimms, line 298: nice! this model is a great addition!

pybamm/parameters/lithium_ion_parameters.py
mleot: I believe Valentine was speaking to this item in [this](https://github.com/pybamm-team/PyBaMM/pull/3593#discussion_r1588319590) comment. Just let me know if you think that keeping it in the subclass is the best way to do this.

pybamm/parameters/lithium_ion_parameters.py
mleot: If I am not mistaken, when the active material volume fraction changes, it is captured in this definition as the `epsilon_s_av` parameter. Do you think it's worth trying to put this in the hysteresis submodel class? If so, which method would it be defined in?

valentinsulzer: The amount of active material is specified by the `active_material` submodels. Constant (default) active material takes the value of that parameter, but if loss of active material is enabled, that parameter is only the initial condition

pybamm/parameters/lithium_ion_parameters.py
valentinsulzer: Yes, let's move this into the submodel in this case

pybamm/parameters/lithium_ion_parameters.py
mleot: Understood! I will take a closer look at that.

pybamm/parameters/lithium_ion_parameters.py
mleot: Done.

pybamm/parameters/lithium_ion_parameters.py
mleot: Done. I changed the calculation to use the epsilon variable instead of parameter.

docs/source/examples/notebooks/models/differential-capacity-hysteresis-state.ipynb
mleot: Done. Removed the hysteresis function.

pybamm/models/submodels/interface/open_circuit_potential/wycisk_ocp.py
```diff
@@ -83,11 +97,34 @@ def get_coupled_variables(self, variables):
...
+            # determine dQ/dU
+            def Q(sto, epsilon_s_av):
+                """Capacity change as a function of stoichiometry and active material volume fraction"""
+                c_max = self.phase_param.c_max
+                # epsilon_s_av = self.epsilon_s_av
+                epsilon_s_av = pybamm.yz_average(epsilon_s_av)
+                V_electrode = self.phase_param.main_param.A_cc * self.domain_param.L
+                Li_max = c_max * V_electrode * epsilon_s_av
+                Q_max = Li_max * self.phase_param.main_param.F / 3600
+                return Q_max * sto
```
valentinsulzer, line 121: Sorry this review process never ends, now this is duplicate of https://github.com/pybamm-team/PyBaMM/blob/fadf094d17d0562319e4d5065a2eee1a9848b004/pybamm/models/submodels/active_material/base_active_material.py#L70 (and also it looks like you defined the `Q` function twice, is there a difference I missed or was it an accident?)

pybamm/models/submodels/interface/open_circuit_potential/wycisk_ocp.py
```diff
@@ -26,12 +26,26 @@ def get_fundamental_variables(self):
...
+
     def get_coupled_variables(self, variables):
         domain, Domain = self.domain_Domain
         phase_name = self.phase_name
         phase = self.phase
 
         if self.reaction == "lithium-ion main":
+            # define custom hysteresis function
+            # hysteresis = pybamm.FunctionParameter(f'{Domain} {phase_name}particle hysteresis [V]')
+            # sto = variables[f'{Domain} {phase_name} particle stoichiometry']
```
valentinsulzer, line 47: remove

valentinsulzer (changes requested):

pybamm/models/submodels/interface/open_circuit_potential/wycisk_ocp.py
mleot: Yes, I tried to use the other submodel, but it looks like it wasn't defined before this submodel, so that variable wasn't accessible. Do you know how to ask for variables from other submodels before they are created?

pybamm/models/submodels/interface/open_circuit_potential/wycisk_ocp.py
mleot: Done

pybamm/models/submodels/interface/open_circuit_potential/wycisk_ocp.py
mleot: I think I just got it to work, not sure what error I was running into earlier.

pybamm/models/submodels/interface/open_circuit_potential/wycisk_ocp.py
```diff
@@ -110,21 +96,28 @@ def get_coupled_variables(self, variables):
...
-                return Q_max * sto
+            # def Q(sto, epsilon_s_av):
+            #     """Capacity change as a function of stoichiometry and active material volume fraction"""
+            #     c_max = self.phase_param.c_max
+            #     # epsilon_s_av = self.epsilon_s_av
+            #     epsilon_s_av = pybamm.yz_average(epsilon_s_av)
+            #     V_electrode = self.phase_param.main_param.A_cc * self.domain_param.L
+            #     Li_max = c_max * V_electrode * epsilon_s_av
+            #     Q_max = Li_max * self.phase_param.main_param.F / 3600
+            #     return Q_max * sto
```
valentinsulzer, line 107: remove

pybamm/models/submodels/interface/open_circuit_potential/wycisk_ocp.py
```diff
@@ -110,21 +96,28 @@ def get_coupled_variables(self, variables):
...
+
+            def Q(sto):
+                """Capacity change as a function of stoichiometry"""
+                if phase_name == "":
+                    C = variables[f"{Domain} electrode capacity [A.h]"]
+                else:
+                    C = variables[
+                        f"{Domain} electrode {phase_name}phase capacity [A.h]"
+                    ]
+                return C * sto
```
valentinsulzer, line 117: this doesn't need to be in a function now

pybamm/models/submodels/interface/open_circuit_potential/wycisk_ocp.py
```diff
@@ -110,21 +96,28 @@ def get_coupled_variables(self, variables):
...
 
             dU = self.phase_param.U(sto_surf, T_bulk).diff(sto_surf)
-            epsilon_s_av = variables[
-                f"X-averaged {domain} electrode {phase_name}active material volume fraction"
-            ]
-            dQ = Q(sto_surf, epsilon_s_av).diff(sto_surf)
+            dQ = Q(sto_surf).diff(sto_surf)
```
valentinsulzer, line 120: by the way, this will just return `C` (`(a * b).diff(a) = b`)

pybamm/models/submodels/interface/open_circuit_potential/wycisk_ocp.py
valentinsulzer: and we usually use `Q` instead of `C` for capacity (it's wrong in the other function I shared)

valentinsulzer (changes requested):

pybamm/models/submodels/interface/open_circuit_potential/wycisk_ocp.py
mleot: Done, sorry forgot to clean this up.

pybamm/models/submodels/interface/open_circuit_potential/wycisk_ocp.py
mleot: Am I able to take the differential if it's not a function?

mleot: I am not sure how to do that in pybamm syntax so I might need some help there.

pybamm/models/submodels/interface/open_circuit_potential/wycisk_ocp.py
mleot: I can give it a try without the function and differentiation. I remember it not working but I could be wrong.

mleot: looks like there are no errors

pybamm/models/submodels/interface/open_circuit_potential/wycisk_ocp.py
mleot: Okay the differential is not needed.

valentinsulzer (approved):
Happy with this, let's merge!

`.diff()` operates on the symbolic expression tree, so it doesn't require a function

mleot:
Thanks for walking me through it! Hopefully the next one goes smoother!

brosaplanella:
@all-contributors please add @mleot for code and tests

DavidMStraub:
Nice addition 👍.  The heat generation due to OCP hysteresis still isn't accounted for, is it?

brosaplanella:
I don't think so. Do you know any references on how this should be accounted for?

DavidMStraub:
Only using terminal voltage & energy balance like in https://doi.org/10.1038/s41560-019-0410-6 ... @rtimms might be able to answer better 🙂

Issue #3530, opened 2023-11-16
Allow custom experiment steps


tinosulzer:
The logic for specifying how experiment steps work is currently in the `Simulation` class. We should move this to the `Step` classes, so that users can specify any custom behavior they want for which variable controls the step, and which variable(s) are used in the termination. This will have the added benefit of reducing some of the complexity of the simulation class.

This involves replacing any if statement that checks `op.type` with a call to a function in the step class, e.g. https://github.com/pybamm-team/PyBaMM/blob/develop/pybamm/simulation.py#L222

tinosulzer:
Adding custom events should be really easy and could be done first separately, we just need to implement a `Step.events()` function that returns the appropriate event(s) https://github.com/pybamm-team/PyBaMM/blob/7c15368aa80d3ed3aa08ae3e0c41682363e06dd3/pybamm/simulation.py#L279-L284

We could allow the `termination` argument of `Step` to take in this function directly

tinosulzer:
For the custom steps, there will be two separate cases:
1. Current is specified explicitly, either as a constant or as a function of other variables, user supplies the explicit expression
2. Current is an algebraic variable, user supplies the algebraic expression for the current variable

Current and C-rate falls into 1, everything else falls into 2.

We will need two different `CustomStep` classes (`e.g. `CustomStepExplicit`, `CustomStepAlgebraic`) with examples for each one.

PR #3596, opened 2023-12-06
Issue 3530 custom termination

tinosulzer:
# Description

Adds ability to create custom termination events, using any variable, for an experiment step

First part of #3530 (second part is using custom steps, not just terminations)

## Type of change

# Key checklist:

## Further checks:


rtimms (approved):
Thanks, looks good! 

Not for this issue, but the mixed `step`/string definitions of experiments is a little weird. I think users like the string experiments and dropping them would be a mistake.

tinosulzer:
> Not for this issue, but the mixed step/string definitions of experiments is a little weird. I think users like the string experiments and dropping them would be a mistake.

Yeah there are good reasons to keep both, what are you suggesting?

rtimms:
> Yeah there are good reasons to keep both, what are you suggesting?

Nothing for now, but I'm just conscious it may be confusing for new users to see both ways of doing things within the same experiment definition.

PR #3627, opened 2023-12-17
Implemented `non_linear_time`, an array of time as an input to `step` function

### LINKED ISSUE #3524: Pass array of times to `step`
When you call `solver.step` you pass in `dt`, the time time increment you want to step forward in. You can also pass `npts` to get the solution back at a linearly spaced array of times between the initial time, `t0`, and `t0+dt`. It would be good to instead be able to pass an array of times, `t_eval`, like you pass to solve, so you can get step solutions returned at non-linearly spaced times.

The motivation for this is that right now you can't pass a `starting_solution` to `Simualtion.solve` if you aren't using an experiment, so if you want to chain solutions together you have to call `Simulation.step` instead. 

An alternative is that we allow starting solutions without experiments in `Simulation.solve`.

The whole logic for `step` and `solve` is pretty clunky both in `Simulation` and `BaseSolver`, but that's a separate issue...

prady0t:
# Description

Fixes #3524 

Now  an array of time can be passed to `step` function which will return solutions at that step times.
Example:

```py
step_solver.step(step_solution, model, non_linear_time=[0, 1000, 1500, 1700, 2000])
```
Output:

image

## Type of change

# Key checklist:

## Further checks:


arjxn-py (commented):
Thanks @prady0t, looks good. Happy to see almost all tests passing, although to fix coverage you might want to modify [test_base_model](https://github.com/pybamm-team/PyBaMM/blob/develop/tests/unit/test_models/test_base_model.py) relative to your changes. Requesting @rtimms review for further refinements.

pybamm/solvers/base_solver.py
```diff
@@ -1110,7 +1110,8 @@ def step(
         self,
         old_solution,
         model,
-        dt,
+        dt=None,
```
rtimms, line 1113: Keep this as a required argument

pybamm/solvers/base_solver.py
```diff
@@ -1162,28 +1166,72 @@ def step(
...
+            step_start_offset = pybamm.settings.step_start_offset
+            if dt <= step_start_offset:
+                raise pybamm.SolverError(
+                    f"Step time must be at least {pybamm.TimerTime(step_start_offset)}"
+                )
 
-        t_start = old_solution.t[-1]
-        t_end = t_start + dt
-        # Calculate t_eval
-        t_eval = np.linspace(t_start, t_end, npts)
```
rtimms, line 1176: if `t_eval` is the optional argument then this can be `t_eval = t_eval or np.linspace(t_start, t_end, 2)` or equivalent

rtimms (changes requested):
Thanks for taking this on. I don't think it is necessary to have both `npts` and `non_linear_time`. Instead, I think we should deprecate `npts`, require users to pass `dt` and make `t_eval` an optional argument. 

In the deprecation warning we can explain that you can calculate `t_eval = np.linspace(t_start, t_end, npts)` and pass that instead.

That way, you don't need to have repeated logic depending on which optional argument is passed, and the API is a bit cleaner - better to have only one way of doing things. It would also mimic [solve_ivp's](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.solve_ivp.html) API where you pass `t_span` (the start and end time) and optionally `t_eval`.

TomTranter:
> Thanks for taking this on. I don't think it is necessary to have both `npts` and `non_linear_time`. Instead, I think we should deprecate `npts`, require users to pass `dt` and make `t_eval` an optional argument.
> 
> In the deprecation warning we can explain that you can calculate `t_eval = np.linspace(t_start, t_end, npts)` and pass that instead.
> 
> That way, you don't need to have repeated logic depending on which optional argument is passed, and the API is a bit cleaner - better to have only one way of doing things. It would also mimic [solve_ivp's](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.solve_ivp.html) API where you pass `t_span` (the start and end time) and optionally `t_eval`.
> 
> tagging @TomTranter as I know you use step a fair bit in other projects

@rtimms suggestions sound sensible to me

prady0t:
> Thanks for taking this on. I don't think it is necessary to have both `npts` and `non_linear_time`. Instead, I think we should deprecate `npts`, require users to pass `dt` and make `t_eval` an optional argument.
> 
> In the deprecation warning we can explain that you can calculate `t_eval = np.linspace(t_start, t_end, npts)` and pass that instead.
> 
> That way, you don't need to have repeated logic depending on which optional argument is passed, and the API is a bit cleaner - better to have only one way of doing things. It would also mimic [solve_ivp's](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.solve_ivp.html) API where you pass `t_span` (the start and end time) and optionally `t_eval`.
> 
> tagging @TomTranter as I know you use step a fair bit in other projects

Yes that makes much more sense! Making changes now along with code coverage.

prady0t:
Also should I change name of variable `non_linear_time` to something more accurate or is it ok? @rtimms

rtimms:
You should always have to pass `dt`, and `t_eval` should be the optional argument. If `t_eval` is `None` then set `t_eval = np.linspace(0, dt, 2)` (this is the same as the default behaviour before with `npts=2`). If `t_eval` is not `None` then check that `t_eval[0] = 0` and `t_eval[-1]=dt`.

tinosulzer:
Why not just require `t_eval` (and no `dt`) so that the API matches `solve` ?

rtimms:
> Why not just require t_eval (and no dt) so that the API matches solve ?

Happy with this too, but `step` was kind of intended to be "step this model forward in time this amount", so having `dt` makes sense in that context.

prady0t:
Now the output of 
```py
while time < end_time:
    step_solution = step_solver.step(step_solution, model, dt=500, non_linear_time=[0, 50, 100, 200, 500])
    time += dt
```
Looks like:

image

prady0t:
Really sorry for all the extra lines and spaces that were added (my editor really wants to stick to the PEP8 style). Should I remove them?

prady0t:
Just noticed the failing test cases 🙃 . Working in it

prady0t:
Should we also remove `npts` from [here](https://github.com/pybamm-team/PyBaMM/blob/ee64eafc7f945fb6684ce22a60169f97f5a0f235/pybamm/simulation.py#L965) ?

tinosulzer:
Yes

prady0t:
I don't understand why [this](https://github.com/pybamm-team/PyBaMM/actions/runs/7379963719/job/20078006973?pr=3627) test case is failing. Can someone please help me out?

prady0t:
I've also added tests.

pybamm/simulation.py
```diff
@@ -954,15 +954,21 @@ def run_padding_rest(self, kwargs, rest_time, step_solution):
...
         )
 
         return step_solution_with_rest
 
     def step(
-        self, dt, solver=None, npts=2, save=True, starting_solution=None, **kwargs
+        self,
+        dt,
+        solver=None,
+        non_linear_time=None,
```
rtimms, line 968: can we rename to `t_eval`?

pybamm/simulation.py
```diff
@@ -974,9 +980,9 @@ def step(
             The timestep over which to step the solution
         solver : :class:`pybamm.BaseSolver`
             The solver to use to solve the model.
-        npts : int, optional
-            The number of points at which the solution will be returned during
-            the step dt. Default is 2 (returns the solution at t0 and t0 + dt).
+        non_linear_time : list or numpy.ndarray, optional
+            An array of time (in ascending order and not necessarily linearly distributed)
```
rtimms, line 984:
```suggestion
            An array of times at which to return the solution during the step (Note: t_eval is the time measured from the start of the step, so should start at 0 and end at dt). By default, the solution is returned at t0 and t0 + dt.
```

pybamm/solvers/base_solver.py
```diff
@@ -1120,9 +1120,9 @@ def step(
             initial_conditions
         dt : numeric type
             The timestep (in seconds) over which to step the solution
-        npts : int, optional
-            The number of points at which the solution will be returned during
-            the step dt. default is 2 (returns the solution at t0 and t0 + dt).
+        non_linear_time : list or numpy.ndarray, optional
```
rtimms, line 1123: update docstring as above

pybamm/solvers/base_solver.py
```diff
@@ -1160,11 +1160,34 @@ def step(
             raise pybamm.SolverError(
                 f"Step time must be at least {pybamm.TimerTime(step_start_offset)}"
             )
-
         t_start = old_solution.t[-1]
-        t_end = t_start + dt
-        # Calculate t_eval
-        t_eval = np.linspace(t_start, t_end, npts)
+        t_end = t_start
```
rtimms, line 1164: should `t_end` be `t_start+dt`?

pybamm/solvers/base_solver.py
```diff
@@ -1160,11 +1160,34 @@ def step(
...
+        t_eval = np.array([t_start])
+
+        if non_linear_time is not None:
+            # Checking if non_linear_time lies within range
+            if non_linear_time[0] != 0 or non_linear_time[-1] != dt:
+                raise pybamm.SolverError(
+                    "Elements inside array non_linear_time must lie in the closed interval 0 to dt"
+                )
+
+            for i in range(len(non_linear_time) - 1):
```
rtimms, line 1174: cany use `any` and remove the loop, then just make `t_eval = non_linear_time + t_start`

pybamm/solvers/base_solver.py
rtimms: in fact, you don't need to define this here?

pybamm/solvers/base_solver.py
```diff
@@ -1160,11 +1160,34 @@ def step(
...
+                        f"Time difference between elements in non_linear_time must be at least "
+                        f"{pybamm.TimerTime(step_start_offset)} "
+                    )
+
+                t_end = t_end + time_diff
+                # Append to t_eval
+                t_eval = np.append(t_eval, t_end)
+
+        else:
+            t_start = old_solution.t[-1]
```
rtimms, line 1187: already defined outside the if statement

rtimms (dismissed):

pybamm/solvers/base_solver.py
prady0t: That's what I did before but removing the line throws this error :
```
  t_end = t_end + time_diff
UnboundLocalError: local variable 't_end' referenced before assignment
```
Maybe `t_end` needs to be initialized before.

pybamm/solvers/base_solver.py
prady0t: So should I not check if difference between elements is larger than `pybamm.TimerTime(step_start_offset)`?

pybamm/solvers/base_solver.py
prady0t: @rtimms what should be done here?

pybamm/solvers/base_solver.py
prady0t: I get it now. But wouldn't we still need a `time_diff` variable to store time difference so that we can increment `t_end` by that much?

pybamm/solvers/base_solver.py
arjxn-py: Maybe you can try doing `t_end = t_start + dt + time_diff` instead of `t_end = t_end + time_diff` in [L1182](https://github.com/pybamm-team/PyBaMM/pull/3627/files?diff=unified&w=0#diff-630d9645dc8c18de32e60b33d118b8dc8c243a8641df8a65fb4ad971b96f6064R1182)

pybamm/simulation.py
```diff
@@ -967,15 +967,21 @@ def run_padding_rest(self, kwargs, rest_time, step_solution):
             step_solution,
             model,
             rest_time,
-            npts=npts,
```
rtimms, line 970: keep `npts` and raise a deprecation warning, then convert to the correct `non_linear_time`

pybamm/solvers/base_solver.py
rtimms: you don't need to do this check. we just need `dt` to be bigger than `step_start_offset`

pybamm/solvers/base_solver.py
```diff
@@ -1160,11 +1160,34 @@ def step(
             raise pybamm.SolverError(
                 f"Step time must be at least {pybamm.TimerTime(step_start_offset)}"
             )
-
         t_start = old_solution.t[-1]
```
rtimms, line 1163: originally we had 

```python3
        # Make sure dt is greater than the offset
        step_start_offset = pybamm.settings.step_start_offset
        if dt <= step_start_offset:
            raise pybamm.SolverError(
                f"Step time must be at least {pybamm.TimerTime(step_start_offset)}"
            )

        t_start = old_solution.t[-1]
        t_end = t_start + dt
        # Calculate t_eval
        t_eval = np.linspace(t_start, t_end, npts)
```

this can be replaced with

```python3
        # Make sure dt is greater than the offset
        step_start_offset = pybamm.settings.step_start_offset
        if dt <= step_start_offset:
            raise pybamm.SolverError(
                f"Step time must be at least {pybamm.TimerTime(step_start_offset)}"
            )

        if non_linear_time is not None:
            # Checking if non_linear_time lies within range
            if non_linear_time[0] != 0 or non_linear_time[-1] != dt:
                raise pybamm.SolverError(
                    "Elements inside array non_linear_time must lie in the closed interval 0 to dt"
                )
        else:
            non_linear_time = np.array([0, dt])

        t_start = old_solution.t[-1]
        t_eval = t_start + non_linear_time
```

and i think achieves what you are trying to do

pybamm/solvers/base_solver.py
rtimms: see above comment

pybamm/solvers/base_solver.py
rtimms: see suggestion above

pybamm/simulation.py
prady0t: should I do this everywhere where `npts=npts` has been converted to `t_eval=np.linspace(0, npts)` and raise deprecation warning in `base_solver.step`?

rtimms: sounds good!

examples/scripts/SPMe_step.py
```diff
@@ -32,12 +32,13 @@
 
 # step model
 dt = 500
+t_eval = np.array([0, 50, 100, 200, 500])
```
rtimms, line 35: can you add a comment that t_eval are the times in the interval [0, dt]?

rtimms:
thanks @prady0t looks great! can you add a changelog entry then this is good to go

examples/scripts/SPMe_step.py
rtimms: and that `dt` is the size of the step (I think it's pretty clear but let's be explicit)

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Renamed "electrode diffusivity" to "particle diffusivity" as a non-breaking change with a deprecation warning ([#3624](https://github.com/pybamm-team/PyBaMM/pull/3624))
 - Add support for BPX version 0.4.0 which allows for blended electrodes and user-defined parameters in BPX([#3414](https://github.com/pybamm-team/PyBaMM/pull/3414))
 - Added the ability to specify a custom solver tolerance in `get_initial_stoichiometries` and related functions ([#3714](https://github.com/pybamm-team/PyBaMM/pull/3714))
+- Modified `step` function to take an array of time `t_eval` as an argument and deprecated use of `npts`. ([#3627](https://github.com/pybamm-team/PyBaMM/pull/3627))
```
agriyakhetarpal, line 9: Small comment, could you move this entry up and sort the entries in decreasing order of PR number?

prady0t:
@agriyakhetarpal why is [this](https://github.com/pybamm-team/PyBaMM/actions/runs/8003981310/job/21860421881?pr=3627) CI job failing?

kratman:
@prady0t that check has been failing for a while. If you did not add any links in your PR then it is probably unrelated. I did not check if you made any URL changes, but I know it was failing before due to stuff in the all-contributors section.

The check looks to see if it can access the URLs found in the codebase, and fails if any of them are unreachable. The idea is that if we put a link to a documentation page, it warns us if the URL stops working.

kratman:
@rtimms Is anything left for your review of this PR?

kratman:
@prady0t I will merge this when CI passes. I think everything is done

kratman (approved):

PR #3642, opened 2023-12-19
Remove initial conditions

### LINKED ISSUE #3629: `initial_conditions` passed to `solve` are ignored.
You can pass a `Symbol` of `initial_conditions` to `solve` in `BaseSolver`, but they never get used - it always uses the initial conditions from the model. My vote is we just remove this, since there are other ways to set initial conditions like updating parameter values directly, or `set_initial_conditions_from` or using `solver.step` or `starting_solution` in the simulation. I don't think there's a strong case for creating a new initial conditions symbol and passing that to the solver.

kratman:
# Description

Removing initial conditions from the solver.

Fixes #3629

## Type of change

Minor fix for an unused variable.

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/solvers/base_solver.py
+++ b/pybamm/solvers/base_solver.py
@@ -698,7 +698,6 @@ class BaseSolver:
         model,
         t_eval=None,
         inputs=None,
-        initial_conditions=None,
         nproc=None,
         calculate_sensitivities=False,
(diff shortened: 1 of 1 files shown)


tinosulzer (approved):
Looks good to me, thanks

PR #3650, opened 2023-12-21
Improved error handling and added type hinting in Experiment

### LINKED ISSUE #3623: Better error or handling when passing a str to Experiment
`Experiment` expects `operating_conditions` to be a list of strings. If you just pass a string, e.g. `pybamm.Experiment("Discharge at 1C until 2.5V")` you get the error `ValueError: Operating conditions must contain keyword 'for' or 'until'.`. 

We should add a check to see if `operating_conditions` is list of strings or a str. If it's a str we should convert to a list, if not we should raise an error (see e.g. `check_input_validity` for `QuickPlot` [here](https://github.com/pybamm-team/PyBaMM/blob/903323efca7fef8f630924d114ce70dcf8a92363/pybamm/plotting/quick_plot.py#L250)).

AbhishekChaudharii:
# Description

1. Improved error handling by raising TypeError when a str is passed to Experiment instead of list of strings
2. Added type hinting for better readability.

Fixes #3623

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/experiment/experiment.py
+++ b/pybamm/experiment/experiment.py
@@ -52,7 +52,7 @@ class Experiment:
         if not (isinstance(operating_conditions, list)):
             raise TypeError("operating_conditions must be list of strings. For example:"
                             f"\n\n [{operating_conditions}]")
-        
+
         if cccv_handling is not None:
             raise ValueError(
                 "cccv_handling has been deprecated, use "
--- a/pybamm/experiment/experiment.py
+++ b/pybamm/experiment/experiment.py
@@ -2,6 +2,7 @@
 # Experiment class
 #
 
+from __future__ import annotations
 import pybamm
 from pybamm.step._steps_util import (
     _convert_time_to_seconds,
@@ -42,16 +43,13 @@ class Experiment:
 
     def __init__(
         self,
-        operating_conditions: list,
+        operating_conditions: list[str],
         period: str = "1 minute",
-        temperature: float = None,
-        termination: list = None,
+        temperature: float | None = None,
+        termination: list[str] | None = None,
         drive_cycles=None,
         cccv_handling=None,
     ):
-        if not (isinstance(operating_conditions, list)):
-            raise TypeError("operating_conditions must be list of strings. For example:"
-                            f"\n\n [{operating_conditions}]")
 
         if cccv_handling is not None:
             raise ValueError(
--- a/pybamm/experiment/experiment.py
+++ b/pybamm/experiment/experiment.py
@@ -42,13 +42,17 @@ class Experiment:
 
     def __init__(
         self,
-        operating_conditions,
-        period="1 minute",
-        temperature=None,
-        termination=None,
+        operating_conditions: list,
+        period: str = "1 minute",
+        temperature: float = None,
+        termination: list = None,
         drive_cycles=None,
         cccv_handling=None,
     ):
+        if not (isinstance(operating_conditions, list)):
+            raise TypeError("operating_conditions must be list of strings. For example:"
+                            f"\n\n [{operating_conditions}]")
+        
         if cccv_handling is not None:
             raise ValueError(
                 "cccv_handling has been deprecated, use "
--- a/pybamm/experiment/experiment.py
+++ b/pybamm/experiment/experiment.py
@@ -22,8 +22,8 @@ class Experiment:
 
     Parameters
     ----------
-    operating_conditions : list
-        List of operating conditions
+    operating_conditions : list[str]
+        List of strings representing the operating conditions.
     period : string, optional
         Period (1/frequency) at which to record outputs. Default is 1 minute. Can be
         overwritten by individual operating conditions.
@@ -31,8 +31,8 @@ class Experiment:
         The ambient air temperature in degrees Celsius at which to run the experiment.
         Default is None whereby the ambient temperature is taken from the parameter set.
         This value is overwritten if the temperature is specified in a step.
-    termination : list, optional
-        List of conditions under which to terminate the experiment. Default is None.
+    termination : list[str], optional
+        List of strings representing the conditions to terminate the experiment. Default is None.
         This is different from the termination for individual steps. Termination for
         individual steps is specified in the step itself, and the simulation moves to
         the next step when the termination condition is met


pybamm/experiment/experiment.py
```diff
@@ -42,13 +42,17 @@ class Experiment:
 
     def __init__(
         self,
-        operating_conditions,
-        period="1 minute",
-        temperature=None,
-        termination=None,
+        operating_conditions: list,
```
agriyakhetarpal, line 45: Could you add another type hint for `operating_conditions` to check if it is a string as requested in https://github.com/pybamm-team/PyBaMM/issues/3623#issue-2043497552? Also, this attribute should have the type hint `list[str]` to indicate that that this accepts a list of strings. Note that using the generic `list` type is for Python 3.9+ only and we support Python 3.8 as well (which has `List`, rather), so we should add

```python
from __future__ import annotations
```

to the top of this file.

pybamm/experiment/experiment.py
```diff
@@ -42,13 +42,17 @@ class Experiment:
...
+        period: str = "1 minute",
+        temperature: float = None,
+        termination: list = None,
         drive_cycles=None,
         cccv_handling=None,
     ):
+        if not (isinstance(operating_conditions, list)):
+            raise TypeError("operating_conditions must be list of strings. For example:"
+                            f"\n\n [{operating_conditions}]")
+
```
agriyakhetarpal, line 55: Could you convert to a string to a list of strings here, based on https://github.com/pybamm-team/PyBaMM/issues/3623#issue-2043497552 again – and raise an error on any other object? Also, you would need to improve the test coverage by adding a test for the condition you added

agriyakhetarpal (dismissed):
Hi @AbhishekChaudharii, thanks for doing this. There are some comments below that you can address

pybamm/experiment/experiment.py
kratman: I believe list[str] is all that is needed for this annotation. I can review and check it more carefully later

pybamm/experiment/experiment.py
kratman: Type checks should handle it already. It should just throw an error instead of trying to recover from bad inputs

pybamm/experiment/experiment.py
agriyakhetarpal: If `str` works with `list[str]`, then yes (but I think it should be `str | list[str]` though and `Union` for older Python versions). I am not well-versed about type hints tbh

Saransh-cpp: A separate `Union` type hint is not required. `from __future__ import annotations` will do the work.

kratman: I originally suggested just type hints because it is better to have a single interface (list[str]) then try to support multiple interfaces. It makes the code simpler and cuts down on bugs and mistakes. Once type hints are in place on the function, it will show mistakes in the IDE.

pybamm/experiment/experiment.py
Saransh-cpp: Type checks are basically just hints in Python. Not everyone has them enabled in their IDEs.

You could check for a list of strings using something like -
```py
all(isinstance(cond, operating_conditions) for cond in operating_conditions)
```

kratman: We could return the helpful error, but we could equally just have it well documented. We probably don't need to handle every possible type of error.

pybamm/experiment/experiment.py
```diff
@@ -42,13 +42,17 @@ class Experiment:
...
     def __init__(
         self,
-        operating_conditions,
-        period="1 minute",
-        temperature=None,
-        termination=None,
+        operating_conditions: list,
+        period: str = "1 minute",
+        temperature: float = None,
+        termination: list = None,
```
Saransh-cpp, line 48: A clever way to account for the `Optional` type:
```suggestion
        temperature: float | None = None,
        termination: list | None = None,
```

pybamm/experiment/experiment.py
Saransh-cpp: Agreed

pybamm/experiment/experiment.py
Saransh-cpp: Also, the `list` type should be more specific - `list[...]`

pybamm/experiment/experiment.py
kratman, line 45:
```suggestion
        operating_conditions: list[str],
```

pybamm/experiment/experiment.py
kratman, line 55: This is not needed. We document the type being list[str]. We don't check the types of the other arguments, so we should not check this one either.
```suggestion
```

kratman: Also I think we should probably work to remove the blocks like
```python
if cccv_handling is not None:
            raise ValueError(
                "cccv_handling has been deprecated, use "
                "`pybamm.step.cccv_ode(current, voltage)` instead to produce the "
                "same behavior as the old `cccv_handling='ode'`"
            )
```
in a future change. If we only check them to warn that they are deprecated, then we should probably clean it up.

kratman (changes requested):

pybamm/experiment/experiment.py
kratman: Additionally if we remove the error handling it is less unit tests needed. For proper coverage you need to cover all error handling. So if we reduce it we need less tests.

kratman (approved):
As long as the tests pass, I am happy with this

agriyakhetarpal (approved):
Thanks @AbhishekChaudharii, I agree with @kratman about reducing the number of checks. In addition, I would suggest updating the docstring for the class so that it is clearer for users and is in line with the type hint added (it should mention that `operating_conditions` is a list of strings, not just a list)

AbhishekChaudharii:
Hi @agriyakhetarpal I have updated the docstrings. Do you have any other suggestions?

agriyakhetarpal:
Looks good to me, no other suggestions!

rtimms (approved):
thanks @AbhishekChaudharii , could you add a changelog entry then this is good to go

rtimms:
@all-contributors please add @AbhishekChaudharii for code and docs

arjxn-py (approved):
Thanks @AbhishekChaudharii, looks good.

PR #3655, opened 2023-12-23
Doctests do not remove built files if they fail #3389

### LINKED ISSUE #3389: Doctests do not remove built files if they fail
In `run-tests.py`, if the doctests fail locally for some reason, they do not delete the built files and cause problems upon invoking the doctests again. This is because `nbsphinx` duplicates some of the doctrees upon rebuilds (it is not meant to work with the doctests). The command should be fixed in order to delete the built files upon invoking the doctests, or just the .doctrees folder inside them so that the error logs can be retained.

shubhambhar007:
# Description

As pointed by @agriyakhetarpal , in case of failure of doctests, built files (inside the docs/build/ folder) are not deleted.
We also have to ensure deletion in case of successful doctests. I added a finally block for the same and put the subprocess call in a try block.

Fixes #3389

## Type of change

- [❌ ] New feature (non-breaking change which adds functionality)
- [❌] Optimization (back-end change that speeds up the code)
- [✅] Bug fix (non-breaking change which fixes an issue)

# Key checklist:

- [✅ ] No style issues: `$ pre-commit run` (or `$ nox -s pre-commit`) (see [CONTRIBUTING.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CONTRIBUTING.md#installing-and-using-pre-commit) for how to set this up to run automatically when committing locally, in just two lines of code)
- [✅] All tests pass: `$ python run-tests.py --all` (or `$ nox -s tests`)
- [✅] The documentation builds: `$ python run-tests.py --doctest` (or `$ nox -s doctests`)

## Further checks:

- [✅] Code is commented, particularly in hard-to-understand areas
- [NA ] Tests added that prove fix is effective or that feature works


shubhambhar007:
image
image

run-tests.py
```diff
@@ -64,36 +64,31 @@ def run_doc_tests():
...
-        print("")
-        sys.exit(1)
-    if ret != 0:
-        print("FAILED")
-        sys.exit(ret)
-    # delete the entire docs/source/build folder + files since it currently
-    # causes problems with nbsphinx in further docs or doctest builds
-    print("Deleting built files.")
-    shutil.rmtree("docs/build")
+            shutil.rmtree("docs/build")
```
agriyakhetarpal, line 89:
```suggestion
            shutil.rmtree("docs/build/html/.doctrees/")
```

This way we get to keep `docs/build/output.txt` if someone wants to inspect the logs, and the doctests don't fail on subsequent runs either

agriyakhetarpal (approved):
Thanks for the fix, @shubhambhar007! Looks good, one small suggestion.

shubhambhar007:
@agriyakhetarpal i think i force pushed on top of your commit 😅
Please review once .

agriyakhetarpal:
I think it was the same suggestion that I updated your branch with, we should be fine – thanks!

shubhambhar007:
> I think it was the same suggestion that I updated your branch with, we should be fine – thanks! Looks like the missing coverage is unrelated.

Yes ,i thought so too, the subrpocess call is being fed arguments by us itself and is not subject to any outside code injection, should be fine in my opinion :D .
Thanks!

arjxn-py (approved):
Looks Good, thanks @shubhambhar007

arjxn-py:
Example notebook tests failing doesn't look like a cause of this PR, should we re-run jobs?

agriyakhetarpal:
> Example notebook tests failing doesn't look like a cause of this PR, should we re-run jobs?

I did so just now, it seems that #3492 broke something even though it shouldn't have

Issue #3489, opened 2023-10-31
Check adherence with `sp-repo-review` guidelines


agriyakhetarpal:
There are a lot of interesting things available in the Scientific Python ecosystem, perhaps we should also try using [`sp-repo-review`](https://learn.scientific-python.org/development/guides/repo-review/)?

_Originally posted by @agriyakhetarpal in https://github.com/pybamm-team/PyBaMM/issues/3390#issuecomment-1786984884_

agriyakhetarpal:
<details><summary>Output from running sp-repo-review locally</summary>
<p>

```
General:

 • Detected build backend: MISSING                                                                                                      

├── PY001 Has a pyproject.toml ❌
│   All projects should have a pyproject.toml file to support a modern build system and support wheel installs properly.                
├── PY002 Has a README.(md|rst) file ✅
├── PY003 Has a LICENSE* file ✅
├── PY004 Has docs folder ✅
├── PY005 Has tests folder ✅
├── PY006 Has pre-commit config ✅
└── PY007 Supports an easy task runner (nox or tox) ✅

PyProject:
├── PP002 Has a proper build-system table [skipped]
├── PP003 Does not list wheel as a build-dep [skipped]
├── PP301 Has pytest in pyproject [skipped]
├── PP302 Sets a minimum pytest to at least 6 [skipped]
├── PP303 Sets the test paths [skipped]
├── PP304 Sets the log level in pytest [skipped]
├── PP305 Specifies xfail_strict [skipped]
├── PP306 Specifies strict config [skipped]
├── PP307 Specifies strict markers [skipped]
├── PP308 Specifies useful pytest summary [skipped]
└── PP309 Filter warnings specified [skipped]

GitHub Actions:
├── GH100 Has GitHub Actions config ✅
├── GH101 Has nice names ✅
├── GH102 Auto-cancel on repeated PRs ✅
├── GH103 At least one workflow with manual dispatch trigger ✅
├── GH200 Maintained by Dependabot ✅
├── GH210 Maintains the GitHub action versions with Dependabot ✅
└── GH211 Do not pin core actions as major versions ✅

Pre-commit:
├── PC100 Has pre-commit-hooks ✅
├── PC110 Uses black or ruff-format ❌
│   Must have https://github.com/psf/black-pre-commit-mirror or https://github.com/astral-sh/ruff-pre-commit + id: ruff-format in       
│   .pre-commit-config.yaml                                                                                                             
[...]

PR #3656, opened 2023-12-23
Migrate to ruff format

### LINKED ISSUE #3477: Switch to `ruff format`

Saransh-cpp:
# Description

Sorry for creating this huge PR just before the holidays 😬 

Fixes #3477 
Also addresses #3489

## Type of change

# Key checklist:

## Further checks:


agriyakhetarpal:
Should we close https://github.com/pybamm-team/PyBaMM/pull/3492?

.pre-commit-config.yaml
```diff
@@ -9,12 +9,14 @@ repos:
...
         types_or: [python, pyi, jupyter]
+      - id: ruff-format
+        types_or: [python, pyi, jupyter]
 
   - repo: https://github.com/adamchainz/blacken-docs
     rev: "1.16.0"
     hooks:
        - id: blacken-docs
-         additional_dependencies: [black==22.12.0]
+         additional_dependencies: [black==23.*]
```
agriyakhetarpal, line 19: I recently got to know that `ruff` is now capable of formatting code blocks in documentation files (Markdown, reST, etc.) as well, should we consider removing `blacken-docs`? Feel free to convert this comment to an issue if yes

agriyakhetarpal (approved):
Thanks, @Saransh-cpp! Looks good and it is nice to get this tool working for notebooks. Is there a way to reduce the line length in the configuration just for the notebooks? See [this notebook in the PR build](https://pybamm--3656.org.readthedocs.build/en/3656/source/examples/notebooks/creating_models/2-a-pde-model.html#Defining-a-geometry-and-mesh) for example – it would be great if we can set the line length to, say, 60 for notebooks, while keeping it at 88 or so for Python files. It reduces the amount of side scrolling required when browsing through the code cells for accessibility reasons.

In hindsight, we should have kept #3492 open since the contributor was active for a while and it had been just three weeks since they opened it. Let's be more patient next time we encounter a delay and wait for at least two months, given that this was not a high priority issue to fix.

.pre-commit-config.yaml
Saransh-cpp: I don't think ruff can do that yet, see - https://github.com/astral-sh/ruff/issues/8237 and https://github.com/astral-sh/ruff/issues/3792

Saransh-cpp:
> it would be great if we can set the line length to, say, 60 for notebooks, while keeping it at 88 or so for Python files

I went through ruff's config and I don't think they support such flexibility at the moment.

agriyakhetarpal:
How about temporarily setting it to 60 and linting the notebooks while ignoring the rest, then setting it back to the default and linting the rest ignoring the notebooks?

Saransh-cpp:
> How about temporarily setting it to 60 and linting the notebooks while ignoring the rest, then setting it back to the default and linting the rest ignoring the notebooks?

ruff will automatically expand the lines with length = 60 to length = 88 whenever we run it again (that is, in the CI), it goes both ways - shrink as well as expand line lengths

agriyakhetarpal:
Ah I didn't know that, I will open up a feature request about this upstream and hope they find it reasonable

PR #3569, opened 2023-11-27
i3558 improve wheel repair + add minimal IDAKLU test for built wheels

agriyakhetarpal:
# Description

This PR sets the runtime search path for the `idaklu` target via CMake to the location of the CasADi libraries so that they can be picked up by `auditwheel` and `delocate` at the time of repairing the wheel. This path was strangely empty in prior versions and therefore a script was used to fix the RPATH for macOS. A small test has been added to each wheel job to check the import of the IDAKLU extension module, which, upon being failed will mark the workflow as such too.

> [!TIP]
> I have triggered a workflow run to build the wheels under these changes, it can be accessed from here: https://github.com/agriyakhetarpal/PyBaMM/actions/runs/7009012889

Closes #3558

## Type of change

Please add a line in the relevant section of [CHANGELOG.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CHANGELOG.md) to document the change (include PR #) - note reverse order of PR #s. If necessary, also add to the list of breaking changes.

- [x] Bug fix (non-breaking change which fixes an issue)

# Key checklist:

- [x] No style issues: `$ pre-commit run` (or `$ nox -s pre-commit`) (see [CONTRIBUTING.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CONTRIBUTING.md#installing-and-using-pre-commit) for how to set this up to run automatically when committing locally, in just two lines of code)
- [x] All tests pass: `$ python run-tests.py --all` (or `$ nox -s tests`)
- [x] The documentation builds: `$ python run-tests.py --doctest` (or `$ nox -s doctests`)

You can run integration tests, unit tests, and doctests together at once, using `$ python run-tests.py --quick` (or `$ nox -s quick`).

## Further checks:

- [X] Code is commented, particularly in hard-to-understand areas
- [X] Tests added that prove fix is effective or that feature works

Issue #3395, opened 2023-10-02
Jax solver mode: only include output variables


BradyPlanden:
### Description

This continues #2796 for the JaxSolver. 

### Motivation

For parameter estimation, it is (computationally) benefitical to store only the required output variables. 

This issue also aims to provide functionality for jit-compiled parameter estimation or optimisation workflows. At the moment this isn't possible since the creation of the `ProcessedVariable` dict is completed with Casadi. Further discussion here: https://github.com/pybop-team/PyBOP/pull/25

agriyakhetarpal:
Hi, @BradyPlanden – you had mentioned that #3658 should have closed this issue, but I think we forgot to do so. Should we close this now?

BradyPlanden:
Yep, marking closed.

PR #2844, opened 2023-03-31
Fix random seed on tests

jsbrittain:
# Description

Some tests in the unit and integration suites call functions that make use of random variables, or include random variables in their calls (specifically numpy.random). This can lead to tests sometimes failing due to reasons that may be unrelated to recent code changes, as outlined (and demonstrated) in issue #2833.

Fixes #2833 

## Type of change

Fixing the numpy random seed within the test routines produces consistent behavior. However, since stochastic elements are present in the PyBaMM codebase in relation to both the Casadi and Idaklu solvers, extensive dependencies on these functions within the test suite are likely to be widespread and opaque. In order to fully rectify the issue (whilst maintaining the same simplicity in test set-up moving forward) a metaclass solution is proposed.

The proposed solution wraps all methods that inherit from `TestCase` with a wrapper function that fixes the random seed to a hash code derived from the method's name. This permits hashes to be changed in problematic circumstances by a hash modifier, such as appending a trailing underscore (or other modifier) to the method name.

Specifically, the `TestCase` class, which is described as the 'Custom TestCase class for pybamm' is given a custom metaclass. However, it was also noted that this test class is not in widespread use across the testbase, and hence `unittest.TestCase` has now been replaced with `TestCase` as imported from `tests` across all test scripts.

Please add a line in the relevant section of [CHANGELOG.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CHANGELOG.md) to document the change (include PR #) - note reverse order of PR #s. If necessary, also add to the list of breaking changes.

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (back-end change that speeds up the code)
- [X] Bug fix (non-breaking change which fixes an issue)

# Key checklist:

- [X] No style issues: `$ pre-commit run` (see [CONTRIBUTING.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CONTRIBUTING.md#installing-and-using-pre-commit) for how to set this up to run automatically when committing locally, in just two lines of code)
- [X] All tests pass: `$ python run-tests.py --all`
- [X] The documentation builds: `$ python run-tests.py --doctest`

You can run unit and doctests together at once, using `$ python run-tests.py --quick`.

## Further checks:

- [X] Code is commented, particularly in hard-to-understand areas
- [ ] Tests added that prove fix is effective or that feature works

PR #3658, opened 2023-12-23
Idaklu-Jax Interface

### LINKED ISSUE #2282: new JAX solver based on IDAKLU for parameter inference
### Description

The existing pybamm JAX solver is useful for incorporation into ML libraries for parameter estimation, but it is not as fast or stable as the casadi or idaklu solvers. 

This issue proposes writting a C++ JAX wrapper around the new IDAKLU solver so that this solver can also be used, see https://dfm.io/posts/extending-jax/

This allows for a powerful solver to be incorporated into libraries such as Tensorflow (probability) for parameter inference (@brosaplanella). 

Note the existing problem that jax does not support sparse matrices. However, the problem can be posed with the jacobian implicit, but defining a jacobian action function (J * a), which can then be passed to the idaklu solver and used.

### Motivation

_No response_

### Possible Implementation

_No response_

### Additional context

_No response_

jsbrittain:
# Description

Provides an IDAKLU-JAX interface by exposing a JAX expression that performs an IDAKLU solve. There is a new IDAKLUJax API page in the docs, and usage examples have been provided in a new jupyter notebook.

Fixes #2282 

## Type of change

# Key checklist:

## Further checks:


jsbrittain:
Hi @martinjrobins  - I'm ready for the IDAKLU-Jax interface to be reviewed. A good starting point would be the example jupyter notebook that I have included as this walks through a usage example.

PS: Benchmarks have run twice and failed on different tests each time (which do not appear related to this PR); I've re-started them again.

CMakeLists.txt
```diff
@@ -50,6 +50,8 @@ pybind11_add_module(idaklu
   pybamm/solvers/c_solvers/idaklu/CasadiSolverOpenMP_solvers.hpp
   pybamm/solvers/c_solvers/idaklu/casadi_sundials_functions.cpp
   pybamm/solvers/c_solvers/idaklu/casadi_sundials_functions.hpp
+  pybamm/solvers/c_solvers/idaklu/idaklu_jax.cpp
+  pybamm/solvers/c_solvers/idaklu/idaklu_jax.hpp
```
agriyakhetarpal, line 54: Usually we add these files here for CMake to ask `pybind11` to create bindings for, but the corresponding list of `sources` in the `Extension` in `setup.py` does not get updated in accordance with these files. Could you add these there as well? We should probably add a comment in both places so that they go hand-in-hand and are in sync on further updates.

agriyakhetarpal (commented):
Hi @jsbrittain – a very unsolicited review, since I cannot review the rest of the changes here with as much fervour. But I have a few suggestions

pyproject.toml
```diff
@@ -130,6 +130,7 @@ odes = [
 all = [
     "autograd>=1.6.2",
     "scikit-fem>=8.1.0",
+    "parameterized>=0.9",
```
agriyakhetarpal, line 133: I would suggest moving this to the `[dev]` extra since the only import of it at this time is in the unit tests (unless it's going to be used for user-facing functionality too in the future, in which case it can stay here, that is)

jsbrittain: I don't think this works, since `nox -s unit` fails (`No module named parameterized`); taking a deeper look, the noxfile installs `[all]` for tests (and selected others), but none of these include `[dev]` dependencies. Do we want to add `dev` to tests, or create a new 'tests' environment (and should that be part of this PR)?

agriyakhetarpal: Yes, that makes sense. We have been using the `[dev]` extra to collect a set of test-related dependencies; we should add them for the `nox` environments for the unit, integration, and coverage tests (should be fine to include in this PR since it is a minimal change at best)

pybamm/solvers/idaklu_jax.py
```diff
@@ -0,0 +1,956 @@
...
+            elif out.ndim == 1:
+                return out[index]
+            else:
+                return out[:, index]
+
+        return f_isolated
+
+    def jax_value(
+        self,
+        *,
```
martinjrobins, line 133: I've not seen this, what does the `*` do here?

pybamm/solvers/idaklu_jax.py
```diff
@@ -0,0 +1,956 @@
...
+        output_variables=None,
+        inputs=None,
+        calculate_sensitivities=True,
+    ):
+        """JAXify the model and solver"""
+
+        self.jax_model = model
+        self.jax_t_eval = t_eval
+        self.jax_output_variables = (
```
martinjrobins, line 474: some of these attributes are essential for most of the functions above, would it work doing the jaxification in the `__init__` so you are sure these attributes are set?

pybamm/solvers/idaklu_jax.py
```diff
@@ -0,0 +1,956 @@
...
+        # Reconstruct dictionary of inputs
+        if self.jax_inputs is None:
+            d = self._hashabledict()
+        else:
+            # Use hashable dictionaries for caching the solve
+            d = self._hashabledict()
+            for key, value in zip(self.jax_inputs.keys(), inputs_values):
+                d[key] = value
+        # Solver
+        logging.debug("_jaxify_solve:")
```
martinjrobins, line 240: Are these meant to be here? I notice you aren't using the normal pybamm logger, which is fair enough since they are quite low level.

martinjrobins: I can see the utility for creating a dev logger for this solver, but it should be given a name and documented somewhere.

tests/unit/test_solvers/test_idaklu_jax.py
```diff
@@ -0,0 +1,901 @@
...
+    from jax.tree_util import tree_flatten
+    import jax
+    import jax.numpy as jnp
+
+    inputs = {
+        "Current function [A]": 0.222,
+        "Separator porosity": 0.3,
+    }
+
+    model = pybamm.lithium_ion.DFN()
```
martinjrobins, line 22: Should use a simpler model here, e.g. exponential decay, so that the test is faster. I don't think using a complicated model has any value since you are re-using the idaklu solver

martinjrobins (changes requested):
excellent work, thanks @jsbrittain! I've got a few questions below and suggestions on the api. Have you done any benchmarking on the final jax expressions (jit and non-jit) compared with pure pybamm calls?

tests/unit/test_solvers/test_idaklu_jax.py
```diff
@@ -0,0 +1,901 @@
...
+@unittest.skipIf(
+    not pybamm.have_idaklu() or not pybamm.have_jax(),
+    "IDAKLU Solver and/or JAX are not available",
+)
+class TestIDAKLUJax(TestCase):
+    # Initialisation tests
+
+    def test_initialise_twice(self):
+        print("Initialise solver")
+        idaklu_jax_solver = idaklu_solver.jaxify(
```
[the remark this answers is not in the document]
jsbrittain: Yes, that will work fine, and we already do this to test the single output `f1` solver and the multiple output `f3` solver as part of the tests. Supporting multiple simultaneous interfaces is why we use unique JAX primitive names in the code; however, we cannot generate multiple JAX expressions from the same idaklu-jax object as instantiation overwrites certain member variables. As an aside, I was initially not going to expose the idaklu-jax object at all (just return the jax expression), but it became necessary to provide easy access to helper functions, such as `get_var`.

pybamm/solvers/idaklu_jax.py
jsbrittain: A bare asterisks prevents any subsequent arguments from being callled positionally; this forces them to be named and so is useful to prevent misordering based on their position alone (see also https://stackoverflow.com/a/14302007).

pybamm/solvers/idaklu_solver.py
```diff
@@ -675,3 +676,41 @@ def _integrate(self, model, t_eval, inputs_dict=None):
...
+        model : :class:`pybamm.BaseModel`
+            The model to be solved
+        t_eval : numeric type, optional
+            The times at which to compute the solution. If None, the times in the model
+            are used.
+        output_variables : list of str, optional
+            The variables to be returned. If None, the variables in the model are used.
+        inputs : dict, optional
+            Any inputs to the model
+        calculate_sensitivities : bool, optional
```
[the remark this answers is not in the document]
jsbrittain: Yes, I wasn't sure whether to retain this argument or not, since it's value can be trivially determined based on whether `inputs` are present during the solve. However, it does form part of the API for the IDAKLU solver, and it is functional (you can specify `False` to prevent sensitivities from being calculated, in which case `grad` will fail with a `ValueError: Cannot compute sensitivities. The 'calculate_sensitivities' argument of the solver.solve should be changed from 'None' to allow sensitivities calculations. Check solver documentation for details.`; by default the argument is `True`). I would be just as happy to remove it as an argument from the interface though.

pybamm/solvers/idaklu_jax.py
```diff
@@ -0,0 +1,956 @@
...
+        ----------
+        f : function
+            The jaxified expression
+        varname : str
+            The name of the variable to extract
+        """
+
+        def f_isolated(*args, **kwargs):
+            out = f(*args, **kwargs)
+            index = self.jax_output_variables.index(varname)
```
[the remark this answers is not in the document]
jsbrittain: Resolved by jaxify'ing upon initialisation of the class.

pybamm/solvers/idaklu_jax.py
```diff
@@ -0,0 +1,956 @@
...
+        f: Callable,
+        varname: str,
+    ):
+        """Helper function to extract a single variable from the jaxified expression
+
+        Returns a JAX expression
+
+        Parameters
+        ----------
+        f : function
```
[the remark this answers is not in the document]
jsbrittain: Yes, this has been refactored a few times and at some point was a static method that confirmed to the typical jax functional form of `x(y(f))(a, b, c)`. However, since the jaxified expression is now bound to the idaklu-jax instance this requirement can be dropped and the `f` in the arguments list can be removed and replaced with the jaxpr. As such I've now removed the `f` such that this function now effectively replaces `f` in user calls when filtering the results by variable name(s). I have amended the docs (specifically the jupyter notebook) and tests to reflect this.

pybamm/solvers/idaklu_jax.py
jsbrittain: Yes, okay. I've jaxified on initialisation.

pybamm/solvers/idaklu_jax.py
jsbrittain: I've named the logger and made reference to it in the class docstring.

pybamm/solvers/idaklu_solver.py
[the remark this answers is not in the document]
martinjrobins: happy to keep it, all sounds good

pybamm/solvers/idaklu_jax.py
martinjrobins: nice :)

tests/unit/test_solvers/test_idaklu_jax.py
jsbrittain: Okay (it was useful for initial testing!), but I've replaced it with a simplified expression set for future CI.

pybamm/solvers/idaklu_solver.py
```diff
@@ -675,3 +676,41 @@ def _integrate(self, model, t_eval, inputs_dict=None):
...
+        Parameters
+        ----------
+        model : :class:`pybamm.BaseModel`
+            The model to be solved
+        t_eval : numeric type, optional
+            The times at which to compute the solution. If None, the times in the model
+            are used.
+        output_variables : list of str, optional
+            The variables to be returned. If None, the variables in the model are used.
+        inputs : dict, optional
```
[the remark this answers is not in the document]
jsbrittain: Yes, while it is true that we don't need `inputs` (or `t_eval`) until evaluation, we do still need knowledge of them in order to setup/jaxify the interface.

I've taken another look at this and I can extract the names of the `inputs` from the model definition (removing the need to include `inputs` in this call). But we also rely on knowing the span of `t_eval` when setting up the idaklu-jax interface. This is because calls to some functions (including `grad`) need to be made over `vmap` (or the user may simply evaluate the expression at only one time point). These details are abstracted away from the user wherever possible in our interface, but we get around this by caching a solve over the predefined `t_span` on a first run.

Hopefully that makes sense. I've removed `inputs` from the call as superfluous, but retained `t_eval` and made a comment in the jupyter notebook to signify that this defines the time-span scope of the solver.

pybamm/solvers/idaklu_jax.py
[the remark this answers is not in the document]
jsbrittain: I'm going to have to row-back on this one. If we use only `jaxpr` for the expression then we cannot pass higher-order functions that the user may have constructed (a prominent example being `jax.jit(f)` into this function); it was designed to be a helper function to isolate variables from any expression (not just the base jaxpr).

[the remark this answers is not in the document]
jsbrittain: On further testing I've decided to refactor this again to accept different call signatures. Now, you *can* call it with only the variable name of interest (in which case it will use jaxpr); but you can also provide a JAX expression (useful if you want to perform some other function, e.g. `jax.jit(f)`) before isolating the variable, or you can provide an `np.ndarray`, which allows you to extract derivates from functions like `jacfwd` more easily. I think this provides the best of both worlds, and have added an example use-case in the jupyter notebook for the latter case, specifically:
```python
d = jax.jacfwd(f, argnums=1)(t_eval, inputs)  # returns a dict of np.ndarrays
data = jax_solver.get_var(d["Current function [A]"], "Voltage [V]")  # Isolates derviate of V wrt current fcn
```

jsbrittain:
> Have you done any benchmarking on the final jax expressions (jit and non-jit) compared with pure pybamm calls?

The following benchmarks were obtained on a Macbook M2. They include building a DFN model with two inputs, running a single solve, then printing a single variable to the screen. Benchmarks were run twice, (T=1,000 and 10,000 samples). Each result is the average over 10 loops, taking the best of 5 executions for each loop.

To run and print a single variable:
IDAKLU: 455 msecs (T=1000), 2.34 secs (T=10,000)
IDAKLU-Jax: 473 msecs (T=1000), 2.34 secs (T=10,000)
IDAKLU-Jax[jit]: 464 msecs (T=1000), 2.35 secs (T=10,000)

For sensitivities / gradient data:
IDAKLU: 455 msecs (T=1000), 2.35 secs (T=10,000)
IDAKLU-Jax: 501 msecs (T=1000), 2.60 secs (T=10,000)
IDAKLU-Jax[jit]: 517 msecs (T=1000), 2.64  secs (T=10,000)

The native IDAKLU solver will be the fastest since our idaklu-jax interface calls IDAKLU for the solve. Computing gradients under idaklu-jax incurs some additional cost due to the need to restructure the data during calls to, e.g. JVP routines. Gradients were computed in this case using the forward method (`jacfwd`).

BradyPlanden:
Excellent job @jsbrittain! Just to note, I believe this completes the functionality we wanted from #3395, so we can probably close that one with this PR as well.

jsbrittain:
@martinjrobins Ready for re-review (Lychee seems to be failing in `develop` at the moment).

CHANGELOG.md
```diff
@@ -5,7 +5,7 @@
 ## Features
 
 - Added a JAX interface to the IDAKLU solver ([#3658](https://github.com/pybamm-team/PyBaMM/pull/3658))
-- The `pybamm_install_odes` command now includes support for macOS systems and can be used to set up SUNDIALS and install the `scikits.odes` solver on macOS ([#3417](https://github.com/pybamm-team/PyBaMM/pull/3417))
+- The `pybamm_install_odes` command now includes support for macOS systems and can be used to set up SUNDIALS and install the `scikits.odes` solver on macOS ([#3417](https://github.com/pybamm-team/PyBaMM/pull/3417), [#3706](https://github.com/pybamm-team/PyBaMM/3706]))
```
martinjrobins, line 8:
```suggestion
- The `pybamm_install_odes` command now includes support for macOS systems and can be used to set up SUNDIALS and install the `scikits.odes` solver on macOS ([#3417](https://github.com/pybamm-team/PyBaMM/pull/3417))
```

martinjrobins (approved):
looks great, just one change in the changlog then it can be merged

agriyakhetarpal:
Small note: please add an entry for `parameterized` to the optional dependencies table in the installation guide

agriyakhetarpal:
Question: since we do not support the JAX solver on Python 3.8 anymore after having updated the versions to `0.4.20`, should we make this explicit by not allowing the usage for `pybamm.IDAKLUJax` for Python < 3.9? I don't know what the best practice is: whether it should be either a helpful error message on instantiation, or a note in the API documentation, or just not importing the class in the public namespace in `pybamm/__init__.py` if `sys.version_info < (3, 9)`, or something else. The tests for this functionality are being ignored because of the combination of the version checks in `noxfile.py` and the `@unittest.skipIf()` decorators, currently.

In any case, we would want clearer communication to the users about the versioning information for this functionality to be used – perhaps in the example notebooks and/or in the CHANGELOG, or any other appropriate place as necessary

jsbrittain:
@agriyakhetarpal good point - note that the idaklu-jax class does have instantiation guards that will throw if either Jax or Idaklu are not installed, as well as import checks so there are no problems when importing the namespace, so from a code perspective it should be safe. Just to be sure, I have added a unit test to make sure that these scenarios throw as expected when testing under py3.8.

From a user/documentation perspective, while we could add messages specifically about this class to the docs (and also JaxSolver, etc.), the fact that PyBaMM itself only supports Jax for python 3.9+ seems sufficient to me, so long as that requirement is clearer visible in the docs - and there does appear to be a prominent note in the optional solvers section of the install guide to this effect. I think I prefer this approach to peppering the code/docs with provisos - but am nevertheless happy to add a note somewhere if you think it is necessary - Let me know.

agriyakhetarpal:
> the idaklu-jax class does have instantiation guards that will throw if either Jax or Idaklu are not installed, as well as import checks so there are no problems when importing the namespace, so from a code perspective it should be safe

Yes, this should work. It would have raised an error on importing PyBaMM if it wasn't done properly anyway – we recommend  using `nox` as a task runner rather than the `python run-tests.py --unit` command for testing in isolated environments and handling dependencies, so we should be good with handling such cases.

> From a user/documentation perspective, while we could add messages specifically about this class to the docs (and also JaxSolver, etc.), the fact that PyBaMM itself only supports Jax for python 3.9+ seems sufficient to me, so long as that requirement is clearer visible in the docs - and there does appear to be a prominent note in the optional solvers section of the install guide to this effect. I think I prefer this approach to peppering the code/docs with provisos - but am nevertheless happy to add a note somewhere if you think it is necessary - Let me know.

I agree with you. Adding admonitions in too many places might bring up synchronisation problems wherein we modify a file and miss out on modifying another, which can leave users misguided – the ones existing can stay. I am not able to access the Plausible Analytics instance we have been using for our website and documentation at this time, but I remember from some previous run-throughs that the Examples page is one of the popular ones that we have and receives a lot of views, many magnitudes higher than those on many other pages. Perhaps the inclusion of an additional note in the notebook should suffice – given that we won't drop Python 3.8 or support Python 3.13 until their EOL and release dates respectively next year.

jsbrittain:
@agriyakhetarpal okay, I've added the following note at the top of the jupyter notebook: "The IDAKLU-JAX interface requires that PyBaMM is installed with the [optional JAX solver enabled](https://docs.pybamm.org/en/stable/source/user_guide/installation/gnu-linux-mac.html#optional-jaxsolver) (`pip install pybamm[jax]`). PyBaMM currently provides JAX support for Linux and MacOS only, and requires at least Python 3.9."

(the link will not be live yet as it references a docs page that only recently been renamed/added and which hasn't propagated through to the stable docs yet).

agriyakhetarpal:
There is a PR preview on Read the Docs here: https://pybamm--3658.org.readthedocs.build/en/3658/

N.B. We have [JAX CPU-only support for Windows](https://jax.readthedocs.io/en/latest/installation.html#pip-installation-cpu) as well now, since they started publishing Windows 64-bit wheels with `0.4.14`. GPU support has a different scenario and continues to exist only for Linux and WSL2 officially

BradyPlanden (approved):

jsbrittain:
@agriyakhetarpal thanks for those links. i've rephrased the requirements slightly in the jupyter notebook to avoid mentioning OSs directly.

jsbrittain:
Ready to merge. A note on the failing tests:
- Lychee is failing because of a documentation link in a new jupyter notebook that references a docs page that does not exist yet in the `stable` deployment (not sure how to avoid this).
- doctests run fine on my machine on py3.11, but fail on py3.12. I am seeing the same issue in other recent PRs too (`the pydata_sphinx_theme extension is not safe for parallel writing`), which appears to be a broader problem with that test.

On a side note, some of the unit tests required multiple runs to pass. This should not happen as they were made deterministic in #2844 - it appears as though some test classes have been introduced that inherit from `unittest.TestCase` instead of PyBaMM's custom `TestCase` class (which fixes the random seed). I will open a separate issue to address this.

agriyakhetarpal:
> - doctests run fine on my machine on py3.11, but fail on py3.12. I am seeing the same issue in other recent PRs too (`the pydata_sphinx_theme extension is not safe for parallel writing`), which appears to be a broader problem with that test.

Thanks, I noticed this too. I don't think there are any larger problems at play since it's just a warning message – I will take a look in a separate PR

agriyakhetarpal:
Yet another note: since this PR has worked on a lot of compiled code, I would suggest running the wheel builds workflow on this branch once, just so that we don't face any problems later.

P.S. my suggestion is to make such checks automatic in some cases through our infrastructure in the long term, such as those being discussed in #3662.

jsbrittain:
> since this PR has worked on a lot of compiled code, I would suggest running the wheel builds workflow on this branch once, just so that we don't face any problems later.

@agriyakhetarpal  Good idea - I just want to check how to do this without breaking anything: I can run the `Build and publish package to PyPI` Action on this branch (and with `publish to testpypi`) to check this without breaking anything, can't I? I think I've done this before, but wanted to double-check.

agriyakhetarpal:
> @agriyakhetarpal Good idea - I just want to check how to do this without breaking anything: I can run the `Build and publish package to PyPI` Action on this branch (and with `publish to testpypi`) to check this without breaking anything, can't I? I think I've done this before, but wanted to double-check.

@jsbrittain yes – the configuration by default is designed to not break anything. You can just set the workflow input as blank to avoid publishing on TestPyPI either. The PyPI publishing is guarded by the condition

https://github.com/pybamm-team/PyBaMM/blob/b84f7ad46993071f36b3811feedb263da2340ca5/.github/workflows/publish_pypi.yml#L197

which helps automatically skip the job regardless of the input on non-release runs.

jsbrittain:
> since this PR has worked on a lot of compiled code, I would suggest running the wheel builds workflow on this branch once, just so that we don't face any problems later.

@agriyakhetarpal wheels build, see https://github.com/jsbrittain/PyBaMM/actions/runs/7632700433.

arjxn-py:
> see https://github.com/jsbrittain/PyBaMM/actions/runs/7632700433.

For some reason, I can't access this build. Although it should be alright but would it be nicer to test the wheels too? 
Since this is such a big PR and involves crucial infrastructural development, wanted to appreciate and thanks a lot @jsbrittain

agriyakhetarpal:
I can't access the link either but I trust the universe with it haha. The wheels are currently being tested by `cibuildwheel` for all platforms through the instantiation of `pybamm.IDAKLUSolver` (#3569). After we migrate to a better test infrastructure by the end of the year (or fall if sooner), the idea is to run the integration tests there

jsbrittain:
> For some reason, I can't access this build. Although it should be alright but would it be nicer to test the wheels too? Since this is such a big PR and involves crucial infrastructural development, wanted to appreciate and thanks a lot @jsbrittain

@arjxn-py @agriyakhetarpal I disabled Actions on my fork after running the workflow - I didn't realise that would prevent you from seeing the report / artifacts. I've re-enabled them and the link is working for me again: https://github.com/jsbrittain/PyBaMM/actions/runs/7632700433

jsbrittain:
@martinjrobins fixed the merge conflicts; all tests passing (apart from lychee).

BradyPlanden:
Any updates on merging this? It would be a great addition from the parameter identification workflow side. @jsbrittain @martinjrobins @pybamm-team/maintainers

jsbrittain:
@BradyPlanden @martinjrobins @pybamm-team/maintainers  I've updated the branch with the latest changes from `develop`, but do not have permissions to merge the other way, so will require someone else to authorise that.

PR #3661, opened 2023-12-24
Build & Push Docker Images on releases

arjxn-py:
# Description

Push Docker images on releases, starting from either `23.9` or `24.1` before the `24.1rc0` release

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -5,6 +5,8 @@ on:
   push:
     branches:
     - develop
+  release:
+    types: [published]
 
 jobs:
   build_docker_images:


arjxn-py:
Related to #3312 
_cc: @agriyakhetarpal_

.github/workflows/docker.yml
```diff
@@ -5,6 +5,8 @@ on:
   push:
     branches:
     - develop
+  release:
+    types: [published]
```
Saransh-cpp, line 9: I'm very rusty at dockerhub stuff at the moment, but IIRC, shouldn't we tag the "release" images with the `git tag` of the pybamm's release?

I mean - won't this overwrite the "latest" (develop) image on dockerhub with the "release" image, and then as soon as a new commit is pushed to develop, the "release" image will be overwritten by a new "latest" image?

agriyakhetarpal: Yes, I'm aware. We're still discussing atm about how the tags will be set since we want to label the image with either the PyBaMM version or the name of the release (since it contains the version) – the latter is easier to extract from the GitHub events API. Marking the PR as draft for now since it's not ready yet

arjxn-py: > shouldn't we tag the "release" images with the `git tag` of the pybamm's release?

Yes, you're right. We'll Push Docker Images Based on release tags only, that way we'll have a different image for each release but as @agriyakhetarpal mentioned we're discussing which Image should we push for release to keep it single and simple for each release (i.e. We're mostly in favor of image with `IDAKLU` or `ALL` solver), We'd be happy to have your opinion on this @Saransh-cpp

agriyakhetarpal: Yes, we're also discussing reducing the number of images we build at this time – the reason is that the `latest` image without any solvers is just a simple git clone command and doesn't make sense to include, the `jax` image can be superseded by `pip install pybamm[jax]` where wheels are currently available for all platforms, which makes the installation a matter of seconds. The images where there are any compilation steps present are `odes`, `idaklu` and subsequently `all` – all of them use the same SUNDIALS that has been compiled in `.local/`. My opinion is towards just pushing the `all` image to accommodate all solvers ehile pushing no other images. We run the image builds on commits to the main branch and the arm64 image workflow runs for ~40 mins due to QEMU emulation, so we are not constrained about time or size either. @arjxn-py, could you open a new issue about this?

agriyakhetarpal: Also, we've been mixing `conda` and `virtualenv` environments in the image – while that does work without issues now, it's not really a good idea to do so, even in late 2023. I would suggest switching to using a manylinux2014-compliant base image (i.e., with `glibc`) in the not-so-long term; it will essentially be the process of containerising PyBaMM from the ground up once again, but the added experience in this area for all of us in the past months would mean it can be done pretty quickly.

arjxn-py: > could you open a new issue about this?

Sure, I'm up for making the `all` image as standard one if we do not consider size. Although, this is also to be noted that `latest`, `jax` & `odes` image with `all` also manages the environment & installation from source which I feel is different than `pip install pybamm[extras]`. Happy to open a new Issue about the same and have a discussion there 🙂

agriyakhetarpal: Yeah ultimately it doesn't make sense for the `jax` image to have SUNDIALS pre-installed. The [odes] extra is the only one where we don't have wheels on PyPI so we have to compile from source – perhaps they are not publishing because of copyleft licensing issues...

.github/workflows/docker.yml
```diff
@@ -48,7 +51,24 @@ jobs:
             echo "tag=all" >> "$GITHUB_OUTPUT"
           fi
 
+      - name: Create tags for Docker images based on build-time arguments (release)
+        if: github.event_name == 'release' && github.event.action == 'published'
+        id: tags-release
+        run: |
+          if [ "${{ matrix.build-args }}" = "No solvers" ]; then
+            echo "tag=${{ github.event.release.name }}-latest" >> "$GITHUB_OUTPUT"
```
agriyakhetarpal, line 59: I am not sure yet if `github.event.release.name` will work, we might need to [use github.ref_name` instead](https://github.com/orgs/community/discussions/26686#discussioncomment-3396593) to get the version. 

I will request @arjxn-py to uncomment the `if:` condition at the top, comment out the login step, change `push: true` to `push: false` and check the tag of the built image manually.

.github/workflows/docker.yml
```diff
@@ -57,5 +77,12 @@ jobs:
           push: true
           platforms: linux/amd64, linux/arm64
 
-      - name: List built image(s)
-        run: docker images
```
agriyakhetarpal, line 61: Turns out this step wasn't really doing anything except for listing the images present on GitHub Actions runners

.github/workflows/docker.yml
arjxn-py: I've done this in https://github.com/arjxn-py/PyBaMM/actions/runs/7350486868, as defined it is taking in release name (i.e. `Test Release` in this case) which looks alright but it would be nicer to have tag instead, hopefully the one you suggested https://github.com/orgs/community/discussions/26686#discussioncomment-3396593 do the work

arjxn-py: Yes now it takes in tags in - https://github.com/arjxn-py/PyBaMM/actions/runs/7350650963/job/20012681428

agriyakhetarpal:
Question: what will happen if the `git tag` is manually adjusted through a force push, a few commits ahead or behind in the tree? Will that trigger a push? 

We won't need to do that usually, but maybe a way out of that situation would be to be able to set the image tags from inputs to the workflow (we already support the `workflow_dispatch:` event) and overwrite the release image – this would tie in with reducing the amount of images we push. But then we currently clone the repository inside the `Dockerfile`, which would always fetch the latest changes from `develop`... perhaps @Saransh-cpp can confirm the most recent instance of having to adjust the GitHub tag, was it for 23.9rc0?

.github/workflows/docker.yml
arjxn-py: > won't this overwrite the "latest" (develop) image on dockerhub with the "release" image, and then as soon as a new commit is pushed to develop, the "release" image will be overwritten by a new "latest" image?

Guess I missed this while answering previously, Sorry.
No, this wont overwrite the previous release image as two different release images would have two different tags and as soon as tags are different the images are distinct. Maybe you can notice [here](https://hub.docker.com/_/python/tags) how there are different tags and different images.

.github/workflows/docker.yml
agriyakhetarpal, line 1: The entire file shows up as changed, possibly due to a difference in line endings schemes?

.github/workflows/docker.yml
```diff
@@ -1,61 +1,88 @@
...
+name: Build and push Docker images to Docker Hub
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+    - develop
+  release:
+    types: [published]
```
agriyakhetarpal, line 9:
```suggestion
  release:
    types: [released]
```

in light of https://github.com/pybamm-team/PyBaMM/pull/3661#issuecomment-1871448550. The changes from this PR won't be available until `24.5rc0` anyway, so we have time to resolve #3692 and #3666 till then – they should be resolved prior to resuming work on this PR.

agriyakhetarpal:
@arjxn-py, we haven't been publishing release candidates for some time now. Could you please revisit this PR? I think we should be unblocked and we don't need anything complex. I think with the Docker sponsored OSS programme we should have more storage as well for the images (if I remember correctly).

agriyakhetarpal:
@arjxn-py, we haven't been publishing release candidates for some time now. Could you please revisit this PR? I think we should be unblocked and we don't need anything complex. I think with Docker's sponsored OSS programme that you steered our application for, we should have more storage as well for the images (if I remember correctly).

agriyakhetarpal:
This will be superseded by #5068, so I shall close this now. Thanks!

PR #3682, opened 2024-01-03
Removing old arguments

kratman:
# Description

These arguments have been deprecated for a while. I figured it would be better to remove them to reduce code and error handling.

## Type of change

Minor cleanup.

# Key checklist:

## Further checks:


agriyakhetarpal:
Deprecated things are kept in the codebase for at least a year (at least three releases), are we in a position to remove these?

kratman:
> Deprecated things are kept in the codebase for at least a year (at least three releases), are we in a position to remove these?

Is this documented somewhere? I don't think I have seen that rule and I could not find it with a text search in our guidelines.

> I cannot exactly ascertain the duration from the git blame

The tests I removed were put in May 2023, so sometime around then.

agriyakhetarpal:
> > Deprecated things are kept in the codebase for at least a year (at least three releases), are we in a position to remove these?
> 
> Is this documented somewhere? I don't think I have seen that rule and I could not find it with a text search in our guidelines.

It's mentioned in the [Versioning section of the README](https://github.com/pybamm-team/PyBaMM#versioning). I agree it's a bit difficult to find though and should be communicated more transparently in the Contributing Guide for developers and future contributors. Please feel free to add this note more explicitly at some place where you think it would be apt.

> The tests I removed were put in May 2023, so sometime around then.

Ah I see, I don't have an opinion on this so I will leave it up to the others to see if removing it now should be fine, or we can come back to merging this PR later at the time of 24.5rc0

pybamm/experiment/experiment.py
```diff
@@ -97,6 +80,8 @@ def __init__(
                 processed_step = pybamm.step.string(step)
             elif isinstance(step, pybamm.step._Step):
                 processed_step = step
+            else:
+                raise TypeError("Operating conditions must be a Step object or string.")
```
rtimms, line 84: can you add a test for this error?

rtimms:
happy to merge this now and have it go in 24.5

kratman:
> happy to merge this now and have it go in 24.5

I will add the test and I also want to do a search for other outdated deprecation warnings. I will update soon

agriyakhetarpal:
Other deprecations that have to be removed from the codebase in v24.5 are all the entry points from `pyproject.toml`, with the exception of `pybamm_install_odes`

kratman:
> Other deprecations that have to be removed from the codebase in v24.5 are all the entry points from `pyproject.toml`, with the exception of `pybamm_install_odes`

@agriyakhetarpal Since you are more familiar with the entry points than I am, feel free to push that change to this branch.

pybamm/experiment/experiment.py
```diff
@@ -129,6 +99,27 @@ def __init__(
...
+
+            if processed_step.period is None:
+                processed_step.period = period
+            if processed_step.temperature is None:
+                processed_step.temperature = temp
+
+            processed_steps[repr(step)] = processed_step
+        return processed_steps
```
kratman, line 121: @rtimms I extracted this method so it was easily testable. I added a test for it.

pybamm/experiment/experiment.py
kratman: Done

agriyakhetarpal:
Actually, `pybamm_install_jax` was deprecated in `v23.9`, and therefore we must refrain from removing it until `v24.9`. I removed the command-line interface to the parameter sets in https://github.com/pybamm-team/PyBaMM/pull/3682/commits/42df246e689cd65ee53a29008d99e49bf030fae8. It was deprecated in `v22.10` and should have been removed by `v24.1` – makes me realise we missed doing that.

rtimms (approved):
thanks @kratman

PR #3706, opened 2024-01-10
Make `pybamm_install_odes` a bit more robust

agriyakhetarpal:
# Description

This was a small detail that was missed; the `pip install` invocation inside a script should be done with a `sys.executable` (i.e., `python`/`python3`/`python3.X`) because some users' systems do not have `pip` added to the `PATH` environment variable explicitly – such as non-`virtualenv`-based virtual environments or those using system package managers such as `apt-get` or `brew` to manage Python dependencies This change makes it more robust and possible to work outside a virtual environment (say, if someone does not wish to use one for their own purposes, even though we strongly recommend that).

I added a `--verbose` command-line flag with the lowest level of verbosity possible just so that some output can be seen in the terminal, even on successful installations.

`wget` and `cmake` were added as Pythonic prerequisites to the installation guide to render the procedure clearer for users.

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ## Features
 
-- The `pybamm_install_odes` command now includes support for macOS systems and can be used to set up SUNDIALS and install the `scikits.odes` solver on macOS ([#3417](https://github.com/pybamm-team/PyBaMM/pull/3417))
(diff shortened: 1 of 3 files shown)


agriyakhetarpal:
Tagging @Saransh-cpp, would be good to include this change into 24.1 when we are ready for that

arjxn-py (approved):
Looks better, thanks @agriyakhetarpal

agriyakhetarpal:
This one is not as essential a bug fix (since this will indeed work inside venvs) and should be fine to not go into non-RC release, but we would hope it won't break things so I would suggest its inclusion to be on the safer side of things.

PR #3719, opened 2024-01-12
Replace wget with urllib for downloading and parallelize with concurrent.futures for efficiency

### LINKED ISSUE #3651: Tracking issue for improvements to `install_KLU_Sundials.py` and build infrastructure
> [!WARNING]
> Following an imminent migration for the build system and/or choice of build-backend for PyBaMM, not all of the listed points may not be applicable for completion. Please refer to #3564 for more information.

There are a number of improvements that can be introduced to improve the installation of the build-time dependencies (SuiteSparse and SUNDIALS) and the IDAKLU extension module and general improvements to the installation scripts:

- [x] Incorporate the SHA-256 hashes of the SuiteSparse and SUNDIALS release tarballs as constants in the script in order to save bandwidth and skip downloading the tarball every time. (#3719)
- [x] Do not install locally if the binaries are already present at the installation location (#3418) (#3719)
- [x] Set parallel builds via the use of the `CMAKE_BUILD_PARALLEL_LEVEL` environment variable (#3646)
- [x] Introduce PEP 723 support (needs triage and should be reviewed at a later time in 2024) (#3647) (#3716)
- [ ] Establish local SUNDIALS and SuiteSparse installations in a folder `sundials_installed/` by adjusting `CMAKE_INSTALL_PREFIX` instead of modifying the `.local/` folder and check this folder to establish whether the IDAKLU extension module should be compiled or not in the `build_ext` command class in `setup.py` (adding `pybind11` as a build-time dependency should be possible in #3480 after this)
- [ ] Build just IDAS instead of all SUNDIALS components, considering that is what we link to for the IDAKLU target
- [ ] Set relevant SUNDIALS and SuiteSparse compilation options such as skipping building the examples (done in #3569), link to OpenBLAS instead of Apple's Accelerate on macOS since NumPy and SciPy are not recommending it for now
- [ ] On macOS, we use compiler options to find `libomp`, they can be removed and we can use `pkg-config` from Homebrew that should be able to do it automatically. The reason for this is because Apple Clang does not come with the OpenMP specification – however, other compilers such as LLVM and GCC do (noted by @kratman: we install `gcc` in CI for macOS, but it is never used!)
- [x] Remove `wget` as a dependency and use `urllib` from the Python standard library for downloading. In addition to this, use `multiprocessing.Pool` or `concurrent.futures` to download both tarballs in parallel to save time (#3719)
- [ ] On platforms we don't support, the `sdist` is used to install PyBaMM, and the IDAKLU solver which is optional cannot be installed unless installing from source. To provide a means to attempt IDAKLU compilation, a `--extra-options` command-line flag can be added to the `pip install pybamm` comand that can pass to `setup.py`. Therefore, the users should have to manually opt-in to the compilation (the build option should be off by default). For this, `pybind11` should be a build-time dependency (see #3480). This requires the points mentioned below:
- [ ] Run `install_KLU_Sundials.py` in a subprocess via a `build_py` command class for macOS and Linux in `setup.py` if the compilation is opted in for. This can eliminate the use of `nox -s pybamm-requires` and a single command `pip install -e .` would do both things
- [ ] Add `CMakeLists.txt` and other relevant files such as `FindSuiteSparse.cmake` and `FindSUNDIALS.cmake` to the SDist via the `MANIFEST.in` file, otherwise CMake will raise an error about a missing project status

These tasks have a mixture of easy and medium difficulty. Some of them have a higher priority than the rest, I am glad to take those up myself.

AlessioBugetti:
# Description

This update replaces the previous wget-based approach with urllib from the Python standard library.
Opted for `urllib.request.urlopen` over `urllib.request.urlretrieve` due to potential future deprecation of `urlretrieve`.
Additionally, it introduces parallelization using `concurrent.futures`, specifically employing `ThreadPoolExecutor` to download SuiteSparse and SUNDIALS tarballs simultaneously. This parallelization significantly reduces the download time.

Related to #3651 (addresses one aspect of the issue).

## Type of change

# Key checklist:

## Further checks:


AlessioBugetti:
Despite adding the following code:

```python
from urllib.parse import urlparse

parsed_url = urlparse(url)
if parsed_url.scheme not in ["http", "https"]:
    raise ValueError(
        f"Invalid URL scheme: {parsed_url.scheme}. Only HTTP and HTTPS
```

the Codacy Static Code Analysis test is not being passed.

agriyakhetarpal:
Thanks for the start @AlessioBugetti! I was already doing some parts of this actually and was going to open a PR soon myself 😅 – this is my fault; I should have commented on the issue as to what parts of it are those that I am taking up. Ideally, this change should be paired with:

1. the completion of #3418, i.e., check for the SUNDIALS and SuiteSparse `.so` or `.dylib` files inside `install_KLU_Sundials.py`, and
2. SHA-256 checksum validation for the downloaded `.tar.gz` files so as to skip downloading them altogether if they are present in the `install_KLU_Sundials/` folder.

I would recommend sequential PRs for these tasks, or doing so in the same PR (to avoid potential merge conflicts that would arise in case of the former)

agriyakhetarpal:
In this case we can skip the Codacy checks – they are mostly for security recommendations which we do not need at this time, especially for a script like this

kratman:
@AlessioBugetti I am not sure you should mark this PR as closing the issue. You are only fixing one aspect of the issue.

AlessioBugetti:
> @AlessioBugetti I am not sure you should mark this PR as closing the issue. You are only fixing one aspect of the issue.

Yes, you're absolutely right, and I apologize for mistakenly marking that issue as closed.

> Thanks for the start @AlessioBugetti! I was already doing some parts of this actually and was going to open a PR soon myself 😅 – this is my fault; I should have commented on the issue as to what parts of it are those that I am taking up. Ideally, this change should be paired with:
> 
> 1. the completion of [Allow `pybamm-requires` to return early if an existing SUNDIALS installation is found #3418](https://github.com/pybamm-team/PyBaMM/issues/3418), i.e., check for the SUNDIALS and SuiteSparse `.so` or `.dylib` files inside `install_KLU_Sundials.py`, and
> 2. SHA-256 checksum validation for the downloaded `.tar.gz` files so as to skip downloading them altogether if they are present in the `install_KLU_Sundials/` folder.
> 
> I would recommend sequential PRs for these tasks, or doing so in the same PR (to avoid potential merge conflicts that would arise in case of the former)

Can I continue on this branch and make a new commit to complete the second task?

agriyakhetarpal:
> Can I continue on this branch and make a new commit to complete the second task?

Yes, please feel free to work on these further in the same PR! (or through what you prefer to be best). I won't mind follow-up PRs for these tasks either

agriyakhetarpal:
Some of these keywords: "fixes", "closes", "solves", "resolves", [and more](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) will close the issue even if you prefix "partially" in front of them. Could you replace that in the PR description with some other word or phrase, say, "related to"?

AlessioBugetti:
> Some of these keywords: "fixes", "closes", "solves", "resolves", [and more](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) will close the issue even if you prefix "partially" in front of them. Could you replace that in the PR description with some other word or phrase, say, "related to"?

I apologize for any inconvenience caused by my initial description. I have now updated the PR description as you requested, replacing the keywords with 'related to' to avoid any automatic issue closures.

AlessioBugetti:
This commit includes:
- A new function, `calculate_sha256`, to compute the checksum.
- Modifications to `download_extract_library` to incorporate checksum validation.

The proposed changes can be seen in the following commit: [link to commit](https://github.com/pybamm-team/PyBaMM/pull/3719/commits/87e39ab1771c561fa1276039f491f29624d025c9).

AlessioBugetti:
I've completed a commit where I addressed the task of ensuring the presence of SUNDIALS and/or SuiteSparse library files (.so or .dylib files) within `install_KLU_Sundials.py`.

Commit [here](https://github.com/pybamm-team/PyBaMM/pull/3719/commits/4622d7709d725eb595b2e5d7f0fff9debded42bd)

agriyakhetarpal:
Looks like the benchmarks are failing because they can't find the SUNDIALS files. About some of the issues listed in #3651, I'm not entirely sure if incorporating them is the best idea at this time, i.e., making it opt-in to compile the IDAKLU solver based on the presence of SUNDIALS and SuitrSparse components will mean it should be opt-in on installing from `pip` + using the SDist, not when building wheels for different platforms (where a wheel is going to be built with the SDist)...

In other things, I think we should look for more files related to the SUNDIALS installation such as those mentioned in #3461. The key reason I was advocating for a local SUNDIALS (with KLU) installation was because we can then point to a local folder `sundials_installed/` for the IDAKLU solver compilation, but if someone wants to use `scikits.odes` (on a non-developer but a user installation), they can install SUNDIALS (without SuiteSparse's KLU) to `.local/`, i.e., in a different place that doesn't overwrite any of the files. I will have to think about the entire compilation procedure in depth so that nothing breaks.

@AlessioBugetti I think these changes look good to me overall once they're done, could you mark the relevant URLs for the build-time components, their checksums, and the installation directories as constants at the top of the file for easier access?

AlessioBugetti:
> Looks like the benchmarks are failing because they can't find the SUNDIALS files. About some of the issues listed in #3651, I'm not entirely sure if incorporating them is the best idea at this time, i.e., making it opt-in to compile the IDAKLU solver based on the presence of SUNDIALS and SuitrSparse components will mean it should be opt-in on installing from `pip` + using the SDist, not when building wheels for different platforms (where a wheel is going to be built with the SDist)...
> 
> In other things, I think we should look for more files related to the SUNDIALS installation such as those mentioned in #3461. The key reason I was advocating for a local SUNDIALS (with KLU) installation was because we can then point to a local folder `sundials_installed/` for the IDAKLU solver compilation, but if someone wants to use `scikits.odes` (on a non-developer but a user installation), they can install SUNDIALS (without SuiteSparse's KLU) to `.local/`, i.e., in a different place that doesn't overwrite any of the files. I will have to think about the entire compilation procedure in depth so that nothing breaks.
> 
> @AlessioBugetti I think these changes look good to me overall once they're done, could you mark the relevant URLs for the build-time components, their checksums, and the installation directories as constants at the top of the file for easier access?

Thank you so much for your feedback! Fun fact: I was actually in the middle of making this new commit when I received your comment.

AlessioBugetti:
In the latest commit, I've made modifications to the code to enhance the search functionality specifically for locating additional files related to the SUNDIALS and SuiteSparse installation as mentioned in #3461 

Commit [here](https://github.com/pybamm-team/PyBaMM/pull/3719/commits/4b8cf03626d8450f7b9f1de83547049ad99b70cc)

agriyakhetarpal:
The [`style` job failures](https://arc.net/l/quote/hdvonlmq) will have to be fixed so that the regular unit and integration tests can proceed to run

agriyakhetarpal:
The installation of `scikits.odes` is failing in CI due to this line: https://github.com/pybamm-team/PyBaMM/actions/runs/7515206009/job/20459589221#step:11:1012.

i.e., it is not able to find the location of the SUNDIALS installation. We set a `SUNDIALS_INST` environment variable for this purpose (see `noxfile.py` and the installation guide for `scikits.odes` in their documentation). I can't figure out what the exact change was that broke the script without having looked at it in depth, but I hope this tidbit helps with some debugging (please try installing `scikits.odes` locally through setting `SUNDIALS_INST` and `pip install scikits.odes` and maybe that can reveal more information)

AlessioBugetti:
> The installation of `scikits.odes` is failing in CI due to this line: https://github.com/pybamm-team/PyBaMM/actions/runs/7515206009/job/20459589221#step:11:1012.
> 
> i.e., it is not able to find the location of the SUNDIALS installation. We set a `SUNDIALS_INST` environment variable for this purpose (see `noxfile.py` and the installation guide for `scikits.odes` in their documentation). I can't figure out what the exact change was that broke the script without having looked at it in depth, but I hope this tidbit helps with some debugging (please try installing `scikits.odes` locally through setting `SUNDIALS_INST` and `pip install scikits.odes` and maybe that can reveal more information)
> 
> Edit: some of the other logs in the matrix show that PyBaMM is not able to find the SUNDIALS configuration as well, so it is definitely a problem with the changes introduced in the script. The macOS tests are passing, however.

I believe I have resolved the issue that was breaking the script. There was an indentation problem within the install_sundials function.

agriyakhetarpal (changes requested):
Thanks @AlessioBugetti for this overhaul of the script, there are many quality-of-life improvements here now and it works great overall. I have requested some changes below based on what I observed by testing the script locally.

In addition, we should remove the `wget` installation present in `noxfile.py` since it is not required anymore.

AlessioBugetti:
Thank you for your valuable feedback, I appreciate the time you took to review the work.

I have now implemented the suggested changes:

- Added the `--force` command-line argument to enable users to force the installation of SuiteSparse and SUNDIALS, even if they are already found in the specified installation directory.

- Resolved an issue where the `--install-dir` argument was not correctly respected. The script now accurately checks for existing libraries in the specified installation directory and creates the folder if it doesn't exist.

- Improved the logic for detecting the presence of SUNDIALS and SuiteSparse libraries.

- Cleaned up the script by removing unused variables.

- Added more descriptive print statements for each step and for individual file checks, providing better feedback and clarity on the script's execution process.

- Corrected the platform check from `linux` to `Linux`.

- Eliminated the unnecessary wget installation from `noxfile.py`, as it's no longer required by the updated script.

- Incorporated details about the `--force` and `--install-dir` flags in the "Installation from source" section of the user guide, providing clear instructions and examples for their usage.

Commit [here](https://github.com/pybamm-team/PyBaMM/pull/3719/commits/00f3269a344cbec391b5e4b488a2df869d6e0189)

scripts/install_KLU_Sundials.py
```diff
@@ -59,132 +269,49 @@ def download_extract_library(url, download_dir):
...
-sundials_version = "6.5.0"
-sundials_url = (
-    "https://github.com/LLNL/sundials/"
-    + f"releases/download/v{sundials_version}/sundials-{sundials_version}.tar.gz"
-)
+if args.force:
+    print(
+        "The '--force' option is activated: installation will be forced, ignoring any existing libraries."
+    )
+    sundials_found, suitesparse_found = False, False
```
agriyakhetarpal, line 292: I ran into an issue where:

1. I invoked the script with `python scripts/install_KLU_Sundials.py` and it proceeded to install the components as requested
2. I used `nox -s pybamm-requires -- --force` (with `session.posargs` enabled locally as mentioned above) and it returned a CMake cache error, asking me to re-run CMake from a different target directory.

IMO to avoid such accidental errors on re-runs of the script from re-runs of the script, we should delete these folders and the files under them, but only if trying to install with `--force`:

```
build_sundials/
SuiteSparse-6.0.3/
sundials-6.5.0/
```

because these are the extracted folders from the downloaded tarballs, and they might contain leftover CMake configuration files from a previous installation (especially SuiteSparse). If `--force` is not passed then I guess it makes sense to keep them

agriyakhetarpal (commented):
Thanks a lot for working on this in such earnest @AlessioBugetti, this is almost complete and ready to go, barring a few final changes to be addressed. I feel we can make the script a bit more robust.

I think there has been a lot done in this PR already and the script has been re-written majorly, so perhaps other improvements in #3651 should be tackled in follow-up PRs (though I am not sure about including all of the points at this time and some are subject to discussion).

AlessioBugetti:
I have just completed the requested modifications to the script. Now it seems more robust.

- In `noxfile.py`, retained `cmake` in the installation command ensuring compatibility for local development environments where `cmake` might not be pre-installed.
- Implemented handling for `--install-dir` and `--force` arguments in `nox` sessions. This allows users to specify custom installation directories and force installation if required. The command usage is now documented in the docstring for clarity.
- In `scripts/install_KLU_Sundials.py`:
  - Modified print statements to include the directory where libraries are found, making the output more informative.
  - Enhanced checksum validation messages to display both actual and expected checksums for better clarity and troubleshooting.
  - Addressed an issue with `--force` installation. The script now removes directories (`build_sundials`, `SuiteSparse-{SUITESPARSE_VERSION}`, `sundials-{SUNDIALS_VERSION}`) if `--force` is used. This ensures a clean state for forced re-installations, preventing CMake cache errors and conflicts with previous configurations.

Commit [here](https://github.com/pybamm-team/PyBaMM/pull/3719/commits/a7fed0e2f73c53f72f2cdb58837671632dc34486)

scripts/install_KLU_Sundials.py
```diff
@@ -15,28 +14,246 @@
...
+        if platform.processor() == "arm":
+            OpenMP_C_FLAGS = (
+                "-Xpreprocessor -fopenmp -I/opt/homebrew/opt/libomp/include"
+            )
+            OpenMP_C_LIB_NAMES = "omp"
+            OpenMP_omp_LIBRARY = "/opt/homebrew/opt/libomp/lib/libomp.dylib"
+        elif platform.processor() == "i386":
+            OpenMP_C_FLAGS = "-Xpreprocessor -fopenmp -I/usr/local/opt/libomp/include"
+            OpenMP_C_LIB_NAMES = "omp"
+            OpenMP_omp_LIBRARY = "/usr/local/opt/libomp/lib/libomp.dylib"
```
kratman, line 116: It is usually best to end an if-elif chain in an else statement. It lets other devs know nothing was missed.

scripts/install_KLU_Sundials.py
```diff
@@ -15,28 +14,246 @@
...
+        "libsundials_sunlinsolspbcgs",
+        "libsundials_sunlinsollapackdense",
+        "libsundials_sunmatrixsparse",
+        "libsundials_nvecserial",
+        "libsundials_nvecopenmp",
+    ]
+    if platform.system() == "Linux":
+        sundials_files = [file + ".so" for file in sundials_files]
+    elif platform.system() == "Darwin":
+        sundials_files = [file + ".dylib" for file in sundials_files]
```
kratman, line 157: Same as above

scripts/install_KLU_Sundials.py
```diff
@@ -59,132 +276,52 @@ def download_extract_library(url, download_dir):
...
+        install_suitesparse(download_dir)
+        install_sundials(download_dir, install_dir)
+    elif not sundials_found and suitesparse_found:
+        # Only SUNDIALS is missing, download and install it
+        parallel_download([(SUNDIALS_URL, SUNDIALS_CHECKSUM)], download_dir)
+        install_sundials(download_dir, install_dir)
+    elif sundials_found and not suitesparse_found:
+        # Only SuiteSparse is missing, download and install it
+        parallel_download([(SUITESPARSE_URL, SUITESPARSE_CHECKSUM)], download_dir)
+        install_suitesparse(download_dir)
```
kratman, line 327: Same as above

scripts/install_KLU_Sundials.py
kratman: I have not worked with this script at all, but to me it looks like those variables are not set for any architecture that is not arm or i386. So I would expect it to end with an error for other processors.

However, I do not know what the correct behavior should be if the hardware is neither arm nor i386. This is sort of why it is best to always end with an else, it helps other devs understand the intended behavior. Since it looks like MacOS, my guess is that it will be an error for unknown architecture if it is not one of those two.

scripts/install_KLU_Sundials.py
```diff
@@ -59,132 +282,55 @@ def download_extract_library(url, download_dir):
...
+    else:
+        # Both libraries are found and no force installation is requested
+        print("Both SUNDIALS and SuiteSparse libraries are already installed.")
```
kratman, line 336: For this case, you might be able to do
```suggestion
    else:
        if not sundials_found:
            parallel_download([(SUNDIALS_URL, SUNDIALS_CHECKSUM)], download_dir)
            install_sundials(download_dir, install_dir)
        if not suitesparse_found:
            parallel_download([(SUITESPARSE_URL, SUITESPARSE_CHECKSUM)], download_dir)
            install_suitesparse(download_dir)
```
This stops you from doing additional checks gainst the other libraries. A small optimization, but it makes the code much simpler.

kratman: @AlessioBugetti This should be my last finding

AlessioBugetti: Thank you for your feedback and observations!

Commit [here](https://github.com/pybamm-team/PyBaMM/pull/3719/commits/63a4ed3e7be0fc9039af2b6f0001c0ffdf70bb5e)

arjxn-py (approved):
Awesome work @AlessioBugetti, this one marks significant quality-of-life improvements with the `install_KLU_Sundials.py` script. 
I've tested this on WSL and it works like breeze. Some minor comments below -
Thanks a lot, Much Appreciated 🚀

arjxn-py:
Should we also do changelog for this? cc: @kratman @agriyakhetarpal?

docs/source/user_guide/installation/install-from-source.rst
```diff
@@ -81,6 +81,24 @@ If you are running windows, you can simply skip this section and jump to :ref:`p
...
+
+      nox -s pybamm-requires -- --install-dir [custom_directory_path]
+
+- ``--force``: Force the installation of SUNDIALS and SuiteSparse, even if they are already found in the specified directory.
+
+  Example:
+
+  .. code:: bash
+
+      nox -s pybamm-requires -- --force
```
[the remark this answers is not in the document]
agriyakhetarpal: They are intended in this case – any arguments passed to a `nox` session after `--` are [positional ones](https://nox.thea.codes/en/stable/config.html#passing-arguments-into-sessions) and are used by what is being run internally

agriyakhetarpal:
I would say we should skip the CHANGELOG for this. It's not entirely user-facing and the improvements cater to just those who want to build from source

AlessioBugetti:
> Awesome work @AlessioBugetti, this one marks significant quality-of-life improvements with the `install_KLU_Sundials.py` script. I've tested this on WSL and it works like breeze. Some minor comments below - Thanks a lot, Much Appreciated 🚀

Thank you for the positive feedback! I'm glad to hear that the improvements to the `install_KLU_Sundials.py` script are making a difference. Thanks again for your support! 🚀

Last commit [here](https://github.com/pybamm-team/PyBaMM/pull/3719/commits/59eb517eadc72619b2b3631c95ba47374d974561)

PR #3726, opened 2024-01-15
Fixed issue #3724

### LINKED ISSUE #3724: `pybamm.Experiment.read_termination`, arg `termination` not documented or given a type. The return value is also not documented

jonchapman1:
Documentation added for pybamm.Experiment.read_termination

# Description

Testing the workflow for participants of the developed training workshop in April 2024 to get used to contributing with simple fixes in advance of the workshop. I added documentation for pybamm.Experiment.read_termination

Fixes #3724

## Type of change

# Key checklist:

## Further checks:


tinosulzer (approved):

pybamm/experiment/experiment.py
```diff
@@ -141,6 +141,21 @@ def __repr__(self):
     def read_termination(self, termination):
         """
         Read the termination reason. If this condition is hit, the experiment will stop.
+
+        Parameters
+        ----------
+        termination : list[str], optional
+           A single string, or a list of strings, representing the conditions to terminate the experiment.
+           Only capacity or voltage can be provided as a termination reason.
+           e.g. '4 Ah capacity' or ['80% capacity', '2.5 V']
```
kratman, line 150: If we are going to say that the type is a ```list[str]``` then we should enforce that as well. The signature can be changed to ```def read_termination(self, termination: list[str]):``` and the check for the string being a single string can be removed.

kratman (changes requested):

pybamm/experiment/experiment.py
brosaplanella: I think at the moment we accept both a string and a list, which is probably not ideal but there quite a few other methods work like this and changing them at the moment is a big change.

For context, this PR from Jon was to test the pre-workshop exercise instructions we are giving attendees of our next Hackathon. To ensure they know how to use Github we ask them to add a docstring somewhere and open a PR (this way we improve documentation too).

I think probably the best solution for the moment would be to write
```
termination : str or list[str], optional
```
which reflects current use.

kratman: If this was only to update documentation that is fine, but I would say we should open a ticket for making it just a ```list[str]```. Long term it reduces code complexity to support only a single type of input for interfaces.

kratman: It is always good to clean up as we go, even for easy PRs

brosaplanella (approved):
Looks good to me, thanks Jon!

kratman (approved):
Removing my requested changes after discussion with @brosaplanella

I do think we should make a ticket to only support ```list[str]``` for the type though. It makes interfaces much nicer when there is only 1 input type and requires less error handling. For users this just requires putting [] around their single strings, so it is not a huge change.

rtimms:
@all-contributors please add @jonchapman1 for docs

PR #3737, opened 2024-01-16
Improve the release workflow

Saransh-cpp:
# Description

See https://github.com/pybamm-team/PyBaMM/pull/3736

A working PR - https://github.com/Saransh-cpp/PyBaMM/pull/17

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:


.github/workflows/update_version.yml
```diff
@@ -49,29 +51,49 @@ jobs:
...
+      # update all the version strings
+      - name: Update version
+        if: github.event_name == 'workflow_dispatch'
+        run: python scripts/update_version.py
+
+      # create a pull request updating versions in develop
+      - name: Create Pull Request
+        id: version_pr
+        uses: peter-evans/create-pull-request@v3
+        with:
```
Saransh-cpp, line 86: Now we don't "sync" the release and develop branches, instead we just update the versions in both the branches independently. This saves us from the conflicts and duplicating the cherry-picked commits.

scripts/update_version.py
```diff
@@ -17,7 +17,11 @@ def update_version():
     Opens file and updates the version number
     """
     release_version = os.getenv("VERSION")[1:]
-    last_day_of_month = date.today() + relativedelta(day=31)
+    release_date = (
+        date.today()
+        if "rc" in release_version
+        else date.today() + relativedelta(day=31)
+    )
```
Saransh-cpp, line 24: The final release will be at the end of the month, but the rc releases will be in between.

scripts/update_version.py
```diff
@@ -72,16 +76,21 @@ def update_version():
...
         output_list[0] = changelog_line1
+        # add a new heading for rc0 releases
         if "rc0" in release_version:
             output_list.insert(2, changelog_line2)
         else:
-            output_list[2] = changelog_line2
+            # for rcX and final releases, update the already existing rc
+            # release heading
+            for i in range(0, len(output_list)):
+                if re.search("[v]\d\d\.\drc\d", output_list[i]):
```
Saransh-cpp, line 92: For rc0 release, we just need to add a new heading under the "Unrelease" heading. But, for rcX and final releases, the workflow now finds the existing "rc" heading and replaces it with an updated version and date.

scripts/update_version.py
kratman: Not all months have 31 days

Saransh-cpp: This always returns the last day of the month. The current workflows use the same line and it can be seen that the headings in the CHANGELOG are correct for months that don't have 31 days.

arjxn-py (approved):
The further comments you made clears most of my doubts and as i've also noticed & observed #3736, this looks good to me.

PR #3740, opened 2024-01-17
Using bibtexparser instead of pybtex for bibtex.

### LINKED ISSUE #3648: Switch to an active citations parser dependency to replace `pybtex`
If `pybtex` is dead, should we switch to something else instead?

I found this active project for parsing bibtex: https://github.com/sciunto-org/python-bibtexparser

_Originally posted by @kratman in https://github.com/pybamm-team/PyBaMM/pull/3645#issuecomment-1864414897_

prady0t:
Fixes #3648 

Still in progress, Opened a PR so that it's easier to communicate changes and progress.


agriyakhetarpal:
Note: the `style` job has to pass in order for code-related tests to proceed

prady0t:
https://github.com/sciunto-org/python-bibtexparser/issues/433
Response here might be helpful.

prady0t:
@agriyakhetarpal I don't understand the aiing test case here : https://github.com/pybamm-team/PyBaMM/actions/runs/7627904153/job/20779262646?pr=3740#step:11:2390:~:text=AssertionError%3A%20ModuleNotFoundError%20not%20raised

This is the only failing test case. Can you help me with this?

agriyakhetarpal:
You haven't added `bibtexparser` to the `[cite]` set of optional dependencies in `pyproject.toml` yet, doing so should fix this error

pybamm/citations.py
```diff
@@ -216,9 +215,6 @@ def print(self, filename=None, output_format="text", verbose=False):
             Marquis2019 was cited due to the use of SPM
 
         """
-        # Parse citations that were not known keys at registration, but do not
-        # fail if they cannot be parsed
-        pybtex = have_optional_dependency("pybtex")
```
agriyakhetarpal, line 221: You have two options here:

1. Either use a different function in the `Citations` class that calls `pybamm.have_optional_dependency()` in `test_util`—see my other comment—instead of `pybamm.print_citations()` (which is really the same as `pybamm.Citations.print()`), or
2. Re-introduce these deleted lines and check for the `bibtexparser` optional dependency before proceeding to the functionality of the method.

I think the latter is better in this case, since it allows for consistent API behaviour across PyBaMM versions despite internal changes in dependencies as they are being done here.

prady0t: pybtex was used here to obtain `text` output format as : 
```py
pybtex.format_from_strings(
                self._cited, style="plain", output_backend="plaintext"
            )
```
due to the lack of such functionality in `bibtexparser` we don't really have a reason to import it here and is left unused. Hence the failing pre-commit.

agriyakhetarpal: We have to clear the style checks for the other PR checks to run. Is there a way you can use the unused variable?

prady0t:
I've added changes. Now only thing there is is changing output format. I might raise a PR in `bibtexparser` repo related to this which once merged should be easy to integrate. Meanwhile can we go with this output format?

pybamm/citations.py
```diff
@@ -150,23 +150,22 @@ def _parse_citation(self, key):
...
-            for key, entry in bib_data.entries.items():
+            for entry in bib_data.entries:
                 # Add to _all_citations dictionary
-                self._add_citation(key, entry)
+                self._add_citation(entry.key, entry)
                 # Add to _papers_to_cite set
-                self._papers_to_cite.add(key)
+                self._papers_to_cite.add(entry.key)
                 return
-        except PybtexError:
```
arjxn-py, line 168: Does `bibtexparser` also have some attribute corresponding to `PybtexError`?

pybamm/citations.py
```diff
@@ -230,37 +226,41 @@ def print(self, filename=None, output_format="text", verbose=False):
...
         else:
             with open(filename, "w") as f:
-                f.write(citations)
+                for entry in citations:
+                    f.write(self._string_formatting(entry))
+
+    def _string_formatting(self, entry):
+        bibtexparser = have_optional_dependency("bibtexparser")
+        if not isinstance(entry, bibtexparser.model.Entry):
+            raise TypeError()
```
arjxn-py, line 244: Coverage is failing as this is not covered by tests, would be nice to add some test case covering this part too.

tests/unit/test_util.py
```diff
@@ -94,10 +94,10 @@ def test_git_commit_info(self):
 
     def test_have_optional_dependency(self):
         with self.assertRaisesRegex(
-            ModuleNotFoundError, "Optional dependency pybtex is not available."
+            ModuleNotFoundError, "Optional dependency bibtexparser is not available."
         ):
-            pybtex = sys.modules["pybtex"]
-            sys.modules["pybtex"] = None
+            bibtexparser = sys.modules["bibtexparser"]
+            sys.modules["bibtexparser"] = None
```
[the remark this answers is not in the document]
arjxn-py: Yes, this is the case indeed. If you feel this should be changed to some other stable optional dependency then it can also be looked into.

arjxn-py (changes requested):
Thanks @prady0t, changes looks nice.
Some comments below -

tests/unit/test_util.py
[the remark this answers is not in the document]
arjxn-py: Yes, this is the case indeed. If you feel this should be changed to some other stable optional dependency then it can also be looked into.

pybamm/citations.py
agriyakhetarpal: I agree, and a helpful error message would be nice as well – a `TypeError` in itself is not descriptive enough

pybamm/citations.py
prady0t: No it doesn't. The only thing close to exception handling in bibtexparser is [this](https://bibtexparser.readthedocs.io/en/main/quickstart.html#:~:text=Step%202%3A%20Error%20Checking%C2%B6).

prady0t:
Why is this test case failing? https://github.com/pybamm-team/PyBaMM/actions/runs/7700868778/job/20989315835?pr=3740

agriyakhetarpal:
Unrelated error, I triggered a re-run which should fix it

agriyakhetarpal (changes requested):
Thanks @prady0t – I haven't looked at the change in functionality in detail as of now, but this PR should – in addition, perform the following changes:

1. Replace `pybtex` with `bibtexparser` in `asv.conf.json` and in the optional dependencies section/table of the installation guide
2. Replace instances of `pybtex` in `CONTRIBUTING.md` and comments in `noxfile.py`
3. Remove any remaining comments referencing `pybtex` in `pybamm/citations.py`

Also, this is the output from `pybamm.print_citations()`, currently:

```text
   {Array programming with NumPy} Harris, Charles R. and Millman, K. Jarrod and van der Walt, St{\'{e}}fan J. and Gommers, Ralf and Virtanen, Pauli and Cournapeau, David and Wieser, Eric and Taylor, Julian and Berg, Sebastian and Smith, Nathaniel J. and others Nature 585 7825 357--362 2020 Nature Publishing Group 10.1038/s41586-020-2649-2 

   {Python Battery Mathematical Modelling (PyBaMM)} Sulzer, Valentin and Marquis, Scott G. and Timms, Robert and Robinson, Martin and Chapman, S. Jon 10.5334/jors.309 Journal of Open Research Software Software Sustainability Institute 9 1 14 2021 
```

which doesn't suggest that the string parsing is working very well at this moment – I know it is unsupported functionality in `bibtexparser`, but we can't compromise on losing proper formatting (missing commas and periods, unneeded curly brackets, missing diacritics)  – it is essential for the NumPy and PyBaMM papers, and no less for any others. 

Another solution is to use something like [`diffrax`'s autocitation functionality](https://github.com/patrick-kidger/diffrax/blob/main/diffrax/_autocitation.py) that I had suggested before and hardcode citation strings inside `__init__` methods for model and solver classes, which would mean losing out on plaintext functionality, but brings us towards having clearer outputs in standard BibTeX – which is something I prefer. @brosaplanella and @kratman, what do you think about this?

pybamm/citations.py
```diff
@@ -178,10 +177,10 @@ def _tag_citations(self):
             for key, entry in self._citation_tags.items():
                 print(f"{key} was cited due to the use of {entry}")
 
-    def print(self, filename=None, output_format="text", verbose=False):
```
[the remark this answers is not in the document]
prady0t: Yes we can convert it into bibtex.

pybamm/citations.py
```diff
@@ -230,37 +226,43 @@ def print(self, filename=None, output_format="text", verbose=False):
...
-            citations = pybtex.format_from_strings(
-                self._cited, style="plain", output_backend="plaintext"
-            )
-        elif output_format == "bibtex":
-            citations = "\n".join(self._cited)
-        else:
-            raise pybamm.OptionError(
-                f"Output format {output_format} not recognised."
-                "It should be 'text' or 'bibtex'."
-            )
```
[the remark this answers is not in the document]
prady0t: One thing that we can do now is having an additional option like `filtered_output` along with `bibtex`. This option can print the essential hardcoded values (inside __init__) which might lead to data loss but we can provide much cleaner output if that's what user is more concerned about. Additionally we can also have a option `raw` which outputs the parsed bibtex as is (which is out only current output).

With this we can solve both problems. Not having data loss and also providing cleaner output.

pybamm/citations.py
```diff
@@ -230,37 +226,43 @@ def print(self, filename=None, output_format="text", verbose=False):
...
 
-def print_citations(filename=None, output_format="text", verbose=False):
+def print_citations(filename=None, verbose=False):
     """See :meth:`Citations.print`"""
     if citations._citation_err_msg is not None:
         raise ImportError(
-            f"Citations could not be registered. If you are on Google Colab - "
-            "pybtex does not work with Google Colab due to a known bug - "
-            "https://bitbucket.org/pybtex-devs/pybtex/issues/148/. "
+            f"Citations could not be registered."
```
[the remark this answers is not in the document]
prady0t: I'll try this out.

prady0t:
@agriyakhetarpal How should we move forward with this?

pybamm/citations.py
```diff
@@ -74,24 +72,25 @@ def read_citations(self):
         """Reads the citations in `pybamm.CITATIONS.bib`. Other works can be cited
         by passing a BibTeX citation to :meth:`register`.
         """
-        parse_file = import_optional_dependency("pybtex.database", "parse_file")
         citations_file = os.path.join(pybamm.root_dir(), "pybamm", "CITATIONS.bib")
-        bib_data = parse_file(citations_file, bib_format="bibtex")
-        for key, entry in bib_data.entries.items():
-            self._add_citation(key, entry)
+        parse_file = have_optional_dependency("bibtexparser", "parse_file")
```
arjxn-py, line 76:
```suggestion
        parse_file = import_optional_dependency("bibtexparser", "parse_file")
```

pybamm/citations.py
```diff
@@ -74,24 +72,25 @@ def read_citations(self):
...
+            self._add_citation(entry.key, entry)
 
     def _add_citation(self, key, entry):
         """Adds `entry` to `self._all_citations` under `key`, warning the user if a
         previous entry is overwritten
         """
 
-        Entry = import_optional_dependency("pybtex.database", "Entry")
         # Check input types are correct
+        Entry = have_optional_dependency("bibtexparser.model", "Entry")
```
arjxn-py, line 88:
```suggestion
        Entry = import_optional_dependency("bibtexparser.model", "Entry")
```

pybamm/citations.py
```diff
@@ -151,23 +150,22 @@ def _parse_citation(self, key):
...
             A BibTeX formatted citation
         """
-        PybtexError = import_optional_dependency("pybtex.scanner", "PybtexError")
-        parse_string = import_optional_dependency("pybtex.database", "parse_string")
+
+        # Parse string as a bibtex citation, and check that a citation was found
         try:
-            # Parse string as a bibtex citation, and check that a citation was found
-            bib_data = parse_string(key, bib_format="bibtex")
+            parse_string = have_optional_dependency("bibtexparser", "parse_string")
```
arjxn-py, line 156:
```suggestion
            parse_string = import_optional_dependency("bibtexparser", "parse_string")
```

pybamm/citations.py
```diff
@@ -231,37 +226,43 @@ def print(self, filename=None, output_format="text", verbose=False):
...
             if verbose:
                 self._tag_citations()  # pragma: no cover
         else:
             with open(filename, "w") as f:
-                f.write(citations)
+                for entry in citations:
+                    f.write(self._string_formatting(entry))
+
+    def _string_formatting(self, entry):
+        bibtexparser = have_optional_dependency("bibtexparser")
```
arjxn-py, line 242:
```suggestion
        bibtexparser = import_optional_dependency("bibtexparser")
```

tests/unit/test_util.py
```diff
@@ -118,8 +118,8 @@ def test_import_optional_dependency(self):
                 sym = pybamm.div(c * pybamm.grad(c)) + (c / d + c - d) ** 5
                 sym.visualise(test_name)
 
-        sys.modules["pybtex"] = pybtex
-        pybamm.util.import_optional_dependency("pybtex")
+        sys.modules["bibtexparser"] = bibtexparser
+        pybamm.util.have_optional_dependency("bibtexparser")
```
arjxn-py, line 122:
```suggestion
        pybamm.util.import_optional_dependency("bibtexparser")
```

arjxn-py (changes requested):
Fixing some conflicts raised with #3866

pybamm/citations.py
```diff
@@ -4,14 +4,12 @@
 # https://firedrakeproject.org/citing.html
 #
 import pybamm
-import os
 import warnings
+import os
 from sys import _getframe
```
arjxn-py, line 9:
```suggestion
from sys import _getframe
from pybamm.util import import_optional_dependency
```

arjxn-py (changes requested):

agriyakhetarpal:
> @agriyakhetarpal How should we move forward with this?

Tagging @Saransh-cpp @kratman since they were involved in the discussions where we discussed replacing `pybtex` earlier – we do not yet have functionality in `bibtexparser` to parse BibTeX in order to receive plaintext. @prady0t is able to parse that in a hacky way to retrieve all fields, join all of the strings and return a plaintext citation, but it doesn't conform to APA/MLA/Chicago style references (of course, doing that is a mighty task): please see [my review](https://github.com/pybamm-team/PyBaMM/pull/3740#pullrequestreview-1851852413) above for more.

P.S. to let go of the `setuptools` installation hack in the `nox` sessions that we had introduced earlier, I had realised that should be fixable with

```python
nox.options.default_venv_backend = "virtualenv"
```

i.e., just the one line in `noxfile.py`. 🙂 This is because `setuptools` is one of the seed packages that gets installed on virtual environment creation (along with `pip` and `wheel`) – this is for `virtualenv`. 
OTOH and fun fact: the `venv` from stdlib does not add seed packages at all, and post Python 3.12 it does not even bootstrap `setuptools` – just `pip` is included (but it does add both `setuptools` and `pip` for Python 3.11 and earlier versions).

Saransh-cpp:
Hmm, using `pybam.print_citations()` now works for me on Google Colab.

We could maybe create a feature request in`bibtexparser` and keep using `pybtex` for now?

agriyakhetarpal:
>Hmm, using pybam.print_citations() now works for me on Google Colab.

Nice!

> We could maybe create a feature request inbibtexparser and keep using pybtex for now?

Yes, @prady0t has created a feature request upstream earlier: https://github.com/sciunto-org/python-bibtexparser/issues/433

So, yes – we can keep using `pybtex` for now in this case, because the above request is still welcoming contributions (which may take an indefinite amount of time). We can add the `noxfile.py` change from https://github.com/pybamm-team/PyBaMM/pull/3740#issuecomment-1984108727 and remove the Google Colab warning from `pybamm/citations.py` now.

Thanks for contributing and trying this out, @prady0t – sorry for the mess that this causes 🙂 Let's hope we can have `pybtex` back to action in the coming years (it is still the best option for a citations parser out there). Could you please revisit this PR and make the above changes, while reverting the rest?

Saransh-cpp:
Or, you could close this PR without deleting the branch so that we have some working `bibtexparser` code on GitHub (given that "squash and merge" will get rid off the intermediate code and commit history). You could create a new PR for the changes listed by @agriyakhetarpal.

prady0t:
Glad to see progress here. I'll close this PR here and open a new one with suggested changes.

prady0t:
Do I have to replace :
https://github.com/pybamm-team/PyBaMM/blob/bc31cacb5e7fcd2d1d16dd6d58d63e2db4474a7b/noxfile.py#L8
with 
```
nox.options.default_venv_backend = "virtualenv"
```

agriyakhetarpal:
> Do I have to replace :
> 
> https://github.com/pybamm-team/PyBaMM/blob/bc31cacb5e7fcd2d1d16dd6d58d63e2db4474a7b/noxfile.py#L8
> 
> 
> with
> ```
> nox.options.default_venv_backend = "virtualenv"
> ```

We should keep both

PR #3750, opened 2024-01-20
Improve Getting Started notebooks

brosaplanella:
# Description

This PR aims to improve and simplify the Getting Started notebooks so they are a better resource to new users (I am assuming users could have very limited Python knowledge). It is still work in progress.

## Main changes

- Tutorial 1: minor text changes.
- Tutorial 2: minor text changes.
- Tutorial 3: removed the part where two simulations (DFN & SPM) got the voltage components plot together as it used some additional plotting functionality outside of PyBaMM. Also implemented the new API for `plot_voltage_components` which matches the functionality of `plot`.
- Tutorial 4: removed the drive cycle example, as I believe we should encourage users to use the experiment class for that. This also means we do not need to import `os` for this notebook.
- Tutorial 5: minor text changes plus extra detail added on the cycle and step definition.
- Tutorial 6: minor text changes.
- Tutorial 7: minor text changes plus list of notebooks with other examples.
- Tutorial 8: minor text changes, including the fact that the fast solver postprocess the solution and trims the interval past the cut-off voltage (so the discussion was wrong).
- Tutorial 9: minor text changes.
- Tutorial 10: I wanted to move it to Creating Models folder, but it seems a duplication of 3-negative-particle-problem.ipynb` so I deleted it.
- Tutorial 11: moved to Creating Models folder (which will be addressed at a later stage).

## Questions
- Tutorial 4: do we want to show how to define your own parameter set as a file? I would say no (or at least not in this notebook),


docs/source/examples/notebooks/getting_started/tutorial-4-setting-parameter-values.ipynb
```diff
@@ -377,121 +372,28 @@
...
+    "### Function parameters"
    ]
   },
   {
    "attachments": {},
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Alternatively, we can define the current to be an arbitrary function of time"
+    "Some parameters, like the current, sometimes need to be defined as a function (in this case a function of time). As an example, we now want to set the current to be sinusoidal. To do this, we need to define the relevant function and pass it as a parameter."
```
kratman, line 391: Can we reword this sentence: "Some parameters, like the current, sometimes need to be defined as a function (in this case a function of time)."

Between the some and sometimes, plus the parentheses, it seems like this got a bit unwieldy

kratman (changes requested):

kratman:
@brosaplanella I only found 1 thing, but I did not run locally or proof read 100% of the text. Overall I think it looks like an improvement

docs/source/examples/notebooks/getting_started/tutorial-4-setting-parameter-values.ipynb
kratman: Maybe: "Some parameters can be defined as a function. In this example we will make current a function of time."

docs/source/examples/notebooks/getting_started/tutorial-4-setting-parameter-values.ipynb
```diff
@@ -278,54 +308,19 @@
...
-    "## Change individual parameters"
-   ]
-  },
-  {
-   "attachments": {},
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
-    "We often want to quickly change a small number of parameter values to investigate how the behaviour or the battery changes. In such cases, we can change parameter values without having to leave the notebook or script you are working in. \n",
+    "## Change individual parameters\n",
```
rtimms, line 311: should we show `print_parameter_info` here to show which parameters can and cannot be functions?

docs/source/examples/notebooks/getting_started/tutorial-7-model-options.ipynb
```diff
@@ -19,8 +19,18 @@
...
+      "\u001b[1m[\u001b[0m\u001b[34;49mnotice\u001b[0m\u001b[1;39;49m]\u001b[0m\u001b[39;49m To update, run: \u001b[0m\u001b[32;49mpip install --upgrade pip\u001b[0m\n",
       "Note: you may need to restart the kernel to use updated packages.\n"
      ]
+    },
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "An NVIDIA GPU may be present on this machine, but a CUDA-enabled jaxlib is not installed. Falling back to cpu.\n"
+     ]
```
rtimms, line 33: I think we should use a lumped model here, not that it matters too much, but it's probably more common that people use the lumped model (x-full is just for a single-layer pouch)

rtimms (commented):
looks good, thanks, just a couple of minor comments

docs/source/examples/notebooks/getting_started/tutorial-4-setting-parameter-values.ipynb
brosaplanella: When dealing with Rob's comments I ended up rephrasing the paragraph so this is gone now.

brosaplanella:
Addressed the comments and fixed a small typo in the parameter argument of `current_with_time`.

kratman (approved):
Looks good to me

agriyakhetarpal:
I can leave a review later in the day today; adding a note to self that broken links should be checked across the documentation and that the links on https://pybamm.org/learn/ would have to be updated in accordance with these changes

kratman:
@brosaplanella Just noticed these in the doc tests. Missed it before
```
/home/runner/work/PyBaMM/PyBaMM/docs/source/user_guide/index.md:45: WARNING: toctree contains reference to nonexisting document 'source/examples/notebooks/getting_started/tutorial-10-creating-a-model'
/home/runner/work/PyBaMM/PyBaMM/docs/source/user_guide/index.md:45: WARNING: toctree contains reference to nonexisting document 'source/examples/notebooks/getting_started/tutorial-11-creating-a-submodel'
/home/runner/work/PyBaMM/PyBaMM/docs/source/examples/index.rst:10: WARNING: toctree contains reference to nonexisting document 'source/examples/notebooks/getting_started/tutorial-10-creating-a-model'
/home/runner/work/PyBaMM/PyBaMM/docs/source/examples/index.rst:10: WARNING: toctree contains reference to nonexisting document 'source/examples/notebooks/getting_started/tutorial-11-creating-a-submodel'
looking for now-outdated files... none found
pickling environment... done
checking consistency... /home/runner/work/PyBaMM/PyBaMM/docs/source/examples/notebooks/creating_models/7-creating-a-submodel.ipynb: WARNING: document isn't included in any toctree
```

kratman:
@brosaplanella Is this ready to go now?

docs/source/examples/notebooks/getting_started/tutorial-4-setting-parameter-values.ipynb
tinosulzer, line 1: could we show or explain how to create a ParameterValues object from scratch in this example?

tinosulzer (approved):
Looks great, thanks, would just suggest giving more info on adding your own parameter set in tutorial 4

docs/source/examples/notebooks/getting_started/tutorial-4-setting-parameter-values.ipynb
kratman: If we want to get this change merged, we could also tack this onto #3844

brosaplanella:
Should I add a short paragraph explaining that the option exists and pointing to the example, or did you mean a fully fleshed example?

tinosulzer:
Something that says you can create custom parameter values by passing in a dictionary, with an example including a couple of scalars and a function, e.g.
```python
def cubed(x):
    return x ** 3

parameter_values = pybamm.ParameterValues({
    "a": 1,
    "b": 2,
    "c": cubed,
})
```
+ explain that the function object itself should be passed in without being called

Then provide a link to the folder of `.py` files where people can find an example to copy for their own custom parameter sets

kratman:
@brosaplanella I will re-review this afternoon so we can get these updates in there

PR #3765, opened 2024-01-25
Fix error message for boolean operator on symbols

### LINKED ISSUE #2520: Boolean operator not defined for Symbols (can't use if-else statements when defining expressions)
### PyBaMM Version

22.9

### Python Version

3.9

### Describe the bug

I want to write a piecewise function , but there is an error:
     18 def Ds_p(sto,T):
---> 19     if T>293.15:
     20         if sto<0.8:
     21             diff=1.0*10**(-11)*(6-5*sto)
    659 def __bool__(self):
--> 660     raise NotImplementedError("Boolean operator not defined for Symbols.")

NotImplementedError: Boolean operator not defined for Symbols.

### Steps to Reproduce

def Ds_p(sto,T):
    if T>293.15:
        if sto<0.8:
            diff=1.0*10**(-11)*(6-5*sto)
        else:
            diff=2.3362*10**(-11)
    return diff
params["Positive electrode diffusivity [m2.s-1]"]=Ds_p

### Relevant log output

_No response_

HarshvirSandhu:
# Description

Update error message for Boolean operators on symbols

Fixes #2520


arjxn-py (approved):
Thanks @HarshvirSandhu, looks good. 
Will wait for one more approving review before merging.

arjxn-py:
Would you also like to implement this in `binary_operators.py` to fullfill the scope of the issue?
```python
def if_else(condition, value_if_true, value_if_false):
    return value_if_true * condition + value_if_false * (1 - condition)
```

kratman:
> Would you also like to implement this in `binary_operators.py` to fullfill the scope of the issue?
> 
> ```python
> def if_else(condition, value_if_true, value_if_false):
>     return value_if_true * condition + value_if_false * (1 - condition)
> ```

I don't think we need an extra function for the if-else case. Plus that name would probably be confusing

kratman (approved):

agriyakhetarpal:
@all-contributors please add @HarshvirSandhu for code

Issue #2553, opened 2022-12-14
Cell-level State of Charge


tinosulzer:
We do not currently provide "State of Charge" as a variable, because it's not clear how to calculate this at the cell level:
- Discharged capacity divided by nominal cell capacity:
  - easy to calculate from existing parameters
  - what is the initial soc?
  - what if nominal cell capacity is wrong (i.e. not equal to the amount of charge that can be passed between voltage limits)?
- Calculation based on min/max electrode stoichiometry and nominal cell capacity
  - requires specifying "Stoichiometry at 100% SOC" in each electrode instead of "Initial concentration"
  - SOC can be calculated as e.g. `(x - x_100)`
     - what if nominal cell capacity is wrong? 
- Calculation based on electrode SOH
   - requires a pre-calculation step to calculate electrode SOH before solving the model
   - after that

## Proposed solution

1. Add a "Calculate SOC" model option which can be "true" or "false"
2. If true, add simple SOC submodel based , and perform `eSOH` calculation in `ParameterValues` class to define the parameters for the SOC submodel

Caveat: this will be wrong as the cell degrades. We can figure out updating parameter values later

tinosulzer:
This will finally enable the much-requested feature "Charge until 98% SOC"

brosaplanella:
Nice! To add to this, we typically calculate SoC as discharged capacity / RPT calculated capacity (or one minus that, to be precise). This way you can update it as the battery ages, but it gets trickier to calculate as it requires a separate simulation.

tinosulzer:
I think all of the following are equivalent:
1. `soc = initial_soc - Q_Ah / Q`
2. `soc = (x-x0)/(x100-x0)`
3. `soc = (y-y0)/(y100-y0)` (both numerator and denominator are negative)

So any way we do it we have to pre-calculate something (`initial_soc`/`Q`/`x0`/`x100`/`y0`/`y100`). The bad news is we can't trust the value the user enters since it might be incompatible with the voltage limits and hence lead to SOC<0 or SOC>1. The good news is that the electrode SOH model can do the calculation for us in a way that guarantees that the SOC is always in [0,1] given the parameter values and voltage limits.

> it gets trickier to calculate as it requires a separate simulation

The capacity calculated by the "electrode SOH" model should give the same answer as the RPT without requiring a whole simulation (just solving an algebraic model) but let me know if you have a case where it doesn't

brosaplanella:
[...]

PR #3771, opened 2024-01-27
Pe degrade

mzzhuo:
# Description

Adds a positive-electrode degradation model based on layered → spinel/rocksalt phase transition (the "core-shell" model) for high-nickel cathodes such as NMC811. Activated by a new option `"PE degradation": "phase transition"` on SPM and DFN. Tracks lithium diffusion in the core, oxygen diffusion in the shell, and a moving core-shell phase boundary driven by an interfacial chemical reaction.

Reference:
> Mingzhao Zhuo, Gregory Offer, Monica Marinescu. *Degradation model of high-nickel positive electrodes: Effects of loss of active material and cyclable lithium on capacity fade*. Journal of Power Sources, 556 (2023): 232461. [doi:10.1016/j.jpowsour.2022.232461](https://doi.org/10.1016/j.jpowsour.2022.232461)

Includes:
- New submodel package `pybamm.pe_degradation.phase_transition_core_shell` (`PhaseTransition`, `BasePhaseTransition`, `TotalConcentration`)
- New `Zhuo2023` parameter set (NMC811 + SiC chemistry)
- New `"positive core"` / `"positive shell"` domains; small extensions to `RAverage` / `XAverage` and the broadcast domain check
- New finite-volume operators `BoundaryCellValue` and `BoundaryCellLength` used by the submodel to enforce Robin-style boundary conditions at the moving phase boundary (operators are general-purpose and unit-tested in `test_unary_operators.py`)
- A small extension hook `_get_pe_shell_potential_drop` on `BaseInverseKinetics` for the SPM/SPMe inverse-Butler-Volmer path; the DFN forward-BV path adds the equivalent term inline in `base_kinetics.py`
- Example script `examples/scripts/run_pe_degradation.py`

Not in scope, **explicitly rejected at runtime via `pybamm.OptionError`**: two-phase positive electrode + PE phase transition (`particle phases: "2"` or per-electrode tuple `("_", "2")`); particle-size distribution + PE phase transition (which also covers MPM, since MPM forces `"particle size": "distribution"`).

## Type of change

Feature. Changelog entry added.

# Important checks

- [x] No style issues: `pre-commit run` passes on all touched files
- [x] All tests pass: targeted unit + integration tests pass locally; submodel-level unit tests, well-posedness on SPM, DFN and NewmanTobias with the option enabled, and an end-to-end `StandardModelTest` for both SPM and DFN. **Patch coverage 100%** (measured locally via `diff-cover` against `develop`).
- [ ] The documentation builds: `nox -s doctests` reports 3 pre-existing failures on `develop` (`Serialise.load_custom_model`, `entry_points.Model`, `entry_points.EntryPoint`) — unrelated to this PR. Worth a separate issue.
- [x] Tests added that prove the feature works


kratman:
@mzzhuo Are you still working on this?

mzzhuo:
> @mzzhuo Are you still working on this?

From my side, the submodel is done, not working on it now but waiting for the feedback from pybamm team. I think Feran and Rob are reviewing the code

kratman:
@mzzhuo I was primarily asking since the CI has been failing for this PR. Can you update the branch to resolve some of the issues? That needs to be done before it is merged anyway

pybamm/expression_tree/unary_operators.py
```diff
@@ -1014,6 +1014,46 @@ def _sympy_operator(self, child):
             return sympy.Symbol(latex_child)
 
 
+class BoundaryCellValue(BoundaryOperator):
```
rtimms, line 1017: It seems odd to me that this is part of `unary_operators` rather than being part of a spatial method. can we talk more about how the problem is posed for the moving boundary?

pybamm/geometry/battery_geometry.py
```diff
@@ -70,6 +70,39 @@ def battery_geometry(
                         }
                     )
 
+    if options is not None and options["PE degradation"] == "phase transition":
+        geo_domain = geo.domain_params["positive"]
+        phases = int(getattr(options, domain)["particle phases"])
+        if phases == 1:
+            geometry.update(
+                {
+                    "positive core": {"r_co": {"min": 0, "max": geo_domain.prim.R_typ}},
```
rtimms, line 79: I was expecting to have the core/shell be solved on [0, 1] and then have transformed PDEs that solve for the lengths to handle the moving boundary. Can you point to the paper or some notes for the implementation?

pybamm/geometry/standard_spatial_vars.py
```diff
@@ -182,3 +182,65 @@
...
+    coord_sys="spherical polar",
+)
+r_sh = pybamm.SpatialVariable(
+    "r_sh",
+    domain=["positive shell"],
+    auxiliary_domains={
+        "secondary": "positive electrode",
+        "tertiary": "current collector",
+    },
+    # note the coordinate system is different from that of r_co
```
rtimms, line 207: if you're doing a transformation anyway, why not pose both eqns on a catersian domain [0, 1], then do the mapping later? assume this is to make plotting easier later so everything lives on [0, R]?

pybamm/models/base_model.py
```diff
@@ -102,7 +102,7 @@ def __init__(self, name="Unnamed model"):
         self._rhs = {}
         self._algebraic = {}
         self._initial_conditions = {}
-        self._boundary_conditions = {}
+        self._boundary_conditions = BoundaryConditionsDict({})
```
rtimms, line 105: don't think this is needed

pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -146,6 +146,12 @@ class BatteryModelOptions(pybamm.FuzzyDict):
                 Sets the model to include a single active particle size or a
                 distribution of sizes at any macroscale location. Can be "single"
                 (default) or "distribution". Option applies to both electrodes.
+            * "PE degradation": str
+                Set the PE degradation submodel (core-shell) to be used. Options are:
```
rtimms, line 150: can you write "Positive Electrode" in these docs so people know what "PE" means

pybamm/models/full_battery_models/lithium_ion/base_lithium_ion_model.py
```diff
@@ -254,6 +281,40 @@ def set_open_circuit_potential_submodel(self):
                     self.param, domain, reaction, self.options, phase
                 )
 
+    def set_pe_degradation_submodel(self):
+        # specify the degradation domain
+        domain = "positive"
+
+        # replace the fickian particle submodel
+        # make sure set_pe_degradation_submodel comes after
```
rtimms, line 289: there is already code that loops over the submodels in a way that the order doesn't matter. does it fail in this case? if so, it would be good to see if we can fix that part of the code rather than having a special case.

pybamm/models/full_battery_models/lithium_ion/base_lithium_ion_model.py
```diff
@@ -254,6 +281,40 @@ def set_open_circuit_potential_submodel(self):
...
+        if self.options["PE degradation"] == "phase transition":
+            for phase in self.options.phases[domain]:
+                par_submod = f"{domain} {phase} particle"
+                if par_submod not in self.submodels:
+                    raise pybamm.ModelError(
+                        "The particle submodel has not been called yet."
+                        "Make sure it is invoked before "
+                        "calling phase transition submodel."
+                    )
+                elif not isinstance(
```
rtimms, line 300: can this be caught earlier as as options check?

pybamm/models/submodels/external_circuit/base_external_circuit.py
```diff
@@ -20,6 +20,8 @@ def get_fundamental_variables(self):
         variables = {
             "Discharge capacity [A.h]": Q_Ah,
             "Throughput capacity [A.h]": Qt_Ah,
+            # initial SoC is 0 as initial Q_Ah is set null
+            "Cell SoC": -Q_Ah / self.param.Q,
```
rtimms, line 24: I don't think we should define cell SoC as everyone has their own definition (see https://github.com/pybamm-team/PyBaMM/issues/2553)

pybamm/models/submodels/pe_degradation/base_phase_transition.py
```diff
@@ -0,0 +1,265 @@
...
+           doi: 10.1016/j.jpowsour.2022.232461.
+    """
+
+    def __init__(self, param, domain, options, phase="primary"):
+        super().__init__(param, domain, options=options, phase=phase)
+
+        # Read from options to see if we have a particle size distribution
+        domain_options = getattr(self.options, domain)
+        self.size_distribution = domain_options["particle size"] == "distribution"
+        if self.size_distribution is True:
```
rtimms, line 43: can this be caught earlier as an option check?

pybamm/models/submodels/pe_degradation/base_phase_transition.py
```diff
@@ -0,0 +1,265 @@
...
+            # For submodel 'positive primary open-circuit potential'
+            # f"{Domain} {phase_name}core surface lithium stoichiometry"
+            f"{Domain} {phase_name}particle surface stoichiometry" "": c_c_surf
+            / c_scale,
+            # ---------------------------------------------------------------------
+        }
+
+        return variables
+
+    def _r_average_core(self, symbol):
```
rtimms, line 221: again, I think this kind of thing should be handled by the spatial methods code, rather than explicitly as part of the model

rtimms (commented):
Thanks for working on this. @mzzhuo can you point to implementation details? The way this has been set up is different from what I was anticipating, so I'm getting a little lost. It seems like the model and spatial implementation are a little tangled here, and I think it would be great to have the general moving boundary stuff be handled by spatial methods and just have the PDEs in the model.

pybamm/expression_tree/unary_operators.py
rtimms: Ok, I see why this might be here. It's essentially `BoundaryValue` but taking cell centre rather than edge. I still kind of feel like this is a concern of spatial methods. Let's talk more.

pybamm/models/submodels/pe_degradation/base_phase_transition.py
mzzhuo: Hi Rob, thanks for reviewing the codes. Regarding your main concern, the PE degradation submodel consists of two diffusion equations residing in a shrinking core and expanding shell, and a set of ODEs describing the moving of the phase boundary. The first step is to do variable change so that the computational domains (core and shell) are fixed, this will naturally change the governing diffusion equations by adding extra terms associated with the moving boundary, which is an unknown variable. The extra terms are caused by both the original temporal differential term and spatial differential terms, so it does not make too much sense to incorporate into the discretization method. Another important reason for me doing the current way is to minimize the modification to current pybamm, so I leave the first step to manual calculation. To be honest, I have no idea whether this can be incorporated into the discretized method, or if yes, how much cost it will take. Regrading the [0, 1] issues, that is how I did in the initial version, now to be consistent with current pybamm release, I scaled both up with a constant R_typ. I share the paper and the formula for implementation: https://doi.org/10.1016/j.jpowsour.2022.232461 and https://hkustconnect-my.sharepoint.com/:b:/g/personal/mzhuo_connect_ust_hk/EYmFXnyOpP9KlUsrHg_0RtIBqQoIkN6qIpVCKu262uSgLA?e=EH0SJX

pybamm/models/submodels/pe_degradation/base_phase_transition.py
DrSOKane: Agreed

pybamm/models/full_battery_models/base_battery_model.py
DrSOKane: We need to agree on a name for this option, for compatibility for a planned acid dissolution model

pybamm/expression_tree/broadcasts.py
```diff
@@ -136,23 +136,25 @@ def check_and_set_domains(self, child: pybamm.Symbol, broadcast_domain: list[str
                 "positive electrode",
             ]
             or "particle" in broadcast_domain[0]
+            or "core" in broadcast_domain[0]
+            or "shell" in broadcast_domain[0]
         ):
             raise pybamm.DomainError(
                 """Primary broadcast from current collector domain must be to electrode
-                or separator or particle or particle size domains"""
+                or separator or particle (core or shell) or particle size domains"""
```
DrSOKane, line 144: Not all particles have core/shell though

pybamm/expression_tree/broadcasts.py
```diff
@@ -162,8 +164,17 @@ def check_and_set_domains(self, child: pybamm.Symbol, broadcast_domain: list[str
                 """Primary broadcast from particle size domain must be to particle
                 domain"""
             )
-        elif child.domain[0] in ["negative particle", "positive particle"]:
-            raise pybamm.DomainError("Cannot do primary broadcast from particle domain")
+        elif child.domain[0] in [
+            "negative particle",
+            "positive particle",
+            "negative core",
```
DrSOKane, line 170: There is no provision for core/shell in the negative electrode, but this might be worth keeping for future-proofing

pybamm/models/submodels/pe_degradation/base_phase_transition.py
```diff
@@ -0,0 +1,265 @@
...
+# phase transition cased degradation, based on paper
+#    Mingzhao Zhuo, Gregory Offer, Monica Marinescu, "Degradation model of
+#    high-nickel positive electrodes: Effects of loss of active material and
+#    cyclable lithium on capacity fade", Journal of Power Sources,
+#    556 (2023): 232461. doi: 10.1016/j.jpowsour.2022.232461.
+#
+import pybamm
+
+
+class BasePhaseTransition(pybamm.BaseSubModel):
```
DrSOKane, line 12: This name will cause problems if anyone wants to add a different positive electrode degradation model, such as acid dissolution. Please change to something like `BasePositiveDegradation` or `BasePositiveElectrodeDegradation`.

DrSOKane (commented):
Hi MZ, thank you so much for writing this code, which must have taken a huge amount of time! Robert Timms and I made some comments in specific areas of the code, but I also have some general comments that must be addressed before we can proceed any further.

Firstly, the model is still incompatible with the particle mechanics submodel as things stand. You can either accept this and add code to `base_battery_model` that prevents users from trying to use both, or alter the code for particle mechanics so that they are compatible.

Secondly, I don't understand why you need a separate set of "cyclable lithium" variables. If the lithium in the shell phase is dead with no posibility of recovery, then all you need to do is arange it so PyBaMM does not count the lithium in the shell when calculating `"Total lithium in particle [mol]"`.

That said, both Ghosh _et al._ (2021) and Zhuo _et al._ (2023) are rigorous pieces of research that deserve to be included in PyBaMM. I also like how you've added a `"positive electrode degradation"` option, so that more models like yours can be added in the future.

mzzhuo:
> Hi MZ, thank you so much for writing this code, which must have taken a huge amount of time! Robert Timms and I made some comments in specific areas of the code, but I also have some general comments that must be addressed before we can proceed any further.
> 
> Firstly, the model is still incompatible with the particle mechanics submodel as things stand. You can either accept this and add code to `base_battery_model` that prevents users from trying to use both, or alter the code for particle mechanics so that they are compatible.
> 
> Secondly, I don't understand why you need a separate set of "cyclable lithium" variables. If the lithium in the shell phase is dead with no posibility of recovery, then all you need to do is arange it so PyBaMM does not count the lithium in the shell when calculating `"Total lithium in particle [mol]"`.
> 
> That said, both Ghosh _et al._ (2021) and Zhuo _et al._ (2023) are rigorous pieces of research that deserve to be included in PyBaMM. I also like how you've added a `"positive electrode degradation"` option, so that more models like yours can be added in the future.

Hi Simone, thank you so much for your reviewing. 1) yes, incompatible at the moment, we can prevent the use of both submodels now for simplicity. 2) what you said about "Total lithium" is what was defined and calculated in the paper with LLI_tot, see Eq. 16 and Fig. 4. I guess you wonder why we bother to define an extra variable called cyclable lithium. OK, when we were discussing the degradation model, Greg had an assumption based on mass conservation at the interface, that is, the total lithium may increase and not always decrease. This scenario is only possible with the definition of cyclable lithium. The results in Fig. 4 with negative LLI_cyc shows that the model (particularly the mass conservation at the interface) and its implementation are validated, this is what we should expect. Besides this, LLI_total is always positive, i.e., we always lose lithium, but this info. is not that relevant to lithium shutting between the two electrodes. The stochiometry change from 0 to 100% corresponds to the cyclable lithium shutting. Note that in Figs. 5, 7 and 9 the results are all about cyclable lithium so that we can see the subtle difference between different scenarios.

pybamm/models/submodels/pe_degradation/base_phase_transition.py
mzzhuo: Good point. How about create a subfolder called "phase transition" inside the pe_degradation folder, and other subfolder like "acid dissolution" etc. pe_degradation is general and can include other non-core-shell submodels.

kratman:
@mzzhuo Are you still working on this PR?

mzzhuo:
> @mzzhuo Are you still working on this PR?

Hi Eric, Thanks for asking. Actually, I am waiting for the pybamm-team to do a thorough review, and I messaged Rob and Simon via Slack for further meetings to understand the model (as can be seen from Rob's comments) but got no response. I can update the branch and complete all the unit tests to pass all checks; meanwhile, could you make a plan to complete the review? FYI, a deep review of the maths and implementation behind the scene (double check) is good for science but will cost a lot of time and energy, in which case a meeting with myself and consultation to the pdf documentation (already uploaded) are helpful (or necessary) to the reviewer; a shallow review of the codes and running of the example notebook is way faster and I recommend this way to proceed.

kratman:
@mzzhuo You will need to update the branch for the review to continue. You have quite a lot of conflicts at the moment

rtimms:
@mzzhuo can you fix the conflicts and respond to the comments above? then we can meet to go through any outstanding issues

mzzhuo:
> @mzzhuo can you fix the conflicts and respond to the comments above? then we can meet to go through any outstanding issues

mzzhuo:
> @mzzhuo can you fix the conflicts and respond to the comments above? then we can meet to go through any outstanding issues

hi @rtimms I have pulled all the changes from pybamm-team develop branch and accordingly update the core-shell model, fixing all conflicts and successfully running 'examples/scripts/run_pe_degradation.py' scripts. Next, I will complete the unit tests and respond to other comments from you and Simon.

kratman:
@mzzhuo Are you still working on this PR?

mzzhuo:
> @mzzhuo Are you still working on this PR?
I am done with pulling the latest updates and resolved the conflict, pre-commit checks and nox documentation run. Now replying Rob/Simon's all comments

pybamm/expression_tree/unary_operators.py
mzzhuo: The moving-boundary problem consists of 1) transforming the changing computational domain to fixed domain and 2) sovling PDEs on fixed domain. Now the first step is done manually by variable change and only the second step implemented into PyBaMM. The spatial methods are suppoesd to do spatial discretization, and it may not be easy to handle the changing computational domain issue.

pybamm/geometry/battery_geometry.py
mzzhuo: With the previous PyBaMM versions, we did solve within [0, 1], and then with the lastest version using dimensional equations, I changed to [0, R] for consistency. here is the link to the equations and their scaling:
https://hkustconnect-my.sharepoint.com/:b:/g/personal/mzhuo_connect_ust_hk/EYmFXnyOpP9KlUsrHg_0RtIB8SksdPP6KAZiP9BfnAfafw?e=rVpY58

pybamm/geometry/standard_spatial_vars.py
mzzhuo: yes for plotting, and also trying to make them consistent with the rest sub-models and equations

pybamm/models/base_model.py
mzzhuo: I think it is gone now (the reason was a failing scenario when assigning bcs)

pybamm/models/full_battery_models/base_battery_model.py
mzzhuo: added "Positive Electrode" in relevant locations. @DrSOKane I put the phase-transition model in another subfolder so that other planned models can be host in parallel and options can be added.

pybamm/models/full_battery_models/lithium_ion/base_lithium_ion_model.py
mzzhuo: the pe_degradation_submodel shares the same key value in submodel dict with PE particle model---they can not coexist as pe fickian is included in the core part. If we want to remove the constraint of 'set pe degradation model' coming after 'set particle model', we need to modify the method 'set_particle_submodel' in both DFN, SPMe etc. To make things safe, I chose this simpler way and would see the pybamm team to further improve.

pybamm/models/full_battery_models/lithium_ion/base_lithium_ion_model.py
mzzhuo: thought of this, the option conflict check between "PE degradation" and "particle" may eliminate the cases that non-fickian submodels used for positive electrode, but at the same time undesirably restricted the choice for anode. I do not think option check can resolve the sequence issue here; maybe you got a better idea, would love to hear from you

pybamm/models/submodels/external_circuit/base_external_circuit.py
mzzhuo: agree! this is used in the paper for plotting e.g. Fig 6. I did not find similar definition in PyBaMM so made my own. You can remove it and find a decent way to handle this.

pybamm/models/submodels/pe_degradation/base_phase_transition.py
mzzhuo: moved to 'base_battery_model' for earlier option check

pybamm/models/submodels/pe_degradation/base_phase_transition.py
mzzhuo: Still I do not thin the spatial method, which in charge of spatial discretization, should take the job of computational domain change; we may be able to encode this manual step into pybamm, but that is out of my capability at the moment.

pybamm/expression_tree/broadcasts.py
mzzhuo: agree, just added in the core/shell scenario

pybamm/expression_tree/variable.py
```diff
@@ -150,7 +150,7 @@ def to_json(
 
 class Variable(VariableBase):
     """
-    A node in the expression tree represending a dependent variable.
+    A node in the expression tree represending an independent variable.
```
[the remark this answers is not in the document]
mzzhuo: done with undoing, but I still do not understand why it is a dependent variable, because it is going to be solved like concentration

mzzhuo:
> > @mzzhuo can you fix the conflicts and respond to the comments above? then we can meet to go through any outstanding issues
> 
> hi @rtimms I have pulled all the changes from pybamm-team develop branch and accordingly update the core-shell model, fixing all conflicts and successfully running 'examples/scripts/run_pe_degradation.py' scripts. Next, I will complete the unit tests and respond to other comments from you and Simon.

Hi @rtimms 1) I checked the unit tests and found none for ficikan diffuion submodels, and thus did not draft one for pe degradation (also need instructions from pybamm team; otherwise it takes too much time for me to learn from the source codebase). Added a script example to reproduce the results in my paper. 2) I went through all the comments from you and simon and replied. If any one missed, pls tag me again. 3) pre-commit checks done and nox documentation done, now I would hand this work over to you.

src/pybamm/models/submodels/external_circuit/discharge_throughput.py
```diff
@@ -18,6 +18,8 @@ def get_fundamental_variables(self):
         variables = {
             "Discharge capacity [A.h]": Q_Ah,
             "Throughput capacity [A.h]": Qt_Ah,
+            # initial SoC is 0 as initial Q_Ah is set null
+            "Cell SoC": -Q_Ah / self.param.Q,
```
DrSOKane, line 22: What would happen if this was removed?

DrSOKane, line 22: Most parameter sets in PyBaMM start from the fully charged condition - not the fully discharged condition - because the negative electrode OCP has a very steep gradient for the fully discharged condition, making solution harder. So this definition of cell SoC will not work in most cases. Please remove it.

kratman:
@mzzhuo Your tests are still failing

DrSOKane:
> @mzzhuo Your tests are still failing

@mzzhuo You need to alter the strings in two of the tests. In `tests/unit/test_models/test_full_battery_models/test_base_battery model.py`, you need to insert an extra line between lines 43 and 44: `'PE degradation': 'none' (possible: ['none', 'phase transition'])`. In `tests/unit/test_expression_tree/test_broadcasts.py`, you need to modify line 75 to read, `"Cannot do primary broadcast from particle (core or shell) domain"`.

src/pybamm/models/submodels/external_circuit/discharge_throughput.py
mzzhuo: removed the cell soc definition. Just not understand why it will make solution harder, because Q_Ah is already defined above and self.param.Q is a parameter, and thus it is just a post-processed variable?

DrSOKane:
> > > @mzzhuo Your tests are still failing
> > 
> > 
> > @mzzhuo You need to alter the strings in two of the tests. In `tests/unit/test_models/test_full_battery_models/test_base_battery model.py`, you need to insert an extra line between lines 43 and 44: `'PE degradation': 'none' (possible: ['none', 'phase transition'])`. In `tests/unit/test_expression_tree/test_broadcasts.py`, you need to modify line 75 to read, `"Cannot do primary broadcast from particle (core or shell) domain"`.
> 
> thanks @DrSOKane for identifying these failed tests, now fixed, but I saw some other failed tests that might not be relevant to my work.

The example script you wrote is failing because it reference `"Cell SoC"`, which has been removed. I'm not sure why the doctests are failing. Adding a test file for the `"Zhuo2023"` parameters will fix most of the coverage errors, and the remaining ones could be fixed in a future PR, for the sake of getting this important feature added.

src/pybamm/models/submodels/external_circuit/discharge_throughput.py
DrSOKane: Sorry MZ, I used the wrong terminology. Solving the PDEs when the one of more of the OCP curves is steep is fine, but it is a lot harder to find the correct initial conditions. It can be done - as you have shown - but even a small error in the initial conditions means the model either does not converge at all, or converges on an incorrect solution.

DrSOKane:
> > > > > @mzzhuo Your tests are still failing
> > > > 
> > > > 
> > > > @mzzhuo You need to alter the strings in two of the tests. In `tests/unit/test_models/test_full_battery_models/test_base_battery model.py`, you need to insert an extra line between lines 43 and 44: `'PE degradation': 'none' (possible: ['none', 'phase transition'])`. In `tests/unit/test_expression_tree/test_broadcasts.py`, you need to modify line 75 to read, `"Cannot do primary broadcast from particle (core or shell) domain"`.
> > > 
> > > 
> > > thanks @DrSOKane for identifying these failed tests, now fixed, but I saw some other failed tests that might not be relevant to my work.
> > 
> > 
> > The example script you wrote is failing because it reference `"Cell SoC"`, which has been removed. I'm not sure why the doctests are failing. Adding a test file for the `"Zhuo2023"` parameters will fix most of the coverage errors, and the remaining ones could be fixed in a future PR, for the sake of getting this important feature added.
> 
> Thanks @DrSOKane I do not quite understand the first two checks---codecover/patch and project, could u elaborate a bit and provide a more detailed guideline to fix them? Regarding the rest, what do you suggest to do?

Codecov checks for lines of code that have been covered by the automated tests. There is no test for `Zhuo2023.py`, so none of that is covered. The changes you made to the operators are not covered either.

Try fixing the example script first. Tests are less critical;  someone else can do those in a future PR.

But the vast majority of tests are now passing! I am so proud of you!

mzzhuo:
> > > > > > @mzzhuo Your tests are still failing
> > > > > 
> > > > > 
> > > > > @mzzhuo You need to alter the strings in two of the tests. In `tests/unit/test_models/test_full_battery_models/test_base_battery model.py`, you need to insert an extra line between lines 43 and 44: `'PE degradation': 'none' (possible: ['none', 'phase transition'])`. In `tests/unit/test_expression_tree/test_broadcasts.py`, you need to modify line 75 to read, `"Cannot do primary broadcast from particle (core or shell) domain"`.
> > > > 
> > > > 
> > > > thanks @DrSOKane for identifying these failed tests, now fixed, but I saw some other failed tests that might not be relevant to my work.
> > > 
> > > 
> > > The example script you wrote is failing because it reference `"Cell SoC"`, which has been removed. I'm not sure why the doctests are failing. Adding a test file for the `"Zhuo2023"` parameters will fix most of the coverage errors, and the remaining ones could be fixed in a future PR, for the sake of getting this important feature added.
> > 
> > 
> > Thanks @DrSOKane I do not quite understand the first two checks---codecover/patch and project, could u elaborate a bit and provide a more detailed guideline to fix them? Regarding the rest, what do you suggest to do?
> 
> Codecov checks for lines of code that have been covered by the automated tests. There is no test for `Zhuo2023.py`, so none of that is covered. The changes you made to the operators are not covered either.
> 
> Try fixing the example script first. Tests are less critical; someone else can do those in a future PR.
> 
> But the vast majority of tests are now passing! I am so proud of you!
> 
> Once the example scripts and doctests pass, I will vote to approve the PR.

Thanks @DrSOKane for your explanation. I fixed the example scripts now and the test passed on my computer. Let us see if doctests fail or not, and if not I will dig into it later, with the help of any suggestions and guildlines from your side.

mzzhuo:
Hi @DrSOKane @kratman regarding the failed doctest, the -W flag in 'sphinx-build -b html -W --keep-going' turns warnings into errors, and we do have 3 warnings of could not finding bibtex key "ploehn2004solvent", "single2018identifying", "safari2008multimodal". it seems not from my PR, could you double check this and see if any solution?

mzzhuo:
> @mzzhuo Last I checked the CI was working on main branch, so I suspect that it is a problem with your code.
> 
> Looking at the key errors:
> 
> ```
> @article{Ploehn2004,
>   title={Solvent 
> ```
> 
> ```
> @article{Safari2008,
>   title={Multimodal
> ```
> 
> ```
> @article{Single2018,
>   title={Identifying
> ```
> 
> It is clearly the citation, followed by first word in the title. Have to checked your new citations/documentation to see if there are any characters which were not properly escaped?

Thanks @kratman for your hints, I think I have fixed the bug now, hopefully the doctests will pass. Then I will leave the rest code-cov/project tests to pybamm team @DrSOKane done from my side.

mzzhuo:
Hey @kratman @DrSOKane, I’ve pulled the latest changes from develop into this feature branch and added tests for all the new code. Local runs suggest the Codecov checks should pass. I’ve also reverted fixes that aren’t directly related to this feature. Could you take a look when you have a moment? Let me know if anything needs to be updated or adjusted.

mzzhuo (commented):
A few changes in this push: the submodel itself has been refactored, and we added unit tests to address Codecov's patch-coverage feedback — even though an end-to-end integration test was already included.

PR #3785, opened 2024-01-30
Resolve broken `scikits.odes` installation on self-hosted M-series runner

agriyakhetarpal:
# Description

The M-series self-hosted runner was using a cached, un-repaired `scikits.odes` wheel which had been linked to the Homebrew rendition of SUNDIALS (which isn't a problem though) instead of the one we regularly use in CI. Apple has a history of [sanitising runtime search paths](https://hynek.me/articles/macos-dyld-env/), which caused the SUNDIALS dynamic libraries to not be found at runtime by the Python extension modules for `scikits.odes`. 

This PR ensures that the build-time dependencies for PyBaMM are always re-compiled on each run, the `SUNDIALS_INST` variable is set, and that `scikits.odes` is built from source every time.

The M-series build passed here: https://github.com/pybamm-team/PyBaMM/actions/runs/7713296977/job/21022885593

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/.github/workflows/run_periodic_tests.yml
+++ b/.github/workflows/run_periodic_tests.yml
@@ -147,6 +147,7 @@ jobs:
           # Don't use Homebrew to install SUNDIALS because scikits.odes looks for
           # in Homebrew folders instead, which we don't want
           brew uninstall sundials --force
+          pip cache remove scikits.odes
           python -m nox -s pybamm-requires -- --force
           python -m nox -s unit
 
--- a/.github/workflows/run_periodic_tests.yml
+++ b/.github/workflows/run_periodic_tests.yml
@@ -134,11 +134,19 @@ jobs:
 
       - name: Install build-time dependencies & run unit tests for M-series macOS runner
         shell: bash
+        env:
+          # Point scikits.odes to the correct SUNDIALS installation
+          SUNDIALS_INST: $HOME/.local/lib
+          # Homebrew environment variables
+          HOMEBREW_NO_INSTALL_CLEANUP: 1
+          NONINTERACTIVE: 1
         run: |
           eval "$(pyenv init -)"
           pyenv activate pybamm-${{ matrix.python-version }}
           python -m pip install --upgrade pip nox
-          brew uninstall sundials
+          # Don't use Homebrew to install SUNDIALS because scikits.odes looks for
+          # in Homebrew folders instead, which we don't want
+          brew uninstall sundials --force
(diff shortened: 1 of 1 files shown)


.github/workflows/run_periodic_tests.yml
```diff
@@ -114,42 +114,54 @@ jobs:
...
+          # Homebrew environment variables
+          HOMEBREW_NO_INSTALL_CLEANUP: 1
+          NONINTERACTIVE: 1
         run: |
           eval "$(pyenv init -)"
           pyenv activate pybamm-${{ matrix.python-version }}
-          python -m pip install --upgrade pip wheel setuptools nox
+          python -m pip install --upgrade pip nox
+          # Don't use Homebrew to install SUNDIALS because scikits.odes looks for
+          # in Homebrew folders instead, which we don't want
```
agriyakhetarpal, line 152: To clarify, using the Homebrew installation of SUNDIALS in this scenario **isn't a problem** (see #3511), but this workflow requires that we build the IDAKLU solver's build-time requirements from source, and therefore it is better to use just one SUNDIALS installation, i.e., the one present in `~/.local/`.

Even though we _can_ use `pybamm_install_odes` to set up `scikits.odes` on developer installations, we shouldn't do so because we limit its use case to user installations (and also because `noxfile.py` doesn't currently have it).

BradyPlanden (approved):
Excellent, thanks for updating this @agriyakhetarpal. LGTM

.github/workflows/run_periodic_tests.yml
```diff
@@ -114,42 +114,54 @@ jobs:
...
-      - name: Install python & create virtualenv
+      - name: Install Python & create virtualenv
         shell: bash
         run: |
           eval "$(pyenv init -)"
           pyenv install ${{ matrix.python-version }} -s
           pyenv virtualenv ${{ matrix.python-version }} pybamm-${{ matrix.python-version }}
 
-      - name: Install dependencies & run unit tests for Windows and MacOS
+      - name: Install build-time dependencies & run unit tests for M-series macOS runner
```
arjxn-py, line 139: Are we not running it for windows now?

agriyakhetarpal: This is a separate job for just the M-series runner, the job step was written wrongly :)

arjxn-py (approved):

kratman:
Let's make sure to squash this so it is easy to cherry pick into releases if needed

PR #3786, opened 2024-01-30
Purge `scikits.odes` cached wheels when using `pybamm_install_odes`

agriyakhetarpal:
# Description

See discussion on #3785. This command exits with a warning if a cached wheel is not present, it does not raise an error and therefore should be safe to add within a subprocess.

We already purge any cached wheels in CI when running the scheduled tests for this entry point – this PR ensures it is done for users running it locally.

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/install_odes.py
+++ b/pybamm/install_odes.py
@@ -191,6 +191,11 @@ def main(arguments=None):
     os.environ["SUNDIALS_INST"] = SUNDIALS_LIB_DIR
     env = os.environ.copy()
     logger.info("Installing scikits.odes via pip")
+    logger.info("Purging scikits.odes whels from pip cache if present")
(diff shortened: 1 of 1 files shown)


arjxn-py:
Should not break anything, hence happy to merge it.

Saransh-cpp:
@arjxn-py please use "squash and merge" unless there is an absolute need of preserving the commit history of a PR.

arjxn-py:
> please use "squash and merge" unless there is an absolute need of preserving the commit history of a PR.

Sure, is it for PRs to be cherry picked in release or should we do it for every PR when commit history is not needed?

Saransh-cpp:
Every PR

Issue #3783, opened 2024-01-30
[Bug]: `casadi_interpolant_linear.dll` not found on Windows with IDAKLUSolver, fails when simulating a drive cycle using Interpolant


chmabaur:
### PyBaMM Version

24.1rc2

### Python Version

3.10.11

### Describe the bug

I want to simulate a drive cycle by passing an Interpolant function to the "Current function [A]" parameter. This however fails as casadi_interpolant_linear.dll is not found.

### Steps to Reproduce

```
import pybamm
import numpy as np

model = pybamm.lithium_ion.DFN()
geometry = model.default_geometry
param = model.default_parameter_values

# Create interpolant
current_interpolant = pybamm.Interpolant(
    np.linspace(0, 5000, 50000),
    np.ones(50000)*0.7,
    pybamm.t
    )

# Set drive cycle
param.update({"Current function [A]": current_interpolant})

# Solve
solver_idaklu = pybamm.IDAKLUSolver(
    rtol=1e-8, atol=1e-8)

sim = pybamm.Simulation(
    model, parameter_values=param, solver=solver_idaklu)
solution_idaklu = sim.solve(calc_esoh=False)
```

### Relevant log output

```shell
File C:\Software\Python\Environments\pybamm_venv\lib\site-packages\pybamm\solvers\idaklu_solver.py:439 in set_up
    rhs_algebraic = idaklu.generate_function(rhs_algebraic.serialize())

RuntimeError: C:\vcpkg\buildtrees\casadi\src\3.6.4-aa7d507b6b.clean\casadi\core\casadi_os.cpp:166: Assertion "handle!=nullptr" failed:
PluginInterface::load_plugin: Cannot load shared library 'casadi_interpolant_linear.dll': 
   (
    Searched directories: 1. casadipath from GlobalOptions
                          2. CASADIPATH env var
                          3. PATH env var (Windows)
                          4. LD_LIBRARY_PATH env var (Linux)
                          5. DYLD_LIBRARY_PATH env var (osx)
    A library may be 'not found' even if the file exists:
          * library is not compatible (different compiler/bitness)
          * the dependencies are not found
   )
  Tried '' :
    Error code (WIN32): 126
  Tried '.' :
    Error code (WIN32): 126
```
```

chmabaur:
@agriyakhetarpal: Are you aware of this issue?

agriyakhetarpal:
Thanks @chmabaur – first time I am looking at this, will explore further.

[...]

Issue #3140, opened 2023-07-11
[Bug]: Integration tests failing M-Series Hardware


BradyPlanden:
### PyBaMM Version

23.5

### Python Version

3.8-3.11

### Describe the bug

Integration tests are currently failing on apple m-series hardware. This seems to be centred around reaching the maximum number of timestep decreases for the casadi solver when testing the cracking submodels. A quick fix might be to decrease the intial step size when running [these](https://github.com/pybamm-team/PyBaMM/blob/1a012e75953d4cd83e6705b4e5c250291fbe75fd/tests/integration/test_models/test_full_battery_models/test_lithium_ion/base_lithium_ion_tests.py#L237C3-L237C3) tests.

### Steps to Reproduce

Run `nox -s integration` on m-series hardware.

### Relevant log output

```shell
https://github.com/pybamm-team/PyBaMM/actions/runs/5515391562/jobs/10061099876
```

brosaplanella:
Sounds like a good fix! Can you check if it works locally? Note that you can pass kwargs to the simulation through `run_basic_processing_test`.

BradyPlanden:
I can take a look at this later in the week. Thinking on it further, we probably need to take a look at why the casadi is decreasing the timesteps to the point of an error for Apple ARM hardware, but not on x64. I'll probe a bit further and come back to this issue.

kratman:
I was looking at this for the examples (```nox -s examples```) which fail on my M1 Mac. I had no luck solving this by changing the dt_max parameter.

I started with **SEI-on-cracks.ipynb** and got it working. I am not sure why yet, but I found that I could stop this error by changing:
```python
var_pts = {
    "x_n": 20,  # negative electrode
    "x_s": 20,  # separator 
    "x_p": 20,  # positive electrode
    "r_n": 30,  # negative particle
    "r_p": 30,  # positive particle
}
```
to
```python
var_pts = {
    "x_n": 20,  # negative electrode
    "x_s": 20,  # separator 
    "x_p": 20,  # positive electrode
    "r_n": 26,  # negative particle
    "r_p": 26,  # positive particle
}
```
Reducing the size of the mesh seems to make it work. Not the best solution, but it lets the tests finish for now.
Adding the same var_pts to **submodel_cracking_DFN_or_SPM.ipynb** also makes that one start to work. Finally I did the same thing in the integration tests, and the fix works there as well.

[...]

PR #3789, opened 2024-01-31
Build M-series (arm64) macOS wheels on GitHub Actions hosted runners

### LINKED ISSUE #3772: Use self-hosted macOS M2 runner to build macOS arm64 or universal wheels
@BradyPlanden and I, in conversation about the usage of the runner for the repository – have decided to look into building wheels for M-series platforms starting with PyBaMM v24.5rcX. This is since cross-compilation for SuiteSparse requires at least a version of 7.2.1, and we support 6.0.3, so we might not be able to do that yet on the standard GitHub Actions runners (see #3462).

In this scenario, this requires setting up a job in the PyPI workflows where the wheels can be built and passed as artifacts to the job that uploads them.

Whether to build a universal macOS wheel or an architecture-specific wheel can be discussed based on the wheel tags we are able to achieve, the macOS X version constraints for such wheels, and the size of the binaries generated.

agriyakhetarpal:
# Description

GitHub Actions M1 runners are free for open-source repositories now – I have configured the PyPI workflow to add a job for these runners.

> [!IMPORTANT]
> The wheels can be tested on M-series hardware via the following link: https://github.com/pybamm-team/PyBaMM/actions/runs/7721068374/

Closes #3772, related to #3462

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/.github/workflows/publish_pypi.yml
+++ b/.github/workflows/publish_pypi.yml
@@ -166,7 +166,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
         with:
-          python-version: 3.8
+          python-version: 3.10
 
       - name: Clone pybind11 repo (no history)
         run: git clone --depth 1 --branch v2.11.1
--- a/.github/workflows/publish_pypi.yml
+++ b/.github/workflows/publish_pypi.yml
@@ -133,7 +133,7 @@ jobs:
           CIBW_REPAIR_WHEEL_COMMAND_LINUX: auditwheel repair -w {dest_dir} {wheel}
           CIBW_TEST_COMMAND: python -c "import pybamm; pybamm.IDAKLUSolver()"
 
-      - name: Build wheels on macOS
+      - name: Build wheels on macOS amd64
         if: matrix.os == 'macos-latest'
         run: pipx run cibuildwheel --output-dir wheelhouse
         env:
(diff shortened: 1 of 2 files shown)


agriyakhetarpal:
Note to self:

- [ ] Look into if archflags configuration is needed for CMake
- [ ] Add CHANGELOG entry
- [ ] Investigate whether to build universal2 wheel via `delocate-fuse`, `delocate-addplat`, or similar or to keep macOS wheels architecture-specific

agriyakhetarpal:
We should start testing PyBaMM on M1 runners as well if we are going to be distributing them – I can do that in a follow-up PR, though I am not sure if #3140 (particle cracking submodels) has been resolved or not.

.github/workflows/publish_pypi.yml
```diff
@@ -213,7 +247,7 @@ jobs:
   open_failure_issue:
     needs: [build_windows_wheels, build_macos_and_linux_wheels, build_sdist]
     name: Open an issue if build fails
-    if: ${{ always() && contains(needs.*.result, 'failure') && github.repository_owner == 'pybamm-team'}}
+    if: ${{ contains(needs.*.result, 'failure') && github.repository_owner == 'pybamm-team'}}
```
Saransh-cpp, line 250: Why was `always()` removed from here?

Saransh-cpp (approved):
Thanks, @agriyakhetarpal! This looks really good!

.github/workflows/publish_pypi.yml
agriyakhetarpal: I removed it because its presence negated the existence of the other conditions – this job ran even in forks despite not being supposed to. The check for failures in the condition seems to be good enough for this purpose.

Saransh-cpp: This would not work without `always()`. If any of the wheel jobs fail, the `open_failure_issue` job will never be executed. Using always is a bit hacky but that is what makes this work.

Maybe we can create a new issue for this and keep it as it is until we find a solution?

Saransh-cpp:
Also, all green ticks in this PR, nice!

.github/workflows/publish_pypi.yml
agriyakhetarpal: Makes sense, Actions are disabled by default on forks anyway. Happy to revert the change for now

agriyakhetarpal, line 250:
```suggestion
    if: ${{ always() && contains(needs.*.result, 'failure') && github.repository_owner == 'pybamm-team'}}
```

Saransh-cpp (approved):

kratman:
> I am not sure if #3140 (particle cracking submodels) has been resolved or not.

I put in a stop-gap fix for that ticket. I did not mark it as closed because all I did was adjust the grid. Not really a full fix or investigation. Having tests that would alert us sooner would probably make fixing stuff like that more visible and we could discuss closing that issue if we start testing more thoroughly on M-series hardware.

agriyakhetarpal:
> Having tests that would alert us sooner would probably make fixing stuff like that more visible and we could discuss closing that issue if we start testing more thoroughly on M-series hardware.

Sure, happy to merge this after you can download and test a wheel or two (I tested them, they are working for me since the `IDAKLUSolver` can be instantiated and can be used to solve a basic Simulation – even though that seems to not be a reliable test, see #3783).

I will set up testing against the M1 runners in a follow-up PR as soon after this is merged.

kratman:
@agriyakhetarpal I downloaded the wheels to test them

kratman (approved):

PR #3792, opened 2024-01-31
Update cron schedule for PyPI publishing workflow and add cache for vcpkg dependencies

### LINKED ISSUE #3664: Utilise caches for Windows nightly tests and test wheels weekly instead of fortnightly
The scheduled tests (unit, integration, examples) are run in a CRON job overnight for all platforms, but they don't compile the IDAKLU solver for Windows. We don't do it in PR tests to save time, but it should be reasonable to do so for the nightly tests – I have noticed that it takes around 15 minutes to compile the build-time dependencies on Windows, and less than three minutes if a `vcpkg` cache hit occurs. Allowing this cache to be set for the nightly tests on an ephemeral system like this would make sure that it will remain available for further tests and when one tries to debug the wheels or publish them, and thereby ensure that the jobs remain speedy. The cache can be invalidated by any changes to the `vcpkg` configuration files anyway so it is ensured that stale caches will never be used.

Building wheels weekly instead of fortnightly ensures that the set cache can be retained for further use in different scenarios, since GitHub Actions caches are evicted after a duration of one week.

R-Yash:
Update cron schedule for PyPI publishing workflow and add cache for vcpkg dependencies

# Description

Added a process to cache vcpkg dependencies in `run_periodic_tests.yaml` and made the tests run every week by editing `publish_pypi.yaml`

Fixes #3664 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/.github/workflows/run_periodic_tests.yml
+++ b/.github/workflows/run_periodic_tests.yml
@@ -110,6 +110,16 @@ jobs:
         if: matrix.os == 'ubuntu-latest'
         run: python -m nox -s scripts
 
+      - name: Cache vcpkg dependencies
+        if: matrix.os == 'windows-latest'
+        uses: actions/cache@v2
+        with:
+          path: ~/.cache/vcpkg
+          key: ${{ runner.os }}-vcpkg-${{ hashFiles('**/vcpkg.json') }}
+          restore-keys: |
+            ${{ runner.os }}-vcpkg-
+
+
   #M-series Mac Mini
   build-apple-mseries:
     if: github.repository_owner == 'pybamm-team'
--- a/.github/workflows/publish_pypi.yml
+++ b/.github/workflows/publish_pypi.yml
@@ -3,8 +3,8 @@ on:
   release:
     types: [published]
   schedule:
-    # Run at 10 am UTC on day-of-month 1 and 15.
-    - cron: "0 10 1,15 * *"
+    # Run every Sunday at 00:00.
+    - cron: "0 0 * * 0"
   workflow_dispatch:
     inputs:
       target:


.github/workflows/run_periodic_tests.yml
```diff
@@ -122,7 +122,17 @@ jobs:
...
+      - name: Cache vcpkg dependencies
+        if: matrix.os == 'windows-latest'
+        uses: actions/cache@v2
+        with:
+          path: ~/.cache/vcpkg
+          key: ${{ runner.os }}-vcpkg-${{ hashFiles('**/vcpkg.json') }}
+          restore-keys: |
+            ${{ runner.os }}-vcpkg-
+
+
```
agriyakhetarpal, line 134: Besides this, we also need the PyBaMM-related environment variables, such as PYBAMM_USE_VCPKG_ON, CMAKE_BUILD_PARALLEL_LEVEL, and the rest present in `publish_pypi.py` so that `setup.py` can use all features of `vcpkg`.

After you add all of those, I would suggest triggering this workflow from this branch to test these changes (if you do it correctly, you should see a `win-amd64` file being built for PyBaMM rather than the current `none-any.whl`).

agriyakhetarpal (changes requested):

.github/workflows/run_periodic_tests.yml
R-Yash: @agriyakhetarpal How can I trigger a workflow from this branch?

arjxn-py: You'd have to go to the `actions` tab on your fork then select the workflow you want to run from the left panel. 
Click on the `Run Workflow` button >> Select the target branch >> Confirm after Clicking on the blue `Run Workflow` button


image

agriyakhetarpal: Thanks @arjxn-py. Also, these environment variables have been set for the entire matrix. We need these just for the Windows steps in the job. We don't have separate Windows steps, so you can create a separate step with `Run integration tests on Windows` and add these environment variables there (similarly for unit tests, etc.)

agriyakhetarpal:
@R-Yash, feel free to ask if you need help with anything!

arjxn-py:
@R-Yash are you still working on this?

agriyakhetarpal:
Let's close this, actually. We could wait for a while, but not only has the contributor been unresponsive, but also that this is _likely_ stalled until we have a resolution for #3783 (the PyBaMM IDAKLU solver is broken on Windows even if one installs from source). Please feel free to reopen this as you feel fit, @R-Yash!

PR #3795, opened 2024-02-01
Added test to check if simulation has been solved in `simulation.py` and `batch_study.py`

### LINKED ISSUE #3725: Check if the solution exists before creating a GIF (`Simulation.create_gif` and `BatchStudy.create_gif`)
I liked how #3723 checks if the simulations have been solved before proceeding ahead to the plotting functionality. It would be nice to have similar checks in -

https://github.com/pybamm-team/PyBaMM/blob/c3def31a04df163769851733657d9b49b3700bca/pybamm/batch_study.py#L184-L187

and 

https://github.com/pybamm-team/PyBaMM/blob/c3def31a04df163769851733657d9b49b3700bca/pybamm/simulation.py#L1047-L1050

prady0t:
# Description

Simple check for if simulations have been solved or not.

Fixes #3725 

## Type of change

# Key checklist:

## Further checks:


pybamm/batch_study.py
```diff
@@ -196,7 +197,8 @@ def create_gif(self, number_of_images=80, duration=0.1, output_filename="plot.gi
             Name of the generated GIF file.
 
         """
-
+        if not self.sims:
```
Saransh-cpp, line 200: You could do something like -
```suggestion
        if not hasattr(self, "sims"):
```
I think this should work.

pybamm/batch_study.py
```diff
@@ -86,6 +86,7 @@ def __init__(
         self.repeats = repeats
         self.permutations = permutations
         self.quick_plot = None
+        self.sims = []
```
Saransh-cpp, line 89: Initializing this in `__init__` would be redundant. We can keep this in its original place.

Saransh-cpp (changes requested):
Thanks, @prady0t! Looks really nice! A couple of comments below -

pybamm/batch_study.py
Saransh-cpp: and using `hasattr` will still work

prady0t:
@Saransh-cpp I've done the suggested changes here.

Saransh-cpp (approved):
Thanks, @prady0t! Looks good to me!

agriyakhetarpal:
Should we add a CHANGELOG entry here?

kratman:
> Should we add a CHANGELOG entry here?

I think it is a small enough change that it does not need it. However that one test keeps failing and I doubt it is due to this small PR. Should we open a PR for that or do you already know what is happening?

agriyakhetarpal:
> Should we open a PR for that or do you already know what is happening?

No worries, it looks like it was coming from a cached wheel for `scikits.odes` – I deleted the caches and the test passed this time. Let's merge

agriyakhetarpal:
@all-contributors please add @prady0t for code, test

Issue #3617, opened 2023-12-13
Use `pytest` for testing


agriyakhetarpal:
### Description

I recently learned that `pytest` supports `unittest`-style tests out of the box and needs little to no extra configuration for setting up test commands and just a tad bit more configuration to use without any differences to how the tests have been written.

While rewriting the bulk of our tests in the `tests/` folder and subdirectories to `pytest` ("Migrate to `pytest`", as it would be called) is a humongous task rife with tedium and should be reserved for potential long-term projects, I would say that we are definitely in a position to start using `pytest` itself.

I ran the unit tests on my macOS machine with eight cores locally with `pytest-xdist` installed and enabled, using the command

```python
pytest tests/unit/
```

to run them in parallel, and it offered a staggering `2.5x` speedup over `python run-tests.py --unit`, i.e., serial execution of tests!

### Motivation

1. `pytest` is a modern testing framework used by most packages in the Scientific Python ecosystem
2. We are already using it with `nbmake` to test the example notebooks
3. It simplifies testing a bit since we can let go of most of the code in the `run-tests.py` file and use the `pytest` built-in logging and path discovery functionalities to extract the most out of our testing
4. Better CI times by a very big margin – unit tests and integration tests can both be completed in under ~10 minutes on GitHub Actions runners in separate jobs, which means we can add them to a single job one after the other as well (it reduces the strain on the runners and we wouldn't need to request so many runners like we do currently).
5. Integration tests can be executed to test the built wheels for a Python version with `cibuildwheel`
6. ...and so on

### Possible Implementation

- The `conftest.py` file can be used to configure the `unittest.TestCase` metaclass we currently use for ensuring the reliability of the test cases (see #2833)
- From @Saransh-cpp: `pytest.importorskip` can be used to test the optional dependencies
- I noticed that some Jax solver test cases in the integration tests are causing some troubles when running in parallel mode, we can either reduce the distribution of the workers to run all tests in the same class in serial, or the better way mark just those tests with a certain decorator or `@pytest.fixture` that instructs them to be assigned to a single worker.
- Other nitty-gritties and niceties such as `pytest-cov` and `pytest-doctest`/`pytest-doctestplus` or other plugins out of the hundreds of plugins can be looked into later as a part of the rewriting process

### Additional context

N/A

Saransh-cpp:
Quick suggestions:

> pytest-cov and pytest-doctest

[...]

PR #3800, opened 2024-02-04
Refactor test_experiment_step_termination.py***

abhicodes369:
(from unit test to pytest)

# Description
converted a unit test case that uses the unit test to pytest. On the official website of pybamm  I saw that they are considering migrating from unit test to pytest.


agriyakhetarpal:
Thanks for starting this @abhicodes369. It looks good, but ideally we want to do this after #3617, so I will let this sit for a while until that is complete and `pytest` is added to the codebase. If you're interested in contributing, please feel free to discuss something over there since multiple people can be assigned to that issue 🙂

agriyakhetarpal:
@abhicodes369, if you're interested, you can help drive #3857 to completion – or we can take this up again after it gets merged (should be about a week at most).

abhicodes369:
I will look into it

agriyakhetarpal (approved):
Codacy raises a warning on assert statements despite the fact that it is the recommended way to write `pytest`-style tests and that these are test files. This can be disabled later, some insights here: https://github.com/fossasia/query-server/issues/332

agriyakhetarpal (approved):

agriyakhetarpal:
@allcontributors please add @abhicodes369 for tests

abhicodes369:
thanks @agriyakhetarpal it really means a lot

Issue #3387, opened 2023-10-01
Add a link checker step in the doctests jobs and fix broken links that appear


agriyakhetarpal:
Currently, we use Lychee to check for broken links, but we ignore a lot of Sphinx-related links because of referencing files internally. `Sphinx` provides a [`linkcheck`](https://www.sphinx-doc.org/en/master/usage/builders/index.html#sphinx.builders.linkcheck.CheckExternalLinksBuilder) builder for this purpose that can be invoked in CI for checking links (either through a new `nox` session, or within the `docs` session using posargs, or just calling the command directly).

It might make sense to rename the Lychee workflow to something like `links.yml` and add a parallel job to run the above command instead of another step.

RohitP2005:
hey @Saransh-cpp is the issue still open ?

agriyakhetarpal:
Hi @RohitP2005, this is still open, thanks for your interest!

RohitP2005:
Thanks, @agriyakhetarpal i will start my work

PR #3818, opened 2024-02-14
Fix broken `Open in Colab` badges linked to example notebooks

### LINKED ISSUE #3817: [Bug]: The link to google colab from the tutorials on docs.pybamm.org does not work.
### PyBaMM Version

v24.1

### Python Version

NA

### Describe the bug

The link to google colab from the tutorials on docs.pybamm.org does not work.

It gives the error:
malformed GitHub path: missing 'blob' before branch name: pybamm-team/PyBaMM/tree/v24.1/docs/source/examples/notebooks/getting_started/tutorial-1-how-to-run-a-model.ipynb

### Steps to Reproduce

From
https://docs.pybamm.org/en/stable/source/examples/notebooks/getting_started/tutorial-1-how-to-run-a-model.html#
clink on 
 Open this notebook in Google Colab

### Relevant log output

_No response_

arjxn-py:
# Description

This issue aims to fix broken `Open in Colab` badges on docs.pybamm.org

Fixes #3817 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -317,7 +317,7 @@ if (os.environ.get("READTHEDOCS_VERSION") == "latest") or (
 
 if os.environ.get("READTHEDOCS_VERSION") == "stable":
     notebooks_version = version
-    append_to_url = f"tree/blob/v{notebooks_version}"
+    append_to_url = f"blob/v{notebooks_version}"
 
 if os.environ.get("READTHEDOCS_VERSION_TYPE") == "external":
     notebooks_version = os.environ.get("READTHEDOCS_GIT_COMMIT_HASH")
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -317,7 +317,7 @@ if (os.environ.get("READTHEDOCS_VERSION") == "latest") or (
 
 if os.environ.get("READTHEDOCS_VERSION") == "stable":
     notebooks_version = version
-    append_to_url = f"tree/v{notebooks_version}"
+    append_to_url = f"tree/blob/v{notebooks_version}"
 
 if os.environ.get("READTHEDOCS_VERSION_TYPE") == "external":
     notebooks_version = os.environ.get("READTHEDOCS_GIT_COMMIT_HASH")


kratman (approved):
The change looks reasonable and it works in the docs build. I did not test it locally though, not sure if that matters

arjxn-py:
I believe that the docs build in this PR is unrelated as it's not stable build. Nice to have @agriyakhetarpal review as well, I hope merging this to develop would fix the badges without having to update `v24.1` branch.

agriyakhetarpal (changes requested):
This should be just `/blob/`, rather than both `/tree/blob/`. 

i.e., this link works: https://githubtocolab.com/pybamm-team/PyBaMM/blob/v24.1/docs/source/examples/notebooks/getting_started/tutorial-3-basic-plotting.ipynb

but these don't:
 https://githubtocolab.com/pybamm-team/PyBaMM/tree/v24.1/docs/source/examples/notebooks/getting_started/tutorial-3-basic-plotting.ipynb (the original, failing one)

and

https://githubtocolab.com/pybamm-team/PyBaMM/blob/tree/v24.1/docs/source/examples/notebooks/getting_started/tutorial-3-basic-plotting.ipynb (current modification)

Sadly, this won't fix the link failure for the 24.1 release. Wish there were a reliable method to test such links, since I don't think even #3387 would cover this scenario.

Please feel free to merge after applying the request changes!

agriyakhetarpal (approved):

arjxn-py:
> Sadly, this won't fix the link failure for the 24.1 release. I wish there were a reliable method to test such links, since I don't think even #3387 would cover this scenario.

Would it be nicer to open a separate issue about this other than #3387 where we could also further discuss about exploratory ways to cover the issue?

agriyakhetarpal:
> Would it be nicer to open a separate issue about this other than #3387 where we could also further discuss about exploratory ways to cover the issue?

I think this issue is minor enough to say that it won't get repeated again

arjxn-py:
The badges are still not fixed for the stable version, should we cherry-pick this to `v24.1`?

agriyakhetarpal:
> The badges are still not fixed for the stable version, should we cherry-pick this to `v24.1`? or maybe wait for next rc

I surfed some links and I am understanding that RTD _does_ build the documentation via GitHub webhooks from repository tags. So, yes, I think cherry-picking this commit to the v24.1 tag should fix this on the `stable` version of the documentation. I will leave this up to what @Saransh-cpp thinks about this, though my preference would be to wait until the next RC. Maybe Google Colab or https://githubtocolab.com changed something from their end, because I remember that this had worked when I had implemented it (which is why I kept it as `tree` instead of `blob`).

Saransh-cpp:
I won't suggest moving the tag around. The maximum we can do is create a `v24.1post1` release, but I don't think that is required. Let's keep this for the next release.

PR #3848, opened 2024-02-28
Make `typing-extensions` and `sympy` required dependencies to fix PyBaMM import

### LINKED ISSUE #3836: [Bug]: non-required dependency `typing_extensions` is imported by default
### PyBaMM Version

24.1

### Python Version

3.8-3.12

### Describe the bug

There might be some instance in #3578 where non required dependency has been imported by default, which is causing failure, [related build logs](https://github.com/arjxn-py/PyBaMM/actions/runs/8032837736/job/21942676799). 

### Steps to Reproduce

* try installing latest built wheels on develop
* do `import pybamm`

### Relevant log output

_No response_

arjxn-py:
# Description

Fix broken import of PyBaMM raised by type hinting.

Fixes #3836 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,9 +18,9 @@
 
 ## Breaking changes
 
+- Integrated the `[latexify]` extra into the core PyBaMM package, deprecating the `pybamm[latexify]` set of optional dependencies. SymPy is now a required dependency and will be installed upon installing PyBaMM ([#3848](https://github.com/pybamm-team/PyBaMM/pull/3848))
 - Renamed "testing" argument for plots to "show_plot" and flipped its meaning (show_plot=True is now the default and shows the plot) ([#3842](https://github.com/pybamm-team/PyBaMM/pull/3842))
 - Dropped support for BPX version 0.3.0 and below ([#3414](https://github.com/pybamm-team/PyBaMM/pull/3414))
-- Integrated the `[latexify]` extra into the core PyBaMM package, deprecating the `pybamm[latexify]` set of optional dependencies. SymPy is now a required dependency and will be installed upon installing PyBaMM ([#3848](https://github.com/pybamm-team/PyBaMM/pull/3848))
 
 # [v24.1](https://github.com/pybamm-team/PyBaMM/tree/v24.1) - 2024-01-31
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 
 - Renamed "testing" argument for plots to "show_plot" and flipped its meaning (show_plot=True is now the default and shows the plot) ([#3842](https://github.com/pybamm-team/PyBaMM/pull/3842))
 - Dropped support for BPX version 0.3.0 and below ([#3414](https://github.com/pybamm-team/PyBaMM/pull/3414))
+- Integrated `[latexify]` module into the core PyBaMM package, deprecating the `pybamm[latexify]` version for installation.([#3848](https://github.com/pybamm-team/PyBaMM/pull/3848))
 
 # [v24.1](https://github.com/pybamm-team/PyBaMM/tree/v24.1) - 2024-01-31
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@
 
 - Renamed "testing" argument for plots to "show_plot" and flipped its meaning (show_plot=True is now the default and shows the plot) ([#3842](https://github.com/pybamm-team/PyBaMM/pull/3842))
 - Dropped support for BPX version 0.3.0 and below ([#3414](https://github.com/pybamm-team/PyBaMM/pull/3414))
-- Integrated `[latexify]` module into the core PyBaMM package, deprecating the `pybamm[latexify]` version for installation.([#3848](https://github.com/pybamm-team/PyBaMM/pull/3848))
+- Integrated the `[latexify]` extra into the core PyBaMM package, deprecating the `pybamm[latexify]` set of optional dependencies. SymPy is now a required dependency and will be installed upon installing PyBaMM ([#3848](https://github.com/pybamm-team/PyBaMM/pull/3848))
 
 # [v24.1](https://github.com/pybamm-team/PyBaMM/tree/v24.1) - 2024-01-31
 
--- a/pybamm/expression_tree/independent_variable.py
+++ b/pybamm/expression_tree/independent_variable.py
@@ -6,7 +6,6 @@ import sympy
 import numpy as np
 
 import pybamm
-from pybamm.util import have_optional_dependency
 from pybamm.type_definitions import DomainType, AuxiliaryDomainType, DomainsType
 
 KNOWN_COORD_SYS = ["cartesian", "cylindrical polar", "spherical polar"]
@@ -58,7 +57,6 @@ class IndependentVariable(pybamm.Symbol):
 
     def to_equation(self) -> sympy.Symbol:
         """Convert the node and its subtree into a SymPy equation."""
-        sympy = have_optional_dependency("sympy")
         if self.print_name is not None:
             return sympy.Symbol(self.print_name)
         else:
@@ -102,7 +100,6 @@ class Time(IndependentVariable):
 
     def to_equation(self):
         """Convert the node and its subtree into a SymPy equation."""
-        sympy = have_optional_dependency("sympy")
         return sympy.Symbol("t")
 
 
--- a/pybamm/expression_tree/independent_variable.py
+++ b/pybamm/expression_tree/independent_variable.py
@@ -2,7 +2,6 @@
 # IndependentVariable class
 #
 from __future__ import annotations
-import sympy
 import numpy as np
 
 import pybamm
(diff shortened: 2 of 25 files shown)


kratman (changes requested):
I would say that we need SymPy. It is a requirement for expression tree if it uses the types from the module. We should probably just make SymPy required instead.

If expression tree is not needed by everything else, then we could make expression tree optional, but I think SymPy is clearly required by expression tree

pybamm/expression_tree/independent_variable.py
```diff
@@ -56,7 +55,7 @@ def _jac(self, variable) -> pybamm.Scalar:
         """See :meth:`pybamm.Symbol._jac()`."""
         return pybamm.Scalar(0)
 
-    def to_equation(self) -> sympy.Symbol:
+    def to_equation(self) -> "sympy.Symbol":  # noqa: F821, UP037
```
arjxn-py, line 58: Have done forward referencing for `sympy` here as mentioned in [PEP-484](https://peps.python.org/pep-0484/#forward-references)

kratman: I guess my point was that we have to do forward referencing and suppressing warnings just to avoid marking a clear dependency as a dependency. I think this is a bit of a code smell and we should think about the code is organized if we really want to make sympy optional

kratman: Usually this sort of thing is for avoiding circular imports or building abstract interfaces. I am not certain this is intended for optional dependencies:

> When a type hint contains names that have not been defined yet, that definition may be expressed as a string literal, to be resolved later.
>
> A situation where this occurs commonly is the definition of a container class, where the class being defined occurs in the signature of some of the methods.

...

> A common use for forward references is when e.g. Django models are needed in the signatures. Typically, each model is in a separate file, and has methods taking arguments whose type involves other models. Because of the way circular imports work in Python, it is often not possible to import all the needed models directly.

arjxn-py: Ah Sorry, guess i started this thread before reading your comment above.

> we have to do forward referencing and suppressing warnings just to avoid marking a clear dependency as a dependency. I think this is a bit of a code smell and we should think about the code is organized if we really want to make sympy optional

As this had to be done at this single specific instance I believe it doesn't affect the integrity of the code plus referring to [#3578(comment)](https://github.com/pybamm-team/PyBaMM/pull/3578#discussion_r1462422642) it should be good to come to an agreement prior to making `sympy` a required dependency.

kratman: I agree that it should be a discussion on the sympy dependency as a whole, but I think this change goes the other way. It hides the sympy dependency even more since this is another part of the code which requires it and we removed the import.

As with my comment on the if-statement import in #3578, we are adding complexity to avoid this dependency

arjxn-py: > I think this change goes the other way. It hides the sympy dependency even more since this is another part of the code which requires it and we removed the import.

Yes, as this PR is meant to fix the priority issue of breaking import. Although we're not removing the import this time as it's being imported same as before #3578, we're just avoiding importing that optional dependency outside it's functional scope.

agriyakhetarpal: I agree with @kratman's original comment. I would say we should add both `sympy` and `typing-extensions` as required dependencies, which should resolve the issue – the expression tree is an integral module in the PyBaMM codebase (or try to make `typing-extensions` optional rather than `sympy`). It will be a breaking change since we will have removed  the `[latexify]` extra, so we should add a note in the CHANGELOG about this as well.

agriyakhetarpal (commented):
Thanks, @arjxn-py – this looks better now. To ensure that this doesn't happen again, could you please add some more tests for the `have_optional_dependency` method?

In particular, to do this we can either open and read the `pyproject.toml` file to extract the list of dependencies we have, or [use `importlib.metadata`](https://docs.python.org/3/library/importlib.metadata.html#distribution-metadata) (the latter is better):

```python
import importlib.metadata

importlib.metadata.metadata("pybamm").json['requires_dist']
```

returns this list of strings:

```
['numpy >=1.23.5',
 'scipy >=1.9.3',
 'casadi >=3.6.3',
 'xarray >=2022.6.0',
 'anytree >=2.8.0',
 "autograd >=1.6.2 ; extra == 'all'",
 "scikit-fem >=8.1.0 ; extra == 'all'",
 "pybamm[bpx,cite,examples,latexify,pandas,plot,tqdm] ; extra == 'all'",
 "bpx >=0.4.0 ; extra == 'bpx'",
 "bibtexparser >=2.0.0b5 ; extra == 'cite'",
 "pre-commit ; extra == 'dev'",
 "ruff ; extra == 'dev'",
 "nox ; extra == 'dev'",
 "coverage[toml] ; extra == 'dev'",
 "pytest >=6 ; extra == 'dev'",
 "pytest-xdist ; extra == 'dev'",
 "nbmake ; extra == 'dev'",
 "sphinx >=6 ; extra == 'docs'",
 "sphinx-rtd-theme >=0.5 ; extra == 'docs'",
 "pydata-sphinx-theme ; extra == 'docs'",
 "sphinx-design ; extra == 'docs'",
 "sphinx-copybutton ; extra == 'docs'",
 "myst-parser ; extra == 'docs'",
 "sphinx-inline-tabs ; extra == 'docs'",
 "sphinxcontrib-bibtex ; extra == 'docs'",
 "sphinx-autobuild ; extra == 'docs'",
 "sphinx-last-updated-by-git ; extra == 'docs'",
 "nbsphinx ; extra == 'docs'",
 "ipykernel ; extra == 'docs'",
 "ipywidgets ; extra == 'docs'",
 "sphinx-gallery ; extra == 'docs'",
 "sphinx-hoverxref ; extra == 'docs'",
 "sphinx-docsearch ; extra == 'docs'",
 "jupyter ; extra == 'examples'",
 'jax ==0.4.20 ; (python_version >= "3.9") and extra == \'jax\'',
 'jaxlib ==0.4.20 ; (python_version >= "3.9") and extra == \'jax\'',
 "sympy >=1.12 ; extra == 'latexify'",
 "scikits.odes ; extra == 'odes'",
 "pandas >=1.5.0 ; extra == 'pandas'",
 "imageio >=2.3.0 ; extra == 'plot'",
 "matplotlib >=3.6.0 ; extra == 'plot'",
 "tqdm ; extra == 'tqdm'"]
```

from which we should be able to clean and parse all of the optional dependencies. The `json` attribute was added in Python 3.10, so for versions prior to that we can perhaps read `importlib.metadata.metadata("pybamm")["Provides-Extra"]` (note that there are multiple keys of the same name `"Provides-Extra"]`).

P.S. The best resource for the job would be the PyPA `packaging` package (https://packaging.pypa.io/en/stable/metadata.html), but that is an external dependency and not part of the stdlib

arjxn-py:
> To ensure that this doesn't happen again, could you please add some more tests for the `have_optional_dependency` method?

Thanks for the review @agriyakhetarpal, but just mocking all optional dependencies to tests won't be enough as we currently we run tests on `pybamm[all]` variant. I should need to create a new job where i just install pybamm base variant and then run necessary tests. But considering the priority of this issue i feel it'd be right to open a separate PR for that.

agriyakhetarpal:
Yes, a follow-up PR should be okay – we should target that soon after this PR so that we are properly testing all dependencies. It should be nothing more than another entry to the GitHub Actions matrix that can be accessed in the context of a step

agriyakhetarpal:
Could you fix the example notebooks failure here? It does look like a small fix – not sure how it was passing earlier

arjxn-py:
> Could you fix the example notebooks failure here? It does look like a small fix – not sure how it was passing earlier

Sure

kratman:
Just a question: Is `have_optional_dependency()` used anywhere else? It looks like it was used pretty heavily for sympy. If it is was not used anywhere else then can it be removed as well?

agriyakhetarpal:
It is currently used for `matplotlib` and `scikit-fem`, which seem to be more optional in their use than `sympy`. `matplotlib` was originally the only dependency that was implemented to be optional before PyBaMM v23.5.

At some point we should consider renaming the function to `import_optional_dependency` to align it with the fact that it is just a wrapper over `importlib.import_module()` and raise a helpful error message

kratman:
> At some point we should consider renaming the function to `import_optional_dependency` to align it with the fact that it is just a wrapper over `importlib.import_module()` to raise a helpful error message

Yeah that would be a way better name. We should just start fixing that sort of thing on the fly as we see it

agriyakhetarpal (approved):

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
 
 - Renamed "testing" argument for plots to "show_plot" and flipped its meaning (show_plot=True is now the default and shows the plot) ([#3842](https://github.com/pybamm-team/PyBaMM/pull/3842))
 - Dropped support for BPX version 0.3.0 and below ([#3414](https://github.com/pybamm-team/PyBaMM/pull/3414))
+- Integrated the `[latexify]` extra into the core PyBaMM package, deprecating the `pybamm[latexify]` set of optional dependencies. SymPy is now a required dependency and will be installed upon installing PyBaMM ([#3848](https://github.com/pybamm-team/PyBaMM/pull/3848))
```
Saransh-cpp, line 23: Should go at the top

Saransh-cpp (approved):
Thanks, @arjxn-py!

CHANGELOG.md
arjxn-py: This should be fine [`ad958a2`](https://github.com/pybamm-team/PyBaMM/pull/3848/commits/ad958a2d7dca2007a90a747a93aeade2d8206ac4), or did you mean something else?

PR #3857, opened 2024-03-02
Pytest for unit testing

### LINKED ISSUE #3617: Use `pytest` for testing

prady0t:
# Description

Still a WIP. 
Since most of the integration tests were failing, they need to be rewritten completely before we can switch to pytest. For now we are using pytest just for unit tests. We still reduce testing time while running `nox` by a considerable factor.

Addresses a part of #3617 

## Type of change

# Key checklist:

## Further checks:


prady0t:
Reason why tests are failing is because `pytest` recognises `arg`, `arg1`, `arg2` from across different function's arguments as [pytest fixtures](https://pytest-with-eric.com/fixtures/pytest-fixtures/#Fixtures-In-The-Test). 
I tried using lambda function in place of 
```py
def test_function(arg):
     return arg + arg
```
using:
```
expr = pybamm.Function(lambda arg: arg+arg, a)
```
insted of 
https://github.com/pybamm-team/PyBaMM/blob/8056d224e75b3926743319a4743abb7c6ab0dac2/tests/unit/test_expression_tree/test_operations/test_evaluate_python.py#L96

But assertion fails (in unittest and pytest):
```
AssertionError: 'function (<lambda>)' != 'function (test_function)'
- function (<lambda>)
+ function (test_function)
```
Maybe if we define `arg fixture` we could solve this.  
@agriyakhetarpal what do you think?

prady0t:
One easy fix is to rename functions like `test_function` to maybe `function1` and others respectively. Due to pytest naming convention, it recognises these as test functions.

agriyakhetarpal:
Does renaming them from `test_function` to something like `function_for_testing` help? Using a lambda function is not only a bit hard on readability but also it would not be able to cover every use case (returns a single expression and you might need to define a different one everywhere)

prady0t:
Why are the CI tasks failing?

agriyakhetarpal:
> Why are the CI tasks failing?

It's probably coming from Matplotlib opening too many figures/plots when running the tests. We should close them after they are opened, I guess. There might be a few other failures due to differences in how `unittest` and `pytest` handle test cases, so you might have to fix the tests.

Also, according to https://github.com/pybamm-team/PyBaMM/pull/3857#issuecomment-2001808198 that the coverage is 19.40%, which means that the coverage percentage is not being reported correctly. We would need to add pytest-cov and configure it with the unit tests, as noted by @Saransh-cpp in https://github.com/pybamm-team/PyBaMM/issues/3617#issuecomment-1854702640.

prady0t:
These are mostly the failures we have to deal with :

```
FAILED tests/unit/test_expression_tree/test_operations/test_latexify.py::TestLatexify::test_latexify -   File "/Users/runner/work/PyBaMM/PyBaMM/pybamm/expression_tree/operations/latexify.py", line 82
    geo_latex = f"\quad {rng_min} < {name} < {rng_max}"
                                                      ^
SyntaxError: invalid escape sequence '\q'
```


```
FAILED tests/unit/test_expression_tree/test_operations/test_latexify.py::TestLatexify::test_latexify_other_variables -   File "/Users/runner/work/PyBaMM/PyBaMM/pybamm/expression_tree/operations/latexify.py", line 82
    geo_latex = f"\quad {rng_min} < {name} < {rng_max}"
                                                      ^
SyntaxError: invalid escape sequence '\q'
```


```
FAILED tests/unit/test_expression_tree/test_operations/test_latexify.py::TestLatexify::test_sympy_preview -   File "/home/runner/work/PyBaMM/PyBaMM/pybamm/expression_tree/operations/latexify.py", line 82
    geo_latex = f"\quad {rng_min} < {name} < {rng_max}"
                                                      ^
SyntaxError: invalid escape sequence '\q'
```


```
FAILED tests/unit/test_util.py::TestUtil::test_import_optional_dependency - AssertionError: "Optional dependency anytree is not available." does not match "Optional dependency anytree.exporter is not available. See https://docs.pybamm.org/en/latest/source/user_guide/installation/index.html#optional-dependencies for more details."
```


```
FAILED tests/unit/test_simulation.py::TestSimulation::test_create_gif - _tkinter.TclError: Can't find a usable init.tcl in the following directories: 
    {C:\hostedtoolcache\windows\Python\3.11.8\x64\tcl\tcl8.6}

C:/hostedtoolcache/windows/Python/3.11.8/x64/tcl/tcl8.6/init.tcl: couldn't read file "C:/hostedtoolcache/windows/Python/3.11.8/x64/tcl/tcl8.6/init.tcl": No error
couldn't read file "C:/hostedtoolcache/windows/Python/3.11.8/x64/tcl/tcl8.6/init.tcl": No error
    while executing
"source C:/hostedtoolcache/windows/Python/3.11.8/x64/tcl/tcl8.6/init.tcl"
    ("uplevel" body line 1)
    invoked from within
"uplevel #0 [list source $tclfile]"


This probably means that Tcl wasn't installed properly.
```


```
FAILED tests/unit/test_plotting/test_plot.py::TestPlot::test_plot
```

I don't understand  failures with `tcl` and `anytree`.  They fail randomly on different envs.

agriyakhetarpal:
I think the `anytree` error in `import_optional_dependency` is related to an incorrect error message for the dependency (use use `anytree.exporter` specifically, but we import `anytree`). Could you try that test with changes from #3892 applied? It should get merged soon.

Re: the Matplotlib error, I'm not sure either. It looks like an unrelated error and is probably coming from `actions/setup-python`? Does the test work for you locally?

prady0t:
I did encounter `anytree` but not the `test_plot` error. So that might be the case.
Also we can remove `invalid escape sequence` failure by using raw string, instead of:

https://github.com/pybamm-team/PyBaMM/blob/39cff38cc7c424285761101c3e046e0d04af5dff/pybamm/expression_tree/operations/latexify.py#L82

we can do:
```py
        geo_latex = fr"\quad {rng_min} < {name} < {rng_max}"
```
Can we do that without any breaking changes?

prady0t:
We should have more number of checks with this.

prady0t:
How to get covergae report?
I tried running Running 
```
 python -m pytest --cov tests/unit  
``` 
but it's unable to find tests.
Also running 
```
coverage run -m unittest discover tests/unit report  
```
only gives me 20% coverage. What am I doing wrong?

Saransh-cpp:
The command should be:
```
pytest --cov=pybamm tests/unit
```
More on coverage with pytest here - https://learn.scientific-python.org/development/guides/coverage/

prady0t:
> The command should be:
> 
> ```
> pytest --cov=pybamm tests/unit
> ```
> 
> More on coverage with pytest here - https://learn.scientific-python.org/development/guides/coverage/

Oh wrong command. This was embarrassing. 🙃 
I'm now at 96% coverage.

prady0t:
Is the doctests failure related to changes in this PR? Do I need to integrate `xdoctest` for it to pass?

kratman:
> Is the doctests failure related to changes in this PR? Do I need to integrate `xdoctest` for it to pass?

The tests appear to be passing on other branches, so most likely it is something in your PR. If you really get stuck then I can try to help out

prady0t:
> > Is the doctests failure related to changes in this PR? Do I need to integrate `xdoctest` for it to pass?
> 
> The tests appear to be passing on other branches, so most likely it is something in your PR. If you really get stuck then I can try to help out

Let me try integrating pytest plugins for doctest and coverage. This might solve it.

agriyakhetarpal:
The `import pytest` error should get fixed if you add the `[dev]` set of optional dependencies to be installed in the `doctests` and the `docs` `nox` sessions.

prady0t:
Reason for failing tests in py3.12 https://www.python.org/downloads/release/python-3122/#:~:text=Invalid%20backslash%20escape,in%20the%20future.)

prady0t:
Most of the tests are passing now. 😄

agriyakhetarpal:
[This JAX test](https://github.com/pybamm-team/PyBaMM/actions/runs/8390154439/job/22978084660?pr=3857#step:11:3814) has previously failed for us sometimes and looks to be flaky – in a separate PR later on, we can explore some of the [available plugins](https://docs.pytest.org/en/stable/explanation/flaky.html#plugins) and decorate those with `@pytest.mark.flaky()`. For now, I re-triggered the failing job

prady0t:
> [This JAX test](https://github.com/pybamm-team/PyBaMM/actions/runs/8390154439/job/22978084660?pr=3857#step:11:3814) has previously failed for us sometimes and looks to be flaky – in a separate PR later on, we can explore some of the [available plugins](https://docs.pytest.org/en/stable/explanation/flaky.html#plugins) and decorate those with `@pytest.mark.flaky()`. For now, I re-triggered the failing job

Should I open an issue regarding this?

agriyakhetarpal:
The parallel testing looks great – GitHub Actions runners do not have a lot of resources so I guess running locally will reveal the speedups more prominently. I can review this in a while – for now, we will also require updates to the Contributing guide and source installation instructions wherever applicable (while mentioning that the unit tests use `pytest` and the integration tests still use `unittest`)

P.S. the failing Lychee job looks fixable, could you replace it with a different link, i.e., CasADi 2.0.0 isn't really relevant anymore and we could add links to the stable documentation?

agriyakhetarpal:
> Should I open an issue regarding this?

Sure! It would be good to track that

prady0t:
> The parallel testing looks great – GitHub Actions runners do not have a lot of resources so I guess running locally will reveal the speedups more prominently. I can review this in a while – for now, we will also require updates to the Contributing guide and source installation instructions wherever applicable (while mentioning that the unit tests use `pytest` and the integration tests still use `unittest`)
> 
> P.S. the failing Lychee job looks fixable, could you replace it with a different link, i.e., CasADi 2.0.0 isn't really relevant anymore and we could add links to the stable documentation?

Let me look into it. Also is there a reason why running tests on macOS based runners are generally slower? (for me running on macos, py 3.10 locally takes about 12-14 min which is still faster than the time taking by runners but is significantly lot when compared to linux and windows )

agriyakhetarpal:
> Also is there a reason why running tests on macOS based runners are generally slower? (for me running on macos, py 3.10 locally takes about 12-14 min which is still faster than the time taking by runners but is significantly lot when compared to linux and windows )

I ~don't~ _do_ think I'm able to reproduce that – up to 94% completion or so on macOS locally the test suite is quite fast, but then I face some heavy slowdowns when running the tests for the IDAKLU solver, IDAKLUJax tests, and the JAX BDF solver tests – and then with some test cases in `tests/unit/test_meshes/test_one_dimensional_submesh.py` and the `scikits.odes` solvers. But since these tests work fine in the serial implementation, could you try this solution in the meantime: https://github.com/pytest-dev/pytest-xdist/issues/385#issuecomment-1304877301? This is the same thread as the one I had shared earlier on Slack in the `infrastructure` channel when we were discussing `pytest-profiling`. I did see later on that JAX itself is using `pytest-xdist`, so it likely has something to do with our solver rather than being an upstream bug.

Also, it looks like the macOS Python 3.12 tests [are failing](https://github.com/pybamm-team/PyBaMM/actions/runs/8390154439/job/22984682748?pr=3857) for a different reason now, oops! :) Could you try fixing that?

kratman:
> okay, it passes now, but we have to ensure that the order of the tests is deterministic (i.e., `test_util.py::test_import_optional_dependency` and related tests should always be tested at the last).

The tests are parallel by default in PyTest, shouldn't run order be irrelevant? I would expect that PyTest is sort of sandboxing the tests.

If the optional dependencies tests are a problem, then they should not be run with the regular tests. That is more of an integration test than a unit test anyway.

kratman:
I am going to merge develop into this to see if the test failures happen again

run-tests.py
```diff
@@ -10,8 +10,9 @@
 import pybamm
 import sys
 import argparse
-import unittest
 import subprocess
+import pytest
+import unittest
```
kratman, line 15: How much extra work would it be to remove unittest so it becomes pure pytest?

tests/unit/test_expression_tree/test_operations/test_evaluate_python.py
```diff
@@ -283,9 +283,9 @@ def test_to_python(self):
         expr = a + b
         constant_str, variable_str = pybamm.to_python(expr)
         expected_str = (
-            "var_[0-9m]+ = y\[0:1\].*\\n"
-            "var_[0-9m]+ = y\[1:2\].*\\n"
-            "var_[0-9m]+ = var_[0-9m]+ \+ var_[0-9m]+"
+            "var_[0-9m]+ = y\\[0:1\\].*\\n"
+            "var_[0-9m]+ = y\\[1:2\\].*\\n"
+            "var_[0-9m]+ = var_[0-9m]+ \\+ var_[0-9m]+"
```
kratman, line 288: Can these be made into raw strings instead of manual escapes or does that break the regex?

tests/unit/test_expression_tree/test_operations/test_jac.py
```diff
@@ -10,7 +10,7 @@
 from tests import get_mesh_for_testing
 
 
-def test_multi_var_function(arg1, arg2):
+def multi_var_function_test(arg1, arg2):
```
kratman, line 13: Seems like these functions repeat a lot, can we move them to a utilities file instead of just renaming them all. Good to clean up a bit when stuff like this is found

kratman (changes requested):
Minor stuff

run-tests.py
kratman: It can be done in another ticket, I am just curious since right now we are using both libraries for testing

agriyakhetarpal: I guess I can answer this a bit – @prady0t had previously mentioned that at least half of the integration tests fail currently with `pytest` – so an estimated timeline could be one chunk of the duration of a standard twelve-week GSoC project (plus or minus a few weeks for any associated technical or non-technical reasons). The tests will have to be rewritten with new syntax either manually or through an AST-based parsing tool.

kratman: 12 weeks sounds like a lot, why is it so high?

agriyakhetarpal: We would need some other things from the project too. Some of them, off the top of my mind:

1.  Rewriting all the tests (unit too, besides integration) to pytest-style tests
2. Testing the built wheels (`pybamm.test()` similar to what is offered by other projects)
4. Improving test coverage with `hypothesis` (our code coverage is already high, but we could deal with a lot less verbosity when writing tests and check for any uncovered cases)
5. Adding pytest plugins based on ideas from other repositories

so all that does end up being a 175 hour project – and also, a potential mentees are mostly newcomers to open source software projects, so we need to give them time to explore the codebase and learn, so all that adds up

agriyakhetarpal: We would need some other things from the project too. Some of them, off the top of my mind:

1.  Rewriting all the tests (unit too, besides integration) to pytest-style tests
2. Testing the built wheels (`pybamm.test()` similar to what is offered by other projects)
4. Improving test coverage with `hypothesis` (our code coverage is already high, but we could deal with a lot less verbosity when writing tests and check for any uncovered cases)
5. Adding pytest plugins based on ideas from other repositories

so all that does end up being a 175 hour project – and also, a potential mentees are mostly newcomers to open source software projects :), so we need to give them time to explore the codebase and learn, so all that adds up

agriyakhetarpal:
To fix the failing Windows tests, let's use the Agg backend for Matplotlib to not open up a GUI. We can either the `MPLBACKEND` environment variable to the `PYBAMM_ENV` dictionary in  `noxfile.py`, or we can do

```python
from matplotlib import use
use("Agg")
```

wherever we have `matplotlib` being imported in the files of the `tests/` folder. The latter would be preferable because it would available without `nox` (i.e., with just `run-tests.py` invocations too).

pyproject.toml
```diff
@@ -249,6 +249,10 @@ filterwarnings = [
     # ignore internal nbmake warnings
     'ignore:unclosed \<socket.socket:ResourceWarning',
     'ignore:unclosed event loop \<:ResourceWarning',
+    # ignore warnings generated while running tests
+    "ignore::DeprecationWarning",
```
agriyakhetarpal, line 253: We should not ignore `DeprecationWarning`s, IMO – they serve as reminders for imminent API changes from dependencies upstream. We do not want our CI to break all of a sudden

pyproject.toml
```diff
@@ -249,6 +249,10 @@ filterwarnings = [
...
+    "ignore::UserWarning",
+    "ignore::RuntimeWarning",
```
agriyakhetarpal, line 255: We probably shouldn't ignore `UserWarning`s and `RuntimeWarning`s either, does not ignoring them fail the tests?

tests/unit/test_expression_tree/test_operations/test_evaluate_python.py
prady0t: Yes we can convert all of them to raw string without any breaking change.

tests/unit/test_expression_tree/test_operations/test_jac.py
prady0t: in `util.py` file right?

pyproject.toml
agriyakhetarpal: Oh, thanks the clarifications! We can revisit this sometime soon, then.

tests/unit/test_expression_tree/test_operations/test_jac.py
kratman: Yeah in a testing utils file, not a main pybamm utils file. Cutting down on duplicated code is just a good thing, especially when you already had to change the names of them all

tests/unit/test_expression_tree/test_operations/test_evaluate_python.py
kratman: Ok then I would go ahead and do it the raw string method

tests/unit/test_expression_tree/test_operations/test_jac.py
agriyakhetarpal: For reference, `tests/shared.py` is what we've been using for this purpose

prady0t: Done!

pyproject.toml
agriyakhetarpal: Unresolving this conversation, @prady0t – if we don't have a lot of `DeprecationWarning`s, could you fix them here? If there are many, then feel free to take that up in a follow-up PR.

agriyakhetarpal (approved):
This looks quite good on my end, and I have just a few additional comments. Let's merge after either @kratman or @Saransh-cpp can review again.

> I noticed that we set MPLBACKEND to "Template" in run-tests.py, which would have set that backend back again when running the tests. Let's switch that environment variable to Agg as well.

This got lost in the conversation, probably – could you please do this too?

kratman (approved):
I don't see any major issues here, however @agriyakhetarpal has an open discussion, so lets make sure to resolve that before we merge the changes from @prady0t

pyproject.toml
prady0t: There are 976 errors but for only one reason.  

DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html

We need to remove `pkg_resources`

agriyakhetarpal: Ah, right. We can't do that at this time due to `pybtex`

agriyakhetarpal:
Almost all good from my end, just the comment about removing the "Template" Matplotlib backend from `run-tests.py` needs to be addressed and a final CI run to be triggered, should be ready to merge after that.

arjxn-py (approved):
Nice work @prady0t, everything looks nice to me.

PR #3859, opened 2024-03-03
Enable Flake8-bugbear

### LINKED ISSUE #3854: Enable flake8-bugbear and resolve the issues that pop up
Recommended by scientific python.

https://github.com/pybamm-team/PyBaMM/blob/2f9f1e25a6f67c53631e13e7fbb500d984c0e301/pyproject.toml#L187

lorenzofavaro:
# Description

Flake8-bugbear has been enabled to improve code quality and make the codebase more consistent and compliant with style standards.

Fixes #3854 

Several style issues and deprecated practices were identified, which have been addressed in this PR:
- [X] [unused-loop-control-variable (B007)](https://docs.astral.sh/ruff/rules/unused-loop-control-variable/)
- [X] [function-uses-loop-variable (B023)](https://docs.astral.sh/ruff/rules/function-uses-loop-variable/)
- [X] [raise-without-from-inside-except (B904)](https://docs.astral.sh/ruff/rules/raise-without-from-inside-except/)
- [X] [mutable-argument-default (B006)](https://docs.astral.sh/ruff/rules/mutable-argument-default/)
- [X] [no-explicit-stacklevel (B028)](https://docs.astral.sh/ruff/rules/no-explicit-stacklevel/)
- [X] [get-attr-with-constant (B009)](https://docs.astral.sh/ruff/rules/get-attr-with-constant/)
- [X] [loop-variable-overrides-iterator (B020)](https://docs.astral.sh/ruff/rules/loop-variable-overrides-iterator/)
- [X] [unary-prefix-increment-decrement (B002)](https://docs.astral.sh/ruff/rules/unary-prefix-increment-decrement/)
- [X] [function-call-in-default-argument (B008)](https://docs.astral.sh/ruff/rules/function-call-in-default-argument/)
- [X] [cached-instance-method (B019)](https://docs.astral.sh/ruff/rules/cached-instance-method/)

## Type of change

# Key checklist:


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 
 ## Bug Fixes
 
+- Enabled Flake8-bugbear linter, fixing potential bugs, improving code quality and detecting design issues in advance ([#3854](https://github.com/pybamm-team/PyBaMM/issues/3854))
```
agriyakhetarpal, line 12: We can skip this note in the CHANGELOG. It is not a user-facing bug fix and affects just developers

pybamm/models/full_battery_models/lithium_ion/electrode_soh_half_cell.py
```diff
@@ -129,9 +129,7 @@ def get_initial_stoichiometry_half_cell(
     return x
 
 
-def get_min_max_stoichiometries(
-    parameter_values, options={"working electrode": "positive"}
-):
+def get_min_max_stoichiometries(parameter_values, options=None):
```
agriyakhetarpal, line 132: I am not sure I understand how this change came up (or which of the bugbear rules was not happy here). Also, if we have to keep `options` as an argument with no default value, would it make sense to add it to the docstring – where it it isn't present yet?

agriyakhetarpal (commented):
This looks good, @lorenzofavaro – thanks! Could you please add the commit hashes where each rule was fixed along with a comment to the `.git-blame-ignore-revs` file? 

There are ten commits here out of twelve where the rules were enforced (all atomic so that is good) – we could consider squashing them and then adding the hash for the single commit generated in a subsequent commit. I do not mind either of those options, but if someone else does have an opinion on this – please let me know :)

pybamm/models/full_battery_models/lithium_ion/electrode_soh_half_cell.py
lorenzofavaro: It is the [mutable-argument-default (B006)](https://docs.astral.sh/ruff/rules/mutable-argument-default/), basically to avoid unexpected behaviour because the same mutable object would be shared across all calls to the function.
Yep, I've now added `options` to the docstring.

pybamm/expression_tree/concatenations.py
```diff
@@ -338,7 +338,7 @@ def create_slices(self, node: pybamm.Symbol) -> defaultdict:
                 """Concatenation and children must have the same number of
                 points in secondary dimensions"""
             )
-        for i in range(second_pts):
+        for _i in range(second_pts):
```
kratman, line 341: Is this setting optional? Do all loop variables need this now?

agriyakhetarpal: This should be the rule: https://docs.astral.sh/ruff/rules/unused-loop-control-variable/. I don't think all loop variables need it and it may not be required in things like notebooks, but for a lengthy file with multiple classes and methods I think it should be fine

kratman: I think if it is just for unused variables then it should be "_" instead of "_i", which is pretty common. Using "_i" just seems weird if there is only a single variable. Not sure how configurable it is though.

When I saw it I was concerned that it was done for scope issues, which are better resolved by shortening functions.

agriyakhetarpal: You're right. `_i` seems to be more like a private variable – not something that is meant to be discarded after an iterable has been called over, `_` is better for the latter

scripts/install_KLU_Sundials.py
```diff
@@ -268,11 +268,11 @@ def parallel_download(urls, download_dir):
 try:
     subprocess.run(["make", "--version"])
 except OSError:
-    raise RuntimeError("Make must be installed.")
+    raise RuntimeError("Make must be installed.") from None
```
Saransh-cpp, line 271: These should not be `from None` as that defeats the point of the check. IIRC, the best practice should look like this -
```py
try:
    ...
except ErrorType as e:
    raise NewErrorType(...) from e
```
This will preserve the stack trace, making it easier to debug for both the users and pybamm developers.

pybamm/solvers/idaklu_jax.py
```diff
@@ -344,10 +344,11 @@ class _hashabledict(dict):
         def __hash__(self):
             return hash(tuple(sorted(self.items())))
 
+    @staticmethod
     @lru_cache(maxsize=1)
-    def _cached_solve(self, model, t_hashable, *args, **kwargs):
+    def _cached_solve(solver, model, t_hashable, *args, **kwargs):
```
Saransh-cpp, line 349: This PR should really not modify the API. Feel free to ignore a few errors using `pyproject.toml` if they look unfixable. Not all the errors reported can be fixed.

Saransh-cpp (changes requested):
Thanks, @lorenzofavaro! See the comments below -

tests/unit/test_solvers/test_processed_variable.py
```diff
@@ -28,9 +28,11 @@ def to_casadi(var_pybamm, y, inputs=None):
 
 
 def process_and_check_2D_variable(
-    var, first_spatial_var, second_spatial_var, disc=None, geometry_options={}
+    var, first_spatial_var, second_spatial_var, disc=None, geometry_options=None
 ):
     # first_spatial_var should be on the "smaller" domain, i.e "r" for an "r-x" variable
+    if geometry_options is None:
+        geometry_options = {}
```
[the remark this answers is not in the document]
lorenzofavaro: The violated rule is [mutable-argument-default (B006)](https://docs.astral.sh/ruff/rules/mutable-argument-default/), basically function defaults are evaluated once so if you have a mutable object, every time you modify it, its modifications will persist across calls. It's to avoid unexpected behavior, although we are in unit tests.

If we want we could ignore it using `# noqa: B006`:

```python
def process_and_check_2D_variable(
    var, first_spatial_var, second_spatial_var, disc=None, geometry_options={}  # noqa: B006
):
```

Or if we want to ignore it at project level, we could add it in `pyproject.toml`:
```toml
ignore = [
    # ...
    "B006",        # Uses of mutable objects as function argument defaults
]
```

scripts/install_KLU_Sundials.py
lorenzofavaro: Right, my only concern was not to make the error message redundant, but it makes sense. I have fixed it now in 03b3a65f72c0f0a69ce6baa5c1f36ac1f3a75367

pybamm/solvers/idaklu_jax.py
lorenzofavaro: Ignored this single check with `# noqa: B019` in 165e9942098c7b14f44be94e9850ed19356fb0ff.

kratman:
@lorenzofavaro I resolved the conflict so that it would trigger CI

pybamm/expression_tree/concatenations.py
lorenzofavaro: I have corrected the naming convention for this in f6b7b8cc11b6290737e694ac92ca4a8e93ba4138

kratman (approved):
I think I am pretty happy with what is in here. Looks like a bunch of good fixes.

@Saransh-cpp Can you re-review?

pybamm/expression_tree/symbol.py
```diff
@@ -901,7 +901,6 @@ def evaluates_to_number(self):
     def evaluates_to_constant_number(self):
         return self.evaluates_to_number() and self.is_constant()
 
-    @lru_cache
```
arjxn-py, line 904: Is this also removed while linting?

arjxn-py (approved):
Thanks @lorenzofavaro, nice work.
Looks Good to me.

pybamm/expression_tree/symbol.py
lorenzofavaro: Yes, to resolve the violation of the [rule cached-instance-method (B019)](https://docs.astral.sh/ruff/rules/cached-instance-method/). 

The `lru_cache` here is redundant as the function already handles the storage of previous results in `self._saved_evaluates_on_edges[dimension]`, which are returned in $O(1)$ time.

pybamm/solvers/idaklu_jax.py
```diff
@@ -370,7 +370,8 @@ def _jaxify_solve(self, t, invar, *inputs_values):
         logger.debug(f"  invar: {invar}")
         logger.debug(f"  inputs: {dict(d)}")
         logger.debug(f"  calculate_sensitivities: {invar is not None}")
-        sim = self._cached_solve(
+        sim = IDAKLUJax._cached_solve(
+            self.solver,
```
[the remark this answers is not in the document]
lorenzofavaro: It's the rule [cached-instance-method (B019)](https://docs.astral.sh/ruff/rules/cached-instance-method/). Its purpose is to avoid memory leaks, caused by the `@lru_cache` decorator on methods, as the global cache will keep a reference to the instance preventing it from being garbage collected.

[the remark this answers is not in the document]
Saransh-cpp: Thanks! Looks good to me!

Saransh-cpp (approved):

agriyakhetarpal (commented):
Thanks for the reviews, everyone – and thanks for contributing, @lorenzofavaro. All we need is to add the relevant commit hashes along with a comment listing the fixed rules (or better squash the commits and add a single hash) to the `.git-blame-ignore-revs` file – it looks like it was removed for some reason.

>[!NOTE]
> In case someone else gets to merge this before I do – please do not squash-merge the PR.

Saransh-cpp:
> All we need is to add the relevant commit hashes along with a comment listing the fixed rules (or better squash the relevant commits and add a single hash) to the .git-blame-ignore-revs file

We can skip this, given that the changes in this PR are not the typical formatting changes but actual code changes.

Saransh-cpp:
So this can be squashed and merged

agriyakhetarpal (approved):

agriyakhetarpal:
The squashed commit has a lot of attributed authors (😄), possibly due to the rebase from `develop` being out of sync. At least it's just a single commit.

Issue #2427, opened 2022-11-03
Make private functions and classes more explicit


tinosulzer:
Make it clearer which methods should not be accessed by users, by adding an underscore at the start of their name

Saransh-cpp:
> We should also look into adding \_\_all\_\_ to pybamm modules

Copied from the PR linked above

agriyakhetarpal:
There was an attempt by @arjxn-py on adding `__all__` with the `mkinit` CLI-based package to speed up the PyBaMM import, but there were problems with that approach (#3732) because it would be messy to add it to several files at once and generate it everytime. Here is an older thread I found, probably not relevant anymore: #2436 

It would make sense to start with smaller modules like `util.py`, `doc_utils.py`, `install_odes.py` (the latter should just have one method of usage – through the entry point). Also, `version.py` should be a private module too (`pybamm._version.__version__` instead of `pybamm.version.__version__`), and it could be altered in favour of using `importlib.metadata.version()` to fetch the version from declarative metadata – in 2024 we should not need to import a package to obtain its version

arjxn-py:
It should be possible to add `__all__` only as well with the help of `mkinit`, how'd we want to have it though - 
- have `__all__` added to all the submodules? (**Not Recommended** - as it would again be messy with several file changes)
- have a recursive `__all__` added to the root `__init__.py` (**Recommended** - See [this commit](https://github.com/arjxn-py/PyBaMM/commit/175788523975a8fbd0200a0e0bc2b14e721075d7))
- have a non-recursive `__all__` added to the root `__init__.py` (See [this commit](https://github.com/arjxn-py/PyBaMM/commit/d71a8a348c0b946e80f8855be18fb18bd8d59600), Makes lesser sense as it only includes direct attributes of pybamm)

agriyakhetarpal:
The second one makes the most sense to me, it should decide the namespace when we import PyBaMM.

The third one would need a bit more refining because we have things like `update_LD_LIBRARY_PATH` which should not be exposed

arjxn-py:
> The third one would need a bit more refining because we have things like `update_LD_LIBRARY_PATH` which should not be exposed

We also have these attributes in the second too, as ultimately: 
- First is the superset of Second & Third & 
- Second is the superset of Third

Currently Protected modules are exposed, but their attributes are not & Private modules and their attributes are not exposed.

[...]

PR #3866, opened 2024-03-05
Rename have_optional_dependency

### LINKED ISSUE #3864: Rename `have_optional_dependency` to `import_optional_dependency`
We use `have_optional_dependency()` to import optional dependencies across codebase which is a wrapper over `importlib.import_module()` to throw a helpful error message.

kratman:
# Description

Renames `have_optional_dependency()` to `import_optional_dependency()`

Fixes #3864 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/util.py
+++ b/pybamm/util.py
@@ -352,25 +352,19 @@ def install_jax(arguments=None):  # pragma: no cover
     )
 
 
-# https://docs.pybamm.org/en/latest/source/user_guide/contributing.html#managing-optional-dependencies-and-their-imports
-def have_optional_dependency(module_name, attribute=None):
+def import_optional_dependency(module_name, attribute=None):
     err_msg = f"Optional dependency {module_name} is not available. See https://docs.pybamm.org/en/latest/source/user_guide/installation/index.html#optional-dependencies for more details."
     try:
-        # Attempt to import the specified module
         module = importlib.import_module(module_name)
-
         if attribute:
-            # If an attribute is specified, check if it's available
             if hasattr(module, attribute):
                 imported_attribute = getattr(module, attribute)
-                return imported_attribute  # Return the imported attribute
+                # Return the imported attribute
+                return imported_attribute
             else:
-                # Raise an ModuleNotFoundError if the attribute is not available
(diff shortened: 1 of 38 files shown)


agriyakhetarpal (approved):
Thanks, @kratman! For next time we should look at removing such functions from the public API – I don't think it is being used by anyone currently and does not provide any useful functionality to users either.

kratman:
> Thanks, @kratman! For next time we should look at removing such functions from the public API – I don't think it is being used by anyone currently and does not provide any useful functionality to users either.

Yeah, python's handling of private methods makes me miss C++

Saransh-cpp (approved):
Thanks! We should also look into adding `__all__` to pybamm modules; not in this PR though.

kratman:
@arjxn-py, @Saransh-cpp I saw this in `latexify()`
```
        if sympy:
            from pybamm.expression_tree.operations.latexify import Latexify
```
I assume this no longer make sense here either. There used to be an "import_optional_dependency" above it

pybamm/util.py
```diff
@@ -352,25 +352,19 @@ def install_jax(arguments=None):  # pragma: no cover
     )
 
 
-# https://docs.pybamm.org/en/latest/source/user_guide/contributing.html#managing-optional-dependencies-and-their-imports
-def have_optional_dependency(module_name, attribute=None):
+def import_optional_dependency(module_name, attribute=None):
```
arjxn-py, line 355:
```suggestion
# https://docs.pybamm.org/en/latest/source/user_guide/contributing.html#managing-optional-dependencies-and-their-imports
def import_optional_dependency(module_name, attribute=None):
```
Not removing this should be fine

arjxn-py (commented):
Looks good to me, thanks

Saransh-cpp:
> I assume this no longer make sense here either. There used to be an "import_optional_dependency" above it

Yes! IIRC we did do some hacky stuff to make sympy optional, maybe this was left behind.

agriyakhetarpal:
> We should also look into adding __all__ to pybamm modules; not in this PR though.

I agree, xref: #2427 for public and private API discussions

PR #3874, opened 2024-03-07
Migrated docker image from miniconda to manylinux2014

### LINKED ISSUE #3666: Reduce the number of Docker Images to just one image
Currently we build and push 5 variants of Docker Images regularly i.e. `latest`, `jax`, `odes`, `idaklu` & `all`.
The idea behind this issue is to have a single standard image variant to be updated regularly to save resources and make things simple for developers.
As discussed in #3661

### LINKED ISSUE #3692: Migrate from `conda` to `venv` inside Docker Images
This would potentially cut down an external system dependency `conda` & would reduce the inconsistency of environments across different OS for IDAKLU solver images.

santacodes:
# Description

Migration from conda virtual environment to venv and using Centos 7 based manylinux2014 as base image for docker which is compliant with [PEP599](https://peps.python.org/pep-0599/).

Fixes #3692 and fixes #3666 

## Type of change

# Key checklist:

## Further checks:


scripts/Dockerfile
```diff
@@ -16,54 +15,34 @@ WORKDIR /home/pybamm/
...
-RUN if [ "$ALL" = "true" ]; then \
-    python scripts/install_KLU_Sundials.py && \
-    rm -rf pybind11 && \
-    git clone https://github.com/pybind/pybind11.git && \
-    pip install --user -e ".[all,dev,docs,jax,odes]"; \
-    fi
-
-RUN if [ -z "$IDAKLU" ] \
-    && [ -z "$ODES" ] \
-    && [ -z "$JAX" ] \
```
Saransh-cpp, line 64: These options should not have been removed. Was there a reason for removing them?

Saransh-cpp (changes requested):
Thanks for taking this up, @santacodes!

scripts/Dockerfile
agriyakhetarpal: We decided to switch to a single Docker image containing all solvers in the last developer meeting on 19/02/2024 – please see #3666 for context

santacodes: These options have been removed to perform a full installation with all the solvers inside the container

santacodes:
> Also, could we get an estimate on the compressed size of the image? The ones we currently push are ~1.6 GB, so remaining close to somewhere around that would be great.

Sure I'd try to minimize the image size as much as possible and update you on the estimate

arjxn-py (commented):
python base image looks much more valid as compared to manylinux, I'd request you to run test suite inside the running docker container at least once and see how many tests are getting skipped.
Plus confirm the solver availability inside the container by executing:
- `pybamm.have_idaklu()`
- `pybamm.have_jax()` 
- `pybamm.have_scikits_odes()`

santacodes:
> python base image looks much more valid as compared to manylinux, I'd request you to run test suite inside the running docker container at least once and see how many tests are getting skipped. Plus confirm the solver availability inside the container by executing:
> 
> * `pybamm.have_idaklu()`
> * `pybamm.have_jax()`
> * `pybamm.have_scikits_odes()`

All of the three methods return true inside the docker container running python venv. I also did run the test suite before pushing the commit, 1 test in each unit and integration tests were getting skipped, I am not exactly sure of the cause. I am awaiting on @agriyakhetarpal response on testing it on ARM.

santacodes:
Upon further researching on [PEP599](https://peps.python.org/pep-0599/) the manylinux2014 platform tag which demands for the container to be compliant with `GLIBC<=2.17`, almost all of the linux distributions fulfilling that constraint have reached EOL, except CentOS 7 which will reach its EOL in June, 2024. 

Most of the docker containers which have a constant maintenance use Debian/Ubuntu, RHEL or Alpine based distributions in their containers to maintain their integrity.

In reference to the above distributions, these are the OS releases that fulfill `glibc<=2.17` - 

- **Debian Wheezy** which had glibc 2.13 and reached its end of life in May 2018. The packages are now archived and not constantly maintained which might cause some packages to break.
- **Ubuntu 13 Saucy** which had glibc 2.17 and reached its EOL in July 2014. 
- **CentOS 7** which still runs glibc 2.17 will reach EOL in June 2024 is the primary base image used by PEP599 manylinux2014 policy, but the downside of it in context to this project is multi-architecture building. Manylinux2014 provides `A single build image and auditwheel profile per architecture` which would mean that: 
  -  We would either have to use 2 different dockerfiles to build for x86_64 and aarch64 distinctly as the base images have a discrete architecture or 
  - Use a multistage build to pull and push different images for each of them. 
  This would also complicate things a bit as CentOS 7 would reach EOL soon. 

To not do things the hacky way the official [Python Bullseye](https://hub.docker.com/_/python) image would make things a bit simple as the base image itself supports multiple architectures and is based on Debian. The size would also not be an issue as there's a slim version of the image and would have a constant support and maintenance for a while. Upgrading to a newer version or migrating would also become easier as the dependencies would be compatible. By default the Python Bullseye image provides `glibc 2.31` which provides backward compatibility and wouldn't have any issues running in a container.


[Reference](https://distrowatch.com/table.php?distribution=centos)

agriyakhetarpal:
Thanks for looking into this @santacodes  – yes, a Python bullseye image makes sense too as @arjxn-py mentions. The manylinux2014 images are indeed architecture-specific implementations of PEP 599 and therefore won't have support for multiple architectures from a single `FROM` command.

@santacodes when pulling this branch I noticed that these changes have been made on the `develop` branch on your for, and not a feature branch. This is generally [not good practice](https://hynek.me/articles/pull-requests-branch/) since it causes issues with syncing the repository if you wish to contribute further, and a host of other reasons.

Could you create a new PR based on these changes, and close this one? You can check out a new branch from the current changes and that should bring your commits there (or you can cherry-pick the commits too – whichever you find easier).

agriyakhetarpal:
Actually, testing this on ARM on macOS has revealed a bug in our current PyBaMM Docker images that we have been pushing to Docker Hub since some while:

```
```
ImportError                               Traceback (most recent call last)
Cell In[4], line 1
----> 1 idaklu = importlib.util.module_from_spec(idaklu_spec)

File <frozen importlib._bootstrap>:573, in module_from_spec(spec)

File <frozen importlib._bootstrap_external>:1233, in create_module(self, spec)

File <frozen importlib._bootstrap>:241, in _call_with_frames_removed(f, *args, **kwds)

ImportError: /home/pybamm/PyBaMM/pybamm/solvers/idaklu.cpython-311-aarch64-linux-gnu.so: undefined symbol: _ZN6casadi8Function11deserializeERKSs
```

this unresolved symbol comes from CasADi upstream (see https://github.com/casadi/casadi/issues/2887, https://github.com/dockcross/dockcross/issues/753, https://github.com/dockcross/dockcross/issues/780) and is quite challenging to fix. This error is the same as what I face when testing this new `bullseye` image – I remember facing this on #3462 and could not find a way towards resolution.

We can keep this PR open for a while – what I remember is that this did work on aarch64 earlier, but somewhere down the line this ended up breaking and therefore the IDAKLU solver cannot be instantiated

Issue #3662, opened 2023-12-25
Too many failing checks / implement smarter CI


adityaraute:
### Description

I was scrolling through the PRs and noticed most PRs failing some checks or the others; often the same ones.

I've attached a screenshot where a missing package led to a check fail, but there are several similar instances.

### Motivation

If checks are failing, and yet they may not be related to the issue at hand, I wonder if it makes the Check or Action itself irrelevant, at least to the majority of PRs. 

### Possible Implementation

Reduce the number of actions/checks attached to certain workflows, such that only the ones that are essential for a PR are run.
Additionally, investigate why certain actions fail repeatedly and rectify any issues in the core repo build.

### Additional context

An example of an action running ubuntu tests on PRs

![image](https://github.com/pybamm-team/PyBaMM/assets/43912470/4f840774-327d-4ce1-afb8-96938aea581e)

agriyakhetarpal:
Thanks for the suggestion, I have myself had a variety of ideas like this in mind since some while based off on recommendations from other repositories. Adding labels to check like `[skip tests]` and `[skip docs]` to group tests would be worthwhile and is super easy to add via already-supported GitHub Actions syntax, but the issue with this is about how maintainable this can be in the long term, since the file globs to conditionally check are too complex to accommodate all such cases. For example, if I end up modifying a docstring incorrectly in the `pybamm/` folder and skip the doctests in a PR as there will be no changes in `docs/`, it would mean that failures in the doctests based on that incorrect change may then start to show up elsewhere in future PRs, which would become troublesome for the PR author or maintainers to fix every time. We will need to decide on how to go about this

I suppose skipping the benchmark runs on documentation-only fixes would be fine. It is the longest job and we don't need it to run everywhere, especially if we're not pushing changes to an external repository like we do with a CRON job. I'm aware that SciPy's CI configuration has a lot of such things we can seek inspiration from; for example, they check for the `[Build wheels]` message in a commit which proves to be useful for debugging when working on compiled code in a PR – and so on. Feel free to pitch in @arjxn-py @Saransh-cpp. If you have any suggestions w.r.t. this based on what you have observed with PyBaMM's infrastructure, please feel free to list them in this thread, @adityaraute. 

There is indeed a lot of redundancy with things like repeated workflows and workflow files (the scheduled tests and regular CI tests could be integrated into the same file, #3518 is still open, benchmarks-related jobs can have a single file, etc.)

<hr>

[...]

PR #3876, opened 2024-03-08
Disable benchmarks on PRs

kratman:
# Description

Stop the benchmarks from running on PRs. The reasons for this are the frequent failures and the long runtime. Benchmarks are still run daily and for main/develop

Related: #3662

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/.github/workflows/benchmark_on_push.yml
+++ b/.github/workflows/benchmark_on_push.yml
@@ -2,7 +2,6 @@ name: Run benchmarks on push
 on:
   push:
     branches: [main, develop]
-  pull_request:
 
 concurrency:
   # Cancel intermediate builds always
--- a/.github/workflows/periodic_benchmarks.yml
+++ b/.github/workflows/periodic_benchmarks.yml
@@ -8,7 +8,7 @@
 # - Publish website
 name: Benchmarks
 on:
-  # Everyday at 3 am UTC
+  # Every day at 3 am UTC
   schedule:
     - cron: "0 3 * * *"
   # Make it possible to trigger the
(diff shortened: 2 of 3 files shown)


tinosulzer (approved):

agriyakhetarpal:
I would say that completely stopping benchmark runs is not a good idea in this case. We could fairly easily write a check that skips running benchmarks on PRs by default, but does run them if `[run bench]` exists in the commit message – it won't be too hard to maintain either. A lot of repositories do it and we can document this somewhere in the Testing section of the Contributing guide.

This should be the required one-liner condition (off the top of my head):

```yaml
if: ${{ contains(github.event.head_commit.message, '[run bench]') }}
```

kratman:
> I would say that completely stopping benchmark runs is not a good idea in this case.

They are not stopped completely. They run every day at 3am and run after a PR is merged into main and develop. This only stops them in PRs.

We also have workflows for manual benchmarks. This just speeds it up and stops a frequent failure. I don't think we need to complicated it further by adding extra checks

agriyakhetarpal:
Yes, they aren't stopped completely, but they are disabled on PRs this way. My opinion is that there should be a way to run the benchmarks on a PR if one needs it, through this condition – and making it explicitly stated how to enable them if needed while reasonably keeping defaults to mark them as neutral/skipped (which speeds up the PR checks too as you mentioned)

The manual benchmarks workflow will have to be triggered manually from a branch through the Actions tab, which isn't convenient if you want to experiment with something – this is where commit messages can help

agriyakhetarpal:
Hi @BradyPlanden, please let us know if we can get started on adding a self-hosted runner for the benchmarks (that will now run nightly and for changes from the the `develop` branch's context)

PR #2547, opened 2022-12-13
Fix citations on Colab + bump v22.11.1

Saransh-cpp:
# Description

Fixes #2543

## Type of change

Adds try-except blocks everywhere the entry points might error out. See the review comments below -

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (back-end change that speeds up the code)
- [ ] Bug fix (non-breaking change which fixes an issue)

# Key checklist:

- [ ] No style issues: `$ flake8`
- [ ] All tests pass: `$ python run-tests.py --unit`
- [ ] The documentation builds: `$ cd docs` and then `$ make clean; make html`

You can run all three at once, using `$ python run-tests.py --quick`.

## Further checks:

- [ ] Code is commented, particularly in hard-to-understand areas
- [ ] Tests added that prove fix is effective or that feature works

PR #3886, opened 2024-03-11
Removed Google colab error warning.

prady0t:
# Description

Follow-up of #3740  

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:


agriyakhetarpal:
Thanks @prady0t, could you remove these lines:

```python
# Temporary fix for Python 3.12 CI. TODO: remove after
# https://bitbucket.org/pybtex-devs/pybtex/issues/169/replace-pkg_resources-with
# is fixed
```

wherever they exist in `noxfile.py`?

prady0t:
Sorry for making changes lazily. 🙂  I've updated the PR.

pybamm/citations.py
```diff
@@ -256,9 +256,7 @@ def print_citations(filename=None, output_format="text", verbose=False):
     """See :meth:`Citations.print`"""
     if citations._citation_err_msg is not None:
         raise ImportError(
-            f"Citations could not be registered. If you are on Google Colab - "
-            "pybtex does not work with Google Colab due to a known bug - "
-            "https://bitbucket.org/pybtex-devs/pybtex/issues/148/. "
+            f"Citations could not be registered."
```
Saransh-cpp, line 259: Could you please take a look at https://github.com/pybamm-team/PyBaMM/pull/2547 and just revert all the changes in the PR? That way we can get rid of some dead code too.

Saransh-cpp (changes requested):
Thanks, @prady0t! This looks nice! See my comment below -

agriyakhetarpal:
Also, it looks like the doctests on Python 3.12 are failing (strange) but all other jobs on Python 3.12 pass. Could we re-add the `setuptools` installation to the doctests session for now?

kratman:
> Also, it looks like the doctests on Python 3.12 are failing (strange) but all other jobs on Python 3.12 pass. Could we re-add the `setuptools` installation to the doctests session for now?

It could be a latex parsing issue: `SyntaxWarning: invalid escape sequence '\m'` is caused by `self.spatial_unit = "$\mu$m"` Maybe we need to make them raw strings or something:
```
self.spatial_unit = r"$\mu$m"
```

pybamm/citations.py
prady0t: Feel like I'm out of sync here. Why do we need to do this?

agriyakhetarpal: This was because we had to check if `pybtex` was working on every `pybamm.print_citations()` or `citations.register()` call. Since it now works on Google Colab, we don't need to keep the try-except blocks anymore.

prady0t:
I've re added setuptools installation to the doctests session and removed checks for error messages.

Saransh-cpp (approved):
Thanks, @prady0t! This looks good to me from the `pybtex` point of view. I'll leave the setuptools bit for @agriyakhetarpal to review.

agriyakhetarpal (commented):
The doctests and example scripts are still failing on Python 3.12, and I'm not sure why. We have two options:

1. Add the `session.install("setuptools")` hack back again for the time being
2. Modify these sessions in the workflows to run on Python 3.11 instead.

I would go forward with the first option at this time, because using the latest available Python version will help us catch any related (or unrelated) issues early.

kratman:
> The doctests and example scripts are still failing on Python 3.12, and I'm not sure why. We have two options:
> 
> 1. Add the `session.install("setuptools")` hack back again for the time being
> 2. Modify these sessions in the workflows to run on Python 3.11 instead.
> 
> I would go forward with the first option at this time, because using the latest available Python version will help us catch any related (or unrelated) issues early.

Not all of our functionality works on 3.12, so 3.11 is probably better

agriyakhetarpal:
> Not all of our functionality works on 3.12, so 3.11 is probably better

Yes, I had considered that when posting my comment – I had checked the example scripts and none of them are currently using unsupported functionality (which is `scikits.odes` for Python 3.12). So Python 3.12 would be fine (for the docs-related tests, it also helps catch warnings/errors early because most documentation dependencies are pure-Python and do not have an upper bound on their Python version requirement). So we can go ahead with either of those, but I tend to lean slightly on 3.12. Both Python versions have a long time till their EOL.

kratman:
> I tend to lean slightly on 3.12. Both Python versions have a long time till their EOL.

I would say that it is better to go with a version where everything runs, we can always upgrade the version later. It is for the documentation generation, not something that must be the newest python

agriyakhetarpal:
> I would say that it is better to go with a version where everything runs, we can always upgrade the version later. It is for the documentation generation, not something that must be the newest python

Sure, let's do it for just the doctests, but keep Python 3.12 for the example scripts (because it matches with the example notebooks tests).

prady0t:
would doing 
```
@nox.session(name="doctests", python="3.11")

```
work?

agriyakhetarpal:
> would doing
> 
> ```
> @nox.session(name="doctests", python="3.11")
> ```
> 
> work?

You don't need to change the `nox` configuration for the doctests. We just need to fix the CI and we should be good to go – just alter `test_on_push.yml` to use Python 3.11 in the doctests job instead of Python 3.12. The only change required in `noxfile.py` is to bring back the `session.install("setuptools")` line that you had removed from the example scripts session.

agriyakhetarpal (approved):
Thanks, @prady0t! All good now. As discussed, let's revisit `python-bibtexparser` at some other time.

PR #3891, opened 2024-03-13
Added a schedule on `needs-reply remove` workflow 

### LINKED ISSUE #3743: Conserve `needs-reply` GitHub Actions workflow runs
The `needs-reply` workflow currently runs on all issue comments or PR comments and it runs too much: https://github.com/pybamm-team/PyBaMM/actions/workflows/need_reply_remove.yml

I am not sure as to how to fix this as of now, and we don't mind it at all (GitHub Actions is completely free for public repositories) but it would be great if the runs for this workflow could be limited in some way or form. Some suggestions could be to:

1. Run only on stale conversations or on different events
2. Switch to a different GitHub App or GitHub Action
3. Skip the workflow on PRs
4. Remove the workflow altogether
5. and so on

santacodes:
# Description

Added a schedule on the [need_reply_remove](https://github.com/pybamm-team/PyBaMM/actions/workflows/need_reply_remove.yml) workflow on every Monday 03:00 hours to limit the resources that get utilized while running this action.

Fixes #3743 

## Type of change

# Key checklist:

## Further checks:


.github/workflows/need_reply_remove.yml
```diff
@@ -12,6 +14,7 @@ jobs:
       github.event.comment.author_association != 'OWNER' &&
       github.event.comment.author_association != 'COLLABORATOR' &&
       github.repository_owner == 'pybamm-team'
+      !github.event.issue.pull_request
```
agriyakhetarpal, line 17: This seems to be missing an AND operator (`&&`) in the previous line. Also, I would suggest using checking for the event name instead – something like `github.event_name != 'pull_request'` should work.

agriyakhetarpal (commented):
Thanks, @santacodes – I have a minor comment below, and this should be good to merge after that.

santacodes:
Sure, sorry for the messy branches

.github/workflows/need_reply_remove.yml
santacodes: whoops sorry for that I will add it also regarding the event name I referred to https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issue_comment to make it not run on PR comments. Though I could change the trigger to the event name itself.

agriyakhetarpal: Thanks – your solution might need to be wrapped with `${{ ... }}`, i.e., curly brackets? Yes, both of these ways would be equivalent, I think.

.github/workflows/need_reply_remove.yml
```diff
@@ -11,6 +11,8 @@ jobs:
     if: |
       github.event.comment.author_association != 'OWNER' &&
       github.event.comment.author_association != 'COLLABORATOR' &&
+      github.repository_owner == 'pybamm-team' &&
```
arjxn-py, line 14:
```suggestion
      github.repository_owner == 'pybamm-team' &&
```
This shuld be causing the style failure.

arjxn-py (changes requested):

.github/workflows/need_reply_remove.yml
arjxn-py: There's a style issue related to whitespaces in `need_reply_remove.yml` as well which also should be fixed with it.

.github/workflows/need_reply_remove.yml
agriyakhetarpal: @santacodes, to run the style checks you can do `pip install pre-commit` and `pre-commit install` to install hooks in your `.git/` folder. This will stop commits from being pushed that do not comply with the style guidelines. Another way is to run `nox -s pre-commit`, which is a little expensive.

arjxn-py:
Thanks for this @santacodes, should be good to merge once you address these minor issues.
No worries for the commit history for now, make sure to take care of it for further contributions 🙂

arjxn-py (approved):
Looks good to me, thanks @santacodes

PR #3914, opened 2024-03-20
Added more Linux distribution specific installation procedures

### LINKED ISSUE #3898: Update source installation instructions for linux to include more distro-specific instructions
Currently the install from source and other areas of installation related references in the documentation lack linux distribution specific steps. Linux distributions can have different package managers and therefore different package names, so to generalize them under `GNU/Linux` wouldn't be specific as it references the linux base kernel.
Instead we could add the following distribution specific steps as they are the most commonly used among developers and personal users.

- Ubuntu/Debian (apt)
- RHEL/Fedora (yum/rpm/dnf)
- Arch Linux (pacman)
- NixOS (nix)
- Windows (chocolatey) if applicable

Most of the other distributions are mostly built around these base distributions, so adding these package managers to the documentation would also cover distributions which use the mentioned package managers.

santacodes:
# Description

Install from source and other prerequisites dependency installation procedures lacked distribution specific steps. Common Linux package managers (`apt`, `dnf`, `pacman`) specific command tabs are added to the documentation.

Fixes #3898 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/docs/source/user_guide/installation/gnu-linux-mac.rst
+++ b/docs/source/user_guide/installation/gnu-linux-mac.rst
@@ -140,7 +140,7 @@ Users can install `scikits.odes <https://github.com/bmcage/odes>`__ to utilize i
 
    .. code:: bash
 
-      dnf install openblas-devel
+      sudo dnf install openblas-devel cmake
       pybamm_install_odes
 
    This will compile and install SUNDIALS for the system (under ``~/.local``), before installing ``scikits.odes``. (Alternatively, one can install SUNDIALS without this script and run ``pip install pybamm[odes]`` to install ``pybamm`` with ``scikits.odes``.)
--- a/docs/source/user_guide/installation/gnu-linux-mac.rst
+++ b/docs/source/user_guide/installation/gnu-linux-mac.rst
@@ -122,7 +122,7 @@ Users can install `scikits.odes <https://github.com/bmcage/odes>`__ to utilize i
 
    .. code:: bash
 
-      brew install openblas gcc gfortran
+      brew install openblas gcc gfortran cmake
       pybamm_install_odes
 
 The ``pybamm_install_odes`` command, installed with PyBaMM, automatically downloads and installs the SUNDIALS library on your
--- a/docs/source/user_guide/installation/gnu-linux-mac.rst
+++ b/docs/source/user_guide/installation/gnu-linux-mac.rst
@@ -14,7 +14,7 @@ To use PyBaMM, you must have Python 3.8, 3.9, 3.10, 3.11, or 3.12 installed.
 
    .. code:: bash
 
-      sudo apt update
+      sudo apt update \
       sudo apt install python3
 
 .. tab:: Fedora/CentOS
@@ -23,7 +23,17 @@ To use PyBaMM, you must have Python 3.8, 3.9, 3.10, 3.11, or 3.12 installed.
 
    .. code:: bash
 
+      sudo dnf upgrade \
       sudo dnf install python3
+
+.. tab:: Arch Linux
+
+  On Arch based systems you can use the pacman package manager. For example
+
+  .. code:: bash
+
+      sudo pacman -Syu \
+      sudo pacman -S python
 
 .. tab:: macOS
 
@@ -39,6 +49,7 @@ To use PyBaMM, you must have Python 3.8, 3.9, 3.10, 3.11, or 3.12 installed.
    .. code:: bash
 
       brew install python3
+
 
 Install PyBaMM
 --------------
@@ -88,7 +99,6 @@ library beforehand.
 
    .. code:: bash
 
-      brew install sundials
       pip install pybamm
 
 PyBaMM’s required dependencies (such as ``numpy``, ``casadi``, etc) will be
@@ -113,14 +123,35 @@ Users can install `scikits.odes <https://github.com/bmcage/odes>`__ to utilize i
    The ``scikits.odes`` solver is not supported on Python 3.12 yet. Please refer to https://github.com/bmcage/odes/issues/162.
    There is support for Python 3.8, 3.9, 3.10, and 3.11.
 
-.. tab:: GNU/Linux
+.. tab:: Debian/Ubuntu/Linux Mint
 
    In a terminal, run the following commands:
 
    .. code:: bash
 
-      apt-get install libopenblas-dev
-      pip install wget cmake
+      apt-get install libopenblas-dev \
+      pybamm_install_odes
+
+   system (under ``~/.local``), before installing ``scikits.odes``. (Alternatively, one can install SUNDIALS without this script and run ``pip install pybamm[odes]`` to install ``pybamm`` with ``scikits.odes``.)
+
+.. tab:: Fedora/CentOS
+
+   In a terminal, run the following commands:
+
+   .. code:: bash
+
+      dnf install openblas-devel \
+      pybamm_install_odes
+
+   system (under ``~/.local``), before installing ``scikits.odes``. (Alternatively, one can install SUNDIALS without this script and run ``pip install pybamm[odes]`` to install ``pybamm`` with ``scikits.odes``.)
+
+.. tab:: Arch Linux
+
+   In a terminal, run the following commands:
+
+   .. code:: bash
+
+      sudo pacman -S openblas \
       pybamm_install_odes
 
    system (under ``~/.local``), before installing ``scikits.odes``. (Alternatively, one can install SUNDIALS without this script and run ``pip install pybamm[odes]`` to install ``pybamm`` with ``scikits.odes``.)
@@ -132,7 +163,6 @@ Users can install `scikits.odes <https://github.com/bmcage/odes>`__ to utilize i
    .. code:: bash
 
       brew install openblas gcc gfortran
-      pip install wget cmake
       pybamm_install_odes
 
 The ``pybamm_install_odes`` command, installed with PyBaMM, automatically downloads and installs the SUNDIALS library on your
--- a/docs/source/user_guide/installation/gnu-linux-mac.rst
+++ b/docs/source/user_guide/installation/gnu-linux-mac.rst
@@ -14,26 +14,8 @@ To use PyBaMM, you must have Python 3.8, 3.9, 3.10, 3.11, or 3.12 installed.
 
    .. code:: bash
 
-      sudo apt-get update \
+      sudo apt-get update
       sudo apt-get install python3
-
-.. tab:: Fedora/CentOS
-
-   On Fedora or CentOS, you can use DNF or Yum. For example
-
-   .. code:: bash
-
-      sudo dnf upgrade \
-      sudo dnf install python3
-
-.. tab:: Arch Linux
-
-  On Arch-based systems you can use the ``pacman`` package manager. For example
-
-  .. code:: bash
-
-      sudo pacman -Syu \
-      sudo pacman -S python
 
 .. tab:: macOS
 
@@ -129,32 +111,10 @@ Users can install `scikits.odes <https://github.com/bmcage/odes>`__ to utilize i
 
    .. code:: bash
 
-      sudo apt-get install libopenblas-dev cmake
+      sudo apt-get install libopenblas-dev cmakeinstall
       pybamm_install_odes
 
    This will compile and install SUNDIALS for the system (under ``~/.local``), before installing ``scikits.odes``. (Alternatively, one can install SUNDIALS without this script and run ``pip install pybamm[odes]`` to install ``pybamm`` with ``scikits.odes``.)
-
-.. tab:: Fedora/CentOS
-
-   In a terminal, run the following commands:
-
-   .. code:: bash
-
-      sudo dnf install openblas-devel cmake
-      pybamm_install_odes
-
-   This will compile and install SUNDIALS for the system (under ``~/.local``), before installing ``scikits.odes``. (Alternatively, one can install SUNDIALS without this script and run ``pip install pybamm[odes]`` to install ``pybamm`` with ``scikits.odes``.)
-
-.. tab:: Arch Linux
-
-   In a terminal, run the following commands:
-
-   .. code:: bash
-
-      sudo pacman -S openblas cmake
-      pybamm_install_odes
-
-   system (under ``~/.local``), before installing ``scikits.odes``. (Alternatively, one can install SUNDIALS without this script and run ``pip install pybamm[odes]`` to install ``pybamm`` with ``scikits.odes``.)
 
 .. tab:: macOS
 
--- a/docs/source/user_guide/installation/gnu-linux-mac.rst
+++ b/docs/source/user_guide/installation/gnu-linux-mac.rst
@@ -14,8 +14,8 @@ To use PyBaMM, you must have Python 3.8, 3.9, 3.10, 3.11, or 3.12 installed.
 
    .. code:: bash
 
-      apt update \
-      apt install python3
+      sudo apt-get update \
+      sudo apt-get install python3
 
 .. tab:: Fedora/CentOS
 
@@ -23,17 +23,17 @@ To use PyBaMM, you must have Python 3.8, 3.9, 3.10, 3.11, or 3.12 installed.
 
    .. code:: bash
 
-      dnf upgrade \
-      dnf install python3
+      sudo dnf upgrade \
+      sudo dnf install python3
 
 .. tab:: Arch Linux
 
-  On Arch based systems you can use the pacman package manager. For example
+  On Arch-based systems you can use the ``pacman`` package manager. For example
 
   .. code:: bash
 
-      pacman -Syu \
-      pacman -S python
+      sudo pacman -Syu \
+      sudo pacman -S python
 
 .. tab:: macOS
 
@@ -132,7 +132,7 @@ Users can install `scikits.odes <https://github.com/bmcage/odes>`__ to utilize i
       apt-get install libopenblas-dev
       pybamm_install_odes
 
-   system (under ``~/.local``), before installing ``scikits.odes``. (Alternatively, one can install SUNDIALS without this script and run ``pip install pybamm[odes]`` to install ``pybamm`` with ``scikits.odes``.)
+   This will compile and install SUNDIALS for the system (under ``~/.local``), before installing ``scikits.odes``. (Alternatively, one can install SUNDIALS without this script and run ``pip install pybamm[odes]`` to install ``pybamm`` with ``scikits.odes``.)
 
 .. tab:: Fedora/CentOS
 
@@ -143,7 +143,7 @@ Users can install `scikits.odes <https://github.com/bmcage/odes>`__ to utilize i
       dnf install openblas-devel
       pybamm_install_odes
 
-   system (under ``~/.local``), before installing ``scikits.odes``. (Alternatively, one can install SUNDIALS without this script and run ``pip install pybamm[odes]`` to install ``pybamm`` with ``scikits.odes``.)
+   This will compile and install SUNDIALS for the system (under ``~/.local``), before installing ``scikits.odes``. (Alternatively, one can install SUNDIALS without this script and run ``pip install pybamm[odes]`` to install ``pybamm`` with ``scikits.odes``.)
 
 .. tab:: Arch Linux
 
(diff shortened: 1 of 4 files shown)


docs/source/user_guide/installation/gnu-linux-mac.rst
```diff
@@ -14,16 +14,26 @@ To use PyBaMM, you must have Python 3.8, 3.9, 3.10, 3.11, or 3.12 installed.
 
    .. code:: bash
 
-      sudo apt update
-      sudo apt install python3
+      apt update \
+      apt install python3
```
kratman, line 18: I would say that most times you need sudo to install on linux and mac

kratman, line 18: It has been a while since I checked, but I thought most distros dropped python2. Is it still necessary to specify python3?

kratman (commented):
How necessary is this? There are a lot of distributions out there and we don't want an exhaustive list. For instance Mint is just a reskin of Ubuntu.

Suggestions:
- Could we just keep instructions for the Debian based systems and note that they will have to be adjusted for other distributions?
- I would expect most linux users would know how to use their respective package managers. Could we just instead list the required packages?

docs/source/user_guide/installation/gnu-linux-mac.rst
agriyakhetarpal, line 18:
```suggestion
      sudo apt-get update \
      sudo apt-get install python3
```

this is what we've been using for our CI, I guess it makes sense? Also, removing `sudo` could be a bit misleading.

Also, most Debian-based distributions would not contain the latest and greatest Python distributions, could you add a sentence and a link to the Python official releases for those who wish to install using them?

docs/source/user_guide/installation/gnu-linux-mac.rst
```diff
@@ -14,16 +14,26 @@ To use PyBaMM, you must have Python 3.8, 3.9, 3.10, 3.11, or 3.12 installed.
...
 
 .. tab:: Fedora/CentOS
 
    On Fedora or CentOS, you can use DNF or Yum. For example
 
    .. code:: bash
 
-      sudo dnf install python3
+      dnf upgrade \
+      dnf install python3
```
agriyakhetarpal, line 27:
```suggestion
      sudo dnf upgrade \
      sudo dnf install python3
```

docs/source/user_guide/installation/gnu-linux-mac.rst
```diff
@@ -14,16 +14,26 @@ To use PyBaMM, you must have Python 3.8, 3.9, 3.10, 3.11, or 3.12 installed.
...
+
+.. tab:: Arch Linux
+
+  On Arch based systems you can use the pacman package manager. For example
+
+  .. code:: bash
+
+      pacman -Syu \
+      pacman -S python
```
agriyakhetarpal, line 36:
```suggestion
.. tab:: Arch Linux

  On Arch based systems you can use the ``pacman`` package manager. For example

  .. code:: bash

      sudo pacman -Syu \
      sudo pacman -S python
```

docs/source/user_guide/installation/gnu-linux-mac.rst
```diff
@@ -84,11 +95,10 @@ library beforehand.
 
 .. tab:: macOS
 
-   In a terminal, run the following commands:
+   In a terminal, run the following command:
```
agriyakhetarpal, line 98: Thanks – `brew install sundials` isn't needed anymore, yes!

docs/source/user_guide/installation/gnu-linux-mac.rst
```diff
@@ -113,14 +123,35 @@ Users can install `scikits.odes <https://github.com/bmcage/odes>`__ to utilize i
...
-.. tab:: GNU/Linux
+.. tab:: Debian/Ubuntu/Linux Mint
 
    In a terminal, run the following commands:
 
    .. code:: bash
 
       apt-get install libopenblas-dev
-      pip install wget cmake
+      pybamm_install_odes
```
agriyakhetarpal, line 133: We should install CMake here, whether through `apt-get` or any other method like `pipx` (though `apt-get` is preferred).

docs/source/user_guide/installation/gnu-linux-mac.rst
```diff
@@ -113,14 +123,35 @@ Users can install `scikits.odes <https://github.com/bmcage/odes>`__ to utilize i
...
+
+   system (under ``~/.local``), before installing ``scikits.odes``. (Alternatively, one can install SUNDIALS without this script and run ``pip install pybamm[odes]`` to install ``pybamm`` with ``scikits.odes``.)
```
agriyakhetarpal, line 135:
```suggestion
   This will compile and install SUNDIALS for the system (under ``~/.local``), before installing ``scikits.odes``. (Alternatively, one can install SUNDIALS without this script and run ``pip install pybamm[odes]`` to install ``pybamm`` with ``scikits.odes``.)
```

(I'm not sure why this was missing)

docs/source/user_guide/installation/gnu-linux-mac.rst
```diff
@@ -113,14 +123,35 @@ Users can install `scikits.odes <https://github.com/bmcage/odes>`__ to utilize i
...
+.. tab:: Fedora/CentOS
+
+   In a terminal, run the following commands:
+
+   .. code:: bash
+
+      dnf install openblas-devel
+      pybamm_install_odes
+
+   system (under ``~/.local``), before installing ``scikits.odes``. (Alternatively, one can install SUNDIALS without this script and run ``pip install pybamm[odes]`` to install ``pybamm`` with ``scikits.odes``.)
```
agriyakhetarpal, line 146:
```suggestion
   This will compile and install SUNDIALS for the system (under ``~/.local``), before installing ``scikits.odes``. (Alternatively, one can install SUNDIALS without this script and run ``pip install pybamm[odes]`` to install ``pybamm`` with ``scikits.odes``.)
```

(same as above)

docs/source/user_guide/installation/gnu-linux-mac.rst
```diff
@@ -113,14 +123,35 @@ Users can install `scikits.odes <https://github.com/bmcage/odes>`__ to utilize i
...
+
+.. tab:: Arch Linux
+
+   In a terminal, run the following commands:
+
+   .. code:: bash
+
+      pacman -S openblas
```
agriyakhetarpal, line 154:
```suggestion
      sudo pacman -S openblas
```

Also, will just `openblas` work or will we need a different OpenBLAS distribution like [`blas64-openblas`](https://archlinux.org/packages/extra/x86_64/blas64-openblas/)? Could you test this and run `pybamm_install_odes` to confirm, if you have a Linux Arch machine, that is?

docs/source/user_guide/installation/install-from-source.rst
```diff
@@ -34,25 +34,42 @@ To install PyBaMM, you will need:
 
 You can install the above with
 
-.. tab:: Ubuntu
+.. tab:: Ubuntu/Debian/Linux Mint
 
 	.. code:: bash
 
-		sudo apt install python3.X python3.X-dev libopenblas-dev gcc gfortran graphviz
+		apt install python3.X python3.X-dev libopenblas-dev gcc gfortran graphviz
```
agriyakhetarpal, line 41:
```suggestion
		sudo apt install python3.X python3.X-dev libopenblas-dev gcc gfortran graphviz
```

docs/source/user_guide/installation/install-from-source.rst
```diff
@@ -34,25 +34,42 @@ To install PyBaMM, you will need:
...
 	Where ``X`` is the version sub-number.
 
-	.. note::
+.. tab:: Fedora/CentOS
 
-		On Windows, you can install ``graphviz`` using the `Chocolatey <https://chocolatey.org/>`_ package manager, or
-		follow the instructions on the `graphviz website <https://graphviz.org/download/>`_.
+	.. code:: bash
+
+		dnf install python3.X-devel openblas-devel gcc gcc-fortran graphviz
```
agriyakhetarpal, line 49:
```suggestion
		sudo dnf install python3.X-devel openblas-devel gcc gcc-gfortran graphviz
```

docs/source/user_guide/installation/install-from-source.rst
```diff
@@ -34,25 +34,42 @@ To install PyBaMM, you will need:
...
+
+	Where ``X`` is the version sub-number.
+
+.. tab:: Arch Linux
+
+	.. code:: bash
+
+		pacman -S python openblas gcc gcc-fortran graphviz
```
agriyakhetarpal, line 57:
```suggestion
		sudo pacman -S python openblas gcc gcc-gfortran graphviz
```

I'm not sure if it is `gcc-fortran` or `gcc-gfortran` – please confirm

docs/source/user_guide/installation/install-from-source.rst
```diff
@@ -34,25 +34,42 @@ To install PyBaMM, you will need:
...
+
+	To install specific python version, refer `Arch Wiki <https://wiki.archlinux.org/title/python>`_.
```
agriyakhetarpal, line 59:
```suggestion
	To install a specific Python version, please refer to the `Arch Wiki <https://wiki.archlinux.org/title/python>`_.
```

docs/source/user_guide/installation/install-from-source.rst
```diff
@@ -332,6 +349,6 @@ configured SUNDIALS, it might say:
    -- LAPACK requires BLAS
 
 If this is the case, on a Debian or Ubuntu system you can install
-OpenBLAS using ``sudo apt-get install libopenblas-dev`` (or
+OpenBLAS using ``apt-get install libopenblas-dev`` (or
```
agriyakhetarpal, line 352:
```suggestion
OpenBLAS using ``sudo apt-get install libopenblas-dev`` (or
```

agriyakhetarpal (changes requested):

docs/source/user_guide/installation/gnu-linux-mac.rst
agriyakhetarpal: On older LTS versions, we do need that because both exist – but since they are all mostly old I added a suggestion about the official Python releases (we left a review at the same time)

agriyakhetarpal:
> I would expect most linux users would know how to use their respective package managers. Could we just instead list the required packages?

I agree with this. Linux Mint is a Debian derivative so it's not really needed

kratman:
> > I would expect most linux users would know how to use their respective package managers. Could we just instead list the required packages?
> 
> I agree with this. Linux Mint is a Debian derivative so it's not really needed (which Ubuntu is also a derivative of, as you mentioned)

Yeah Mint was just an example though. This should still be a single linux section, but with more generic instructions.  If we want to support all distros equally, then just a requirements list is probably better. We should only really need to tell them which libraries need to be installed.

santacodes:
> > > I would expect most linux users would know how to use their respective package managers. Could we just instead list the required packages?
> > 
> > 
> > I agree with this. Linux Mint is a Debian derivative so it's not really needed (which Ubuntu is also a derivative of, as you mentioned)
> 
> Yeah Mint was just an example though. This should still be a single linux section, but with more generic instructions. If we want to support all distros equally, then just a requirements list is probably better. We should only really need to tell them which libraries need to be installed.

Yeah the idea was just to cover all the common package managers, not specifically all kinds of distributions. But to have a single command with a list of libraries that need to be installed makes more sense.

agriyakhetarpal:
> Yeah the idea was just to cover all the common package managers, not specifically all kinds of distributions. But to have a single command with a list of libraries that need to be installed makes more sense.

Usually this is where most other packages I have seen provide an `environment.yml` file where you can set up an environment with `conda` or `mamba` and list both Pythonic and non-Pythonic dependencies. But that should be a separate issue

docs/source/user_guide/installation/gnu-linux-mac.rst
```diff
@@ -113,14 +123,35 @@ Users can install `scikits.odes <https://github.com/bmcage/odes>`__ to utilize i
...
+
+   This will compile and install SUNDIALS for the system (under ``~/.local``), before installing ``scikits.odes``. (Alternatively, one can install SUNDIALS without this script and run ``pip install pybamm[odes]`` to install ``pybamm`` with ``scikits.odes``.)
+
+.. tab:: Fedora/CentOS
+
+   In a terminal, run the following commands:
+
+   .. code:: bash
+
+      dnf install openblas-devel
```
arjxn-py, line 143:
```suggestion
      sudo dnf install openblas-devel cmake
```

docs/source/user_guide/installation/gnu-linux-mac.rst
```diff
@@ -132,7 +163,6 @@ Users can install `scikits.odes <https://github.com/bmcage/odes>`__ to utilize i
    .. code:: bash
 
       brew install openblas gcc gfortran
-      pip install wget cmake
```
arjxn-py, line 135: We still require `cmake` to run `pybamm_install_odes`, right?

arjxn-py (commented):
Good work @santacodes, some comments below:

docs/source/user_guide/installation/gnu-linux-mac.rst
```diff
@@ -113,14 +123,35 @@ Users can install `scikits.odes <https://github.com/bmcage/odes>`__ to utilize i
...
    There is support for Python 3.8, 3.9, 3.10, and 3.11.
 
-.. tab:: GNU/Linux
+.. tab:: Debian/Ubuntu
+
+   In a terminal, run the following commands:
+
+   .. code:: bash
+
+      apt-get install libopenblas-dev cmake
```
kratman, line 132:
```suggestion
      sudo apt-get install libopenblas-dev cmake
```

docs/source/user_guide/installation/gnu-linux-mac.rst
kratman: @santacodes Why was this resolved with no response?

docs/source/user_guide/installation/install-from-source.rst
kratman: I think the sudo in the suggestion should be added. This probably should not have been resolved yet

docs/source/user_guide/installation/gnu-linux-mac.rst
```diff
@@ -8,23 +8,33 @@ Prerequisites
...
+      sudo apt-get update \
+      sudo apt-get install python3
 
 .. tab:: Fedora/CentOS
 
    On Fedora or CentOS, you can use DNF or Yum. For example
 
    .. code:: bash
 
+      sudo dnf upgrade \
```
kratman, line 26: Why is this "\" added here? I think the slashes should probably be removed from the instructions. It is two different commands, so if anything it should be something like `sudo dnf upgrade && sudo ....`

Usually separate lines in bash code blocks are considered different commands. I would think that we want to keep them distinct commands

kratman (changes requested):
Minor stuff really.

In general though, I still think that it is better to just list the required dependencies and then have the linux specific steps for pybamm install from there. We are mostly ending up with repeated nearly identical instructions.

For example:
"PyBaMM requires the following dependencies:
`python3, cmake, gcc, gfortran, openblas-dev`

On Debian based systems these could be installed with
`sudo apt-get python3 cmake gcc gfortran openblas-dev`
These packages should all be installable from the linux package manager if you use a different distribution. Once the dependencies are installed, you can run the following commands to finish the installation:
`....`"

docs/source/user_guide/installation/install-from-source.rst
kratman:
```suggestion
		sudo pacman -S python openblas gcc gcc-fortran graphviz
```

santacodes:
> Minor stuff really.
> 
> In general though, I still think that it is better to just list the required dependencies and then have the linux specific steps for pybamm install from there. We are mostly ending up with repeated nearly identical instructions.
> 
> For example: "PyBaMM requires the following dependencies: `python3, cmake, gcc, gfortran, openblas-dev`
> 

You are right, we could maybe just make everything a bit more cleaner and simple. I think we could list out the package dependencies and remove the distribution specific commands.
>
> On Debian based systems these could be installed with `sudo apt-get python3 cmake gcc gfortran openblas-dev` These packages should all be installable from the linux package manager if you use a different distribution and some of them may already be installed. Once the dependencies are installed, you can run the following commands to finish the installation: `....`"

We could add a note under the command for this. I am open to hear any more suggestions that would make the install docs more readable and clean.

agriyakhetarpal (commented):
I did a brief reading of the installation docs, and I wanted to highlight two issues that are not a part of the changes in this PR, but can be fixed here itself:

1. The user installation for macOS still mentions that the command is `brew install sundials && pip install pybamm` – we can remove the SUNDIALS installation from here
2. In the source installation instructions for WSL, it further links to the user installation for Linux rather than the source installation for Linux.

Besides these, the changes look quite better to me now and the instructions seem to read well. ~We still need to resolve [#3914 (comment)](https://github.com/pybamm-team/PyBaMM/pull/3914/files#r1533773820) and add CMake before asking users to run `pybamm_install_odes`~ never mind, it's already there now 🙂

docs/source/user_guide/installation/gnu-linux-mac.rst
agriyakhetarpal: This looks to be resolved now, with https://github.com/pybamm-team/PyBaMM/pull/3914/commits/6522857446b745e01bebb358c1b1b6a0e83b7856.

kratman: @agriyakhetarpal Actually those are the MacOS instructions, so apt is not used for that

docs/source/user_guide/installation/gnu-linux-mac.rst
```diff
@@ -132,7 +123,6 @@ Users can install `scikits.odes <https://github.com/bmcage/odes>`__ to utilize i
    .. code:: bash
 
       brew install openblas gcc gfortran
```
kratman, line 125: Since cmake was not installed below we can install it with brew
```suggestion
      brew install openblas gcc gfortran cmake
```

docs/source/user_guide/installation/gnu-linux-mac.rst
kratman: I added it above instead

agriyakhetarpal: Oops, thanks for noticing!

docs/source/user_guide/installation/install-from-source.rst
```diff
@@ -34,25 +34,26 @@ To install PyBaMM, you will need:
 
 You can install the above with
 
-.. tab:: Ubuntu
+.. tab:: Ubuntu/Debian
 
 	.. code:: bash
 
 		sudo apt install python3.X python3.X-dev libopenblas-dev gcc gfortran graphviz
```
kratman, line 41:
```suggestion
		sudo apt install python3.X python3.X-dev libopenblas-dev gcc gfortran graphviz cmake
```

docs/source/user_guide/installation/install-from-source.rst
```diff
@@ -34,25 +34,26 @@ To install PyBaMM, you will need:
...
-	.. note::
-
-		On Windows, you can install ``graphviz`` using the `Chocolatey <https://chocolatey.org/>`_ package manager, or
-		follow the instructions on the `graphviz website <https://graphviz.org/download/>`_.
-
 .. tab:: MacOS
 
 	.. code:: bash
 
 		brew install python openblas gcc gfortran graphviz libomp
```
kratman, line 49:
```suggestion
		brew install python openblas gcc gfortran graphviz libomp cmake
```

kratman (commented):
@santacodes After my next few suggestions, this should be good

santacodes:
> I did a brief reading of the installation docs, and I wanted to highlight two issues that are not a part of the changes in this PR, but can be fixed here itself:
> 
> 1. The user installation for macOS still mentions that the command is `brew install sundials && pip install pybamm` – we can remove the SUNDIALS installation from here
> 2. In the source installation instructions for WSL, it further links to the user installation for Linux rather than the source installation for Linux.
> 
> Besides these, the changes look quite better to me now and the instructions seem to read well. ~We still need to resolve [#3914 (comment)](https://github.com/pybamm-team/PyBaMM/pull/3914/files#r1533773820) and add CMake before asking users to run `pybamm_install_odes`~ never mind, it's already there now 🙂

Hopefully 6b09d1e fixes this 😃

kratman (approved):
Looks good to me

agriyakhetarpal (approved):

PR #3919, opened 2024-03-22
Lithium plating upgrades

DrSOKane:
# Description

- Lithium plating model now works on composite electrodes
- Added plating parameters to Ecker2015, Ecker2015_graphite_halfcell and Chen2020_composite parameter sets
- Fixed bug where the dead lithium concentration variable had no scale

## Type of change

# Key checklist:

## Further checks:


tinosulzer (changes requested):
Submodel code looks good to me, but I am now against adding new parameters to existing parameter sets from the literature (e.g. adding lithium plating parameters to Ecker). In an ideal world, PyBaMM parameter sets would be based only on peer-reviewed papers, and only contain the parameters included in those papers. I know it's not the case now (lots of parameter sets contain extra parameters not in the main paper), but we should strive to move towards that, not away from it. 

If you want to add a parameter set based on Ecker with your own plating parameters, can you put it in a separate public repository, with entry points? I assume you're working on a paper related to this, once that is published we can add the full parameter set under "OKane2024" or whatever

DrSOKane:
> Submodel code looks good to me, but I am now against adding new parameters to existing parameter sets from the literature (e.g. adding lithium plating parameters to Ecker). In an ideal world, PyBaMM parameter sets would be based only on peer-reviewed papers, and only contain the parameters included in those papers. I know it's not the case now (lots of parameter sets contain extra parameters not in the main paper), but we should strive to move towards that, not away from it.
> 
> If you want to add a parameter set based on Ecker with your own plating parameters, can you put it in a separate public repository, with entry points? I assume you're working on a paper related to this, once that is published we can add the full parameter set under "OKane2024" or whatever

The O'Kane2020 paper did in fact use the Ecker2015 parameter set, but it was in COMSOL, not PyBaMM. The charging protocol used to validate the 2020 model is not supported by PyBaMM, for reasons we have previously discussed.

I agree that technically the Chen2020_composite model should not technically have any plating parameters. There is [another paper](https://www.sciencedirect.com/science/article/pii/S0378775323004196?via%3Dihub) with a composite lithium plating model, and I could add that paper instead.

tinosulzer:
> The charging protocol used to validate the 2020 model is not supported by PyBaMM

Is this the SOC-based cutoff? That should now be possible using custom events

If you have parameter sets that follow existing references then please add those. It just needs to be easy for future PyBaMM maintainers to point to a peer-reviewed reference to justify the parameter set, rather than trying to chase down old maintainers

DrSOKane:
I decided to just not have a parameter set for the composite plating. We plan to make one in the future, but this is a necessary first step.

valentinsulzer (approved):
Thanks, looks good to me now

PR #3920, opened 2024-03-22
Allow SEI layers choice

### LINKED ISSUE #2333: Add single-layer SEI

lorenzofavaro:
# Description
Currently, the default for SEI is one inner and one outer layer. 
Continuing the work of #3457, this change is to include an option to have a single layer SEI model, rather than just the 2-layer version.

Closes #2333 

## Type of change

# Key checklist:


kratman:
@lorenzofavaro This was being worked on by another person. Did you ask @GBlanka if they were still working on it? In general you don't want to take over the work of another developer unless they were not interested in continuing

lorenzofavaro:
> @lorenzofavaro This was being worked on by another person. Did you ask @GBlanka if they were still working on it? In general you don't want to take over the work of another developer unless they were not interested in continuing

You're right, I forgot to write it down but I asked her about it on another platform (on Github it is no longer active) and she confirmed that she was no longer interested in working on the PR #3457.

pybamm/models/submodels/interface/sei/base_sei.py
```diff
@@ -73,24 +75,26 @@ def _get_standard_thickness_variables(self, L_inner, L_outer):
             The variables which can be derived from the SEI thicknesses.
         """
         domain, Domain = self.domain_Domain
-        variables = {
-            f"{Domain} inner {self.reaction_name}thickness [m]": L_inner,
-            f"{Domain} outer {self.reaction_name}thickness [m]": L_outer,
-        }
+        L_sei = L_outer
+        variables = {f"{Domain} outer {self.reaction_name}thickness [m]": L_outer}
+
```
tinosulzer, line 80: I wouldn't add outer thickness in the single-layer case, only total thickness

tinosulzer: or just call it "SEI thickness" (not outer or total)

pybamm/models/submodels/interface/sei/sei_growth.py
```diff
@@ -133,8 +134,12 @@ def get_coupled_variables(self, variables):
...
         elif SEI_option == "electron-migration limited":
             # Scott Marquis thesis (eq. 5.94)
-            eta_inner = delta_phi - phase_param.U_inner
-            j_sei = (eta_inner < 0) * phase_param.kappa_inner * eta_inner / L_sei_inner
+            j_sei = 0
+            if self.double_layer_sei:
+                eta_inner = delta_phi - phase_param.U_inner
+                j_sei = (
+                    (eta_inner < 0) * phase_param.kappa_inner * eta_inner / L_sei_inner
+                )
```
tinosulzer, line 142: electron-migration limited should still work with the single sei layer, we would just assume that the single layer is the inner layer

tinosulzer (changes requested):
Looks good to me except a few comments. I'm wondering whether it would be cleaner to have entirely separate classes for double-layer or single-layer SEIs. There are a lot of if statements now in `sei_growth.py`

lorenzofavaro:
> I'm wondering whether it would be cleaner to have entirely separate classes for double-layer or single-layer SEIs. There are a lot of if statements now in sei_growth.py

@tinosulzer Could we think of adding two classes `SingleLayer` and `DoubleLayer` in this way?

UML-SEI_Layers

After that, update the function `set_six_submodel` in the `base_lithium_ion_model` module, so that it handles the correct initialization of the submodel, based on the `double SEI layer` option.

pybamm/models/submodels/interface/sei/base_sei.py
lorenzofavaro: Right, we could rename it to just `SEI thickness`. 
This renaming also apply to the other parameters in the single-layer case, right?
Ex: 
```python
f"X-averaged {domain} electrode outer {self.reaction_name}interfacial current density [A.m-2]"
```
would become:
```python
f"X-averaged {domain} electrode {self.reaction_name}interfacial current density [A.m-2]"
```

pybamm/models/submodels/interface/sei/sei_growth.py
lorenzofavaro: > we would just assume that the single layer is the **inner layer**

All right, but in the single-layer case we usually assume that the layer to be used is the outer layer right? Should we make an exception for this case and continue with the previous assumption in the remaining code?

tinosulzer: If we just make it a separate file it will be easier as there won't be a need for an exception

tinosulzer:
NoSEI and ConstantSEI should only have the single-SEI case (as it's pointless to model a double-layer SEI if it's not there or constant). TotalSEI is an aggregator anyway so that should deal with both single and double cases. That only leaves SEIGrowth and DoubleLayerSEIGrowth. In fact we may want to put the "double" option inside the SEI option e.g. "solvent-diffusion limited" and "solvent-diffusion limited (double-layer)"

Btw, thanks for your work on this and sorry for changing the specs along the way, it just became clearer when thinking about it more

tinosulzer:
The other (simpler) option is to just get rid of double layer SEI, which is just confusing and hard to parameterize. Thoughts @rtimms @brosaplanella @DrSOKane ?

rtimms:
I would happily remove the double layer stuff. If people really really want it they can implement as a custom submodel or basic model.

tinosulzer:
There's pros and cons to separating into different classes. I think a good middle ground is to have different submodel classes when the variables in `submodel.rhs` or `submodels.algebraic` change, and use if statements if we keep the same general structure but only vary the form of the equations. So in this case, keep it all in one submodel class

tinosulzer:
@lorenzofavaro the conclusion is let's just go ahead and get rid of the double SEI entirely. It should make the changes much cleaner

brosaplanella:
Hi @lorenzofavaro! Any updates on this?

brosaplanella:
Hi @lorenzofavaro! Any updates on this? Otherwise we can take this over as we need this feature for some work we are doing with @kawaMANMI.

PR #3932, opened 2024-03-26
Remove ODES solver

kratman:
# Description

Since the ODES solver was causing issues for upgrading to python 3.12 and might not offer anything outside of our other solvers, it might be able to be removed.

# Key checklist:

## Further checks:


kratman:
I created this PR very early since I want to test on CI. This is in no way ready for review

tinosulzer:
🎵  Should Old Acquaintance be forgot... 🎵

agriyakhetarpal:
For the macOS failures, removing some of the recent `pybamm-requires` caches should work (this comes from time to time whenever CMake puts out a new release, and I always forget about fixing it)

.github/workflows/test_on_push.yml
```diff
@@ -39,7 +39,6 @@ jobs:
         os: [ubuntu-latest, macos-12, windows-latest]
         python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
         # We check coverage on Ubuntu with Python 3.11, so we skip unit tests for it here
-        # TODO: check coverage with Python 3.12 when [odes] supports it
         exclude:
           - os: ubuntu-latest
             python-version: "3.11"
```
kratman, line 44: We should probably remove this. If we change the coverage test, and forget this line then we skip an architecture

.github/workflows/run_periodic_tests.yml
```diff
@@ -150,19 +150,13 @@ jobs:
       - name: Install build-time dependencies & run unit tests for M-series macOS runner
         shell: bash
         env:
-          # Point scikits.odes to the correct SUNDIALS installation
-          SUNDIALS_INST: $HOME/.local/lib
           # Homebrew environment variables
```
kratman, line 153:
```suggestion
```

.github/workflows/test_on_push.yml
agriyakhetarpal: Could you bump the coverage job to Python 3.12? It will get us some good [speedup](https://nedbatchelder.com/blog/202312/coveragepy_with_sysmonitoring.html) for it (the article is a bit dated so the feature is a bit more stable now)

kratman:
Since tests seem to be passing, I am going to open this up for review

Extra things for reviewers to check:
- [ ] Updated documentation
- [ ] Make sure there are no tests/scripts outside of CI that I missed
- [ ] If there are use-cases for ODEs that I missed, please mention them

kratman:
> Thanks, @kratman! This looks good!
> 
> [Dockerfile](https://github.com/pybamm-team/PyBaMM/blob/develop/scripts/Dockerfile) should also be updated to remove the extra `odes` dependency. Additionally, the CHANGELOG note should ideally be placed under the "Breaking changes" section and should specify why the solver was removed.

Yeah I was making the changelog now. Good catch on the dockerfile

.github/workflows/test_on_push.yml
```diff
@@ -38,11 +38,6 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-12, windows-latest]
         python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
```
agriyakhetarpal, line 40: On the below line, could you exclude Ubuntu with Python 3.12 again which you removed? It creates a redundant job otherwise

.github/workflows/test_on_push.yml
```diff
@@ -38,11 +38,6 @@ jobs:
...
-        # We check coverage on Ubuntu with Python 3.11, so we skip unit tests for it here
-        # TODO: check coverage with Python 3.12 when [odes] supports it
-        exclude:
-          - os: ubuntu-latest
-            python-version: "3.11"
```
agriyakhetarpal, line 45: i.e., exclude Ubuntu and Python 3.12 here (one cannot suggest changes on removed lines)

.github/workflows/test_on_push.yml
```diff
@@ -272,8 +260,7 @@ jobs:
       - name: Run integration tests for ${{ matrix.os }} with Python ${{ matrix.python-version }}
         run: python -m nox -s integration
 
-# Runs only on Ubuntu with Python 3.12. Skips IDAKLU module compilation
-# for speedups, which is already tested in other jobs.
+  # Skips IDAKLU module compilation for speedups, which is already tested in other jobs.
```
agriyakhetarpal, line 263: We should be able to bump this job back to Python 3.12 now

README.md
```diff
@@ -130,7 +130,6 @@ conda install -c conda-forge pybamm
 
 Following GNU/Linux and macOS solvers are optionally available:
```
agriyakhetarpal, line 131:
```suggestion
The following solvers are optionally available:
```

(JAX is available on Windows too)

docs/source/user_guide/installation/index.rst
```diff
@@ -47,7 +47,6 @@ Optional solvers
 
 Following GNU/Linux and macOS solvers are optionally available:
```
agriyakhetarpal, line 48:
```suggestion
The following solvers are optionally available:
```

(same as above)

kratman:
@pybamm-team/maintainers If you have checked any of the things in my checklist, please check them off

docs/source/user_guide/installation/install-from-docker.rst
```diff
@@ -26,12 +26,6 @@ Use the following command to pull the PyBaMM Docker image from Docker Hub:
 
             docker pull pybamm/pybamm:latest
 
-.. tab:: Scikits.odes solver
-
-      .. code:: bash
-
-            docker pull pybamm/pybamm:odes
-
```
agriyakhetarpal, line 34: `scripts/Dockerfile` needs to be updated too to remove the `ODES` arg

noxfile.py
```diff
@@ -63,19 +62,7 @@ def run_coverage(session):
     set_environment_variables(PYBAMM_ENV, session=session)
     session.install("coverage", silent=False)
     if sys.platform != "win32":
```
agriyakhetarpal, line 64: We can now remove the `sys.platform != "win32"` checks from everywhere in the file because all dependencies will now support all platforms

The only condition we will still need should be the Python <3.9 version check (where JAX is not available)

noxfile.py
```diff
@@ -178,35 +141,15 @@ def set_dev(session):
     # is fixed
     session.run(python, "-m", "pip", "install", "setuptools", external=True)
     if sys.platform == "linux":
```
agriyakhetarpal, line 143: This platform-specific check should not be needed anymore

agriyakhetarpal (changes requested):
I cannot find any more references besides these, thanks! Also, could you exclude the `IDAKLUJax` notebook from running in `nox -s examples` for Python <3.9 because JAX is unavailable? It has been failing our scheduled tests on 3.8 recently

agriyakhetarpal:
Oops, it looks like @Saransh-cpp reviewed at the same time as I did, lol

> Do any of our linux/mac dependencies in the instructions/docker get used for only ODES?

I checked the Dockerfile and it has just `SUNDIALS_INST` as an unneeded environment variable, all of the system-level dependencies in it (and the docs) are required

.github/workflows/test_on_push.yml
kratman: I removed it because it was easy to forget if we change the coverage test. I figured one redundant test was better than potentially missing a test. It also assumes that coverage tests will always be the same as this test

docs/source/user_guide/installation/install-from-docker.rst
kratman: Done already

.github/workflows/test_on_push.yml
kratman: See above

.github/workflows/test_on_push.yml
agriyakhetarpal: Let's still exclude it because it is one more unneeded job after all. You can move the coverage job to the same job if you want (but the reason why we kept it separate was because of being able to alter the `name:`

noxfile.py
kratman: JAX is fine on windows then? I left it because it was unrelated to ODES

agriyakhetarpal: Yes, JAX works on Windows

.github/workflows/test_on_push.yml
kratman: Done

noxfile.py
kratman: Since that change is related to enabling JAX on windows and not removing ODES, I think that should be a different PR. I can still do it though

noxfile.py
kratman: See above, I think that should be a different PR

noxfile.py
agriyakhetarpal: JAX is already enabled on Windows right now, so feel free to remove it here

.github/workflows/test_on_push.yml
kratman: I will make adjustments for this when I do the JAX changes. I think getting rid of the exclude is still a good thing since it prevents an error in the future. It is a good design principle to only have a single point of change, so if we change the coverage version then we should not have to remember to un-exclude a different test

agriyakhetarpal: Sure, resolving this comment for now

Saransh-cpp: I'll prefer keeping the `exclude` block too, will save us CI time and space. The coverage job has been a separate job for quite some time now (~3 years) and we made it a distinct job in order to save CI time. This should go away anyways with the GSoC `pytest` project, so we should keep it as it for now.

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@
 
 ## Breaking changes
 
+- The ODES solver was removed ([#3932](https://github.com/pybamm-team/PyBaMM/pull/3932))
```
Saransh-cpp, line 30: Could you please state the reason and alternatives here (for users who're using this solver).

Saransh-cpp (changes requested):

.github/workflows/test_on_push.yml
kratman: I still strongly disagree with this skipped test, but I put it back. For the very reason that when I put it back I had to go search to find out what the coverage was currently using. It is not a good approach

CHANGELOG.md
kratman: Done

.github/workflows/test_on_push.yml
agriyakhetarpal: We'll make a note about this as something to be fixed during the summer

agriyakhetarpal (approved):
So long, old friend

tinosulzer (approved):

agriyakhetarpal:
On another note, we shouldn't need to install two of the macOS system dependencies now with this, just `graphviz` and `libomp` should be required. We are fine with using Accelerate as a BLAS vendor and we needed `gcc` to get `gfortran` earlier for compiling `scikits.odes` (it wasn't being used for compiling IDAKLU anyway, we used the system Clang for that). I guess these changes would probably be better in a separate PR.

kratman:
> On another note, we shouldn't need to install two of the macOS system dependencies now with this, just `graphviz` and `libomp` should be required. We are fine with using Accelerate as a BLAS vendor and we needed `gcc` to get `gfortran` earlier for compiling `scikits.odes` (it wasn't being used for compiling IDAKLU anyway, we used the system Clang for that). I guess these changes would probably be better in a separate PR.

Can you add this to #3941 and I will look at it there

kratman:
Coverage will be fixed in #3942

PR #3942, opened 2024-03-29
Remove old JAX restrictions for Winows and test cleanup

### LINKED ISSUE #3941: Update windows logic for JAX in Nox and remove redundant Homebrew installations
Followup to comments from @agriyakhetarpal in #3932 

Changing settings for the JAX build was outside the scope of the other ticket, so this one is my reminder to go fix the issues.

> We can now remove the sys.platform != "win32" checks from everywhere in the file because all dependencies will now support all platforms
>
> The only condition we will still need should be the Python <3.9 version check (where JAX is not available)

kratman:
# Description

Support for JAX on windows was improved. This change removes the nox file restrictions on JAX. This builds on #3932 and keeps that one focused on just ODES.

Fixes #3941 

# Key checklist:

## Further checks:


kratman:
This will be rebased/forced pushed after #3932 is merged

PR #3945, opened 2024-03-30
Add CI jobs to upload nightly wheels to Anaconda

### LINKED ISSUE #3251: Infrastructure for nightly releases
### Description

I found some resources for our proposed nightly builds and hosting solutions plus options as to how we can configure them to work with our existing release infrastructure.

1. Anaconda provides artifact storage as a custom PyPI repository and offers 5 GB of storage on a free plan. The paid plans start from $9/month. Note: this method will require the creation of our own [Anaconda organisation](https://docs.anaconda.com/pro/organizations/) since the `scientific-python` repository is not open for packages outside its ecosystem. Some examples of this are
    a. NumPy and SciPy store their weekly wheels in this repository: https://anaconda.org/scipy-wheels-nightly/repo 
    b. Some packages under the `scientific-python` umbrella (`scikit-learn`, `pandas`, `matplotlib`, `xarray`, et cetera) publish at https://anaconda.org/scientific-python-nightly-wheels/repo
    c. Anaconda’s [own repository]( https://anaconda.org/anaconda/repo)
    d. AstroPy [used Azure earlier](https://docs.astropy.org/en/latest/changelog.html#id94), now they use [Anaconda](https://anaconda.org/astropy/repo).
2. [Cloudsmith.io](https://cloudsmith.io/user/login/) provides an expensive artifact repository priced at $89 per user per month.
3. [Google Cloud Artifact Registry](https://cloud.google.com/artifact-registry/docs/python/manage-packages) provides hosting and storage for Python packages. This is an enterprise solution so it might be private-only, I am not sure if it can be used for public, open-source projects.
4. A Homebrew formula using Ruby for PyBaMM (this one is a mostly undocumented and niche solution and does not have many resources to seek. Requires submission to and acceptance from `homebrew-core`. Here is a tutorial I found: https://til.simonwillison.net/homebrew/packaging-python-cli-for-homebrew). `cookiecutter` is an example of a Python package that has been packaged for Homebrew, since it is possible to install it with `brew install cookiecutter`. This solution would be available for Linux and macOS only, since Homebrew does not support Windows.
5. [Nexus Repository from Sonatype](https://www.sonatype.com/products/sonatype-nexus-oss) has support for many packaging formats and supports proxying and hosting for PyPI packages. This might also be a private-only solution.
6. [Artifactory](https://jfrog.com/artifactory/) also provides a super expensive artifact registry plan priced at $150 per user per month.
7. [PEP 503](https://peps.python.org/pep-0503/) describes how we can make our own PyPI repository for indexing. Anaconda is compliant with this specification. Some self-hosted solutions that are compatible with this comprise
    a. [devpi](https://devpi.net/docs/devpi/devpi/stable/)
    b. [pypiserver](https://pypi.org/project/pypiserver/), and a [tutorial by Linode](https://www.linode.com/docs/guides/how-to-create-a-private-python-package-repository/) to accompany it
    c. [Artipie](https://opensource.com/article/22/12/python-package-index-repository-artipie), an open-source PyPI package repository
    d. Pulp is another self-hosted package index that can be used as described in [this blog by Red Hat](https://developers.redhat.com/articles/2022/01/17/how-self-host-python-package-index-using-pulp)
8. I looked at GitHub Packages (https://github.com/features/packages) and found it to be pretty cool, but it doesn’t support Python packages yet and probably will not. There have been [discussions](https://github.com/orgs/community/discussions/8542) around this in their [roadmap](https://github.com/orgs/github/projects/4247/views/1?filterQuery=pypi) earlier, but they closed the feature request, sadly: https://github.com/github/roadmap/issues/94, possibly because Microsoft owns GitHub; so they have more precedence over what the folks at GitHub end up doing—plus I assume that they would like to keep Azure as the standard for a Python package registry, rather than offer a competing standard.
9. However, related to point 7, we could create a PEP-503 compliant `pybamm-team/pybamm-nightly` repository and push wheels to it daily. The size of a PyBaMM installation is around 160 MB, sourced from ([libraries.io](https://libraries.io/pypi/pybamm)). We could write a workflow to delete releases older than 30 days so that the size of the repository remains limited. `scientific-python` has a workflow to do this in their Anaconda index. Though there is no real-time vulnerability detection in this case like how other commercial solutions provide, we have more control over our release infrastructure anyway and we can mitigate bad actors. It is easy to install a package from GitHub as well since we can use `pip install git+` with the version from the release tag. There are many resources available, some of them I found are
    a. https://www.freecodecamp.org/news/how-to-use-github-as-a-pypi-server-1c3b0d07db2/
    b. https://medium.com/network-letters/using-github-as-a-private-python-package-index-server-798a6e1cfdef
    c. An example of a PyPI index hosted as a GitHub repository: https://github.com/astariul/github-hosted-pypi which works with GitHub Releases. The total size of cumulative GitHub Releases has [no limits](https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases#storage-and-bandwidth-quotas), but individual releases have to be below 2 GB each, which is many margins above our release size.
10. The [GitLab package registry](https://docs.gitlab.com/ee/user/packages/package_registry/supported_package_managers.html) supports PyPI packages. We could host a read-only mirror of PyBaMM on GitLab that gets updated with every commit to the `develop` and `main` branches. The better way would be to create a package registry there and write a GitLab CI pipeline to download artifacts from a GitHub Actions pipeline that uploads them, therefore starting a chain of CI/CD pipelines. Some resources (GitHub Actions in the marketplace, StackOverflow answers, and blogs) that might be useful in this case are listed below.
    a. https://github.com/marketplace/actions/trigger-gitlab-ci
    b. https://stackoverflow.com/questions/63308904/push-to-gitlab-with-access-token-using-github-actions
    c. https://github.com/marketplace/actions/trigger-gitlab-ci-through-webhooks
    d. https://github.com/marketplace/actions/trigger-gitlab-ci-job
    e. https://github.com/marketplace/actions/gitlab-pipeline-trigger
    f. https://dev.to/edersonbrilhante/gitlab-runners-as-a-service-with-github-action-149n
    g. https://www.anapaulagomes.me/2021/04/publishing-your-python-package-in-your-gitlab-package-registry/

We might need to ensure that the guide to downloading and using nightly releases is documented properly and warn unsuspecting users from using them. An edge case to take care of is that `pip` does not fall back to using PyPI to download via the `--extra-index` flag if a package is not found on the custom index which is a common modus operandi for dependency hijacking attacks. Source: https://discuss.python.org/t/advice-to-avoid-extra-index-url-to-install-private-packages-from-gitlab-ci/18242/11

### Motivation

_No response_

### Possible Implementation

_No response_

### Additional context

_No response_

agriyakhetarpal:
# Description

This PR configures the `publish_pypi.yml` workflow (maybe we should rename it to `publish_wheels.yml` now?) to add a job that uploads wheels to Anaconda. Currently, wheels will be uploaded on the same cadence as the workflow, so they are more like "fortnightlies" rather than nightlies.

We 

Fixes #3251 

## Type of change

# Key checklist:

## Further checks:


.github/workflows/publish_pypi.yml
```diff
@@ -212,6 +213,35 @@ jobs:
           path: ./dist/*.tar.gz
           if-no-files-found: error
 
+  publish_anaconda:
+    # This job is only of value to PyBaMM and would always be skipped in forks
+    # It is meant to run on a schedule, when manually triggered, and does not
+    # run on releases.
```
agriyakhetarpal, line 219: Note: this job is configured differently and is isolated from the PyPI upload job for the reason that the latter uses the `pypi` repository environment, which must not be accessed by other jobs.

.github/workflows/publish_pypi.yml
```diff
@@ -212,6 +213,35 @@ jobs:
...
+    needs: [build_macos_and_linux_wheels, build_macos_arm64_wheels, build_windows_wheels, build_sdist]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Download all artifacts
+        uses: actions/download-artifact@v4
+
+      - name: Move all package files to files/
+        run: |
+          mkdir files
+          mv windows_wheels/* linux_wheels/* macos_amd64_wheels/* macos_arm64_wheels/* files/
```
agriyakhetarpal, line 234: Note: the source distribution is not uploaded to Anaconda because it does not make sense to do so. The nightly uploads are supposed to be binary-only. It is therefore not copied into the `files/` staging area for upload.

.github/workflows/publish_pypi.yml
```diff
@@ -249,7 +279,7 @@ jobs:
           repository-url: https://test.pypi.org/legacy/
 
   open_failure_issue:
-    needs: [build_windows_wheels, build_macos_and_linux_wheels, build_sdist]
+    needs: [build_windows_wheels, build_macos_and_linux_wheels, build_macos_arm64_wheels, build_sdist]
```
agriyakhetarpal, line 282: This could have made our wheel builds fail silently, just caught it

.github/workflows/publish_pypi.yml
```diff
@@ -212,6 +213,34 @@ jobs:
...
+      - name: Download all artifacts
+        uses: actions/download-artifact@v4
+
+      - name: Move all package files to files/
+        run: |
+          mkdir files
+          mv windows_wheels/* linux_wheels/* macos_amd64_wheels/* macos_arm64_wheels/* files/
+
+      - name: Push to Anaconda PyPI index
+        uses: scientific-python/upload-nightly-action@b67d7fcc0396e1128a474d1ab2b48aa94680f9fc  # v0.5.0
```
agriyakhetarpal, line 237: It's recommended to pin this action to a commit hash. I think we should do this for all of our other actions in other jobs too.

agriyakhetarpal:
This is still a draft because I need to add instructions for how to let people use these nightly wheels in the installation guide. I think I can prepare that locally, but it would make more sense to push those changes when we will have the repository secrets and the custom index configured for uploads (and subsequently downloads).

agriyakhetarpal:
>  Do we have upload limits on anaconda? Nightly builds would make a lot of releases

Yes, there is a 5 GB upload limit, but the action handles deleting releases older than N versions or M days automatically if one pushes multiple versions (N and M can be configured). I forgot to mention here that since we are not configuring any SCM-based versioning for the nightlies, they will be overwritten on each upload and we will not hit any limits. 

This is what I had sent to @tinosulzer the other day about the requirements here:

1. A nightly channel to push wheels to, to be filled in `anaconda_nightly_upload_organization`
2. An API key generated to push to this channel and then added as a repository secret
3. A manual upload for a single wheel to this channel for the first time (any platform/architecture shall work, we can delete it later).

arjxn-py:
> I forgot to mention here that since we are not configuring any SCM-based versioning for the nightlies, they will be overwritten on each upload and we will not hit any limits.

Do we plan to configure any SCM-based versioning?

agriyakhetarpal:
> Do we plan to configure any SCM-based versioning?

Not for now until we migrate to a new build system. If we were to set up `setuptools-scm` and then we start removing `setuptools` soon, it's going to cause issues with file inclusion and exclusion later, because `scikit-build-core` uses different ways to handle that and we would need to add something like `versioneer` and go through it all over again.

What I will suggest for this right now is to use [local version identifiers](https://packaging.python.org/en/latest/specifications/version-specifiers/#local-version-identifiers) as noted in PEP 440 to distinguish between PyPI releases and these nightlies, and therefore rename the wheel to add `+dev` after the version tag and before the platform and ABI tags, similar to [what JAX does](https://storage.googleapis.com/jax-releases/jax_cuda_releases.html) for its GPU wheels. I'll push another commit to make this change.

agriyakhetarpal:
Bumping the version for built wheels would require a bit of restructuring in the workflow to update the version before building the nightlies and that means this should probably be done when we have some SCM-based versioning. I think we can go ahead with this workflow right now and (I merged the latest changes from `develop` just now).

The MO to test the nightlies is described in [SPEC-0004](https://scientific-python.org/specs/spec-0004/#test-with-nightly-wheels), which means that PyBOP as a dependent will be implementing a weekly workflow which downloads the PyBOP and PyBaMM nightly wheels as such, something like:

```yaml
      - name: Test nightly builds
        run: |
          pip install pybop pybamm --pre --extra-index-url https://pypi.anaconda.org/pybamm-and-friends/simple
          pytest --pyargs pybop  # or any other test invocation command
```

(cc: @BradyPlanden), which can then open an issue for PyBOP about incoming failures due to changes in PyBaMM. It is to be noted that the use of `--extra-index-url` here rather than `--index-url` means that we open up indices such as PyPI to search in for the case of downloading PyBaMM's dependencies (NumPy, SciPy, CasADi, etc.) because their wheels would not exist in our own nightly index. If it is needed (which IMO it isn't), we can specify multiple index URLs to test in CI and add the `https://pypi.anaconda.org/scientific-python-nightly-wheels/simple` index. 

Users should also look to use `--extra-index-url` though, because PyBaMM doesn't have a lot of dependents right now (four, based on GitHub's insights). I probably will have to update the installation guide about this so that they are not misled with not being able to download `scipy`/`numpy`/`casadi`/`xarray` from our index, but still warn them enough that they should use these command-line flags carefully and at their own discretion.

.github/workflows/publish_pypi.yml
agriyakhetarpal: I resolved this in #4015 and merged `develop` to get this here.

.github/workflows/build_wheels_and_publish.yml
```diff
@@ -203,6 +204,33 @@ jobs:
...
+        run: |
+          mkdir files
+          mv windows_wheels/* linux_wheels/* macos_amd64_wheels/* macos_arm64_wheels/* files/
+
+      - name: Push to Anaconda PyPI index
+        uses: scientific-python/upload-nightly-action@b67d7fcc0396e1128a474d1ab2b48aa94680f9fc  # v0.5.0
+        with:
+          artifacts_path: files/
+          anaconda_nightly_upload_organization: # TODO: create
+          anaconda_nightly_upload_token: ${{ secrets.ANACONDA_ORG_UPLOAD_TOKEN }} # TODO: configure
```
kratman, line 232:
```suggestion
          anaconda_nightly_upload_token: ${{ secrets.ANACONDA_ORG_UPLOAD_TOKEN }}
```

arjxn-py: Best

PR #3947, opened 2024-03-30
Added pybamm user id to 1000 and set its group to root

santacodes:
# Description

Changed the pybamm userID to 1000 and groupID to 0 (root).

Related to #3312 
docker

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/docs/source/user_guide/installation/install-from-docker.rst
+++ b/docs/source/user_guide/installation/install-from-docker.rst
@@ -85,6 +85,9 @@ Once you have pulled the Docker image, you can run a Docker container with the P
 
 3. You can execute PyBaMM-related commands, run tests develop & contribute from the container.
 
+.. note::
+    The default user for the container is ``pybamm`` with ``pybamm`` as password. The user belongs to ``sudoers`` and ``root`` group so the sudo command can be issued to install additional packages to the container.  After a clean install, ``sudo apt-get update`` should be executed to update the source list. Additional packages can be installed using ``sudo apt-get install [package_name]``.
+
 Exiting the Docker container
 ----------------------------
 
--- a/docs/source/user_guide/installation/install-from-docker.rst
+++ b/docs/source/user_guide/installation/install-from-docker.rst
@@ -86,7 +86,11 @@ Once you have pulled the Docker image, you can run a Docker container with the P
 3. You can execute PyBaMM-related commands, run tests develop & contribute from the container.
 
 .. note::
-    The default user for the container is ``pybamm`` with ``pybamm`` as password. The user belongs to ``sudoers`` and ``root`` group so the sudo command can be issued to install additional packages to the container.  After a clean install, ``sudo apt-get update`` should be executed to update the source list. Additional packages can be installed using ``sudo apt-get install [package_name]``.
+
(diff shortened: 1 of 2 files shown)


santacodes:
By default the default user `pybamm` inside the docker container doesn't have root permissions which would restrict the user from using `fake root` environment or `sudo` operations to install additional packages and also to use commands like `chmod` which might be useful to run executables. To solve this `pybamm` user is added to the root group (0).

agriyakhetarpal:
Thanks for this @santacodes – could you spin up a container and verify your changes if you haven't done so already? The expected behaviour is that the `pybamm` user should be able to install packages as you mentioned, so installing something via `apt-get` should work via the entry point for the container – if that does, we should be good to go. Another thing you could test out is whether you can copy things to and from the location pointed to by `LD_LIBRARY_PATH`.

santacodes:
@agriyakhetarpal Sure I gave it a try and these are the points to be noted -

- The docker container doesn't have the `sudo` command by default due to security issues as typically they are used for microservices. So to overcome this `apt-get install sudo` is added to dockerfile and `sudo` is installed at build-time. This shouldn't be a problem as this container is aimed for development purposes and not for deploying services.
- I have added the user `pybamm` to the sudoers group which would enable it to use the `sudo` command.
- The default password for the user `pybamm` is `pybamm` which is assigned at build-time. We can change it to something else and mention it in docs or something.
- By entry point, I am guessing bash, as for a new container the `apt-get` packages are not synchronized so to ensure that `sudo apt-get update` needs to be issued before using any `apt-get install` command.
aptdocker
- Also for the `LD_LIBRARY_PATH` which is directed at `/home/pybamm/.local/lib` directory the `cp` command for copying files works. 
copyperms

arjxn-py (commented):
Thanks for this @santacodes, can you also add some related useful documentation for this on the `Install from Source (Docker)` page?

santacodes:
> Thanks for this @santacodes, can you also add some related useful documentation for this on the `Install from Source (Docker)` page?

Sure I will add them 😃

docs/source/user_guide/installation/install-from-docker.rst
```diff
@@ -85,6 +85,9 @@ Once you have pulled the Docker image, you can run a Docker container with the P
 
 3. You can execute PyBaMM-related commands, run tests develop & contribute from the container.
 
+.. note::
+    The default user for the container is ``pybamm`` with ``pybamm`` as password. The user belongs to ``sudoers`` and ``root`` group so the sudo command can be issued to install additional packages to the container.  After a clean install, ``sudo apt-get update`` should be executed to update the source list. Additional packages can be installed using ``sudo apt-get install [package_name]``.
```
agriyakhetarpal, line 89:
```suggestion
.. note::

    The default user for the container is ``pybamm`` with ``pybamm`` as password. The user belongs to
    ``sudoers`` and ``root`` group, so the sudo command can be issued to install additional packages to
    the container.  After a clean install, ``sudo apt-get update`` should be executed to update the source
    list. Additional packages can be installed using ``sudo apt-get install [package_name]``.
```

Just a formatting change, please split text into multiple lines if possible and indent the text if you are using an admonition

arjxn-py (approved):
Looks good to me, thanks @santacodes

PR #3961, opened 2024-04-03
Remove support for python 3.8

kratman:
# Description

Removes support for python 3.8

Fixes #3390

## Type of change

This is a breaking change since it reduces support for a python version. On the plus side it reduces the number of CI jobs.

# Key checklist:

- [x] No style issues: `$ pre-commit run` (or `$ nox -s pre-commit`) (see [CONTRIBUTING.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CONTRIBUTING.md#installing-and-using-pre-commit) for how to set this up to run automatically when committing locally, in just two lines of code)
- [x] All tests pass: `$ python run-tests.py --all` (or `$ nox -s tests`)
- [x] The documentation builds: `$ python run-tests.py --doctest` (or `$ nox -s doctests`)

You can run integration tests, unit tests, and doctests together at once, using `$ python run-tests.py --quick` (or `$ nox -s quick`).

## Further checks:

- [x] Code is commented, particularly in hard-to-understand areas
- [x] Tests added that prove fix is effective or that feature works

Issue #3948, opened 2024-03-31
Hangs encountered in `IDAKLUJax` unit tests (`test_jacrev_vmap`)


agriyakhetarpal:
### PyBaMM Version

develop

### Python Version

3.11.8

### Describe the bug

The `test_jacrev_vmap` test case in the `TestIDAKLUJax` class (present in `tests/unit/test_solvers/test_idaklu_jax.py`) hangs quite a lot during local development. It is one of the slowest tests to pass, to the point that coverage logging almost gets stuck indefinitely on 99% and that this test, in particular, takes time in several orders of magnitude more to complete when compared to the rest of the tests.

This is most likely coming from the recent migration to using `pytest` for running the unit tests (#3857), which also brought support for `pytest-xdist` for parallel execution of unit tests, where JAX-related unit tests take up a lot of time in CI in parallel mode.

Here's an SVG from a profiling sample with the `pytest-profiling` plugin from @prady0t earlier in the `#infrastructure` channel on Slack:

<details>
<summary>Expand to view</summary>

![combined](https://github.com/pybamm-team/PyBaMM/assets/74401230/b12b6fb7-232b-4052-a1c2-34bfd4e6444d)

</details>

which reveals that something is up with the JAX-related tests.

### Steps to Reproduce

There isn't a better reproducer at this time, but to reproduce one can run `nox -s coverage` or its `pytest --cov` equivalent in the root directory – it is a bit slower than `nox -s unit`, but both of them seem to have the same issue.

agriyakhetarpal:
Temporary solution: wrap all classes and the methods with `@pytest.mark.xdist_group(name="serial execution")` to run them inside the same worker in serial mode.

This resolves the `test_jacrev_vmap` execution and it runs as normal – but `test_jacrev_vector_getvars`, and `test_solver_` + `test_solver_sensitivities` from the JAX BDF solver are three other culprits which still aren't happy – they still take a lot longer than any other test. I assume we can extract more speedups with the newly enabled parallel testing, however, these test cases do not seem to budge and are causing bottlenecks. Maybe @jsbrittain would have some suggestions here?

agriyakhetarpal:
To bring some rudimentary sense of the issue at hand, here's what I can see locally on an M-series macOS machine:

Running the entire coverage suite with the JAX tests included:
```
1596 passed, 2 skipped in 663.93s (0:11:03)
```

[...]

PR #3963, opened 2024-04-03
Remove time check from Jax tests to improve stability

### LINKED ISSUE #3925: Flaky JAX test cases
While running unit-tests, some tests appaers to be flaky. See https://github.com/pybamm-team/PyBaMM/pull/3857#issuecomment-2015300918

#3857 Is a prerequisite to this. Once it gets merged, we can discuss more such tests here and come up with a solution for them.

kratman:
# Description

Jax tests checked runtime, which is inconsistent in the CI suite. This should keep the same coverage, but stop checking the runtime. Note: the changelog is also updated since that was missed in #3961 

Fixes #3925

## Type of change

Only test code.

# Key checklist:

## Further checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@
 
 ## Breaking changes
 
+- Removed support for Python 3.8 ([#3961](https://github.com/pybamm-team/PyBaMM/pull/3961))
 - Renamed "ocp_soc_0_dimensional" to "ocp_soc_0" and "ocp_soc_100_dimensional" to "ocp_soc_100" ([#3942](https://github.com/pybamm-team/PyBaMM/pull/3942))
 - The ODES solver was removed due to compatibility issues. Users should use IDAKLU, Casadi, or JAX instead. ([#3932](https://github.com/pybamm-team/PyBaMM/pull/3932))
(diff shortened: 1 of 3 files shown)


agriyakhetarpal (approved):

kratman:
I don't see anywhere that the operator usage should have changed, so I think that coverage is outdated

tinosulzer:
Are these tests actually failing, or is this a "we shouldn't test times in CI"? They do test real functionality, if they are failing that indicates there may be a bug

cc @martinjrobins

kratman:
Both
1. Time checks are not great because multiple jobs run on the same server.
2. They do end up failing randomly. Re-triggering makes them pass. This got worse since we switched to pytest

agriyakhetarpal:
I think the failures are macOS-specific, because these tests are failing in CI but only for macOS and I have never seen them fail on Ubuntu or Windows. These take a really long time to complete on my local macOS machine too (#3948) and fail quite erratically. It makes me wonder if something is amiss upstream or a memory leak of some kind

kratman:
Yeah I intend to keep looking into the Jax stuff, so I will try to find out if there is more going on

tinosulzer (approved):
Got it. Removing makes sense in that case

PR #3968, opened 2024-04-07
Fix optional dependencies error detection and remove `pybtex` hidden dependency

### LINKED ISSUE #3940: PyBaMM import error - Cannot run anything (import broken due to `pybtex`)
### Discussed in https://github.com/pybamm-team/PyBaMM/discussions/3928

<sup>Originally posted by **abhishek-appana** March 25, 2024</sup>
Hello,

I am running Python 3.11.8 and installed PyBaMM 24.1 using pip on a Windows system. Earlier I was using Python 3.10 and everything was fine. But I am getting the below error after upgrading Python.

When I run `import pybamm`, it gives me the error:

ModuleNotFoundError: Optional dependency pybtex.database is not available. See https://docs.pybamm.org/en/latest/source/user_guide/installation/index.html#optional-dependencies for more details.

I tried installing pybamm using `pip install pybamm[all]`. I was still getting this error. How do I resolve this?</div>

<hr>

I think that this is a valid bug and it is still failing, and `import pybamm` should work without installing `pybamm[all]`. It can be noticed in the logs here: https://github.com/pybamm-team/PyBaMM/actions/runs/8461852686/job/23182296194, so I've converted this discussion to an issue instead (cc: @Saransh-cpp @arjxn-py). 

Perhaps this isn't being properly checked through #3892, @lorenzofavaro?

lorenzofavaro:
# Description
The goal of this PR is to solve two problems.
- Remove optional dependencies present in the base version of PyBaMM, so that `pybamm` can be imported without optional dependencies installed
- Fix error detection: when an optional dependency is present in non-optional modules, we need to detect these errors and report them.

Fixes #3940 

# Key checklist:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -126,26 +126,18 @@ This allows people to (1) use PyBaMM without importing optional dependencies by 
 
 **Writing Tests for Optional Dependencies**
 
-Whenever a new optional dependency is added for optional functionality, it is recommended to write a corresponding unit test in `test_util.py`. This ensures that an error is raised upon the absence of said dependency. Here's an example:
+Below, we list the currently available test functions to provide an overview. If you find it useful to add new test cases please do so within `tests/unit/test_util.py`.
 
-```python
-from tests import TestCase
-import pybamm
+Currently, there are three functions to test what concerns optional dependencies:
+- `test_import_optional_dependency`
+- `test_pybamm_import`
+- `test_optional_dependencies`
 
+The `test_import_optional_dependency` function extracts the optional dependencies installed in the setup environment, makes them unimportable (by setting them to `None` among the `sys.modules`), and tests that the `pybamm.util.import_optional_dependency` function throws a `ModuleNotFoundError` exception when their import is attempted.
 
-class TestUtil(TestCase):
-    def test_optional_dependency(self):
-        # Test that an error is raised when pybtex is not available
-        with self.assertRaisesRegex(
-            ModuleNotFoundError, "Optional dependency pybtex is not available"
-        ):
-            sys.modules["pybtex"] = None
-            pybamm.function_using_pybtex(x, y, z)
+The `test_pybamm_import` function extracts the optional dependencies installed in the setup environment and makes them unimportable (by setting them to `None` among the `sys.modules`), unloads `pybamm` and its sub-modules, and finally tests that `pybamm` can be imported successfully. In fact, it is essential that the `pybamm` package is importable with only the mandatory dependencies.
 
-        # Test that the function works when pybtex is available
-        sys.modules["pybtex"] = pybamm.util.import_optional_dependency("pybtex")
-        pybamm.function_using_pybtex(x, y, z)
-```
+The `test_optional_dependencies` function extracts `pybamm` mandatory distribution packages and verifies that they are not present in the optional distribution packages list in `pyproject.toml`. This test is crucial for ensuring the consistency of the released package information and potential updates to dependencies during development.
 
 ## Testing
 
--- a/pybamm/citations.py
+++ b/pybamm/citations.py
@@ -179,7 +179,7 @@ class Citations:
         except ModuleNotFoundError:
             pybamm.logger.warning(
                 f"Could not parse citation for '{key}' because the 'pybtex' library is not installed. "
-                "Install 'pybtex' to enable citation parsing."
+                "Install 'pybamm[cite]' to enable citation parsing."
             )
 
     def _tag_citations(self):
--- a/pybamm/citations.py
+++ b/pybamm/citations.py
@@ -76,7 +76,7 @@ class Citations:
         except ModuleNotFoundError:
             pybamm.logger.warning(
                 "Citations could not be read because the 'pybtex' library is not installed. "
-                "Install 'pybtex' to enable citation reading."
+                "Install 'pybamm[cite]' to enable citation reading."
             )
 
     def _add_citation(self, key, entry):
--- a/pybamm/citations.py
+++ b/pybamm/citations.py
@@ -100,7 +100,7 @@ class Citations:
         except ModuleNotFoundError:
             pybamm.logger.warning(
                 f"Could not add citation for '{key}' because the 'pybtex' library is not installed. "
-                "Install 'pybtex' to enable adding citations."
+                "Install 'pybamm[cite]' to enable adding citations."
             )
 
     def _add_citation_tag(self, key, entry):
--- a/pybamm/citations.py
+++ b/pybamm/citations.py
@@ -73,7 +73,7 @@ class Citations:
             bib_data = parse_file(citations_file, bib_format="bibtex")
             for key, entry in bib_data.entries.items():
                 self._add_citation(key, entry)
-        except ModuleNotFoundError: # pragma: no cover
+        except ModuleNotFoundError:  # pragma: no cover
             pybamm.logger.warning(
                 "Citations could not be read because the 'pybtex' library is not installed. "
                 "Install 'pybamm[cite]' to enable citation reading."
@@ -97,7 +97,7 @@ class Citations:
 
             # Add to database
             self._all_citations[key] = new_citation
-        except ModuleNotFoundError: # pragma: no cover
+        except ModuleNotFoundError:  # pragma: no cover
             pybamm.logger.warning(
                 f"Could not add citation for '{key}' because the 'pybtex' library is not installed. "
                 "Install 'pybamm[cite]' to enable adding citations."
@@ -176,7 +176,7 @@ class Citations:
                 raise KeyError(
                     f"Not a bibtex citation or known citation: {key}"
                 ) from error
-        except ModuleNotFoundError: # pragma: no cover
+        except ModuleNotFoundError:  # pragma: no cover
             pybamm.logger.warning(
                 f"Could not parse citation for '{key}' because the 'pybtex' library is not installed. "
                 "Install 'pybamm[cite]' to enable citation parsing."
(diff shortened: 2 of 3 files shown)


agriyakhetarpal:
Thanks, @lorenzofavaro – with the failing tests we can confirm that we are now testing the PyBaMM import correctly and properly :) I guess the next step is the one you highlighted about `pybtex`'s import being required for PyBaMM's import, which should get revised so that it is only done when someone does `pybamm.print_citations()` and not when one does `import pybamm`. Maybe doing so is as easy as either moving `self.read_citations()` to `Citations.print()` plus moving `self._reset()` out of `__init__` (since the docstring mentions that it is used for testing purposes), or using try-except blocks again (which would be less desirable).

lorenzofavaro:
> Maybe doing so is as easy as either moving self.read_citations() to Citations.print() plus moving self._reset() out of __init__ (since the docstring mentions that it is used for testing purposes, or moving the citations.register() bits), or using try-except blocks again (which would be less desirable).

Thanks @agriyakhetarpal, I tried this way but the problem was being able to handle a single instance of the object. The solution I found was using the singleton pattern.

I also took the opportunity to add a handy api to register new citations with `pybamm.register_citation(...)`. Let me know if it will work out for you.

pybamm/citations.py
```diff
@@ -16,15 +16,28 @@ class Citations:
...
+        def __get__(self, cls, owner):
+            return classmethod(self.fget).__get__(None, owner)()
+
+    @classproperty
+    def citations(cls):
+        if cls._instance is None:
+            cls._instance = cls.__new__(cls)
+            cls._instance.__init__()
+        return cls._instance
+
```
agriyakhetarpal, line 40: This is a really neat and convenient decorator, but

pybamm/citations.py
```diff
@@ -245,8 +247,28 @@ def print(self, filename=None, output_format="text", verbose=False):
...
+    Parameters
+    ----------
+    key : str
+        - The citation key for an entry in `pybamm/CITATIONS.bib` or
+        - A BibTeX formatted citation
+    """
+    citations = Citations.citations
+    citations._register(key)
+
```
agriyakhetarpal, line 267: I'm afraid that this leans towards being a breaking change for users, so we'll probably have to revert this or restructure the function to match the current API. We can't issue a breaking change without a warning message and a deprecation cycle across two/three releases being involved. Is it possible to stick with `pybamm.citations.register("<...>")` for now?

agriyakhetarpal (changes requested):
Thanks, @lorenzofavaro, this is great – just a single comment about the new API for registering citations which probably needs a revisit :)

pybamm/citations.py
agriyakhetarpal: I see. Yes, `pybamm.Citations.citations.register(key)` is also a bit messy, and the try-except block was also masking the import error, which we probably don't want (because it was masking any other `Exception`s too). I think a better way would be to rethink the citations workflow from start to end. The current one is using `self.read_citations()` during `__init__`, and we are importing the class itself when we run `import pybamm` – here we could make the `citations.register` functionality optional across the codebase, which can work, but then we want to encourage people to cite their references properly and not make it optional to do so. One way out is to make `pybtex` a mandatory dependency and remove the `[cite]` extra – which should be the last way out.

One option that seems reasonable at the time of writing is to put import guards across all of the `pybtex`-related functionality. Wherever we are using `import_optional_dependency("pybtex")`, if `pybtex` is not present, we are currently not implementing any try-except blocks to ensure that the `pybtex`-specific functions are not called. Maybe doing so and adding a helpful error message can fix it? i.e., `read_citations`, `_add_citation`, and `_parse_citation`, and `print` should not run if `pybtex` is not installed.

agriyakhetarpal:
Also, could we update the documentation for the optional dependencies section in the Contributing guide in this PR as necessary, as we had discussed?

pybamm/citations.py
lorenzofavaro: All right, yes we can definitely guard `pybtex` imports for the moment :)
I changed this in e02662f

pybamm/citations.py
```diff
@@ -67,29 +67,41 @@ def read_citations(self):
...
+        try:
+            parse_file = import_optional_dependency("pybtex.database", "parse_file")
+            citations_file = os.path.join(pybamm.root_dir(), "pybamm", "CITATIONS.bib")
+            bib_data = parse_file(citations_file, bib_format="bibtex")
+            for key, entry in bib_data.entries.items():
+                self._add_citation(key, entry)
+        except ModuleNotFoundError:
+            pybamm.logger.warning(
+                "Citations could not be read because the 'pybtex' library is not installed. "
+                "Install 'pybtex' to enable citation reading."
```
agriyakhetarpal, line 79:
```suggestion
                "Install 'pybamm[cite]' to enable citation reading."
```

I felt that adding the link to the documentation might make the error message too long and difficult to scroll, so maybe we can use `pybamm[cite]` for now.

pybamm/citations.py
```diff
@@ -67,29 +67,41 @@ def read_citations(self):
...
+            new_citation = entry.to_string("bibtex")
+            if key in self._all_citations and new_citation != self._all_citations[key]:
+                warnings.warn(f"Replacing citation for {key}", stacklevel=2)
+
+            # Add to database
+            self._all_citations[key] = new_citation
+        except ModuleNotFoundError:
+            pybamm.logger.warning(
+                f"Could not add citation for '{key}' because the 'pybtex' library is not installed. "
+                "Install 'pybtex' to enable adding citations."
```
agriyakhetarpal, line 103:
```suggestion
                "Install 'pybamm[cite]' to enable adding citations."
```

pybamm/citations.py
```diff
@@ -143,24 +155,32 @@ def _parse_citation(self, key):
...
+                    return
+            except PybtexError as error:
+                # Unable to parse / unknown key
+                raise KeyError(
+                    f"Not a bibtex citation or known citation: {key}"
+                ) from error
+        except ModuleNotFoundError:
+            pybamm.logger.warning(
+                f"Could not parse citation for '{key}' because the 'pybtex' library is not installed. "
+                "Install 'pybtex' to enable citation parsing."
```
agriyakhetarpal, line 182:
```suggestion
                "Install 'pybamm[cite]' to enable citation parsing."
```

pybamm/citations.py
```diff
@@ -211,38 +231,44 @@ def print(self, filename=None, output_format="text", verbose=False):
...
+                print(citations)
+                if verbose:
+                    self._tag_citations()  # pragma: no cover
+            else:
+                with open(filename, "w") as f:
+                    f.write(citations)
+        except ModuleNotFoundError:
+            pybamm.logger.warning(
+                "Could not print citations because the 'pybtex' library is not installed. "
+                "Please, install 'pybtex' to print citations."
```
agriyakhetarpal, line 270:
```suggestion
                "Please, install 'pybamm[cite]' to print citations."
```

pybamm/citations.py
lorenzofavaro: Right, thanks!

agriyakhetarpal (approved):
Looks all great to me now, thanks for the fixes, @lorenzofavaro! Let's hope CI passes on this one, I have no other comments.

agriyakhetarpal:
@all-contributors please add @lorenzofavaro for docs

CONTRIBUTING.md
```diff
@@ -126,26 +126,18 @@ This allows people to (1) use PyBaMM without importing optional dependencies by
...
-        ):
-            sys.modules["pybtex"] = None
-            pybamm.function_using_pybtex(x, y, z)
-
-        # Test that the function works when pybtex is available
-        sys.modules["pybtex"] = pybamm.util.import_optional_dependency("pybtex")
-        pybamm.function_using_pybtex(x, y, z)
-```
+Currently, there are three functions to test what concerns optional dependencies:
+- `test_import_optional_dependency`
```
arjxn-py, line 132: It's be nice to link the valid URLs of these functions so that one can be directed to the code.

pybamm/citations.py
```diff
@@ -143,24 +155,32 @@ def _parse_citation(self, key):
...
+                    self._add_citation(key, entry)
+                    # Add to _papers_to_cite set
+                    self._papers_to_cite.add(key)
+                    return
+            except PybtexError as error:
+                # Unable to parse / unknown key
+                raise KeyError(
+                    f"Not a bibtex citation or known citation: {key}"
+                ) from error
+        except ModuleNotFoundError:  # pragma: no cover
```
arjxn-py, line 179: For now it's alright but can u add related specific testcase instead of using `# pragma: no cover`, Maybe in other PR?

arjxn-py (approved):
Thanks @lorenzofavaro, This looks good to me and I'd be happy to merge this seeing the priority. but -

CONTRIBUTING.md
agriyakhetarpal: The URLs for the tests can change and their permalinks for the relevant can also get outdated (in case we move the tests to somewhere else, rewrite the tests later, etc.), do we have a better way to link these?

pybamm/citations.py
agriyakhetarpal: I personally think it's fine to skip this block – were testing for the other errors in `test_citations.py` and this will get caught in other ways if it fails, such as through `test_pybamm_import`. What do you think?

Saransh-cpp (approved):

agriyakhetarpal:
Resolving previous conversations for now and all tests are passing – let's merge

arjxn-py:
Yes, the macOS integration failure looks unrelated

Issue #3392, opened 2023-10-02
Improve documentation (ongoing)


brosaplanella:
There are quite a few features that are not in the docs, and which would be fairly minor fixes. Opening a list here so we can keep track of them as we go and fix them in batches. Feel free to add more in the comments, and we can update the list.

- [x] `print_parameter_info` not documented in `base_model.py` (#3584)
- [ ] Experiment steps docstrings do not mention that arrays can be passed as values for drive cycles.

User Guide
- [ ] *Submodel docs*: At the moment the documentation for the base models in pyamm just points to the relevent papers, but this is not really sufficient given that users can customise each model via submodels. As a proposal, I think there should be hierarchical user documentation on each of the models. At the top level there are the three main *default* models (SPM, SPMe, DFN), with a description of the relevent equations that is divided up using the submodel structure, so that a user can see the list of equations that are in each submodel, *along with the parameters in that submodel* (at the moment, its very diffiicult to know what parameters are needed for what submodels). Below this there should be an individual discussion of each submodel and the different options for each one. For each option the corresponding equations and parameters should be provided. Incompatible submodels should be highlighted.   (edit by @martinjrobins)
- [ ] *Getting started tutorials*: The getting started tutorials should have an expanded section on submodels, how to vary them, and how to determine what parameters are needed by your particular model configuration. (@martinjrobins)
- [ ] *Getting started tutorials*: These docs introduce a lot of user-facing pybamm classes like `Simulation`, and should link to the API docs for these classes so that users can see the full range of options for these classes/functions. (edit by @martinjrobins)
- [x] #4532

- [ ] The user-guide, or main page, needs a "How to get help" section explictly pointing to discussions, issues, slack channel and explaining how to use each (@martinjrobins)
- [x] In the install from source we should make clearer that you do not need to create a virtual environment if using nox, as nox will create one for you.

API Docs
- [ ] The overall structure of the classes is not documented or explained. Needs to be a diagram, plus a lot of explanation on the class structure and dependencies (@martinjrobins )
- [x] #4535
- [ ] Generally make clear where classes implement __getitem__ and/or __setitem__ and so these are the expected calling patterns. (from @ejfdickinson )
- [ ] `pybamm.Symbol`. This is only a one-liner docstring, needs to explain the overall expression tree structure, with links to example notebooks, and include some examples of use, particularly explaining how to use all the overloaded operators (@martinjrobins )
[...]

PR #3970, opened 2024-04-08
docstring fixes to pybamm.plotting.quick_plot

neilsengupta-elysia:
# Description

Practise contributing to issue for Hackathon:

In the file pybamm.plotting.quick_plot.py, added "optional" keyword to "step" argument in dynamic_plot function. Removed round brackets around optional keywords in the arguments "number_of_images", "duration", "output_filename" in create_gif function.

Related #3392 - Improve documentation

## Type of change
Documentation Update

### DIFF
--- a/pybamm/plotting/quick_plot.py
+++ b/pybamm/plotting/quick_plot.py
@@ -464,6 +464,8 @@ class QuickPlot:
         ----------
         t : float
             Dimensional time (in 'time_units') at which to plot.
+        dynamic : bool
+
         """
 
         plt = import_optional_dependency("matplotlib.pyplot")
@@ -647,7 +649,7 @@ class QuickPlot:
 
         Parameters
         ----------
-        step : float
+        step : float, optional
             For notebook mode, size of steps to allow in the slider. Defaults to 1/100th
             of the total time.
         show_plot : bool, optional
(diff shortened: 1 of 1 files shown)


pybamm/plotting/quick_plot.py
```diff
@@ -464,6 +464,8 @@ def plot(self, t, dynamic=False):
         ----------
         t : float
             Dimensional time (in 'time_units') at which to plot.
+        dynamic : bool
+
```
brosaplanella, line 468: You could include a short description, e.g. "Whether the plot should be dynamic or not"

brosaplanella (commented):
Just a small comment, but looks good overall!

pybamm/plotting/quick_plot.py
kratman: It is also optional

kratman (commented):
While the syntax used in your issue description
```
Fixes # Issue 3392 - Improve documentation (partial fix)
```
will not work because it is incorrectly formatted, you should not mark a partial fix as with the `fixes` tag. When used correctly the `fixes` keyword causes the issue to be closed. This would me that this 10 line change would close a much larger ticket.

Usually you can do
```
Related #3392
```
to tag the issue without closing it

pybamm/plotting/quick_plot.py
neilsengupta-elysia: added a short description, as well as the optional tag.

neilsengupta-elysia:
Fixed PR description to remove "fixes" keyword

agriyakhetarpal (approved):

kratman:
@neilsengupta-elysia I updated your PR description. Having `Related # issue 3392` does not work, it needs to be `Related #3392`. If there are any spaces between # and the number it will not link properly

kratman (approved):

PR #3975, opened 2024-04-09
Standardise BaseSolver.solve() multiprocessing context to `fork`

### LINKED ISSUE #3974: [Bug]: BaseSolver.solve() multiprocessing is broken on system's defaulting to `spawn` context
### PyBaMM Version

24.1

### Python Version

3.11

### Describe the bug

Multiprocessing inputs within BaseSolver.solve() is broken on operating systems that default to `spawn` (Windows, MacOS) instead of `fork` (Linux). A minimal working example is shown below; however, when run inside a [Jupyter notebook](https://github.com/pybamm-team/PyBaMM/blob/develop/docs/source/examples/notebooks/parameterization/change-input-current.ipynb), the code completes without problem, which seems to be how it didn't get caught in tests. I suspect this is due to the difference in call structure, i.e. jupyter might wrap the cells in `main()`. The code in question is: https://github.com/pybamm-team/PyBaMM/blob/c204bdd891d886c955b01bc14049ace5167ed2c3/pybamm/solvers/base_solver.py#L915

Modifying this to `with mp.get_context("fork").Pool(processes=nproc) as p:` appears to solve the issue; however, as of [python 3.14](https://discuss.python.org/t/switching-default-multiprocessing-context-to-spawn-on-posix-as-well/21868/23), `fork` is being depreciated for `spawn` or `forkserver`.

Linking https://github.com/pybop-team/PyBOP/pull/283

### Steps to Reproduce

MWE:
```python
import pybamm
import numpy as np

# create the model

# set the default model parameters

# change the current function to be an input parameter
param["Current function [A]"] = "[input]"

simulation = pybamm.Simulation(model, parameter_values=param)

# solve the model at the given time points, passing the current as an input
t_eval = np.linspace(0, 600, 300)
current_dict = [{"Current function [A]": x} for x in range(1,3)]
sol = simulation.solve(t_eval, inputs=current_dict)
```

### Relevant log output

```shell
RuntimeError:
        An attempt has been made to start a new process before the
        current process has finished its bootstrapping phase.

        This probably means that you are not using fork to start your
        child processes and you have forgotten to use the proper idiom
        in the main module:

            if __name__ == '__main__':
                freeze_support()
                ...

        The "freeze_support()" line can be omitted if the program
        is not going to be frozen to produce an executable.

        To fix this issue, refer to the "Safe importing of main module"
        section in https://docs.python.org/3/library/multiprocessing.html
```

BradyPlanden:
# Description

This PR updated the multiprocess pool to be constructed with context `fork` for `darwin` and `linux` platforms. It also adds an example and updates integration tests to use the list of dictionaries input functionality which should ensure this get tested in script formation.

Fixes #3974 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/tests/integration/test_models/test_full_battery_models/test_lithium_ion/test_compare_outputs_two_phase.py
+++ b/tests/integration/test_models/test_full_battery_models/test_lithium_ion/test_compare_outputs_two_phase.py
@@ -145,8 +145,8 @@ class TestCompareOutputsTwoPhase(TestCase):
 
         sim = pybamm.Simulation(model, parameter_values=param)
         t_eval = np.linspace(0, 9000, 1000)
-        sol1 = sim.solve(t_eval, inputs={"x": 0.01})
-        sol2 = sim.solve(t_eval, inputs={"x": 0.1})
+        inputs = [{"x": 0.01}, {"x": 0.1}]
+        sol = sim.solve(t_eval, inputs=inputs)
 
         # Starting values should be close
         for var in [
@@ -155,11 +155,11 @@ class TestCompareOutputsTwoPhase(TestCase):
             "Average negative secondary particle concentration",
         ]:
             np.testing.assert_allclose(
-                sol1[var].data[:20], sol2[var].data[:20], rtol=1e-2
+                sol[0][var].data[:20], sol[1][var].data[:20], rtol=1e-2
             )
 
         # More silicon means longer sim
-        self.assertLess(sol1["Time [s]"].data[-1], sol2["Time [s]"].data[-1])
+        self.assertLess(sol[0]["Time [s]"].data[-1], sol[1]["Time [s]"].data[-1])
 
     def test_compare_SPM_silicon_graphite(self):
         model_class = pybamm.lithium_ion.SPM
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Features
 
+- Updates multiprocess `Pool` in `BaseSolver.solve()` to be constructed with context `fork`. Adds small example for multiprocess inputs. ([#3974](https://github.com/pybamm-team/PyBaMM/pull/3974))
 - Added custom experiment steps ([#3835](https://github.com/pybamm-team/PyBaMM/pull/3835))
 - Added support for macOS arm64 (M-series) platforms. ([#3789](https://github.com/pybamm-team/PyBaMM/pull/3789))
 - Added the ability to specify a custom solver tolerance in `get_initial_stoichiometries` and related functions ([#3714](https://github.com/pybamm-team/PyBaMM/pull/3714))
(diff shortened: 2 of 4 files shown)


tests/integration/test_models/test_full_battery_models/test_lithium_ion/test_compare_outputs_two_phase.py
```diff
@@ -155,11 +155,11 @@ def compare_outputs_two_phase_silicon_graphite(self, model_class):
             "Average negative secondary particle concentration",
         ]:
             np.testing.assert_allclose(
-                sol1[var].data[:20], sol2[var].data[:20], rtol=1e-2
+                sol[0][var].data[:20], sol[1][var].data[:20], rtol=1e-2
             )
 
         # More silicon means longer sim
-        self.assertLess(sol1["Time [s]"].data[-1], sol2["Time [s]"].data[-1])
+        self.assertLess(sol[0]["Time [s]"].data[-1], sol[1]["Time [s]"].data[-1])
```
kratman, line 162: Time tests have been a bit problematic. I would recommend removing this part of the test.

Runtime can vary based on what else is running on the system, but the tests are run in parallel. So unless the time difference is quite large, the test will be fragile

kratman (changes requested):

tests/integration/test_models/test_full_battery_models/test_lithium_ion/test_compare_outputs_two_phase.py
tinosulzer: It could be replaced by testing "Discharge capacity [A.h]" instead to make it clearer

tinosulzer (approved):

agriyakhetarpal:
The Windows tests are still failing, I guess it is related to the fact that `fork()` syscalls are not available on Windows (can we fix this just conditionally in case the MWE you posted works on Windows but fails on Unix-like systems?)

BradyPlanden:
This is now ready for merge for those that have permissions.

It's worth noting that this only fixes #3974 for `darwin` systems and protects `linux` systems from the default context change in Python 3.14. `Windows` systems will still require the `__main__` guard discussed in #3974. 

Longer term, @martinjrobins and I discussed moving to a more robust solution by implementing the multiprocessing in C++.

martinjrobins (approved):
looks good, just need approval from @kratman, then can merge

BradyPlanden:
I would recommend we merge this and reopen (or unlink) #3974 to continue discussion on a longer-term solution that doesn't depend on the `multiprocessing`. Open to other suggestions though.

kratman:
> I would recommend we merge this and reopen (or unlink) #3974 to continue discussion on a longer-term solution that doesn't depend on the `multiprocessing`. Open to other suggestions though.

Yeah I think it should be unlinked or a new ticket opened for the windows side

kratman (approved):

PR #3981, opened 2024-04-10
Removed `odes` build-time argument from `docker.yml` workflow

santacodes:
# Description

Removed `odes` build-time argument in `docker.yml` workflow file in addition to #3932. I think it was missed.

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        build-args: ["No solvers", "JAX", "ODES", "IDAKLU", "ALL"]
+        build-args: ["No solvers", "JAX", "IDAKLU", "ALL"]
       fail-fast: true
 
     steps:


agriyakhetarpal (approved):

arjxn-py (approved):

agriyakhetarpal:
I think we should remove the existing ODES image now as well (cc: @arjxn-py @Saransh-cpp because I do not have permissions for the Docker Hub stuff)

arjxn-py:
> I think we should remove the existing ODES image now as well (cc: @arjxn-py @Saransh-cpp because I do not have permissions for the Docker Hub stuff)

Yes, I'll look into the same if removing existing tags permanently is possible and will update about the latter.

arjxn-py:
> I think we should remove the existing ODES image now as well

ODES tag have been removed successfully, as other tags are also deprecated we can keep them as it is till the next rc or release maybe and they can also be removed further upon agreement

agriyakhetarpal:
> ODES tag have been removed successfully, as other tags are also deprecated we can keep them as it is till the next rc or release maybe and they can also be removed further upon agreement

Thanks, @arjxn-py. I would be happy to remove the other ones too in case not a lot of people are using them. https://hub.docker.com/r/pybamm/pybamm/tags shows 6.7K pulls, but, alas, we don't have data for a specific tag's pulls (or do we?)

arjxn-py:
> we don't have data for a specific tag's pulls (or do we?)

No we don't 😬 , I'd have no worries about removing the other deprecated tags hence that'll be done.

agriyakhetarpal:
Okay, will leave it up to you on when we should do this :)

PR #3988, opened 2024-04-10
Remove RMSE function

### LINKED ISSUE #3982: pybamm.rmse. document return type and args types
### PyBaMM Version

23.9

### Python Version

3.11.7

### Describe the bug

Minor issue for the PyBaMM training workshop

### Steps to Reproduce

Read the documentation of pybamm.rmse

### Relevant log output

_No response_

kratman:
# Description

Minor removal of code and some cleanup of IDE warnings

Fixes #3982

## Type of change

Removes old code

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/__init__.py
+++ b/pybamm/__init__.py
@@ -1,38 +1,13 @@
 import sys
 import os
 
-
-#
-# Version info
-#
 from pybamm.version import __version__
-
-#
-# Constants
-#
-# Float format: a float can be converted to a 17 digit decimal and back without
-# loss of information
-FLOAT_FORMAT = "{: .17e}"
-# Absolute path to the PyBaMM repo
-script_path = os.path.abspath(__file__)
 
 from .util import root_dir
 
-ABSOLUTE_PATH = root_dir()
-PARAMETER_PATH = [
-    root_dir(),
-    os.getcwd(),
-    os.path.join(root_dir(), "pybamm", "input", "parameters"),
-]
-
-
-#
-# Utility classes and methods
-#
 from .util import Timer, TimerTime, FuzzyDict
 from .util import (
     root_dir,
-    rmse,
     load,
     is_constant_and_can_evaluate,
 )
@@ -48,9 +23,6 @@ from .logger import logger, set_logging_level, get_new_logger
 from .settings import settings
 from .citations import Citations, citations, print_citations
 
-#
-# Classes for the Expression Tree
-#
 from .expression_tree.symbol import *
 from .expression_tree.binary_operators import *
 from .expression_tree.concatenations import *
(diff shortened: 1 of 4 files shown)


pybamm/__init__.py
```diff
@@ -214,38 +160,38 @@
...
+FLOAT_FORMAT = "{: .17e}"
+# Absolute path to the PyBaMM repo
+script_path = os.path.abspath(__file__)
+
+ABSOLUTE_PATH = root_dir()
+PARAMETER_PATH = [
+    root_dir(),
+    os.getcwd(),
+    os.path.join(root_dir(), "pybamm", "input", "parameters"),
+]
```
kratman, line 194: @tinosulzer These don't appear to be used anywhere are they OK to remove as well?

tinosulzer (changes requested):
All the code removed is fine. Please don't remove those comments. They add structure to the code and make it easier to scan. Removing the empty comment lines (`#` and nothing else) is fine.

kratman:
> All the code removed is fine. Please don't remove those comments. They add structure to the code and make it easier to scan. Removing the empty comment lines (`#` and nothing else) is fine. 

I removed them because they basically all said something along the lines of
```python
# functions from file X
from x import y
```
Which does not seem to add anything

tinosulzer:
It adds structure and makes the file easier to scan since they are highlighted in different colors by most IDEs

Saransh-cpp:
Imports in pybamm's `__init__.py` file must be in a particular order (autoformatting tools usually mess up this order), so adding a structure with the comments makes sense.

kratman:
> It adds structure and makes the file easier to scan since they are highlighted in different colors by most IDEs

To me it did not add anything since it literally said what was right below that in the code, but I will add it back if you insist. In general my changes like this are just trying to clean up stuff that does not add anything like:
```python
#
# Batch Study
#
from .batch_study import BatchStudy
```
which ends up adding a lot of comments for the exact thing that is below it.

> Imports in pybamm's `__init__.py` file must be in a particular order (autoformatting tools usually mess up this order), so adding a structure with the comments makes sense.

I did not run any auto-formatting here and left the imports in groups. The only thing I did for my IDE was move the constants to the bottom of the file in order to stop the warnings about imports coming after code. The comments did not add any hints about which order things needed to be imported, no warnings about problematic stuff, etc. So I don't think they solve the import order problem either

However, the constants I moved look like they are not actually used and can probably be removed as well. PyBaMM triggers so many warnings in PyCharm that I often cannot see the real problems from the inspections. I try to clean up what I can as I work

kratman:
Ok everything here is passing now besides Lychee

valentinsulzer (approved):

PR #3989, opened 2024-04-10
Fix Lychee link failures from stack overflow

kratman:
# Description

Removes stack overflow links due to lychee failures

## Type of change

Fixes Lychee link checker

# Key checklist:

## Further checks:


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -34,7 +34,7 @@ body:
     id: reproduce
     attributes:
       label: Steps to Reproduce
-      description: Tell us how to reproduce this behaviour. Ideally, this should take the form of a [Minimum Workable Example](https://stackoverflow.com/help/minimal-reproducible-example)
+      description: Tell us how to reproduce this behaviour. Ideally, this should take the form of a minimum workable example.
```
tinosulzer, line 37: We should tell users what a minimum workable example is

CONTRIBUTING.md
```diff
@@ -87,7 +90,8 @@ Class names are CamelCase, and start with an upper case letter, for example `MyO
...
 For these reasons, all dependencies in PyBaMM should be thought about carefully, and discussed on GitHub.
 
-Direct inclusion of code from other packages is possible, as long as their license permits it and is compatible with ours, but again should be considered carefully and discussed in the group. Snippets from blogs and [stackoverflow](https://stackoverflow.com/) can often be included without attribution, but if they solve a particularly nasty problem (or are very hard to read) it's often a good idea to attribute (and document) them, by making a comment with a link in the source code.
+Direct inclusion of code from other packages is possible, as long as their
+license permits it and is compatible with ours, but again should be considered carefully and discussed in the group. Snippets from blogs can often be included without attribution, but if they solve a particularly nasty problem (or are very hard to read) it's often a good idea to attribute (and document) them, by making a comment with a link in the source code.
```
tinosulzer, line 94: link removal is fine by why remove "and stackoverflow"?

pybamm/solvers/solution.py
```diff
@@ -15,7 +15,6 @@
 class NumpyEncoder(json.JSONEncoder):
     """
     Numpy serialiser helper class that converts numpy arrays to a list
```
tinosulzer, line 17: Since the link is removed, add explanation about why this is required

tinosulzer (changes requested):

CONTRIBUTING.md
kratman: I can add plain text stack overflow in again. I just figured it was redundant without the links

agriyakhetarpal:
I would be slightly against this change. Usually link checkers are used to prevent link rot and make sure that dead links are not scattered across codebases. However, StackOverflow permalinks will never go extinct as long as StackOverflow never goes down, so, could we ignore the website altogether?

pybamm/logger.py
```diff
@@ -1,11 +1,3 @@
-#
-# Logging class for PyBaMM
-# Includes additional logging levels inspired by verboselogs
-# https://pypi.org/project/verboselogs/#overview-of-logging-levels
-#
-# Implementation from stackoverflow
-# https://stackoverflow.com/questions/2183233/how-to-add-a-custom-loglevel-to-pythons-logging-facility
-#
```
Saransh-cpp, line 8: Should not remove this. Stackoverflow code should be attributed. This goes against our own guidelines in CONTRIBUTING.md haha!

Saransh-cpp (dismissed):
> However, StackOverflow permalinks will never go extinct as long as StackOverflow never goes down, so, could we ignore the website altogether?

+1 for this.

Updating `numpy` and `python` links is a good idea.

pybamm/logger.py
kratman: Contributing guidelines: "Snippets from blogs and [stackoverflow](https://stackoverflow.com/) can often be included without attribution"

kratman:
@agriyakhetarpal, @Saransh-cpp In general I don't think adding links to code snippets is all that useful. I don't think many people would just copy the link to look at it. They would then have to read a whole blog post or forum post just to find the key pieces of information. If there is something that important in the link, then it should be explained in plain text comments. I strongly believe comments should be minimized, but their main purpose is to explain extremely complex parts of the code. If somebody changes a code block commented by a link, then the person updating that code block has to read the link to determine if they changed what was commented in there. This can easily lead to code rot in the comments.

While we can mostly assume that some sites will never change their domains or delete pages, the whole point of link checking is to confirm that the assumption is correct. If we suppress any site which is not perfect for the link checker, then what is the point of the link checker? There are some things like the paths to executables which the link checker will never be able to handle, but the point is to check if links work. If somebody adds a new stack overflow link in a comment do you think all reviewers are going to copy the link to check it or do you think they will see the link checker pass and assume that the link is correct?

Additionally the link checker has repeatedly marked PRs as failing their CI checks, which hides real problems. I am going to go back through and update the comments in the parts @tinosulzer flagged, but I think in general reducing the ways in which things can fail or break is an improvement.

.github/ISSUE_TEMPLATE/bug_report.yml
kratman: Done

CONTRIBUTING.md
kratman: Done

pybamm/logger.py
kratman: The coding guidelines say that we do not have to do the attribution

pybamm/solvers/solution.py
kratman: Done

kratman:
Real link failure that was hidden by the StackOverflow ones:
[https://github.com/keras-team/keras/blob/master/keras/callbacks/callback.py](https://github.com/keras-team/keras/blob/master/keras/callbacks/callback.py) | Failed: Network error: Not Found

kratman:
> I had pitched for the creation of an automated issue in a workflow that runs fortnightly (or monthly would be better) instead of on PRs

I think it is better to have the link checker run on individual PRs if we are going to do it. The purpose should be to check if a new change has added a bad link

kratman:
@valentinsulzer, @Saransh-cpp Can you re-review I would like to get this link checker fixed

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -34,7 +34,7 @@ body:
     id: reproduce
     attributes:
       label: Steps to Reproduce
-      description: Tell us how to reproduce this behaviour. Ideally, this should take the form of a [Minimum Workable Example](https://stackoverflow.com/help/minimal-reproducible-example)
+      description: Tell us how to reproduce this behaviour. Ideally, this should include a code block which produces the error.
```
Saransh-cpp, line 37: A code block is not necessarily always an MWE.

.readthedocs.yaml
```diff
@@ -25,16 +20,11 @@ build:
...
-    # You can also specify other tool versions:
-    # nodejs: "19"
-    # rust: "1.64"
-    # golang: "1.19"
   jobs:
     # Unshallow the git clone otherwise this may cause issues with Sphinx extensions
     post_checkout:
       - git fetch --unshallow
-    # Altered PDF build and upload job, attributed to
-    # https://stackoverflow.com/a/76992101/14001839
```
Saransh-cpp, line 37: I'll still prefer attributing StackOverflow answers. Moreover, in this case, the question was asked by @agriyakhetarpal on StackOverflow to specifically cater to the needs of PyBaMM.

pybamm/callbacks.py
```diff
@@ -1,10 +1,3 @@
-#
-# Base class for callbacks and some useful callbacks for pybamm
-# Callbacks are used to perform actions (e.g. logging, saving)
-# at certain points in the simulation
-# Inspired by Keras callbacks
-# https://github.com/keras-team/keras/blob/master/keras/callbacks/callback.py
```
Saransh-cpp, line 6: Can we replace this with the updated link? It is often useful to look at the original implementations, especially for people trying to extend PyBaMM (or any software) and not just use it.

pybamm/logger.py
Saransh-cpp, line 8: Ditto!

.github/ISSUE_TEMPLATE/bug_report.yml
Saransh-cpp: This is the only comment that is a blocker from my end.

Saransh-cpp (commented):
> If somebody adds a new stack overflow link in a comment do you think all reviewers are going to copy the link to check it or do you think they will see the link checker pass and assume that the link is correct?

This is an issue, but I believe the person reviewing the PR will most probably look at the link to understand the implementation.

I would still prefer to keep the StackOverflow links. As I mentioned below, they're not useful for direct users but are somewhat useful for people trying to extend the software or depend on it as a library. The majority (maybe all) of the libraries that I work on attribute StackOverflow answers and even GitHub issues/discussions in comments. I know that I will definitely go through the links attached in the comments when I'm trying to understand a functionality on a deeper level.

I won't block the PR. Please feel free to merge if everyone else is happy with it.

pybamm/solvers/solution.py
```diff
@@ -14,8 +14,9 @@
 
 class NumpyEncoder(json.JSONEncoder):
     """
-    Numpy serialiser helper class that converts numpy arrays to a list
-    https://stackoverflow.com/questions/26646362/numpy-array-is-not-json-serializable
```
Saransh-cpp, line 18: Ditto!

pybamm/callbacks.py
agriyakhetarpal: Replacing with a GitHub permalink (pinned to a commit hash) would also be helpful, so that we can know what the implementation was at the time of writing and the link won't break.

.github/ISSUE_TEMPLATE/bug_report.yml
kratman: I opted to do it this way because explaining what a MWE is was quite wordy, so instead of even mentioning that, I just switched it to asking for them to put in code that produces the error.

pybamm/callbacks.py
kratman: I don't think we need to mark were we were inspired by other projects. Especially when concepts like callbacks are pretty common.

We should use links only where it is necessary. The fact is that this keras link broke just 6 months ago too and this time it was not appearing in the link checker log until the stackoverflow stuff was removed.

pybamm/solvers/solution.py
kratman: I added plain text that explains, so we don't need the link anymore. If the explanation is not good enough then we can just update the text

.github/ISSUE_TEMPLATE/bug_report.yml
agriyakhetarpal: Perhaps could we add the link to Wikipedia – https://en.wikipedia.org/wiki/Minimal_reproducible_example, it shouldn't fail as such? I checked it with `lychee` locally and it works as usual

kratman: But why even bother with a link when we can just use words

agriyakhetarpal: Makes sense, could you an extended description? Like not to be excessively wordy, but just another sentence about what additional context people can provide besides their snippets in code blocks

valentinsulzer, line 37:
```suggestion
      description: Tell us how to reproduce this behaviour. Ideally, this should include a code block which produces the error. Strive to make this example as small and simple as possible: it should contain the code required to reproduce the error, and no additional code. For example, if your code includes lines to run a simulation, then lines to plot the results, and the lines to run the simulation fail, only include those lines (and not the plotting lines, which are irrelevant). Often, the act of simplifying code to pinpoint errors can help you find bugs in your own code. For more information, see [these references](https://en.wikipedia.org/wiki/Minimal_reproducible_example#External_links)
```

valentinsulzer (changes requested):
In the interest of closing this bikeshed PR, I'm happy to merge once the suggested change is made.

valentinsulzer (approved):

valentinsulzer:
Thanks! @Saransh-cpp I think that addresses your main concern as well - just waiting for your approval now

agriyakhetarpal (approved):
I'm approving as well; it has been a recent revelation that content from StackOverflow links is brought under the CC-BY-SA license, which is not compatible with our BSD-3-Clause license without the due permission of the author of the code snippet being referenced. [SciPy's development guide](https://scipy.github.io/devdocs/dev/core-dev/index.html#licensing) has more information. I would consider adding this to our Contributing Guide somewhere and it should be fine within this PR itself, @kratman – thanks!

kratman:
I dismissed @Saransh-cpp's review since he said previously:
> Please feel free to merge if everyone else is happy with it.

Issue #3908, opened 2024-03-20
[Roadmap area] Model refactoring


brosaplanella:
We don't have as many contributors adding models to PyBaMM as we'd like. There are a few reasons for this
- PyBaMM’s submodel structure is too complex, huge barrier to entry
- You have to add things to the PyBaMM repo, uncertain ownership

We propose "model entry points" as a solution to this. Using entry points, contributors can create models in their own repositories and make them easily available to the rest of the community, without the models having to be added to PyBaMM itself. This will enable a community for battery models that use the PyBaMM infrastructure but are not owned by the PyBaMM team.

### Advantages
- Contributors get to keep models in their own repositories, retain ownership, choose license terms
- PyBaMM team doesn’t have to endorse models we don’t like
- All of GitHub’s functionality (issues, discussions, pull requests) are available for each model

### How to start

- Set up a separate repository with an example of adding a model separately outside of PyBaMM
- Make required changes to PyBaMM to enable this
    - Entry points for models
    - Making basic models compatible with experiments
    - Figure out how to deal with new parameters (define parameters inside the model itself?)
- Make cookiecutter repository to help with setting up, adding entry points, publishing to pypi, tests, docs, specifying pybamm version(s), etc (as part of the GSoC coookie-cutter project)

### Later changes

After this is done, we will be in a much better position to consider fully separating the model code and PDE code

_Originally posted by @tinosulzer in https://github.com/pybamm-team/PyBaMM/issues/3839#issuecomment-1966614301_

valentinsulzer:
Some additional thoughts on refactoring submodels. Perhaps deserves its own issue (or three) but relevant to the roadmap.

Although in general we should move towards encouraging more standalone models, the composability of submodels is still useful. The things that make the submodel structure complex are:
1. The relationship between options strings and submodels is opaque
2. Submodels are nested (e.g. `BaseParticle` -> `FickianDiffusion`)
3. Separation into `get_fundamental_variables`, `get_coupled_variables`, `set_rhs`, etc

Each of these can be addressed and simplified, although in all cases it's a big undertaking. 

1. Get rid of options and instead only allow creating models by passing in a dictionary of models. This  is heavily breaking and may not be worth it. The options handle quite a few special cases and putting that onus on the user could be quite annoying.
2. The role of nested submodels is to add a bunch of extra variables, but this could be done with helper functions instead to make it less confusing - it doesn't need to be object oriented. This can be done in a non-breaking way for the front end.
[...]

Issue #1826, opened 2021-11-26
Adding termination events to experiment class


sklein-ise:
Hey, 

in reference to #1796 I wanted to open this issue for future termination events in the experiment class.
It would be nice to set a voltage limit to end the experiment completely. I was also thinking a termination event after each step inside the cycle might be useful.

First one would be helpful for example in GITT experiments, when you're trying to cycle pulses until a voltage limit. Second one could be used when they are multiple instructions like a cycling procedure and a pulsetest.

```
experiment_gitt = pybamm.Experiment(
        [("GITT Pulse")]*60, termination = "Voltage limit"
        )

experiment_aging = pybamm.Experiment(
        [
        ("Pulsetest")+
        ("Cycling procedure")*400, termination = ""
        ]*5, termination = ""
        )
```

PR #3995, opened 2024-04-11
Make basic models compatible with experiments

rtimms:
# Description
Makes "basic" models compatible with experiments. For a model to work with experiments it needs to have 

1. All the `Variable` objects in the `variables` dict
2. The "Current function [A]" and "Battery voltage [V]" variables 
~~3. Define `self.summary_variables`~~

Related #3908 

## Type of change

# Key checklist:

## Further checks:


tinosulzer (commented):
These changes are a step in the right direction, so ok to merge as is if you don't have time to work on it further, but in the longer term it would be better to change how summary variables are processed in `Solution` so that basic models work out of the box without needing to add all the right variables.

Easy fixes:
- make the default summary variables for a base lithium ion model be `[]`, and only set the specific summary variables for DFN, SPM, SPMe, etc
- don't calculate summary variables based on Discharge capacity and Battery voltage if those are not included in `self.summary_variables`

Harder fixes that also address the fact that summary variable calculations can slow down the simulation quite a lot:
- create a `SummaryVariable` object that saves `solution.last_state` and uses it to calculate the esoh variables **only** when they are called (the same way as `Solution` calculates the variables only when they are called)
- get rid of the min/max voltage and capacity calculations as a default and allow users to pass in functions that calculate summary variables based on other variables

rtimms:
Most of the new variables were added so that copying from one model to the next in experiments worked (you need to `Variable` objects in the `variables` dict.

> make the default summary variables for a base lithium ion model be [], and only set the specific summary variables for DFN, SPM, SPMe, etc
where would you put these to avoid code duplication?

> don't calculate summary variables based on Discharge capacity and Battery voltage if those are not included in self.summary_variables
I can do this

I'll leave the harder fixes for a separate issue.

tinosulzer:
> where would you put these to avoid code duplication?

rename `lithium_ion.BaseModel.set_summary_variables` to `set_standard_summary_variables` and only call it from within `DFN.__init__`, `SPM.__init__`, etc (instead of `BaseModel.__init__`).

> "Battery voltage [V]" has to be in variables for experiments to work and isn't in summary_variables

Let's still only calculate the related min/max voltage if it's there (not there by default but possible to add) i.e. change
```python
if "Battery voltage [V]" in model.variables
```
to 
```python
if "Min/max battery voltage [V]" in model.summary_variables
```
and same for "Discharge energy". Calculating the min/max voltage for every step requires evaluating the full voltage, which is much slower than solving the model for SPM/SPMe.

rtimms:
Ah yeah, thanks, obvious implementation now!

For the harder issues, I think we should should store `all_last_states` in the same way we store `all_first_states`, then we can just lazily evaluate summary variables. The only catch is you won’t be able to do things like min/max voltage as you may not have the whole cycle solution (you may not have saved that cycle). I think for summary variables we should make it so they can only depend on the first and last state of a cycle, so you can have things that are (some combination of) variables at the start/end of a cycle. Then you can still do “Change in…” summary variables. 

For custom summary variables I wonder if it’s simpler to just add new variables to the model rather than allowing functions to be passed to summary variables? Then summary variables can still just be a list. 

If this sounds like a sensible approach I’ll do it as part of this PR.

rtimms:
Alternatively we can calculate the “easy” summary variables on the fly and do the above for the esoh. Is that what you had in mind? 

I’m leaning towards just allowing summary variables to depend on first and last state, and if people want something else then they should save that cycle or do some other manual calculation.

rtimms:
> I was talking about custom functions that allow you to do things over different time steps (min/max, integrate, etc), which wouldn't be possible to define as a model variable, but let's leave that for later if there is a good use case

For these you wouldn't be able to do from first/last state anyway, so I say let's leave it for later.

> This is probably ok but we should profile it. There is some overhead to creating a ProcessedVariable which you then never use

If we aren't doing min/max etc then let's just lazily evaluate everything when you call.

brosaplanella:
Sounds good to me. We should keep track of what's needed to build a successful basic model so we can document it well.

rtimms:
@valentinsulzer I would like to merge this and have made a new issue for evaluating summary variables only when they are called https://github.com/pybamm-team/PyBaMM/issues/4058

pybamm/callbacks.py
```diff
@@ -201,20 +201,6 @@ def on_cycle_end(self, logs):
                     f"is below stopping capacity ({cap_stop:.3f} Ah)."
                 )
 
-        voltage_stop = logs["stopping conditions"]["voltage"]
```
rtimms, line 204: We can't do this if we don't allow calculating summary variables on the fly. I'm not sure how useful setting a voltage cut-off at the experiment level in this way is anyway? Surely we would set cut-offs at the step level?

pybamm/models/submodels/external_circuit/base_external_circuit.py
```diff
@@ -41,23 +42,25 @@ def get_fundamental_variables(self):
         return variables
 
     def set_initial_conditions(self, variables):
-        Q_Ah = variables["Discharge capacity [A.h]"]
-        Qt_Ah = variables["Throughput capacity [A.h]"]
-        self.initial_conditions[Q_Ah] = pybamm.Scalar(0)
-        self.initial_conditions[Qt_Ah] = pybamm.Scalar(0)
+        if self.add_discharge_capacity:
```
rtimms, line 45: We add these by default when using this as a submodel, but pass `add_discharge_capacity=False` when creating the submodel to update the "Current [A]" variable for experiments to avoid adding these equations to models that don't already contain them. I'd welcome alternative implementations.

valentinsulzer: How about just having a separate submodel for discharge capacity and related variables?

pybamm/callbacks.py
valentinsulzer: It was added in response to this request https://github.com/pybamm-team/PyBaMM/issues/1826. I completely forgot we had this feature but it's quite useful for GITT! The lines commented here are just for logging, the line that does the check is in the `Simulation` class, we should keep it but evaluate the voltage inside the `Simulation` class instead of using the summary variable. It would only get called if `voltage_stop is not None`, so no performance hit in general.

pybamm/experiment/experiment.py
```diff
@@ -167,12 +167,14 @@ def read_termination(termination):
                         "'80%', '4Ah', or '4A.h'"
                     )
             elif term.endswith("V"):
-                end_discharge_V = term.split("V")[0]
-                termination_dict["voltage"] = (float(end_discharge_V), "V")
+                raise ValueError(
+                    "Voltage termination at the experiment-level is no longer supported. "
+                    "Please set voltage termination conditions at the step level."
+                )
```
valentinsulzer, line 173: See other comment, the use case is where the voltage termination is not expected to be hit in every step e.g. GITT https://github.com/pybamm-team/PyBaMM/issues/1826

valentinsulzer (changes requested):
Main changes:
- Keep voltage termination at the experiment level (e.g. for GITT) but evaluate the voltage directly instead of using the summary variable
- Try having a separate submodel for discharge capacity (there may be a reason why this isn't possible)

valentinsulzer (approved):

PR #4001, opened 2024-04-12
Update multiprocessing context to be Pool specific

### LINKED ISSUE #3997: [Bug]: PyBaMM fails to import when user has already imported a library that uses `mp.set_start_method`
### PyBaMM Version

develop

### Python Version

3.11

### Describe the bug

PyBaMM fails to import when importing a library that uses multiprocessing and has already set the start method to something other than fork, as a result of #3975. This should be apparent when reading the documentation for `mp.set_start_method`

Pending a long-term solution, I think that #3975 should at least have `mp.set_start_method` wrapped in a try-except -- most people, most of the time, do not really need the multiprocessing functionality, and I don't think pybamm should force those users to modify their multiprocessing configuration. It can throw a warning saying multiprocessing is not available or whatever.

Other solutions are welcome, of course. 
 

### Steps to Reproduce

module `thing`:
```
import multiprocessing as mp

mp.set_start_method("spawn")

def hello():
    print("hello, world")
```

script:
```
import thing
import pybamm
``` 

### Relevant log output

```shell
❯ python main.py 
  File "/Users/abills/.../main.py", line 2, in <module>
    import pybamm
  File "/Users/abills/.../python3.11/site-packages/pybamm/__init__.py", line 204, in <module>
    from .solvers.base_solver import BaseSolver
  File "/Users/abills/.../python3.11/site-packages/pybamm/solvers/base_solver.py", line 18, in <module>
    mp.set_start_method("fork")
  File "/Users/abills/.../python3.11/multiprocessing/context.py", line 247, in set_start_method
    raise RuntimeError('context has already been set')
RuntimeError: context has already been set
```

BradyPlanden:
# Description

Convert's the global `mp.set_start_method` context to an individual pool context.

Fixes #3997 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/solvers/base_solver.py
+++ b/pybamm/solvers/base_solver.py
@@ -66,12 +66,7 @@ class BaseSolver:
         self.algebraic_solver = False
         self._on_extrapolation = "warn"
         self.computed_var_fcns = {}
-
-        # Set context for parallel processing depending on the platform
-        if platform.system() == "Darwin" or platform.system() == "Linux":
-            self._mp_context = "fork"
-        else:
-            self._mp_context = "spawn"
+        self._mp_context = self.get_platform_context(platform.system())
 
     @property
     def root_method(self):
@@ -1396,6 +1391,12 @@ class BaseSolver:
                         "You may need to provide additional interpolation points "
                         "outside these bounds."
                     )
+
+    def get_platform_context(self, system_type: str):
+        # Set context for parallel processing depending on the platform
+        if system_type.lower() in ["linux", "darwin"]:
+            return "fork"
+        return "spawn"
 
     @staticmethod
     def _set_up_model_inputs(model, inputs):
--- a/pybamm/solvers/base_solver.py
+++ b/pybamm/solvers/base_solver.py
@@ -13,10 +13,6 @@ import numpy as np
 import pybamm
 from pybamm.expression_tree.binary_operators import _Heaviside
 
-# Set context for parallel processing depending on the platform
-if platform.system() == "Darwin" or platform.system() == "Linux":
-    mp.set_start_method("fork")
-
 
 class BaseSolver:
     """Solve a discretised model.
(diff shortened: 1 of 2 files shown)


BradyPlanden:
This PR is ready for review. It's a small change to fix #3997. Coverage decreases since we only run the CI on Linux machines, open to suggestions on how to capture the missing line, otherwise it's a small decrease. Not sure a CHANGELOG entry is needed, but happy to add one if so.

agriyakhetarpal:
It's not the best solution as such, but as a stop-gap, you can mark the conditions with a `# pragma: no cover` comment. It is also possible to exclude coverage for just a specific platform, like `# pragma: win32 no cover` – not sure about macOS or Linux, though. For them, I think
```python
# pragma: no cover windows # pragma: no cover macos
```

should work, based on GitHub code search results and [this StackOverflow answer](https://stackoverflow.com/q/41778073/14001839) which works based on specifying these lines in `[tool.coverage]` in `pyproject.toml` (and possibly adding another dependency or two in `[dev]` besides `coverage`).

kratman (changes requested):
@agriyakhetarpal, @BradyPlanden I think a general no-cover is better then adding multiple of them, but there is another solution.

If you make a function that takes platform.system() as an argument, then you can write a test for it.
```python
def get_mp_context(system_type: str):
    if system_type.lower() in ["linux", "darwin"]:
        return "fork"
    return "spawn"

self._mp_context = get_mp_context(platform.system())
```

This requires only a trivial test and is not really any extra complexity or assumptions about coverage.

valentinsulzer (dismissed):
[Removing review - missed Eric's comment]

valentinsulzer:
Wouldn't the `return "spawn"` line still not be covered in the suggested function?

I wonder if we have been over-using "no cover" in the chase for 100% coverage. 99.5 % coverage is fine

kratman:
> Wouldn't the `return "spawn"` line still not be covered in the suggested function?
> 
> I wonder if we have been over-using "no cover" in the chase for 100% coverage. 99.5 % coverage is fine

No because you can write a test for the function I put there. My function is basic input output so the test is just a couple lines:

```python
def test_mp_spawn():
    assert get_mp_context("fake_system") == "spawn"
```

kratman:
> I wonder if we have been over-using "no cover" in the chase for 100% coverage. 99.5 % coverage is fine

In general, I feel 100% coverage usually requires faking coverage in some way:
- Tests on large entry points that don't really "test" anything meaningful
- Pragmas to ignore coverage
- etc

However, there are often simple functions that can be made to make trivial tests. Generally places that need long comments or switches to turn off tests/coverage are code smells that show the code needs to be broken up or refactored

kratman (approved):

PR #4021, opened 2024-04-18
add plot_thermal_components

valentinsulzer:
# Description

Similar to `plot_voltage_components`, breaks down the temperature / heating into components.
This currently is oriented towards the lumped thermal model.

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/models/submodels/thermal/base_thermal.py
+++ b/pybamm/models/submodels/thermal/base_thermal.py
@@ -218,13 +218,8 @@ class BaseThermal(pybamm.BaseSubModel):
         Q_rev_vol_av = Q_rev_W / V
         Q_vol_av = Q_W / V
 
-        # Newton cooling, accounting for surface area to volume ratio
-        T_vol_av = variables["Volume-averaged cell temperature [K]"]
-        T_amb = variables["Volume-averaged ambient temperature [K]"]
-        Q_cool_W = -self.param.h_total * (T_vol_av - T_amb) * self.param.A_cooling
-        Q_cool_vol_av = Q_cool_W / V
-
         # Effective heat capacity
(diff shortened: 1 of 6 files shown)


pybamm/models/submodels/thermal/base_thermal.py
```diff
@@ -249,6 +258,12 @@ def _get_standard_coupled_variables(self, variables):
                 # Current collector
                 "Negative current collector Ohmic heating [W.m-3]": Q_ohm_s_cn,
                 "Positive current collector Ohmic heating [W.m-3]": Q_ohm_s_cp,
+                # Lumped cooling
```
rtimms, line 261: Can calculating and adding these variables be a separate method that for now only gets called by the lumped model? Other models handle cooling differently and use different parameters so right now this will either not work as the parameters aren’t there or will work but give a result that may be different from averaging the actual cooling in the model.

rtimms (changes requested):

pybamm/models/submodels/thermal/base_thermal.py
valentinsulzer: I moved it to lumped model. Is there an easy equivalent for x-full or should we leave it to only work with the lumped model?

rtimms: You’ll have to sum the cooling terms over the boundaries, but you’ll have to do all those calculations manually because of how the model is set up. It assumes a pouch and cooling from the two large faces normal to the through cell direction show up as boundary conditions either end. The other cooking all shows up as a source term (start from 3D and integrate over y,z so the boundary terms become sources). 

Probably not worth doing unless someone asks for it. The x-full model isn’t really that useful since it’s just for a single layer.

PR #4023, opened 2024-04-19
Generalised symbol copying

### LINKED ISSUE #3798: Refactor: Generalise copy for node of expression tree
### Description

Provide a general function to copy a `pybamm.Symbol` with a new set of children

### Motivation

Pybamm expresses models using an expression tree and `pybamm.Symbol` and derived classes. It is a useful operation to create a new model from an old one with minor modifications (.e.g parameter values does this), and for this it is useful to have a general function to copy a node of an expression tree with a new set of children. We already have similar functions like `Symbol.new_copy` and `create_copy`, but these do not allow for new children. We have more similar functions for specific derived classes like `_binary_new_copy`, `_unary_new_copy` that *do* allow for new children.  

### Possible Implementation

generalise `new_copy` or `create_copy` to take (optionally) a list of new children. This should work across all derived classes.

This should simplify some of the logic in `pybamm.ParameterValues.process_symbol`

### Additional context

_No response_

pipliggins:
# Description

Provides a general `new_copy()` function for pybamm symbols which can accept new children.
Some symbols are copied within pybamm using private functions like `_unary_new_copy`, usually to use overloaded or convenience functions to perform simplifications before creating a new symbol.
As this behaviour can sometimes result in a different symbol class to the one originally copied when new children are provided, this can be overridden using the `perform_simplifications` argument.

Fixes #3798

## Type of change

# Key checklist:

## Further checks:


pybamm/discretisations/discretisation.py
```diff
@@ -874,7 +874,7 @@ def _process_symbol(self, symbol):
 
         elif isinstance(symbol, pybamm.Function):
             disc_children = [self.process_symbol(child) for child in symbol.children]
-            return symbol._function_new_copy(disc_children)
+            return symbol.new_copy(disc_children)
```
martinjrobins, line 877: can this new_copy replace the _unary_new_copy and _binary_new_copy calls above?

pybamm/expression_tree/averages.py
```diff
@@ -38,9 +39,22 @@ def __init__(self, child: pybamm.Symbol) -> None:
...
+        Creates a new copy of the operator with the child `child`.
+
+        Uses the convenience function :meth:`x_average` to perform checks before
+        creating an XAverage object.
+        """
+        if perform_simplifications:
+            return x_average(child)
+        else:
+            raise NotImplementedError(
+                f"{self.__class__.__name__} should always be copied using "
```
martinjrobins, line 55: why do we have to use simplification checks here?

martinjrobins: and others below....

pybamm/expression_tree/symbol.py
```diff
@@ -952,21 +952,47 @@ def to_casadi(
...
-    def create_copy(self):
+    def _children_for_copying(self, children: list[Symbol] | None = None) -> Symbol:
+        """
+        Gets existing children for a symbol being copied if they aren't provided.
+        """
+        if children is None:
+            children = [child.create_copy() for child in self.children]
+        return children
+
+    def create_copy(
```
martinjrobins, line 963: what is the difference between `create_copy` and `new_copy`? The docs say that `new_copy` adds attributes, but not sure what this means?

martinjrobins (commented):
thanks @pipliggins, this looks great. I had a quick look through and had a couple of questions, see below. Also I was thinking it might be good to put in an integration test that copies a larger expression tree, e.g. the rhs of the DFN model, and evaluates it a couple of times to check the output is the same?

pybamm/expression_tree/symbol.py
pipliggins: new_copy used the print_name setter to ensure the print name was run through the pretty-print functions. No tests broke when I took this out, so I've deprecated new_copy in favour of just using `create_copy()` (to avoid confusion with `model.new_copy()`)

pybamm/expression_tree/averages.py
pipliggins: Turns out we didn't, I've removed these to follow the same format as other overwritten functions

martinjrobins (approved):
thanks @pipliggins for resolving these, looks great!

PR #4028, opened 2024-04-20
Adds additional tests for Casadi modes

### LINKED ISSUE #4027: Additional tests for Casadi modes
### Description

At the moment, I believe the Casadi solver modes are tested on a toy model: https://github.com/pybamm-team/PyBaMM/blob/90c1c357a97dfd5c8c6a9092a70dddf0dac978db/tests/unit/test_solvers/test_casadi_solver.py#L17, but not formally across the particle models.

This issue is to add additional tests on an SPM or equivalent with assert for ensuring the results for each solver mode. 

### Motivation

As we are moving towards the `"safe with events"` mode in PyBOP, it would be ideal to have the mode tested for across the pybamm models.

### Possible Implementation

_No response_

### Additional context

_No response_

BradyPlanden:
# Description

This PR adds additional tests for the Casadi modes: `"safe"`, `"fast"`, `"fast with events"` with asserts comparing the results from each mode on an SPM.

Fixes #4027 

## Type of change

# Key checklist:

## Further checks:


agriyakhetarpal:
The notebook that timed out (`loss_of_active_materials.ipynb`) in the notebook tests has been doing so much more frequently recently, I just triggered a re-run (twice) and it still fails. It's not entirely clear from the logs which code cell is timing out and I don't see it happening when I try locally. Maybe we should ignore the `pip install` cells in the notebooks from execution (https://github.com/treebeardtech/nbmake#ignore-a-code-cell)?

martinjrobins (changes requested):
as I understand it this is not testing anything new, just testing the different solver modes with the SPM model instead of toy models? If so I think it should probably live in the integration tests (e.g. in https://github.com/pybamm-team/PyBaMM/tree/develop/tests/integration/test_solvers)

PR #4040, opened 2024-04-23
Adds recursive `__all__` dictionary

arjxn-py:
# Description

Been proposed in #3866 

## Type of change

# Key checklist:

## Further checks:


arjxn-py:
Also related to #2427

agriyakhetarpal (commented):
Thanks, @arjxn-py – overall, I believe that adding `__all__` to the codebase makes a lot of sense, it's the more modern way of bringing structure to the public API, plus providing efficient tab completions for IDEs and text editors.

I wonder if doing this in a staggering manner, i.e., across multiple `pybamm/module/submodule/__init__.py` files makes more sense, in comparison to adding them in the same file? We have way too many changes in a single file right now, it would become difficult to keep track of where these symbols and attributes come from (some in-line comments could also be of help). Keeping everything split across files will be just a one-time change to review (though it would have to be thorough), and then adding to items to multiple `__all__` lists will be simpler, too, since they will be smaller. I suppose you're using the `mkinit` tool right now that has come up in our previous discussions – the main issue with it was that we would have needed to ask contributors to run a command every time they modified something for the API, and we don't want that additional complexity. Splitting this across multiple files should make the structure easier to follow and then manually modify on future changes, without the use of the tool.

pybamm/solvers/c_solvers/__init__.py
```diff
@@ -0,0 +1 @@
+__all__ = []
```
agriyakhetarpal, line 1: This one is mentioned by CodeCov as a line we have missing coverage for, so I think we can remove the line.

```suggestion
```

agriyakhetarpal (commented):
Thanks, @arjxn-py! I think this looks all good to me, unless there is something more that you plan to do here and push further changes.

arjxn-py:
Planned to add some directives to contribution guide as well but I believe it should be fine without that, we just have to little bit more careful on the PRs modifying method & attribute names or adding new of them.

arjxn-py:
Feel free to merge it after approval.

agriyakhetarpal (approved):
Looks fine to me, @arjxn-py! I'll request another review from @Saransh-cpp, we can merge soon after.

kratman:
Is this something that needs to be updated frequently?

agriyakhetarpal:
No, as long as we're not changing the API frequently. It does need to be updated when adding new features, though.

The goal will be that once we have a stable public-private API demarcation, `__all__` will reflect the public API.

arjxn-py:
Just a gentle ping to @Saransh-cpp to review this and go forward with it if looks fine, Thanks :)

Saransh-cpp (approved):
Thanks, @arjxn-py! Looks good to me!

PR #3624, opened 2023-12-16
Rename "positive electrode diffusivity" to "positive particle diffusivity"

julian-evers:
# Description

The "electrode diffusivity" parameter have been renamed to "particle diffusivity". parameter_values["Positive electrode diffusivity"] returns "Positive particle diffusivity" with a deprecation warning.

Fixes https://github.com/pybamm-team/PyBaMM/issues/3229

## Type of change

Please add a line in the relevant section of [CHANGELOG.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CHANGELOG.md) to document the change (include PR #) - note reverse order of PR #s. If necessary, also add to the list of breaking changes.

- [ x] New feature (non-breaking change which adds functionality)
- [ ] Optimization (back-end change that speeds up the code)
- [ ] Bug fix (non-breaking change which fixes an issue)

# Key checklist:

- [ x] No style issues: `$ pre-commit run` (or `$ nox -s pre-commit`) (see [CONTRIBUTING.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CONTRIBUTING.md#installing-and-using-pre-commit) for how to set this up to run automatically when committing locally, in just two lines of code)
- [ x] All tests pass: `$ python run-tests.py --all` (or `$ nox -s tests`)
- [ ] The documentation builds: `$ python run-tests.py --doctest` (or `$ nox -s doctests`)

You can run integration tests, unit tests, and doctests together at once, using `$ python run-tests.py --quick` (or `$ nox -s quick`).

## Further checks:

- [ ] Code is commented, particularly in hard-to-understand areas
- [ x] Tests added that prove fix is effective or that feature works

Issue #2028, opened 2022-04-20
Use `deprecation` package for deprecation warnings


brosaplanella:
Replace existing deprecation warnings/errors with [`deprecation` package](https://pypi.org/project/deprecation/#description) so tests fail after a few releases.

brosaplanella:
I have been looking into this and I have a (probably very basic) question. The `deprecation` package uses decorators and it seems designed to deprecate entire methods directly. How should I use them when we want to deprecate either a few lines of code, or some arguments? For an example, see l84 of [simulation.py](https://github.com/pybamm-team/PyBaMM/blob/develop/pybamm/simulation.py).

Saransh-cpp:
The [deprecat](https://github.com/mjhajharia/deprecat) package (built on top of the `deprecate` package) allows us to put deprecation warnings for arguments - https://github.com/mjhajharia/deprecat#deprecated-kwargs. Not sure how we can deprecate some particular lines of code.

agriyakhetarpal:
@brosaplanella I was looking through some solutions for this and while I couldn't find anything to deprecate some lines of code, there are indeed ways for deprecating arguments: in addition to [deprecat](https://github.com/mjhajharia/deprecat#deprecated-kwargs) mentioned by @Saransh-cpp, there's also [pyDeprecate](https://github.com/Borda/pyDeprecate) – a similar library with many more customisations available, applicable on both classes and methods along with forwarding inputs to newer arguments.  

Both of them will come at the cost of an extra dependency, so here is a purely Pythonic way which could perhaps be implemented as helper functions in `pybamm/doc_utils.py`? (see #2597). It transfers parameter values too but can be modified to just raise a warning on deprecation: https://gist.github.com/ZoomQuiet/f0113711181750e56ddfb51756d5c7fc

Saransh-cpp:
`pyDeprecate` looks really nice. We could also implement a new helper function, but having an extra dependency doesn't sound too bad to me (if it fulfills all our requirements).

RohitP2005:
hey @Saransh-cpp  i would like to work on it , if this issue isnt resolved yet .

Saransh-cpp:
HI @RohitP2005, please feel free to take this up. You might have to experiment a bit with the solutions/deprecation packages you find on the internet.

RohitP2005:
[...]

PR #4072, opened 2024-05-07
Fixes breaking changes from #3624

### LINKED ISSUE #4071: [Bug]: `Particle electrode diffusivity` rename breaks ParameterSet.update()
### PyBaMM Version

[Develop]

### Python Version

3.12

### Describe the bug

The parameter rename of `Positive electrode diffusivity` -> `Positive particle diffusivity` currently breaks the `ParameterSet.update()` method.

Linking https://github.com/pybamm-team/PyBaMM/pull/3624

### Steps to Reproduce

```python
import pybamm
parameter_values.update({"Positive electrode diffusivity [m2.s-1]": 1e-14})
```

### Relevant log output

```shell
{
	"name": "KeyError",
	"message": "\"Cannot update parameter 'Positive electrode diffusivity [m2.s-1]' as it does not have a default value. ('Positive electrode diffusivity [m2.s-1]' not found. Best matches are ['Positive particle diffusivity [m2.s-1]', 'Positive electrode conductivity [S.m-1]', 'Electrolyte diffusivity [m2.s-1]']). If you are sure you want to update this parameter, use param.update({{name: value}}, check_already_exists=False)\"",
	"stack": "---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
File ~/Documents/Git/PyBaMM/pybamm/util.py:59, in FuzzyDict.__getitem__(self, key)
     58 try:
---> 59     return super().__getitem__(key)
     60 except KeyError as error:

KeyError: 'Positive electrode diffusivity [m2.s-1]'

KeyError                                  Traceback (most recent call last)
File ~/Documents/Git/PyBaMM/pybamm/parameters/parameter_values.py:247, in ParameterValues.update(self, values, check_conflict, check_already_exists, path)
    246 try:
--> 247     self._dict_items[name]
    248 except KeyError as err:

File ~/Documents/Git/PyBaMM/pybamm/util.py:103, in FuzzyDict.__getitem__(self, key)
    100         raise KeyError(
    101             f\"'{key}' not found. Use the dimensional version '{k}' instead.\"
    102         ) from error
--> 103 raise KeyError(
    104     f\"'{key}' not found. Best matches are {best_matches}\"
    105 ) from error

KeyError: \"'Positive electrode diffusivity [m2.s-1]' not found. Best matches are ['Positive particle diffusivity [m2.s-1]', 'Positive electrode conductivity [S.m-1]', 'Electrolyte diffusivity [m2.s-1]']\"

KeyError                                  Traceback (most recent call last)
Cell In[4], line 2
      1 parameter_values = pybamm.ParameterValues(\"Chen2020\")
----> 2 parameter_values.update({\"Positive electrode diffusivity [m2.s-1]\": 1e-14})

File ~/Documents/Git/PyBaMM/pybamm/parameters/parameter_values.py:249, in ParameterValues.update(self, values, check_conflict, check_already_exists, path)
    247         self._dict_items[name]
    248     except KeyError as err:
--> 249         raise KeyError(
    250             f\"Cannot update parameter '{name}' as it does not \"
    251             + f\"have a default value. ({err.args[0]}). If you are \"
    252             + \"sure you want to update this parameter, use \"
    253             + \"param.update({{name: value}}, check_already_exists=False)\"
    254         ) from err
    255 # if no conflicts, update
    256 if isinstance(value, str):

KeyError: \"Cannot update parameter 'Positive electrode diffusivity [m2.s-1]' as it does not have a default value. ('Positive electrode diffusivity [m2.s-1]' not found. Best matches are ['Positive particle diffusivity [m2.s-1]', 'Positive electrode conductivity [S.m-1]', 'Electrolyte diffusivity [m2.s-1]']). If you are sure you want to update this parameter, use param.update({{name: value}}, check_already_exists=False)\""
}
```

BradyPlanden:
# Description

This PR updates `ParameterSet.update()` and `Solver.solve()` so that #3624 is non-breaking.

Fixes #4071 

## Type of change

# Key checklist:

## Further checks:


kratman:
Out of curiosity, if we did the renaming of the parameter, where is the old name coming from in PyBaMM usage?

In #4071 it looks like you manually passed the old name into the function. Is there another tool in the PyBaMM ecosystem that is still using the old name?

BradyPlanden:
Fair question. The previous name would be used in any script/notebook/package built before v24.5, and in those cases, if `ParameterSet.update()` or `Solver.solve(inputs={old_name:value})` is called, the change would be breaking. #3624 both renames the parameter and does the conversion in the [FuzzyDict](https://github.com/pybamm-team/PyBaMM/blob/fadf094d17d0562319e4d5065a2eee1a9848b004/pybamm/util.py#L60), but misses the above methods. This PR fixes that.

kratman:
I know we have deprecation warnings for the parameter name changes, I think if anything we should update them to catch it better. Putting it is the base solver seems like the wrong approach. Once we are out of the deprecation window then we would expect to be able to remove the code checking for it.

BradyPlanden:
> Putting it is the base solver seems like the wrong approach. Once we are out of the deprecation window then we would expect to be able to remove the code checking for it.

I agree it's not a very clean solution, but given the `inputs` arg is a user defined dictionary and directly passed from user-level I don't believe we can catch it before that point. Unless you see an alternative?

How do we track removal of depreciation warnings at the moment? Perhaps adding depreciation warnings around the additions in this PR and adding them to a log for removal after two release would ensure they get removed when ready.

brosaplanella:
We discussed about deprecation warnings in https://github.com/pybamm-team/PyBaMM/issues/2028 a while ago, but not much progress has been made.

valentinsulzer:
Can you share the scripts that are broken on `develop` and work with this fix?

kratman:
Summary of my comments from the meeting: "Can the check for the old parameter name be done only in the update function and add a warning"

BradyPlanden:
> Can you share the scripts that are broken on develop and work with this fix?

Sure @valentinsulzer, [any of these](https://github.com/pybamm-team/PyBaMM/tree/develop/docs/source/examples/notebooks/parameterization) with the old parameter name used in `parameter_values.update()` or the inputs object for `solve()` will break in develop and be fixed by this PR. It's any time `ParameterSet.update({name:value})` or `solver.solve(input={name:value})` uses the previous parameter name, both of which are used throughout the examples.

> "Can the check for the old parameter name be done only in the update function and add a warning"
Yep, I'll sort this and drop a line once it's ready for review.

valentinsulzer:
Nice preview of how painful it is to rename a parameter, for you guys leading the parameter refactor effort

brosaplanella (approved):
Looks good to me, thanks! It would be good to get a second review, just for a second opinion on how to handle the warnings. BTW, could you add a line to the CHANGELOG?

BradyPlanden:
There are a few tests that are failing periodically at the moment, but it seems to be across the repo. For coverage, the lines in question are being missed periodically, but not as a function of this PR. This should be good for the final review now.

PR #4092, opened 2024-05-15
Use wider-compatible LLVM-OpenMP, bundle `libgfortran` dylibs for macOS wheels

### LINKED ISSUE #4091: Fortnightly build for wheels failed
The build is failing with the following logs - https://github.com/pybamm-team/PyBaMM/actions/runs/9093914939

agriyakhetarpal:
# Description

Closes #4091, binaries on macOS need to be built with OpenMP which is more compatible with older macOS versions, rather than Homebrew OpenMP, which is not built for use cases where one needs to set MACOSX_DEPLOYMENT_TARGET to a lower value than the version specifier of the host macOS machine, e.g., at the time of building wheels with `cibuildwheel`.

This change is required due to stricter requirements across macOS version compatibility when repairing the wheel with tools like `delocate`.

## Type of change

# Key checklist:

## Further checks:


agriyakhetarpal:
Looks like https://github.com/matthew-brett/delocate/issues/211 and https://github.com/matthew-brett/delocate/pull/212 are some related threads.

https://github.com/agriyakhetarpal/PyBaMM/actions/runs/9096310836/job/25001433738 has some progress because the SuiteSparse and SUNDIALS dylibs are now exported correctly, but the build is still failing because we use gfortran and OpenMP headers from Homebrew, and Homebrew binaries are not necessarily good to publish Python wheels with. It is not recommended to downgrade the macOS runner version, since they shall be phased out soon... I'll look further into how we can replace these dependencies with other alternatives.

agriyakhetarpal:
Another solution is to set them to 12.0 and 14.0 as requested – however, this would break compatibility for a big proportion of users

agriyakhetarpal:
Yet another option: set up C and C++ compilers with the LLVM-OpenMP variant from https://scikit-learn.org/dev/developers/advanced_installation.html#macos-compilers-from-conda-forge rather than using Homebrew

Point of breakage is https://github.com/pypa/cibuildwheel/pull/1766, but this was just a documentation update, so our builds might have been incompatible silently: https://github.com/matthew-brett/delocate/pull/198 is where all dylibs are being checked for compatibility, and https://github.com/matthew-brett/delocate/issues/214 reveals that this was previously undetected across projects, so in any case we will need to migrate to using `gcc` which provides `libgomp` headers or use a different installation of `libomp` (LLVM variant) that has greater compatibility across macOS versions.

agriyakhetarpal:
See also: https://github.com/Homebrew/legacy-homebrew/issues/48437

agriyakhetarpal:
I was able to handle OpenMP, the other remaining task is to handle `gfortran` dynamic libraries. We are still getting them from Homebrew, so they have to either be replaced with a different installation. Using gfortran and gcc as compilers doesn't bring issues, but we have a tricky dependency from `suitesparseconfig.dylib` on `libgfortran.dylib` which is why the latter gets bundled into our wheels, and therefore breaks compatibility... A way out could be to load at runtime instead and excluding it from the wheels, but this should be used as a last resort. The `macos-11` images have been deprecated and will be removed in Q2 2024 (this quarter), so we can't use those, and we will need to migrate from `macos-12` to `macos-13` sometime soon.

agriyakhetarpal:
Almost all the way, here: https://github.com/agriyakhetarpal/PyBaMM/actions/runs/9116260227/job/25064471991

CasADi wheels are not repaired properly and it has a higher macOS minimum version requirement than expected – `libcasadi.3.7.dylib` requires 11.1, so I have opened an issue on their repository: https://github.com/casadi/casadi/issues/3698

I'll simply rename the wheels to get the builds passing, since we cannot do much about it at this time.

agriyakhetarpal:
Wheel builds are finally passing (🎉): https://github.com/agriyakhetarpal/PyBaMM/actions/runs/9117439816

I can open this up for review now, cc: @kratman and @martinjrobins. The Codacy failures are irrelevant for the most part since we need to call a few subprocesses, I don't seem to remember how one can ignore those.

.github/workflows/publish_pypi.yml
```diff
@@ -134,11 +125,54 @@ jobs:
...
+            mkdir -p $HOME/.local/lib
+            lib_dir=$(pwd)/gfortran_installed/gfortran-darwin-x86_64-native/lib
+            for lib in libgfortran.5.dylib libgfortran.dylib libquadmath.0.dylib libquadmath.dylib libgcc_s.1.dylib libgcc_s.1.1.dylib; do
+              cp $lib_dir/$lib $HOME/.local/lib/
+              install_name_tool -id $HOME/.local/lib/$lib $HOME/.local/lib/$lib
+              # can be removed I think
+              # install_name_tool -change $lib_dir/$lib $HOME/.local/lib/$lib $HOME/.local/lib/libgfortran.5.dylib
+              # install_name_tool -change $lib_dir/$lib $HOME/.local/lib/$lib $HOME/.local/lib/libgfortran.dylib
+              # install_name_tool -change $lib_dir/$lib $HOME/.local/lib/$lib $HOME/.local/lib/libquadmath.0.dylib
+              # install_name_tool -change $lib_dir/$lib $HOME/.local/lib/$lib $HOME/.local/lib/libquadmath.dylib
```
agriyakhetarpal, line 168:
```suggestion
```

.github/workflows/publish_pypi.yml
```diff
@@ -134,11 +125,54 @@ jobs:
...
 
       - name: Build wheels on macOS amd64
-        if: matrix.os == 'macos-12'
+        if: matrix.os == 'macos-13'
         run: pipx run cibuildwheel --output-dir wheelhouse
         env:
-          CIBW_BEFORE_BUILD_MACOS: >
-            python -m pip install --upgrade cmake casadi setuptools wheel
+          MACOSX_DEPLOYMENT_TARGET: 11.0
+          # Sourced from
```
agriyakhetarpal, line 132:
```suggestion
          # Sourced from
          # License: BSD-3-Clause
```

.github/workflows/publish_pypi.yml
```diff
@@ -162,18 +196,56 @@ jobs:
...
       - name: Build wheels on macOS arm64
-        run: python -m pipx run cibuildwheel --output-dir wheelhouse
+        run: |
+          python -m pip install cibuildwheel
+          python -m cibuildwheel --output-dir wheelhouse
         env:
-          CIBW_BEFORE_BUILD: python -m pip install cmake casadi setuptools wheel delocate
-          CIBW_REPAIR_WHEEL_COMMAND: delocate-listdeps {wheel} && delocate-wheel -v -w {dest_dir} {wheel}
+          MACOSX_DEPLOYMENT_TARGET: 11.0
+          # Sourced from
```
agriyakhetarpal, line 205:
```suggestion
          # Sourced from
          # License: BSD-3-Clause
```

arjxn-py (commented):
@agriyakhetarpal, is there anything still left in this one? Looks pretty good to me. 
Let know if you still plan to do something in this PR

agriyakhetarpal:
I don't think there should be anything else. I checked the `--force` and `--install-dir` posargs combinations for `pybamm-requires` as well locally – everything seems to install in the right place and work as desired. Until XCode's Clang adds support for the OpenMP specification someday in a later macOS version, this piece of code can stay for now. I would note that the same bundling process for `libgfortran` is present for the NumPy wheels as well, besides SciPy.

@arjxn-py,  you can go ahead with your approval and merge this. If we receive any complaints from macOS arm64 users about the IDAKLU solver, we'll just ask them to upgrade from macOS X 1A.B to 11.1 or later, that should fix their issues (because the CasADi issue above will probably remain open for a while).

cc: @BradyPlanden – this fixes the issue that was being discussed at the PyBOP 16/05/2024 developer meeting :)

agriyakhetarpal:
Also, cc: @cringeyburger @prady0t this PR, while this should be okay to merge for now, is one of the reasons why we should be working towards a mechanism for testing the built wheels this year ;) currently we just check if the IDAKLU solver can be imported correctly, but it can hide issues related to the solver at runtime—see the older issue #3783 for example—running the full test suite against the wheels (or a fraction of it) would be great

arjxn-py (approved):
Thanks @agriyakhetarpal 🎉

PR #4098, opened 2024-05-16
Added pooch registries for data files

### LINKED ISSUE #4026: Decide how to bundle `.pickle` files from COMSOL solution results data
`.pickle` files are binary files and not inherently secure. They therefore should be best included elsewhere in PyBaMM's packaging infrastructure when publishing to common distribution channels like PyPI owing to security reasons in the case where these binaries are modified with malicious intent and make it to the `develop` branch or a PyPI release where PyBaMM can be downloaded from (binary or non-binary). This means they should neither be included in the source distribution (controlled by `MANIFEST.in`), nor in the wheels (in the case where the wheel is built from the sdist). Binary platform-specific wheels for PyBaMM cannot be built from sdists right now and have to be built in-tree (#3651, #3564), and require restructuring the build system configuration at least, if not rewriting what files are copied into the package structure – since `pybamm/input/comsol_results/` is a namespace package (via `setuptools.command.install_data`).

We have the following options:

1. Store in JSON, CSV, or TXT file formats (non-binary) instead of `.pickle`
2. Use [`pooch`](https://learn.scientific-python.org/development/patterns/data-files/) to reproducibly download these files when needed and adjust PyBaMM's public API plus example notebooks for this, while moving these files to a separate repository in the `pybamm-team` organisation, and marking the repository as an archive to prevent tampering.
3. Use non-binary storage solutions from `xarray` or `zarr` because they can handle multi-dimensional data

Out of these, point 2 requires an internet connection and those wishing to load COMSOL results from these files would need access to the internet to download them, and therefore should be the last option to consider.

<hr>

## Task list, as of 10/05/2024

There are two major things to be done:

- [x] Convert pickle files to non-binary files (done in #4064)
- [ ] Use https://github.com/fatiando/pooch as an optional (`[dev]`) dependency to fetch these files plus adjust for the change here
    - [x] Set up a new repository: https://github.com/pybamm-team/pybamm-data
        - [ ] Add a README and some documentation
        - [ ] Copy the files and delete them from the PyBaMM main (this) repository (and additionally store checksums in a text file somewhere), and publish a GitHub Release there to store them perhaps (shorter URL)
    - [ ] Refer to the Scientific Python guide: https://learn.scientific-python.org/development/patterns/data-files/#downloading-larger-files-on-demand
    - [ ] Use `pytest`'s conditional markers or regular markers to skip tests where access to the internet is not available (adjust `noxfile.py`, test-related files, etc.)
    - [ ] Ensure files are not copied into the sdist (control `MANIFEST.in` and `setuptools`-specific package data)
    - [ ] Unsure about this point – maybe provide a public API to fetch these files (with the URLs kept hardcoded)?

santacodes:
# Description

- Added pooch as a [dev] dependency and added a file to create and store data file registries for [pybamm-data](https://github.com/pybamm-team/pybamm-data). 
- Added dataloader method to fetch data files from [pybamm-data](https://github.com/pybamm-team/pybamm-data) and store it under local cache folder and import it in notebooks wherever needed with specified paths.
- Removed data files from local repo under `pybamm/input`
- Data files under `pybamm/input/parameters` have not been removed or included in the pooch registries.
- Added documentation, tests and an example notebook for `pybamm.DataLoader` class. 

Fixes #4026 

## Type of change

# Key checklist:

## Further checks:


pybamm/pybammdata.py
agriyakhetarpal, line 1: Could you rename the file to `pybamm_data.py`?

pybamm/pybammdata.py
```diff
@@ -0,0 +1,48 @@
+import pooch
+
+
+class DataLoader:
+    def __init__(self):
```
agriyakhetarpal, line 5: This class should include a docstring with Google-style docstring parameters to explain what it is about and a usage example or two. You should add the associated `.rst` files as well for it to render on the Sphinx documentation

pybamm/pybammdata.py
```diff
@@ -0,0 +1,48 @@
...
+        self.path = pooch.os_cache("pybamm")
+        self.odie = pooch.create(
+            path=self.path,
+            base_url="https://github.com/pybamm-team/pybamm-data/releases/download/v1.0.0",
+            registry={
```
agriyakhetarpal, line 10: Could you add a comment that the data under `pybamm/input/parameters/` has not been included at this time?

pybamm/pybammdata.py
```diff
@@ -0,0 +1,48 @@
...
+                "stp_2C.txt": "sha256:6fe73b3a18e5fcfb95151dfd7d34c3cbe929792631447ed3ec88c047c9778223",
+                # Drive cycles
+                "UDDS.csv": "sha256:9fe6558c17aad3cc08109186923aeb7459cd3097a381c44e854bf22dd12a5a2a",
+                "US06.csv": "sha256:5909eb2ec7983fae86a050ff3b35a2041d0ab698710a6b0f95d5816e348077ba",
+                "WLTC.csv": "sha256:bb2f95018a44ac1425cb9c787c34721192af502c7385f1358f28e4f75df11fd8",
+                "car_current.csv": "sha256:4305b91b9df073cb048c25dd3fae725e06a94fe200e322e5c08db290d6799e36",
+            },
+        )
+
+    def getdata(self, filename: str):
```
agriyakhetarpal, line 43:
```suggestion
    def get_data(self, filename: str):
```

pybamm/pybammdata.py
```diff
@@ -0,0 +1,48 @@
...
+        try:
+            self.odie.fetch(filename)
+            return f"{self.path}/{filename}"
```
agriyakhetarpal, line 46: This will most likely fail on Windows where the path separator is `\` rather than `/`. Could you construct and return a `Path` object instead (or otherwise, use `os.normpath` on this string)?

pybamm/pybammdata.py
```diff
@@ -0,0 +1,48 @@
...
+        except NameError:
+            raise NameError(f"Unable to find {filename} in the registry") from None
```
agriyakhetarpal, line 48: These lines need a test. A few simple test cases with dummy file names should work plus a few cases with actual files. You would need to add a marker to check if internet access is available, and skip the tests if not running under a network.

pyproject.toml
```diff
@@ -110,6 +110,8 @@ dev = [
     "nbmake",
     # To access the metadata for python packages
     "importlib-metadata; python_version < '3.10'",
+    # To fetch data from pybamm-data
+    "pooch",
```
agriyakhetarpal, line 114: Could you make this a required dependency? Also, could you update the installation guide in the documentation about this – we have a table there that lists all of our dependencies and their minimum versions. It might also be good to add a minimum version here (one that is available for Python 3.9+ on both conda-forge and PyPI).

pybamm/pybammdata.py
santacodes: sure!

pybamm/pybammdata.py
santacodes: added this to the PR description

pybamm/pybammdata.py
santacodes: `self.path` refers to the `pooch.os_cache` method which returns a path object, so I think it should be good.

agriyakhetarpal: No, I meant the filename after `self.path`, which is still separated with `/`. This should be normalised. I would suggest using `pathlib`'s `Path` object, you can use `/` that way.

santacodes: resolved in the latest commit

pybamm/pybammdata.py
santacodes: I have added test cases with one dummy file and one randomly selected files from the registry

pybamm/pybamm_data.py
```diff
@@ -0,0 +1,137 @@
...
+        -------
+        POSIX/Windows Path to the data file.
+        """
+        try:
+            self.registry.fetch(filename)
+            return pathlib.Path(f"{self.path}/{self.version}/{filename}")
+        except NameError:
+            raise NameError(f"Unable to find {filename} in the registry") from None
+
+    def show_registry(self, checksum: bool = False):
```
agriyakhetarpal, line 123:
```suggestion


    def show_registry(self, checksum: bool = False):
```

(style guide): it is generally better to leave out two newlines between methods.

pybamm/pybamm_data.py
```diff
@@ -0,0 +1,137 @@
...
+            "car_current.csv": "sha256:4305b91b9df073cb048c25dd3fae725e06a94fe200e322e5c08db290d6799e36",
+        }
+        self.registry = pooch.create(
+            path=self.path,
+            base_url=f"https://github.com/pybamm-team/pybamm-data/releases/download/{self.version}",
+            version=self.version,
+            registry=self.files,
+        )
+
+    def get_data(self, filename: str):
```
agriyakhetarpal, line 105:
```suggestion


    def get_data(self, filename: str):
```

pybamm/pybamm_data.py
```diff
@@ -0,0 +1,137 @@
...
+    registry: pooch.core.Pooch object
+        Creates a registry of data files available to fetch using pooch.
+    """
+
+    def __init__(self):
+        """
+        Create a pooch registry with the following data files available upstream at https://github.com/pybamm-team/pybamm-data/
+
+        COMSOL RESULTS
+        ---------------
```
agriyakhetarpal, line 29:
```suggestion
        COMSOL Results
        ---------------
```

pybamm/pybamm_data.py
```diff
@@ -0,0 +1,137 @@
...
+        comsol_05C.json
+        comsol_1C.json
+        comsol_1plus1D_3C.json
+        comsol_2C.json
+        comsol_3C.json
+
+        Kokam SLPB 75106100 discharge data from Ecker et al (2015)
+        ----------------------------------------------------------
+        Ecker_1C.csv
+        Ecker_5C.csv
```
agriyakhetarpal, line 40: Could you add the relevant papers here (and similarly for other data files under the other headings where applicable)? They should be present in the `CITATIONS.bib` file. If they are not, you can get the BibTeX reference for them from Google Scholar or similar and add them to the file. Then, you can reference them in the docstring with something like

```rest

:footcite:t:`AuthorYear`

```

pybamm/pybamm_data.py
```diff
@@ -0,0 +1,137 @@
...
+    ----------
+    path: POSIX/Windows PATH
+       POSIX/Windows path to the data file.
+       The folder locations are defined by the `platformdirs <https://platformdirs.readthedocs.io/en/latest/>`_ package using the `user_cache_dir <https://platformdirs.readthedocs.io/en/latest/api.html#cache-directory>`_ function. The default cache paths for pybamm are located under -
+
+        - Mac: ``~/Library/Caches/v1.x.x/pybamm``
+
+        - Unix: ``~/.cache/v1.x.x/pybamm`` or the value of the XDG_CACHE_HOME environment variable, if defined.
+
+        - Windows: ``C:\\Users\\<user>\\AppData\\Local\\pybamm\\v1.x.x\\Cache``
```
agriyakhetarpal, line 19:
```suggestion
       The folder locations are defined by the `platformdirs <https://platformdirs.readthedocs.io/en/latest/>`_ package using the `user_cache_dir <https://platformdirs.readthedocs.io/en/latest/api.html#cache-directory>`_ function.
```

I would suggest skipping these, it directs too much emphasis on users who do not have to deal with this information when using the `DataLoader` class, and therefore we can keep this abstracted away from them.

agriyakhetarpal (commented):
Thanks for this, @santacodes! Could you fix the coverage by adding additional tests for the `get_data` method (to check the `NameError` that you raised), and for the `checksum` option in the `show_registry` method? @valentinsulzer, could you review this as well?

Besides this, I will recommend building the source distribution and the wheels on your fork by triggering the manual job we have available, and locally using `python -m build` to ensure that we aren't missing out on any files and that the sizes of the distributions are alright.

pybamm/pybamm_data.py
santacodes: Sure!

tests/unit/test_parameters/test_current_functions.py
```diff
@@ -21,8 +20,9 @@ def test_constant_current(self):
 
     def test_get_current_data(self):
         # test process parameters
+        data_loader = pybamm.DataLoader()
         current_data = pd.read_csv(
-            os.path.join(pybamm.__path__[0], "input", "drive_cycles", "US06.csv"),
+            data_loader.get_data("US06.csv"),
```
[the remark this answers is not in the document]
santacodes: Yeah oops I forgot to add those marks in other tests

tests/unit/test_pybamm_data.py
```diff
@@ -0,0 +1,36 @@
+import pybamm
+import pytest
+import random
+import requests
```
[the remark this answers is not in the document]
santacodes: I think it is a required dependency of pooch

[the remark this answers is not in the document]
santacodes: https://www.fatiando.org/pooch/latest/install.html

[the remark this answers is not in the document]
agriyakhetarpal: Makes sense. I am not sure if we should include it as well – explicit is better than implicit so I would say we should, but happy to hear from others as well

santacodes:
Ran the Build and publish package CI to build whls on my branch, and the sizes have been reduced by almost 15-20MB for POSIX and Windows distributions - https://github.com/santacodes/PyBaMM/actions/runs/9162618687.

tests/unit/test_pybamm_data.py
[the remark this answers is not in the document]
agriyakhetarpal: Here, I would say we should use something from the standard library, like

```python
import socket

host = socket.gethostbyname("www.github.com")
conn.close()
```

inside a try-except block rather than importing a transitive dependency like `requests` that we do not explicitly mark as ours. This should be reasonable and diverts the question altogether, plus it helps us in case `pooch` ever decides to drop `requests` and aims to download files using plain-old `urllib` (not that it seems it will, though).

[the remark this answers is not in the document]
santacodes: Sure that would work, I'll do that then!

pybamm/pybamm_data.py
```diff
@@ -0,0 +1,166 @@
...
+        Returns
+        -------
+        POSIX/Windows Path to the data file.
+        """
+        self.registry.fetch(filename)
+        return pathlib.Path(f"{self.path}/{self.version}/{filename}")
+
+    def show_registry(self, checksum: bool = False):
+        """
+        Prints all the files present in the registry along with their checksums (optional).
```
agriyakhetarpal, line 149:
```suggestion
        Prints the names all the files present in the registry along with their checksums (optional).
```

pybamm/pybamm_data.py
```diff
@@ -0,0 +1,166 @@
...
+        """
+        Fetches the data file from upstream and stores it in the local cache directory under pybamm directory.
+
+        Parameters
+        ----------
+        filename : str
+            Name of the data file to be fetched from the registry.
+        Returns
+        -------
+        POSIX/Windows Path to the data file.
```
agriyakhetarpal, line 142:
```suggestion

        Returns
        -------
        POSIX/Windows Path to the data file.
```

Also, could you add a proper return type for this?

pybamm/pybamm_data.py
```diff
@@ -0,0 +1,166 @@
...
+        Prints all the files present in the registry along with their checksums (optional).
+
+        Parameters
+        ----------
+        checksum : bool, optional
+            Prints file registries with checksums if True.
+        Returns
+        -------
+        list, dict (if checksum = True)
+        """
```
agriyakhetarpal, line 158: I am not sure how I feel about a method that returns different data types based on the input argument, do we have a better way to do this?

docs/source/user_guide/installation/index.rst
```diff
@@ -69,6 +69,7 @@ Package                                                             Minimum supp
 `SymPy <https://docs.sympy.org/latest/index.html>`__                1.9.3
 `typing-extensions <https://pypi.org/project/typing-extensions/>`__ 4.10.0
 `pandas <https://pypi.org/project/pandas/>`__                       1.5.0
+`pooch <https://www.fatiando.org/pooch/>`__       1.8.1
```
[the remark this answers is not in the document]
santacodes: sure I fiexed it, I don't remember indenting it that way maybe the linter did it.

pybamm/pybamm_data.py
santacodes: I don't feel there might be a practical application for the return data types, it usually just prints the the registry which is the intention of that method and returns what it prints (to assert it in pytest unit tests).

pybamm/pybamm_data.py
```diff
@@ -0,0 +1,166 @@
...
+    """
+    Data Loader class for downloading and loading data files upstream at https://github.com/pybamm-team/pybamm-data/
+
+    Attributes
+    ----------
+    path: POSIX/Windows PATH
+       POSIX/Windows path to the data file.
+       The folder locations are defined by the `platformdirs <https://platformdirs.readthedocs.io/en/latest/>`_ package using the `user_cache_dir <https://platformdirs.readthedocs.io/en/latest/api.html#cache-directory>`_ function.
+    registry: pooch.core.Pooch object
+        Creates a registry of data files available to fetch using pooch.
```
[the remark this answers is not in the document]
santacodes: I think we might be well off removing the specifications when they are already mentioned in the `get_data()` method, and creating an `intersphinx` mapping for pooch and `platfordirs`. Will add it in the latest commit.

pybamm/pybamm_data.py
agriyakhetarpal: We could create a separate function `show_registry_by_checksum()` that returns a dictionary, while this one can return a list. I agree that there is not much of a practical application for this and I don't think end users need to view the checksums for any of the files (because they have been present inside the PyBaMM repository before this PR).

Another option could be to implement something like `get_data("US06.csv").checksum` that can the checksum for an individual file as a `str`, and move away from this one.

santacodes: I decided to remove the checksum parameter overall as it doesn't have any practical application, rather if a user wants to check the checksum of a particular file they could access the `files` attribute of the `data_loader` class which would give them a `dict` of all files and their checksums. I also updated its execution in the example notebook.

agriyakhetarpal (approved):
Looks good to me now, thanks, @santacodes! I'll wait for some additional thoughts on this before we merge.

cc @cringeyburger @prady0t – this PR has removed (most of) the PyBaMM data files from the PyBaMM source code. Testing the built wheels using `CIBW_TEST_COMMAND` should now be a bit easier. I feel that the `pytest --pyargs pybamm` invocation would be better than a `pybamm.test()` one – that way we would not need to move the `tests/` directory inside `pybamm/`

pyproject.toml
```diff
@@ -42,7 +42,8 @@ dependencies = [
     "anytree>=2.8.0",
     "sympy>=1.12",
     "typing-extensions>=4.10.0",
-    "pandas>=1.5.0"
+    "pandas>=1.5.0",
+    "pooch>=1.8.1",
```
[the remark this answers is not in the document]
arjxn-py: No worries on keeping it as an optional dependency too, but can you specify which previously existent file? Feel free to resolve this conversation further.

docs/source/examples/notebooks/models/Validating_mechanical_models_Enertech_DFN.ipynb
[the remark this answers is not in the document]
arjxn-py: > this serves as a personal reminder to switch to MyST-NB

+1 for this as far as it is not a major breaking change, feel free to open an issue for the same or let me know to open it.

arjxn-py (approved):
I've no further concerns with this one, good work @santacodes.
@agriyakhetarpal are we good to merge this or we still seek +1 review?

pyproject.toml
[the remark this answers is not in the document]
agriyakhetarpal: Ah, for example, earlier to this PR one could do

```
drive_cycle = pd.read_csv(
    "pybamm/input/drive_cycles/US06.csv", comment="#", header=None
).to_numpy()
```

in a drive cycle simulation to retrieve a NumPy array from the CSV-converted `DataFrame` object – but, after this PR, one needs to use `data_loader.get_data("US06.csv")` to access it. We have added an entry about this in the CHANGELOG, which is fine, but we would also like people to receive functionalities to work with these files without having to install any additional optional dependencies such as `pybamm[data]` after they try to upgrade their workflows to v24.5. 

(keeping it optional would lead to another error about `pooch` not being installed, which could be disorienting to unsuspecting users who are not reading the CHANGELOG as often)

agriyakhetarpal:
I think this is good to go on my end, but I just had a question about skipping tests without networks – @santacodes, the `pytest` decorator + `no_internet_connection()` we have will prevent the tests from failing, but this is just for the unit and the integration tests. Is there a way we can cleanly extend this behaviour for the example scripts and notebooks which make use of the data files, in specific? It's not a big deal though, because we expect that most people have internet access nowadays somehow or the other.

santacodes:
> I think this is good to go on my end, but I just had a question about skipping tests without networks – @santacodes, the `pytest` decorator + `no_internet_connection()` we have will prevent the tests from failing, but this is just for the unit and the integration tests. Is there a way we can cleanly extend this behaviour for the example scripts and notebooks which make use of the data files, in specific? It's not a big deal though, because we expect that most people have internet access nowadays somehow or the other.

For notebooks and scripts, I am thinking we could add an error-handling method to raise an Error, pooch raises `socket.gaierror` by default when there is no network connection as it cannot create a socket connection with the upstream URL.
The second option would be to check for the internet connection before actually fetching data files from upstream and warning the user about network connectivity. 
I think both can be implemented, but I am not sure which would be more optimal for scripts and notebooks specifically.

agriyakhetarpal:
> For notebooks and scripts, I am thinking we could add an error-handling method to raise an Error, pooch raises `socket.gaierror` by default when there is no network connection as it cannot create a socket connection with the upstream URL. The second option would be to check for the internet connection before actually fetching data files from upstream and warning the user about network connectivity. I think both can be implemented, but I am not sure which would be more optimal for scripts and notebooks specifically.

I thought of that but both of these options should ideally stay outside and abstracted away from the content inside the notebooks and scripts, which isn't quite viable and seems over-engineered... I opine that we should skip them for now. We can help out on a per-user basis if someone tries to use `nox -s scripts` or `nox -s examples` without an internet connection. Another reason why I say this is because we do not include any of the example scripts or notebooks inside the sdist or the wheel, which means that people have to come to GitHub or browse the online documentation to access them anyway, and accessing GitHub requires access to an internet connection.

tests/unit/test_pybamm_data.py
```diff
@@ -0,0 +1,37 @@
...
+@pytest.mark.skipif(
+    no_internet_connection(),
+    reason="Network not available to download files from registry",
+)  # Skip if no internet
+def test_fetch():
+    # Fetch a file from the registry and check it in local cache folder for its presence
+    data_loader = pybamm.DataLoader()
+    random_file = random.choice(list(data_loader.files.keys()))
+    file_path = data_loader.get_data(random_file)
+    return file_path.is_file()
```
kratman, line 18: I agree with codacy here, random in tests is not great. Can we just check the first file instead?
```suggestion
def test_fetch():
    test_file = list(pybamm.DataLoader().files.keys())[0]
    return data_loader.get_data(test_file).is_file()
```

tests/unit/test_pybamm_data.py
```diff
@@ -0,0 +1,37 @@
+import pybamm
+import pytest
+import random
```
kratman, line 3: With the other change, this should be removed
```suggestion
```

kratman (changes requested):
Minor changes to the tests

kratman (commented):
For any other reviewers: I checked the link failure. It should start passing once this PR is merged. The link checker goes with the stable documentation, so it cannot see the new page

agriyakhetarpal:
Maybe we should refer to the reST-based heading for PyBaMM Data instead of adding an HTTP(S) link to our documentation? Otherwise, this failing link will keep failing for all the release candidates until we have the v24.5 release out – the link is pointing to `stable`, not `latest`. @santacodes, could you make this change in the notebook?

agriyakhetarpal:
Changing to `../api/pybamm_data.rst` won't help – could you add a heading under the PyBaMM Data page and use reST syntax to link to it? Something under [this SO post](https://stackoverflow.com/questions/18632781/how-to-make-an-internal-link-to-a-heading-in-sphinx-restructuredtext-without-cre) should help

kratman (changes requested):
Marking @agriyakhetarpal's last comment as a change request to keep this from being merged until it is fixed:

> Changing to `../api/pybamm_data.rst` won't help – could you add a heading under the PyBaMM Data page and use reST syntax to link to it? Something under [this SO post](https://stackoverflow.com/questions/18632781/how-to-make-an-internal-link-to-a-heading-in-sphinx-restructuredtext-without-cre) should help

santacodes:
> Changing to `../api/pybamm_data.rst` won't help – could you add a heading under the PyBaMM Data page and use reST syntax to link to it? Something under [this SO post](https://stackoverflow.com/questions/18632781/how-to-make-an-internal-link-to-a-heading-in-sphinx-restructuredtext-without-cre) should help

I tried referencing headers, but sphinx doesn't seem to render it as a link no matter what, even when I reference the `pybamm_data.rst` to reference the documentation header as they are not under the same directory. It only works by referencing the notebook's path like in my last commit, or by referencing `docs/source/api/pybamm_data.rst:#pybamm.DataLoader`. They both mean the same thing, so I will leave it up to you to decide further.

agriyakhetarpal:
I think this should be okay, then, thanks for the explanation. Not sure if there is a better way to link them unless we switch to MyST completely for the codebase, which is something for a later time

PR #4103, opened 2024-05-20
Updated scipy and jax versions to fix `linalg.tri` deprecation

### LINKED ISSUE #3959: JAX BDF solver tests failing / update `[jax]` versions (due to `scipy.linalg.tril` deprecation)
The JAX BDF solver tests are [failing](https://github.com/pybamm-team/PyBaMM/actions/runs/8535878801/job/23383249254?pr=3846) on all PRs (#3846, #3945, etc.) for Python 3.9 and later because SciPy removed some linear algebra routines in v1.13.0. The Python 3.8 tests are passing because SciPy has dropped support for it earlier

I'm guessing we need to bump the `jax` and `jaxlib` versions now or relax the pin in the requirements, because there have been quite many releases since `v0.4.20` – the current version available at the time of writing is `v0.4.25`.

### Checklist
- [x] Fix Jax test issues with updated versions (#3962)
- [ ] Unpin scipy version (#4103)

brosaplanella:
# Description

Updated jax & jaxlib to most recent version and scipy to >= 1.11.0 (the one that introduced the deprecation warning).

Fixes #3959

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ classifiers = [
 ]
 dependencies = [
     "numpy>=1.23.5",
-    "scipy>=1.9.3,<1.13.0",
+    "scipy>=1.11.0",
     "casadi>=3.6.5",
     "xarray>=2022.6.0",
     "anytree>=2.8.0",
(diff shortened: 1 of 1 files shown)


kratman:
@brosaplanella It looks like the remaining failures are about precision

brosaplanella:
Ok, I found the issue. Basically they have changed the solver for the interpolant and now it is an iterative one (see [Scipy docs](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.RegularGridInterpolator.html)). I have now defined the interpolant to use a direct solver (as before) so accuracy should remain. If needed, in the future, we can allow users to specify the solver when creating the interpolant.

kratman:
@brosaplanella If the only issue was precision before switching to the direct solver, then I would guess that the iterative solver would be fine. If they made that the default, then there is probably a reason for it

kratman (approved):
Assuming tests pass, then I am happy with this

PR #4105, opened 2024-05-20
Use `delvewheel` to repair wheels on Windows, add MSVCRT DLLs

agriyakhetarpal:
# Description

Adds what was missing to be done in #3783 – this should ensure that we get some extra non-UCRT DLLs in a `wheel.libs` folder inside the Windows wheel. This is not required but would be good to have because we use MSVC (Visual Studio 17 2022) as our compiler and not Clang/MinGW (which use UCRT) and therefore we should provide DLLs like `MSVCP140.dll` that have a tendency of not being present on many systems, in order to boost compatibility (this is mostly because of the absence of the MSVC Redistributable runtime components on quite a few systems). For most systems running Windows 10 and above, these files are generally present.

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/.github/workflows/publish_pypi.yml
+++ b/.github/workflows/publish_pypi.yml
@@ -83,7 +83,7 @@ jobs:
             CMAKE_GENERATOR_PLATFORM=x64
             CMAKE_BUILD_PARALLEL_LEVEL=${{ steps.get_num_cores.outputs.count }}
           CIBW_ARCHS: AMD64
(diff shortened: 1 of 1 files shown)


agriyakhetarpal:
Triggered a run at https://github.com/agriyakhetarpal/PyBaMM/actions/runs/9164324329/job/25195376665, it should pass most likely

arjxn-py (approved):
Thanks @agriyakhetarpal, coverage failure seems to be unrelated (i.e. related with #4103)

PR #4108, opened 2024-05-24
Running tests together on github-actions.

### LINKED ISSUE #4099: Run unit tests and integration tests together to speed up PR builds
## Description

The PR builds run unit tests and integration tests separately right now, which leads to a lot of runner requests from GitHub Actions (more than what are available to us concurrently with a non-team or non-enterprise plan), and as a result they usually take up to an hour to complete. A way to improve the PR build times would be to run `nox -s unit` and `nox -s integration` together till the time #3617 is finished and the integration tests are migrated to support `pytest` invocations, because the `unit` tests after #3857 run in parallel, and usually pass within ~5-6 minutes across platforms (and are much faster within the range of <3 minutes on larger runners such as the `macos-14` ones).

## Proposed implementation

Modify `test_on_push.yml` to run `nox -s unit` and `nox -s integration` in the same step, remove one (either) of the redundant unit or the integration jobs, and adjust the job names as necessary. 

## Additional context

Discussed in the community bonding GSoC meeting on 17/05/2024 with @Saransh-cpp and @arjxn-py – cc: @prady0t (you'll have to comment on the issue in order for us to assign you to this)

prady0t:
# Description

Running unit and integration tests together for speedup.

Fixes #4099


.github/workflows/test_on_push.yml
```diff
@@ -30,18 +30,14 @@ jobs:
...
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-12, windows-latest]
-        python-version: ["3.9", "3.10", "3.11", "3.12"]
-        # We check coverage on Ubuntu with Python 3.12, so we skip unit tests for it here
-        exclude:
-          - os: ubuntu-latest
-            python-version: "3.12"
+        os: [ ubuntu-latest, macos-12, windows-latest ]
```
agriyakhetarpal, line 39:
```suggestion
        os: [ubuntu-latest, macos-12, macos-14, windows-latest]
```
Besides this change, we can remove the `include:` section and the separate entries for macOS 14 below. We kept them because Python 3.8 and Python 3.9 were not available back then: https://github.com/actions/setup-python/issues/808

prady0t:
> I guess this duplicates the coverage job but we can look at this further here itself, could you please fix the conflicts first before I leave another review?

Using the exclude block would have resulted in integration tests also getting excluded. Can we also run coverage test together with this?

agriyakhetarpal:
> Using the exclude block would have resulted in integration tests also getting excluded. Can we also run coverage test together with this?

You can do something like

```yaml
      - name: Run unit/coverage and integration tests
        run: |
          if [ ${{ matrix.python-version }} == "3.12" && ${{ matrix.os }} == "ubuntu-latest" ]; then
            python -m nox -s coverage
          else
            python -m nox -s unit
          fi
          python -m nox -s integration
```

you might need to either add `shell: bash` to the step or adjust the syntax for `pwsh` on Windows (former is better)

prady0t:
> > Using the exclude block would have resulted in integration tests also getting excluded. Can we also run coverage test together with this?
> 
> You can do something like
> 
> ```yaml
>       - name: Run unit/coverage and integration tests
>         run: |
>           if [ ${{ matrix.python-version }} == "3.12" && ${{ matrix.os }} == "ubuntu-latest" ]; then
>             python -m nox -s coverage
>           else
>             python -m nox -s unit
>           fi
>           python -m nox -s integration
> ```
> 
> and then include the excluded Linux 3.12 job. You might need to either add `shell: bash` to the step or adjust the syntax for `pwsh` on Windows (former is better)

Wouldn't the above if else condition already do the job? Do we still need to add "exclude block" after this?

agriyakhetarpal:
> Wouldn't the above if else condition already do the job? Do we still need to add "exclude block" after this?

We're referring to the same thing, yes – "include the excluded block" would be the same as not excluding it. We can then move the coverage upload step to this job and run it with an `if:` condition.

prady0t:
If I also upload the coverage results in this same step here can we completely remove `check_coverage`?

.github/workflows/test_on_push.yml
```diff
@@ -109,144 +109,22 @@ jobs:
...
-        run: python -m nox -s pybamm-requires
-
-      - name: Run unit tests for Ubuntu with Python 3.11 and generate coverage report
-        run: python -m nox -s coverage
+          if [ ${{ matrix.python-version }} == "3.12" && ${{ matrix.os }} == "ubuntu-latest" ]; then
+            python -m nox -s coverage
+          else
+            python -m nox -s unit
+          fi
+          python -m nox -s integration
```
agriyakhetarpal, line 120: Maybe we can add the integration tests as a separate step after the unit/coverage tests? It would make it easier to view the logs since they would then be under a different heading/dropdown

.github/workflows/test_on_push.yml
```diff
@@ -109,144 +109,22 @@ jobs:
...
 
       - name: Upload coverage report
+        if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.12'
         uses: codecov/codecov-action@v4.4.1
```
agriyakhetarpal, line 124: This isn't running in any jobs right now because the Linux 3.12 job gets excluded. You shall have to remove the `exclude:` item from the matrix defined above

agriyakhetarpal (approved):
Thanks, @prady0t! Looks all good to me. The Windows integration tests look like they were on a slower runner, and we can't do much about this until we migrate all the integration tests to `pytest` and thereby make use of `pytest-xdist`. I'll merge this in a while if there are no objections.

For additional context for any other reviewers: we can receive 12 or so concurrent jobs from GitHub runners on a free plan, so requesting 40 or so jobs means we need to wait until some of them pass in order to run further requested jobs, which meant that we were not benefiting from the parallelisation between unit and integration tests. With this PR, the number of requested jobs is now 24 (including the style and the link checkers).

PR #4117, opened 2024-05-29
Running doctests using doctest-plus

### LINKED ISSUE #4031: Add support for `pytest-doctestplus`
**Description**

Now that #3857 is merged, we can add support for `xdoctest` for testing doctests.
[Xdoctest](https://xdoctest.readthedocs.io/en/latest/) is a pytest equivalent to python's build-in `doctest`. See more : https://github.com/pybamm-team/PyBaMM/issues/3617#issuecomment-1854702640 

Related to : #3617

> [!NOTE]
> It was decided on 27/05/2024 to switch to `pytest-doctestplus` rather than `xdoctest`.

prady0t:
# Description

Running doctests on `/pybamm` using pytest's doctest-plus plugin. We decided to move ahead with `doctestplus` as `xdoctest` was throwing random internal errors. See discussion [here](https://pybamm.slack.com/archives/C04ARM97HAQ/p1716694895908549).

Fixes #4031
Also addresses #3617


prady0t:
Still working on latexify error.

agriyakhetarpal:
I think it makes sense to skip the failing doctest. The functionality is being tested in the unit tests, and testing through a doctest whether a PNG or the LaTeX form of an equation can be printed out or not is quite inconvenient. Please see https://github.com/scientific-python/pytest-doctestplus#skipping-tests for instructions.

prady0t:
> I think it makes sense to skip the failing doctest. The functionality is being tested in the unit tests, and testing through a doctest whether a PNG or the LaTeX form of an equation can be printed out or not is quite inconvenient. Please see https://github.com/scientific-python/pytest-doctestplus#skipping-tests for instructions.

Makes sense. Skipping these tests.

run-tests.py
```diff
@@ -86,6 +86,15 @@ def run_doc_tests():
             ],
             check=True,
         )
+        print("Checking for doctests.")
+        subprocess.run(
+            [
+                "pytest",
+                "--doctest-plus",
```
agriyakhetarpal, line 93:
```suggestion
            [
                f"{sys.executable}",
                "-m"
                "pytest",
                "--doctest-plus",
```

run-tests.py
```diff
@@ -86,6 +86,15 @@ def run_doc_tests():
             ],
             check=True,
         )
+        print("Checking for doctests.")
```
agriyakhetarpal, line 89: Could you remove the subprocess above this and [configure `pytest-doctestplus` to run for docs-related files](https://github.com/scientific-python/pytest-doctestplus#setup-and-configuration) as well? Furthermore, if we get that done, we don't need the Sphinx version of it, so you can then remove `sphinx.ext.doctest` and `doctest_global_setup` from `docs/conf.py`.

run-tests.py
```diff
@@ -86,6 +86,15 @@ def run_doc_tests():
...
+        subprocess.run(
+            [
+                "pytest",
+                "--doctest-plus",
+                "pybamm",
+            ],
+            check=True,
+        )
     except subprocess.CalledProcessError as e:
```
agriyakhetarpal, line 98: This subprocess call can now be made a bit cleaner in general since there would be no files to remove if Sphinx is not run. We can wrap everything in a simple try-except block and remove the `finally` block.

pyproject.toml
```diff
@@ -103,6 +103,8 @@ dev = [
     "nox",
     # For coverage
     "pytest-cov",
+    # For doctest
+    "pytest-doctestplus",
```
agriyakhetarpal, line 107: Could you add an entry for this to the table in the installation guide? Here: https://docs.pybamm.org/en/latest/source/user_guide/installation/index.html#dev-dependencies

pybamm/models/full_battery_models/lithium_metal/dfn.py
```diff
@@ -2,7 +2,7 @@
 # Doyle-Fuller-Newman (DFN) Model
 #
 import pybamm
-from ..lithium_ion.dfn import DFN as LithiumIonDFN
+from pybamm.models.full_battery_models.lithium_ion.dfn import DFN as LithiumIonDFN
```
agriyakhetarpal, line 5: Doing this across the codebase would be much appreciated, though in a separate PR – thanks!

agriyakhetarpal: Yes, that is the plan. We had a discussion about this the other day in the `#infrastructure` channel on Slack, I shared [`mkinit`](https://pypi.org/project/mkinit/) for this to do this as a one-time thing, but additionally, a pre-commit hook that checks/converts relative imports would be great as well. 

I found https://github.com/MarcoGorelli/absolufy-imports from this [Reddit post](https://www.reddit.com/r/learnpython/comments/1d3n991/is_there_a_python_precommit_tool_that_can_change/), and while it's archived, it might still work

kratman: https://docs.astral.sh/ruff/rules/relative-imports/

kratman: Ruff has it, no new tools needed

agriyakhetarpal: Awesome, thanks!

run-tests.py
prady0t: Isn't the above subprocess for building documentation? How to test documentation builds if we remove this?

run-tests.py
prady0t: Again if we do not run Sphinx, how to test for documentation builds?

run-tests.py
agriyakhetarpal: > Isn't the above subprocess for building documentation?

The documentation is built using Sphinx's HTML builder (`-b html`) while this is the `doctest` builder (`-b doctest`). The doctest builder is enabled by the `doctest` extension, see https://www.sphinx-doc.org/en/master/usage/builders/index.html#builders

> How to test documentation builds if we remove this?

Good question. We can do that with `nox -s docs` instead, which runs separate commands in CI versus locally – we never added the HTML builder to `run-tests.py` I think and we always kept it in `noxfile.py` (and earlier in `tox.ini`). We will either completely remove `run-tests.py` or clear out major parts of it anyway, so reducing repetition will be nice in the coming weeks.

run-tests.py
agriyakhetarpal: Answered above, I hope that helps

run-tests.py
prady0t: This is really helpful! Thanks. However `--doctest-plus` cannot recognise tests in `docs`. Could it be because of [this](https://github.com/scientific-python/pytest-doctestplus#setup-and-configuration:~:text=Sphinx%20Doctest%20Directives) directive?

agriyakhetarpal: I don't think we have many (or any, AFAIR) doctests inside `docs/`. We have most of them in the docstrings for the public API classes and methods, and they are in the main `pybamm/` folder – as long as they are being tested and they pass, we should be good.

prady0t: Running pytest on docs Fails with exit code 5. Pytest treats `no test run` as [error](https://pytest.org/en/stable/reference/exit-codes.html#:~:text=Exit%20code%205,tests%20were%20collected). 
This stops running further tests.
We can maybe remove running tests on docs for now until we have any tests written. Is there a way to suppress this error? One hackish way could be to add a dummy test somewhere in the docs folder.

agriyakhetarpal: Why do we need to run `pytest` on the `docs/` folder, though?

prady0t: Aren't the docs-related files inside the `docs` folder? Am I missing something here?

agriyakhetarpal: Sorry, I somehow missed the context of the comment I posted originally. All the doctests are in `pybamm/`, inside the docstrings for the public API (marked with `.. doctest::`), so we don't need to run doctests across `docs/` – we don't have anything to test there. 

Therefore, from my earlier comment:

> and [configure pytest-doctestplus to run for docs-related files](https://github.com/scientific-python/pytest-doctestplus#setup-and-configuration) as well?

this isn't needed. We just need to remove `sphinx.ext.doctest` and `doctest_global_setup`, and the Sphinx subprocess above, and this comment can be resolved.

prady0t: This clears the confusion. Thanks!

pybamm/models/full_battery_models/lithium_metal/dfn.py
arjxn-py: Opened #4134 for this, can be further assigned and discuss there.
Marking this one as resolved for now.

pyproject.toml
prady0t: I've also modified other outdated entries as well.

agriyakhetarpal:
The doctests failed because of the now redundant `.. doctest::` directive that Sphinx won't recognise without `sphinx.ext.doctest`. The Windows build is unrelated, I triggered a re-run for it.

docs/source/user_guide/installation/index.rst
```diff
@@ -144,9 +144,10 @@ Dependency
 `pre-commit <https://pre-commit.com/index.html>`__                               \-                 dev                For managing and maintaining multi-language pre-commit hooks.
 `ruff <https://beta.ruff.rs/docs/>`__                                            \-                 dev                For code formatting.
 `nox <https://nox.thea.codes/en/stable/>`__                                      \-                 dev                For running testing sessions in multiple environments.
-`coverage <https://coverage.readthedocs.io/en/>`__                               \-                 dev                For calculating coverage of tests.
+`pytest-cov <https://pytest-cov.readthedocs.io/en/latest/>`__                    \-                 dev                For calculating coverage of tests.
```
agriyakhetarpal, line 147: (nit)

```suggestion
`pytest-cov <https://pytest-cov.readthedocs.io/en/stable/>`__                    \-                 dev                For calculating test coverage.
```

docs/source/user_guide/installation/index.rst
```diff
@@ -144,9 +144,10 @@ Dependency
...
 `parameterized <https://github.com/wolever/parameterized>`__                     \-                 dev                For test parameterization.
-`pytest <https://docs.pytest.org/en/stable/>`__                                  6.0.0              dev                For running Jupyter notebooks tests.
+`pytest <https://docs.pytest.org/en/stable/>`__                                  6.0.0              dev                For running Unit, Integration and Jupyter notebooks tests.
```
agriyakhetarpal, line 149:
```suggestion
`pytest <https://docs.pytest.org/en/stable/>`__                                  6.0.0              dev                For running the test suites.
```

pybamm/parameters/parameter_sets.py
```diff
@@ -17,6 +17,7 @@ class ParameterSets(Mapping):
 
     .. doctest::
```
agriyakhetarpal, line 18:
```suggestion
```

agriyakhetarpal (commented):
Looks good to me! I added a few suggestions plus a fix for the failing doctest. Thanks, @prady0t!

agriyakhetarpal:
Looks like there's another one of the `.. doctest::` directives below the one that we removed – could you fix that, @prady0t? This should be good to merge after that.

prady0t:
Done!

prady0t:
Test failure due to non-matching hashes happens quite frequently. Why does it happen? Is there a way to fix it? 
```
Downloading types_python_dateutil-2.9.0.20240316-py3-none-any.whl (9.7 kB)
ERROR: THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.
    unknown package:
        Expected sha256 d37f1d8cab6fca11d8ba44d6404fdbb9b796d4e6f447d12317e4ce0660310072
             Got        f6974e35bc2c42ad0002e80e04815ae8f65a012154c0de8b4daf00c9417b8bb4
```

agriyakhetarpal:
> Test failure due to non-matching hashes happens quite frequently. Why does it happen? Is there a way to fix it?

This is a bug with `pip`'s hash-checking functionality. I assume there are three intermittent factors that can lead to the error: 
1. `pip` either computes the SHA-256 checksum of the cached `types_python_dateutil` package incorrectly
2. `pip` either receives an incorrect SHA-256 checksum and that for the cached `types_python_dateutil` is correct
3. The GitHub Actions cache mechanisms are corrupted or that they corrupt this file (a bit unlikely).

Either way, a re-run typically fixes the issue, but it's strange that it happens with `types_python_dateutil` everytime I see it.

agriyakhetarpal (approved):

agriyakhetarpal:
Thanks for your work, @prady0t!

PR #4122, opened 2024-05-31
Updated SEI on cracks example

DrSOKane:
# Description

Updated plots in SEI on cracks example notebook to improve visibility


kratman:
@DrSOKane In the updated notebook, all the lines in the plots sit on top of each other. Is this intended?

kratman:
image

kratman:
Ok I just wanted to make sure since the change was changing the plotting styles, but then the plots also changed

DrSOKane:
> Ok I just wanted to make sure since the change was changing the plotting styles, but then the plots also changed

What happened was the code got changed, but the plots never updated. So now I've re-run the code to get the plots right.

kratman (approved):

Issue #2152, opened 2022-07-05
[Bug]: PyBaMM Unable to accurately re-create Ecker 2015 figures.


Tazarus:
### PyBaMM Version

22.5

### Python Version

3.9.7

### Describe the bug

**In recreating figures 4a), 7a), and 7b), from Ecker et al's 2015 paper "Parameterization of a Physico-Chemical Model of a Lithium-Ion Battery" I was unable to accurately replicate the groups simulated results using PyBaMM.**

I recreated figures 4a), 7a), and 7b). Including digitizing the original measurement data and overlaying it with the PyBaMM results to compare against the simulations used by the original authors.

**The PyBaMM DFN and SPMe results were noticeably less accurate then the discharge curve simulations done by Ecker et al in the original paper for most figures,** while still doing a reasonable job at normal operating temperatures.

However, **at higher charge rates and lower temperatures the discrepancies got quite bad**. Particularly for figures 7b) at discharge 1.3C , and figure 7a) at -10 celcius.

Considering there is already a premade Ecker 2015 parameter set for this very paper, **I was surprised that I was unable to reproduce the simulation results, even using a full DFN model.**

**I have included both the original figures and the figures I produced using PyBaMM with mirrored styles for easily comparison. I also included a simplified .py file with the basics for reproducing the most troublesome discharge curve from figure 7b).** 

I also included my full .ipynb file for the entire experiment, and the digitized measurement data from the figures from the paper (digitized using grapher) if anyone wants to have access to the experimental data for comparison purposes.

**The full .ipynb file is included primarily for completeness, it is much longer, and not really intended for troubleshooting**, as it is mostly just looping through lists of parameters with slight experimental or temperature variations to reproduce the ~4 discharge curves on each figure. But it is included in case anyone is interested.

[PyBaMM_Ecker.zip](https://github.com/pybamm-team/PyBaMM/files/9048592/PyBaMM_Ecker.zip)
[Ecker_figures.zip](https://github.com/pybamm-team/PyBaMM/files/9048593/Ecker_figures.zip)

### Steps to Reproduce

```
import pybamm
import matplotlib.pyplot as plt

model = pybamm.lithium_ion.DFN(name="DFN")

chemistry = pybamm.parameter_sets.Ecker2015

K = 273.15 # Celsius to Kelvin conversion factor
params = pybamm.ParameterValues(chemistry=chemistry)
params["Ambient temperature [K]"] = -10 + K
[...]

PR #4141, opened 2024-06-05
Update Ecker2015 solid-phase diffusivity functions to more closely match the paper

DrSOKane:
The solid-phase diffusivity functions in `Ecker2015` and `Ecker2015_graphite_halfcell` are updated to better match the functions in the original paper. A new citation `Yuan2023`, for which these parameters were used, is also added.

Advantages:
* The new functions reproduce phase separation behaviour in a similar way to the Cahn-Hilliard equations
* With them, the results in `Yuan2023` can be reproduced.

Disadvantages:
* This level of physical accuracy comes at a noticeable computational cost, comparable with that of `ORegan2022`
* The results in `Richardson2019` are no longer reproduced. However, that paper still made a major contribution towards our understanding of the `Ecker2015` parameters and is therefore still cited.


rtimms:
My preference would be to add a new `Yuan2023` parameter set so that the Ecker results don't change from version to version.

valentinsulzer (commented):
What's the formula for D_s_n and D_s_p actually used by Ecker? I can't find it in Part I or Part II.

CHANGELOG.md
```diff
@@ -47,6 +48,7 @@
 
 ## Breaking changes
 
+- Updated solid-phase diffusivity functions in `Ecker2015` and `Ecker2015_graphite_halfcell` parameter sets to better match the original papers ([#4141](https://github.com/pybamm-team/PyBaMM/pull/4141))
```
kratman, line 51: This should probably be under only "breaking changes" or "features" instead of both

kratman:
@DrSOKane Is this one fixed by this change as well?

#2152

DrSOKane:
> What's the formula for D_s_n and D_s_p actually used by Ecker? I can't find it in Part I or Part II.

They don't give a formula, only data. It's possible that they used an interpolation between the data points. LiionDB has this data (I put it there!) so an alternative approach would be to use that.

kratman:
@DrSOKane This PR was marked as merged because the commits were also present in #4168

PR #4150, opened 2024-06-07
3937 events with idaklu output variables

### LINKED ISSUE #3937: [Bug]: output variables option in idaklu solver fails when event triggered
### PyBaMM Version

24.1

### Python Version

3.10

### Describe the bug

When the output_variables option of the IDAKLU solver is used in a simulation with an event, the solve fails with the error message below when trying to get the event reason

### Steps to Reproduce

```python
def test(solver):

    sim = pybamm.Simulation(
        model,
        solver=solver,
    )
    sol = sim.solve(np.linspace(0, 3600, 1000))
    sol = sol["Terminal voltage [V]"].entries

test(pybamm.IDAKLUSolver(output_variables=["Terminal voltage [V]"]))
```

### Relevant log output

```shell
  File "/home/mrobins/git/tmp/test.py", line 17, in <module>
    test(pybamm.IDAKLUSolver(output_variables=["Terminal voltage [V]"]))
  File "/home/mrobins/git/tmp/test.py", line 14, in test
    sol = sim.solve(np.linspace(0, 3600, 1000))
  File "/home/mrobins/git/tmp/env/lib/python3.10/site-packages/pybamm/simulation.py", line 584, in solve
    self._solution = solver.solve(self.built_model, t_eval, **kwargs)
  File "/home/mrobins/git/tmp/env/lib/python3.10/site-packages/pybamm/solvers/base_solver.py", line 959, in solve
    solutions[i], termination = self.get_termination_reason(
  File "/home/mrobins/git/tmp/env/lib/python3.10/site-packages/pybamm/solvers/base_solver.py", line 1303, in get_termination_reason
    final_event_values[event.name] = event.expression.evaluate(
  File "/home/mrobins/git/tmp/env/lib/python3.10/site-packages/pybamm/expression_tree/binary_operators.py", line 132, in evaluate
    right = self.right.evaluate(t, y, y_dot, inputs)
  File "/home/mrobins/git/tmp/env/lib/python3.10/site-packages/pybamm/expression_tree/binary_operators.py", line 132, in evaluate
    right = self.right.evaluate(t, y, y_dot, inputs)
  File "/home/mrobins/git/tmp/env/lib/python3.10/site-packages/pybamm/expression_tree/binary_operators.py", line 132, in evaluate
    right = self.right.evaluate(t, y, y_dot, inputs)
  File "/home/mrobins/git/tmp/env/lib/python3.10/site-packages/pybamm/expression_tree/symbol.py", line 778, in evaluate
    return self._base_evaluate(t, y, y_dot, inputs)
  File "/home/mrobins/git/tmp/env/lib/python3.10/site-packages/pybamm/expression_tree/state_vector.py", line 284, in _base_evaluate
    raise ValueError(
ValueError: y is too short, so value with slice is smaller than expected
(env) ➜  tmp pip list |grep pybamm
```

pipliggins:
# Description

Simulations using the IDAKLU solver with output variables specified which terminated with an event rather than the final time step would crash trying to find the termination event.

Termination events are found in the Solution class by using the state vector of the final time-step to evaluate each potential termination event and find the minimum (the event which caused the termination). When output variables are specified in the IDAKLU solver only those variables are output, not the entire state vector, so the events could not be evaluated outside the solver.

This PR adds a check to see if the events specified in a model can be calculated using the values of the output variables rather than the state vector. If they can, the events are calculated using these values after the solve. If not, an error is raised warning that events will not be able to be calculated with the current output variables, and provides a list of variables to choose from to add to the output_variables list, e.g.:
```

sim2 = pybamm.Simulation(
  model,
  solver=pybamm.IDAKLUSolver(output_variables=["Cell temperature [K]"]),
)
sim2.solve(np.linspace(0, 3600, 1000))
```
raises
```
ValueError: IDA KLU solver: Event 'Minimum voltage [V]' cannot be calculated using the current output variables.
Add one of the following to `output_variables`:
Positive current collector potential [V]
Local voltage [V]
Terminal voltage [V]
Voltage [V]
Battery voltage [V]
```

Fixes #3937 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/solvers/base_solver.py
+++ b/pybamm/solvers/base_solver.py
@@ -1468,21 +1468,12 @@ class BaseSolver:
                 for k, v in model.variables.items():
                     if v in event_children:
                         required_vars.append(k)
-                if required_vars:
-                    joined_vars = "\n".join(required_vars)
-                    raise ValueError(
-                        f"{self.name}: Event '{event.name}' cannot be calculated "
-                        "using the current output variables.\n"
-                        f"Add one of the following to `output_variables`:\n{joined_vars}"
-                    )
-                else:
-                    # no variables identified which match the event children
-                    raise ValueError(
-                        f"{self.name}: Event '{event.name}' cannot be calculated "
-                        "using the current output variables.\n"
-                        f"Add one of {event_children} as a variable in the model, "
-                        "then add the variable to `output_varaibles`."
-                    )
+                joined_vars = "\n".join(required_vars)
+                raise ValueError(
+                    f"{self.name}: Event '{event.name}' cannot be calculated "
+                    "using the current output variables.\n"
+                    f"Add one of the following to `output_variables`:\n{joined_vars}"
+                )
 
 
 def process(
--- a/pybamm/solvers/base_solver.py
+++ b/pybamm/solvers/base_solver.py
@@ -261,6 +261,8 @@ class BaseSolver:
             self.computed_var_fcns = {}
             self.computed_dvar_dy_fcns = {}
             self.computed_dvar_dp_fcns = {}
+            if self.output_variables:
+                self._compare_events_to_output_variables(model, self.output_variables)
             for key in self.output_variables:
                 # ExplicitTimeIntegral's are not computed as part of the solver and
                 # do not need to be converted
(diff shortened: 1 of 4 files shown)


pybamm/solvers/base_solver.py
```diff
@@ -1425,6 +1429,52 @@ def _set_up_model_inputs(model, inputs):
...
+            matching_var = {k: v for k, v in output_vars.items() if v in event_children}
+
+            # search for variables that can be used to calculate the event
+            if len(matching_var) == 0:
+                required_vars = []
+                for k, v in model.variables.items():
+                    if v in event_children:
+                        required_vars.append(k)
+                joined_vars = "\n".join(required_vars)
+                raise ValueError(
```
jsbrittain, line 1472: Would it be better to raise a [warning](https://docs.python.org/3/library/warnings.html) here rather than an Error, as this lets the user decide whether identification of the specific termination event is worth the expense of adding multiple additional variables (and their sensitivities) to the solution.

pybamm/solvers/idaklu_solver.py
```diff
@@ -711,3 +711,87 @@ def jaxify(
...
+                }
+                if len(matching_var) != 1:
+                    try:
+                        final_event_values[event.name] = event.expression.evaluate(
+                            solution.t_event,
+                            solution.y_event,
+                            inputs=solution.all_inputs[-1],
+                        )
+                        continue
+                    except ValueError as e:  # pragma: no cover
```
jsbrittain, line 763: Likewise, is this cause for a critical failure? I would consider replacing with a warning.

pybamm/solvers/idaklu_solver.py
```diff
@@ -711,3 +711,87 @@ def jaxify(
...
+                    solution.t_event,
+                    solution.y_event,
+                    inputs=solution.all_inputs[-1],
+                )
+            termination_event = min(final_event_values, key=final_event_values.get)
+
+            # Check that it's actually an event
+            if final_event_values[termination_event] > 0.1:  # pragma: no cover
+                # Hard to test this
+                raise pybamm.SolverError(
```
jsbrittain, line 789: dito.

pybamm/solvers/base_solver.py
pipliggins: Raising a warning and filling in the termination event with an empty string/"unknown" or similar would work for single simulations, but when used with an experiment the termination event is required to check when to move through experiment steps. This would require adding in different behaviour for using the solver with/without an experiment, perhaps auto-adding appropriate output variables when used with experiments. Currently sensitivities can't be used with experiments so there's minimal change to the computational expense.

An alternative implementation to consider in the future might be to change the IDAKLU solver output to include the state vector for the final timestep in the return value. This shouldn't impact the memory usage, and would allow calculation of any termination events using the pre-existing methods without the additional overhead of storing and calculating unwanted output variables/sensitivities during the solve.

pybamm/solvers/idaklu_solver.py
pipliggins: See above

pybamm/solvers/base_solver.py
jsbrittain: Sounds good!

valentinsulzer: > An alternative implementation to consider in the future might be to change the IDAKLU solver output to include the state vector for the final timestep in the return value. This shouldn't impact the memory usage, and would allow calculation of any termination events using the pre-existing methods without the additional overhead of storing and calculating unwanted output variables/sensitivities during the solve.

This is a great solution for all solvers since the final state vector is required to initialize the next step in an experiment.

PR #4157, opened 2024-06-09
Rename string(string) to avoid a scope issue

kratman:
# Description

The function
```
def string(string, **kwargs):
```
used the same name for the argument as the function.

I slightly improved this by changing the argument to "text" instead.
```
def string(text, **kwargs):
```

# Key checklist:

## Further checks:

### DIFF
--- a/docs/source/examples/notebooks/change-settings.ipynb
+++ b/docs/source/examples/notebooks/change-settings.ipynb
@@ -36,9 +36,9 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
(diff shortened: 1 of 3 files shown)


agriyakhetarpal (approved):

agriyakhetarpal:
Does this need a CHANGELOG entry or is it minor enough to leave it out?

agriyakhetarpal:
Does this rename need a CHANGELOG entry or is it minor enough to leave it out?

kratman:
> Does this rename need a CHANGELOG entry or is it minor enough to leave it out?

It is a change to an internal variable name. No need for a change log entry, it won't break anything

PR #4158, opened 2024-06-09
Minor refactoring on deserialize

kratman:
# Description

Removes some duplicated code and fixes some warnings in the documentation

# Key checklist:

## Further checks:


pybamm/models/submodels/base_submodel.py
```diff
@@ -33,15 +30,15 @@ class BaseSubModel(pybamm.BaseModel):
     ----------
     param: parameter class
         The model parameter symbols
-    rhs: dict
+    _rhs: dict
```
valentinsulzer, line 33: Could we document the properties instead of these attributes? Or just remove the attributes from the docstring? We don't want to tell users to use `model._rhs`

kratman: Yeah I can remove all the internal ones

kratman: Ok I removed all the private ones from the modified files. If you want me to go through remove all attributes and/or make sure all properties are covered then I will do that part in another ticket

valentinsulzer: Thanks. It's not great having only some of the attributes documented either. Maybe we should just remove documentation of all attributes?

kratman: At least right now it only documents the public ones. I will make a ticket for fixing the docs properly. I only renamed them in the first place to remove some IDE warnings. The main goal of this PR was to remove the duplicated code

kratman: Covered by #4170

valentinsulzer (approved):

PR #4102, opened 2024-05-20
Issue 4082 Documentation CI Fix

cringeyburger:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.
The documentation built on CI fails with an error. It was found that we still need to install setuptools before docs/doctests sessions to avoid erroring out. 
Also, the scheduled jobs are parallelized for faster CI

Fixes #4082 

## Type of change

Please add a line in the relevant section of [CHANGELOG.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CHANGELOG.md) to document the change (include PR #) - note reverse order of PR #s. If necessary, also add to the list of breaking changes.

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (back-end change that speeds up the code)
- [x] Bug fix (non-breaking change which fixes an issue)

# Key checklist:

- [x] No style issues: `$ pre-commit run` (or `$ nox -s pre-commit`) (see [CONTRIBUTING.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CONTRIBUTING.md#installing-and-using-pre-commit) for how to set this up to run automatically when committing locally, in just two lines of code)
- [x] All tests pass: `$ python run-tests.py --all` (or `$ nox -s tests`)
- [x] The documentation builds: `$ python run-tests.py --doctest` (or `$ nox -s doctests`)

You can run integration tests, unit tests, and doctests together at once, using `$ python run-tests.py --quick` (or `$ nox -s quick`).

## Further checks:

- [ ] Code is commented, particularly in hard-to-understand areas
- [ ] Tests added that prove fix is effective or that feature works

PR #4172, opened 2024-06-12
Sync periodic tests with the push workflow

kratman:
# Description

Syncs the test on push workflows with the periodic tests.

# Key checklist:

## Further checks:


kratman:
This needs some more work before it will be ready for review. I just pushed it in advance as a reminder to myself for the weekend.

agriyakhetarpal:
Thanks, @kratman – I know see that you mentioned needs some work, but I am not very supportive of these changes in general. Two reasons:

1. We want to avoid the caching stuff in the scheduled tests so as to always download newer versions packages from PyPI and foresee things in advance if they are going to fail due to upstream changes. The scheduled tests are meant to fail to catch things early (same as the philosophy for dependents that will make use of the nightly releases in #3945).
2. I opine that the integration and the unit tests in this workflow should stay in parallel jobs (they were very recently split up in #4102 to run in parallel), because the run time for a nightly job is not a concern. If they run in serial, it's not possible to find any potential failures in the integration tests if the unit tests fail.

I guess a better way of refactoring this would be to create a `setup-pybamm.yml` workflow that runs on the `workflow_call:` event which can be reused both in PR tests and the scheduled tests (it can take `cache` as a string-based workflow input that can be turned `on` or `off` based on where the workflow is triggered from).

kratman:
@agriyakhetarpal I started this because the periodic tests were failing, so I started by syncing the workflows. If things like caching are problematic then I can remove them from this workflow as I go. The goal is to get the periodic tests running smoothly

agriyakhetarpal:
Okay, makes sense, thanks for getting this done. I can review after this is ready of course

kratman:
This seemed to run locally besides the mseries block which only runs on the main pybamm repo. Not sure if that CI job is even needed anymore now that we have the macos-13 and macos-14 jobs

kratman:
I did a quick fix for a lychee issue, but it just exposed more issues with links from tinyurl. I will address those elsewhere

kratman:
Remaining lychee issues are fixed by #4219

agriyakhetarpal (commented):
> The integration and the unit tests in this workflow should stay in parallel jobs because the run time for a nightly job is not a concern. If they run in serial, it's not possible to find any potential failures in the integration tests if the unit tests fail.

The changes here look good to me. I still feel that the above point applies, though. Creating a reusable workflow with inputs for disabling/enabling etc. that responds to a `workflow_call:` event, or separating CI jobs into different files by either OS or the type of the test suite (unit, integration), etc. seem to be the better approaches and are similar to how they are structured in other projects I have looked at

kratman:
> The integration and the unit tests in this workflow should stay in parallel jobs because the run time for a nightly job is not a concern. If they run in serial, it's not possible to find any potential failures in the integration tests if the unit tests fail.

Yeah I figure that making them match brings us closer to using reusable workflows. This makes sure that the same test procedures are used for the regular PRs and the periodic tests. I think the reusable workflow and additional configuration can be revisited in the future. Our infrastructure is already quite complicated so the simpler we can make it the better

kratman:
@agriyakhetarpal Is this good to go then?

agriyakhetarpal:
I feel the reusable workflow is easy enough to be added here itself before we merge, but if you wish to do it in a follow-up after this PR, I shall approve it

kratman:
> I feel the reusable workflow is easy enough to be added here itself before we merge, but if you wish to do it in a follow-up after this PR, I shall approve it

I created a followup ticket since I suspect that more can be turned into re-usable workflows. For now, this improves the scheduled tests, so I think I can safely revisit the reusable workflows

.github/workflows/run_periodic_tests.yml
```diff
@@ -45,6 +42,7 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install gfortran gcc graphviz pandoc libopenblas-dev texlive-latex-extra dvipng
+          sudo dot -c
```
agriyakhetarpal, line 45:
```suggestion
```

This won't be needed if we're not caching `graphviz`

.github/workflows/run_periodic_tests.yml
```diff
@@ -65,6 +63,7 @@ jobs:
         run: choco install graphviz --version=8.0.5
 
       - name: Set up Python ${{ matrix.python-version }}
+        id: setup-python
```
agriyakhetarpal, line 66:
```suggestion
```

Since we're not accessing this step in any further steps

.github/workflows/run_periodic_tests.yml
```diff
@@ -78,98 +77,29 @@ jobs:
...
-        run: python -m nox -s pybamm-requires
-
       - name: Run integration tests for ${{ matrix.os }} with Python ${{ matrix.python-version }}
         run: python -m nox -s integration
 
   # Skips IDAKLU module compilation for speedups, which is already tested in other jobs.
   run_doctests:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
```
agriyakhetarpal, line 100:
```suggestion
```

This is a single job and not a matrix and therefore does not require a strategy definition

.github/workflows/run_periodic_tests.yml
```diff
@@ -179,9 +109,11 @@ jobs:
       - name: Install Linux system dependencies
         run: |
           sudo apt-get update
-          sudo apt-get install graphviz pandoc libopenblas-dev texlive-latex-extra dvipng
+          sudo apt-get install graphviz pandoc texlive-latex-extra dvipng
+          sudo dot -c
```
agriyakhetarpal, line 113:
```suggestion
```

Similar reason as above

.github/workflows/run_periodic_tests.yml
```diff
@@ -179,9 +109,11 @@ jobs:
...
 
-      - name: Set up Python 3.11
+      - name: Set up Python
+        id: setup-python
```
agriyakhetarpal, line 116:
```suggestion
```

agriyakhetarpal (approved):

PR #4178, opened 2024-06-14
Extend Step String Tagging Functionality

### LINKED ISSUE #4177: Extend Step String Tagging Functionality
### Description

This is a feature request covering multiple related feature additions surrounding the `pybamm.step.string` tags functionality. Currently there is the ability to add tags to pybamm step strings, and there is the ability to search an experiment for the indices where specific tags exist. There is not currently any additional functionality of the step string tags. The proposed added functionality is the following:

- The ability to search a solution object for specific steps based on tags
- The added functionality to export tags along with cycle and step indices when using the `Solution.get_data_dict()` method

### Motivation

I believe these features are the would-be natural evolution of additional functionality surrounding step string tags. I believe the true utility of these step string tags come when you can use them to filter a solution object, or to filter data exported from a solution object.

## Core Implementation
I believe there are a number of steps for the proposed core implementation.

1. Add a property to the `Solution` class: `.tags`
2. By default, when running an experiment, each time a `step_solution` is [generated](https://github.com/pybamm-team/PyBaMM/blob/91bfed4cc9abaceedfdbf93532df367e9809f8ee/pybamm/simulation.py#L692), the tags from the associated step (from the experiment) should then be assigned as a property to the `Solution` object
3. The method `Solution.get_data_dict()` should export a 'Tags' column along with 'Cycle' and 'Step' columns.
4. An additional method should be added: `Solution.search_tags()`, similar to `Experiment.search_tag()`, allowing for searching of one of many tags in the solution sub_solutions

## Additional Features
I also believe there are a number of additional changes that may be useful to include as well

1. The inclusion of tags in the `__repr__` of the `Solution` object
2. Similar to how tags would be associated with a `Solution` object (as a property), there may also be the properties `.cycle` and `.step` which are associated with a `Solution` when the solution results from an experiment.
3. The `Solution` object holds the additional variables, accessible through the `__getitem__()` method, being "Cycle", "Step", and "Tags"

## Remaining Questions

- Should calling `Solution.get_data_dict()` when there are no tags result in returning an empty "Tags" column, or no "Tags" column at all?

### Additional context

_No response_

mleot:
# Description

Adds additional funcitonality around step string tags. Allows for filtering a solution object based on tags. Allows for inclusion of tags when exporting data through the `Solution.get_data_dict()` method.

Fixes #4177

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -931,6 +931,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "Ubham16",
+      "name": "Ubham16",
+      "avatar_url": "https://avatars.githubusercontent.com/u/173074476?v=4",
+      "profile": "https://github.com/Ubham16",
+      "contributions": [
+        "code"
+      ]
     }
   ],
   "contributorsPerLine": 7,
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,7 @@ ci:
 
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: "v0.4.9"
+    rev: "v0.4.10"
     hooks:
       - id: ruff
         args: [--fix, --show-fixes]
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@
 [![code style](https://img.shields.io/endpoint?url=https://raw.githubusercontent.com/charliermarsh/ruff/main/assets/badge/v2.json)](https://github.com/astral-sh/ruff)
 
 <!-- ALL-CONTRIBUTORS-BADGE:START - Do not remove or modify this section -->
-[![All Contributors](https://img.shields.io/badge/all_contributors-87-orange.svg)](#-contributors)
+[![All Contributors](https://img.shields.io/badge/all_contributors-88-orange.svg)](#-contributors)
 <!-- ALL-CONTRIBUTORS-BADGE:END -->
 
 </div>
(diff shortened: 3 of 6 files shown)


tests/unit/test_experiments/test_simulation_with_experiment.py
```diff
@@ -438,6 +438,111 @@ def test_save_at_cycles(self):
...
+        sub_solutions = solution.search_tags(["tag1"])
+        assert len(sub_solutions) == 5
+
+        sub_solutions = solution.search_tags(["tag3"])
+        assert len(sub_solutions) == 5
+
+        sub_solutions = solution.search_tags(["tag1", "tag3"])
+        assert len(sub_solutions) == 10
+
+    def test_experiment_solution_step_tags(self):
```
valentinsulzer, line 474: I don't think this test adds any additional checks compared to the previous test?

tests/unit/test_experiments/test_simulation_with_experiment.py
```diff
@@ -438,6 +438,111 @@ def test_save_at_cycles(self):
...
+        model = pybamm.lithium_ion.SPM()
+
+        # O'Kane 2022: pos = function, neg = data
+        param = pybamm.ParameterValues("OKane2022")
+        sim = pybamm.Simulation(model, experiment=experiment, parameter_values=param)
+        solution = sim.solve(solver=pybamm.CasadiSolver("fast with events"))
+
+        assert solution.cycles[0].steps[0].tags == ["tag1", "tag2"]
+        assert solution.cycles[1].steps[0].tags == ["tag3"]
+        assert solution.cycles[1].steps[1].tags == ["tag4"]
```
valentinsulzer, line 499: just add these asserts to the previous test

tests/unit/test_solvers/test_solution.py
```diff
@@ -359,6 +359,30 @@ def test_get_data_cycles_steps(self):
...
+            ]
+            * 5,
+        )
+        sim = pybamm.Simulation(
+            model=model,
+            experiment=experiment,
+        )
+        solution = sim.solve()
+        data_dict = solution.get_data_dict(["Terminal voltage [V]", "Time [s]"])
+        assert set(data_dict["Tags"]).issubset({"tag1;tag2", "tag3", "tag4"})
```
valentinsulzer, line 384: This test could quite easily have false positives. Could you create a simpler experiment (e.g. just two steps of known length) and test `data_dict["Tags"]` for an exact match?

valentinsulzer (changes requested):
Implementation looks good to me, just a few comments about the tests

tests/unit/test_solvers/test_solution.py
mleot: Sure! Good point. Thanks for the feedback.

tests/unit/test_experiments/test_simulation_with_experiment.py
mleot: I think you are correct. I will add both to a single test.

mleot:
@valentinsulzer I added an additional feature, being the .step and .cycle attributes to the solution objects, which are assigned during the solve step of a simulation, when running 'with experiment'.

agriyakhetarpal:
> Are assert statements generally frowned upon in testing? I noticed that it raises issues in the static code analysis.

They are usually frowned upon in regular code, but I think we need to make Codacy more lenient about this a bit, because these are all in `tests/`, where `assert`s are commonplace. Soon, the use of asserts will grow a lot in PyBaMM (see #4156). I am unsure who has access to the relevant settings and if we can configure it with a `codacy.yml` file of sorts.

kratman:
@mleot are you still working on this?

PR #4180, opened 2024-06-14
Removing `run-tests.py` file.

prady0t:
# Description

Trying to remove run-tests.py file.


prady0t:
Yet to add `conftest.py` file

noxfile.py
```diff
@@ -120,7 +127,7 @@ def run_scripts(session):
     # is fixed
     session.install("setuptools", silent=False)
     session.install("-e", ".[all,dev]", silent=False)
-    session.run("python", "run-tests.py", "--scripts")
+    session.run("python", "-m", "pytest", "tests/test_examples.py")
```
agriyakhetarpal, line 130:
```suggestion
    session.run("python", "-m", "pytest", "-v", "tests/test_examples.py")
```

Soon enough, we should enable more options here when this PR is closer to merging. Personally, I like to use `pytest -svra`.

agriyakhetarpal (commented):
This is looking good! I tried locally and

```bash
==================================================================== 44 passed in 100.46s (0:01:40) ====================================================================
nox > Session scripts was successful.
```

much faster than the serial implementation. I checked and the slowest scripts are `run_ecm.py`, `experiment_drive_cycle.py`,  and `pouch_cell_cooling.py`. It might be worth improving those, but I guess 100 seconds for 44 scripts isn't too bad at all, afterall, and is rather great in terms of speedups.

noxfile.py
kratman: Do the options have any impact on the speed of the tests. I assume it would be negligible, but I am curious

arjxn-py: I too believe that i'd be negligible since there's no extensive logging but I guess we can definitely do some benchmarking before merging to quantify the range of time difference

agriyakhetarpal: Yes, as Arjun says, the impact is indeed negligible and it isn't measurable either; `-v` is to increase the verbosity, `-s` controls any print statements or logging statements in stdout, and `-ra` will display a summary of everything: failed + skipped + xfailed/xpassed + errors, after the test suite completes.

agriyakhetarpal: For reference, after we'll have most of this ready and when we'll be starting with the migration of the tests (which is from this week onwards), a useful guide with some tips is present at https://learn.scientific-python.org/development/guides/pytest/

prady0t: should i include `-svra` in all nox commands?

prady0t:
I think I messed up the branch. I'm having problems while pushing changes. I've made the changes.

noxfile.py
agriyakhetarpal: Yes, should be okay

prady0t: Output with `-svra` is actually pretty clean. Thanks for the suggestion!

kratman:
@prady0t Is your branch still having problems? Do you need help fixing the branch?

prady0t:
> @prady0t Is your branch still having problems? Do you need help fixing the branch?

I can work with this, although I made a lot of commits.

agriyakhetarpal:
No worries, we'll squash merge anyway, but to learn how to rewrite history would be up to you as an exercise

conftest.py
```diff
@@ -0,0 +1,60 @@
+import pytest
+import numpy as np
+import pybamm
+
+
+def pytest_addoption(parser):
+    parser.addoption(
+        "--scripts", action="store_true", default=False, help="run examples tests"
```
agriyakhetarpal, line 8:
```suggestion
        "--scripts", action="store_true", default=False, help="execute the example scripts"
```

conftest.py
```diff
@@ -0,0 +1,60 @@
...
+    parser.addoption(
+        "--integration",
+        action="store_true",
+        default=False,
+        help="run integration tests",
+    )
+
+
+def pytest_configure(config):
+    config.addinivalue_line("markers", "scripts: mark test as an example")
```
agriyakhetarpal, line 22:
```suggestion
    config.addinivalue_line("markers", "scripts: mark test as an example script")
```

noxfile.py
```diff
@@ -98,7 +104,7 @@ def run_unit(session):
     set_environment_variables(PYBAMM_ENV, session=session)
     session.install("setuptools", silent=False)
     session.install("-e", ".[all,dev,jax]", silent=False)
-    session.run("python", "run-tests.py", "--unit")
+    session.run("python", "-m", "pytest", "-svra", "-m", "unit")
```
agriyakhetarpal, line 107: I know I suggested adding this earlier, but I think we have something better. Let's add these to `addopts` in `pyproject.toml` under `[tool.pytest.ini_options]` instead.
```suggestion
    session.run("python", "-m", "pytest", "-m", "unit")
```

agriyakhetarpal: Also, can we make them run with `pytest --unit` instead of `pytest -m unit`? No big deal if we can't, just wanted to know if it's possible to do so with how you have configured the markers

noxfile.py
```diff
@@ -120,7 +128,7 @@ def run_scripts(session):
     # is fixed
     session.install("setuptools", silent=False)
     session.install("-e", ".[all,dev]", silent=False)
-    session.run("python", "run-tests.py", "--scripts")
+    session.run("python", "-m", "pytest", "-svra", "--scripts", "tests/test_scripts.py")
```
agriyakhetarpal, line 131:
```suggestion
    session.run("python", "-m", "pytest", "--scripts", "tests/test_scripts.py")
```

agriyakhetarpal: For example, this session runs `pytest --scripts`, while the `unit` session runs `pytest -m unit` (related to my question above). I think `pytest --unit`, `pytest --integration`, etc. are shorter and better to use.

Also, it would be great if we could configure `pytest --scripts` to run just `tests/test_scripts.py` itself, rather than us having to specify the file.

noxfile.py
```diff
@@ -151,7 +159,7 @@ def run_tests(session):
     set_environment_variables(PYBAMM_ENV, session=session)
     session.install("setuptools", silent=False)
     session.install("-e", ".[all,dev,jax]", silent=False)
-    session.run("python", "run-tests.py", "--all")
+    session.run("python", "-m", "pytest", "-svra")
```
agriyakhetarpal, line 162:
```suggestion
    session.run("python", "-m", "pytest")
```

This command will run not only the unit and integration tests but also the example scripts, which contradicts the docstring for the function. Can we change it to `pytest --unit --integration` or similar?

pyproject.toml
```diff
@@ -233,7 +238,7 @@ addopts = [
     "--strict-markers",
 ]
 testpaths = [
-    "docs/source/examples/",
+    "tests",
```
agriyakhetarpal, line 241: If we were to add both `tests/` and `docs/source/examples/` here, could we skip specifying the path in `pytest --nbmake` above?

pyproject.toml
```diff
@@ -221,6 +221,11 @@ ignore = [
 
 [tool.pytest.ini_options]
 minversion = "6"
```
agriyakhetarpal, line 223:
```suggestion
minversion = "8"
```

Let us bump this to `8` here because `6` is too old now (please update in the `[dev]` extras and the installation guide table in the docs as well).

pyproject.toml
```diff
@@ -221,6 +221,11 @@ ignore = [
...
+markers = [
+      "examples: mark a test as an example test",
+]
+# Use pytest-xdist to run tests in parallel by default, exit with
+# error if not installed
 required_plugins = [
     "pytest-xdist",
     "pytest-mock",
```
agriyakhetarpal, line 231: We can now add `-svra` to `addopts` below this line

tests/test_scripts.py
```diff
@@ -0,0 +1,24 @@
+import os
+import runpy
+
+import pytest
+from pathlib import Path
+
+
+ROOT_DIR = Path(os.path.join(os.path.dirname(__file__), ".."))
```
agriyakhetarpal, line 8: We can use `pathlib.Path()` even further by replacing this as follows:

```suggestion
ROOT_DIR = Path(__file__).parent.parent
```

which is cleaner!

tests/test_scripts.py
```diff
@@ -0,0 +1,24 @@
...
+
+class TestExamples:
+    """
+    A class to test the example scripts.
+    """
+
+    def list_of_files():
+        base_dir = Path.joinpath(ROOT_DIR, "examples", "scripts")
+        file_list = base_dir.rglob("*py")
+        return file_list
```
agriyakhetarpal, line 19:
```suggestion
    def list_of_files():
        file_list = (ROOT_DIR / "examples" / "scripts").rglob("*.py")
        return file_list
```

tests/testcase.py
```diff
@@ -2,52 +2,12 @@
...
-            if isinstance(attribute, FunctionType):
-                attribute = FixRandomSeed(attribute)
-            newClassDict[attributeName] = attribute
-        return type.__new__(meta, classname, bases, newClassDict)
-
-
-class TestCase(unittest.TestCase, metaclass=MakeAllTestsDeterministic):
-    """
-    Custom TestCase class for pybamm
+    Custom TestCase class for pybamms
```
agriyakhetarpal, line 9:
```suggestion
    Custom TestCase class for PyBaMM
```

tests/testcase.py
```diff
@@ -2,52 +2,12 @@
...
+    TO BE REMOVED
     """
 
     def assertDomainEqual(self, a, b):
```
agriyakhetarpal, line 13: Note: could you move `assertDomainEqual` to `shared.py` in another PR later? (I think we discussed this somewhere)

agriyakhetarpal: Never mind, you've already done it in https://github.com/pybamm-team/PyBaMM/pull/4231!

agriyakhetarpal:
@prady0t, could you fix the failure in the style jobs so that the tests can run?

prady0t:
> @prady0t, could you fix the failure in the style jobs so that the tests can run?

Done. I removed lot of code lines from `conftest.py` file and now we are using `-m` for everything. If a user runs pytest without any marker value, it runs all the tests, otherwise only those specified by `-m` flag.

noxfile.py
agriyakhetarpal: Discussed on Slack, and `pytest -m unit` is the better one to use

pyproject.toml
agriyakhetarpal: Resolved in the meeting and we'll have to pass the path to the notebooks explicitly, it's already how it is like in our documentation: https://docs.pybamm.org/en/stable/source/user_guide/contributing.html#testing-the-example-notebooks

prady0t:
Added documentation fixes.

CONTRIBUTING.md
```diff
@@ -149,9 +149,11 @@ The `test_optional_dependencies` function extracts `pybamm` mandatory distributi
 
 ## Testing
 
-All code requires testing. We use the [unittest](https://docs.python.org/3.3/library/unittest.html) package for our tests. (These tests typically just check that the code runs without error, and so, are more _debugging_ than _testing_ in a strict sense. Nevertheless, they are very useful to have!)
+All code requires testing. We use the [pytest](https://docs.pytest.org/en/latest/) package for our tests. (These tests typically just check that the code runs without error, and so, are more _debugging_ than _testing_ in a strict sense. Nevertheless, they are very useful to have!)
 
-We also use [pytest](https://docs.pytest.org/en/latest/) along with the [nbmake](https://github.com/treebeardtech/nbmake) and the [pytest-xdist](https://pypi.org/project/pytest-xdist/) plugins to test the example notebooks.
+We also use it along with the [nbmake](https://github.com/treebeardtech/nbmake) and the [pytest-xdist](https://pypi.org/project/pytest-xdist/) plugins to test the example notebooks.
+
+Warning : PyBaMM is currently in the process of test migration to pytest. We strongly recommend using pytest for any testing outside nox.
```
agriyakhetarpal, line 156:
```suggestion
```

We should let go of the warning for now. Not many people ask about how to run tests, and I don't think this needs emphasis, because we are removing any mentions of `unittest` here anyway.

tests/test_scripts.py
```diff
@@ -0,0 +1,22 @@
...
+    """
+
+    def list_of_files():
+        file_list = (ROOT_DIR / "examples" / "scripts").rglob("*.py")
+        return file_list
+
+    @pytest.mark.parametrize("files", list_of_files())
+    @pytest.mark.scripts
+    def test_example_scripts(self, files):
+        runpy.run_path(files)
```
agriyakhetarpal, line 22: One thing I'm finding difficult with this approach right now is about seeing the names of the files – `pytest` displays just numbers, and the `-s` option means that it prints out a lot of information, coupled with the additional logging we've set through the fixture. Is it possible to maintain a clear separation between the file names and actually print the file names in the test summary info at the end? You may refer to the difference between the current logs and the logs in other PRs. I don't think the verbose logging is helping all that much – scrolling through 14K+ lines of logs isn't useful to us.

CONTRIBUTING.md
```diff
@@ -149,9 +149,11 @@ The `test_optional_dependencies` function extracts `pybamm` mandatory distributi
 
 ## Testing
 
-All code requires testing. We use the [unittest](https://docs.python.org/3.3/library/unittest.html) package for our tests. (These tests typically just check that the code runs without error, and so, are more _debugging_ than _testing_ in a strict sense. Nevertheless, they are very useful to have!)
+All code requires testing. We use the [pytest](https://docs.pytest.org/en/latest/) package for our tests. (These tests typically just check that the code runs without error, and so, are more _debugging_ than _testing_ in a strict sense. Nevertheless, they are very useful to have!)
 
-We also use [pytest](https://docs.pytest.org/en/latest/) along with the [nbmake](https://github.com/treebeardtech/nbmake) and the [pytest-xdist](https://pypi.org/project/pytest-xdist/) plugins to test the example notebooks.
+We also use it along with the [nbmake](https://github.com/treebeardtech/nbmake) and the [pytest-xdist](https://pypi.org/project/pytest-xdist/) plugins to test the example notebooks.
```
agriyakhetarpal, line 154: Could we revise this to say something on the lines: "we use the following plugins:" followed by a one-line description of the three of them?

tests/test_scripts.py
arjxn-py: I agree with that, unit tests logs are also a little too much ~75,000.

About improving the clarity of test output, maybe we can try modifying the way pytest parametrize displays the file names.

Instead of just displaying numbers, we can try displaying a more descriptive name by using `pytest.param` while returning file list, try returning a `pytest.param` object with an id set to the file name like `[pytest.param(file, id=file.name) for file in file_list]`

arjxn-py (changes requested):

CONTRIBUTING.md
```diff
@@ -162,14 +166,14 @@ nox -s unit
...
-python run-tests.py --unit
+pytest -m unit
 ```
 
 ### Writing tests
 
 Every new feature should have its own test. To create ones, have a look at the `test` directory and see if there's a test for a similar method. Copy-pasting this is a good way to start.
 
-Next, add some simple (and speedy!) tests of your main features. If these run without exceptions that's a good start! Next, check the output of your methods using any of these [assert methods](https://docs.python.org/3.3/library/unittest.html#assert-methods).
+Next, add some simple (and speedy!) tests of your main features. If these run without exceptions that's a good start! Next, check the output of your methods using  [assert statement](https://docs.pytest.org/en/7.1.x/how-to/assert.html).
```
agriyakhetarpal, line 176:
```suggestion
Next, add some simple (and speedy!) tests of your main features. If these run without exceptions that's a good start! Next, check the output of your methods using [assert statements](https://docs.pytest.org/en/7.1.x/how-to/assert.html).
```

pyproject.toml
```diff
@@ -228,20 +228,21 @@ ignore = [
...
 ]
 addopts = [
     "-nauto",
-    "-v",
-    "-ra",
+    "-vrA",
```
agriyakhetarpal, line 240:
```suggestion
    "-vra",
```

I figured out the problem with the extra lines of logging output in the tests for the example scripts – this slipped right under our noses. This should be `-a`, not `-A` – the latter prints out the outputs from the tests that passed, and not just their names. Using `-a` should reduce the length of the logs by a significant amount.

agriyakhetarpal (approved):
Thanks for your work here, @prady0t! This looks great now – just a couple of comments, and this is ready to go.

tests/test_scripts.py
agriyakhetarpal: This should be resolved by https://github.com/pybamm-team/PyBaMM/pull/4180#discussion_r1693373168.

agriyakhetarpal:
This will need approval from @arjxn-py since he previously requested changes, we can merge after that

arjxn-py (approved):
This looks good, thanks @prady0t.
Will just wait for CI to pass to observe logs once, then we can merge.

agriyakhetarpal:
No unnecessary logs this time around: https://github.com/pybamm-team/PyBaMM/actions/runs/10115401416/job/27976388229

arjxn-py:
> No unnecessary logs this time around: https://github.com/pybamm-team/PyBaMM/actions/runs/10115401416/job/27976388229

Noticed that, just that we'd need to update branch to merge. My bad, I should have done it earlier.

PR #4184, opened 2024-06-16
[WIP] Improve "Creating Models" series

### LINKED ISSUE #3844: Restructure "Creating Models" notebooks
After reviewing the "Getting Started" notebooks, the next step is to restructure the "Creating Models" ones. The idea is to have a cohesive set of notebooks that build progressively more complex models up to DFN (potentially with degradation). The inspiration is the ["12 steps to Navier Stokes" tutorial](https://lorenabarba.com/blog/cfd-python-12-steps-to-navier-stokes/).

We will not have 12 steps, at least initially, but so far the structure would be:
1. Reservoir model (ODEs)
2. SPM (parabolic PDEs)
3. SPMe (parabolic PDEs in multiple domains, multiscale formulation appears)
4. DFN (parabolic & elliptic PDEs in multiple domains)
5. DFN + degradation (maybe SEI? Includes all of the above, but more complicated)

This is subject to change, as the notebooks are being edited/written (e.g. some might need to be split into two).

This issue would go hand-in-hand with this: https://github.com/pybamm-team/PyBaMM/issues/3839#issuecomment-1966614301

brosaplanella:
# Description

Refactor and improve the Creating Models series notebooks. I have significantly refactored the existing notebooks, but the underpinning philosophy should be similar. The first few notebooks follow very closely the material from the training session in Oxford (Apr 24) which worked quite well.

Fixes #3844

## List of notebooks
1. Creating a simple ODE model: demonstrate the basics of PyBaMM models and illustrate with a simple ODE model. I have used the reservoir model to stick to battery models. I solved the model using the simulation class.
2. Creating a simple PDE model: extends the previous model to solving a PDE model, in this case a simplified version of SPM for a half cell to continue with the battery-flavoured models. I solved the model using the simulation class.
3. Wrapping a model into a class: takes the model from notebook 2 and wraps it into a class. 

## Other additions
- `BasicReservoir` model

## Type of change

# Key checklist:

## Further checks:


valentinsulzer:
I think it's ok to leave out explanations of the simulation class in this series, and just stick to the models

brosaplanella:
The PR is not yet ready, but I have hit a point where I would really benefit from some feedback before continuing. I believe that the 3 notebooks I have written introduce all the main features needed to write models except for Broadcasts. I was originally planning on introducing broadcasts in later notebooks, but they would fit well into notebook 2.

Most of the questions I have on how to proceed are related to the model refactoring area #3908. I think everything revolves around the following question: how do we envision users will write their models? Will they write everything from scratch inheriting from the `pybamm.BaseModel` template, or will they use the features in `pybamm.lithium_ion.BaseModel` class? If the latter, this could be demonstrated in additional notebooks (e.g. the one I started writing but haven't pushed yet is how to build SPM, and makes extensive use of the lithium ion parameter class). But if the former, any additional notebooks will focus solely on the battery modelling side (happy to do so if that's useful).

Another thing to include, either in notebook 3 or future notebooks, is how to get a Basic model to work with experiments.

## TLDR
The three models so far document all the main features someone would need to know to write their own models. There are other features that we could show, but the usefulness of this depends on how we expect users to write models.

## Additional questions
Here is a list of additional questions I have. They might open several cans of worms so please only spend time on them after you have given your thoughts on the main points above.
- If we choose to show how to use the lithium ion parameters class, how do we document them? At the moment the syntax is not documented anywhere.
- How do we want to show models in notebooks? Classes need to be defined in a single cell which gets really verbose.
- Some Basic models use slighlty different notation (e.g. some parameters are defined as `pybamm.Parameter`, some call the `LithiumIonParameters` class). Is there any reason for this or should I unify them?
- Submodels: I was planning on writing a final notebook on submodels, but I believe it is better to hold this until we have implemented the fix suggested by Valentin in https://github.com/pybamm-team/PyBaMM/issues/3908#issuecomment-2153307269 (unless we don't see this happening any time soon).

PR #4192, opened 2024-06-17
Added contact resistance to model.

### LINKED ISSUE #4140: Add contact resistance heat source
Add a heat source due to the parameter "Contact resistance [Ohm]"

parkec3:
# Description

Added a term for the contact resistance within heating. Still needs to be tested for functionality as well as the typical tests. The contact resistance parameter within pybamm is [ohm] instead of [ohm m2], so I left off the surface area when calculating the heating term. The volume parameter also needs to be checked to ensure it's the correct volume calculation.

Fixes #4140

## Type of change

# Key checklist:

## Further checks:


pybamm/models/submodels/thermal/base_thermal.py
```diff
@@ -175,15 +175,25 @@ def _get_standard_coupled_variables(self, variables):
             Q_rev_n, pybamm.FullBroadcast(0, "separator", "current collector"), Q_rev_p
         )
 
+        # Contact resistance heating Q_cr
+        # TODO check that the volume is the correct volume for contact resistance
+        if self.options["contact resistance"] == "true":
+                I = variables["Current [A]"]
+                V = param.V_cell
+                Q_cr = I**2 * param.R_contact / V
+            else:
```
valentinsulzer, line 184: `else` is not aligned with the `if` statement so the tests won't run

valentinsulzer: you can install the "ruff" linter in vscode to highlight things like this

pybamm/models/submodels/thermal/base_thermal.py
```diff
@@ -216,6 +229,7 @@ def _get_standard_coupled_variables(self, variables):
         Q_ohm_vol_av = Q_ohm_W / V
         Q_rxn_vol_av = Q_rxn_W / V
         Q_rev_vol_av = Q_rev_W / V
+        Q_cr_vol_av = Q_cr_W / V
```
valentinsulzer, line 232: I think this term should be `I^2R/V` and we need to reverse engineer the previous terms to get to that. i.e. `Q_cr = I**2 / (n_elec * A) * R_contact`. Need input from @rtimms @brosaplanella @TomTranter about how to think about contact resistance heating in "volume-averaged" sense

valentinsulzer (changes requested):

pybamm/models/submodels/thermal/base_thermal.py
agriyakhetarpal: another way is to do `pip install pre-commit` and then `pre-commit install` which will block changes that don't pass from being committed

parkec3: Thanks. I updated the code to have the proper syntax.

pybamm/models/submodels/thermal/base_thermal.py
rtimms: I'm not sure the contact resistance heating makes sense outside of the lumped model - it's not obvious "where" the heating is taking place. For higher-dimensional models, it's certainly not the case that the contact resistance heating should appear as a uniform source everywhere in the cell.

valentinsulzer (changes requested):
Code looks good but needs tests. The following tests would be good:
- a unit test in `base_lithium_ion_tests` that checks contact resistance + lumped thermal
- an integration test that runs a lumped thermal model with and without contact resistance heating and makes sure that the temperature is always larger when there is contact resistance heating

rtimms:
It would also be good to update [this notebook](https://github.com/pybamm-team/PyBaMM/blob/develop/docs/source/examples/notebooks/models/thermal-models.ipynb) to include contact resistance heating in the lumped model. We should also update the docs of the thermal models to be more explicit about which heating terms are included.

tests/integration/test_models/test_full_battery_models/test_lithium_ion/test_thermal_models.py
```diff
@@ -103,6 +103,56 @@ def err(a, b):
...
+        # get the average temperature from each model
+        avg_cell_temp = sols[0]["X-averaged cell temperature [K]"].entries
+        avg_cell_temp_cr = sols[1]["X-averaged cell temperature [K]"].entries
+
+        # check that the time-averaged cell temperature of the lumped thermal model
+        # with contact resistance is higher than without contact resistance
+        self.assertGreater(
+            sum(avg_cell_temp_cr) / len(avg_cell_temp_cr),
+            sum(avg_cell_temp) / len(avg_cell_temp),
+        )
```
valentinsulzer, line 154: you can also use `np.testing.assert_array_less` to check the whole array

kratman (approved):

PR #4199, opened 2024-06-19
Add support for MLIR-based expression evaluation

### LINKED ISSUE #3826: Add support for MLIR-based expression evaluation
### Description

Provide an MLIR-based backend for expression evaluation that can be utilised by SUNDIALS for either CPU or GPU-accelerated solves.

### Motivation

At present the PyBaMM IDAKLU solver makes use of SUNDIALS which provides [support for GPU](https://sundials.readthedocs.io/en/latest/sundials/GPU_link.html), but this support cannot be harnessed as IDAKLUs method of function evaluation relies on casadi, which would require user-side compilation of PyBaMM models. Of PyBaMMs solvers, currently only JAX (a python library) translates naturally to GPU. By providing an MLIR-based backend, the code required for expression evaluation can be lowered to CPU or GPU devices, allowing expression-trees to be passed for evaluation at runtime. There have been previous efforts to expose GPU support for SUNDIALS in PyBaMM [e.g. #2644 ], which stalled as the model equations could not be easily lowered to device (GPU) code.

### Possible Implementation

[MLIR](https://mlir.llvm.org/) is a multi-level process of lowering into [LLVM](https://llvm.org/)'s Intermediate Representation (IR), which can then be cross-compiled onto multiple platforms, including GPU. This is also the (broad) mechanism for how JAX supports multiple platforms.

### Additional context

https://github.com/pybamm-team/PyBaMM/issues/3766#issuecomment-1910627686

jsbrittain:
# Description

Add a new expression evaluation backend to the IDAKLU solver. MLIR expression evaluation is now supported by lowering PyBaMM's Jax-based expressions into [MLIR](https://mlir.llvm.org/), which are then compiled and executed as part of the IDAKLU solver using [IREE](https://iree.dev/).

To enable the IREE/MLIR backend, set the (new) `PYBAMM_IDAKLU_EXPR_IREE` compiler flag `ON` via an environment variable and install PyBaMM using the developer method (by default `PYBAMM_IDAKLU_EXPR_IREE` is turned `OFF`):
```bash
export PYBAMM_IDAKLU_EXPR_IREE=ON
nox -e pybamm-requires && nox -e dev
```

Expression evaluation in IDAKLU is enabled by constructing the model using Jax expressions (`model.convert_to_format="jax"`) and setting the solver backend (`jax_evaluator="iree"`). Example:
```python
import pybamm
import numpy as np

model.convert_to_format = "jax"
param.process_model(model)
param.process_geometry(model.default_geometry)
disc.process_model(model)

solver = pybamm.IDAKLUSolver(
	root_method="hybr",  # change from default ("casadi")
	options={"jax_evaluator": "iree"}
)
solution = solver.solve(model, np.linspace(0, 3600, 2500))

print(solution["Voltage [V]"].entries[:100])
```

**Note** that IREE currently only supports single-precision floating-point operations, which requires the model to be demoted from 64-bit to 32-bit precision before the solver can run. This is handled within the solver logic, but the operation is performed *in-place* on the PyBaMM battery model (we display a warning when run). Operating at lower precision requires tolerances to be relaxed for convergence on larger [e.g. DFN] models, and leads to memory transfers and type casting in the solver which are currently causing slow-downs (at least until 64-bit computation is natively supported).

Comparative performance on the above SPM problem on an Apple M2 Macbook Pro (with events=[] to allow comparison to the JaxSolver):
- IDAKLU-IREE took 1.4 secs (1.3 secs to demote and compile the expressions; <0.1 secs for each subsequent solve).
- IDAKLU-Casadi: took 0.14 secs (<0.1 secs setup; <0.1 secs for each subsequent solve).
- JaxSolver [BDF]: took 1.0 secs (0.9 secs compilation; 0.1 secs for each subsequent solve).

Substituting a DFN model (and reducing `atol = 1e-1`) the times become:
- IDAKLU-IREE took 12.1 secs (8.7 secs to demote and compile the expressions; 3.4 secs for each subsequent solve).
- IDAKLU-Casadi: took 0.6 secs (0.3 secs setup; 0.3 secs for each subsequent solve).
- JaxSolver [BDF]: took 22.8 secs (7.6 secs compilation; 15.2 secs for each subsequent solve).

There is a noticeable performance deficit for the IDAKLU-MLIR solver compared to Casadi, due to 1) initial compilation of MLIR to bytecode, 2) demotion strategies, and 3) memory transfers casting between types in the solver. We anticipate improvements in the second and third points with native 64-bit IREE support, and as our IREE approach compiles on the model expressions (not the solver) compilation times quickly out-perform the JaxSolver with increasing model complexity / time steps (while also taking full advantage of the capabilities already provided by the IDAKLU solver, such as events). The IREE/MLIR approach offers a pathway to compiling expressions across a wide variety of backends, including metal and cuda, although additional code adjustment (principally host/device transfers) will be required before those can be supported.

Resolves #3826

## Type of change

# Key checklist:

## Further checks:


jsbrittain:
_Note:_ Codacy seems to be struggling with the template / inheritance structures in C++, hence the inclusion of additional `// cppcheck-suppress` comments in the IDAKLU solver.

noxfile.py
```diff
@@ -13,11 +15,54 @@
...
+    """
+    state = "ON" if os.getenv("PYBAMM_IDAKLU_EXPR_IREE", "OFF") == "ON" else "OFF"
+    if state == "ON":
+        if sys.platform == "win32":
+            warnings.warn(
+                (
+                    "IREE is not enabled on Windows yet. "
+                    "Setting PYBAMM_IDAKLU_EXPR_IREE=OFF."
+                ),
+                stacklevel=2,
```
agriyakhetarpal, line 36: Hi @jsbrittain, I see that IREE is available as a wheel on x86_64 Windows as well now – based on https://pypi.org/project/iree-compiler/#files?

Or is this because the `IDAKLU_EXPR_IREE` files do not compile on Windows yet?

noxfile.py
```diff
@@ -40,7 +85,11 @@ def set_environment_variables(env_dict, session):
 
 @nox.session(name="pybamm-requires")
 def run_pybamm_requires(session):
-    """Download, compile, and install the build-time requirements for Linux and macOS. Supports --install-dir for custom installation paths and --force to force installation."""
+    """Download, compile, and install the build-time requirements for Linux and macOS.
+
+    Supports --install-dir for custom installation paths and --force to force
+    installation.
+    """
```
agriyakhetarpal, line 92: `nox` has an annoyance (https://github.com/wntrblm/nox/issues/694) where `nox --list` prints out just the first line of the docstring when trying to list the details of all sessions. Could we please keep this in a single line?

noxfile.py
```diff
@@ -59,6 +108,30 @@ def run_pybamm_requires(session):
...
+            session.chdir("iree")
+            session.run(
+                "git",
+                "submodule",
+                "update",
+                "--init",
+                "--recursive",
+                external=True,
+            )
+            session.chdir("..")
```
agriyakhetarpal, line 134: `session.chdir` can be used as a context manager, so something like
```suggestion
            with session.chdir("iree"):
                session.run(
                   "git",
                    "submodule",
                    "update",
                    "--init",
                    "--recursive",
                    external=True,
                )
```

should also work

pyproject.toml
```diff
@@ -114,11 +114,17 @@ dev = [
...
+# Note: These must be kept in sync with the versions defined in pybamm/util.py, and
+#       must remain compatible with IREE (see noxfile.py for IREE compatibility).
 jax = [
     "jax==0.4.27",
     "jaxlib==0.4.27",
 ]
+# For MLIR expression evaluation (IDAKLU Solver)
+iree = [
+    "iree-compiler==20240507.886",  # see IREE compatibility notes in noxfile.py
+]
```
agriyakhetarpal, line 127: Please add a note about this in the table of dependencies in the docs: https://docs.pybamm.org/en/latest/source/user_guide/installation/index.html and additionally in the Optional solvers section as well

pyproject.toml
```diff
@@ -114,11 +114,17 @@ dev = [
...
 # Contains all optional dependencies, except for jax and dev dependencies
```
agriyakhetarpal, line 128:
```suggestion
# Contains all optional dependencies, except for jax, iree, and dev dependencies
```

agriyakhetarpal (commented):
ccing @cringeyburger here because this PR will be adding and modifying a lot of compiled code, we shall need to make a few adjustments in the migration to `scikit-build-core` as needed. Though, as long as the wheels builds pass (@jsbrittain, could you please trigger them on your fork?), it should be fine.

jsbrittain:
> ccing @cringeyburger here because this PR will be adding and modifying a lot of compiled code, we shall need to make a few adjustments in the migration to `scikit-build-core` as needed. Though, as long as the wheels builds pass (@jsbrittain, could you please trigger them on your fork?), it should be fine.

@agriyakhetarpal Wheels build fine: https://github.com/jsbrittain/PyBaMM/actions/runs/9665563143

pybamm/expression_tree/functions.py
```diff
@@ -61,7 +61,7 @@ def __str__(self):
         out = f"{self.name[10:-1]}("
         for child in self.children:
             out += f"{child!s}, "
-        out = out[:-2] + ")"
+        out = out[:-2] + ")"  # pragma: no cover
```
martinjrobins, line 64: why only the one line, should this pragma be on the entire function?

pybamm/expression_tree/operations/evaluate_python.py
```diff
@@ -229,7 +229,7 @@ def find_symbols(
                         f"{children_vars[0]}.scalar_multiply(1/{children_vars[1]})"
                     )
                 else:
-                    symbol_str = f"{children_vars[0]}.multiply(1/{children_vars[1]})"
+                    symbol_str = f"{children_vars[0]}.multiply(1/{children_vars[1]})"  # pragma: no cover
```
martinjrobins, line 232: I think this would need a test to cover it.

pybamm/expression_tree/operations/evaluate_python.py
```diff
@@ -310,7 +310,7 @@ def find_symbols(
 
         elif isinstance(symbol, pybamm.SparseStack):
             if len(children_vars) == 1:
-                symbol_str = children_vars[0]
+                symbol_str = children_vars[0]  # pragma: no cover
```
martinjrobins, line 313: again a test would be good for this. if you don't want to do it in this PR (understandable) then an issue for reminder would be good too.

martinjrobins: more below which I won't mention. Looks like we were not covering a lot of it :(

pybamm/solvers/c_solvers/idaklu.cpp
```diff
@@ -79,13 +85,41 @@ PYBIND11_MODULE(idaklu, m)
...
+    py::arg("var_fcns"),
     py::arg("dvar_dy_fcns"),
     py::arg("dvar_dp_fcns"),
     py::arg("options"),
     py::return_value_policy::take_ownership);
 
-  m.def("generate_function", &generate_function,
+#ifdef IREE_ENABLE
+  m.def("create_iree_solver", &create_idaklu_solver<IREEFunctions>,
+    "Create a casadi idaklu solver object",
```
martinjrobins, line 96:
```suggestion
    "Create a iree idaklu solver object",
```

pybamm/solvers/c_solvers/idaklu/Expressions/Base/Expression.hpp
```diff
@@ -0,0 +1,69 @@
...
+   */
+  Expression() = default;
+
+  /**
+   * @brief Evaluation operator
+   */
+  virtual void operator()() = 0;
+
+  /**
+   * @brief Evaluation operator given data vectors
```
martinjrobins, line 23: can you please expand on the descriptions of the inputs/outputs? You mention they are vectors, but then later functions indicate that the output can be matrices?

pybamm/solvers/c_solvers/idaklu/Expressions/IREE/IREEFunctions.cpp
```diff
@@ -0,0 +1,303 @@
...
+  const std::string& mlir = f.get_mlir();
+
+  // Parse IREE (MLIR) function string
+  if (mlir.size() == 0) {
+    DEBUG("Empty function --- skipping...");
+    return;
+  }
+
+  // Parse module name
+  std::regex module_name_regex("module @([^\\s]+)");  // Match until first whitespace
```
martinjrobins, line 21: does iree have a parser class that can you use to do this? (e.g. https://mlir.llvm.org/doxygen/classmlir_1_1detail_1_1Parser.html). This seems quite manual

pybamm/solvers/c_solvers/idaklu/Expressions/IREE/IREEFunctions.cpp
```diff
@@ -0,0 +1,303 @@
...
+      count *= input_shape[j][k];
+    }
+    numel[j] = count;
+    input_data[j].resize(numel[j]);
+  }
+
+  // Allocate memory for input arguments
+  m_arg.resize(m_func.n_args, nullptr);
+
+  // Size iree results vector (single precision) and idaklu results vector (double precision)
```
martinjrobins, line 180:
```suggestion
  // Size iree results vector (single precision) and casadi results vector (double precision)
```

pybamm/solvers/c_solvers/idaklu/IDAKLUSolverOpenMP.hpp
```diff
@@ -40,7 +36,8 @@ using Function = casadi::Function;
  *   19. Destroy objects
  *   20. (N/A) Finalize MPI
  */
-class CasadiSolverOpenMP : public CasadiSolver
+template <class CExprSet>
```
martinjrobins, line 39: is the `C` here for casadi?

martinjrobins (changes requested):
thanks @jsbrittain this looks excellent. I've made a few suggestions below, see what you think. also, is `jax_evaluator="iree"` needed for the options, it there a case where you want to convert to jax but not use the jax_evaluator?

pybamm/expression_tree/functions.py
jsbrittain: Codacy doesn't like this _one_ line, but passes when it is excluded from cover! I will try to dig out the logs...

jsbrittain: Here it is: https://app.codecov.io/gh/pybamm-team/PyBaMM/commit/7bce94ed32246c1e77fd1f9eec7b2ded431410c8/blob/pybamm/expression_tree/functions.py#L70

jsbrittain:
> thanks @jsbrittain this looks excellent. I've made a few suggestions below, see what you think. also, is `jax_evaluator="iree"` needed for the options, it there a case where you want to convert to jax but not use the jax_evaluator?

@martinjrobins yes, there is actually an existing python-idaklu interface that will run if we don't redirect using the (new) `jax_evaluator` option. I think it's a legacy item (`idaklu/python.cpp`) (it can be quite slow, even on these toy examples).

pybamm/expression_tree/operations/evaluate_python.py
jsbrittain: Quite a few of these were introduced as 'indirect' changes during one of the commits. I will go back and see whether these can be relaxed again, or if new tests are needed...

pybamm/expression_tree/functions.py
martinjrobins: ha, that is a weird one! ok then lets leave it in then

martinjrobins:
that reminds me, we should get rid of python-idaklu, I don't think it serves a useful purpose anymore. I'll add an issue

pybamm/solvers/c_solvers/idaklu/Expressions/Base/Expression.hpp
jsbrittain: This refers to a 'vector' of arguments (input data), where each argument can be of arbitrary size (although in reality this is limited within the solver). I have rephrased the docstrings to `@brief Evaluation operator (for use after setting input and output data references)` and `Evaluation operator (supplying data references)` as-well as clarifying `get_rows()` and `get_cols()` to e.g. `Returns row indices in COO format (where the output data represents sparse matrix elements)`. Note that the spartsity pattern is computed in Python before calling the solver and is assumed fixed.

noxfile.py
jsbrittain: IREE does officially support Windows. However, when trying to build for the platform I ran into compile errors in the IREE codebase. Specifically, these were due to the use of 'compound literals' which are not part of the C++ standard and are not supported by MSVC (C4576). Since IREE is based on the MLIR/LLVM codebase it is probably more compatible with clang++ (which does provide this support, but is not our chosen toolchain). So this is going to need some further consideration. @agriyakhetarpal @martinjrobins

pybamm/solvers/c_solvers/idaklu/Expressions/IREE/IREEFunctions.cpp
jsbrittain: This is proving surprisingly difficult to implement. I can see two options:
1) working with IREE: the routines provided only seem to permit access to low-level memory details, using representations below the `tensor` syntax, which makes extraction difficult.
2) working with llvm-mlir: we have to compile-in the llvm/mlir libraries alongside iree; set-up new contexts and register dialects in order to parse the mlir string. This approach has potential but is so far proving non-trivial to get working in harmony with the iree solution we already have in place.

As this is turning into a bit of a blocker, I have decided to refactor the existing module parsing code out into it's own ModuleParser class that can be modified or wholesale replaced as needed at a later date.

agriyakhetarpal:
Hi again, @jsbrittain – https://github.com/pybamm-team/PyBaMM/pull/4205 is trying to migrate PyBaMM's package structure to an `src` layout from the current flat one, which will move all the files from `pybamm/` into `src/pybamm/`. It has a lot of potential for delays because of inevitable merge conflicts across several PRs, so we plan to merge it as soon as stable v24.5 hits next week, just a heads up :)

pybamm/expression_tree/operations/evaluate_python.py
jsbrittain: I've gone through and removed all of the 'no cover' statements that I inserted due to indirect changes in previous revisions. Codecov seems to have settled itself down now and is passing everything again. So we should be good.

pybamm/solvers/c_solvers/idaklu/Expressions/IREE/IREEFunctions.cpp
martinjrobins: fair enough. Happy to go with the separate parsing class for now.

martinjrobins (approved):
Thanks @jsbrittain for the changes, all looks good to me now. I'm happy to merge, but just need to check with #4249 how we're going to put these two PRs in.

PR #4208, opened 2024-06-24
#4120 improved error message

### LINKED ISSUE #4120: Better error on unrecognised parameter set
Doing `pybamm.ParameterValues("bad")` gives `ValueError: Invalid Parameter Value`. It would be better to say something like `ValueError: 'bad' is not a recognised parameter set` and possibly list the valid options (including custom parameter sets that have been registered as entry points).

Ubham16:
# Description

Added a more informative error message when I passed an invalid parameter set, which rejects that parameter set and relays a list of potential paramters that could be inputted

Fixes #4120

## Type of change

# Key checklist:

## Further checks:


arjxn-py (approved):
Looks good, thanks @Ubham16

pybamm/parameters/parameter_values.py
```diff
@@ -65,7 +65,10 @@ def __init__(self, values, chemistry=None):
                 values.pop("chemistry", None)
                 self.update(values, check_already_exists=False)
             else:
-                raise ValueError("Invalid Parameter Value")
+                valid_sets = "\n".join(pybamm.parameter_sets.keys())
+                raise ValueError(
+                    f"{values} is not a valid parameter set. Parameter set must be one of:\n {valid_sets} "
```
rtimms, line 70:
```suggestion
                    f"{values} is not a valid parameter set. Parameter set must be one of:\n{valid_sets} "
```

rtimms (changes requested):
Looks good, just need to fix the broken test. Thanks!

pybamm/parameters/parameter_values.py
kratman: There should also be single quotes or something around `{values}` to make the error message nicer:

For example the test is: "Junk is not a valid parameter set."
It would probably be nicer to have "'Junk' is not a valid parameter set."

kratman: Just an update to Rob's suggestion. I removed the trailing space and added the quotes. I also updated the test to match
```suggestion
                    f"'{values}' is not a valid parameter set. Parameter set must be one of:\n{valid_sets}"
```

pybamm/parameters/parameter_values.py
```diff
@@ -65,7 +65,10 @@ def __init__(self, values, chemistry=None):
                 values.pop("chemistry", None)
                 self.update(values, check_already_exists=False)
             else:
-                raise ValueError("Invalid Parameter Value")
+                valid_sets = "\n".join(pybamm.parameter_sets.keys())
+                raise ValueError(
+                    f"{values} is not a valid parameter set. Parameter set must be one of:\n{valid_sets} "
```
kratman, line 70:
```suggestion
                    f"'{values}' is not a valid parameter set. Parameter set must be one of:\n{valid_sets}"
```

agriyakhetarpal (approved):

rtimms:
@all-contributors please add @Ubham16 for code

rtimms (approved):

PR #4214, opened 2024-06-25
Moving `test_util.py` to pytest

prady0t:
# Description

Refactored `test_util.py` from unittest style test to pytest.

Part of #4156

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -107,10 +107,11 @@ dev = [
     "pytest-doctestplus",
     # For test parameterization
     "parameterized>=0.9",
-    # For testing Jupyter notebooks
+    # pytest and its plugins
     "pytest>=6",
     "pytest-xdist",
     "pytest-mock",
+    # For testing Jupyter notebooks
     "nbmake",
     # To access the metadata for python packages
     "importlib-metadata; python_version < '3.10'",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -218,11 +218,8 @@ ignore = [
 "**.ipynb" = ["E402", "E703"]
 "docs/source/examples/notebooks/models/lithium-plating.ipynb" = ["F821"]
 
-# NOTE: currently used only for notebook tests with the nbmake plugin.
 [tool.pytest.ini_options]
 minversion = "6"
-# Use pytest-xdist to run tests in parallel by default, exit with
-# error if not installed
 required_plugins = [
     "pytest-xdist",
     "pytest-mock",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -121,7 +121,6 @@ jax = [
 ]
 # Contains all optional dependencies, except for jax and dev dependencies
 all = [
-    "autograd>=1.6.2",
     "scikit-fem>=8.1.0",
     "pybamm[examples,plot,cite,bpx,tqdm]",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -110,6 +110,7 @@ dev = [
     # For testing Jupyter notebooks
     "pytest>=6",
     "pytest-xdist",
+    "pytest-mock",
     "nbmake",
     # To access the metadata for python packages
     "importlib-metadata; python_version < '3.10'",
@@ -224,6 +225,7 @@ minversion = "6"
 # error if not installed
 required_plugins = [
     "pytest-xdist",
+    "pytest-mock",
 ]
 addopts = [
     "-nauto",
--- a/tests/unit/test_util.py
+++ b/tests/unit/test_util.py
@@ -176,9 +176,3 @@ def test_url_gets_to_stdout(mocker):
         assert fake_out.getvalue() == "test\t10\n"
 
 
-if __name__ == "__main__":
-    print("Add -v for more debug output")
-
-    if "-v" in sys.argv:
-        debug = True
-    pybamm.settings.debug_mode = True
--- a/tests/unit/test_util.py
+++ b/tests/unit/test_util.py
@@ -13,175 +13,170 @@ from tests import (
 )
 
 
-def test_is_constant_and_can_evaluate():
-    symbol = pybamm.PrimaryBroadcast(0, "negative electrode")
-    assert not pybamm.is_constant_and_can_evaluate(symbol)
-    symbol = pybamm.StateVector(slice(0, 1))
-    assert not pybamm.is_constant_and_can_evaluate(symbol)
-    symbol = pybamm.Scalar(0)
-    assert pybamm.is_constant_and_can_evaluate(symbol)
+class TestUtil():
+    """
+    Test the functionality in util.py
+    """
 
+    def test_is_constant_and_can_evaluate(self):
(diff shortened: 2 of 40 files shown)


pyproject.toml
```diff
@@ -110,6 +110,7 @@ dev = [
     # For testing Jupyter notebooks
```
agriyakhetarpal, line 110: Could you move this comment to be above `nbmake`?


```suggestion
    # pytest and its plugins
```

pyproject.toml
```diff
@@ -110,6 +110,7 @@ dev = [
...
     "pytest>=6",
     "pytest-xdist",
+    "pytest-mock",
```
agriyakhetarpal, line 113:
```suggestion
    "pytest-mock",
    # For testing Jupyter notebooks
```

pyproject.toml
```diff
@@ -224,6 +225,7 @@ minversion = "6"
 # error if not installed
 required_plugins = [
```
agriyakhetarpal, line 226: Now that we use `pytest` virtually everywhere, could you remove the comments on lines 221 and 224–225?

tests/unit/test_util.py
```diff
@@ -15,180 +13,178 @@
...
-        with patch("sys.stdout", new=StringIO()) as fake_out:
-            param.search("test")
-            self.assertEqual(fake_out.getvalue(), "test\t10\n")
+    # Test param search (default returns key, value)
+    with mocker.patch("sys.stdout", new=StringIO()) as fake_out:
+        param.search("test")
+        assert fake_out.getvalue() == "test\t10\n"
 
 
 if __name__ == "__main__":
```
agriyakhetarpal, line 190: Do we still need this block for being able to run the tests, or can `pytest` run individual files without it? Also, we run `pytest` with `-v` by default, so we probably don't need to check `sys.argv` either.

tests/unit/test_util.py
```diff
@@ -197,4 +193,3 @@ def test_url_gets_to_stdout(self):
     if "-v" in sys.argv:
         debug = True
     pybamm.settings.debug_mode = True
```
agriyakhetarpal, line 195: Ideally, we should set this `pybamm.settings.debug_mode = True` by default when running the tests, perhaps in `conftest.py` as a fixture (with `autouse` set to `True`) in your other PR.

tests/unit/test_util.py
prady0t: Yes we can safely remove this block.

tests/unit/test_util.py
```diff
@@ -15,186 +13,175 @@
...
-                "test": 1,
-                "test2": 2,
-                "SEI current": 3,
-                "Lithium plating current": 4,
-                "A dimensional variable [m]": 5,
-                "Positive electrode diffusivity [m2.s-1]": 6,
-            }
+def test_is_constant_and_can_evaluate():
+    symbol = pybamm.PrimaryBroadcast(0, "negative electrode")
+    assert False is pybamm.is_constant_and_can_evaluate(symbol)
```
arjxn-py, line 18:
```suggestion
    assert not pybamm.is_constant_and_can_evaluate(symbol)
```
I think this should be same but a bit simplified, make sure to test once tho.

tests/unit/test_util.py
```diff
@@ -15,186 +13,175 @@
...
+    symbol = pybamm.StateVector(slice(0, 1))
+    assert False is pybamm.is_constant_and_can_evaluate(symbol)
```
arjxn-py, line 20:
```suggestion
    assert not pybamm.is_constant_and_can_evaluate(symbol)
```
Ditto

arjxn-py (commented):
Thanks @prady0t, looks good. Just a vey minor comment

tests/unit/test_util.py
```diff
@@ -15,186 +13,175 @@
...
-
-        with self.assertRaisesRegex(KeyError, "open circuit voltage"):
-            d.__getitem__("Measured open circuit voltage [V]")
-
-        with self.assertRaisesRegex(KeyError, "Lower voltage"):
-            d.__getitem__("Open-circuit voltage at 0% SOC [V]")
+def test_get_parameters_filepath():
+    tempfile_obj = tempfile.NamedTemporaryFile("w", dir=".")
+    assert pybamm.get_parameters_filepath(tempfile_obj.name) == tempfile_obj.name
+    tempfile_obj.close()
```
agriyakhetarpal, line 65: Can we use a context manager to handle the temporary files? That way, we won't have to close the files every time

tests/unit/test_util.py
```diff
@@ -15,186 +13,175 @@
...
+
+    package_dir = os.path.join(pybamm.root_dir(), "pybamm")
+    tempfile_obj = tempfile.NamedTemporaryFile("w", dir=package_dir)
+    path = os.path.join(package_dir, tempfile_obj.name)
+    assert pybamm.get_parameters_filepath(tempfile_obj.name) == path
+    tempfile_obj.close()
```
agriyakhetarpal, line 71: Same here

tests/unit/test_util.py
```diff
@@ -15,186 +13,175 @@
...
+
+
+def test_is_jax_compatible():
+    if pybamm.have_jax():
+        compatible = pybamm.is_jax_compatible()
+        assert compatible
```
agriyakhetarpal, line 77: Instead of having a conditional statement inside the test, you can move it up and use a decorator, i.e., `@pytest.mark.skipIf(pybamm.have_jax(), reason="The JAX solver is not installed")`

tests/unit/test_util.py
```diff
@@ -15,186 +13,175 @@
...
-        optional_distribution_deps = get_optional_distribution_deps("pybamm")
-        present_optional_import_deps = get_present_optional_import_deps(
-            "pybamm", optional_distribution_deps=optional_distribution_deps
+    # Check that optional dependencies are not present in the core PyBaMM installation
+    optional_present_deps = optional_distribution_deps & required_distribution_deps
+    assert not bool(optional_present_deps), (
+        f"Optional dependencies installed: {optional_present_deps}.\n"
+        "Please ensure that optional dependencies are not present in the core PyBaMM installation, "
+        "or list them as required."
+    )
```
agriyakhetarpal, line 160: I don't think this is correct. While it may have been passing, `&` is used for bitwise operations, the `and` keyword is better to use here, i.e., for boolean operations. Also, you should cast `optional_present_deps` as boolean prior to the assert statement.

prady0t: I think `optional_distribution_deps` and `required_distribution_deps` here are sets and hence `&` checks for the common elements if found. Using `and` here leads to test failure.

prady0t:
I added the class structure back.

tests/unit/test_util.py
agriyakhetarpal: Ah, I see, I think we should fix that in a different PR. Thanks for looking!

agriyakhetarpal (approved):
Thanks, @prady0t! Adding the class structure back makes the diff smaller and the PR is nicer to review. The changes look good to me, I shall merge once CI passes.

PR #4215, opened 2024-06-25
V24.5

Saransh-cpp:
# Description

Release update for the develop branch


agriyakhetarpal:
Triggered another deployment after the last commit: https://github.com/pybamm-team/PyBaMM/actions/runs/9668643828, cancelling the tests here for now

agriyakhetarpal:
Oops, I updated the tag but I missed pushing the tag 😅 I don't think the workflow ran at all because it was stuck in the queue, I cancelled it just now. I think I will let you do it because I am about to log off

kratman:
> Oops, I updated the tag but I missed pushing the tag 😅 I don't think the workflow ran at all because it was stuck in the queue, I cancelled it just now. I think I will let you do it because I am about to log off

Yeah I already triggered the correct one. Hopefully you did not cancel mine

agriyakhetarpal:
I think I didn't, it's still running: https://github.com/pybamm-team/PyBaMM/actions/runs/9668726233

kratman:
Tests and deployment passed except for lychee

agriyakhetarpal (approved):

Saransh-cpp:
Merging this

PR #4219, opened 2024-06-26
Fix lychee issues from tinyurl

kratman:
# Description

Lychee was having issues with tinyurl, so this should avoid future issues.

## Type of change

# Key checklist:

## Further checks:


kratman:
I will mark this as ready for review after I test that the links still work in the docs

kratman:
While the tinyurl issues could be intermittent like some of the stack overflow ones, some of them also pointed to old documentation. So getting rid of the obfuscation is a plus

valentinsulzer (approved):

PR #4221, opened 2024-06-26
Bug fix for cracking options

### LINKED ISSUE #4119: Raise OptionError if "SEI on cracks" is "true" but cracking isn't enabled.
If you do e.g. 
```python3
model = pybamm.lithium_ion.SPMe(
    {
        "SEI": "solvent-diffusion limited",
        "lithium plating porosity change": "true",  # alias for "SEI porosity change"
        "particle mechanics": ("swelling only", "swelling only"),
        "SEI on cracks": "true",
        "loss of active material": "stress-driven",
    }
)
```
you get a `ZeroDivisionError`. It doesn't make sense to have SEI n cracks if there are no cracks, so we should catch this and raise an `OptionError`

pipliggins:
# Description

Adds an OptionError if the option `SEI on cracks` is True but `particle mechanics` does not contain cracks. This issue only occurs if both options are set explicitly in the model's additional options - the user's `particle mechanics` option overrides the defaults which automatically set the mechanics to "swelling and cracking" if `SEI on cracks` is true.

Fixes #4119

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/models/full_battery_models/base_battery_model.py
+++ b/pybamm/models/full_battery_models/base_battery_model.py
@@ -633,6 +633,18 @@ class BatteryModelOptions(pybamm.FuzzyDict):
                     "be 'none': 'particle mechanics', 'loss of active material'"
                 )
 
+        if "true" in options["SEI on cracks"]:
+            sei_on_cr = options["SEI on cracks"]
+            p_mechanics = options["particle mechanics"]
+            if any(
+                sei == "true" and mech != "swelling and cracking"
+                for mech, sei in zip(p_mechanics, sei_on_cr)
+            ):
+                raise pybamm.OptionError(
+                    "If 'SEI on cracks' is 'true' then 'particle mechanics' must be "
+                    "'swelling and cracking'."
+                )
+
         # Check options are valid
         for option, value in options.items():
             if isinstance(value, str) or option in [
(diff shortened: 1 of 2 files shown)


pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -633,6 +633,20 @@ def __init__(self, extra_options):
...
+            p_mechanics = options["particle mechanics"]
+            if isinstance(p_mechanics, str) and isinstance(sei_on_cr, tuple):
+                p_mechanics = (p_mechanics, p_mechanics)
+            if any(
+                sei == "true" and mech != "swelling and cracking"
+                for mech, sei in zip(p_mechanics, sei_on_cr)
+            ):
+                raise pybamm.OptionError(
+                    "If 'SEI on cracks' is 'true' then 'particle mechanics' must be "
+                    "'swelling and cracking'."
```
kratman, line 647: If swelling and cracking is required for "SEI on cracks", then can't we just enable swelling and cracking if the user wants SEI on cracks? That seems closer to the users intent

pipliggins: The pattern you suggest is already implemented if a user *only* provides `SEI on cracks` in the options list, and not `particle mechanics`. However, in a scenario such as 
```
model = pybamm.lithium_ion.SPMe(
    {
        "particle mechanics": ("swelling only", "swelling only"),
        "SEI on cracks": "true",
    }
)
```
where both options are supplied, the users `particle mechanics` choice will override the default (where `swelling and cracking` was previously enabled based on the `SEI on cracks` value).

In a case where the user has specified both options, it seems unhelpful to assume which option is of more importance to the user by automatically overriding their mechanics choice. Throwing an error indicating an incompatible combination of options feels more appropriate to me - particularly given that this is such an easy fix for the user.

kratman:
@rtimms Does this cover everything you wanted in #4119? The code looks fine to me, but I have not tested it myself

rtimms (approved):

Issue #4187, opened 2024-06-17
Allow functions of time in experiment step


valentinsulzer:
Users should be able to pass in arbitrary functions of time as the argument for a `pybamm.step` in an experiment, e.g.
```python
def oscillating(t):
    return 1 + 0.1*np.sin(2*np.pi*t)

pybamm.step.current(oscillating, duration=3600)
```
Time should be taken to start at the beginning of the step, just like with drive cycles.

Note: this can be already be done using custom steps, but it requires writing a more complex step object

PR #4222, opened 2024-06-26
Add functions of time in experiment step 

MarcBerliner:
# Description

Adds functionality to pass in arbitrary functions of time as the argument for a `pybamm.step`

Fixes https://github.com/pybamm-team/PyBaMM/issues/4187

## Type of change

# Key checklist:

- [ ] All tests pass: `$ python run-tests.py --all` (or `$ nox -s tests`) (I haven't gotten `idaklu` installed so I'm not certain all tests pass)

## Further checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Features
 
+- Added functionality to pass in arbitrary functions of time as the argument for a (`pybamm.step`). ([#4222](https://github.com/pybamm-team/PyBaMM/pull/4222))
 - Added new parameters `"f{pref]Initial inner SEI on cracks thickness [m]"` and `"f{pref]Initial outer SEI on cracks thickness [m]"`, instead of hardcoding these to `L_inner_0 / 10000` and `L_outer_0 / 10000`. ([#4168](https://github.com/pybamm-team/PyBaMM/pull/4168))
 - Added `pybamm.DataLoader` class to fetch data files from [pybamm-data](https://github.com/pybamm-team/pybamm-data/releases/tag/v1.0.0) and store it under local cache. ([#4098](https://github.com/pybamm-team/PyBaMM/pull/4098))
 - Added `time` as an option for `Experiment.termination`. Now allows solving up to a user-specified time while also allowing different cycles and steps in an experiment to be handled normally. ([#4073](https://github.com/pybamm-team/PyBaMM/pull/4073))
--- a/pybamm/experiment/step/base_step.py
+++ b/pybamm/experiment/step/base_step.py
@@ -105,27 +105,6 @@ class BaseStep:
             duration = self.default_duration(value)
         self.duration = _convert_time_to_seconds(duration)
 
-        # Record all the args for repr and hash
-        self.repr_args = f"{value}, duration={duration}"
-        self.hash_args = f"{value}"
-        if termination:
-            self.repr_args += f", termination={termination}"
-            self.hash_args += f", termination={termination}"
-        if period:
-            self.repr_args += f", period={period}"
-        if temperature:
-            self.repr_args += f", temperature={temperature}"
-            self.hash_args += f", temperature={temperature}"
-        if tags:
-            self.repr_args += f", tags={tags}"
-        if start_time:
-            self.repr_args += f", start_time={start_time}"
-        if description:
-            self.repr_args += f", description={description}"
-        if direction:
-            self.repr_args += f", direction={direction}"
-            self.hash_args += f", direction={direction}"
-
         # If drive cycle, repeat the drive cycle until the end of the experiment,
         # and create an interpolant
         if self.is_drive_cycle:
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Features
 
+- Added functionality to pass in arbitrary functions of time as the argument for a (`pybamm.step`). ([#4222](https://github.com/pybamm-team/PyBaMM/pull/4222))
```
valentinsulzer, line 5: This should be made into a new section as it won't be included in this release

pybamm/experiment/step/base_step.py
```diff
@@ -83,6 +84,21 @@ def __init__(
...
+            except TypeError:
+                raise TypeError(
+                    "Input function must have only 1 positional argument for time"
+                ) from None
+
+            # Check if the value at t0 is feasible
+            if not (np.isfinite(value_t0) and np.isscalar(value_t0)):
+                raise ValueError(
+                    f"Input function must return a real number output at t = {t0}"
+                )
```
valentinsulzer, line 101: will need a test for this for coverage

valentinsulzer (commented):
Thanks, this is perfect, just need tests for coverage

CHANGELOG.md
MarcBerliner: This is under Unreleased right now. Do you want me to move everything else under Unreleased into a new section?

valentinsulzer: Yes - though let's check with release team @Saransh-cpp @agriyakhetarpal @kratman to see the best way to deal with this

agriyakhetarpal: Since this PR is going into the `develop` branch and is a feature addition, the merged commit here won't be cherry-picked into further release candidates – it will go straight into v24.9. I think it should be fine to leave as is. Cherry-picked bug fixes will carry their own CHANGELOG entries into the next release candidate or final release.

Since the release notes are still copied over manually, we'll have a chance there to check for unneeded entries

agriyakhetarpal: Since this PR is going into the `develop` branch and is a feature addition, the merged commit here won't be cherry-picked into further release candidates – it will go straight into v24.9. I think it should be fine to leave as is. Ideally, cherry-picked bug fixes will carry their own CHANGELOG entries into the next release candidate or final release.

Since the release notes are still copied over manually, we'll have a chance there to check for unneeded entries

pybamm/experiment/step/base_step.py
MarcBerliner: added tests

MarcBerliner:
For steps that compute their charge/discharge direction (`Current`, `CRate`, `Power`, `Resistance`), there were a few redundancies in the `BaseStep` initialization. I moved the `value_based_charge_or_discharge` inside the `BaseStep` class and evaluate it after sanitizing the step `value`. This should fix the codecov (I believe)

pybamm/experiment/step/base_step.py
```diff
@@ -73,6 +73,7 @@ def __init__(
     ):
         # Check if drive cycle
         self.is_drive_cycle = isinstance(value, np.ndarray)
+        is_python_function = callable(value)
```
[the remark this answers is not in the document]
MarcBerliner: Just pushed this change

pybamm/experiment/step/base_step.py
```diff
@@ -135,10 +130,42 @@ def __init__(
...
+            self.hash_args += f", temperature={temperature}"
+        if tags:
+            self.repr_args += f", tags={tags}"
+        if start_time:
+            self.repr_args += f", start_time={start_time}"
+        if description:
+            self.repr_args += f", description={description}"
+        if direction:
+            self.repr_args += f", direction={direction}"
+            self.hash_args += f", direction={direction}"
```
kratman, line 167: Minor but can we break this out into a separate method? Just to start shortening this long code block in the function

MarcBerliner: I agree that this block is a bit awkward, especially because it's placed in the middle of the function where some but not all of the attributes are set. We can put this in its own separate method and make it the final step in initialization, but this is technically breaking (and I'm not certain how breaking it is), and it's unrelated to this issue, so I think this should be its own PR.

kratman: It would be something like this:
```python
self.repr_args, self.hash_args = self.record_tags(...)
```
Then inside the function there would be no state variables updated inside the function.
```python
def record_tags(...):
        repr_args = f"{value}, duration={duration}"
        hash_args = f"{value}"
        if termination:
            repr_args += f", termination={termination}"
            hash_args += f", termination={termination}"
        if period:
            repr_args += f", period={period}"
        if temperature:
            repr_args += f", temperature={temperature}"
            hash_args += f", temperature={temperature}"
        if tags:
            repr_args += f", tags={tags}"
        if start_time:
            repr_args += f", start_time={start_time}"
        if description:
            repr_args += f", description={description}"
        if direction:
            repr_args += f", direction={direction}"
            hash_args += f", direction={direction}"
        return repr_args, hash_args
```
Not the prettiest, but long functions can be broken up safely without breaking things.

Then if we want to use it as part of initialization, we can do that part of the refactoring later

MarcBerliner: Ok, I just made `record_tags` its own method

kratman: Thanks

PR #4239, opened 2024-07-03
Issue 4224 duration bug

### LINKED ISSUE #4224: [Bug]: [24.5rc0] Simulations with `pybamm.Experiment` terminate after 24 h
### PyBaMM Version

24.5rc0

### Python Version

3.9.13

### Describe the bug

Simulations using `pybamm.Experiment` with a voltage cut-off only appear to terminate with a time cut-off after 24 h.

I'm afraid I haven't had time to get into logs to figure out why this is happening!

### Steps to Reproduce

The following simulation should proceed for approximately 100 h = 360000 s, until the specified voltage cut-off. However, it terminates after 24 h.

```python
import pybamm

    [
        "Discharge at 0.01C until 2.5 V",
    ]
)

sim = pybamm.Simulation(
    model,
    experiment=experiment,
)

sol = sim.solve()
sim.plot()

print(f'Last time step at {sim.solution["Time [s]"].entries[-1] / 3600} h')
```

### Relevant log output

_No response_

valentinsulzer:
# Description

Fixes #4224

- ~~longer default duration (1000 hours)~~ 
- default duration is 2 / C-rate hours for C-rate steps
- warning if step finishes due to hitting the default duration

Currently working on adding a warning if maximum duration is reached

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/experiment/step/base_step.py
+++ b/pybamm/experiment/step/base_step.py
@@ -259,7 +259,7 @@ class BaseStep:
             t = value[:, 0]
             return t[-1]
         else:
-            return 1000 * 3600  # 1000 hours in second
+            return 1000 * 3600  # 1000 hours in seconds
 
     def process_model(self, model, parameter_values):
         new_model = model.new_copy()
--- a/pybamm/experiment/step/base_step.py
+++ b/pybamm/experiment/step/base_step.py
@@ -100,8 +100,12 @@ class BaseStep:
                     f"Input function must return a real number output at t = {t0}"
                 )
 
+        # Record whether the step uses the default duration
+        # This will be used by the experiment to check whether the step is feasible
+        self.uses_default_duration = duration is None
+        self.input_duration = duration
         # Set duration
-        if duration is None:
+        if self.uses_default_duration:
             duration = self.default_duration(value)
         self.duration = _convert_time_to_seconds(duration)
 
@@ -196,7 +200,7 @@ class BaseStep:
         """
         return self.__class__(
             self.value,
-            duration=self.duration,
+            duration=self.input_duration,
             termination=self.termination,
             period=self.period,
             temperature=self.temperature,
--- a/pybamm/experiment/step/base_step.py
+++ b/pybamm/experiment/step/base_step.py
@@ -416,9 +416,15 @@ _type_to_units = {
 
 def _convert_time_to_seconds(time_and_units):
     """Convert a time in seconds, minutes or hours to a time in seconds"""
-    # If the time is a number, assume it is in seconds
-    if isinstance(time_and_units, numbers.Number) or time_and_units is None:
+    if time_and_units is None:
         return time_and_units
+
+    # If the time is a number, assume it is in seconds
+    if isinstance(time_and_units, numbers.Number):
+        if time_and_units <= 0:
+            raise ValueError("time must be positive")
+        else:
+            return time_and_units
 
     # Split number and units
     units = time_and_units.lstrip("0123456789.- ")
--- a/pybamm/experiment/step/base_step.py
+++ b/pybamm/experiment/step/base_step.py
@@ -259,7 +259,7 @@ class BaseStep:
             t = value[:, 0]
             return t[-1]
         else:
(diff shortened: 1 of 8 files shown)


ejfdickinson:
@valentinsulzer You need to take `abs(value)` in `CRate.default_duration()`, because `value` comes through negative for charge steps.

brosaplanella (approved):
Looks good to me, let's just add a line to CHANGELOG before merging.

ejfdickinson:
@brosaplanella @valentinsulzer To confirm, this is working as expected for the use cases on which I was testing.

valentinsulzer:
Changing default duration to 1000 hours caused a severe performance degradation, investigating further

aabills:
Thanks! I just tried it and though I'd prefer a longer duration barring a substantial performance hit, in my opinion, the bug is fixed by the warning alone. 

I wonder if 100 + ε hours might be a happy medium?

valentinsulzer:
This PR adds the warning and also fixes the C/100 case which I assume is more common than specifying a small current (backed up by a n=2 sample size from @aabills and @ejfdickinson ). The default duration is 24h and we can reconsider once @MarcBerliner is done with updates to solvers, which will make solve times much less sensitive to things like duration, period, etc

pybamm/simulation.py
```diff
@@ -768,12 +769,19 @@ def solve(
 
                     logs["termination"] = step_solution.termination
                     # Only allow events specified by experiment
-                    if not (
+                    if (
+                        step_termination == "final time"
+                        and step.uses_default_duration is True
```
kratman, line 774: We can probably skip the "is True" part here since it is a bool in the definition

pybamm/simulation.py
```diff
@@ -768,12 +769,19 @@ def solve(
...
+                        break
+                    elif not (
                         isinstance(step_solution, pybamm.EmptySolution)
                         or step_termination == "final time"
                         or "[experiment]" in step_termination
                     ):
-                        callbacks.on_experiment_infeasible(logs)
+                        callbacks.on_experiment_infeasible_event(logs)
                         feasible = False
                         break
```
kratman, line 786: It is usually good to end the chains in an else of some sort to be clearer for future devs. A quick workaround is to just break up the last elif:
```python
elif condition:
    code()
```
becomes
```python
else:
    if condition:
        code()
```

Not the prettiest, but it is clearer that the chain is done.

pybamm/experiment/step/base_step.py
```diff
@@ -100,8 +102,11 @@ def __init__(
                     f"Input function must return a real number output at t = {t0}"
                 )
 
+        # Record whether the step uses the default duration
+        # This will be used by the experiment to check whether the step is feasible
+        self.uses_default_duration = duration is None
```
kratman, line 107: I assume a zero duration should also trigger the default, not just if one is not passed in, correct?

pybamm/experiment/step/base_step.py
```diff
@@ -71,6 +71,8 @@ def __init__(
         description=None,
         direction=None,
     ):
+        self.input_duration = duration
+        self.input_value = value
```
kratman, line 75: If these are only used for the copy method, is there a strong reason not to just use the python copy.deepcopy() method? For the array items the constructor is looping over the arrays anyway, so I would assume python's built in copy would be safer, cleaner and potentially faster than writing our own copy operation.

kratman (changes requested):

pybamm/experiment/step/base_step.py
valentinsulzer: Zero duration should error, I'll add a separate check for that

pybamm/experiment/step/base_step.py
valentinsulzer: In general I've found deepcopy to be a huge performance bottleneck, and there's a few attributes of step that we don't want to copy (start time and end time, which are calculated based on the adjacent steps). The proper fix willl be to refactor steps so we don't have to copy at all, but that is outside the scope of this PR

kratman (approved):

PR #3853, opened 2024-02-29
chore: cleanup mypy config

Saransh-cpp:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes #3489

## Type of change

Please add a line in the relevant section of [CHANGELOG.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CHANGELOG.md) to document the change (include PR #) - note reverse order of PR #s. If necessary, also add to the list of breaking changes.

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (back-end change that speeds up the code)
- [ ] Bug fix (non-breaking change which fixes an issue)

# Key checklist:

- [ ] No style issues: `$ pre-commit run` (or `$ nox -s pre-commit`) (see [CONTRIBUTING.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CONTRIBUTING.md#installing-and-using-pre-commit) for how to set this up to run automatically when committing locally, in just two lines of code)
- [ ] All tests pass: `$ python run-tests.py --all` (or `$ nox -s tests`)
- [ ] The documentation builds: `$ python run-tests.py --doctest` (or `$ nox -s doctests`)

You can run integration tests, unit tests, and doctests together at once, using `$ python run-tests.py --quick` (or `$ nox -s quick`).

## Further checks:

- [ ] Code is commented, particularly in hard-to-understand areas
- [ ] Tests added that prove fix is effective or that feature works

PR #4242, opened 2024-07-04
`scikit-build-core` CI Builds

cringeyburger:
# Description

Added Linux and MacOS builds using `scikit-build-core`; there is some redundant code that needs to be cleared

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:


agriyakhetarpal:
The RTD build and the Windows builds because it doesn't have `pybind11`. We should move away from having to git-clone it – just use `pybind11[global]` as a build-time dependency in `pyproject.toml`. Doing this will bump the installation time because compilation will now be involved, so it's probably best to start looking at how to make IDAKLU optional while building.

cringeyburger:
Task: Trying to fix RTD builds by shifting `pybind` to `pyproject` table

cringeyburger:
Just fixed the merge conflicts. Please let me know if there are any problems!

cringeyburger:
Just ran `sp-repo-review` and here is the review. 
```
General:

 • Detected build backend: scikit_build_core.build                                                                                                                                                        
 • Detected license(s): BSD License                                                                                                                                                                       

├── PY001 Has a pyproject.toml ✅
├── PY002 Has a README.(md|rst) file ✅
├── PY003 Has a LICENSE* file ✅
├── PY004 Has docs folder ✅
├── PY005 Has tests folder ✅
├── PY006 Has pre-commit config ✅
└── PY007 Supports an easy task runner (nox or tox) ✅

PyProject:
├── PP002 Has a proper build-system table ✅
├── PP003 Does not list wheel as a build-dep ✅
├── PP301 Has pytest in pyproject ✅
├── PP302 Sets a minimum pytest to at least 6 ✅
├── PP303 Sets the test paths ✅
├── PP304 Sets the log level in pytest ✅
├── PP305 Specifies xfail_strict ✅
├── PP306 Specifies strict config ✅
├── PP307 Specifies strict markers ✅
├── PP308 Specifies useful pytest summary ✅
└── PP309 Filter warnings specified ✅

GitHub Actions:
├── GH100 Has GitHub Actions config ✅
├── GH101 Has nice names ✅
├── GH102 Auto-cancel on repeated PRs ✅
├── GH103 At least one workflow with manual dispatch trigger ✅
├── GH104 Use unique names for upload-artifact ✅
├── GH200 Maintained by Dependabot ✅
├── GH210 Maintains the GitHub action versions with Dependabot ✅
├── GH211 Do not pin core actions as major versions ✅
└── GH212 Require GHA update grouping ✅

Pre-commit:
├── PC100 Has pre-commit-hooks ✅
├── PC110 Uses black or ruff-format ✅
├── PC111 Uses blacken-docs ✅
├── PC140 Uses mypy ❌
│   Must have https://github.com/pre-commit/mirrors-mypy repo in .pre-commit-config.yaml                                                                                                                  
├── PC160 Uses codespell ❌
│   Must have https://github.com/codespell-project/codespell repo in .pre-commit-config.yaml                                                                                                              
├── PC170 Uses PyGrep hooks (only needed if RST present) ✅
├── PC180 Uses prettier ❌
│   Must have https://github.com/pre-commit/mirrors-prettier repo in .pre-commit-config.yaml                                                                                                              
├── PC190 Uses Ruff ✅
├── PC191 Ruff show fixes if fixes enabled ✅
└── PC901 Custom pre-commit CI message ✅

MyPy:
├── MY100 Uses MyPy (pyproject config) ✅
├── MY101 MyPy strict mode ❌
│   Must have strict in the mypy config. MyPy is best with strict or nearly strict configuration. If you are happy with the strictness of your settings already, ignore this check or set strict = false  
│   explicitly.                                                                                                                                                                                           
│   
│                                                                                                                                                                                                         
│    [tool.mypy]                                                                                                                                                                                          
│    strict = true                                                                                                                                                                                        
│                                                                                                                                                                                                         
├── MY102 MyPy show_error_codes deprecated ✅
├── MY103 MyPy warn unreachable ❌
│   Must have warn_unreachable (true/false) to pass this check. There are occasionally false positives (often due to platform or Python version static checks), so it's okay to set it to false if you    
│   need to. But try it first - it can catch real bugs too.                                                                                                                                               
│   
│                                                                                                                                                                                                         
│    [tool.mypy]                                                                                                                                                                                          
│    warn_unreachable = true                                                                                                                                                                              
│                                                                                                                                                                                                         
├── MY104 MyPy enables ignore-without-code ❌
│   Must have "ignore-without-code" in enable_error_code = [...]. This will force all skips in your project to include the error code, which makes them more readable, and avoids skipping something      
│   unintended.                                                                                                                                                                                           
│   
│                                                                                                                                                                                                         
│    [tool.mypy]                                                                                                                                                                                          
│    enable_error_code = ["ignore-without-code", "redundant-expr", "truthy-bool"]                                                                                                                         
│                                                                                                                                                                                                         
├── MY105 MyPy enables redundant-expr ❌
│   Must have "redundant-expr" in enable_error_code = [...]. This helps catch useless lines of code, like checking the same condition twice.                                                              
│   
│                                                                                                                                                                                                         
│    [tool.mypy]                                                                                                                                                                                          
│    enable_error_code = ["ignore-without-code", "redundant-expr", "truthy-bool"]                                                                                                                         
│                                                                                                                                                                                                         
└── MY106 MyPy enables truthy-bool ❌
    Must have "truthy-bool" in enable_error_code = []. This catches mistakes in using a value as truthy if it cannot be falsey.                                                                           
    
                                                                                                                                                                                                          
     [tool.mypy]                                                                                                                                                                                          
     enable_error_code = ["ignore-without-code", "redundant-expr", "truthy-bool"]                                                                                                                         
                                                                                                                                                                                                          

Ruff:
├── RF001 Has Ruff config ✅
├── RF002 Target version must be set ✅
├── RF003 src directory specified if used [skipped]
├── RF101 Bugbear must be selected ✅
├── RF102 isort must be selected ❌
│   Must select the isort I checks. Recommended:                                                                                                                                                          
│   
│                                                                                                                                                                                                         
│    [tool.ruff.lint]                                                                                                                                                                                     
│    extend-select = [                                                                                                                                                                                    
│      "I",  # isort                                                                                                                                                                                      
│    ]                                                                                                                                                                                                    
│                                                                                                                                                                                                         
├── RF103 pyupgrade must be selected ✅
├── RF201 Avoid using deprecated config settings ✅
└── RF202 Use (new) lint config section ✅

Documentation:
├── RTD100 Uses ReadTheDocs (pyproject config) ✅
├── RTD101 You have to set the RTD version number to 2 ✅
├── RTD102 You have to set the RTD build image ✅
└── RTD103 You have to set the RTD python version ✅

```

pyproject.toml
```diff
@@ -268,7 +285,7 @@ concurrency = ["multiprocessing"]
 
 [tool.repo-review]
 ignore = [
-    "PP003"  # list wheel as a build-dep
+    "PP003", # list wheel as a build-dep
 ]
```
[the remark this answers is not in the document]
agriyakhetarpal: Resolving this conversation as per the results in https://github.com/pybamm-team/PyBaMM/pull/4242#issuecomment-2241918092.

agriyakhetarpal:
Perfect, thanks for checking, @cringeyburger! We can ignore the mypy checks and the other ruff-isort ones for now, since we are tracking them elsewhere, in #3489 and #3853 (@Saransh-cpp, gentle ping on that PR!)

agriyakhetarpal:
Note that the Windows tests and the Ubuntu doctests are still failing because we haven't made compiling the IDAKLU solver optional. The environment variables to check for and the associated behaviour should be:

1. `CIBUILDWHEEL` – if `cibuildwheel` is running, make it mandatory to build compiled wheels, so that we always publish platform-specific wheels to PyPI
2. `BUILD_IDAKLU` can be set by the user to force the compilation of the IDAKLU solver. However, if we set this, it will make the IDAKLU solver an opt-in piece of functionality (it has always been so, but users don't know that — see below), and people will have to manually enable compilation (but `noxfile.py` could potentially set this variable to a non-zero value).

We don't currently explain to the users that compiling the IDAKLU solver is optional for using PyBaMM, so updates to the documentation and the installation guide are very much welcome, maybe even in the same PR. We need additional instructions and other refinements to be added, such as Windows-specific instructions.

cringeyburger:
>However, libomp.dylib is being copied into sundials_KLU_libs/, but delocate is searching for it inside sundials_KLU_libs/lib/ (note the extra /lib/). I would advise looking into why this is happening.

From what I read, this happens when one of the libraries gives a library dependency with a relative path. [Here](https://github.com/matthew-brett/delocate?tab=readme-ov-file#delocationerror-library-does-not-exist) is the link to the resource.
Continuing, it is said that this is not a `delocate` problem but a `CMake` one. 
I think I will try installing `libomp.dylib` inside `sundials_KLU_libs/lib/` and check if the relative path issue is fixed.

cringeyburger:
From this [job](https://github.com/cringeyburger/PyBaMM/actions/runs/10046244024), it seems that just using `pybind11` instead of `pybind11[global]` works fine

pyproject.toml
```diff
@@ -1,25 +1,25 @@
...
-    "setuptools>=64",
-    "wheel",
+    "scikit-build-core>=0.9",
     # On Windows, use the CasADi vcpkg registry and CMake bundled from MSVC
     "casadi>=3.6.5; platform_system!='Windows'",
     # Note: the version of CasADi as a build-time dependency should be matched
     # cross platforms, so updates to its minimum version here should be accompanied
     # by a version bump in https://github.com/pybamm-team/casadi-vcpkg-registry.
     "cmake; platform_system!='Windows'",
+    "pybind11[global]",
```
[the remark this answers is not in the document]
agriyakhetarpal: Resolving based on https://github.com/pybamm-team/PyBaMM/pull/4242#issuecomment-2243556473

agriyakhetarpal:
Thanks for trying to investigate the RPATH issue more thoroughly. Could you generate a build artifact by temporarily setting `CIBW_REPAIR_WHEEL_COMMAND` to something like `exit 0` in a commit so that the wheel builds/repairs pass, and the workflow succeeds? I think it will help to examine the files locally, I can do that on my macOS machine.

agriyakhetarpal:
Sorry, I should have mentioned more clearly – `delocate-wheel` still fails because it still needs a wheel to inspect, so what I wanted to suggest was `<wheel repair command> || exit 0`, i.e. append `exit 0`. To avoid that and disable repairing wheels completely, you can set the repair wheel command to an empty string, i.e., `""`.

agriyakhetarpal:
Thanks, @cringeyburger. We'll need not only the wheel but also the other dylibs since they will not be copied into the wheel anymore. If you could copy them all from `sundials_KLU_libs` (and also `libomp.dylib`) to the folder where the wheel is expected to be after `cibuildwheel` finishes, we can have all of the files ready to download for local testing. Once this is done, please paste a link to the workflow run here and ping me.

cringeyburger:
Hey @agriyakhetarpal, [here](https://github.com/cringeyburger/PyBaMM/actions/runs/10088349384/job/27893976968) is the link to the job. I've uploaded the dylibs along with the MacOS wheels. Please let me know if you need anything else!

agriyakhetarpal:
Output from running `otool` is

```
❯ otool -vL libomp.dylib
libomp.dylib:
	{project}/sundials_KLU_libs/libomp.dylib (compatibility version 5.0.0, current version 5.0.0)
	time stamp 1 Thu Jan  1 05:30:01 1970
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1292.0.0)
	time stamp 2 Thu Jan  1 05:30:02 1970
```

agriyakhetarpal:
I can't suggest changes directly from the Files tab because I can only do so for modified or deleted lines. Would it be okay if I were to push a few commits to your branch here? After that, you can use the `git pull` command to retrieve those commits locally and keep working ahead. Let me know if this works for you.

cringeyburger:
Yes, please do

agriyakhetarpal:
@cringeyburger, I fixed the wheel repairs for you. You can now pull the commits I pushed here to your branch locally and continue working on the rest of the tasks we talked about. I think the primary way in order to get this PR to a preliminary reviewable state would be to clean up redundant code and fix all of the PR tests (Windows test suite and Linux doctests) and the RTD build, which would be by making the compilation of the IDAKLU solver optional like we have been discussing.

agriyakhetarpal:
From my action item for this week: I tested the changes here on my M-series macOS machine with `nox -s pybamm-requires` (which is unchanged, except for the installation directory) and then using

```bash
pip install --no-build-isolation --config-settings=editable.rebuild=true -Cbuild-dir=build -ve.
```

based on the docs for `scikit-build-core`. This works perfectly, and CMake recognises and performs partial rebuilds when I type `import pybamm` and I place a few modifications in any of the C++ files listed in `pybind11_add_module` for the IDAKLU solver in `CMakeLists.txt`, and it tries to recompile and link the modified module. Good progress!

One thing to note here is that if we are going to use this setting, we need to follow the guidelines on both editable installations and their partial rebuilds recpmmended in the `scikit-build-core` documentation. This requires modifying `noxfile.py` a bit, such as to enable `--no-build-isolation` when installing the package (we are countering this by the use of virtual environments that `nox` provides). Subsequently, once we enable `--no-build-isolation`, the build-time dependencies (pybind11, CMake, etc.) won't be installed by default, and the setup will raise an error on their unavailability, so we will need to install them on our own beforehand. One way to do this is to have `nox` read those from `pyproject.toml`'s `[build-system.requires]` TOML table through a TOML reader library (I recommend [`tomlkit`](https://pypi.org/project/tomlkit/) for Python <3.11 until we switch to Python 3.11+), we can pass them to `session.install()` everywhere. Lastly, we should now have either an upper pin on or we should completely pin all of our build-time dependencies. This will serve two grounds: i. we can avoid unwanted breaking changes, since both `scikit-build-core`'s build backend and the editable-rebuild config setting are a part of an experimental package and an experimental feature respectively (no v1 release yet), and ii. adding bounds helps with reproducibility in some aspects if someone were to build PyBaMM's older releases from source some time down the line.

Please let me know in case any of this is unclear, and I can try to summarise what is needed to be done.

P.S. I tried the available `-Ceditable.mode=inplace` config setting as well, and doesn't make sense for us to use it – since we have all of the built files in the source code directory (in `build/` and `sundials_KLU_libs`) anyway, and it clutters the source directory even further without offering any advantages.

P.P.S. CMake doesn't recognise any changes in SuiteSparse/SUNDIALS when reinstalling in editable mode, of course, given that they are not interfaced with `pybind11` directly in any way).

agriyakhetarpal:
Also, instead of environment variables, we should use https://scikit-build-core.readthedocs.io/en/latest/configuration.html#configuring-cmake-arguments-and-defines to set build options we need such as `USE_PYTHON_CASADI` and the replace the ones we use for the Windows compilation: https://github.com/pybamm-team/PyBaMM/blob/48b4e48d776ad2c254e5fae8db94706c57a8cc38/.github/workflows/publish_pypi.yml#L78-L83

(hopefully we can reduce some of these sometime soon)

.github/workflows/periodic_benchmarks.yml
```diff
@@ -44,7 +44,8 @@ jobs:
           asv machine --machine "GitHubRunner"
           asv run --machine "GitHubRunner" NEW --show-stderr -v
         env:
-          SUNDIALS_INST: $HOME/.local
+          # TODO: test this
+          SUNDIALS_INST: sundials_KLU_libs/
           LD_LIBRARY_PATH: $HOME/.local/lib
```
agriyakhetarpal, line 49:
```suggestion
```

Both of these shouldn't be needed anymore. You'll need to update the build command in `asv.conf.json`, though, it still uses `setup.py`. Once you manage to get the benchmarks working, let me know with a link to the logs :)

CMakeLists.txt
```diff
@@ -32,11 +34,25 @@ endif()
...
+#   set(BUILD_IDAKLU ON)
+# elseif(DEFINED ENV{BUILD_IDAKLU})
+#   set(BUILD_IDAKLU "$ENV{BUILD_IDAKLU}")
+# else()
+#   set(BUILD_IDAKLU OFF)
+# endif()
 
+# message(STATUS "BUILD_IDAKLU is set to: ${BUILD_IDAKLU}")
+
+# if(BUILD_IDAKLU)
```
agriyakhetarpal, line 54: Should this be cleaned up?

CMakeLists.txt
```diff
@@ -32,11 +34,25 @@ endif()
...
+message(STATUS "Building with IDAKLU support")
```
agriyakhetarpal, line 55:
```suggestion
```

I think this is obvious, since it makes sense that CMake will be invoked only if something has to be compiled, which in this case is the IDAKLU solver.

CMakeLists.txt
```diff
@@ -149,14 +169,16 @@ target_link_libraries(idaklu PRIVATE ${SUNDIALS_LIBRARIES} casadi)
 # link suitesparse
 # if using vcpkg, use config mode to
 # find suitesparse. Otherwise, use FindSuiteSparse module
+set(SUITESPARSE_ROOT ${CUSTOM_INSTALL_DIR})
+message("SUITESPARSE_ROOT: ${SUITESPARSE_ROOT}")
 if(DEFINED VCPKG_ROOT_DIR)
-  find_package(SuiteSparse CONFIG REQUIRED)
+  find_package(SUITESPARSE CONFIG REQUIRED)
```
agriyakhetarpal, line 175:
```suggestion
  find_package(SuiteSparse CONFIG REQUIRED)
```

We should be good to go back to the original naming convention

CMakeLists.txt
```diff
@@ -149,14 +169,16 @@ target_link_libraries(idaklu PRIVATE ${SUNDIALS_LIBRARIES} casadi)
...
 else()
-  find_package(SuiteSparse REQUIRED)
+  find_package(SUITESPARSE REQUIRED)
```
agriyakhetarpal, line 177:
```suggestion
  find_package(SuiteSparse REQUIRED)
```

CMakeLists.txt
```diff
@@ -149,14 +169,16 @@ target_link_libraries(idaklu PRIVATE ${SUNDIALS_LIBRARIES} casadi)
...
   message("SuiteSparse found in ${SuiteSparse_INCLUDE_DIRS}: ${SuiteSparse_LIBRARIES}")
 endif()
-include_directories(${SuiteSparse_INCLUDE_DIRS})
-target_link_libraries(idaklu PRIVATE ${SuiteSparse_LIBRARIES})
+include_directories(${SUITESPARSE_INCLUDE_DIRS})
+target_link_libraries(idaklu PRIVATE ${SUITESPARSE_LIBRARIES})
```
agriyakhetarpal, line 181:
```suggestion
include_directories(${SuiteSparse_INCLUDE_DIRS})
target_link_libraries(idaklu PRIVATE ${SuiteSparse_LIBRARIES})
```

CMakeLists.txt
```diff
@@ -170,3 +192,8 @@ if(${PYBAMM_IDAKLU_EXPR_IREE} STREQUAL "ON" )
   target_link_libraries(idaklu PRIVATE iree_compiler_bindings_c_loader)
   target_link_libraries(idaklu PRIVATE iree_runtime_runtime)
 endif()
+
+install(TARGETS idaklu DESTINATION pybamm/solvers)
+# else()
+#   message(STATUS "Building without IDAKLU support")
+# endif()
```
agriyakhetarpal, line 199:
```suggestion
```

CONTRIBUTING.md
```diff
@@ -55,7 +55,7 @@ You now have everything you need to start making changes!
 ### C. Merging your changes with PyBaMM
 
 10. [Test your code!](#testing)
-11. PyBaMM has online documentation at http://docs.pybamm.org/. To make sure any new methods or classes you added show up there, please read the [documentation](#documentation) section.
+11. PyBaMM has online documentation at <http://docs.pybamm.org/>. To make sure any new methods or classes you added show up there, please read the [documentation](#documentation) section.
```
agriyakhetarpal, line 58:
```suggestion
11. PyBaMM has online documentation at https://docs.pybamm.org/. To make sure any new methods or classes you added show up there, please read the [documentation](#documentation) section.
```

Not sure why this is needed, since this is a Markdown file? Also, we can switch to HTTPS.

CONTRIBUTING.md
```diff
@@ -373,7 +375,7 @@ These docstrings can be fairly simple, but can also make use of [reStructuredTex
 
 In addition, we write a (very) small bit of documentation in separate reStructuredText files in the `docs` directory. Most of what these files do is simply import docstrings from the source code. But they also do things like add tables and indexes. If you've added a new class to a module, search the `docs` directory for that module's `.rst` file and add your class (in alphabetical order) to its index. If you've added a whole new module, copy-paste another module's file and add a link to your new file in the appropriate `index.rst` file.
 
-Using [Sphinx](http://www.sphinx-doc.org/en/stable/) the documentation in `docs` can be converted to HTML, PDF, and other formats. In particular, we use it to generate the documentation on http://docs.pybamm.org/
+Using [Sphinx](http://www.sphinx-doc.org/en/stable/) the documentation in `docs` can be converted to HTML, PDF, and other formats. In particular, we use it to generate the documentation on <http://docs.pybamm.org/>
```
agriyakhetarpal, line 378:
```suggestion
Using [Sphinx](http://www.sphinx-doc.org/en/stable/), the documentation in `docs` can be converted to HTML, PDF, and other formats. In particular, we use it to generate the documentation on https://docs.pybamm.org/.
```

CONTRIBUTING.md
```diff
@@ -417,17 +419,17 @@ wherever code is called that uses that citation (for example, in functions or in
 
 ### Installation
 
-Installation of PyBaMM and its dependencies is handled via [pip](https://pip.pypa.io/en/stable/) and [setuptools](http://setuptools.readthedocs.io/). It uses `CMake` to compile C++ extensions using [`pybind11`](https://pybind11.readthedocs.io/en/stable/) and [`casadi`](https://web.casadi.org/). The installation process is described in detail in the [source installation](https://docs.pybamm.org/en/latest/source/user_guide/installation/install-from-source.html) page and is configured through the `CMakeLists.txt` file.
+Installation of PyBaMM and its dependencies is handled via [pip](https://pip.pypa.io/en/stable/) and [scikit-build-core](https://scikit-build-core.readthedocs.io/en/latest/getting_started.html). It uses `CMake` to compile C++ extensions using [`pybind11`](https://pybind11.readthedocs.io/en/stable/) and [`casadi`](https://web.casadi.org/). The installation process is described in detail in the [source installation](https://docs.pybamm.org/en/latest/source/user_guide/installation/install-from-source.html) page and is configured through the `CMakeLists.txt` file.
```
agriyakhetarpal, line 422:
```suggestion
Installation of PyBaMM and its dependencies is handled via [pip](https://pip.pypa.io/en/stable/) and [scikit-build-core](https://scikit-build-core.readthedocs.io/en/stable/getting_started.html). It uses `CMake` to compile C++ extensions using [`pybind11`](https://pybind11.readthedocs.io/en/stable/) and [`casadi`](https://web.casadi.org/). The installation process is described in detail in the [source installation](https://docs.pybamm.org/en/latest/source/user_guide/installation/install-from-source.html) page and is configured through the `CMakeLists.txt` file.
```

CONTRIBUTING.md
```diff
@@ -417,17 +419,17 @@ wherever code is called that uses that citation (for example, in functions or in
...
 
 Configuration files:
 
 ```
-setup.py
 pyproject.toml
-MANIFEST.in
+.gitignore
```
agriyakhetarpal, line 428:
```suggestion
```

We can remove this, since `scikit-build-core` is configured entirely inside `pyproject.toml` and the reliance on a VCS for file inclusion/exclusion is purely a matter of preference. You've mentioned the use of `.gitignore` below, which seems to be sufficient.

CONTRIBUTING.md
```diff
@@ -417,17 +419,17 @@ wherever code is called that uses that citation (for example, in functions or in
...
 ```
 
-Note: `MANIFEST.in` is used to include and exclude non-Python files and auxiliary package data for PyBaMM when distributing it. If a file is not included in `MANIFEST.in`, it will not be included in the source distribution (SDist) and subsequently not be included in the binary distribution (wheel).
+Note: `pyproject.toml` and `.gitignore` are used to include and exclude non-Python files and auxiliary package data for PyBaMM when distributing it. If a file is mentioned in `.gitignore`, it will not be included in the source distribution (SDist) by `scikit-build-core`, and subsequently not be included in the binary distribution (wheel).
```
agriyakhetarpal, line 431:
```suggestion
Note: `pyproject.toml` and `.gitignore` are used to include and exclude non-Python files and auxiliary package data for PyBaMM when distributing it. If a file is mentioned in `.gitignore`, it will not be included in the source distribution (SDist), and subsequently not be included in the binary distribution (wheel).
```

The build frontend (`pip wheel` or `build`) is what invokes the build backend (`scikit-build-core`), but I think it's better to exclude that detail, at least from our docs.

CONTRIBUTING.md
```diff
@@ -443,7 +445,7 @@ Configuration files for various GitHub actions workflow can be found in `.github
 
 ### Codecov
 
-Code coverage (how much of our code is actually seen by the (linux) unit tests) is tested using [Codecov](https://docs.codecov.io/), a report is visible on https://codecov.io/gh/pybamm-team/PyBaMM.
+Code coverage (how much of our code is actually seen by the (linux) unit tests) is tested using [Codecov](https://docs.codecov.io/), a report is visible on <https://codecov.io/gh/pybamm-team/PyBaMM>.
```
agriyakhetarpal, line 448: Same question as above

CONTRIBUTING.md
```diff
@@ -453,7 +455,7 @@ Configuration files:
 
 ### Read the Docs
 
-Documentation is built using https://readthedocs.org/ and published on http://docs.pybamm.org/.
+Documentation is built using <https://readthedocs.org/> and published on <http://docs.pybamm.org/>.
```
agriyakhetarpal, line 458: Same question as above

FindSUITESPARSE.cmake
agriyakhetarpal, line 1: Could you revert the file rename?

pyproject.toml
```diff
@@ -1,25 +1,26 @@
...
-    "wheel",
+    "scikit-build-core>=0.9.9",
     # On Windows, use the CasADi vcpkg registry and CMake bundled from MSVC
     "casadi>=3.6.5; platform_system!='Windows'",
     # Note: the version of CasADi as a build-time dependency should be matched
     # cross platforms, so updates to its minimum version here should be accompanied
     # by a version bump in https://github.com/pybamm-team/casadi-vcpkg-registry.
     "cmake; platform_system!='Windows'",
+    # TODO: check if just pybind11 works or not
+    "pybind11",
```
agriyakhetarpal, line 11:
```suggestion
    "pybind11",
```

Now that we have confirmed it, we can remove this line :)

pyproject.toml
```diff
@@ -156,15 +149,19 @@ Xu2019 = "pybamm.input.parameters.lithium_ion.Xu2019:get_parameter_values"
...
+[tool.scikit-build]
+minimum-version = "0.9"
+build-dir = "build/{wheel_tag}"
+
+[tool.scikit-build.logging]
+level = "INFO"
+
+[tool.scikit-build.wheel]
+cmake = false
+expand-macos-universal-tags = true
```
agriyakhetarpal, line 161:
```suggestion
cmake = false
```

We won't be compiling or publishing universal2 wheels for macOS anytime soon, so this setting is not needed.

.readthedocs.yaml
```diff
@@ -27,6 +27,7 @@ build:
     # Altered PDF build and upload job
     # This also runs on PR builds, but does not upload the PDF
     post_build:
+      - export BUILD_IDAKLU=OFF
```
agriyakhetarpal, line 30: Since `wheel.cmake` is now `false` by default, can the RTD builds work without setting this environment variable?

pyproject.toml
```diff
@@ -173,16 +170,23 @@ pybamm = [
...
-[tool.setuptools.packages.find]
-include = ["pybamm", "pybamm.*"]
+[tool.scikit-build.cmake]
+version = ">=3.13"
+verbose = true
+build-type = "Release"
+
+[[tool.scikit-build.overrides]]
+if.any.env.BUILD_IDAKLU = true
+if.any.env.CIBUILDWHEEL = true
```
agriyakhetarpal, line 180: I think we should decide on a format for this – whether to set it to `true`, `True`, or `ON`. I think `ON/OFF` sounds nice, but it is a matter of preference.

CMakeLists.txt
cringeyburger: I thought it would work well with the UX aspect. I will remove it 👍

CONTRIBUTING.md
cringeyburger: Markdownlint says that we shouldn't use bare URLs. Here is the [rule](https://github.com/updownpress/markdown-lint/blob/master/rules/034-no-bare-urls.md). Basically, the URL isn't converted to a link in many parsers without the angular brackets.

.readthedocs.yaml
cringeyburger: I will test it and let you know

CONTRIBUTING.md
agriyakhetarpal: Fair enough, Markdownlint has some nice rules, but we use the [GFM (GitHub-flavoured Markdown)](https://github.github.com/gfm/) and [MyST Markdown](https://mystmd.org/) variants for rendering on GitHub and our Sphinx docs, which tie in nicely with where these files are hosted.

These changes are welcome indeed and the cleanup is well-deserved, but I would prefer if you could open a separate PR that adds Markdownlint as a pre-commit hook, so that this PR is limited to just build/packaging infra changes.

pyproject.toml
cringeyburger: Yes, `ON/OFF` sounds nice

.readthedocs.yaml
cringeyburger: Just tested the RTD builds, looks like it is working fine: [LINK](https://readthedocs.org/projects/migrate-test/builds/25225826/)

agriyakhetarpal:
> From what I can see, setting `if.any.env.* = "ON"` isn't triggering `wheel.cmake`.

I just read the docs, where it is said that any non-null value for the environment variable is okay: https://scikit-build-core.readthedocs.io/en/latest/overrides.html#env-string-or-bool. We can keep it as `true` in that case.

cringeyburger:
Setting the new CMake policy helped Linux builds to work again, but now there is a problem with the MacOS repair step. I will try looking at how you fixed it and try to replicate the steps.

agriyakhetarpal:
> I will try looking at how you fixed it and try to replicate the steps.

Sure – I fixed it in the end by avoiding the `{project}` placeholder altogether and working in the root directory. Also, I fixed the path where `libomp.dylib` was being copied. If it helps, you may also post a link to the logs here to provide more context.

cringeyburger:
@agriyakhetarpal, I went through your commits and didn't understand what to change. Everything looks fine, and the MacOS builds should work. Here is the latest [build](https://github.com/cringeyburger/PyBaMM/actions/runs/10267452686/job/28408028523) I tried working on.

agriyakhetarpal:
Thanks, @cringeyburger – I will be able to take a look at this tomorrow or the day after. I feel you should restore the SuiteSparse name, though – I un-resolved the conversation(s) on those suggestions. I don't think a new CMake policy is needed, how were the Linux builds passing earlier?

cringeyburger:
@agriyakhetarpal The Linux builds were working before the renaming.

cringeyburger:
Why are doctests failing? Is there something I can do?

cringeyburger:
I was trying to fix the benchmarks workflow. It seems that we may have to install `libmambapy`, which is missing. Here is the link to the latest [job](https://github.com/cringeyburger/PyBaMM/actions/runs/10284599375/job/28460972514#step:7:14). 
We can install the module using `conda` or `mamba`, if I am not wrong.

agriyakhetarpal:
The error seems to be masked a few lines below that and `libmambapy` not being present isn't the actual error, based on what I got from a Google search: https://github.com/sympy/sympy/issues/26239. Could you try to fix `benchmarks_on_push.yml` in the meantime instead of `periodic_benchmarks.yml`? You can temporarily enable it by allowing it to run manually using a `workflow_dispatch:` trigger. We can adapt the changes to the rest of the benchmarks workflows after that.

cringeyburger:
I ran the benchmarks, here is the [job](https://github.com/cringeyburger/PyBaMM/actions/runs/10287876181/job/28471968604). It failed though.

agriyakhetarpal:
We just merged #4311 to move to an `src` layout, which means that this PR now has some conflicts (apologies for the disruption). I think they should be fairly resolvable and the builds should mostly work, but if you face any issues with this, let us know! We have a guide for you that should make it easier.

agriyakhetarpal:
Superseded by #4352

PR #4252, opened 2024-07-08
Pin `sympy` to `<1.13.0`

MarcBerliner:
# Description

`sympy` `v1.13.0` breaks CI. Temporary pin to fix it

## Type of change

Please add a line in the relevant section of [CHANGELOG.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CHANGELOG.md) to document the change (include PR #) - note reverse order of PR #s. If necessary, also add to the list of breaking changes.

- [x] Bug fix (non-breaking change which fixes an issue)

# Key checklist:

- [x] No style issues: `$ pre-commit run` (or `$ nox -s pre-commit`) (see [CONTRIBUTING.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CONTRIBUTING.md#installing-and-using-pre-commit) for how to set this up to run automatically when committing locally, in just two lines of code)
- [x] All tests pass: `$ python run-tests.py --all` (or `$ nox -s tests`)
- [x] The documentation builds: `$ python run-tests.py --doctest` (or `$ nox -s doctests`)

You can run integration tests, unit tests, and doctests together at once, using `$ python run-tests.py --quick` (or `$ nox -s quick`).

## Further checks:

- [ ] Code is commented, particularly in hard-to-understand areas
- [ ] Tests added that prove fix is effective or that feature works

PR #4253, opened 2024-07-08
Fixes for sympy update

kratman:
# Description

SymPy updated and seems to no longer recognize PyBaMM symbols as strings.

It gets to this block (N.B. warning text was edited for readability):
```
    if not isinstance(a, str):
        try:
            a = str(a)
        except Exception as exc:
            raise SympifyError(a, exc)
        sympy_deprecation_warning("The string fallback in sympify() is deprecated.")
```
Then fails due to an exception. Even if it did convert to a string, it would have caused a deprecation warning.

Failing tests looked like this:
```
a = pybamm.Symbol("a", domain="test")
self.assertEqual(pybamm.Arcsinh(a).to_equation(), sympy.asinh(a))
```

There seemed to be two options:
1.  Add a `pybamm.Symbol._sympy_()` method. This would be basically identical to `pybamm.Symbol.to_equation()`, and would not be a good test.
2. Explicitly put the expected string into the sympy function. This works as a test of `pybamm.Symbol.to_equation()` by confirming that `pybamm.Symbol.print_name()` returns the original name, but could be a bit redundant.

I went with option (2) since it keeps the tests working, but it could be good to determine if `pybamm.Symbol.print_name()` and `pybamm.Symbol.to_equation()` could be just replaced by a `pybamm.Symbol._sympy_()`method.

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/tests/unit/test_expression_tree/test_functions.py
+++ b/tests/unit/test_expression_tree/test_functions.py
@@ -76,22 +76,24 @@ class TestFunction(TestCase):
         self.assertEqual(func.to_equation(), sympy.Symbol("test"))
 
         # Test Arcsinh
-        self.assertEqual(pybamm.Arcsinh(a).to_equation(), sympy.asinh(a))
+        self.assertEqual(pybamm.Arcsinh(a).to_equation(), sympy.asinh("a"))
 
         # Test Arctan
-        self.assertEqual(pybamm.Arctan(a).to_equation(), sympy.atan(a))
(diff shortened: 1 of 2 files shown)


kratman:
When this passes CI then it should fix the CI in #4250 and replace #4252

brosaplanella:
Looks good, happy to approve once coverage passes.

kratman:
@brosaplanella I am pretty sure that failure is a false positive. It is the same failure in #4252, which does not change any code

valentinsulzer (approved):

PR #4254, opened 2024-07-09
Ecm with diffusion

MehrdadBabazadeh:
# Description
The existing Equivalent Circuit Model (ECM) has been extended to create a more accurate model named the ECMD (ECM+Diffusion). This improvement involved adding a "diffusion_element" to the circuit alongside updates to the voltage_model and other relevant files. The key difference lies in how terminal voltage is calculated. The original ECM relied on average State of Charge (SoC). With the diffusion option enabled, the ECMD utilizes a newly calculated Surface SoC, providing a more precise representation. Additionally, a new term, eta_diffusion, is introduced. This term represents the voltage drop caused by diffusion limitations and is placed in series with the existing Open Circuit Voltage (OCV) within the circuit. This approach isolates the pure effect of diffusion on the voltage calculations. Finally, a Python script named Compare_ECM_ECMD.py is included as an example to demonstrate the comparison between the original ECM and the new ECMD.

## Type of change

- [*] New feature (non-breaking change which adds functionality)

# Key checklist:

- [*] No style issues: `$ pre-commit run` (or `$ nox -s pre-commit`) (see [CONTRIBUTING.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CONTRIBUTING.md#installing-and-using-pre-commit) for how to set this up to run automatically when committing locally, in just two lines of code)

## Further checks:

- [*] Code is commented, particularly in hard-to-understand areas

### DIFF
--- a/examples/scripts/Compare_ECM_ECMD.py
+++ b/examples/scripts/Compare_ECM_ECMD.py
@@ -1,0 +1,51 @@
+import pybamm
+
+model1 = pybamm.equivalent_circuit.Thevenin(options={"diffusion element": "false"})   # Model1: ECM Model definition
+model2 = pybamm.equivalent_circuit.Thevenin(options={"diffusion element": "true"})    # Model2: ECMD Model definition
+
+parameter_values1 = model1.default_parameter_values
+parameter_values2 = model2.default_parameter_values
+parameter_values2.update({"Diffusion time constant [s]": 580}, check_already_exists=False)
+
+experiment = pybamm.Experiment(
+    [
+        (
+            "Discharge at C/10 for 10 hours or until 3.3 V",
+            "Rest for 30 minutes",
+            "Rest for 2 hours",
+            "Charge at 100 A until 4.1 V",
+            "Hold at 4.1 V until 5 A",
+            "Rest for 30 minutes",
+            "Rest for 1 hour",
+        ),
+    ]
+     
+)
+
+models = [
+    model1,
+    model2,
+]
+
+sims = []
+for model, parameter_values in zip([model1, model2], [parameter_values1, parameter_values2]):
+    sim = pybamm.Simulation(model, experiment=experiment, parameter_values=parameter_values)
+    sim.solve()  # Solve for specific time range (0 to 3600 seconds)
+    sims.append(sim)
+
+labels = [
+    "Thevenin (No Diffusion)",
+    "Thevenin (Diffusion, τ=580s)",  # Adjust labels as needed
+]
+pybamm.dynamic_plot(sims, time_unit="seconds", labels=labels)
+
+default_params = model1.default_parameter_values
+nominal_capacity = default_params["Nominal cell capacity [A.h]"]
+print(f"Nominal Cell Capacity (Ah): {nominal_capacity}")
+
+#Second plot to compare SoC and Surface SoC
+output_variables = ["SoC", "Surface SoC"]
+
+sim_diffusion = sims[1]  # Assuming model2 is at index 1 in sims list
+sim_diffusion.plot(output_variables=output_variables)
+
--- a/pybamm/models/full_battery_models/equivalent_circuit/thevenin.py
+++ b/pybamm/models/full_battery_models/equivalent_circuit/thevenin.py
@@ -30,6 +30,9 @@ class Thevenin(pybamm.BaseModel):
                 throughput capacity in addition to discharge capacity. Must be one of
                 "true" or "false". "false" is the default, since calculating discharge
                 energy can be computationally expensive for simple models like SPM.
+            * "diffusion element" : str
+                Whether to include the diffusion element to the model. Must be one of
+                "true" or "false". "false" is the default.
             * "operating mode" : str
                 Sets the operating mode for the model. This determines how the current
                 is set. Can be:
@@ -73,6 +76,7 @@ class Thevenin(pybamm.BaseModel):
     def set_options(self, extra_options=None):
         possible_options = {
             "calculate discharge energy": ["false", "true"],
+            "diffusion element": ["false", "true"],
             "operating mode": OperatingModes("current"),
             "number of rc elements": NaturalNumberOption(1),
         }
@@ -164,6 +168,16 @@ class Thevenin(pybamm.BaseModel):
                 self.param, self.element_counter, self.options
             )
             self.element_counter += 1
+            
+    def set_diffusion_submodel(self):
+        if self.options == "false":
+            self.submodels["Diffusion"] = pybamm.equivalent_circuit_elements.NoDiffusion(
+                self.param, self.options
+            )
+        elif self.options == "true":
+            self.submodels["Diffusion"] = pybamm.equivalent_circuit_elements.DiffusionElement(
+                self.param, self.options
+            )            
 
     def set_thermal_submodel(self):
         self.submodels["Thermal"] = pybamm.equivalent_circuit_elements.ThermalSubModel(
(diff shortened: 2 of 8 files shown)


valentinsulzer:
Thanks! Is there any reference for this in the literature?

pybamm/models/submodels/equivalent_circuit_elements/ocv_element.py
```diff
@@ -56,7 +56,7 @@ def set_initial_conditions(self, variables):
 
     def set_events(self, variables):
         soc = variables["SoC"]
-        self.events = [
+        self.events += [
             pybamm.Event("Minimum SoC", soc),
             pybamm.Event("Maximum SoC", 1 - soc),
         ]
```
kratman, line 62: Every time this is called, it would append these events. I think that is a bit counter intuitive to what "set_events()" would typically mean. I would expect a set function to replace what is currently there, not add to it. Can we remove this?

kratman (changes requested):

pybamm/models/full_battery_models/equivalent_circuit/thevenin.py
```diff
@@ -30,6 +30,9 @@ class Thevenin(pybamm.BaseModel):
                 throughput capacity in addition to discharge capacity. Must be one of
                 "true" or "false". "false" is the default, since calculating discharge
                 energy can be computationally expensive for simple models like SPM.
+            * "diffusion element" : str
+                Whether to include the diffusion element to the model. Must be one of
+                "true" or "false". "false" is the default.
             * "operating mode" : str
```
kratman, line 36: Can we make this a Boolean rather than a string?

valentinsulzer: We make all options strings so that the types are consistent

MehrdadBabazadeh:
> Thanks! Is there any reference for this in the literature?

Hi. The used references are in the following. We have validated the results with an implementation of Chebyshev Collocation method developed by separate Python code:
1. Fan, Chuanxin, et al. "Data-driven identification of lithium-ion batteries: A nonlinear equivalent circuit model with diffusion dynamics." Applied Energy 321 (2022): 119336.
2. Binous, Housam, Abdullah A. Shaikh, and Ahmed Bellagi. "Chebyshev orthogonal collocation technique to solve transport phenomena problems with matlab® and mathematica." Computer Applications in Engineering Education 23.3 (2015): 422-431.

The address for the developed Python code, used to validate SoC profile is:
https://github.com/MehrdadBabazadeh/Chebyshev-Collocation.git

pybamm/models/submodels/equivalent_circuit_elements/ocv_element.py
brosaplanella: I suggested this change as I was afraid that otherwise, if this submodel was processed after some events had already been defined it would erase them. Not sure what's the best way forward.

kratman: This adds a side effect to what would be expected to be a simple setter. At minimum it should be changed to `append_events()` or `add_events()`. Was there an actual bug that this was fixing or just not sure if it would cause issues?

pybamm/models/submodels/equivalent_circuit_elements/diffusion_element.py
```diff
@@ -0,0 +1,109 @@
...
+            z: {"left": (lbc, "Neumann"), "right": (rbc, "Neumann")}
+        }
+
+    def set_initial_conditions(self, variables):
+        z = variables["Distributed SoC"]
+        self.initial_conditions = {z: self.param.initial_soc}
+
+    def set_events(self, variables):
+        z_surf = variables["Surface SoC"]
+        self.events += [
```
kratman, line 106: I just noticed that the same append behavior is in this version of the function. Same as below, at minimum the function name needs to change, but if there is not known issue then can it stay as a basic setter?

valentinsulzer:
@MehrdadBabazadeh thanks for sharing the references, can you document these as described here? https://github.com/pybamm-team/PyBaMM/blob/develop/CONTRIBUTING.md#citations 

Also, you can link that git repository in the docstring of your submodel. These references will be useful in future if someone wants to know more about the model

pybamm/models/submodels/equivalent_circuit_elements/diffusion_element.py
valentinsulzer: I'm guessing @MehrdadBabazadeh followed how we do this in several other places, which I agree is the wrong way to do it. We're pretty haphazard throughout the submodels codebase about creating new dictionaries vs updating the pre-defined empty dictionaries. One way to enforce the better approach you suggest here would be to not initialize an empty dictionary at all

brosaplanella: Yes, the `set_events` function is pretty much everywhere. For this PR, I just suggested to Mehrdad to replace the `=` to a `+=` to avoid overwriting existing events. Going forward it would be probably better to have events as a dictionary, which would also make it easier to access specific events if the users want to change anything. But I would suggest doing this a separate PR.

pybamm/models/submodels/equivalent_circuit_elements/ocv_element.py
brosaplanella: It was not sure, but before only the OCV component defined events, while now there are the OCV and the diffusion components.

examples/scripts/Compare_ECM_ECMD.py
```diff
@@ -0,0 +1,59 @@
...
+
+sims = []
+for model, parameter_values in zip(
+    [model1, model2], [parameter_values1, parameter_values2]
+):
+    sim = pybamm.Simulation(
+        model, experiment=experiment, parameter_values=parameter_values
+    )
+    sim.solve()  # Solve for specific time range (0 to 3600 seconds)
+    sims.append(sim)
```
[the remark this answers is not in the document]
brosaplanella: We just had a chat with Mehrdad, and agreed this example is not needed anymore as that was for validation. We already have the other example script on how to run ECMD, which should be enough.

pybamm/models/submodels/equivalent_circuit_elements/diffusion_element.py
```diff
@@ -0,0 +1,110 @@
...
+    ----------
+    param : parameter class
+        The parameters to use for this submodel
+    options : dict, optional
+        A dictionary of options to be passed to the model.
+    """
+
+    def __init__(self, param, options=None):
+        super().__init__(param)
+        pybamm.citations.register("Fan2022")
```
kratman, line 19: The paper mentions ECM with diffusion, and it is cited in the `NoDiffusion` class. I would expect this citation to be used in the `DiffusionElement` class. Should it be in both?

kratman (commented):
One last thing now that #4266 was created to address `set_events` being misnamed.

pybamm/models/submodels/equivalent_circuit_elements/diffusion_element.py
MehrdadBabazadeh: We fixed it now!

kratman (approved):
Seems OK now as long as the tests pass

kratman:
@MehrdadBabazadeh Are you still working on the coverage for this?

MehrdadBabazadeh:
Yes. Ferran and me will go on the remaining issues and continue the work next week.

brosaplanella:
The coverage issues seem to be unrelated to this PR, so that should be good to go. Waiting to merge as I don't know if we need to wait for #4311

arjxn-py:
> I don't know if we need to wait for #4311

@brosaplanella, we can merge this one without waiting for #4311. Coverage failure seems unrelated to me too as giving failure in `pybamm/util.py` which is [failing for old changes](https://app.codecov.io/gh/pybamm-team/PyBaMM/pull/4254/blob/pybamm/util.py)

brosaplanella:
@all-contributors please add @MehrdadBabazadeh for code and tests

MehrdadBabazadeh:
Thanks to @brosaplanella and other team members!

PR #4263, opened 2024-07-13
Fix directory in testpypi upload step

Saransh-cpp:
# Description

@kratman changes should not be made directly to the release branch. The changes should only be carried from the develop branch to the release branch.

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:


kratman:
I made the fix to the release workflow to make it work, just like we did when doing the training rc0. I figured it was not a big deal since rc0 got merged into develop shortly after

Saransh-cpp:
It is not a big deal imo, but it might cause the branches to diverge. If everything flows from develop to other branches then we know for sure that all the new code is going into develop and the other branches are just following it.

PR #4179, opened 2024-06-14
Fixes #4176 electrode diffusivity rename

BradyPlanden:
# Description

Fixes #4176. In addition to this issue, in the `develop` implementation changes to `electrode diffusivity` via `parameter_values.update()` will not be captured. To account for this, the I've taken the depreciated key and appended the new key into parameter_set dictionary in the event that the depreciated parameter is supplied. This also cleans up the `bpx.py` implementation since the logic for changing the key is moved into the parameters class.

Fixes # (issue)

## Type of change

Please add a line in the relevant section of [CHANGELOG.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CHANGELOG.md) to document the change (include PR #) - note reverse order of PR #s. If necessary, also add to the list of breaking changes.

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (back-end change that speeds up the code)
- [X] Bug fix (non-breaking change which fixes an issue)

# Key checklist:

- [ ] No style issues: `$ pre-commit run` (or `$ nox -s pre-commit`) (see [CONTRIBUTING.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CONTRIBUTING.md#installing-and-using-pre-commit) for how to set this up to run automatically when committing locally, in just two lines of code)
- [ ] All tests pass: `$ python run-tests.py --all` (or `$ nox -s tests`)
- [ ] The documentation builds: `$ python run-tests.py --doctest` (or `$ nox -s doctests`)

You can run integration tests, unit tests, and doctests together at once, using `$ python run-tests.py --quick` (or `$ nox -s quick`).

## Further checks:

- [ ] Code is commented, particularly in hard-to-understand areas
- [ ] Tests added that prove fix is effective or that feature works

PR #4267, opened 2024-07-15
Corrects "electrode diffusivity" error catch

### LINKED ISSUE #4176: [Bug]: "electrode diffusivity" -> "particle diffusivity" is still breaking change on `develop` when using pre-24.5 pybamm.ParameterValues
### PyBaMM Version

develop

### Python Version

3.9.13

### Describe the bug

Since #4072 re-inverted the logic fix in #3871, the release candidate on `develop` causes the rename of "electrode diffusivity" -> "particle diffusivity" to be a breaking change. Parameter sets using the old key names do not simulate in the 24.5 RC.

### Steps to Reproduce

```python
import pybamm

# Dummy a pybamm.ParameterValues object with a legacy parameter name
parameter_values.update(
    {
        "Negative electrode diffusivity [m2.s-1]": parameter_values["Negative particle diffusivity [m2.s-1]"],
    },
    check_already_exists=False,
)
del parameter_values["Negative particle diffusivity [m2.s-1]"]

# Try to run a simulation with the old-style parameter set
    [
        "Discharge at 1C until 3 V",
    ]
)
sim = pybamm.Simulation(
    model,
    experiment=experiment,
)

# KeyError provoked here on call into pybamm.ParameterValues with new key name
sol = sim.solve()
```

### Relevant log output

_No response_

BradyPlanden:
# Description

This reverts an incorrect change applied in #4179. This adds to the error catch for the old parameter name `"{domain} electrode diffusivity [m2.s-1]"` to use the new name, likewise if the new name `"{domain} particle diffusivity [m2.s-1]" `is provided the error catch uses the old name.

I've confirmed it fixes the MWE in #4176, and currently works with the old parameter name within [PyBOP](https://github.com/pybop-team/PyBOP/actions/runs/9944318010). A quick MWE to test this fix is below:

```python
import pybamm
# create the model

# set the default model parameters

# use both parameter names
neg_old = param["Negative electrode diffusivity [m2.s-1]"] = 2.0
pos_new = param["Positive particle diffusivity [m2.s-1]"] = 2.0
assert neg_old == pos_new
```

Fixes #4176

## Type of change

# Key checklist:

## Further checks:


pybamm/util.py
```diff
@@ -57,14 +57,22 @@ def __getitem__(self, key):
...
-                warn(
-                    f"The parameter '{key.replace('particle', 'electrode')}' "
-                    f"has been renamed to '{key}'",
-                    DeprecationWarning,
-                    stacklevel=2,
+            if "electrode diffusivity" in key or "particle diffusivity" in key:
+                old_term, new_term = (
+                    ("electrode", "particle")
+                    if "electrode diffusivity" in key
+                    else ("particle", "electrode")
```
brosaplanella, line 64: In which cases would we want to have the old version?

BradyPlanden: It's to ensure the parameter name change is not breaking. This allows users to continue with the old version until we deprecate it in a future release.

pybamm/util.py
```diff
@@ -57,14 +57,22 @@ def __getitem__(self, key):
...
                 )
-                return super().__getitem__(key.replace("particle", "electrode"))
+                alternative_key = key.replace(old_term, new_term)
+
+                if old_term == "electrode":
+                    warn(
+                        f"The parameter '{alternative_key}' has been renamed to '{key}'",
```
agriyakhetarpal, line 70:
```suggestion
                        f"The parameter '{alternative_key}' has been renamed to '{key}' and will be removed in a future release. Using '{key}'",
```

Would it make sense to let users know which parameter is being used regardless of the one they input?

agriyakhetarpal (approved):
This LGTM, thanks, @BradyPlanden – just one question since this was deprecated before the v24.1 release cycle and thus with the release of v24.1 (but incorrectly as far as I understand), so the plan would have been to remove it in v25.1 (three releases). Since the behaviour change is being made "not breaking" now but has existed beforehand, when should/could this be deprecated? It would help reduce the code for dependents like PyBOP where it has caused much trouble, I assume, so there's a case to be made for deprecating earlier.

kratman (approved):
Seems fine

PR #4290, opened 2024-07-24
replacing rounded faraday constant with exact value in bpx.py

dion-w:
# Description

The faraday constant has an exact value, calculated by the exact values of the avogadro constant and the elementray charge [[see here](https://www.nist.gov/si-redefinition/meet-constants)]. Using a rounded value may introduce systemic errors further down the line that can be easily avoided with this new value.

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:

- [not needed] Tests added that prove fix is effective or that feature works


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ## Features
 
+- replaced rounded faraday constant with it's exact value in bpx.py for better comparison between different tools
```
agriyakhetarpal, line 7:
```suggestion
- Replaced rounded faraday constant with its exact value in `bpx.py` for better comparison between different tools
```

agriyakhetarpal: Additionally, could you please move it to the unreleased section, @dion-w?

dion-w: yeah sure

arjxn-py (approved):
Thanks @dion-w, looks good

rtimms:
FYI you can do `pybamm.constants.F` or `pybamm.constants.F.value` if you need a number instead of a symbol.

kratman:
> FYI you can do `pybamm.constants.F` or `pybamm.constants.F.value` if you need a number instead of a symbol.

I agree, using a constant from a library is better than updating this file.

One other thing to check would be if the BPX standard specifies a value for Faraday constant. Sometimes standards specify the precision to ensure that things are constant between implementations

pybamm/parameters/bpx.py
```diff
@@ -240,7 +240,7 @@ def _entropic_change(sto, c_s_max, dUdT, constant=False):
     # (cs-cs_max)) in BPX exchange current is defined j0 = F * k_norm * sqrt((ce/ce0) *
     # (cs/cs_max) * (1-cs/cs_max))
     c_e = pybamm_dict["Initial concentration in electrolyte [mol.m-3]"]
-    F = 96485
+    F = 96485.3321233100184
```
kratman, line 243: If the BPX spec does not specify a value, then this is better
```suggestion
    F = pybamm.constants.F.value
```

Note: I did not test this myself

kratman (changes requested):

agriyakhetarpal:
Re-requested review/approval from @kratman since he previously requested changes

kratman (approved):

PR #4294, opened 2024-07-26
Adding a plugin to annotate pytest failures on GitHub

prady0t:
# Description

Using `pytest-github-actions-annotate-failures` to annotate failures on GitHub.


noxfile.py
```diff
@@ -78,6 +81,12 @@ def run_integration(session):
     """Run the integration tests."""
     set_environment_variables(PYBAMM_ENV, session=session)
     session.install("setuptools", silent=False)
+    if (
+        "CI" in os.environ
+        and sys.version_info[:2] == (3, 12)
+        and sys.platform == "linux"
+    ):
+        session.install("pytest-github-actions-annotate-failures")
```
agriyakhetarpal, line 89:
```suggestion
```

I think you forgot to remove these lines from the integration tests?

prady0t: If we remove these lines, integration test failures wont be annotated since coverage runs just the unit tests.
See : https://github.com/pybamm-team/PyBaMM/blob/48b4e48d776ad2c254e5fae8db94706c57a8cc38/noxfile.py#L150

agriyakhetarpal: Oops – that slipped my mind. Thanks!

agriyakhetarpal (approved):
Thanks, @prady0t! I'm linking previous PRs where this was tested: https://github.com/prady0t/PyBaMM/pull/3 and https://github.com/prady0t/PyBaMM/pull/2. The PRs don't contain annotations anymore, so if we want an additional test, we can do so here or fix things by following up.

noxfile.py
```diff
@@ -78,6 +81,12 @@ def run_integration(session):
     """Run the integration tests."""
     set_environment_variables(PYBAMM_ENV, session=session)
     session.install("setuptools", silent=False)
+    if (
+        "CI" in os.environ
+        and sys.version_info[:2] == (3, 12)
+        and sys.platform == "linux"
```
arjxn-py, line 87: The annotations will be same for all platforms, right?

agriyakhetarpal: Yes, they will be the same, and we found that not restricting to one platform and one Python version introduces redundant annotations (14 annotations for one test failure). This won't help where tests fail on, say, Python 3.11 but pass on Python 3.12, but it's the best one we have for now – I think

arjxn-py: Sounds Good.

arjxn-py (approved):
Thanks @prady0t

prady0t:
See https://github.com/prady0t/PyBaMM/pull/3 for how annotations are working

Saransh-cpp (approved):

PR #4305, opened 2024-07-30
Address RTD deprecation for build-time Sphinx context injection 

agriyakhetarpal:
# Description

This PR performs the action item received by the PyBaMM RTD deployment admins via email. The Read the Docs add-ons [will be enabled by default](https://about.readthedocs.com/blog/2024/07/addons-by-default/) for all projects post October 07, 2024 – we've been using them in their beta stage already.

The snippet I added uses the [`READTHEDOCS_CANONICAL_URL`](https://docs.readthedocs.io/page/reference/environment-variables.html#envvar-READTHEDOCS_CANONICAL_URL) environment variable to define [`html_baseurl`](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_baseurl) and injects [`READTHEDOCS`](https://docs.readthedocs.io/page/reference/environment-variables.html#envvar-READTHEDOCS) variable into the [`html_context`](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_context) to keep the previous behaviour.

## Type of change

# Key checklist:

## Further checks:


agriyakhetarpal:
Requested a review from the rest of the RTD admins – I don't think anything should break.

docs/conf.py
```diff
@@ -176,13 +176,21 @@
...
+        "github_version": "develop",
+        "doc_path": "docs/",
+    }
+)
+
+# Set canonical URL from the Read the Docs Domain
+html_baseurl = os.environ.get("READTHEDOCS_CANONICAL_URL", "")
 
+# Tell Jinja2 templates the build is running on Read the Docs
+if os.environ.get("READTHEDOCS", "") == "True":
```
arjxn-py, line 192:
```suggestion
if os.getenv("READTHEDOCS") == "True":
```
This should be ok too, right?

arjxn-py (commented):
Looks good. Thanks Agriya.
A couple of suggestions below.

docs/conf.py
```diff
@@ -176,13 +176,21 @@
...
+    {
+        "github_user": "pybamm-team",
+        "github_repo": "pybamm",
+        "github_version": "develop",
+        "doc_path": "docs/",
+    }
+)
+
+# Set canonical URL from the Read the Docs Domain
+html_baseurl = os.environ.get("READTHEDOCS_CANONICAL_URL", "")
```
arjxn-py, line 189:
```suggestion
html_baseurl = os.getenv("READTHEDOCS_CANONICAL_URL", "")
```

If we decide to use `os.getenv` below, we should use the same here too. Makes it a little more readable.

Missed this one in the last review.

agriyakhetarpal: I would have liked to push back on these suggestions since readability shouldn't be a concern with an API difference, and they mean the same thing, see here:

https://github.com/python/cpython/blob/ef21e48cac548025aa76704b84b489f0c111e1de/Lib/os.py#L807-L813

but I won't wish to hold up the PR :)

agriyakhetarpal: Applied the suggestion

arjxn-py (approved):
Thanks @agriyakhetarpal

PR #4205, opened 2024-06-23
Moving from flat layout to src layout.

prady0t:
# Description

Moving to a src layout is a better option overall. You can read more about it here: https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/#:~:text=Auto%20color%20theme-,src%20layout%20vs%20flat%20layout,-%23

It was decided internally to move to a `src` layout. 

## Type of change

Please add a line in the relevant section of [CHANGELOG.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CHANGELOG.md) to document the change (include PR #) - note reverse order of PR #s. If necessary, also add to the list of breaking changes.

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (back-end change that speeds up the code)
- [ ] Bug fix (non-breaking change which fixes an issue)

# Key checklist:

- [ ] No style issues: `$ pre-commit run` (or `$ nox -s pre-commit`) (see [CONTRIBUTING.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CONTRIBUTING.md#installing-and-using-pre-commit) for how to set this up to run automatically when committing locally, in just two lines of code)
- [ ] All tests pass: `$ python run-tests.py --all` (or `$ nox -s tests`)
- [ ] The documentation builds: `$ python run-tests.py --doctest` (or `$ nox -s doctests`)

You can run integration tests, unit tests, and doctests together at once, using `$ python run-tests.py --quick` (or `$ nox -s quick`).

## Further checks:

- [ ] Code is commented, particularly in hard-to-understand areas
- [ ] Tests added that prove fix is effective or that feature works

PR #4311, opened 2024-08-02
Moving to src layout

prady0t:
# Description

Related to #4205

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:


agriyakhetarpal (approved):
Thanks, @prady0t! Could you trigger the wheel builds from this branch on your fork and link the workflow run here? I shall merge this in 24 hours unless someone raises any objections. Note that if any other PR gets merged before this one does, there might be conflicts again, so it will be better to resolve them here (or learn how to).

Saransh-cpp (approved):
Thanks, @prady0t! Looks good to me. I am not sure why the tests are failing but I'll try taking a look soon.

agriyakhetarpal:
The tests are failing because

`pybamm/solvers/c_solvers/idaklu/Expressions/Casadi/CasadiFunctions.cpp`

is not found. To fix this, these lines have to be changed, which were missed:

https://github.com/pybamm-team/PyBaMM/blob/8983f43a9f2ce54c6afa14e2e06edb250d426dc4/CMakeLists.txt#L46-L54

agriyakhetarpal:
Note: the paths in `scripts/update_version.py` will likely also be broken. Let's ensure we test this thoroughly throughout the codebase from start to end, identify potential breakages, and fix them here.

agriyakhetarpal (changes requested):
A summary of the required changes:

1. ✅ Path to `pybamm.root_dir()`
2. ✅ Changes in `scripts/update_version.py`, and its corresponding `update_version.yml` file as needed
3. ✅ Paths to `CITATIONS.bib` in `pyproject.toml` and elsewhere
4. ✅ Changes in `[tool.setuptools.package-data]`
5. ✅ Changes to paths in `conf.py` (`bibtex_bibfiles`)
6. `citations.py`
7. ✅ Paths to `version.py`
8. ✅ Changes in `CONTRIBUTING.md`
9. ✅ A link to passing wheel builds triggered from this branch on your fork

There is a chance I might have missed something, so please feel free to suggest more.

prady0t:
@agriyakhetarpal I've tried adding all the changes. Here's the workflow for wheels build : https://github.com/prady0t/PyBaMM/actions/runs/10228436876
Lets see how it goes

prady0t:
Tests are still failing on the local branch so they will fail here too. Let me look into it further.

agriyakhetarpal:
Thanks, nice to see wheels passing – besides fixing the CI failures, don't forget the rest of the points!

agriyakhetarpal:
Let's focus on getting this PR merged before any of your other PRs. I posted about this PR in the #maintainers channel on Slack,nand there seem to be no objections, so we will be good to go once the changes are addressed.

prady0t:
https://github.com/prady0t/PyBaMM/pull/4
local branch tests are passing except for ubuntu

noxfile.py
```diff
@@ -180,7 +180,7 @@ def run_doctests(session):
         "-m",
         "pytest",
         "--doctest-plus",
-        "pybamm",
+        "src/pybamm",
```
agriyakhetarpal, line 183:
```suggestion
        "src",
```

This can be just `src` and you would not need to deal with different slashes for paths (such as on Windows)

agriyakhetarpal (changes requested):
Keep going, this is almost there! To fix the failing builds, these paths need to be fixed:

https://github.com/pybamm-team/PyBaMM/blob/109b9c442e8df77224bbfcf2a7909c69ec11e848/CMakeLists.txt#L62-L75

Also, things like `scripts/update_version.py` and `pybamm.root_dir()`, etc., from the previous points are left to be updated; I checked off the items that have been addressed.

prady0t:
`pybamm.root_dir()` automatically returns `PyBaMM/src` do we still need to change it?

agriyakhetarpal:
> `pybamm.root_dir()` automatically returns `PyBaMM/src` do we still need to change it?

Yes, we need to go one level back – it returns the repository's root directory (`PyBaMM/`), and we call the function in a few tests and scripts, so it would be better to modify those instead of modifying the output of a public function (even if not many people are using it)

prady0t:
> > `pybamm.root_dir()` automatically returns `PyBaMM/src` do we still need to change it?
> 
> Yes, we need to go one level back – it returns the repository's root directory (`PyBaMM/`), and we call the function in a few tests and scripts, so it would be better to modify those instead of modifying the output of a public function (even if not many people are using it)

Makes sense.

prady0t:
This should cover everything. Lets wait for CI jobs

agriyakhetarpal:
From the RTD failure:
```
pybtex.exceptions.PybtexError: unable to open /home/docs/checkouts/readthedocs.org/user_builds/pybamm/envs/4311/lib/python3.12/src/pybamm/CITATIONS.bib. No such file or directory
```

This should be easy to fix and should be related to point 6

agriyakhetarpal:
Triggered it from the admin dashboard, let's see if it fails again

agriyakhetarpal:
I think the SPM notebook's failure is related to `os.chdir(pybamm.__path__[0] + "/..")` in the first cell. The coverage failure is unrelated. I don't understand the RTD failure right now. I think point 8 in the requested changes still stands – the links to the `CITATIONS.bib` file in `README.md` and in `CONTRIBUTING.md` will break as soon as this PR is merged into `develop`, so we should change the links here and let the link checker fail on that basis.

src/pybamm/citations.py
```diff
@@ -68,7 +68,9 @@ def read_citations(self):
         """
         try:
             parse_file = import_optional_dependency("pybtex.database", "parse_file")
-            citations_file = os.path.join(pybamm.root_dir(), "pybamm", "CITATIONS.bib")
+            citations_file = os.path.join(
+                pybamm.root_dir(), "src", "pybamm", "CITATIONS.bib"
+            )
```
agriyakhetarpal, line 73: I figured out the problem with the failing RTD builds, and it's coming from these lines:

Note the location of the `CITATIONS.bib` file in the logs:

```
FileNotFoundError: [Errno 2] No such file or directory: '/home/docs/checkouts/readthedocs.org/user_builds/pybamm/envs/4311/lib/python3.12/src/pybamm/CITATIONS.bib'
```

i.e., the path is `lib/python3.12/src/pybamm/CITATIONS.bib`, which is obviously incorrect – it's not going to exist. The correct path to check should be `lib/python3.12/site-packages/pybamm/CITATIONS.bib`. 

This means that the output from `pybamm.root_dir()` was always incorrect in the first place, and the problem has just been exposed when migrating to an `src` layout. The reason why this doesn't fail outside of the RTD builds is because we use an editable installation with `nox` in the PR checks (the `-e` command-line flag), while the RTD builds don't install in editable mode. You can verify for yourself and reproduce the same error locally:

``` bash
# works
pip install -e .
python -c "import pybamm; print(pybamm.root_dir())"
# fails
pip install .
python -c "import pybamm; print(pybamm.root_dir())"
```

I notice that `CITATIONS.bib` exists in the same folder as `citations.py`. Maybe you could modify these lines to fix this?

agriyakhetarpal (changes requested):

src/pybamm/citations.py
prady0t: Thanks for the suggestions. I was able to reproduce this locally.

prady0t:
@agriyakhetarpal RTD failure is now resolved.

agriyakhetarpal:
I don't think `"site-packages"` would fix it – I triggered the workflows, but the editable installations are going to raise an error now.

prady0t:
I see. Let me try something else

examples/scripts/drive_cycle.py
```diff
@@ -5,7 +5,7 @@
 import pandas as pd
 import os
 
-os.chdir(pybamm.__path__[0] + "/..")
+os.chdir(pybamm.__path__[0] + "/.." + "/..")
```
agriyakhetarpal, line 8: Also, could you check if this line is needed anymore? If not, we should remove it unless we have inclinations to keep it to historical reasons :)

```suggestion
```

examples/scripts/experiment_drive_cycle.py
```diff
@@ -5,7 +5,7 @@
 import pandas as pd
 import os
 
-os.chdir(pybamm.__path__[0] + "/..")
+os.chdir(pybamm.__path__[0] + "/.." + "/..")
```
agriyakhetarpal, line 8:
```suggestion
```

Same as above

prady0t:
We can do something like :
```
            if os.path.isfile(os.path.join(pybamm.root_dir(), "pybamm.root_dir()", "site-packages", "pybamm", "CITATIONS.bib")):
                citations_file = os.path.join(
                    pybamm.root_dir(), "site-packages", "pybamm", "CITATIONS.bib"
                )
            else:
                citations_file = os.path.join(pybamm.root_dir(), "src", "pybamm", "CITATIONS.bib")
                
```
Do you think it's correct? We can also use : https://stackoverflow.com/questions/122327/how-do-i-find-the-location-of-my-python-site-packages-directory#:~:text=python3%20%2Dc%20%27import%20sysconfig%3B%20print(sysconfig.get_paths()%5B%22purelib%22%5D)%27
to obtain file location.

agriyakhetarpal:
How about avoiding `pybamm.root_dir()` altogether and just using `pybamm.__path__[0]`? That way, we can avoid an if-else condition.

examples/scripts/drive_cycle.py
prady0t: yes we can

noxfile.py
agriyakhetarpal: I think this is still a valid suggestion

agriyakhetarpal (approved):
Thanks, @prady0t, this LGTM now! As a note to other reviewers: the failing links are expected and will be resolved once this PR goes into `develop`.

noxfile.py
prady0t: Sorry must have missed it.

arjxn-py (approved):
Thanks @prady0t, looks good

agriyakhetarpal:
The example notebook looks like it is still failing: https://github.com/pybamm-team/PyBaMM/actions/runs/10300421066/job/28509802852?pr=4311

prady0t:
Is it related to changes in this PR?

agriyakhetarpal:
Not sure what it was, but I removed the custom path because the PNGs are in the same directory as the notebook

PR #4319, opened 2024-08-05
Update simulation class private members

BradyPlanden:
# Description

A small PR to remove usage of the simulation class properties, within the class.

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybamm/simulation.py
+++ b/pybamm/simulation.py
@@ -243,7 +243,7 @@ class Simulation:
                     initial_soc, param=param, inplace=False, options=options
                 )
             )
-            # elif options["working electrode"] == "positive":
+        elif options["working electrode"] == "positive":
             self._parameter_values = (
                 self._unprocessed_parameter_values.set_initial_stoichiometry_half_cell(
                     initial_soc,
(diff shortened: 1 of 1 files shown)


valentinsulzer (commented):
Don't see any problem with doing this but what is wrong with the other way out of interest?

BradyPlanden:
Fair question, I'd say it mainly removes confusion for developers and users looking through the codebase. Since the attributes I've corrected were using the property, we're presenting these objects as modifiable outside the class without issues, which is not the case (or if it is, we can remove them as private attrs and the corresponding properties) while also having the private attributes used elsewhere in the class (i.e. mixing them up throughout the class).

This actually came from a PyBOP PR, as we are using a few of the simulation class methods during build. One of the attributes (`self.model`) uses both the property and the private attribute, so we needed to do this in one spot. This will clean things up for both packages.

As a side note, PyBOP is moving towards doing this automatically with ruff: https://github.com/pybop-team/PyBOP/pull/435 I suspect doing something similar would be beneficial for PyBaMM as well; however, it takes a bit of work to fix everything at the beginning..

valentinsulzer (approved):
Thanks for the info, definitely sounds like a useful check to include

agriyakhetarpal (approved):

PR #3901, opened 2024-03-18
Migrated from miniconda to official python bullseye docker image

santacodes:
# Description

This draft is in continuation to #3874 which was closed due to messy branches and the image failing to build IDAKLU solver on aarch64 linux. Initially the idea was to migrate to a single docker image of manylinux2014 which was compliant to [PEP599](https://peps.python.org/pep-0599/), but due to some complications, the official python image along with venv were decided upon implementation, check #3874 discussions for more information.

Fixes #3692
Fixes #3666 

## Type of change

Please add a line in the relevant section of [CHANGELOG.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CHANGELOG.md) to document the change (include PR #) - note reverse order of PR #s. If necessary, also add to the list of breaking changes.

- [x] New feature (non-breaking change which adds functionality)
- [ ] Optimization (back-end change that speeds up the code)
- [ ] Bug fix (non-breaking change which fixes an issue)

# Key checklist:

- [x] No style issues: `$ pre-commit run` (or `$ nox -s pre-commit`) (see [CONTRIBUTING.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CONTRIBUTING.md#installing-and-using-pre-commit) for how to set this up to run automatically when committing locally, in just two lines of code)
- [x] All tests pass: `$ python run-tests.py --all` (or `$ nox -s tests`)
- [x] The documentation builds: `$ python run-tests.py --doctest` (or `$ nox -s doctests`)

You can run integration tests, unit tests, and doctests together at once, using `$ python run-tests.py --quick` (or `$ nox -s quick`).

## Further checks:

- [x] Code is commented, particularly in hard-to-understand areas
- [ ] Tests added that prove fix is effective or that feature works

PR #4323, opened 2024-08-07
Fixed JAX solver compatibility test

santacodes:
# Description

I caught this while running `nox -s unit` and also in CI [here](https://github.com/pybamm-team/PyBaMM/actions/runs/10285560857/job/28464203323).
Instead of checking the compatibility of the JAX solver the test was checking if JAX solver was present and logging `JAX is not installed`, when it should log something like `JAX is compatible` when the check passes inside the `skipif` marker of pytest.

```bash
SKIPPED [1] tests/unit/test_solvers/test_idaklu_jax.py:91: Both IDAKLU and JAX are available
SKIPPED [1] tests/unit/test_util.py:91: The JAX solver is not installed
```

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/tests/unit/test_util.py
+++ b/tests/unit/test_util.py
@@ -88,6 +88,10 @@ class TestUtil:
             path = os.path.join(package_dir, tempfile_obj.name)
             assert pybamm.get_parameters_filepath(tempfile_obj.name) == path
 
+@pytest.mark.skipif(not pybamm.have_jax(), reason="JAX is not installed")
+def test_is_jax_compatible():
+    compatible = pybamm.is_jax_compatible()
+    assert compatible
     def test_git_commit_info(self):
         git_commit_info = pybamm.get_git_commit_info()
         assert isinstance(git_commit_info, str)
--- a/tests/unit/test_util.py
+++ b/tests/unit/test_util.py
@@ -88,12 +88,6 @@ class TestUtil:
             path = os.path.join(package_dir, tempfile_obj.name)
             assert pybamm.get_parameters_filepath(tempfile_obj.name) == path
 
-    @pytest.mark.skipif(
-        pybamm.is_jax_compatible(), reason="The JAX solver is compatible"
-    )
-    def test_is_jax_compatible(self):
-        assert True
-
     def test_git_commit_info(self):
         git_commit_info = pybamm.get_git_commit_info()
         assert isinstance(git_commit_info, str)
--- a/tests/unit/test_util.py
+++ b/tests/unit/test_util.py
@@ -88,10 +88,12 @@ class TestUtil:
             path = os.path.join(package_dir, tempfile_obj.name)
             assert pybamm.get_parameters_filepath(tempfile_obj.name) == path
 
+
 @pytest.mark.skipif(not pybamm.have_jax(), reason="JAX is not installed")
 def test_is_jax_compatible():
     compatible = pybamm.is_jax_compatible()
     assert compatible
+
     def test_git_commit_info(self):
         git_commit_info = pybamm.get_git_commit_info()
         assert isinstance(git_commit_info, str)
--- a/tests/unit/test_util.py
+++ b/tests/unit/test_util.py
@@ -88,7 +88,6 @@ class TestUtil:
             path = os.path.join(package_dir, tempfile_obj.name)
             assert pybamm.get_parameters_filepath(tempfile_obj.name) == path
 
-
     @pytest.mark.skipif(not pybamm.have_jax(), reason="JAX is not installed")
     def test_is_jax_compatible():
         assert pybamm.is_jax_compatible()
--- a/tests/unit/test_util.py
+++ b/tests/unit/test_util.py
@@ -89,7 +89,7 @@ class TestUtil:
             assert pybamm.get_parameters_filepath(tempfile_obj.name) == path
 
     @pytest.mark.skipif(not pybamm.have_jax(), reason="JAX is not installed")
-    def test_is_jax_compatible():
+    def test_is_jax_compatible(self):
         assert pybamm.is_jax_compatible()
 
     def test_git_commit_info(self):
--- a/tests/unit/test_util.py
+++ b/tests/unit/test_util.py
@@ -89,10 +89,9 @@ class TestUtil:
             assert pybamm.get_parameters_filepath(tempfile_obj.name) == path
 
 
-@pytest.mark.skipif(not pybamm.have_jax(), reason="JAX is not installed")
-def test_is_jax_compatible():
-    compatible = pybamm.is_jax_compatible()
-    assert compatible
+    @pytest.mark.skipif(not pybamm.have_jax(), reason="JAX is not installed")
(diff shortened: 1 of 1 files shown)


tests/unit/test_util.py
```diff
@@ -88,7 +88,9 @@ def test_get_parameters_filepath(self):
             path = os.path.join(package_dir, tempfile_obj.name)
             assert pybamm.get_parameters_filepath(tempfile_obj.name) == path
 
-    @pytest.mark.skipif(pybamm.have_jax(), reason="The JAX solver is not installed")
+    @pytest.mark.skipif(
+        pybamm.is_jax_compatible(), reason="The JAX solver is compatible"
+    )
     def test_is_jax_compatible(self):
```
kratman, line 94: Shouldn't the skip behavior be `!pybamm.have_jax()`? it is testing if jax is compatible.

The current logic:
```
jax installed -> Skip test with a message that jax is not installed
jax not installed -> check if jax is compatible, if so assert true
```

New logic:
```
jax compatible -> Skip test with a message that jax is installed
jax not installed -> run a test that asserts true
```

I don't think the test is working in either setup

kratman (changes requested):

tests/unit/test_util.py
kratman: Maybe a better approach is to remove this test or rework what is actually being tested

santacodes: I think there are two different tests, one for testing if JAX is present and the other for checking if it is compatible. The presence of JAX and IDAKLU solvers are already being tested and working well, while this function had to test the compatibility and it should probably `assert False` inside the function if the test doesn't skip. So it should probably be like - 
```python 
    @pytest.mark.skipif(
        pybamm.is_jax_compatible(), reason="The JAX solver is compatible"
    )
    def test_is_jax_compatible(self):
        assert False
```
But pre-commit does not want me to assert False in that case, would there be a way to implement this logic? Also, I previously brought up this [here](https://github.com/pybamm-team/PyBaMM/pull/3901#issuecomment-2273383056).
Or probably we could implement like you mentioned 
```python 
    @pytest.mark.skipif(
        !pybamm.is_jax_compatible(), reason="The JAX solver is not compatible"
    )
    def test_is_jax_compatible(self):
        assert True
```

kratman: I think it is just a bad test. The skipping mechanism is being used for internal logic. The tests should test behavior, not that a module was installed by our installer

kratman: As a further point a assert false would make tests fail any time jax was not installed

santacodes: Should I just remove it then?

santacodes: > I think it is just a bad test. The skipping mechanism is being used for internal logic. The tests should test behavior, not that a module was installed by our installer

Also, there are tests for testing if IDAKLU and JAX are installed in that case maybe we could just make one collective test/validation of some kind for that?

kratman: There is no sense in keeping tests which don't test what you expect. This test and other similar tests can probably be removed. If it is not testing the behavior of a function, then it is probably a bad test.

Any time you see just a bare `assert True` without an accompanying logic block, then it should raise a red flag.

kratman (approved):
This works for me

agriyakhetarpal:
I think this came about as a migration done incorrectly in 68f71f278501b8bc670aed1514d4b32a817f42d4? I think we should indeed test that the JAX version is compatible (i.e., `0.4`, since we don't support `0.5`) – which was the earlier behaviour of the test.

kratman:
> I think this came about as a migration done incorrectly in [68f71f2](https://github.com/pybamm-team/PyBaMM/commit/68f71f278501b8bc670aed1514d4b32a817f42d4)? I think we should indeed test that the JAX version is compatible (i.e., `0.4`, since we don't support `0.5`) – which was the earlier behaviour of the test.

While I agree that the change was done poorly in 68f71f278501b8bc670aed1514d4b32a817f42d4, I am still in favor of removing the test. Installing the correct version of jax is the job of the build system, not the unit tests. If we need to check for it, then it should be done in the build process not in the tests after the build is done

agriyakhetarpal:
Okay, fair enough – we have `jax` and `jaxlib` pinned already, and there's a chance the tests will fail elsewhere if someone undoes the pin

agriyakhetarpal (approved):

arjxn-py (approved):

arjxn-py:
Oh, just wait for me before you merge.

arjxn-py:
```
@pytest.mark.skipif(not pybamm.have_jax(), reason="JAX is not installed")
def test_is_jax_compatible():
    compatible = pybamm.is_jax_compatible()
    assert compatible
```
This should be the correct test, @santacodes.

tests/unit/test_util.py
```diff
@@ -88,10 +88,6 @@ def test_get_parameters_filepath(self):
             path = os.path.join(package_dir, tempfile_obj.name)
             assert pybamm.get_parameters_filepath(tempfile_obj.name) == path
```
arjxn-py, line 90:
```suggestion

@pytest.mark.skipif(not pybamm.have_jax(), reason="JAX is not installed")
def test_is_jax_compatible():
    compatible = pybamm.is_jax_compatible()
    assert compatible
```

arjxn-py (changes requested):

tests/unit/test_util.py
```diff
@@ -88,9 +88,11 @@ def test_get_parameters_filepath(self):
...
 
-    @pytest.mark.skipif(pybamm.have_jax(), reason="The JAX solver is not installed")
-    def test_is_jax_compatible(self):
-        assert True
+
+@pytest.mark.skipif(not pybamm.have_jax(), reason="JAX is not installed")
+def test_is_jax_compatible():
+    compatible = pybamm.is_jax_compatible()
+    assert compatible
```
kratman, line 95: I still think this is unneeded, but if it is being put in, then it needs to be cleaned up
```suggestion
    @pytest.mark.skipif(not pybamm.have_jax(), reason="JAX is not installed")
    def test_is_jax_compatible():
        assert pybamm.is_jax_compatible()
```

arjxn-py (approved):

tests/unit/test_util.py
```diff
@@ -88,9 +88,9 @@ def test_get_parameters_filepath(self):
             path = os.path.join(package_dir, tempfile_obj.name)
             assert pybamm.get_parameters_filepath(tempfile_obj.name) == path
 
-    @pytest.mark.skipif(pybamm.have_jax(), reason="The JAX solver is not installed")
-    def test_is_jax_compatible(self):
-        assert True
+    @pytest.mark.skipif(not pybamm.have_jax(), reason="JAX is not installed")
+    def test_is_jax_compatible():
```
kratman, line 92:
```suggestion
    def test_is_jax_compatible(self):
```

PR #4330, opened 2024-08-09
ECM with split OCV

parkec3:
# Description

This creates a simple equivalent circuit model with split OCV for the two electrodes. The OCV functions are from the lithium ion parameters, and this model is housed underneath full lithium ion models.

## Type of change

# Key checklist:

## Further checks:


tests/integration/test_models/test_full_battery_models/test_lithium_ion/test_ecm_split_OCV.py
```diff
@@ -0,0 +1,58 @@
+#
+# Test that the model works with an example parameter set
+#
+import pybamm
+import numpy as np
+
+
+class TestECMSplitOCVModel:
+    def test_run_model_with_parameters(self):
```
kratman, line 9: This test does not asset anything. If a test does not assert anything then the only thing that gets tested is that the code does not crash and the coverage is artificially high.

A lot of the other models use something like this:
```
    def test_basic_processing(self):
        model = pybamm.equivalent_circuit.Thevenin()
        modeltest = tests.StandardModelTest(model)
        modeltest.test_all()
```
Is it possible to set that up instead?

docs/source/api/models/lithium_ion/ecm_split_ocv.rst
```diff
@@ -0,0 +1,7 @@
+Equivalent Circuit Model with Split OCV (ECMsplitOCV)
+=====================================================
+
+.. autoclass:: pybamm.lithium_ion.ECMsplitOCV
```
valentinsulzer, line 4: I think `pybamm.lithium_ion.SplitOCVR` would be more in line with what people would call this model

src/pybamm/models/full_battery_models/lithium_ion/basic_ecm_split_OCV.py
```diff
@@ -0,0 +1,100 @@
...
+        super().__init__(name)
+
+        ######################
+        # Variables
+        ######################
+        # All variables are only time-dependent
+        # No domain definition needed
+
+        c_n = pybamm.Variable("Negative particle SOC")
+        c_p = pybamm.Variable("Positive particle SOC")
```
valentinsulzer, line 29: The standard we use is `theta` for the dimensionless stoichiometry (that goes from 0 to 1) as opposed to `c` for the dimensional concentration (that goes from 0 to c_max), and at the electrode level we call it "stoichiometry" instead of "SOC" to avoid confusion with full cell SOC

valentinsulzer (changes requested):

valentinsulzer:
Looks good, just a few comments to make it more in line with the naming standards in the rest of the codebase

tests/integration/test_models/test_full_battery_models/test_lithium_ion/test_ecm_split_OCV.py
parkec3: Done

docs/source/api/models/lithium_ion/ecm_split_ocv.rst
parkec3: Updated.

src/pybamm/models/full_battery_models/lithium_ion/basic_ecm_split_OCV.py
parkec3: Got it. I updated this within the code.

valentinsulzer (dismissed):

valentinsulzer (approved):

PR #4341, opened 2024-08-14
Testing the built wheels with `cibuildwheel`

prady0t:
# Description

WIP


prady0t:
local run : https://github.com/prady0t/PyBaMM/actions/runs/10443133199

arjxn-py (approved):
Looks pretty good. 
The idea of using marker is much better than that discussed in the meeting. Thanks :)

agriyakhetarpal (approved):
Looks quite great! Could you add an additional test where the IDAKLU solver creates a Simulation object from a DFN model and solves it for an hour? We might have such a test in `test_simulation.py` already, though – I haven't checked.

prady0t:
> test_simulation.py

Do we not have those simulations already inside `TestIDAKLUSolver` class (the one on which we have put `cibw` marker)?

agriyakhetarpal:
> Do we not have those simulations already inside `TestIDAKLUSolver` class (the one on which we have put `cibw` marker)?

Oh, I see that it runs on the entire class of tests, not just one test. Apologies for the oversight!

agriyakhetarpal:
I'll merge this for now since we will have another release cycle coming up soon for 24.9, and it would be great to have the wheels getting tested for it. Thanks, @prady0t!

prady0t:
Ok lets merge this. I'll open another pr to add changes to the docs

agriyakhetarpal:
Yes, docs are much appreciated

PR #4351, opened 2024-08-16
IDA adaptive time stepping

### LINKED ISSUE #4312: Time stepping in IDA - minimise solver restarts
proposal by @MarcBerliner (https://github.com/pybamm-team/PyBaMM/issues/3910#issuecomment-2260736100)

### Time stepping
In PyBaMM, the solver currently stops at all values in the `t_eval` vector. It seems that we use `t_eval` and set some `dt` for a few distinct purposes:
1. To enforce time-dependent discontinuities within a single step, like with a drive cycle
2. To set the data collection frequency (as in the `period` experiment kwarg)
3. Setting a small `dt` to force better solver convergence (take smaller time steps)

These three reasons for setting a `dt` are all valid, but stopping the simulation at all time steps can have a drastic impact on the adaptive time stepping and performance. For example, consider a full C/10 discharge with
- a. `t_eval = [0, 36000]` (i.e., no stops)
- b. `t_eval = np.arange(0, 36060, 60)` (pybamm default)

If we compare the solver stats, 
- Number of steps: a. `165` vs b. `715`
- Number of residual calls: a. `288` vs b. `823`
- Number of linear solver setups: a. `28` vs b. `91`
- Number of nonlinear iterations: a. `286` vs b. `821`
- DAE integration time: a. `25.5 ms` vs b. `97.1 ms`

Even though we solve the same system, the dense `t_eval` b. is nearly 4x slower! To address these issues, I propose the following changes that align the time-stepping options with [Julia's differential equation solvers (see Output Control)](https://docs.sciml.ai/DiffEqDocs/stable/basics/common_solver_opts/#solver_options):
- [ ] **(Breaking)** By default, save every `t` and `y` determined by IDA's adaptive time stepping algorithm. This eliminates issues like the one above with the C/10 discharge. We can accurately post-interpolate the solution onto specific times with IDA's Hermite interpolator. This is a huge benefit for parameter estimation because we will always obtain the full solution accuracy regardless of the data collection frequency.
- [ ] **(Non-breaking)** Change the description of `t_eval` to match Julia's `tstops`: "Denotes _extra_ times that the time-stepping algorithm must step to. This should be used to help the solver deal with discontinuities and singularities, since stepping exactly at the time of the discontinuity will improve accuracy." With this option, drive cycles in 1. still work
- [ ] **(Non-breaking)** Add a solver option that matches Julia's `saveat`: "Denotes specific times to save the solution at, during the solving phase. The solver will save at each of the timepoints in this array in the most efficient manner available to the solver." This addresses the data collection frequency in 2. without negatively affecting performance since it interpolates the solution with `IDAGetDky()`.
- [ ] **(Non-breaking)** Discourage modifying `t_eval` for performance issues and encourage modifying appropriate solver options (`rtol`, `atol`, `dt_max`, etc.), which addresses 3.

MarcBerliner:
# Description

This PR makes several changes to adaptive time-stepping in IDA. Previously, the `t_eval` decided both the points at which we store the solution and the points to explicitly stop the IDA solver. Now, the time-stepping has been broken down into `t_eval` (which retains the same behavior) and `t_interp`, which uses IDA's internal Hermite interpolator to efficiently store the solution. Additionally, we default to saving the adaptive time steps, which gives us full-order solution accuracy without negatively impacting the solve speed with periods or dense `t_eval` vectors.

This PR cleans up the main `solve` loop of the IDA solver and clearly delineates saving the full solution `y` vs. specific output variables.

### Example: adaptive time outputs (new) vs. fixed outputs
Screenshot 2024-08-16 at 1 37 59 PM

The above picture shows the difference in resolution between the new adaptive time-stepping scheme and the previous approach without affecting the performance. The markers show every data point available in `sol.t`

### Period/drive cycles improvements

```python
parameter_values.set_initial_stoichiometries(1)

experiment = pybamm.step.CRate(0.1, period = period, duration = 36000)
sim = pybamm.Simulation(
    model,
    solver=solver,
    experiment=experiment,
)

sim.solve()
```

image

The above figure shows the `solve` time for a full C/10 discharge. The new IDA solver is strictly faster than the old IDA and the Casadi solver. For the SPM, the IDA vs. Casadi timings are closer because IDA is a DAE solver while Casadi is using an ODE solver. Adding CVODE to pybamm could be future work to further improve our ODE speed.

For small periods, the new IDA solver is about an order of magnitude faster than the old IDA. For large periods, it's about half an order of magnitude faster than Casadi. 

In a parameter estimation context, you can also post-interpolate the adaptive time-stepping solution without specifying a `t_interp` or `period` for just the variable of interest (such as voltage) which will dramatically improve the performance.

### Adaptive time-stepping (no period)

```python
parameter_values.set_initial_stoichiometries(0)

experiment = pybamm.Experiment([f"Charge at {C_rate}C until 4.2 V"])
sim = pybamm.Simulation(
    model,
    solver=solver,
    experiment=experiment,
)

sim.solve()
```

Screenshot 2024-08-16 at 1 54 16 PM

When `t_eval` or a `period` is not set for an experiment, the new IDA is strictly faster than the previous IDA. Additionally, adaptive time-stepping has much better performance for slower simulations that previously used a dense `t_eval` vector by default.

Fixes #4312

## Type of change

# Key checklist:

## Further checks:


src/pybamm/experiment/step/base_step.py
```diff
@@ -266,6 +266,84 @@ def default_duration(self, value):
...
+        Parameters
+        ----------
+        solver: :class`pybamm.BaseSolver`
+            The solver
+        tf: float
+            The final time
+        t_interp: np.array | None
+            The time points at which to interpolate the solution
+        """
+        if solver._supports_interp:
```
valentinsulzer, line 295: We don't do this well in general, but private methods should only be accessed from within the same file/class - in this case each `Solver` class should implement a property `supports_interp` in order to make it truly read-only.

src/pybamm/experiment/step/base_step.py
```diff
@@ -266,6 +266,84 @@ def default_duration(self, value):
...
+            return self._setup_timestepping(solver, tf, t_interp)
+        else:
+            return self._setup_timestepping_dense_t_eval(solver, tf, t_interp)
+
+    def _setup_timestepping(self, solver, tf, t_interp):
+        """
+        Setup timestepping for the model. This returns a t_eval vector that stops
+        only at the first and last time points. If t_interp and the period are
+        unspecified, then the solver will use adaptive time-stepping. For a given
+        period, t_interp willbe set to return the solution at the end of each period
```
valentinsulzer, line 305:
```suggestion
        period, t_interp will be set to return the solution at the end of each period
```

src/pybamm/solvers/base_solver.py
```diff
@@ -65,6 +65,7 @@ def __init__(
         self.name = "Base solver"
         self.ode_solver = False
         self.algebraic_solver = False
+        self._supports_interp = False
```
valentinsulzer, line 68: make this a property instead (and we should have made `ode_solver` and `algebraic_solver` properties too so might as well do that while you're there)

valentinsulzer (commented):
Looks good to me other than some very minor comments. I have a feeling there are several bits of code we'll be able to refactor once this is in as the default everywhere but that can be a job for later. As usual will leave the C++ code to others to review

src/pybamm/simulation.py
```diff
@@ -400,6 +401,9 @@ def solve(
             Whether to show a progress bar for cycling. If true, shows a progress bar
             for cycles. Has no effect when not used with an experiment.
             Default is False.
+        t_interp : None, list or ndarray, optional
```
martinjrobins, line 404: `t_interp` is not passed to  `solver.solve` for the case `self.operating_mode in ["without experiment", "drive cycle"]`

martinjrobins: also, does the docstring for t_eval need to be updated? It currently says "Can be provided as an array of times at which to return the solution", but this is the role of `t_interp` now.

martinjrobins (changes requested):
Really great work @MarcBerliner, this looks excellent and some really great speed-ups! :) I've suggested a few changes below that need to be fixed but mostly minor. One general question I had: many of the idaklu tests use a dense t_eval and an identical t_interp, which is now the "slow" way of running the solver. When not specificly testing the interpolation code, I would use the most efficient / standard way of running `solver.solve`, which I believe would be `t_eval = [start, stop]` and `t_interp = None` or `t_interp = np.linspace(start, stop, N)`?

src/pybamm/solvers/c_solvers/idaklu/IDAKLUSolverOpenMP.inl
```diff
@@ -384,54 +432,127 @@ Solution IDAKLUSolverOpenMP<ExprSet>::solve(
...
+  DEBUG("IDASolve");
+  while (true) {
+    // Progress one step
+    retval = IDASolve(ida_mem, tf, &t_val, yy, yp, IDA_ONE_STEP);
+
+    if (retval < 0) {
+      // failed
+      break;
+    } else if (t_prev == t_val) {
+      // IDA sometimes returns an identical time point twice
```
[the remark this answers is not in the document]
MarcBerliner: I'm not sure if this is an IDA bug, but this issue is documented in [Sundials.jl](https://github.com/SciML/Sundials.jl/blob/d7e83c086d40e1b5b7ce470cccfea5dfd5586ec3/src/common_interface/solve.jl#L1418):
```
# NB: CVode, ARKode may warn and then recover if integrator.t == integrator.tprev so don't flag this as an error 
```
Whenever I've encountered this issue it's due to an ill-posed model, which can happen a lot in parameter estimation. It does continue the solve, but in my experience it uses miniscule time steps that cause floating point arithmetic issues and continues to use these tiny steps until it exhausts the maxiters.

MarcBerliner:
> One general question I had: many of the idaklu tests use a dense t_eval and an identical t_interp, which is now the "slow" way of running the solver. When not specificly testing the interpolation code, I would use the most efficient / standard way of running `solver.solve`, which I believe would be `t_eval = [start, stop]` and `t_interp = None` or `t_interp = np.linspace(start, stop, N)`?

@martinjrobins good point. The most efficient way to run it is using the `t_eval = [start, stop]` and `t_interp = None`. A current limitation with this PR is that we do not have the Hermite interpolator available in python (adding it is difficult), so all the post-interpolation tests like `sol[v](t)` use linear interpolation, which is less accurate. I changed most tests to use `t_interp = np.linspace(start, stop, N)`, which uses the Hermite interpolator.

Once [this PR merges](https://github.com/pybamm-team/PyBaMM/pull/4326), we can fully migrate all the tests to using `t_eval = [start, stop]` and `t_interp = np.linspace(start, stop, N)`. When we have more accurate post-interpolation, we can then change it to `t_eval = [start, stop]` and `t_interp = None`

src/pybamm/solvers/c_solvers/idaklu/common.cpp
```diff
@@ -0,0 +1,59 @@
+#include "common.hpp"
+#include <iostream>
+
+#include <idas/idas.h>                 /* prototypes for IDAS fcts., consts.    */
```
martinjrobins, line 4: all these headers are already included in common.hpp and so are not needed here

src/pybamm/solvers/base_solver.py
```diff
@@ -65,10 +65,43 @@ def __init__(
...
         self._on_extrapolation = "warn"
         self.computed_var_fcns = {}
         self._mp_context = self.get_platform_context(platform.system())
 
+    @property
+    def ode_solver(self):
+        return self._ode_solver
+
+    @ode_solver.setter
+    def ode_solver(self, value):
```
martinjrobins, line 78: I don't think we need a setter here as ode_solver should be read-only for users

src/pybamm/solvers/base_solver.py
```diff
@@ -65,10 +65,43 @@ def __init__(
...
+        if not isinstance(value, bool):
+            raise ValueError("ode_solver must be a boolean value")
+        self._ode_solver = value
+
+    @property
+    def algebraic_solver(self):
+        return self._algebraic_solver
+
+    @algebraic_solver.setter
+    def algebraic_solver(self, value):
```
martinjrobins, line 88: same here

src/pybamm/solvers/base_solver.py
```diff
@@ -65,10 +65,43 @@ def __init__(
...
+        if not isinstance(value, bool):
+            raise ValueError("algebraic_solver must be a boolean value")
+        self._algebraic_solver = value
+
+    @property
+    def supports_interp(self):
+        return self._supports_interp
+
+    @supports_interp.setter
+    def supports_interp(self, value):
```
martinjrobins, line 98: and here

src/pybamm/solvers/base_solver.py
martinjrobins: internally we can still set this via `self._ode_solver = X`

martinjrobins (changes requested):
thanks @MarcBerliner, just a few more minor changes below

martinjrobins (approved):
Looks great, thanks @MarcBerliner

PR #4354, opened 2024-08-18
Migrating unittest to pytest 6

prady0t:
I've tried using fixture and parameterized in this PR. Do let me know places I've missed (inside this PR).

#4156


tests/unit/test_geometry/test_battery_geometry.py
```diff
@@ -1,28 +1,30 @@
...
             )
-            for domain_geoms in geometry.values():
-                all(
-                    self.assertIsInstance(spatial_var, str)
-                    for spatial_var in domain_geoms.keys()
-                )
         geometry.print_parameter_info()
 
-    def test_geometry(self):
+    def test_geometry(self, geometry):
```
agriyakhetarpal, line 27: We initialise the `geometry` variable below, do we need it in the arguments of the function?

agriyakhetarpal: Also, this test can be refactored since we are testing the same thing repeatedly. Please add this to the tracking issue.

tests/unit/test_meshes/test_scikit_fem_submesh.py
```diff
@@ -444,23 +425,19 @@ def test_mesh_creation(self):
...
             if domain == "current collector":
                 # NOTE: only for degree 1
                 npts = var_pts["y"] * var_pts["z"]
-                self.assertEqual(mesh[domain].npts, npts)
+                assert mesh[domain].npts == npts
             else:
-                self.assertEqual(len(mesh[domain].edges), len(mesh[domain].nodes) + 1)
+                assert len(mesh[domain].edges) == len(mesh[domain].nodes) + 1
 
     def test_exceptions(self):
```
agriyakhetarpal, line 442: This test can be refactored, too.

tests/unit/test_parameters/test_parameter_sets/test_OKane2022.py
```diff
@@ -1,55 +1,54 @@
...
-class TestOKane2022(unittest.TestCase):
-    def test_functions(self):
-        param = pybamm.ParameterValues("OKane2022")
-        sto = pybamm.Scalar(0.9)
-        T = pybamm.Scalar(298.15)
 
-        fun_test = {
+class TestOKane2022:
+    @pytest.mark.parametrize(
+        "name, inputs, expected_output",
```
agriyakhetarpal, line 7: Looks great, definitely add more like these! 👍

tests/unit/test_models/test_full_battery_models/test_lithium_ion/test_electrode_soh.py
```diff
@@ -2,11 +2,22 @@
...
+@pytest.fixture()
+def options():
+    options = {
+        "open-circuit potential": "MSMR",
+        "particle": "MSMR",
+        "number of MSMR reactions": ("6", "4"),
+        "intercalation kinetics": "MSMR",
+    }
+    return options
+
```
agriyakhetarpal, line 18: Could you move this fixture down to just above `TestElectrodeSOHMSMR`, so that we know it is relevant only for those tests?

agriyakhetarpal: Either that, or (maybe better) you can add in a comment here the names of the classes that this applies to.

tests/unit/test_models/test_base_model.py
```diff
@@ -187,15 +184,15 @@ def test_get_parameter_info(self):
...
-        self.assertIn("i", parameter_info)
+        assert parameter_info["a"][1] == "InputParameter"
+        assert parameter_info["b"][1] == "InputParameter in ['test']"
+        assert "c" in parameter_info
+        assert "d" in parameter_info
+        assert "e" in parameter_info
+        assert "f" in parameter_info
+        assert parameter_info["g"][1] == "Parameter"
+        assert "h" in parameter_info
+        assert "i" in parameter_info
```
agriyakhetarpal, line 195: Wherever you find repetition like this, it's probably better to parameterize the function and gain benefits, both in terms of readability and slightly faster execution by parallelization.

tests/unit/test_models/test_base_model.py
```diff
@@ -270,34 +267,25 @@ def set_initial_conditions(self, variables):
...
+        assert parameter_info["sub1"]["w"][1] == "InputParameter"
+        assert parameter_info["sub1"]["e"][1] == "InputParameter"
+        assert parameter_info["sub1"]["g"][1] == "Parameter"
+        assert parameter_info["sub1"]["x"][1] == "Parameter"
+        assert parameter_info["sub1"]["f"][1] == "InputParameter in ['test']"
+        assert parameter_info["sub2"]["b"][1] == "InputParameter in ['test']"
+        assert parameter_info["sub2"]["h"][1] == "Parameter"
+        assert parameter_info["sub1"]["c"][1] == "FunctionParameter with inputs(s) ''"
+        assert parameter_info["sub2"]["d"][1] == "FunctionParameter with inputs(s) ''"
+        assert parameter_info["sub2"]["i"][1] == "FunctionParameter with inputs(s) ''"
```
agriyakhetarpal, line 288: This looks like it can be parameterized as well.

tests/unit/test_models/test_base_model.py
```diff
@@ -385,18 +373,18 @@ def test_print_parameter_info_submodel(self):
...
+        assert "Parameter" in result
+        assert "InputParameter" in result
+        assert "FunctionParameter with inputs(s) ''" in result
+        assert "InputParameter in ['test']" in result
+        assert "g" in result
+        assert "a" in result
+        assert "c" in result
+        assert "h" in result
+        assert "b" in result
+        assert "d" in result
```
agriyakhetarpal, line 387: This can be parameterized as well. I would recommend opening a subsequent PR that implements nothing but parameterization (and similarly for fixtures, etc.) and changing this test there.

tests/unit/test_expression_tree/test_variable.py
```diff
@@ -2,105 +2,95 @@
...
-        self.assertEqual(a.domain, [])
+        assert a.name == "a"
+        assert a.domain == []
         a = pybamm.Variable("a", domain=["test"])
-        self.assertEqual(a.domain[0], "test")
-        self.assertRaises(TypeError, pybamm.Variable("a", domain="test"))
-        self.assertEqual(a.scale, 1)
-        self.assertEqual(a.reference, 0)
+        assert a.domain[0] == "test"
+        pytest.raises(TypeError, match=pybamm.Variable("a", domain="test"))
```
agriyakhetarpal, line 20: Could you take a look at this failure? – I noticed from the annotation, which was quite helpful. It looks like a `pytestify` bug, where, instead of converting `assertRaises` to a

```python
with pytest.raises(TypeError):
    pybamm.Variable("a", domain="test")
```

it was converted to a regular test where `match` expects a string, but it received a `pybamm.Variable` instance, which won't work.

tests/unit/test_expression_tree/test_variable.py
```diff
@@ -2,105 +2,95 @@
...
         a = pybamm.VariableDot("a'")
-        self.assertEqual(a.name, "a'")
-        self.assertEqual(a.domain, [])
+        assert a.name == "a'"
+        assert a.domain == []
         a = pybamm.VariableDot("a", domain=["test"])
-        self.assertEqual(a.domain[0], "test")
-        self.assertRaises(TypeError, pybamm.Variable("a", domain="test"))
+        assert a.domain[0] == "test"
+        pytest.raises(TypeError, match=pybamm.Variable("a", domain="test"))
```
agriyakhetarpal, line 79: Same error as above here

tests/unit/test_expression_tree/test_variable.py
prady0t: Also not sure if `TypeError` should be raised here anyways. See : https://github.com/pybamm-team/PyBaMM/blob/396741eec6076a83c3104aa9b73d2f7957201ab7/src/pybamm/type_definitions.py#L17
i.e. domain can be of type `string` or a list of it as well.

agriyakhetarpal:
```python
self.assertRaises(TypeError, pybamm.Variable("a", domain="test"))
```

works fine, as you mentioned, but

even the `unittest` equivalent of the `pytest.raises` context manager, i.e.,

```python
with self.assertRaises(TypeError):
    pybamm.Variable("a", domain="test")
```

fails with the same error not being raised.

I don't think this test was written correctly, since we should always use a context manager to catch errors. Either that, or we never enforced the type in the first place. For example, here is a short snippet of what happens if we enforce the type and raise an error:

```python
from typing import TypeAlias, Union
import unittest
import pytest


DomainType: TypeAlias = Union[list[str], str, None]


def foo(name: str, domain: DomainType = None) -> tuple[str, DomainType]:
    if isinstance(domain, int):
        raise TypeError
    return (name, domain)

# both should work
foo(name="a", domain="test")
foo(name="a", domain=["test"])


class TestFoo(unittest.TestCase):
    def test_foo(self): # raises an error
        self.assertRaises(TypeError, foo(name="a", domain=1))
    def test_foo2(self): # passes
        with self.assertRaises(TypeError):
            foo(name="a", domain=1)
    def test_foo3(self): # passes
        with pytest.raises(TypeError):
            foo(name="a", domain=1)

```

Here, `self.assertRaises(TypeError, foo(name="a", domain=1))` raises an error instead of failing because `foo` already raises a `TypeError` at runtime, which can't be caught out of context. We could enforce this, but it would contradict the docstring of `pybamm.Variable()`, which says that the argument is "domain : iterable of str", and changing the behaviour of the function means this is a breaking change – not recommended and outside the scope of migrating our testing infrastructure. I think we should be okay with removing this test case, since the behaviour is clearly correct – it shouldn't raise an error since we don't validate the type in the code. Do you have any thoughts, @Saransh-cpp?

prady0t: Totally agreed. We should remove this and the following test. Waiting on Saransh reply.

Saransh-cpp: Removing the test case sounds good to me too

tests/unit/test_geometry/test_battery_geometry.py
prady0t: Now sure how to refactor. Could you help me more?

agriyakhetarpal: Usually, in such situations, it's better to make the test shorter and distribute it into multiple tests. For example, here, you could rewrite `test_geometry` into an array of smaller tests (haven't added the `self` keyword here for brevity):

```python
def test_battery_geometry_current_collector_tabs():
    geometry = pybamm.battery_geometry(
        options={
            "particle size": "distribution",
            "dimensionality": 1,
        },
    )
    assert "tabs" in geometry["current collector"]
```

and

```python
def test_battery_geometry_cylindrical_dimensionality():
    geo = pybamm.geometric_parameters
    geometry = pybamm.battery_geometry(
        form_factor="cylindrical", options={"dimensionality": 1}
    )
    assert geometry["current collector"]["r_macro"]["min"] == geo.r_inner
    assert geometry["current collector"]["r_macro"]["max"] == 1
```

However, readability is important. So, if refactoring introduces too many conditions or makes the code hard to read, it is best to not bother and keep the test as is.

Another aspect is to remove the loops by parametrizing. For example, the loop `for cc_dimension in [0, 1, 2]:` can be refactored into

```python
@pytest.mark.parametrize("cc_dimension", [0, 1, 2])
def test_battery_geometry_basics(cc_dimension):
```

and similarly

```python
@pytest.mark.parametrize("electrode, particle_type", [
    ("negative", "primary"),
    ("negative", "secondary"),
    ("positive", "primary"),
    ("positive", "secondary"),
])
def test_battery_geometry_particle_phases(electrode, particle_type):
    options = {"particle phases": "2"}
    geometry = pybamm.battery_geometry(options=options)
    geo = pybamm.GeometricParameters(options=options)
# followed by the rest of the test
```

and so on. I hope this can get you going!

agriyakhetarpal: I'm happy to write more examples if needed, it's easier than you would think :)

arjxn-py: These examples looks enough to me. Thanks :)
I guess we don't need case or if/else then as we using `parametrize` marker?

agriyakhetarpal: In some cases, we might still need them.

```python

@pytest.mark.parametrize("foo, bar", [0, 1, 2, 3, 4])
def test_something(foo, bar):
    # Some logic for initialising "foo" and "bar" here
    # and setting attributes, etc.
    # [...]

    # Conditional logic
    match foo.value:
        case 1:
            # do something
            assert foo in bar
        case 3.value:
            # do something else
            assert bar in foo

    # not (1) and not (3) – i.e., for (0) or (2) or (4)
    # Some other logic that applies to the rest of the cases
    assert foo.value + bar.value > 30
```

but it depends on the test. It _might_ be better to handle such cases without parameterizing them

prady0t:
Not sure why the error is. I was getting another regex error locally can you run CI jobs again?

tests/unit/test_parameters/test_parameter_sets/test_OKane2022.py
```diff
@@ -1,37 +1,45 @@
...
+        [
```
arjxn-py, line 8:
```suggestion
        {
```

agriyakhetarpal:
Since this PR seems to have taken a bit of time (for all the right reasons), let's nudge this towards a merge asap and work on the parameterization in further, dedicated PRs in your coming weeks – once the `pytest` migration gets finished.

tests/unit/test_parameters/test_parameter_sets/test_OKane2022.py
```diff
@@ -1,37 +1,45 @@
...
 
-class TestOKane2022(unittest.TestCase):
-    def test_functions(self):
-        param = pybamm.ParameterValues("OKane2022")
-        sto = pybamm.Scalar(0.9)
-        T = pybamm.Scalar(298.15)
 
-        fun_test = {
+class TestOKane2022:
+    @pytest.mark.parametrize(
```
Saransh-cpp, line 6: The `(` is not close below

prady0t:
I reverted the changes in `test_OKane2022.py` I still need to figure out the regex error. In the meantime, I'm also working on the last migration. I'll try to open PR by today or tomorrow morning.

prady0t:
I've removed pytest conversion from `test_base_battery_model.py`. This should work

test_spm_2024_09_13-PM03_52.json
agriyakhetarpal, line 1: This generated test file has crept into the changes for some reason, it should be removed

agriyakhetarpal (changes requested):
Looks nice, thanks, @prady0t! Just one more change, and we can merge this if the tests pass.

test_spm_2024_09_13-PM03_52.json
prady0t: Ah almost forgot about this

agriyakhetarpal (approved):

PR #4362, opened 2024-08-19
Remove install_jax and rename have_jax

### LINKED ISSUE #4131: Remove `pybamm_install_jax` and references to it
Based on https://github.com/pybamm-team/PyBaMM/issues/4127#issuecomment-2145762187 and pointed out by @DavidMStraub, after v24.5 gets released on PyPI, the `pybamm_install_jax` entry point and its corresponding method `install_jax` in `pybamm/util.py` can be removed from the source code for v24.9, following the completion of its deprecation cycle since v23.9.

There might be a few references in the documentation about it as well, I have not checked those at this time.

kratman:
# Description

Removes the deprecated install_jax() method

Fixes #4131

# Key checklist:

## Further checks:


src/pybamm/util.py
```diff
@@ -267,7 +264,7 @@ def get_parameters_filepath(path):
         return os.path.join(pybamm.__path__[0], path)
 
 
-def have_jax():
```
Saransh-cpp, line 270: Given that this is a public API, it should be a deprecation before renaming - https://github.com/pybamm-team/PyBaMM?tab=readme-ov-file#versioning :(

Saransh-cpp (changes requested):
Thanks, @kratman!

src/pybamm/util.py
kratman: I did not bother deprecating it since it seems like more of an internally used function for testing installations than one that would be commonly used by our users

agriyakhetarpal: I agree that it could be handled without a deprecation warning, since `pybamm_install_jax` was removed more than a year ago and we haven't had any complaints from users about `pybamm[jax]`.

Saransh-cpp: We _explicitly_ list `util` as a public module -

https://github.com/pybamm-team/PyBaMM/blob/977dcf9cdfc2fc978da3b1b56b7423454a2a2929/src/pybamm/__init__.py#L225

making all the functions in the module (that does not start with `_`) public.

Users might not be using the function, but _assuming_ that it is not being used is really not a good practice. If we go ahead with the PR, it would be good to rename it to `_has_jax` to clearly indicate that the function is private.

I still incline towards deprecation, but won't block the PR!

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Features
 
+- Remove deprecated function `pybamm_install_jax` and renamed `have_jax` to
```
Saransh-cpp, line 5: Should be in "Breaking changes" or a "Misc" category? Definitely not in "Features"

kratman: Ok I will switch it to a different section

kratman: Done

src/pybamm/util.py
agriyakhetarpal: Oh, I see. I didn't remember that `have_jax` was part of `pybamm.util`. Yes, then either a deprecation or marking it as a private attribute would be fine. I would suggest the latter. I would have suggested going through the three-release deprecation timeline too, but since it isn't used on an everyday basis in people's scripts, it should be fine to mark private directly as long as we have a CHANGELOG entry. Moreover, if one imports `pybamm.JaxSolver()` nowadays with `jax` and `jaxlib` missing, they already receive a nice error message and links to the documentation with instructions on the optional dependencies.

The `pybamm.have_idaklu()` method, on the other hand, will remain quite useful (and it is currently used a lot, too) until we have a separate IDAKLU solver available via PyPI wheels.

kratman: We can't make it `_has_jax()`, that marks it as a private member. Since it is used outside of `util.py`, then it is not private.

I was mainly trying to change this to make it read a bit better: "if pybamm has jax" instead of "if pybamm have jax". I would generally expect these functions to be internal checks, which is why I figured changelog was sufficient for updating the name.

Saransh-cpp: Private functions can be used internally without any restrictions, so it should be okay to add that `_` ✅

kratman: @Saransh-cpp According to Ruff and PyCharm it is incorrect. I know we don't have these inspections on yet, but they are good practice.

PyCharm:
image

Ruff:
image

Saransh-cpp: I see, but I have never seen that rule being used anywhere. The original implementation looks pretty unmaintained at the moment too - https://github.com/Korijn/flake8-self.

Also, scientific python does not recommend that rule, so I think we can ignore that. Plus, it looks like it is meant for private class members, and I don't know why it is complaining on private functions. Using private functions for internal work in a library should be okay because that is what they are meant for.

image

kratman:
@Saransh-cpp, @agriyakhetarpal I pull out the has_jax() part for now

kratman:
Coverage failure is unrelated

rtimms (approved):

PR #4376, opened 2024-08-26
Streamline release process

kratman:
# Description

Streamlining the release process to make primary releases and bugfix releases instead of release candidates.

## Type of change

Process update

# Key checklist:

## Further checks:


.github/release_workflow.md
```diff
@@ -1,10 +1,13 @@
...
-This file contains the workflow required to make a `PyBaMM` release on GitHub, PyPI, and conda-forge by the maintainers.
+This file contains the workflow required to make a `PyBaMM` release on
+GitHub, PyPI, and conda-forge by the maintainers.
 
-## rc0 releases (automated)
+## Initial release (automated)
 
-1. The `update_version.yml` workflow will run on every 1st of January, May and September, updating incrementing the version to `vYY.MMrc0` by running `scripts/update_version.py` in the following files -
+1. The `update_version.yml` workflow will run on every 1st of January, May
+   and September, updating incrementing the version to `vYY.MM.0` by running
```
Saransh-cpp, line 9: Are we moving away from the concept of release candidates? The "rc" term is a PyPI and PEP thing. "rc" allows PyPI to identify if it is a final or a pre release.

Saransh-cpp: If we are, then we don't need this entire infrastructure. We can just do normal releases with bug fixes whenever we want to, just how it was before.

.github/workflows/update_version.yml
```diff
@@ -20,54 +20,36 @@ jobs:
...
         if: github.event_name == 'schedule'
         run: |
-          echo "VERSION=$(date +'v%y.%-m')rc0" >> $GITHUB_ENV
-          echo "NON_RC_VERSION=$(date +'v%y.%-m')" >> $GITHUB_ENV
+          echo "VERSION=$(date +'v%y.%-m' + ".0")" >> $GITHUB_ENV
 
       - name: Get current date for a manual release
         if: github.event_name == 'workflow_dispatch'
         run: |
           echo "VERSION=$(date +'v%y.%-m')${{ github.event.inputs.append_to_tag }}" >> $GITHUB_ENV
```
Saransh-cpp, line 28: This should be be updated to account for the new ".X" format in version.

.github/workflows/update_version.yml
```diff
@@ -20,54 +20,36 @@ jobs:
...
       - uses: actions/checkout@v4
-        if: github.event_name == 'schedule'
         with:
           ref: 'develop'
 
-      # the dispatch workflow is for rcX and final releases
-      - uses: actions/checkout@v4
-        if: github.event_name == 'workflow_dispatch'
-        with:
-          ref: '${{ env.NON_RC_VERSION }}'
```
Saransh-cpp, line 42: Why are we not checking out to the version branch for bug fixes? Checking out to develop will bring on all the changes, including any new features that went in.

Saransh-cpp (changes requested):
Thanks for taking this up, @kratman! Could you please try this on your fork and post the results here?

.github/workflows/update_version.yml
```diff
@@ -76,37 +58,16 @@ jobs:
...
-          title: Bump to ${{ env.VERSION }}
-          body: |
-            - [x] Update to ${{ env.VERSION }}
-            - [ ] Check the [release workflow](https://github.com/pybamm-team/PyBaMM/blob/develop/.github/release_workflow.md)
-
-      # checkout to the version branch for the final release
-      - uses: actions/checkout@v4
-        if: github.event_name == 'workflow_dispatch' && !startsWith(github.event.inputs.append_to_tag, 'rc')
-        with:
-          ref: '${{ env.NON_RC_VERSION }}'
```
Saransh-cpp, line 100: This would still be required to update the version in develop, right?

.github/workflows/update_version.yml
```diff
@@ -20,54 +20,36 @@ jobs:
...
-          message: 'Bump to ${{ env.VERSION }}'
-          new_branch: '${{ env.NON_RC_VERSION }}'
-
-      # use the already created release branch for rcX + final releases
-      # and commit
-      - uses: EndBug/add-and-commit@v9
-        if: github.event_name == 'workflow_dispatch'
         with:
           message: 'Bump to ${{ env.VERSION }}'
+          new_branch: '${{ env.VERSION }}'
```
Saransh-cpp, line 52: Creating a new branch for each release will bring in the non-bug-fix commits from develop too. Do we want bug fix releases to have features in them too? The version scheme gets very hazy here.

Saransh-cpp (changes requested):
I am a bit confused with the workflow. Maybe it will become more clear once you try it out on your fork.

.github/release_workflow.md
kratman: I like the idea of the initial release branch being created automatically. I agree that manually triggering the workflow is a bit useless. I can remove that.

Yeah in general I thought it would be good to move away from the release candidates. It seems to make our process take a month for each release. Late releases are not the end of the world when there is a lot of time between them, but it would be nicer to get some of our changes out faster.

.github/workflows/update_version.yml
kratman: This was because I was assuming that `append_to_tag` was `.1`, `.2`, etc.

As discussed above, I will remove the "workflow_dispatch" portion and simplify this process further

.github/workflows/update_version.yml
kratman: I was trying to go for something like this: 
image

Where if the releases were fast, then the release branch goes right to main, then main updates develop for the changelog/version.

Then if we do a bugfix release, it is manual anyway

kratman:
@Saransh-cpp I will run the workflow to test it and update the github workflow so that gets simpler for bugfix releases

kratman:
@Saransh-cpp, @agriyakhetarpal The release workflow is working on my PR (with the target branch changed to develop instead of main):
image

If you want to look at the PR, it is here: https://github.com/kratman/PyBaMM/pull/2

.github/workflows/update_version.yml
kratman: Replaced

kratman:
@Saransh-cpp Can you take another look? I would like to get the release process updated before the end of the mont

Saransh-cpp (changes requested):
Hi, @kratman, thanks for the explanation! I have some concerns with this PR -

1. The first release (a user-facing one) is automated. This means that it will be prepared even if developers know that there exists a critical bug in PyBaMM. The release PR can be stopped from being merged, but this will require manual work to ensure that the release does not contain any critical bugs. It was okay to have the `rc0` release automated as having bugs in `rc` releases is common and acceptable.
2. The bug fix releases are completely manual now with no GH Action automations. We don't usually create PRs manually to main; the PRs going into main are always (?) automated. Do we want to change that?
3. The linked example shows the version branch going into develop. Shouldn't it go into main for the release (according to the release instructions)? I will still prefer it to go into develop and then develop to go into main. We don't want main to go into develop to update versions.

Other than these points, it looks good!

kratman:
> 3. The linked example shows the version branch going into develop. Shouldn't it go into main for the release (according to the release instructions)? I will still prefer it to go into develop and then develop to go into main. We don't want main to go into develop to update versions.

Sorry I thought my comment was clear, but forks of PyBaMM only copy the develop branch by default. Since my fork does not have a main branch, I temporarily changed the target branch to develop. The workflow still goes to main:
```yaml
      # Create a PR from version branch to main
      - name: Make a PR from ${{ env.VERSION }} to main
        id: release_pr
        uses: repo-sync/pull-request@v2
        with:
          source_branch: '${{ env.VERSION }}'
          destination_branch: "main"
```

kratman:
> 1. The first release (a user-facing one) is automated. This means that it will be prepared even if developers know that there exists a critical bug in PyBaMM. The release PR can be stopped from being merged, but this will require manual work to ensure that the release does not contain any critical bugs. It was okay to have the `rc0` release automated as having bugs in `rc` releases is common and acceptable.

This automatically creates the release PR, but it needs to be checked before merging like all the other releases. It does not automatically merge the PR so we can always wait for it to be tested. This is also part of going to a typical bug fix release process where we could release a bugfix PR if there is a problem with the first one.

The goal is faster releases for bug fixes, while waiting for major releases when there are breaking changes.

Same as before:
- Whatever is merged into develop by the first day of the release month ends up in the release.
- Release PRs wait for devs to review/test them.
- Release PRs can be updated before merging.
- The process of tagging and triggering the formal release is manual.

Changes:
- The RC process is replaced by bugfix updates.
- The process is less time intensive, so we can easily make more releases. This gets changes out faster and lets us fix known issues without having to wait 1/3 of a year.
- Release workflows will not break during late releases.

Overall I think the RC version and bugfix release version are just two different paradigms which is why it is difficult to reconcile some of the differences.
1. The RC version was very careful about what was in the release, but the end result is releases take a full month of cautious work.
2. The bugfix release version accepts some risk of bugs being in the first release, but allows it to be fixed immediately.

I tend to be in favor of the bug fix release process because it is less work and PyBaMM is not critical to infrastructure. Our users get faster updates, and we get less work. While the RC process is more careful about looking for bugs, it does not create a 100% bug free release and any of the bugs that make it through the RC process will remain for 4 months.

kratman:
> 2. The bug fix releases are completely manual now with no GH Action automations. We don't usually create PRs manually to main; the PRs going into main are always (?) automated. Do we want to change that?

I removed the automations for bug fix releases based on this comment you made in the previous review:
image

Even before we had to do those cherry picks manually. Hopefully bug fixes are rare enough that we don't have to do this often. I can put something back for the `append_to_tag` part so that it uses the `update_version.yml` workflow for the bug fix releases as well, but I will also have to fix the date issue that was broken in the last release. Whenever we have a late release (twice in the last year), then the PRs to main would have to be manual anyway since the workflow gets `vYY.MM` from the current month.

Saransh-cpp (approved):
Sorry for sitting on this too long and thank you for the details! Looks good to me!

kratman:
Thanks @Saransh-cpp

Issue #3909, opened 2024-03-20
[Roadmap area] Parameter refactor


brosaplanella:
Refactor parameter values to use the BPX format by default.
Using nested dictionaries allows cleaner definition of negative / positive and also different components of the battery.
This is a big breaking change so we should add a conversion class from legacy parameter sets to new ones, and legacy parameter sets should still work for a while.

_Originally posted by @tinosulzer in https://github.com/pybamm-team/PyBaMM/issues/3839#issuecomment-1966635559_

rtimms:
FYI we (Ionworks) have a FUSE internship open for applications that is related to this, so we can support this over the summer

tinosulzer:
Also, when we do this, it would be great to do a deep review of existing parameter sets, including removing parameters that are not provided in the original reference (e.g SEI parameters for a performance-only model). See https://github.com/pybamm-team/PyBaMM/pull/3919#pullrequestreview-1954878606

brosaplanella:
Following https://github.com/pybamm-team/PyBaMM/issues/5064 this area is now paused until further notice, as it is no longer a priority.

PR #4378, opened 2024-08-26
Add base parameter class

kratman:
# Description

Adds a base class for parameters and groups them into other groups aligned with BPX

Related: #3909

## Type of change

# Key checklist:

## Further checks:


kratman:
@BradyPlanden Can you take a look at this?

src/pybamm/input/parameters/base_parameter_set.py
```diff
@@ -0,0 +1,39 @@
+class AbstractBaseParameters:
```
brosaplanella, line 1: Why is it AbstractBaseParameters and not just BaseParameters? If anything, these would be battery parameters (rather than abstract parameters). Also, might be worth to use BaseParameterSet or BaseParameterValues instead. I think we are not 100% consistent with this, but in theory parameters are the concept (e.g. diffusion coefficient) while parameter values/sets are the value.

brosaplanella (approved):
Just a picky notation comment, but looks good, thanks! Make sure to include a line in the CHANGELOG before merging.

src/pybamm/input/parameters/base_parameter_set.py
kratman: Abstract as in an abstract class where implementation details are not fully defined. The idea was that this provides an interface that we can extend for the various parameter sets.

kratman:
@brosaplanella What did you think of my groupings of parameters? Are the close enough or do they still need work?

src/pybamm/input/parameters/base_parameter_set.py
valentinsulzer: +1 for changing to `AbstractBaseParameterSets`

valentinsulzer (changes requested):
To be totally honest I don't like this change. I like the split to different components, but by making it a class it feels like it moves us further away from serialization, with everything having to be defined in a class, rather than closer to it. Ideally an entire parameter set should be able to be defined in a single json file.

kratman:
> To be totally honest I don't like this change. I like the split to different components, but by making it a class it feels like it moves us further away from serialization, with everything having to be defined in a class, rather than closer to it. Ideally an entire parameter set should be able to be defined in a single json file.

The way I was looking at it was that this class could be used as the interface between BPX files and the pybamm functionality. It lets us group and test if features are available. So we could easily add BPX reading/writing to the class plus sanity checking

brosaplanella:
> @brosaplanella What did you think of my groupings of parameters? Are the close enough or do they still need work?

I think they are in the right direction but work is needed. However, I believe this goes beyond parameters themselves and also covers models. For example, we have the mechanical properties, which currently are under the electrodes, but they could also fall under a new mechanics category. LAM is also another one that it is not clear. However, unless we need to set them in stone now, I am happy going ahead as things are and improving in the future.

I also agree with Valentin that serialisation and being able to define parameters as a single script should be the priority. I don't have a strong opinion regarding classes, but I think we should try to do whatever is easier for users (e.g. can we achieve similar testing results by just having some functions that read the .py or .json file and process it?).

valentinsulzer:
I think the class is good, but parameter sets should be stored in as-close-to-json-as-possible format, not embedded into child classes and split across different methods. Right now as-close-to-json-as-possible format is python functions + dicts like we have now, since there is no good way to include the python functions in a json file.

kratman:
This is badly outdated, so I am going to close it

PR #4383, opened 2024-08-27
Remove pybtex warning when importing PyBaMM

### LINKED ISSUE #4304: Pybtex warnings on downstream projects
When importing projects which use PyBaMM (e.g. pybamm-eis) users can get the following warning:
```
>>> import pybammeis
2024-07-29 18:14:25.383 - [WARNING] citations.read_citations(76): Citations could not be read
because the 'pybtex' library is not installed. Install 'pybamm[cite]' to enable citation reading.
```
This warning should probably be removed since it is popping up in other projects

kratman:
# Description

Updates to the bibtex parsing to avoid seeing warnings when importing PyBaMM.

Fixes #4304

## Type of change

# Key checklist:

## Further checks:


src/pybamm/citations.py
```diff
@@ -165,21 +154,13 @@ def _parse_citation(self, key):
 
                 # Add and register all citations
                 for key, entry in bib_data.entries.items():
-                    # Add to _all_citations dictionary
                     self._add_citation(key, entry)
-                    # Add to _papers_to_cite set
                     self._papers_to_cite.add(key)
-                    return
+                return
```
kratman, line 159: This would skip the loop after the first item. That seemed like a bug, so I fixed it while doing the rest of the changes. If anyone thinks this was intended behavior, then I can switch it back or rewrite the block of code to be less confusing.

src/pybamm/citations.py
```diff
@@ -138,7 +127,7 @@ def register(self, key):
                     self._add_citation_tag(key, entry=caller)
                     # Don't add citation tags if the citation is registered manually
                 except KeyError:  # pragma: no cover
-                    pass
+                    self._module_import_error = True
```
kratman, line 130: This was a mistake, fixing it
```suggestion
                    pass
```

rtimms (approved):

PR #4352, opened 2024-08-16
Migrate to `scikit-build-core`

cringeyburger:
# Description

This PR is about migrating to a new build system, `scikit-build-core` from `setuptools` and `wheel`, while incorporating necessary changes in GitHub Workflows, Benchmarks and Documentation.

CHANGES IN DISCUSSION:
- [x] Task 27: Use SUNDIALS installation script to set the environmental variables for windows --- We currently don't support Windows users to run `nox -s pybamm-requires` or `install_KLU_Sundials.py` script. Adding this feature would change a lot of things

CHANGES REQUESTED:
- [x] Task 22: Remove images from `.gitignore`
- [x] Task 23: Include the whole `sundials_KLU_libs` in `.gitignore`
- [x] Task 24: Remove import sys from bottom call in test_idaklu_solver
- [x] Task 25: Skip the parameter_sets test
- [x] Task 26: Change the version requirements given the scikit-build-core docs
- [x] Task 28: Update docs for DAE solver (remove extra text from the tabs)

CHANGES IN PROGRESS:

COMPLETED, need reviews:
- [x] Task 4: Remove pins on other build-time dependencies like CasADi --- unpinned all instances of build-time dependencies, including `scikit-build-core` in the direction to remove editable-rebuilds, or at least give a warning to the users before using it.
- [x] Task 5: Remove `pip update` commands from `nox.`
- [x] Task 13: Remove `set CMAKE_GENERATOR="Visual Studio 17 2022"` and `set CMAKE_GENERATOR_PLATFORM=x64` and let CMake set it by itself.
- [x] Task 14: Set defaults for Windows for the remaining 4 environment variables in `CMakeLists.txt`. Use "if set, use that otherwise use default" logic for `VCPKG_ROOT` (rename `VCPKG_ROOT_DIR` to `VCPKG_ROOT`) --- Can not fix using CMake variables, have to set environment variables, hence wrote a batch script to set the env vars permanently for the Windows user. The CI still use pre-defined env vars because this avoids issues related to interactive scripts.
- [x] Task 19: Clean the `find BLAS` script
- [x] Task 3: Run tests, including IDAKLU ones, on all platforms (except doctests). --- random windows IDAKLU tests fail because of workers crashes. IREE tests had to be skipped because of attribute failure. Rest linux and macos tests are working.
- [x] Task 1: Check if the error in Doctests `(pybamm.parameter_sets)` can be resolved using an example instead of explicitly setting all the names. --- need to set all names
[...]

Issue #3100, opened 2023-07-04
[Bug]: IDAKLU / casadi serialization error - Object written in version 2 but can only read in version 1


Mrzhang-hub:
### PyBaMM Version

23.4.1

### Python Version

3.9

### Describe the bug

RuntimeError:C:\..\casadi\core\serialization_stream.cpp:355:Assertion"load_version==v" failed:ProtoFunction failed. Object written in version 2 but can only read in version 1.

### Steps to Reproduce

```
import pybamm
import numpy as np
model = pybamm.lithium_ion.SPM()
geometry = model.default_geometry
param = model.default_parameter_values
param.process_model(model)
param.process_geometry(geometry)
mesh = pybamm.Mesh(geometry,model.default_submesh_types,model.default_var_pts)
disc = pybamm.Discretisation(mesh,model.default_spatial_methods)
disc.process_model(model)
t_eval = np.linspace(0,3600,100)
if pybamm.have_idaklu():
     klu_sol = pybamm.IDAKLUSolver(atol=1e-8,rtol=1e-8).solve(model,t_eval)
plot = pybamm.Quickplot(klu_sol)
plot.dynamic_plot()
```

brosaplanella:
Strange, it works at my end (even with version 23.4.1). Could be that when you upgraded to the newer PyBaMM version, casadi didn't. Can you try upgrading both PyBaMM and Casadi to their respective latest versions?

Mrzhang-hub:
> 

I have upgrade pybamm and casadi to their newest version.(pybamm=23.5 casadi=3.6.3) But the problem still exist. Is that has something to do with the version of c++?

Mrzhang-hub:
I downgraded  pybamm to version 22.10 and casadi to version 3.5.5, which solved the problem.

tinosulzer:
Reopening as suggested in https://github.com/pybamm-team/PyBaMM/discussions/3193#discussioncomment-6563103

We need to find out why the newer pybamm or casadi version breaks this

jsbrittain:
Reopening as the issue has reemerged, including in pip installs of pybamm. See #3193 for related discussion.

jsbrittain:
Some progress - Given the new information I have managed to recreate the serialisation error on my (working) machine by taking the following actions:
1. Removed/renamed `~/local` to remove any existing (local) `casadi` installs.
[...]

PR #4391, opened 2024-08-28
Fix CasADi libs workaround for macOS wheels, update to CasADi 3.6.6 for Windows wheels

agriyakhetarpal:
# Description

I was going to split these into two PRs, but I decided that it would be fine to do it in the same one:

- This removes a workaround previously used to repair the wheels for Intel macOS devices. We will now have better compatibility with macOS version 10.13 and later. The M-series wheels still have something broken, so I will need to reach out to CasADi upstream again.
- For Windows, I bumped CasADi to version 3.6.6 in our registry at https://github.com/pybamm-team/casadi-vcpkg-registry and updated the baseline for our `vcpkg` configuration.

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/.github/workflows/publish_pypi.yml
+++ b/.github/workflows/publish_pypi.yml
@@ -149,9 +149,6 @@ jobs:
       - name: Clone pybind11 repo (no history)
         run: git clone --depth 1 --branch v2.12.0 https://github.com/pybind/pybind11.git -c advice.detachedHead=false
 
-      - name: Set macOS-specific environment variables
-        run: echo "MACOSX_DEPLOYMENT_TARGET=11.0" >> $GITHUB_ENV
-
       - name: Install cibuildwheel
         run: python -m pip install cibuildwheel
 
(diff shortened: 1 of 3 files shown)


agriyakhetarpal:
Wheel builds passing: https://github.com/agriyakhetarpal/PyBaMM/actions/runs/10601940680/job/29382886771.

Opened an issue about CasADi 3.6.6 in their feedstock: https://github.com/conda-forge/casadi-feedstock/issues/114. Either way, it's not needed too much since 3.6.X releases maintain minor version compatibility and don't break in the ways we have noticed in #3100. We should be fine without it even if it takes slightly more time (since we don't distribute the IDAKLU solver in our conda-forge recipe anyway).

kratman (approved):
Changes look fine, but I won't merge this yet since it sounds like you are still waiting on Anaconda and CasADi updates

agriyakhetarpal:
Yes, let's hope that the update gets merged soon. We should be safe in general, though.

kratman:
> Yes, let's hope that the update gets merged soon. We should be safe in general, though (PyBaMM on conda-forge will be released only after PyPI, so we always have time to update our recipe).

If the CasADi/Anaconda issues don't get resolved soon, then we have to revert the change in the vcpkg registry for casadi correct?

agriyakhetarpal:
No, we don't need to, because we can stall the PyBaMM conda-forge release during the delay. PyBaMM does not need the build-time (host) dependency on CasADi for its conda-forge recipe (because the IDAKLU solver is not compiled there yet), just the run-time dependency (which is fine with 3.6.X).

agriyakhetarpal:
Also, no users have reported the IDAKLU solver's absence in conda-forge to us earlier. I had earlier noted somewhere that most people seem to be installing via PyPI, and that seems to still be the case: https://anaconda.org/conda-forge/pybamm/files. Let's wait for a few days before we merge this, nevertheless.

kratman:
It is a bit off topic, but what needs to happen for IDAKLU to be included with conda-forge? If the switch to IDAKLU as the default solver ends up happening, then that is pretty critical

agriyakhetarpal:
Oh, that's not off-topic, I had that in mind recently – if the switch to the IDAKLU solver being the default ends up happening, then `pyidaklu` will become a required dependency on PyBaMM, and we will need to create a new recipe for it to conda-forge via a `pyidaklu-feedstock` repository. Assuming that we make it the default in PyBaMM version 25.1 while also trying to unvendor it at the same time, we have the following three months and the next release cycle to set up a recipe for https://github.com/pybamm-team/pyidaklu and publish it on both PyPI and conda-forge.

cc: @santacodes

agriyakhetarpal:
We can include the IDAKLU solver even right now in the PyBaMM recipe without having unvendored it – we just need to add the relevant build scripts and compile it the right way across platforms – I've thought about it before and it isn't difficult, but it's better to do so after #4352 gets merged and becomes a part of a release (which would bring us back to the 25.1 release).

agriyakhetarpal:
CasADi 3.6.6 is now up on conda-forge, too: https://anaconda.org/conda-forge/casadi/files. We can now merge this safely.

agriyakhetarpal:
I see that the required check for auto-merge is just the RTD one, we should add the unit tests to this, at least, so that we don't auto-merge something in the future that can potentially fail.

PR #4394, opened 2024-08-29
I3769 kolzenberg2020 (Another option for the SEI growth)

### LINKED ISSUE #3769: SEI Growth by Electron Diffusion & Migration
### Description

The feature contains SEI growth models based on the diffusion and migration of electrons through the SEI (L.v. Kolzenberg, A. Latz, B. Horstmann, ChemSusChem 2020, 13, 3901-3910). The model is dervied to describe SEI growth during storage and battery operation.

### Motivation

This feature supports the simulation of  SEI growth under different operating conditions. Thereby it extends the set of available SEI growth models on PyBaMM and enables the comparison of SEI growth for different operating protocols. 

### Possible Implementation

The simplest to implement is to add these lines into `sei_growth.py`

```
OCP_n = variables["Negative electrode open-circuit potential [V]"]
j0 = variables["Negative electrode exchange current density [A.m-2]"]

elif self.options["SEI"] == "electron-diffusion limited":
            # L. v. Kolzenberg, ChemSusChem 2020 (eq. 25)
            chem_pot_Li = 0.1803
            eta_int = 2/F_RT*pybamm.arcsinh(j/(2*j0))
            j_ED = -(phase_param.D_li * phase_param.c_li_0 * param.F / L_sei_outer) * pybamm.exp(-F_RT*(eta_int + OCP_n + chem_pot_Li))
            j_sei = j_ED

elif self.options["SEI"] == "electron-migration limited":
            # L. v. Kolzenberg, ChemSusChem 2020 (eq. 27a & 27b)
            chem_pot_Li = 0.1803
            eta_int = 2/F_RT*pybamm.arcsinh(j/(2*j0))
            j_migr = pybamm.minimum(param.F**2 * phase_param.D_li * phase_param.c_li_0 / (2 * param.R * T * phase_param.kappa_inner) *j*pybamm.exp(-F_RT*(eta_int + OCP_n + chem_pot_Li)),0)
            j_sei = j_migr
```

### Additional context

In the following all cited equations refer to the equations from this work (L.v. Kolzenberg, A. Latz, B. Horstmann, ChemSusChem 2020, 13, 3901-3910).

The equations described in "Possible Implementation" origin from the general equation for the SEI current density under different operating conditions (eq. 19). Looking at different limits this equation reduces to the reaction limited (eq. 23), electron diffusion limited (eq. 25) and migration limited cases (eq. 27 a+b). Note that this "electron-migration limited" case differs from the mechanism already implemented in PyBaMM under the same name. Even though the basic equation for electromigration (see eq. 5) can yield both "migration" mechanisms, the major difference lies in the interpretation of conductivity and the origin of the gradient in the electrical potential in the SEI. Here the intercalation current of Li-ions is assumed to cause the gradient in the electrical potential. This leads then to SEI growth proportional to the intercalation current density and it goes linear with time (eq. 28).

Both mechanisms (electron diffusion & migration) depend on the overpotential of the SEI (eta_SEI), which is directly linked to the intercalation overpotential (eta_int). The latter is defined through the difference in potentials (eq. 10) and linked to the resulting intercalation current (eq. 9). In the above suggested implementation of the models we assume that j_SEI << j_int and therefore j_int ~ j. This let's us use eq. 9 directly to compute eta_int and insert it in the needed SEI overpotential (see eq. 18). Depending on the application this assumption may not be justified. Some attempts were made to keep the intercalation current a variable, and let the solver get a correct solution for j_int and j_SEI. However, I was not able to get numerical stability. 

To fully capture the correct dependencies and smooth transitions from reaction to diffusion to migration limited SEI growth it would be necessary to implement eq. 19. In theory this captures the linear to square-root to linear growth regimes of the SEI, depending on transport parameters and current SEI thickness. Unfortunately  i was not able to achieve any solutions due to solver problems.

Looking forward to thoughts on this.

kawaMANMI:
# Description

Another option for the SEI gowth has been added based on teh model of [von Kolzenberg et al. ](https://chemistry-europe.onlinelibrary.wiley.com/doi/full/10.1002/cssc.202000867) 
 spefically equation (19) has been implemented. Tunilening distance and lithoum ion conductivity in teh SEI as parameters need to be defined  here is  example to run a case

```python3
import pybamm
import numpy as np


no_cycles = 1
exp = pybamm.Experiment(
    [("Charge at C/10 until 4.2 V",
      "Rest for 5 hours",
      "Discharge at C/10 until 2.5 V",
      "Rest for 5 hours",
      )]* no_cycles )

param=pybamm.ParameterValues("Chen2020")
param.update({
  'tunneling distance for electrons [m]': 4e-9,
  'SEI lithium ion conductivity [S.m-1]': 1.0e-3,}, check_already_exists=False) 


sei_option = "Kolzenberg2020"
model = pybamm.lithium_ion.SPM({"particle": "uniform profile",
                                  "SEI": sei_option}) 
sim = pybamm.Simulation(model, parameter_values=param,
                          experiment=exp, solver=solver) 
sol = sim.solve(initial_soc=0) 
plt.plot(sol["Time [h]"].entries, sol["X-averaged negative total SEI thickness [m]"].entries)
```

Fixes #3769 

 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/src/pybamm/expression_tree/unary_operators.py
+++ b/src/pybamm/expression_tree/unary_operators.py
@@ -611,10 +611,13 @@ class Integral(SpatialOperator):
     A node in the expression tree representing an integral operator.
 
     .. math::
-        I = \\int_{a}^{b}\\!f(u)\\,du,
+        I = \\int_{u_{min}}^{u_{max}}\\!f(u)\\,dq,
 
-    where :math:`a` and :math:`b` are the left-hand and right-hand boundaries of
-    the domain respectively, and :math:`u\\in\\text{domain}`.
+    where :math:`u\\in\\text{domain}` is a spatial variable, :math:`u_{min}` and :math:`u_{max}` are the values of
+    :math:`u` at the left-hand and right-hand boundaries of the domain respectively, and :math:`dq` is given by, \n
+    :math:`dq=du` for cartesian coordinates, \n
+    :math:`dq=2\\pi udu` for cylindrical coordinates, \n
+    :math:`dq=4\\pi u^2 du` for spherical coordinates.
 
     Parameters
     ----------


brosaplanella:
Great, thanks @kawaMANMI! Will take a look later :)

.gitignore
```diff
@@ -137,3 +137,4 @@ results/
 
 # tests
 test_callback.log
+.mypy_cache/
```
brosaplanella, line 140: Can you remove this line from the Github commit? You can still keep it locally if you want.

src/pybamm/CITATIONS.bib
```diff
@@ -814,3 +814,15 @@ @article{Fan2022
   year={2022},
   publisher={Elsevier}
 }
+
+
+@article{von2020solid,
```
brosaplanella, line 819:
```suggestion
@article{VonKolzenberg2020,
```

src/pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -302,6 +302,7 @@ def __init__(self, extra_options):
                 "interstitial-diffusion limited",
                 "ec reaction limited",
                 "ec reaction limited (asymmetric)",
+                "Kolzenberg2020",
```
brosaplanella, line 305:
```suggestion
                "VonKolzenberg2020",
```

I suggest using this naming, to match how we usually refer to parameters.

src/pybamm/models/submodels/interface/sei/sei_growth.py
```diff
@@ -135,6 +137,26 @@ def get_coupled_variables(self, variables):
...
+            # Solid electrolyte interphase during battery cycling:
+            # Theory of growth regimes. ChemSusChem. 2020 Aug 7;13(15):3901-10.
+            eta_bar = F_RT * (delta_phi)
+            L_diff = (
+                phase_param.D_li * phase_param.c_li_0 * param.F / phase_param.j0_sei
+            ) * pybamm.exp(-(1 - alpha_SEI) * eta_bar)
+            L_tun = phase_param.L_tunneling
+            L_app = (L_sei - L_tun) * ((L_sei - L_tun) > 0)
+            L_mig = (
+                2 / F_RT * phase_param.kappa_Li_ion / (pybamm.AbsoluteValue(j) + 1e-38)
```
brosaplanella, line 153: Maybe we could use `pybamm.max()` instead, like we do in other places. This way you can also set a larger tolerance if you want as errors won't accumulate.

src/pybamm/models/full_battery_models/base_battery_model.py
brosaplanella: Note that adding the options made some test fail (tests/unit/test_models/test_full_battery_models/test_base_battery_model.py). You just need to go into that test and add the new option in the long string that contains all model options.

brosaplanella (commented):
Hi Kawa! Nice work, getting there. I have made some suggestions, including one that will hopefully make tests pass. We can address coverage once tests pass.

.gitignore
kratman: Whats the problem with including this in the gitignore?

brosaplanella: Good point! I thought it was better to remove as I wasn't sure whether it might add clutter or mess things in the repo (not sure if we are actually using mypy or not).

brosaplanella:
@pipliggins the integration failing test (`test_DFN.py`, in particular the `test_sei_VonKolzenberg2020` is failing because the `Sign` symbol does not have a `_from_json` method. I am happy to skip the output tests for this one, but I wanted to check first if that is an easy fix.

brosaplanella:
Apologies, marked it as ready by accident. Need to wait for #4470 to be merged first.

pipliggins:
> @pipliggins the integration failing test (`test_DFN.py`, in particular the `test_sei_VonKolzenberg2020` is failing because the `Sign` symbol does not have a `_from_json` method. I am happy to skip the output tests for this one, but I wanted to check first if that is an easy fix.

@brosaplanella Sorry it took a while to get to this - `Sign` doesn't have a `_from_json` method because my understanding was that all `Sign` classes should be converted to scalar/array values after discretisation (and as only discretised models can be serialised, there was no need for the function). I'm taking a look to see why that doesn't seem to be happening here, but if that assumption was just incorrect it should be a relatively quick fix to implement the method. Alternatively can be left as-is if the logic operation introduced in [bfab8a0](https://github.com/pybamm-team/PyBaMM/pull/4394/commits/bfab8a0df34f1477c92dee12a9b2918b7651e77c) is sufficient.

brosaplanella:
This PR includes the changes in #4470 so that one needs merging before this one can be reviewed.

brosaplanella (commented):
Looks good to me, but would be good to have an extra pair of eyes on it. Not marking it as approved so it doesn't get accidentally merged.

DrSOKane (dismissed):
This looks excellent! I noticed you coded the equation for `L_mig` to avoid division by zero, while still ensuring the correct behaviour at zero current by making it a very large number. Well done!

.gitignore
```diff
@@ -137,3 +137,4 @@ results/
 
 # tests
 test_callback.log
+# .mypy_cache/
```
kratman, line 140:
```suggestion
```

=2.42.2
kratman, line 1: This file should be deleted

CHANGELOG.md
kratman, line 1: The change log is messed up. The new items should be at the top and the other changes should be reverted

tests/unit/test_citations.py
```diff
@@ -36,7 +38,8 @@ def test_citations(self):
 
         # Test key error
         with pytest.raises(KeyError):
-            citations._parse_citation("not a citation")  # this should raise key error
+            # this should raise key error
```
kratman, line 41:
```suggestion
```

kratman (approved):
From the infra/conflicts side this looks good to me

kratman:
I will squash this after tests pass.

@all-contributors please add @kawaMANMI for code

PR #4415, opened 2024-09-04
Support sensitivities for Experiments

### LINKED ISSUE #3834: generalise simulation, include sensititivies
### Description

at the moment simulation is a specific to the battery models and experiment class within pybamm. I'd like to propose a simpler, more general purpose mechanism so that a sequence of pybamm models can be solved (in sequence), with the next model initialised from the previous (including sensitivities). This doesn't have to be anything fancy, just a way to reinitialise a model from the solution of another model (although you could also have a base simulation class that works for a general pybamm model). I'd imagine that the simulation class could use this to simplify some of its logic as well, and it would mean that sensitivities could be used with experiments

### Motivation

This was motivated by https://github.com/pybamm-team/PyBaMM/discussions/3804
and https://github.com/pybamm-team/PyBaMM/issues/3176

### Possible Implementation

_No response_

### Additional context

_No response_

martinjrobins:
# Description

Support sensitivity calculation for `pybamm.Simulation`, including experiments

Fixes #3834 

Note that although the sensitivity calculation for `pybamm.ScipySolver` appears to work, it seems like there are bugs in this solver for experiments (see #4429) so this is currently not tested.

## Type of change

# Key checklist:

## Further checks:


brosaplanella (approved):
Looks good to me! Asking @MarcBerliner to review the PR too, as he knows a lot more about this than me.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - Added additional user-configurable options to the (`IDAKLUSolver`) and adjusted the default values to improve performance. ([#4282](https://github.com/pybamm-team/PyBaMM/pull/4282))
 - Added the diffusion element to be used in the Thevenin model. ([#4254](https://github.com/pybamm-team/PyBaMM/pull/4254))
 - Added lumped surface thermal model ([#4203](https://github.com/pybamm-team/PyBaMM/pull/4203))
+- Added sensitivity calculation support for `pybamm.Simulation` and `pybamm.Experiment` ([#4415](https://github.com/pybamm-team/PyBaMM/pull/4415))
```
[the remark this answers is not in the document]
martinjrobins: oops, I missed that header, thanks @kratman, I'll fix that

src/pybamm/simulation.py
```diff
@@ -182,6 +182,27 @@ def set_up_and_parameterise_experiment(self):
...
+            try:
+                param = parameter_values[key]
+                if isinstance(param, pybamm.InputParameter):
+                    raise pybamm.ModelError(
+                        f"Cannot use '{key}' as an input parameter in this experiment. "
+                        f"This experiment is controlled via the following parameters: {restrict_list}. "
+                        f"None of these parameters are able to be input parameters."
+                    )
+            except KeyError:
+                pass
```
MarcBerliner, line 204: Can we check with `has_key` instead of the try/except block?

tests/unit/test_solvers/test_casadi_solver.py
```diff
@@ -1022,7 +1022,8 @@ def test_solve_sensitivity_algebraic(self):
             model, t_eval, inputs={"p": 0.1}, calculate_sensitivities=True
         )
         np.testing.assert_array_equal(solution.t, t_eval)
-        np.testing.assert_allclose(solution.y[0], 0.1 * solution.t)
+        print(solution.y.shape)
```
MarcBerliner, line 1025: Debug line left in

MarcBerliner (changes requested):
Thanks Martin! This is great, just a few minor comments on the code.

There are some experimental setups where a linearized sensitivity analysis might produce inaccurate results. Sequential experiment steps with a fixed end time will always work correctly. However, sequential steps with event-based termination conditions (like max voltage) may give the correct answer since the end time of each step is indeterminate and may change wrt the parameters. The only situation where an experiment step with an event-based termination is valid is if it’s the final step in the experiment. It would be nice to add a check for this in the code.

martinjrobins:
> There are some experimental setups where a linearized sensitivity analysis might produce inaccurate results. Sequential experiment steps with a fixed end time will always work correctly. However, sequential steps with event-based termination conditions (like max voltage) may give the correct answer since the end time of each step is indeterminate and may change wrt the parameters. The only situation where an experiment step with an event-based termination is valid is if it’s the final step in the experiment. It would be nice to add a check for this in the code.

Yea, its a good point. For event-based termination conditions there would be an additional term in the sensitivity equations which we are assuming is close to zero. In our particular case where the model changes over the event I'm not sure if this term can even be defined, although it would be interesting to look into this further. I'd be reluctant to make this an error, and to allow users to still use this for event-based terminations. As long as the additional term is small the sensitivities can still be used within, for example, an optimisation algorithm that is robust to errors in the gradient (which many are). I'll put something in the docstrings to warn users about this, and I'll do a run-time check and emit a warning if the experiment contains an event-based termination, how does that sound?

MarcBerliner:
> I'll put something in the docstrings to warn users about this, and I'll do a run-time check and emit a warning if the experiment contains an event-based termination, how does that sound?

Sounds good to me.

MarcBerliner (approved):

itsjacobhere:
This feature would enable gradient computation while using `pybamm.Experiment` to describe a current profile right? Has this been implemented yet? Thanks!

PR #4416, opened 2024-09-04
Remove unnecessary `dt` checks

MarcBerliner:
# Description

I removed a few unnecessary checks for minimum `dt` values to proceed with the simulation.

I also removed the `start_step_offset = 1e-9` heuristic setting. For large `t_start` values, you can get floating point arithmetic errors like `t == t + 1e-9` that cause duplicate `t` values in the solution. I replaced `start_step_offset` with `np.nextafter(t, np.inf)`, which prevents these errors by giving the next largest floating point number. cc @valentinsulzer 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # [Unreleased](https://github.com/pybamm-team/PyBaMM/)
 
+## Optimizations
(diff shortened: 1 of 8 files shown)


rtimms (approved):
thanks, @MarcBerliner looks good to me!

PR #4417, opened 2024-09-05
Composite porosity

DrSOKane:
# Description

Modified reaction_driven_porosity.py so that porosity change still works for composite electrodes

Fixes # (issue)

## Type of change


brosaplanella (commented):
Hi Simon! The PR looks nice, just the tests need addressing. Note that once the unit tests are updated to increase coverage, it will also have the same parameter issues as the integration tests.

DrSOKane:
> Hi Simon! The PR looks nice, just the tests need addressing. Note that once the unit tests are updated to increase coverage, it will also have the same parameter issues as the integration tests.

That's odd. My working example ran just fine 10 days ago before I went on holiday, but it's now throwing the same parameter error as the automated tests. Was there a recent change to how phase parameters are handled?

valentinsulzer:
@parkec3 has also been working on some changes to make composite electrodes compatible with more cases

parkec3:
It looks like the degradation parameters need to be updated to be defined by phase. I'm finishing up work getting the LAM submodel compatible with composite electrodes. I had to make a custom parameter set from Chen2020 composite and Ai2020 for my own local testing. The default parameter set doesn't have those defined.

DrSOKane:
The SEI parameters are defined by phase, but not by domain, which was the underlying cause of the bug. I rewrote the code with if statements to cover three possibilities:
* Both electrodes have only one phase, in which case `phase_name` and `pref` are empty strings
* `domain` has only one phase, in which case `phase_name` is empty but `pref = "Primary: "`
* `domain` has more than one phase, in which case `phase_name` and `pref` are different for each phase
There is still a potential error if the electrode with SEI has one phase and the other has two, but that can be avoided by putting `"Primary: "` in front of the names of all SEI parameters.

DrSOKane:
UPDATE: I've simply set it so that `L_sei_0` is zero if there is no SEI on that domain. Much more elegant and means the code will run in some edge cases that would have caused an error before.

brosaplanella:
Labelled this as skip release, given that we need to wait for #3457

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@
 - Added `BasicDFN` model for sodium-ion batteries ([#4451](https://github.com/pybamm-team/PyBaMM/pull/4451))
 - Added sensitivity calculation support for `pybamm.Simulation` and `pybamm.Experiment` ([#4415](https://github.com/pybamm-team/PyBaMM/pull/4415))
 - Added OpenMP parallelization to IDAKLU solver for lists of input parameters ([#4449](https://github.com/pybamm-team/PyBaMM/pull/4449))
+- Porosity change now works for composite electrode ([#4417](https://github.com/pybamm-team/PyBaMM/pull/4417))
```
brosaplanella, line 33: This needs to be moved to the unreleased section

src/pybamm/models/submodels/porosity/reaction_driven_porosity.py
```diff
@@ -25,36 +25,61 @@ def __init__(self, param, options, x_average):
     def get_coupled_variables(self, variables):
         eps_dict = {}
         for domain in self.options.whole_cell_domains:
+            delta_eps_k = 0
             if domain == "separator":
-                delta_eps_k = 0  # separator porosity does not change
+                pass  # separator porosity does not change
             else:
```
brosaplanella, line 31: Maybe this could be refactored to use

```python3
if domain != "separator"
```

brosaplanella (commented):
Just a small couple of cosmetic changes

rtimms:
@aabills might be good for you to review this one since you've been looking at the composite code a lot recently

aabills:
> @aabills might be good for you to review this one since you've been looking at the composite code a lot recently

I'm going to review this after a meeting this morning

aabills (approved):

PR #4435, opened 2024-09-12
Plot-voltage-components-for-composite-electrodes

### LINKED ISSUE #4434: Plot Voltage Components for Composite Electrodes
### Description

I would like to see the functionality for plotting the different over potentials separated by particle phases for a composite electrode, in the `plot_voltage_components` function.

### Motivation

Having visibility into the contributions of each phase toward the net overpotentials would allow for better understanding of the dynamics in composite electrode systems.

### Possible Implementation

I suggest that the definitions of overpotentials and labels be done differently when the cell is a half-cell. I also suggest that the new 'battery' overpotential variables be defined when the composite electrode model is being used.

### Additional context

I attempted to make the changes locally, when I noticed a few interesting things that may deserve attention. I noticed in the `BaseBatteryModel.set_voltage_variables()` method, it appeared that there was no case where the secondary phase overpotentials would be extracted from the variables.

```
def set_voltage_variables(self):
        if self.options.negative["particle phases"] == "1":
            # Only one phase, no need to distinguish between
            # "primary" and "secondary"
            phase_n = ""
        else:
            # add a space so that we can use "" or (e.g.) "primary " interchangeably
            # when naming variables
            phase_n = "primary "
        if self.options.positive["particle phases"] == "1":
            phase_p = ""
        else:
            phase_p = "primary "
```

I wasn't sure if this behavior was a bug, or if this is the intended behavior. Maybe I am missing it, but I don't see where exactly the secondary phase overpotentials would be stored or accessed.

mleot:
# Description

The goal here is to allow for plotting of overpotential components in a composite electrode. Attempting to do so first in a composite half-cell

Fixes #4434

## Type of change

# Key checklist:

## Further checks:


DrSOKane:
I added a Chen2020_composite parameter set as part of PR #4463, which may be useful for this

Daniel-Nicolae23:
Reviving this discussion as it may be of interest to some people.

Stacking the primary and secondary overpotentials in this plot doesn't really make sense to me as these values are not additive in the model. Rather, the mathematical relationship for a 2-phase model is: 

$$OCV_{prim} + \eta_{conc, prim} + \eta_{reac, prim}  = OCV_{sec} + \eta_{conc, sec} + \eta_{reac, sec}$$

That's why your dashed line is below the top of the fill_betweens, you're not meant to add up all of these quantities. If you wanted to show all 6 of these voltages together, you would need to stack the 3 components of the primary starting from 0V, then overlap the 3 components of the secondary, again starting from 0V, and finally stack the remaining OPs on top of all that. 

IMO that wouldn't make for a very useful plot, as the voltages you're probably most interested in will be so cluttered. 
What we can do is add a keyword argument to `sol.plot_voltage_components()` to select which phase you want to plot. Something like this:

```python
sol.plot_voltage_components(split_by_electrode=True, electrode_phases=("secondary", "primary"))
```
image

This is a Si-Gr model, you can see the different behaviours during the rest. 

I'm happy to open another draft PR with the code for the `plot_voltage_components()`, if people would like this feature.

PR #4441, opened 2024-09-15
Add telemetry

valentinsulzer:
# Description

Adds very basic telemetry - records when a simulation is solved.

The following section has been added to the user guide to explain the telemetry we are doing:

> PyBaMM collects anonymous usage data to help improve the library. This telemetry is enabled by default but can be easily disabled. Here's what you need to know:
> 
> - **What is collected**: Basic usage information like PyBaMM version, Python version, and which functions are run.
> - **Why**: To understand how PyBaMM is used and prioritize development efforts.
> - **Opt-out**: To disable telemetry, set the environment variable `PYBAMM_OPTOUT_TELEMETRY=true` or use `pybamm.telemetry.disable()` in your code.
> - **Privacy**: No personal information (name, email, etc) or sensitive information (parameter values, simulation results) is ever collected.


src/pybamm/config.py
```diff
@@ -0,0 +1,85 @@
...
+import pathlib
+import os
+
+
+def _get_config_file():
+    # Get the home directory
+    home_dir = pathlib.Path.home()
+
+    # Create the file path for pybamm config
+    config_file = home_dir / ".config" / "pybamm" / "config.yml"
```
agriyakhetarpal, line 11: I don't have a strong opinion on adding telemetry as a whole. However, a better cross-platform alternative for these lines instead of hardcoding is to use `user_config_dir` from [`platformdirs`](https://pypi.org/project/platformdirs/). I think it would be okay to add this as a dependency, because it is minimal and not likely to break things.

agriyakhetarpal:
We could also make `pybamm.telemetry.disable()` a top-level `pytest` fixture to disable telemetry when running tests

src/pybamm/config.py
valentinsulzer: I tested it out and it looks like the files in `platformdirs.user_config_dir()` (`/Users/vsulzer/Library/Application Support`) are local apps like Zotero, Cursor, etc, whereas packages like `dask` and `julia` have saved their configs under `~/.config`. Are julia and dask just doing it wrong?

valentinsulzer:
> We could also make pybamm.telemetry.disable() a top-level pytest fixture as a cleaner way to disable telemetry when running tests

Yes, but we also want to disable it when our dependencies run their tests, so that wouldn't be enough right?

agriyakhetarpal:
> > We could also make pybamm.telemetry.disable() a top-level pytest fixture as a cleaner way to disable telemetry when running tests
> 
> Yes, but we also want to disable it when our dependencies run their tests, so that wouldn't be enough right?

Oh, yes, forgot about that! In that case, we would need both a fixture and the logic you added.

src/pybamm/config.py
agriyakhetarpal: 🤔 I think there is no right answer. `pipx` prior to version 1.5.0 used the former location to store binaries, but it now avoids it because there is a space between the path ("Application Support"). However, other packages such as `pooch`, SciPy, `hatch`, Jupyter, uv, and more save files under there by default, since they don't store binaries there. The latter location has a dot (".config") in its path, which makes it a hidden folder and harder to discover by default. While that wouldn't matter if we provide a Pythonic interface to handle it, it would not be a standard location on Windows, where `%AppData%` is a better choice. I would suggest going with the former, because we already use `pooch` and it seems to be the more popular choice across applications.

kratman (commented):
I understand why we want to know which parts of PyBaMM our users are using, but I am not sure this is the best approach.

My concerns:
- I feel like I would be upset to find out a free research tool was tracking my usage and sending it to a server. I think for a paid tool the expectation that they want to know your usage is a little more reasonable.
- Users installing through pip might not realize we are tracking their usage. It would be better to allow an opt-in instead so that users know we are collecting their data.
- Are we only going to be tracking simulation data so that we can just keep it in the simulation class or are we going to have to add the tracking to a lot of different places and make sure we add it to new functionality?

pyproject.toml
```diff
@@ -280,3 +281,6 @@ module = [
         "pybamm.models.full_battery_models.base_battery_model.*"
         ]
 disable_error_code = "attr-defined"
+
+[tool.poetry.scripts]
+post-install = "scripts.post_install:run"
```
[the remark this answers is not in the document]
valentinsulzer: removed

examples/scripts/compare_lithium_ion.py
```diff
@@ -4,6 +4,7 @@
 import pybamm
 
 pybamm.set_logging_level("INFO")
+pybamm.telemetry.disable()
```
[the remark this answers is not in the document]
valentinsulzer: fixed

src/pybamm/config.py
```diff
@@ -0,0 +1,123 @@
...
+    print(
+        "PyBaMM can collect usage data and send it to the PyBaMM team to "
+        "help us improve the software.\n"
+        "We do not collect any sensitive information such as models, parameters, "
+        "or simulation results - only information on which parts of the code are "
+        "being used and how frequently.\n"
+        "This is entirely optional and does not impact the functionality of PyBaMM.\n"
+        "For more information, see https://docs.pybamm.org/en/latest/source/user_guide/index.html#telemetry"
+    )
+    while True:
```
martinjrobins, line 62: this should time-out after a short period with default no so we don't break everyone's automated stuff (say 10-20 sec?)

src/pybamm/config.py
```diff
@@ -0,0 +1,123 @@
...
+        user_input = input("Do you want to enable telemetry? (Y/n): ").strip().lower()
+        if user_input in ["yes", "y"]:
```
martinjrobins, line 64:
```suggestion
        if user_input in ["yes", "y", ""]:
```

So if user presses enter it will be interpreted as a yes (as per the prompt)

martinjrobins (changes requested):
Thanks @valentinsulzer. I suggest a timeout with default opt-out so we don't break ppl's CI, see below

src/pybamm/config.py
agriyakhetarpal: A better option could be to check for the `ci_env_vars` above (can be set as a constant at the top of the file) and assume non-interactive if set, and return early (like `nox` does it with `session.interactive`).

valentinsulzer: This function is only ever called by `generate` which already returns early if it detects an automated environment, but I'll also add the time check as a backup

valentinsulzer:
Added an input with timeout but can't figure out why the docs are failing as I can't reproduce it locally

martinjrobins (dismissed):
looks good to me, I'm not sure what is causing the doctest failure though...?

src/pybamm/config.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+def ask_user_opt_in(timeout=10):
+    """
+    Ask the user if they want to opt in to telemetry.
+
+    Parameters
+    ----------
+    timeout: float, optional
+        The timeout for the user to respond to the prompt. Default is 20 seconds.
```
kratman, line 52: Default above appears to be 10 seconds
```suggestion
        The timeout for the user to respond to the prompt. Default is 10 seconds.
```

src/pybamm/config.py
```diff
@@ -0,0 +1,141 @@
...
+        except TimeoutOccurred:
+            print("\nTimeout reached. Defaulting to not enabling telemetry.")
+            return False
+
+        if user_input in ["yes", "y", ""]:
+            return True
+        elif user_input in ["no", "n"]:
+            return False
+        else:
+            print("\nInvalid input. Please enter 'yes'/'y' or 'no'/'n'.")
```
kratman, line 87: If yes, y, and no response are already yes, then we can probably just assume that anything else is no
```suggestion
        else:
            return False
```

valentinsulzer: I would prefer to avoid the situation where people do want to enable telemetry but disable it by accident

agriyakhetarpal (commented):
The code looks fine, the only thing problematic here, at least to me, seems to be the dependency added on `inputimeout`: https://pypi.org/project/inputimeout hasn't had a release in the last six years. Could we replace it with a different dependency that can do the same task? I'm just worried on the prospect of it causing issues with constraints or with us upgrading to new Python versions as they come some time down the line.

src/pybamm/config.py
```diff
@@ -0,0 +1,138 @@
...
+
+    # Check for other common CI environment variables
+    ci_env_vars = ["CI", "TRAVIS", "CIRCLECI", "JENKINS_URL", "GITLAB_CI"]
+    if any(var in os.environ for var in ci_env_vars):
+        return True
+
+    # Check for common test runner names in command-line arguments
+    test_runners = ["pytest", "unittest", "nose", "trial", "nox", "tox"]
+    if any(runner in sys.argv[0].lower() for runner in test_runners):
+        return True
```
agriyakhetarpal, line 37: This should check across the entire list and not just the first index – this would work for commands like `pytest -m unit`, but not for cases where one runs tools through `nox`, or `uvx pytest`/`pipx pytest`, and similar.

src/pybamm/config.py
```diff
@@ -0,0 +1,138 @@
...
+
+    # Check if building docs with Sphinx
+    if "sphinx" in sys.modules:
+        return True
```
agriyakhetarpal, line 41: I think this should be stricter: https://stackoverflow.com/a/51637094/14001839. Also, some `conf.py` files across projects I noted when reviewing this don't use `import sphinx`, but they import only a sub-package, such as `from sphinx.util import foobar`, et cetera, so we should check for those cases too if we can.

src/pybamm/telemetry.py
```diff
@@ -0,0 +1,36 @@
...
+    project_api_key="phc_bLZKBW03XjgiRhbWnPsnKPr0iw0z03fA6ZZYjxgW7ej",
+    host="https://us.i.posthog.com",
+)
+
+
+def disable():
+    _posthog.disabled = True
+
+
+_opt_out = os.getenv("PYBAMM_DISABLE_TELEMETRY", "false").lower()
```
agriyakhetarpal, line 17: We should add this environment variable in the `PYBAMM_ENV` constant in `noxfile.py` as an additional measure

agriyakhetarpal (commented):
Thanks, @valentinsulzer! There are many `#pragma: no cover` statements here, I think they can be cleaned up. We can add this to `.github/codecov.yml`:

```yml
ignore:
  - src/pybamm/config.py
  # and so on
```

Also, there is another problem here that I feel deserves another look. I tried a notebook from this branch, and there's no option for me to provide a Y/n answer for opting in/out of telemetry. One would have to wait for ten seconds to default to a "no", which isn't a lot, but this doesn't make for good UX. Is there a way to detect if one is in an IPython kernel and reduce the timeout (or accept stdin in a different manner)?

pyproject.toml
```diff
@@ -44,6 +44,7 @@ dependencies = [
     "typing-extensions>=4.10.0",
     "pandas>=1.5.0",
     "pooch>=1.8.1",
+    "posthog",
```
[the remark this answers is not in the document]
valentinsulzer: done

src/pybamm/config.py
valentinsulzer: Modified based on this test, which prints True:
```
from sphinx.util import logger

import sys

if any(mod.startswith("sphinx") for mod in sys.modules):
    print("True")
else:
    print("False")
```

valentinsulzer: This found `sphinxcontrib` as a false positive

valentinsulzer:
Updated to work for notebooks. Re: coverage, I don't want to skip the whole file since it's possible to test some parts of the file

agriyakhetarpal (dismissed):
Thanks, @valentinsulzer! Two more things:
- I started a PR to include this in the release notes just in case we forget it (please feel free to modify what I wrote and push any changes to it directly): https://github.com/pybamm-team/pybamm.org/pull/214 and I read up a bit on PostHog's data protection policies: https://posthog.com/docs/privacy/gdpr-compliance. Should we switch to their EU-based cloud instance instead of the current US one to better achieve GDPR compliance (not that we collect anything non-anonymised right now, but just in case we start collecting more data points in the future)? We use Plausible for the docs analytics already which is compliant, so that makes an additional case for changing.
- Since we're tracking when a simulation is solved, it might also make sense to capture the host operating system (just once at the time of initial config setup) since our compiled solvers are in some ways supported differently across platforms and that can help with platform-specific issues later on.

agriyakhetarpal:
Thanks, @valentinsulzer, I've approved the code changes already but my points above still apply

rtimms (approved):

valentinsulzer:
Thanks @agriyakhetarpal , let's chat on the pybamm.org PR about the announcement. Feel free to open additional PRs with more logging as necessary

PR #4443, opened 2024-09-16
Final pytest migration

prady0t:
WIP


prady0t:
Just two files left 
`test_idaklu_jax.py` : remove parameterized dependency.
`test_solution.py` : problem with converting `assertLogs` to pytest

agriyakhetarpal:
Unfortunately, there's no standard way to replace `self.assertLogs()` as far as I could find. I think the best way is to hack into PyBaMM's logger and retrieve the result you want, which is okay since we have just one test that uses this. For example,

```python
import io
import logging
log_capture = io.StringIO()
handler = logging.StreamHandler(log_capture)
handler.setLevel(logging.ERROR)
logger = logging.getLogger('pybamm.logger')
logger.addHandler(handler)
```

and then, if you run the code that should produce the logger warning: `pybamm.Solution(ts, bad_ys, model, {})`, you can check the output using `log_capture.getvalue()` and check if `"exceeds the maximum"` string is in there. Later, you can destroy the handler using `logger.removeHandler(handler)` and clean it up, so that it doesn't affect any other tests. This worked for me as a quick experiment, could you try and see how it holds up in the test suite?

agriyakhetarpal:
@prady0t, how is this going? The tests are all already passing here, and the approach I suggested should be okay in general – it works on my machine. Please let me know if you need additional help with this.

prady0t:
Hi. I've updated the files. Lets wait for CI to run.

pyproject.toml
```diff
@@ -105,12 +105,11 @@ dev = [
     "pytest-cov",
     # For doctest
     "pytest-doctestplus",
-    # For test parameterization
-    "parameterized>=0.9",
     # pytest and its plugins
     "pytest>=6",
     "pytest-xdist",
     "pytest-mock",
+    "pytest-subtests",
```
agriyakhetarpal, line 112: Could you add this to the installation guide? Here: https://docs.pybamm.org/en/latest/source/user_guide/installation/index.html#optional-dependencies

agriyakhetarpal (commented):
Thanks, @prady0t – I guess the next step to do after this PR should be to disallow the use of `unittest` through the Ruff rule as we discussed earlier

Edit: I notice some failing tests with the parameterisation, could you try fixing them locally?

pyproject.toml
prady0t: Let me add this in another commit

prady0t: Done

agriyakhetarpal (approved):
Thanks for completing the migration, @prady0t!

PR #4449, opened 2024-09-17
I4087-multiprocessing-2

### LINKED ISSUE #4087: refactor multiprocessing and multiple inputs

martinjrobins:
# Description

- idaklu solver can run multiple simulations in parallel using openmp
- Previously this was done using Python multiprocessing, but this has been removed and instead a list of inputs can be passed directly to the `_integrate` function of the idaklu solver, the solver then runs these simulations in parallel using openmp

Partially fixes #4087 (just for IDAKLU solver). This is a redo of some of #4260 (plan is to split this old PR into multiple smaller PRs, starting with the openmp parallisation of the IDAKLU solver).

## Type of change

# Key checklist:

## Further checks:


agriyakhetarpal:
@martinjrobins, I'm unsure how the Homebrew OpenMP detection on macOS was working before this PR (my guess is that CVODE was able to find it on its own, but maybe since we now use the header directly, we need to pass its path?). Could you also trigger the wheel build from this branch? We use a different `libomp.dylib` there rather than the one from Homebrew, so that the wheels don't receive a higher macOS deployment target

agriyakhetarpal:
@martinjrobins, I'm unsure how the Homebrew OpenMP detection on macOS was working before this PR (my guess is that CVODE was able to find it on its own, but maybe since we now use the header directly, we need to pass its path?). Could you also trigger the wheel build from this branch? We use a different `libomp.dylib` there rather than the one from Homebrew, so that the wheels don't receive a higher macOS deployment target.

MarcBerliner:
> I'm unsure how the Homebrew OpenMP detection on macOS was working before this PR (my guess is that CVODE was able to find it on its own, but maybe since we now use the header directly, we need to pass its path?). Could you also trigger the wheel build from this branch? We use a different `libomp.dylib` there rather than the one from Homebrew, so that the wheels don't receive a higher macOS deployment target.

I think I'm running into this issue. I'm trying to install this on macOS, and I get the error
```cpp
/PyBaMM/src/pybamm/solvers/c_solvers/idaklu/IDAKLUSolverGroup.cpp:2:10: fatal error: 'omp.h' file not found
      #include <omp.h>
```

agriyakhetarpal:
@MarcBerliner, I think passing `-I$(brew --prefix libomp)/include` should work

agriyakhetarpal:
`CXXFLAGS="-I $(brew --prefix libomp)/include" nox -s dev` works for me locally. I guess we should set it in the noxfile when running on macOS, so that we get passing builds for both CI tests and locally.

src/pybamm/solvers/base_solver.py
```diff
@@ -896,36 +896,37 @@ def solve(
...
-                new_solution = self._integrate(
-                    model,
-                    t_eval[start_index:end_index],
-                    model_inputs_list[0],
-                    t_interp=t_interp,
-                )
-                new_solutions = [new_solution]
-            elif model.convert_to_format == "jax":
-                # Jax can parallelize over the inputs efficiently
+            if isinstance(self, (pybamm.JaxSolver, pybamm.IDAKLUSolver)):
```
MarcBerliner, line 899: Can we add a solver property for this

src/pybamm/solvers/base_solver.py
```diff
@@ -1350,7 +1351,13 @@ def step(
         # Step
         pybamm.logger.verbose(f"Stepping for {t_start_shifted:.0f} < t < {t_end:.0f}")
         timer.reset()
-        solution = self._integrate(model, t_eval, model_inputs, t_interp)
+
+        # API for _integrate is different for JaxSolver and IDAKLUSolver
+        if isinstance(self, (pybamm.JaxSolver, pybamm.IDAKLUSolver)):
```
MarcBerliner, line 1356: ^likewise

src/pybamm/solvers/c_solvers/idaklu/common.hpp
```diff
@@ -31,8 +31,8 @@
 #include <pybind11/stl.h>
 
 namespace py = pybind11;
-using np_array = py::array_t<realtype>;
-using np_array_dense = py::array_t<realtype, py::array::c_style | py::array::forcecast>;
+// note: we rely on c_style ordering for numpy arrays so don't change this!
```
MarcBerliner, line 34: This is outside the scope of this PR, but we should be more careful about this in pybamm. We transpose the `y` when we store it with the python `Solution` object, so it's f-contiguous there. Not a big deal, but it'd make life simpler to keep it consistent.

src/pybamm/solvers/idaklu_solver.py
```diff
@@ -736,22 +742,30 @@ def _integrate(self, model, t_eval, inputs_dict=None, t_interp=None):
...
         else:
             inputs = np.array([[]])
 
-        y0full = model.y0full
-        ydot0full = model.ydot0full
+        # stack y0full and ydot0full so they are a 2D array of shape (number_of_inputs, number_of_states + number_of_parameters * number_of_states)
+        # note that y0full and ydot0full are currently 1D arrays (i.e. independent of inputs), but in the future we will support
+        # different initial conditions for different inputs (see https://github.com/pybamm-team/PyBaMM/pull/4260). For now we just repeat the same initial conditions for each input
+        y0full = np.vstack([model.y0full] * len(inputs_list))
+        ydot0full = np.vstack([model.ydot0full] * len(inputs_list))
```
MarcBerliner, line 768: The `y0full` values include the initial values of the differential terms, which may depend on input parameters. These should all be initialized

MarcBerliner (changes requested):
Thanks @martinjrobins, this will be really nice for optimization.

One high-level question: how does this handle solver convergence errors? If we batch several simulations, we don't want a single ill-conditioned set of parameters to ruin the whole batch with an exception. Do we always receive a solution for the simulations that converge?

Also, this fix to the `noxfile.py` solved my local build issue with macOS (thanks @agriyakhetarpal):
```python
PYBAMM_ENV = {
    "LD_LIBRARY_PATH": f"{homedir}/.local/lib",
    "PYTHONIOENCODING": "utf-8",
    "MPLBACKEND": "Agg",
    # Expression evaluators (...EXPR_CASADI cannot be fully disabled at this time)
    "PYBAMM_IDAKLU_EXPR_CASADI": os.getenv("PYBAMM_IDAKLU_EXPR_CASADI", "ON"),
    "PYBAMM_IDAKLU_EXPR_IREE": set_iree_state(),
    "IREE_INDEX_URL": os.getenv(
        "IREE_INDEX_URL", "https://iree.dev/pip-release-links.html"
    ),
    "CXXFLAGS": f"-I {os.popen('brew --prefix libomp').read().strip()}/include",
}
```

martinjrobins:
> `CXXFLAGS="-I $(brew --prefix libomp)/include" nox -s dev` works for me locally. I guess we should set `OpenMP_ROOT` (or `CXXFLAGS`) in the noxfile when running on macOS, so that we get passing builds for both CI tests and locally.

Yea, I put this into the CI and it works fine (see last commit). Let me know if that is the best way to do it?

src/pybamm/solvers/c_solvers/idaklu/common.hpp
martinjrobins: Agreed, I had to write the code assuming a particular ordering hence the specification as `c_style`. If an f_style ordered array is passed in then it gets converted to c_style by pybind11 (according to their docs: https://pybind11.readthedocs.io/en/stable/advanced/pycpp/numpy.html). This isn't ideal since it would involve a copy, but I wasn't sure about a better way of doing it. Let me know if you have any ideas

src/pybamm/solvers/idaklu_solver.py
martinjrobins: these should have already been initialised during the setup correct?

src/pybamm/solvers/c_solvers/idaklu/common.hpp
MarcBerliner: Interesting, I didn't realize pybind11 made a copy here. IMO, I think we should stick with `c_style` and make python-side changes to `y` to avoid the transpose and change its ordering from `y[time, var]` to `y[var, time]`

martinjrobins:
> One high-level question: how does this handle solver convergence errors?

It should be the same behaviour as currently. If the retval from IDASolve indicates it has failed the individual solver will stop early and return. All the other solves, if they converge ok, carry on and solve for the complete duration and return normally. The retval is stored in the solution so the user can tell if that individual solve has failed or not.

src/pybamm/solvers/idaklu_solver.py
MarcBerliner: By setup, do you mean in `IDACalcIC`? This gets handled during the call to `_set_consistent_initialization`, but initial values of differential terms cannot be resolved by `IDACalcIC`

agriyakhetarpal:
> > `CXXFLAGS="-I $(brew --prefix libomp)/include" nox -s dev` works for me locally. I guess we should set `OpenMP_ROOT` (or `CXXFLAGS`) in the noxfile when running on macOS, so that we get passing builds for both CI tests and locally.
> 
> Yea, I put this into the CI and it works fine (see last commit). Let me know if that is the best way to do it?

Yes. Though, this would fail if someone tries to install without our `nox -s dev` helper (i.e., with `pip install -e .`). Could we include this in our CMake subprocess call in `setup.py`? That should work everywhere.

I'm off to a conference this week, so I haven't re-reviewed #4352 fully yet. We'll have to think how to do it there, because `setup.py` would be removed with that PR.

MarcBerliner:
> It should be the same behaviour as currently. If the retval from IDASolve indicates it has failed the individual solver will stop early and return. All the other solves, if they converge ok, carry on and solve for the complete duration and return normally. The retval is stored in the solution so the user can tell if that individual solve has failed or not.

It looks like we throw an error for any simulation failure in `_post_process_solution`, so the user wouldn't be able to examine the `retval`s
https://github.com/pybamm-team/PyBaMM/blob/508767342feabd75941daa0201010a67a4a4e151/src/pybamm/solvers/idaklu_solver.py#L830-L835

We could change this to a warning (or make error vs. warning a solver option).

src/pybamm/solvers/idaklu_solver.py
martinjrobins: no, I did mean _set_consistent_initialization. But this function only has a single input dict, if you pass in a list of parameters that affect the initial condition you get an error. Part of #4260 was that if you pass an input list (that affects the initial conditions) this would be handled appropriately so that y0full is a 2d array where the rows are possible different corresponding to different parameters in the list (unlike now where each row is identical)

martinjrobins:
> We could change this to a warning (or make error vs. warning a solver option).

hmmm, this would be a big change to the API if someone was relying on the solve throwing an error if it failed. Plus the other solvers raise errors as well (even for multiple input parameters). I agree it would sometime be better to carry on if a solve fails though, but I'd say this would be a different issue (and probably one to discuss at the developers meeting)

src/pybamm/solvers/idaklu_solver.py
MarcBerliner: Ah ok, got it

martinjrobins:
> Yes. Though, this would fail if someone tries to install without our `nox -s dev` helper (i.e., with `pip install -e .`). Could we include this in our CMake subprocess call in `setup.py`? That should work everywhere.

I'll see if I can do it in CMakeLists.txt, since we're getting rid of setup.py

MarcBerliner:
> hmmm, this would be a big change to the API if someone was relying on the solve throwing an error if it failed. Plus the other solvers raise errors as well (even for multiple input parameters). I agree it would sometime be better to carry on if a solve fails though, but I'd say this would be a different issue (and probably one to discuss at the developers meeting)

Yeah, if this behavior is consistent with the other solvers' multiprocessing, then let's table this for now and bring this up at the meeting. Otherwise LGTM (besides the build issue)

MarcBerliner (approved):

kratman:
@martinjrobins Did you trigger the wheel build to make sure that works?

PR #4451, opened 2024-09-18
Implement sodium-ion DFN

brosaplanella:
# Description

Implemented DFN for sodium-ion (same as lithium-ion but in a separate file to avoid conflicts). Implemented parameter value from Chayambuka et al (2022) (actual values and data taken from [this COMSOL example](https://www.comsol.com/model/1d-isothermal-sodium-ion-battery-117341)). 

When adding the CITATION I also modified many citation keys to be consistent with the format (FirstAuthorYear). Hence the many files changed.

Fixes #4397 (not an issue but a discussion)

## Type of change

# Key checklist:

## Further checks:


src/pybamm/models/full_battery_models/sodium_ion/basic_dfn.py
```diff
@@ -0,0 +1,273 @@
+#
+# Basic Doyle-Fuller-Newman (DFN) Model
+#
+import pybamm
+
+
+class BasicDFN(pybamm.lithium_ion.BaseModel):
+    """Doyle-Fuller-Newman (DFN) model of a sodium-ion battery, from
+    :footcite:t:`Marquis2019`.
```
rtimms, line 9: Does the Na-ion paper have the same model equations?  It would be good to reference that paper here for more context I think, and then say eqns are from Marquis (except they aren't quite ahah)

rtimms (commented):
thanks @brosaplanella looks good, can you add a test in the basic model tests?

src/pybamm/models/full_battery_models/sodium_ion/basic_dfn.py
brosaplanella: Not quite, but because I took the parameters from the COMSOL implementation it is actually quite similar to the Marquis et al. Happy to cite either, not sure what will be less confusing to the users.

brosaplanella:
Doctests are failing but I don't know why. @agriyakhetarpal can you help with this?

src/pybamm/parameters/parameter_sets.py
agriyakhetarpal, line 1: @brosaplanella, I can't comment on the line directly since it hasn't been modified, so I'm commenting on the file – but the failure is in line 29. Changing it to `:footcite:t:`Rieger2016` should fix the doctest failure.

src/pybamm/input/parameters/sodium_ion/Chayambuka2022.py
```diff
@@ -0,0 +1,341 @@
...
+        "Negative electrode exchange-current density [A.m-2]"
+        "": HC_electrolyte_exchange_current_density_Chayambuka2022,
+        "Negative electrode OCP entropic change [V.K-1]": 0,
+        # positive electrode
+        "Positive electrode conductivity [S.m-1]": 50,
+        "Maximum concentration in positive electrode [mol.m-3]": 15320,
+        "Positive particle diffusivity [m2.s-1]": NVPF_diffusivity_Chayambuka2022,
+        "Positive electrode OCP [V]": NVPF_ocp_Chayambuka2022,
+        "Positive electrode porosity": 0.23,
+        "Positive electrode active material volume fraction": 1 - 0.23 - 0.22,
```
kratman, line 310: Any reason why this is not just a single number?

src/pybamm/input/parameters/sodium_ion/Chayambuka2022.py
```diff
@@ -0,0 +1,341 @@
...
+        "Nominal cell capacity [A.h]": 3e-3,
+        "Current function [A]": 3e-3,
+        "Contact resistance [Ohm]": 0,
+        # negative electrode
+        "Negative electrode conductivity [S.m-1]": 256,
+        "Maximum concentration in negative electrode [mol.m-3]": 14540,
+        "Negative particle diffusivity [m2.s-1]": HC_diffusivity_Chayambuka2022,
+        "Negative electrode OCP [V]": HC_ocp_Chayambuka2022,
+        "Negative electrode porosity": 0.51,
+        "Negative electrode active material volume fraction": 1 - 0.51 - 0.001,
```
kratman, line 296: Same as the other comment

src/pybamm/models/full_battery_models/sodium_ion/basic_dfn.py
```diff
@@ -0,0 +1,273 @@
...
+
+    """
+
+    def __init__(self, name="Doyle-Fuller-Newman model"):
+        super().__init__(name=name)
+        pybamm.citations.register("Marquis2019")
+        # `param` is a class containing all the relevant parameters and functions for
+        # this model. These are purely symbolic at this stage, and will be set by the
+        # `ParameterValues` class when the model is processed.
+        param = self.param
```
kratman, line 24: if these are the same, then why not just use self.param instead?

src/pybamm/input/parameters/sodium_ion/Chayambuka2022.py
brosaplanella: This number is not directly reported in the paper, they provide the other two and say everything adds up to one. Given that one needs to go through the paper to find this, I thought it would be better to leave it this way so it is apparent what's going on. I will add a comment to make it clear.

src/pybamm/models/full_battery_models/sodium_ion/basic_dfn.py
brosaplanella: Actually, I just copied this from the lithium_ion BasicDFN. I believe we do that in all the basic models, so we don't have to do self.param every time (there any a lot of occurrences) and the code is easier to read. I personally think it is useful as it is easier to read through the model.

src/pybamm/input/parameters/sodium_ion/Chayambuka2022.py
kratman: Personally, I think this just looks confusing. I don't see any harm in making it a single number

DrSOKane (approved):
Amazing work! The next step will be to implement the full DFN (with submodels) for sodium-ion, but this is a welcome and necessary first step.

DrSOKane:
> Amazing work! The next step will be to implement the full DFN (with submodels) for sodium-ion, but this is a welcome and necessary first step.

All you need to do is edit CHANGELOG and this should be ready to merge.

tests/integration/test_models/test_full_battery_models/test_sodium_ion/test_basic_models.py
```diff
@@ -0,0 +1,25 @@
...
+        model = self.model
+        experiment = pybamm.Experiment(
+            [
+                "Discharge at C/3 until 3.5V",
+                "Hold at 3.5V for 1 hour",
+                "Rest for 10 min",
+            ]
+        )
+        sim = pybamm.Simulation(model, experiment=experiment)
+        sim.solve(calc_esoh=False)
```
kratman, line 19: This test does not assert anything. It does not really seem to add any value

kratman: Can this be changed or removed?

kratman (changes requested):

src/pybamm/models/full_battery_models/sodium_ion/basic_dfn.py
brosaplanella: There's 63 occurrences of `param = self.param` across models, so will open a separate issue & PR for that (will also make it easier to review).

kratman (approved):
I can't vouch for the physics, but the code seems fine to me

kratman:
Docs all failed, so I will check on that too

kratman:
Docs failures are unreleated. I am looking into it, but will fix it in a different PR

PR #4453, opened 2024-09-19
Faster (re)initialization of ODEs in `IDA`

MarcBerliner:
# Description

In `IDA`, we can compute an analytical solution for consistent initialization/reinitialization of ODEs, which avoids an expensive call to `IDACalcIC`. This change gives a modest speed improvement to all the ODE models solved by the `IDAKLUSolver` and will give a bigger speedup to those with a dense `t_eval` vector. For example,

```python
import pybamm
import numpy as np

sim = pybamm.Simulation(model, solver=solver)

t_eval = np.linspace(0, 3600, 10000)
sol = sim.solve(t_eval)

print(sol.integration_time)
```
- New `integration_time`: **367 ms**
- Old `integration_time`: **514 ms**
- Speedup: **1.4x**

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/src/pybamm/solvers/c_solvers/idaklu/IDAKLUSolverOpenMP.inl
+++ b/src/pybamm/solvers/c_solvers/idaklu/IDAKLUSolverOpenMP.inl
@@ -298,9 +298,10 @@ void IDAKLUSolverOpenMP<ExprSet>::Initialize() {
   // Determine if the system is an ODE
   is_ODE = number_of_states > 0;
   for (int ii = 0; ii < number_of_states; ii++) {
-    const bool id_i = id_np_val[ii];
-    id_val[ii] = id_i;
-    is_ODE &= id_i;
+    id_val[ii] = id_np_val[ii];
+    // check if id_val[ii] approximately equals 1 (>0.999) handles
+    // cases where id_val[ii] is not exactly 1 due to numerical errors
+    is_ODE &= id_val[ii] > 0.999;
   }
 
   // Variable types: differential (1) and algebraic (0)
--- a/src/pybamm/solvers/c_solvers/idaklu/IDAKLUSolverOpenMP.inl
+++ b/src/pybamm/solvers/c_solvers/idaklu/IDAKLUSolverOpenMP.inl
@@ -83,6 +83,10 @@ IDAKLUSolverOpenMP<ExprSet>::IDAKLUSolverOpenMP(
   if (this->setup_opts.preconditioner != "none") {
     precon_type = SUN_PREC_LEFT;
   }
+
+  // The default is to solve a DAE for generality. This may be changed
+  // to an ODE during the Initialize() call
+  is_ODE = false;
 }
 
 template <class ExprSet>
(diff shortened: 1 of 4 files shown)


src/pybamm/solvers/c_solvers/idaklu/IDAKLUSolverOpenMP.inl
```diff
@@ -289,9 +295,13 @@ void IDAKLUSolverOpenMP<ExprSet>::Initialize() {
   realtype *id_val;
   id_val = N_VGetArrayPointer(id);
 
+  // Determine if the system is an ODE
+  is_ODE = number_of_states > 0;
   int ii;
   for (ii = 0; ii < number_of_states; ii++) {
```
kratman, line 301: Unrelated, but cleaner this way.
```suggestion
  for (int ii = 0; ii < number_of_states; ii++) {
```

src/pybamm/solvers/c_solvers/idaklu/IDAKLUSolverOpenMP.inl
```diff
@@ -289,9 +295,13 @@ void IDAKLUSolverOpenMP<ExprSet>::Initialize() {
...
-    id_val[ii] = id_np_val[ii];
+    const bool id_i = id_np_val[ii];
+    id_val[ii] = id_i;
+    is_ODE &= id_i == 1;
```
kratman, line 304: If `id_i` is a bool, then the `id_i == 1` should be redundant

martinjrobins: This bit is a confusing jumble of types, these are not really bool, but are floats indicating if the state is differential (1.0) or algebraic (0.0), this is just the format that sundials expects. I think introducing a bool here is just confusing an already confusing sundials thing (e.g. they never say in the docs what happens to values that are not 0 or 1!), perhaps:
```c++
id_val[ii] = id_np_val[ii];
is_ODE &= id_val[ii] == 1.0
```

martinjrobins (changes requested):
looks great @MarcBerliner, just that one minor comment about the bool. I'm supprised that sundials doesn't do this in CalcIC, it has the id vector so can tell if it is an ode or not. Great find!

src/pybamm/solvers/c_solvers/idaklu/IDAKLUSolverOpenMP.inl
kratman: Float comparisons are not reliable either. Float equality checks are not good to have in the code

MarcBerliner: @martinjrobins good point! I'll make this change

MarcBerliner: I made this an inequalty check to handle minor floating point arithmetic issues

martinjrobins: @kratman: totally agree! but this is how sundials docs define this vector (as a vector of floats, 1.0 = differential and 0.0 == algebraic) so I think we have to follow this definition, no matter how strange it is.

martinjrobins: saying that, its not a big deal so happy with `is_ODE &= id_val[ii] > SMALL_NUMBER`, as long as the type conversions are removed. it shouldn't make any difference to the result

martinjrobins (approved):

kratman (approved):

src/pybamm/solvers/c_solvers/idaklu/IDAKLUSolverOpenMP.inl
kratman: Yeah sorry for the confusion, I was talking to Marc separately removing the type conversion was already part of it

PR #4464, opened 2024-09-24
Fast Hermite interpolation and observables

### LINKED ISSUE #4419: Post-solve Hermite interpolation in Sundials solvers
### Description

This will give higher accuracy interpolation for observables after solving the system of equations. 

### Motivation

_No response_

### Possible Implementation

_No response_

### Additional context

_No response_

MarcBerliner:
# Description

This PR adds Hermite interpolation to the `IDAKLUSolver`, which gives more accurate post-solve interpolation accuracy and improves performance by moving code from Python to C++. The high-frequency example below is **25x** faster.

image

For simulations that save the adaptive time stepping states `y`, the new solver option `hermite_interpolation` uses the time derivatives `yp` to obtain a more accurate interpolator. Previously, our post-solve interpolation involved:
1. Computing a variable `v` for the given raw data `sol.all_ts` and `sol.all_ys` (like `v_raw = sol[v].data`)
2. Linear interpolating `v_raw` onto new time points `t_new` (like `np.interp1d(t_new, sol.t, v_raw)`)

This was problematic because linear interpolation does not always provide an accurate solution (like in the example picture above). With the new Hermite interpolation, the two steps are combined into a single step that uses the state derivatives:
- Create a `y_new` matrix by interpolating `sol.all_ys` onto `t_new` accurately using `sol.all_yps`, then observe `v` at the `t_new` and `y_new` values. The Hermite interpolation API is just `sol[v](t)`
- We can still run `sol[v].data` to access the `v_raw` data points at the `sol.t` points

These changes are most visible with very large datasets. Here are the new timings to get an accurate version of the `SPM` with `t_new = np.linspace(t0, tf, 1000000)`:
- Dense `t_interp` (old, Python): **10.6 s**
- Hermite interp (new, C++): **0.421 s**

**giving a 25x speedup for this example.**

Main changes in the code:
- Moves the solution indexing (including `sol["Voltage [V]"].data` or `sol["Voltage [V]"](t)`) from Python to C++
- Refactors the `ProcessedVariable` class and adds subclasses like `ProcessedVariable0D`, `ProcessedVariable1D`, etc.
- Adds an `all_yps` term to the `Solution` class
- Remove unused `warn` argument from the interpolator

Notes:
- Hermite interpolation can only work if we have all the adaptive time stepping `y` and `yp` values. This means simulations with `output_variables` or `t_interp` values cannot be supported.
- Sensitivity analyses are not currently using the C++ observable interface.
- There are a few places in `ProcessedVariable` that will be cleaned up once we move to a standalone `IDAKLU` package

Fixes #4419

## Type of change

# Key checklist:

## Further checks:


src/pybamm/solvers/processed_variable.py
```diff
@@ -76,45 +72,416 @@ def __init__(
...
 
         # xr_data_array is initialized
-        self._xr_data_array = None
+        self._xr_array_raw = None
+        self._entries_raw = None
+        self._entries_for_interp_raw = None
+        self._coords_raw = None
+
+    def initialise(self):
+        if self.entries_raw_initialized:
```
martinjrobins, line 81: Why do we need to set these variables that are not used? Be simpler just to return if already initialised?

src/pybamm/solvers/processed_variable.py
```diff
@@ -126,22 +493,67 @@ def initialise_0D(self):
...
+    ----------
+    base_variables : list of :class:`pybamm.Symbol`
+        A list of base variables with a method `evaluate(t,y)`, each entry of which
+        returns the value of that variable for that particular sub-solution.
+        A Solution can be comprised of sub-solutions which are the solutions of
+        different models.
+        Note that this can be any kind of node in the expression tree, not
+        just a :class:`pybamm.Variable`.
+        When evaluated, returns an array of size (m,n)
+    base_variable_casadis : list of :class:`casadi.Function`
```
kratman, line 532: Guessing it was supposed to be this 
```suggestion
    base_variables_casadi : list of :class:`casadi.Function`
```

src/pybamm/solvers/processed_variable.py
```diff
@@ -189,26 +601,71 @@ def initialise_1D(self, fixed_t=False):
...
```
kratman, line 629: Same as above
```suggestion
    base_variables_casadi : list of :class:`casadi.Function`
```

src/pybamm/solvers/processed_variable.py
```diff
@@ -291,38 +762,68 @@ def initialise_2D(self):
...
```
kratman, line 795:
```suggestion
    base_variables_casadi : list of :class:`casadi.Function`
```

src/pybamm/solvers/processed_variable.py
```diff
@@ -76,45 +72,416 @@ def __init__(
...
+        pybamm.logger.debug("Observing the variable raw data in C++")
+        ts, ys, _, funcs, inputs, is_f_contiguous = self._setup_cpp_inputs()
+        shapes = self._shape(self.t_pts)
+
+        return pybamm.solvers.idaklu_solver.idaklu.observe(
+            ts, ys, inputs, funcs, is_f_contiguous, shapes
+        )
+
+    def _observe_raw_python(self):
+        pass  # pragma: no cover
```
martinjrobins, line 174:
```suggestion
        raise NotImplementedError  # pragma: no cover
```

src/pybamm/solvers/processed_variable.py
```diff
@@ -76,45 +72,416 @@ def __init__(
...
+
+    def _observe_postfix(self, entries, t):
+        return entries
+
+    def _interp_setup(self, entries, t):
+        pass  # pragma: no cover
```
martinjrobins, line 180:
```suggestion
        raise NotImplementedError  # pragma: no cover
```

src/pybamm/solvers/processed_variable.py
martinjrobins: see https://docs.python.org/3/library/exceptions.html#NotImplementedError

src/pybamm/solvers/processed_variable.py
```diff
@@ -76,45 +72,416 @@ def __init__(
...
+
+    def _shape(self, t):
+        pass  # pragma: no cover
```
martinjrobins, line 183:
```suggestion
        raise NotImplementedError  # pragma: no cover
```

kratman (dismissed):
Pretty minor changes

src/pybamm/solvers/processed_variable.py
```diff
@@ -76,45 +72,416 @@ def __init__(
...
+        return processed_entries
+
+    def observe_raw(self):
+        """
+        Evaluate the base variable at the given time points and y values.
+        """
+        t = self.t_pts
+
+        # For small number of points, use Python
+        if pybamm.has_idaklu() and (t.size > 1):
```
martinjrobins, line 119: 1 is very small! So even for 10 or so points cpp is faster (I'm worried about the overhead of serialising and regenerating the casadi functions....)?

martinjrobins (changes requested):
great stuff, thanks @MarcBerliner! I like the refactor of processed variable into subclasses. A few minor points below

src/pybamm/solvers/c_solvers/idaklu/IDAKLUSolverOpenMP.inl
```diff
@@ -2,6 +2,9 @@
 #include "sundials_functions.hpp"
 #include <vector>
 
+// import the IDAMem class
```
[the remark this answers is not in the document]
MarcBerliner: Yeah this isn't needed

src/pybamm/solvers/processed_variable.py
MarcBerliner: Thanks, I forgot to remove this from a previous version of the function

src/pybamm/solvers/processed_variable.py
MarcBerliner: Thanks

src/pybamm/solvers/solution.py
```diff
@@ -588,8 +627,38 @@ def process_casadi_var(self, var_pybamm, inputs, ys_shape):
...
+            "variable",
+            [t_MX, y_MX, inputs_MX],
+            [var_sym],
+            opts,
+        )
+
+        # Some variables, like interpolants, cannot be expanded
+        try:
+            var_casadi_out = var_casadi.expand()
+        except RuntimeError:
```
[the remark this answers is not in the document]
MarcBerliner: This is just catching casadi's possible error. I believe they don't have a way of checking if the expansion will work ahead of time, otherwise this would be an if statement

src/pybamm/solvers/c_solvers/idaklu/IDAKLUSolverOpenMP.inl
```diff
@@ -409,23 +436,27 @@ SolutionData IDAKLUSolverOpenMP<ExprSet>::solve(
...
 
-  // Set the initial stop time
-  IDASetStopTime(ida_mem, t_eval_next);
+  // Reset the states at t = t_val. Sensitivities are handled in the while-loop
+  CheckErrors(IDAGetDky(ida_mem, t_val, 0, yy));
 
   // Solve the system
-  int retval;
   DEBUG("IDASolve");
   while (true) {
```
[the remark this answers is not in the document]
MarcBerliner: Yeah, I will revisit this section with the `CVODE` PR.

src/pybamm/solvers/solution.py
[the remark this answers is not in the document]
kratman: My concern was more that other runtime errors could be caught by this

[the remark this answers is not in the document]
MarcBerliner: Ah ok, I understand. Changed this to
```python
except RuntimeError as error:
    if "'eval_sx' not defined for" not in str(error):
        raise error  # pragma: no cover
    var_casadi_out = var_casadi
```

src/pybamm/solvers/processed_variable.py
```diff
@@ -76,45 +72,416 @@ def __init__(
...
+            entries = self._observe_raw_cpp()
+        else:
+            entries = self._observe_raw_python()
```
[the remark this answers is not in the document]
MarcBerliner: C++ used to be slower for one entry, but now they are the same. Removing the size check.

src/pybamm/solvers/processed_variable.py
MarcBerliner: C++ used to be slower for one entry, but now they are the same. Removing the size check.

To address the issue with serializing casadi functions, I added a pre-processing step in `_setup_cpp_inputs` to only serialize the subset of casadi functions that will actually be used

src/pybamm/solvers/c_solvers/idaklu/IDAKLUSolverOpenMP.inl
```diff
@@ -127,6 +130,26 @@ void IDAKLUSolverOpenMP<ExprSet>::InitializeStorage(int const N) {
           vector<realtype>(length_of_return_vector, 0.0)
       )
   );
+
+  if (save_hermite) {
```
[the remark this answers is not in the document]
MarcBerliner: I added an initialization for safety (see below). This is not always equal to `solver_opts.hermite_interpolation` because we need to satisfy additional checks that the full solution `y` is available and that we are saving all adaptive steps. If we don't, then we cannot guarantee the interpolation accuracy.

https://github.com/pybamm-team/PyBaMM/blob/a4aa9fe17dceadad6ac70e566663b824c3b78ddf/src/pybamm/solvers/c_solvers/idaklu/IDAKLUSolverOpenMP.inl#L90-L91

kratman (commented):
From my side it looks good, but lets wait for @martinjrobins  to approve of the changes

martinjrobins (approved):
happy with this, thanks @MarcBerliner

PR #4470, opened 2024-09-26
make SEI single layer only

### LINKED ISSUE #2333: Add single-layer SEI

valentinsulzer:
# Description

Remove the double-layer SEI and make it single layer only
This PR falls in the category of "breaking changes to the parameter names" so let's discuss how to handle this in the developer meeting - draft only for now

Fixes #2333
Replaces #3920

## Type of change

# Key checklist:

## Further checks:


brosaplanella:
We should probably merge this one before #4394 (tagging @kawaMANMI so he is aware of this PR).

src/pybamm/models/submodels/interface/sei/base_sei.py
```diff
@@ -268,37 +221,24 @@ def _get_standard_concentration_variables(self, variables):
...
 
         Returns
         -------
         variables : dict
             The variables which can be derived from the SEI currents.
         """
         domain, Domain = self.domain_Domain
-        j_inner_av = pybamm.x_average(j_inner)
-        j_outer_av = pybamm.x_average(j_outer)
-        j_sei = j_inner + j_outer
```
DrSOKane, line 240: Why is there no jsei_av variable?

src/pybamm/models/submodels/interface/sei/sei_growth.py
```diff
@@ -134,16 +123,16 @@ def get_coupled_variables(self, variables):
         elif SEI_option == "electron-migration limited":
             # Scott Marquis thesis (eq. 5.94)
             eta_inner = delta_phi - phase_param.U_inner
```
DrSOKane, line 125: I noticed `U_outer` doesn't get used anywhere. Deleting `U_inner` and `U_outer` and just having `U_sei` would reduce the total number of parameters, but I'm aware that `U_inner` and `U_sei` are for different mechanisms

DrSOKane (changes requested):
I agree with this change. As we were able to show in our joint paper, the effect of the two-layer structure on SEI growth rate is dwarfed by that of altering the parameters. The only exception I can think of is if there are two different mechanisms occuring simultaneously. [In-situ measurements](https://www.nature.com/articles/s41565-018-0284-y) of SEI growth and kinetic Monte-Carlo simulations both support such a scenario. I'm not aware of it being incorporated into a continuum-scale model yet, although [it has been mooted](https://www.sciencedirect.com/science/article/abs/pii/S0378775323000265?via%3Dihub). It would almost certainly require a separate class if it were ever to be implemented in PyBaMM.

I see no reason why `_get_standard_thickness_variables` and `_get_standard_total_thickness_variables` cannot be merged into a single function.

src/pybamm/models/submodels/interface/sei/sei_growth.py
brosaplanella: Yes, I spotted that too. I guess the main question is, if we switch to just `U_sei`, which parameter value should we use? I believe at the moment one is set at 0.1 V and the other at 0.8 V.

src/pybamm/models/submodels/interface/sei/base_sei.py
brosaplanella: It is defined later on, in lines 248-253

CHANGELOG.md
```diff
@@ -2,14 +2,16 @@
 
 ## Features
 
+- Added the `pybamm.DiscreteTimeSum` expression node to sum an expression over a sequence of data times, and accompanying `pybamm.DiscreteTimeData` class to store the data times and values ([#4501](https://github.com/pybamm-team/PyBaMM/pull/4501))
+- Double-layer SEI models have been removed (with the corresponding parameters). All models assume now a single SEI layer. ([#4470](https://github.com/pybamm-team/PyBaMM/pull/4470))
```
kratman, line 6: This entry is duplicated, since it is already in the breaking changes, can it be removed here

brosaplanella: Sorry, I thought it had to be in both!

src/pybamm/models/submodels/interface/sei/sei_growth.py
```diff
@@ -170,21 +159,15 @@ def get_coupled_variables(self, variables):
                 name = f"{Domain} EC surface concentration [mol.m-3]"
             variables.update({name: c_ec, f"X-averaged {name}": c_ec_av})
 
-        if SEI_option.startswith("ec reaction limited"):
-            inner_sei_proportion = 0
-        else:
-            inner_sei_proportion = phase_param.inner_sei_proportion
-
         # All SEI growth mechanisms assumed to have Arrhenius dependence
         Arrhenius = pybamm.exp(
```
kratman, line 163: Fixing a merge mistake
```suggestion
        arrhenius = pybamm.exp(
```

DrSOKane (changes requested):
Hi Ferran, thank you very much for considering my feedback. I have two more minor edits. However, a broader discussion around the three different SEI open-circuit potentials, and what they used to do before being unified, will help avoid any present and future misunderstandings.
* `U_sei_inner` is used by the `"electron-migration limited"` model, but no other. Substituting `U_sei_inner` with `U_sei` in line 124 of base_sei.py will significantly alter the results of this model, unless the user changes the value `U_sei` accordingly. However, according to @Scottmar93 thesis, this model is not widely used, so it's a sacrifice worth making. 
* `U_sei` is used by the `"reaction limited"` and `"ec reaction limited"` models. This PR does not change this.
* `U_sei_outer` was not used by any model.

There is therefore no basis for changing the value of `U_sei` in the `OKane2022` parameter set. Please revert this change.

Also, I tried running half_cell.ipynb with your proposed value of `U_sei = 0.1` and the results are massively different. Please revert this change as well.

I would apologize for the wall of text but I know you enjoy a good SEI discussion even more than I do!

brosaplanella:
Hi @DrSOKane! Thanks for the comments, I hadn't realised that `U_outer` was actually never used and I also didn't realised that your parameter sets were used with models that only use `U_sei`. I thought they used the inner and outer models, hence my confusion.

I have also noticed that diffusivity is defined to `2.5000000000000002e-22`, which probably resulted from copy pasting from some other code. I will chop them to 2 significant digits.

DrSOKane (approved):

kratman:
I made this a draft to keep it from being merged accidentally

kratman:
@valentinsulzer Removing the skip-release tag now

CHANGELOG.md
```diff
@@ -11,11 +15,10 @@
...
 - Added basic telemetry to record which functions are being run. See [Telemetry section in the User Guide](https://docs.pybamm.org/en/latest/source/user_guide/index.html#telemetry) for more information. ([#4441](https://github.com/pybamm-team/PyBaMM/pull/4441))
 - Added `BasicDFN` model for sodium-ion batteries ([#4451](https://github.com/pybamm-team/PyBaMM/pull/4451))
-- Added sensitivity calculation support for `pybamm.Simulation` and `pybamm.Experiment` ([#4415](https://github.com/pybamm-team/PyBaMM/pull/4415))
 - Added OpenMP parallelization to IDAKLU solver for lists of input parameters ([#4449](https://github.com/pybamm-team/PyBaMM/pull/4449))
 - Added phase-dependent particle options to LAM ([#4369](https://github.com/pybamm-team/PyBaMM/pull/4369))
 - Added a lithium ion equivalent circuit model with split open circuit voltages for each electrode (`SplitOCVR`). ([#4330](https://github.com/pybamm-team/PyBaMM/pull/4330))
-- Added the `pybamm.DiscreteTimeSum` expression node to sum an expression over a sequence of data times, and accompanying `pybamm.DiscreteTimeData` class to store the data times and values ([#4501](https://github.com/pybamm-team/PyBaMM/pull/4501))
```
kratman, line 18: I assume this should be put back in, it looks like it is in the release and I did not find a second version of it.

CHANGELOG.md
```diff
@@ -11,11 +15,10 @@
 - Added Hermite interpolation to the (`IDAKLUSolver`) that improves the accuracy and performance of post-processing variables. ([#4464](https://github.com/pybamm-team/PyBaMM/pull/4464))
 - Added basic telemetry to record which functions are being run. See [Telemetry section in the User Guide](https://docs.pybamm.org/en/latest/source/user_guide/index.html#telemetry) for more information. ([#4441](https://github.com/pybamm-team/PyBaMM/pull/4441))
 - Added `BasicDFN` model for sodium-ion batteries ([#4451](https://github.com/pybamm-team/PyBaMM/pull/4451))
-- Added sensitivity calculation support for `pybamm.Simulation` and `pybamm.Experiment` ([#4415](https://github.com/pybamm-team/PyBaMM/pull/4415))
```
kratman, line 14: I assume this should be put back in, it looks like it is in the release and I did not find a second version of it.

kratman (approved):

Brushnie:
@DrSOKane @brosaplanella @valentinsulzer The SEI predicted by 25.1 version when using "ec reaction limited" sei growth model is much less than 24.11.2 (and previous versions), please see the issue here: https://github.com/pybamm-team/PyBaMM/issues/4773

PR #4472, opened 2024-09-28
Removing all instances of unittest

prady0t:
WIP, will be adding ruff rule here.


prady0t:
I've added two rules, PT027 and PT009. These won't detect the usage of unittest, but is enough to prevent someone from writing tests in future as it prevents using unittest style assertions.

pyproject.toml
```diff
@@ -196,6 +196,8 @@ extend-select = [
     "YTT",           # flake8-2020
     "TID252",        # relative-imports
     "S101",          # to identify use of assert statement
+    "PT027",         # remove unittest style assertion
+    "PT009",         # Use pytest.raises instead of unittest-style
```
Saransh-cpp, line 200: Can you instead try turning on the "PT" rule commented above and see what happens?

Saransh-cpp (changes requested):
This looks amazing, thanks, @prady0t!

pyproject.toml
prady0t: There are a lot of style failures. Should I try fixing those and keep the `PT` rule?

agriyakhetarpal: If there are too many, then a separate PR for this would be better.

Saransh-cpp: A separate PR sounds good then. Could you create one after this, @prady0t?

Saransh-cpp (approved):

PR #4475, opened 2024-09-30
Pin casadi to 3.6.6

kratman:
# Description

I am seeing failures in some workflows due to casadi installs. Casadi just did a release this afternoon. I made this PR to pin casadi to a specific version while I investigate the issues.

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ requires = [
     "setuptools>=64",
     "wheel",
     # On Windows, use the CasADi vcpkg registry and CMake bundled from MSVC
-    "casadi>=3.6.6; platform_system!='Windows'",
+    "casadi==3.6.6; platform_system!='Windows'",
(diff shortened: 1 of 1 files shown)


agriyakhetarpal:
Maybe the problem was because the wheels were not available when the sdist was up? https://pypi.org/project/casadi/#files has wheels for 3.6.7 available, now.

kratman:
@agriyakhetarpal It also looks like the 39 wheel is missing from the release

agriyakhetarpal:
They are now available. I guess we should wait for a while and try to unpin, otherwise, okay to pin it. I'll update the vcpkg configuration later in the day for this.

kratman:
@agriyakhetarpal Pinning is often a good idea. The old pin was ">=3.6.6" and there was no higher version. So this literally changes nothing from what it was before

PR #4480, opened 2024-10-02
Adds doc strings for attributes in `base_model` and `base_submodel`

### LINKED ISSUE #4170: Fix attributes documentation in base_model and base_submodel
The doc strings cover the public attributes, but ignore the private ones.

Options:
- Remove all attributes from the doc strings
- Document attributes which are marked as properties

Related to #4158

medha-14:
# Description

Related #4170 

## Type of change

# Key checklist:

## Further checks:


medha-14:
I have added doc strings for a few private attributes. Am I going in the right direction with this or do I have to add the doc strings defining the attributes to the main doc string defining the class?

brosaplanella:
Hi! I think you are going in the right direction (but asked a review from @kratman as he opened the issue). Could you also add docstrings to the properties that return an empty dictionary. After defining what the method does, you can say that it returns an empty dictionary and that is defined in inheriting classes.

kratman:
@brosaplanella, @medha-14 The #4170 was a much larger ticket and probably should not be marked as "fixed" by this PR. It originally came from when I was fixing IDE inspections that said things like `_rhs` were not documented. Essentially the doc strings were out of date because the `_` was missing. The whole ticket is probably quite a lot more than this because there are plenty of classes where the arguments and attributes should be checked.

medha-14:
@kratman Could you confirm if the approach I took to address the issue was correct? If so, should I proceed by implementing similar changes in other classes as well? Additionally, could you please advise on which specific classes I should apply the required changes to?

kratman:
@medha-14 I was suggesting that you edit your PR description to remove the `fixes` part. Change it to "Related #4170" instead. Otherwise it will close the ticket when the ticket is probably larger than a single PR.

As far as the approach, I don't see any issues with adding more doc strings, but the original topic was about the class-wide doc string block. You want to turn on inspections in your IDE to make sure you also clean up any undocumented or incorrectly documented attributes there too.

The end goal should be making sure that all publicly accessible attributes appear in the documentation and none of the private ones are shown. Many of those are the properties, so this is a good start. It is a fairly open ended and visual task, you just have to go into the documentation and make sure everything is appearing.

You have some docs warnings to sort out as well:
```
/home/runner/work/PyBaMM/PyBaMM/src/pybamm/models/base_model.py:docstring of pybamm.BaseModel.boundary_conditions:1: WARNING: duplicate object description of pybamm.BaseModel.boundary_conditions, other instance in source/api/models/base_models/base_model, use :no-index: for one of them
/home/runner/work/PyBaMM/PyBaMM/src/pybamm/models/base_model.py:docstring of pybamm.BaseModel.name:1: WARNING: duplicate object description of pybamm.BaseModel.name, other instance in source/api/models/base_models/base_model, use :no-index: for one of them
/home/runner/work/PyBaMM/PyBaMM/src/pybamm/models/base_model.py:docstring of pybamm.BaseModel.variables:1: WARNING: duplicate object description of pybamm.BaseModel.variables, other instance in source/api/models/base_models/base_model, use :no-index: for one of them
```

medha-14:
Thanks for the review. I have added all public attributes in `base_model` and `base_submodel` to the respective class wide doc strings.

src/pybamm/models/base_model.py
```diff
@@ -19,29 +19,90 @@ class BaseModel:
...
+        A string representing the name of the model.
+    rhs: dict
+        A dictionary mapping expressions (variables) to expressions that represent
+        the right-hand side (RHS) of the model's differential equations.
+    algebraic: dict
+        A dictionary mapping expressions (variables) to expressions that represent
+        the algebraic equations of the model.
+    initial_conditions: dict
+        A dictionary mapping expressions (variables) to expressions that represent
+        the initial conditions for the state variables.
```
kratman, line 31: I think what you did here was revert some of my changes. These were removed because they are not public attributes:
```
        self._rhs = {}
        self._algebraic = {}
        self._initial_conditions = {}
        self._boundary_conditions = {}
        self._variables_by_submodel = {}
        self._variables = pybamm.FuzzyDict({})
        self._summary_variables = []
        self._events = []
        self._concatenated_rhs = None
        self._concatenated_algebraic = None
        self._concatenated_initial_conditions = None
        self._mass_matrix = None
        self._mass_matrix_inv = None
        self._jacobian = None
        self._jacobian_algebraic = None
        self._parameters = None
        self._input_parameters = None
        self._parameter_info = None
        self._variables_casadi = {}
        self._geometry = pybamm.Geometry({})
```
You can see here that these start with `_` and as such are private members. This is why I removed the docs for them in the first place.

kratman (changes requested):

kratman:
@medha-14 As a helpful guide, you should turn on inspections in your IDE then a lot of this shows up. PyCharm is what alerted me to the problem in the first place.

Basically we had docs for variables that used to be public (e.g. `rhs`), once they were made private like (e.g. `_rhs`), the docs were not updated so the variables do not even exist in the class. Ticket #4170 was about going through and making sure are the private attributes' documentation was removed, but then making sure the properties are properly documented.

I split that into a separate task from the original PR because I was not sure of the best way to handle it. It might be doc strings on the properties, it might be putting documentation in the main class's doc string block while properly marking them as properties. I am not sure what the correct fix is off the top of my head. You need to turn on IDE inspections so you can find stuff that is incorrect in the existing doc strings (most IDEs have plugins for doc warnings), and you need to view the generated documentation to make sure it all appears correctly. Sphinx probably has suggestions for how to document properties properly.

medha-14:
Thank you for your review and feedback. I've successfully resolved all warnings related to unresolved references in the documentation and removed all private attributes. I utilized PyCharm to address these warnings and ensured that properties are documented according to the Sphinx format. If there are any additional suggestions for improvements, I would be happy to work on them.

kratman:
Thanks @medha-14 I will go through another review later

kratman (approved):

medha-14:
Thank you for the prompt reviews and guidance @kratman.

PR #4487, opened 2024-10-03
Download IDAKLU from pybammsolvers

### LINKED ISSUE #3564: Tracking issue: migrate to `scikit-build-core`
Now that we have a declarative format for build metadata provided by `pyproject.toml` and build isolation enabling a more stable installation procedure via PEP 517/518, it eases the migration to either the [`scikit-build-core`](https://scikit-build-core.readthedocs.io/en/latest/index.html) or the [`meson-python`](https://meson-python.readthedocs.io/en/latest/) build-backend(s). Both of them include support for editable installations and are slowly becoming more mature for downstream users, considering that NumPy[^1] and SciPy[^2] have successfully transitioned this year.

The `scikit-build-core` build backend would still use CMake as a build system as we do currently, but `meson-python` requires [`ninja`](https://ninja-build.org/) as a build system, which would render the IDAKLU target compilation procedure radically different from the current one.

Some advantages of these new build-backends:

1. Better support for compilers plus ease of use[^3]
2. Reliable build caching for faster installation
3. Both of them are faster at compilation and more modern (multithreaded builds, cross-compilation, do not use `distutils` or `setuptools` at all)
4. Packagers for libraries in the Scientific Python ecosystem are slowly migrating and we should too
5. Might enable faster IDAKLU compilation with other compilers like MinGW? i.e., on Windows when building from source (which we do only when building the wheel)

[^1]: https://rgoswami.me/posts/numpy-meson-f2py/
[^2]: https://labs.quansight.org/blog/building-scipy-with-flang
[^3]: https://labs.quansight.org/blog/2021/07/moving-scipy-to-meson

## Task list

These tasks will be useful to look at, but not all of them might end up being needed or valid – subject to further discussion:

- [ ] Establish local SUNDIALS and SuiteSparse installations in a folder `sundials_installed/` or similar by adjusting `CMAKE_INSTALL_PREFIX` instead of modifying the `.local/` folder in the HOME directory` for a user
- [x] Specify `pybind11` as a build-time dependency in `pyproject.toml`
- [ ] Maybe build just IDAS instead of all SUNDIALS components, considering it is what we link to for the IDAKLU target?
- [x] On macOS, use a better way to find and use `libomp.dylib` that's set with the correct `MACOSX_DEPLOYMENT_TARGET` at least when deploying the wheels (not the Homebrew one). The reason for this is that Apple Clang does not come with the OpenMP specification – however, other compilers such as LLVM and GCC do. See also: #4091, #4092.
- [ ] On platforms we don't support, the `sdist` is used to install PyBaMM, and the IDAKLU solver which is optional cannot be installed unless one is installing from source or using `pip install pybamm --no-binary pybamm` to force installing from the sdist on PyPI. To provide a means to attempt IDAKLU compilation, a `--config-settings` command-line flag can be added to the `pip install pybamm` command that can be parsed by the build backend (`scikit-build-core` in this case). Therefore, users should have to manually opt-in to the compilation (the build option should be off by default). For this, `pybind11` should be a build-time dependency (see above).
- [ ] Look at ways to improve the user experience when installing SuiteSparse and SUNDIALS (which is by `nox -s pybamm-requires` on macOS and Linux, possibly remove this command) and add docs on how to do this on Windows
- [x] Add `CMakeLists.txt` (#4101) and other relevant files such as `FindSuiteSparse.cmake` and `FindSUNDIALS.cmake` to the SDist via the `MANIFEST.in` file, otherwise, CMake will raise an error about a missing project status. This needs to be done as a part of the migration and a job could be added that tests building a wheel from the sdist nightly (and/or we can use the scheduled tests).

### LINKED ISSUE #3603: Revisit linkage with Python `casadi` for Windows wheels and move away from `pybamm-team/casadi-vcpkg-registry`
Please read through #3100 and #3193 for background discussion. This proposition is about getting back to linking against the Python-based CasADi distribution from PyPI when building the IDAKLU extension for the Windows wheels. We currently use a custom `vcpkg` registry at https://github.com/pybamm-team/casadi-vcpkg-registry/ for this reason, but further releases mean that the versions for `casadi` at build-time and that at run-time can sometimes run out of sync with each other due to lack of periodic updates to the repository, and might cause issues with the import of the IDAKLUSolver after importing PyBaMM such as the one introduced in the PyPI wheels for certain users where a serialisation error was introduced (#3193).

This would serve two purposes:

1. Provide a more maintainable solution for building the Windows wheels, since the only `vcpkg` registry we would need to interface would be the one for SUNDIALS – which is needed because we require it to be built and distributed with KLU
2. Use the same version for build-time and run-time dependencies which ensures a stable binary.

<hr>

I have tried building the Windows wheels and used `casadi==3.6.4` from PyPI for the linkage in a [workflow run](https://github.com/agriyakhetarpal/PyBaMM/actions/runs/7137512881/job/19437616278) for my fork, but it resulted in an error where the `libcasadi.lib` file was requested but just `casadi.lib` exists in the `casadi/` folder, this might be a mismatch from their end or some error in our configuration (see https://github.com/pybamm-team/PyBaMM/issues/3100#issuecomment-1845352546)

### LINKED ISSUE #4611: Mark JAX IDAKLU solver as experimental on Windows
@martinjrobins @kratman can I leave you to fill in details, please?

kratman:
# Description

This will separate the IDAKLU C++ code from pybamm.

Fixes #3564
Fixes #4611

This ticket is something that can be handled in the other repo
Closes #3603 

## Type of change

This should speed up CI by skipping the build of the C++ code.

# Key checklist:

## Further checks:


kratman:
A new link error cropped up, but it looks like we could get a lot of savings on time with this update

agriyakhetarpal:
The linkage error is the same one as #3783, coming from CasADi's plugin system. I am not sure if it's worth fixing it, since it was fixed by @martinjrobins for the linear interpolant case by dropping down to Python but IIRC there wasn't a way in CasADi for doing it for the cubic

kratman:
@agriyakhetarpal Yeah I was looking at that issue as well. As far as I can tell CasADI sets a path for plugins. I am trying to see if there is a decent workaround since this was part of #4464

My guess is that the wheels for the next release will be broken as well, but I have not confirmed it yet

agriyakhetarpal:
There is a workaround for Linux and macOS, but not for Windows (different toolchain); sadly, it's not decent enough to include. I think I'll raise a PR upstream in CasADi to get one part of the linkage going and see if we can migrate to a non-MSVC toolchain (which can potentially help fix this later on). It's been on my list of things to do for a while, but I've yet to do it.

kratman:
This is fixed locally with this: `export CASADIPATH=.venv/lib/python3.12/site-packages/casadi`

agriyakhetarpal:
> This is fixed locally with this: `export CASADIPATH=.venv/lib/python3.12/site-packages/casadi`
> 

Yes, won't work with Windows

martinjrobins:
I had a look at this. The linker error is the same as I came across for the case of linear interpolation. The solution there was to swap to using the direct casadi function rather than their plugin system, which won't work if the casadi function is evaluated in C++ for windows as we compile everything statically. 

I think I might be able to access the direct bspline interface by calculating the spline coefficients in scipy and then use the casadi `Function.bspline` function to construct a bspline. Cross fingers this doesn't use the plugin system anywhere!

kratman:
> I had a look at this. The linker error is the same as I came across for the case of linear interpolation. The solution there was to swap to using the direct casadi function rather than their plugin system, which won't work if the casadi function is evaluated in C++ for windows as we compile everything statically.
> 
> I think I might be able to access the direct bspline interface by calculating the spline coefficients in scipy and then use the casadi `Function.bspline` function to construct a bspline. Cross fingers this doesn't use the plugin system anywhere!

Yeah I was going to approach this by seeing if I could just change the build itself. It is something that should work if we are compiling and delivering everything correctly. If that does not work, then I will look at workarounds for interpolation

kratman:
I expect to work on this again next week, I have been caught up with other stuff

martinjrobins:
looks like there is still issues with the idaklu jax solver on windows, I can look into these?

martinjrobins:
I tried to figure this one out today but no luck :( It's crashing with a fatal exception when jax tries to jit compile, I'm still in the dark as to why. It might be a threading issue as the problem is intermittant (occurs in about 95% of test runs). It might be cause with some interaction with pytest because when I copy the test into a stand-alone script it works fine

agriyakhetarpal:
For a stopgap solution, we can isolate these tests into their own xdist_group and allow only one worker to touch them at a time.

kratman:
> For a stopgap solution, we can isolate these tests into their own xdist_group and allow only one worker to touch them at a time.

Yeah that is my fallback option.

I want to take a closer look at the linking/delivery as well. We have failures on windows when you download the wheels:
- pybammsolvers has some crashes workers
- my i5 (without AVX-512 instructions) has ~45 test failures on both 24.9.0 and 24.11.0 when running tests with the wheels
- A colleague's i7 (with AVX-512 instructions) has ~25 test failures on both 24.9.0 and 24.11.0 when running tests with the wheels

So it appears that the tests are working when you test in the build environment, but not in a different environment. I will be digging into this more and see what I come up with

jsbrittain:
Hi - I took a very quick look at this yesterday and agree that it seems to be a threading issue. More specifically, `jaxify()` can only be called once per solver instance (this is the first test), which then caches the full solve result so the jax-wrapper can query samples without repeatedly re-running the solver. My suspicion is that running these tests in parallel is causing test pollution, probably because the test script currently instantiates the solver and jax wrapper objects at the start of the test script, not as a fixture for each test (although the ubuntu tests should also be failing?). Refactoring the tests with fixtures would be good to see if that resolves things - I can take a look at that if you like but if I'm right then the `xdist_group` solution should also work if you need a quick fix. I can't remember the precise details as to why we can't jaxify more than once per object, but I do remember that it was more complex than just the cache issue (something to do with the jax primitives...).

kratman:
Testing out the skips and test refactor now with CI. I have some docs stuff to update then this should be mostly ready to go. I will do one more update to pybammsolvers to make sure that the versions of IDAKLU match

kratman:
Tests pass, just need to do some documentation fixes

kratman:
Additional documentation will be added to the pybammsolvers repo

kratman:
@MarcBerliner, @martinjrobins Ok I think this is finally working. I am working on tests and docs for the other repo now

docs/source/user_guide/installation/gnu-linux-mac.rst
```diff
@@ -76,10 +76,6 @@ library beforehand.
 
 .. tab:: conda
 
-   .. note::
-
-      At the moment, PyBaMM's conda-forge distribution does not include ``IDAKLUSolver``.
-
```
Saransh-cpp, line 82: I think `pybammsolvers` is not yet available on conda-forge? This can be updated with something better near the "required dependencies" section.

docs/source/user_guide/installation/index.rst
```diff
@@ -19,10 +19,6 @@ It can be installed using ``pip`` or ``conda``, or from source.
 
    The ``pybamm`` package on conda-forge installs PyBaMM with all the `required and optional dependencies <#dependencies>`_ available on conda-forge.
```
Saransh-cpp, line 20: The required dependencies section should be updated to include `pybammsolvers`.

Saransh-cpp (commented):
This is fantastic. Took a quick view and it looks good overall!

docs/source/user_guide/installation/gnu-linux-mac.rst
agriyakhetarpal: cc @prady0t: we will now like to get the IDAKLU solver packaged and ready, since the v25.1 release is fast approaching. So, while it is okay to wait for a while (since PyBaMM's conda-forge package is downloaded much less than the PyPI wheels, and feedstocks sometimes do tend to lag behind releases), it would be great if your work could be expedited so that both distributions are at par soon after the release.

Please feel free to [start a new feedstock](https://conda-forge.org/docs/maintainer/understanding_conda_forge/feedstocks/#initialization) for `pybammsolvers` and ping Saransh and/or me where you have any questions with the recipe!

prady0t: Just to be clear, `pybammsolvers` is all the C++ part related to IDAKLU solver. For packaging pybamm with IDAKLU, we can specify one of the dependencies as `pybamsolvers` in pybamm feedstock and later continue with building `suitsparse` and `sundials`(Which I'm already been trying for a few days)

And for that, we need to package `pybammsolvers` as a conda-forge recipe, and hence a new feedstock is needed.

Am I correct?

kratman: @prady0t pybammsolvers should have all of the C++ code that is normally distributed with PyBaMM, so suitesparse and sundials are being built there. It should be the only C++ requirement for PyBaMM after this PR is done

What are the tickets for the suitespares and sundials work? I am probably going to need to do something there, and want to see if work is being duplicated

docs/source/user_guide/installation/index.rst
kratman: Ok I will fix this

docs/source/user_guide/installation/gnu-linux-mac.rst
prady0t: I was trying to build `suitsparse` and `sundials` as extension modules in pybamm feedstock. https://github.com/conda-forge/pybamm-feedstock/issues/65

docs/source/user_guide/installation/index.rst
kratman: Done. Maintaining that list seems like a bit of wasted effort. We should probably just link people to our pyproject.toml file or something. Nobody needs to manually install that list of dependencies, they are just installed by pip or conda

docs/source/user_guide/installation/gnu-linux-mac.rst
kratman: @Saransh-cpp I assumed that pybammsolvers would be uploaded to conda before the release, so I hope we don't need a note about it. A goal is to also switch IDAKLU to be the default solver (I have a branch locally, but have not pushed yet)

Saransh-cpp: That makes sense, thanks for the clarification!

agriyakhetarpal: > I assumed that pybammsolvers would be uploaded to conda before the release

This means we have around ~2 weeks, echoing Eric's message from Slack (if we need PyBaMM's v25.1 `conda-forge` release to be on time and as close as possible to the PyPI release, that is).

.github/workflows/benchmark_on_push.yml
```diff
@@ -31,12 +31,9 @@ jobs:
 
       - name: Install python dependencies
         run: |
-          python -m uv pip install --upgrade pip wheel setuptools wget cmake casadi numpy
+          python -m uv pip install --upgrade pip wheel setuptools wget numpy
```
agriyakhetarpal, line 34:
```suggestion
```

I don't think even this should be needed now, `asv` sets up and manages environments automatically from its config file.

.github/workflows/periodic_benchmarks.yml
```diff
@@ -39,19 +39,13 @@ jobs:
 
       - name: Install python dependencies
         run: |
-          python -m uv pip install --upgrade pip wheel setuptools wget cmake casadi numpy
+          python -m uv pip install --upgrade pip wheel setuptools wget numpy
```
agriyakhetarpal, line 42:
```suggestion
```

.github/workflows/publish_pypi.yml
```diff
@@ -104,24 +100,17 @@ jobs:
...
         run: pipx run cibuildwheel --output-dir wheelhouse
         env:
           CIBW_ENVIRONMENT: >
             PYBAMM_DISABLE_TELEMETRY="true"
           CIBW_ARCHS_LINUX: x86_64
-          CIBW_BEFORE_ALL_LINUX: >
-            yum -y install openblas-devel lapack-devel &&
-            bash scripts/install_sundials.sh 6.0.3 6.5.0
-          CIBW_BEFORE_BUILD_LINUX: python -m pip install cmake casadi setuptools wheel
+          CIBW_BEFORE_BUILD_LINUX: python -m pip install setuptools wheel
```
agriyakhetarpal, line 109:
```suggestion
          CIBW_BEFORE_BUILD_LINUX: python -m pip install setuptools
```

From #4652

.github/workflows/publish_pypi.yml
```diff
@@ -155,95 +141,17 @@ jobs:
...
-            python scripts/install_KLU_Sundials.py
             python -m cibuildwheel --output-dir wheelhouse
         env:
           CIBW_ENVIRONMENT: >
             PYBAMM_DISABLE_TELEMETRY="true"
           # 10.13 for Intel (macos-13), 11.0 for Apple Silicon (macos-14 and macos-latest)
           MACOSX_DEPLOYMENT_TARGET: ${{ matrix.os == 'macos-14' && '11.0' || '10.13' }}
           CIBW_ARCHS_MACOS: auto
-          CIBW_BEFORE_BUILD: python -m pip install cmake casadi setuptools delocate
+          CIBW_BEFORE_BUILD: python -m pip install setuptools wheel delocate
```
agriyakhetarpal, line 154:
```suggestion
          CIBW_BEFORE_BUILD: python -m pip install setuptools delocate
```

CHANGELOG.md
```diff
@@ -17,6 +17,9 @@
 package to install PyBaMM with only the required dependencies. ([conda-forge/pybamm-feedstock#70](https://github.com/conda-forge/pybamm-feedstock/pull/70))
 - Separated extrapolation options for `pybamm.BoundaryValue` and `pybamm.BoundaryGradient`, and updated the default to be "linear" for the value and "quadratic" for the gradient. ([#4614](https://github.com/pybamm-team/PyBaMM/pull/4614))
 - Double-layer SEI models have been removed (with the corresponding parameters). All models assume now a single SEI layer. ([#4470](https://github.com/pybamm-team/PyBaMM/pull/4470))
+- Moved the IDAKLU solver to a standalone `pybammsolvers` package. This will
+  make pybamm a pure python package and make installing and using the solver
+  easier. ([#4487](https://github.com/pybamm-team/PyBaMM/pull/4487))
```
agriyakhetarpal, line 22:
```suggestion
- Moved the IDAKLU solver to a standalone `pybammsolvers` package. This will
  make PyBaMM a pure Python package and make installing and using the solver
  easier. ([#4487](https://github.com/pybamm-team/PyBaMM/pull/4487))
```

(nit)

benchmarks/different_model_options.py
```diff
@@ -37,17 +37,6 @@ class SolveModel:
...
-
-        idaklu_spec = importlib.util.find_spec("pybamm.solvers.idaklu")
-        if idaklu_spec is not None:
-            try:
-                idaklu = importlib.util.module_from_spec(idaklu_spec)
-                idaklu_spec.loader.exec_module(idaklu)
-            except ImportError as e:  # pragma: no cover
-                print("XXXXX cannot find klu", e)
-                idaklu_spec = None
-
```
agriyakhetarpal, line 50: This will make the IDAKLU solver non-optional, is this also intended here? This would cause non-supported platforms such as Linux arm64 (which we provide Docker images for) try to compile `pybammsolvers` from the sdist, which is going to be broken until https://github.com/pybamm-team/pybammsolvers/pull/7 is merged.

 I think it's better to split both tasks (separating the solver, and making the solver the default) in two separate PRs.

pyproject.toml
```diff
@@ -1,12 +1,7 @@
...
 requires = [
-    "setuptools>=70.1.1",
-    # On Windows, use the CasADi vcpkg registry and CMake bundled from MSVC
-    "casadi>=3.6.7; platform_system!='Windows'",
-    # Note: the version of CasADi as a build-time dependency should be matched
-    # across platforms, so updates to its minimum version here should be accompanied
-    # by a version bump in https://github.com/pybamm-team/casadi-vcpkg-registry.
-    "cmake; platform_system!='Windows'",
+    "setuptools",
+    "wheel",
```
agriyakhetarpal, line 4:
```suggestion
```

Now dropped with #4652

pyproject.toml
```diff
@@ -34,6 +29,7 @@ classifiers = [
     "Topic :: Scientific/Engineering",
 ]
 dependencies = [
+    "pybammsolvers",
```
agriyakhetarpal, line 32: This makes it a non-optional dependency, and is a breaking change as well – please see my comment above.

scripts/Dockerfile
```diff
@@ -28,11 +28,7 @@ RUN uv venv $VIRTUAL_ENV
 RUN #!/bin/bash && source /home/pybamm/venv/bin/activate;
 ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 
-RUN uv pip install --upgrade setuptools wget cmake
-
-RUN python scripts/install_KLU_Sundials.py && \
-rm -rf pybind11 && \
-git clone https://github.com/pybind/pybind11.git && \
-uv pip install -e ".[all,dev,docs,jax]";
+RUN uv pip install --upgrade setuptools wget
```
agriyakhetarpal, line 31:
```suggestion
```

src/pybamm/solvers/idaklu_solver.py
```diff
@@ -23,20 +19,18 @@
...
-        print(f"Error loading idaklu: {e}")
-        idaklu_spec = None
+try:
+    import pybammsolvers.idaklu as idaklu
+
+    idaklu_imported = True
+except ImportError as e:  # pragma: no cover
+    idaklu = None
+    print(f"Error loading idaklu: {e}")
+    idaklu_imported = False
```
agriyakhetarpal, line 29: Same as above

src/pybamm/solvers/idaklu_jax.py
```diff
@@ -8,19 +8,14 @@
...
-        if idaklu_spec.loader:
-            idaklu_spec.loader.exec_module(idaklu)
-    except ImportError:  # pragma: no cover
-        idaklu_spec = None
+try:
+    import pybammsolvers.idaklu as idaklu
+except ImportError as e:  # pragma: no cover
+    idaklu = None
+    print(f"Error loading idaklu: {e}")
+
```
agriyakhetarpal, line 18: The IDAKLU solver is optional here, but `pybammsolvers` also needs to be an optional dependency for this to work correctly and be guarded under `pybamm[idaklu]` or a similar name (maybe `[solvers]` because the plan is to add CVODES too?)

agriyakhetarpal (changes requested):

benchmarks/different_model_options.py
kratman: Yeah changing the default solver is intended to be a separate task.

I thought the linux Arm64 stuff was already fixed. I go back to working on that

src/pybamm/solvers/idaklu_jax.py
kratman: I had this in here for testing and error handling. I think there is probably other cleanup that can be done as well.

benchmarks/different_model_options.py
agriyakhetarpal: Sure. The last time I worked on Linux arm64, cross-compilation wasn't helpful with SuiteSparse's symbols turning up missing, and CasADi faced https://github.com/dockcross/dockcross/issues/780 when I tried compiling under QEMU. But maybe the situation has improved by a lot now...

src/pybamm/solvers/idaklu_solver.py
kratman: Removed

pyproject.toml
kratman: The intention is to make it required, but not default yet

benchmarks/different_model_options.py
kratman: I am working on getting the scikit build up and running again while also looking at the conda-forge recipe. Hopefully those will resolve these issues

agriyakhetarpal (approved):
Yes, let's do that, @kratman. Linux arm64 is quite common but lacks readily available CI providers for OSS; Windows arm64 CI is not present either outside of Windows Arm dev boxes. FWIW, we can mark both platforms/architectures as unsupported for now and ask users to compile it themselves (the same as we did when macOS arm64 was not common).

Based on https://stackoverflow.com/a/64921347, we will have to wait a bit for PyBaMM 25.1 to be supported on conda-forge until https://github.com/conda-forge/staged-recipes/pull/28748 is resolved, but at the time if someone needs a new version of PyBaMM in a conda environment, they should be able to `pip`-install without any issues as long as their environment files specify that.

kratman:
Note: The changes from #4736 are not in pybammsolvers yet. The tests pass without the C++ side for now though. I am working on the pybammsolvers v0.0.5 release, but I have to do a bit of testing before it is ready. Hopefully I will finish that off today

PR #3752, opened 2024-01-21
[Bug]: `pybamm.Simulation.set_parameters`, `pybamm.Simulation.set_up_and_parameterise_experiment` and `pybamm.Simulation.set_up_and_parameterise_model_for_experiment` made private in `simulation.py`

Akhil-Sharma30:
# Description

`pybamm.Simulation.set_parameters`, `pybamm.Simulation.set_up_and_parameterise_experiment` and `pybamm.Simulation.set_up_and_parameterise_model_for_experiment` made private in `pybamm/simulation.py`

Fixes #3751 

## Type of change

Please add a line in the relevant section of [CHANGELOG.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CHANGELOG.md) to document the change (include PR #) - note reverse order of PR #s. If necessary, also add to the list of breaking changes.

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (back-end change that speeds up the code)
- [ ] Bug fix (non-breaking change which fixes an issue)

# Key checklist:

- [ ] No style issues: `$ pre-commit run` (or `$ nox -s pre-commit`) (see [CONTRIBUTING.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CONTRIBUTING.md#installing-and-using-pre-commit) for how to set this up to run automatically when committing locally, in just two lines of code)
- [x] All tests pass: `$ python run-tests.py --all` (or `$ nox -s tests`)
- [ ] The documentation builds: `$ python run-tests.py --doctest` (or `$ nox -s doctests`)

You can run integration tests, unit tests, and doctests together at once, using `$ python run-tests.py --quick` (or `$ nox -s quick`).

## Further checks:

- [ ] Code is commented, particularly in hard-to-understand areas
- [ ] Tests added that prove fix is effective or that feature works

PR #4489, opened 2024-10-04
[Bug]: `pybamm.Simulation.set_parameters` and `pybamm.Simulation.set_up_and_parameterise_experiment` made private in `simulation.py`

### LINKED ISSUE #3751: [Bug]: `pybamm.Simulation.set_parameters`, `pybamm.Simulation.set_up_and_parameterise_experiment` and `pybamm.Simulation.set_up_and_parameterise_model_for_experiment` should be private (i.e. begin with underscore _) and not documented
### PyBaMM Version

latest

### Python Version

python3.11

### Describe the bug

Took #3392, This is a part of `API Docs` in this issue.

### Steps to Reproduce

Not applicable

### Relevant log output

_No response_

Akhil-Sharma30:
# Description

`pybamm.Simulation.set_parameters`, `pybamm.Simulation.set_up_and_parameterise_experiment` and `pybamm.Simulation.set_up_and_parameterise_model_for_experiment` made private in `pybamm/simulation.py` 
(continuation of #3752 PR)

Fixes #3751 

## Type of change

# Key checklist:

## Further checks:


src/pybamm/simulation.py
```diff
@@ -182,7 +182,8 @@ def set_up_and_parameterise_experiment(self, solve_kwargs=None):
         reduces simulation time since the model formulation is efficient.
         """
         parameter_values = self._parameter_values.copy()
-
+        msg = "pybamm.simulation.set_up_and_parameterise_experiment is deprecated and not meant to be accessed by users."
+        warnings(msg, DeprecationWarning)
```
kratman, line 186: Two problems here:
1. You need to make it warnings.warn
2. You need to explicitly set stacklevel

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # [Unreleased](https://github.com/pybamm-team/PyBaMM/)
 
+- Deprecated `pybamm.Simulation.set_parameters` and `pybamm.Simulation.set_up_and_parameterise_experiment` functions in `pybamm.simulation.py`. ([#3752](https://github.com/pybamm-team/PyBaMM/pull/3752))
+
```
kratman, line 4: This should probably be in breaking changes since it is intended as a deprecation.

src/pybamm/simulation.py
```diff
@@ -256,7 +257,10 @@ def set_parameters(self):
         """
         A method to set the parameters in the model and the associated geometry.
         """
-
+        msg = (
+            "pybamm.set_paramters is deprecated and not meant to be accessed by users."
+        )
+        warnings(msg, DeprecationWarning)
```
kratman, line 263: Same as my other comment

kratman (changes requested):
I think the other question for this is what we do about the function names and tests.

### Improvements:
I think that if we intend to deprecate the public functions we should:
1. First rename all the usages to `_set_parameters()` and `_set_up_and_parameterise_experiment()`
2. Create new functions called `set_parameters()` and `setup_and_parametrise_experiments()`
3. The new functions should be more like this
    ```
    def set_parameters(...):
        msg = "pybamm.simulation.set_parameters() is deprecated"
        warnings,warn(msg, DeprecationWarning, stacklevel=2)
        self._set_parameters(...)
    ```
 
 The renaming is done first so you make sure to catch all usages in the codebase before adding the new one.
 
 ### Problems:
 - The tests use these functions, so this means that we will be accessing private members outside the class.
 - In the future I would like to turn on inspections to prevent accessing private methods. This becomes a problem if the tests need to access these methods.
 - We should find a solution to having the tests access these members.

src/pybamm/simulation.py
Akhil-Sharma30: Like this?
image

agriyakhetarpal: Yes

kratman (dismissed):

kratman (approved):

kratman:
@Akhil-Sharma30 I just fixed a couple small things, should be good for now

Issue #4406, opened 2024-09-03
Add `pybamm-cookie` guides and references in the documentation


santacodes:
As discussed in the `2024-09-02 Developer Meeting` moving forward the template would be the way to add third-party parameter sets and models as it supports its entry points. This would require references in the documentation and guides on how to use the template and map entry points. In addition to that the documentation would also contain steps on how to package it to make it more accessible for the community to fetch it and run it locally out of the box. 

P.S. I am happy to take up this issue so feel free to assign me if someone is not already working on it after discussing it in the meeting.

PR #4490, opened 2024-10-04
Unified entry point for models and parameter sets

santacodes:
# Description

Modified the existing `parameter_sets` API and unified entry points to accommodate `model entry points`. This is an extension of the existing model entry points in the `pybamm-cookie` project. [Relevant discussion](https://github.com/pybamm-team/pybamm-cookie/issues/1#issuecomment-1959971400).

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:


src/pybamm/entry_points.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+#: Singleton Instance of :class:EntryPoint initialised with pybamm_parameter_sets"""
+parameter_sets = EntryPoint(group="pybamm_parameter_sets")
+
+#: Singleton Instance of :class:EntryPoint initialised with pybamm_models"""
+model_sets = EntryPoint(group="pybamm_models")
+
+
+def Model(model: str):  # doctest: +SKIP
```
arjxn-py, line 123: Why skipping doctests?

arjxn-py, line 123: Would `Model()` as a method be sufficient? My idea is that a class would add more utility and you can have similar required methods as parameter sets i.e. get_docstring or other model operations.

arjxn-py (commented):
Good start @santacodes, looks like this is still a WIP? Still I have commented just a tip hoping it'd make sense.

There are some conflicts that you may want to resolve and also feel free to add [WIP] prefix to your PR title while you're on it.

src/pybamm/entry_points.py
santacodes: All of the parameter_sets function also works for models. The Model() method is just an addition to that. As I mentioned this is a unified entry point for both models and parameter-sets. They operate on the same API, just different instances for models and parameter-sets.

santacodes: The model function returns an object of an initialised model along with the address of the object. This might vary per instance in a personal computer so the doctests couldn't assert the example I have mentioned in docstring. For example something like  <pybamm.models.spm.SPM object 0x7381> is returned when the Model() function is called, the address might be different across various systems which would cause the doctests to fail. In brief I think doctesting and asserting two objects of a class would be a bad idea, unless there's a workaround for it. Do let me know if there is, I'm happy to implement it! :)

src/pybamm/entry_points.py
```diff
@@ -0,0 +1,144 @@
...
+            pass  # Handle the attribute error normally
+
+        return super().__getattribute__(name)
+
+
+#: Singleton Instance of :class:EntryPoint initialised with pybamm_parameter_sets"""
+parameter_sets = EntryPoint(group="pybamm_parameter_sets")
+
+#: Singleton Instance of :class:EntryPoint initialised with pybamm_models"""
+model_sets = EntryPoint(group="pybamm_models")
```
valentinsulzer, line 120: this should just be called `models`, a `parameter_set` is a set of parameters (and a model is a set of equations), there is no set of models

santacodes: Yes that was intentional, `models` wasn't working because the `pybamm.models` namespace is already taken by the collection of battery models which gets overridden first. We could either change the variable name which was my temporary fix (`model_sets`) to something more sensible or we could change the namespace for the entry point models to something like `pybamm.entrypoint.models`. I'd prefer the first option though if you could come up with a sensible variable name for the models instance because I have a limited knowledge with the battery library.

valentinsulzer: Understood! In order to keep it consistent with parameter sets let's keep it as is then

santacodes: Sorry for the late response. I have abstracted the `models` instance inside the `dispatch` namespace as @agriyakhetarpal mentioned. I did not move the `parameter_sets` instance inside the dispatch namespace, as it would be a breaking change and also fail a lot of tests because `pybamm.parameter_sets` wouldn't work instead, people would have to use `pybamm.dispatch.parameter_sets`, though both of these work as of now. The `models` instance (entry point) can now be accessed through `pybamm.dispatch.models` and `list(pybamm.dispatch.models)` would list all the battery models available through the entry point. To load the model, `pybamm.Model("SPM")` would return the model object.

santacodes:
> @santacodes Are you still working on this?

Yes I am, it has yet to be reviewed, I will update this branch and request a review again.

santacodes:
Resolving merge conflicts on my branch was causing a lot of problems and it was staging a lot of files that weren't modified by me during the merge for some reason, anyway it was a small change so I manually wrote a patch file and applied the patch after resetting this branch so the history is gone. This PR still needs a review to get it merged, and this PR is directly related to - https://github.com/pybamm-team/pybamm.org/pull/258 so it might be good if we have a base entry points API before the GSoC timeline.

santacodes:
Bumping this @valentinsulzer @agriyakhetarpal @arjxn-py as I cannot request a re-review

valentinsulzer (changes requested):
This implementation looks good to me! My only comment is that the user should be able to pass args and kwargs to the model i.e.
```python
pybamm.Model("SPM", options)
```
returns
```python
pybamm.lithium_ion.SPM(options)
```
But let's keep it as a general `*args, **kwargs` so that developers can enable other arguments beyond just `options` if they want

santacodes:
> This implementation looks good to me! My only comment is that the user should be able to pass args and kwargs to the model i.e.
> 
> ```python
> pybamm.Model("SPM", options)
> ```
> 
> returns
> 
> ```python
> pybamm.lithium_ion.SPM(options)
> ```
> 
> But let's keep it as a general `*args, **kwargs` so that developers can enable other arguments beyond just `options` if they want

Added this in the latest commit, works too! :) 
pybammmep

pyproject.toml
```diff
@@ -148,6 +148,9 @@ ECM_Example = "pybamm.input.parameters.ecm.example_set:get_parameter_values"
 MSMR_Example = "pybamm.input.parameters.lithium_ion.MSMR_example_set:get_parameter_values"
 Chayambuka2022 = "pybamm.input.parameters.sodium_ion.Chayambuka2022:get_parameter_values"
 
+[project.entry-points."pybamm_models"]
+SPM = "pybamm.models.full_battery_models.lithium_ion.spm:SPM"
+
```
agriyakhetarpal: Should we also add the DFN model, or are you looking to expose just one model for now?

src/pybamm/dispatch/entry_points.py
```diff
@@ -0,0 +1,152 @@
...
+import textwrap
+import warnings
+from collections.abc import Mapping
+from typing import Callable
+
+
+class EntryPoint(Mapping):
+    """
+    Access via :py:data:`pybamm.parameter_sets` for parameter_sets
+    Access via :py:data:`pybamm.Model` for Models
```
agriyakhetarpal: Could you please expand this docstring to be a bit more descriptive?

src/pybamm/dispatch/entry_points.py
```diff
@@ -0,0 +1,152 @@
...
+
+        >>> print(pybamm.dispatch.models.get_docstring("SPM"))  # doctest: +ELLIPSIS
+        <BLANKLINE>
+        Single Particle Model (SPM) of a lithium-ion battery...
+    """
+
+    _instances = 0
+
+    def __init__(self, group):
+        """Dict of entry points for parameter sets or models, lazily load entry points as"""
```
agriyakhetarpal: Is this docstring incomplete, or is there a grammatical error? i.e.,:

```suggestion
        """Dict of entry points for parameter sets or models to lazily load as entry points"""
```

or did you mean something else?

src/pybamm/dispatch/entry_points.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def __getitem__(self, key) -> dict:
+        return self._load_entry_point(key)()
+
+    def _load_entry_point(self, key) -> Callable:
+        """Check that ``key`` is a registered ``parameter_sets`` or ``models` ,
+        and return the entry point for the parameter set/model, loading it needed."""
+        if key not in self._all_entries:
+            raise KeyError(f"Unknown parameter set or model: {key}")
+        ps = self._all_entries[key]
```
agriyakhetarpal: (nit): `ps` doesn't seem like a readable variable name, maybe `parameter_set` or `param_set` would be nicer?

src/pybamm/dispatch/entry_points.py
```diff
@@ -0,0 +1,152 @@
...
+        Model object of the initialised model.
+    Examples
+    --------
+    Listing available models:
+        >>> import pybamm
+        >>> list(pybamm.dispatch.models)
+        ['SPM']
+        >>> pybamm.Model('SPM') # doctest: +SKIP
+        <pybamm.models.full_battery_models.lithium_ion.spm.SPM object>
+    """
```
agriyakhetarpal: Could you document the issue with the doctests here, either as an inline comment or on GitHub in our reviews? I think it would be nice to try fixing it.

src/pybamm/dispatch/entry_points.py
```diff
@@ -0,0 +1,152 @@
...
+    model_class = models.get_class(model)
+
+    # Pass options as the first argument if provided, then *args and **kwargs
```
agriyakhetarpal:
```suggestion
```

I think this is self-explanatory, so it doesn't need a comment :)

tests/unit/test_parameters/test_parameter_sets_class.py
```diff
@@ -35,3 +35,28 @@ def test_iter(self):
...
+        print(docstring)
+        assert re.search(
+            "Single Particle Model",
+            docstring,
+        )
+
+    def test_iter(self):
+        """Test that iterating `pybamm.models` iterates over keys"""
+        for k in pybamm.dispatch.models:
+            assert isinstance(k, str)
```
agriyakhetarpal, line 62: There should be another file `test_dispatch.py` where you can check for:
- `pybamm.Model("SPM")` works on import
- another test but with model options

Edit: looks like the tests were updated when writing my review, so you could move some of the newer ones to `test_dispatch.py` instead.

pyproject.toml
santacodes: We could add more ofcourse!

src/pybamm/dispatch/entry_points.py
santacodes: I think it is a grammatical mistake

src/pybamm/dispatch/entry_points.py
santacodes: @agriyakhetarpal I think it is resolved as I dont see doctests failing anymore

tests/unit/test_dispatch.py
```diff
@@ -0,0 +1,44 @@
+#
+# Test dispatching mechanism in entry points
+#
+import pybamm
+
+
+class TestDispatch:
```
agriyakhetarpal, line 7: Could this be parametrised across `pybamm.dispatch.models.get_entries("pybamm_models")`?

tests/unit/test_dispatch.py
```diff
@@ -0,0 +1,44 @@
...
+        model = pybamm.Model("SPM", options=options)
+        assert model.__class__.__name__ == "SPM"
+
+    def test_model_function_with_args_kwargs(self):
+        """Test Model function with *args and **kwargs"""
+        model = pybamm.Model("SPM", build=False)
+        assert model.__class__.__name__ == "SPM"
+        options = {"thermal": "isothermal"}
+        model = pybamm.Model("SPM", options=options, build=False)
+        assert model.__class__.__name__ == "SPM"
```
agriyakhetarpal: I think these three tests can be parametrised.

tests/unit/test_parameters/test_parameter_sets_class.py
```diff
@@ -35,3 +35,106 @@ def test_iter(self):
         """Test that iterating `pybamm.parameter_sets` iterates over keys"""
         for k in pybamm.parameter_sets:
             assert isinstance(k, str)
+
+
+class TestModelEntryPoints:
```
agriyakhetarpal, line 40: These tests have been moved to the new file now, so I suppose you can remove them from here.

santacodes: No, only the tests that test the dispatch function have been moved, meaning the tests that load the model instance and pass options to them, the mechanism of model entry points is still being tested here.

agriyakhetarpal: - Thanks for the explanation!

agriyakhetarpal: Thanks for the explanation!

src/pybamm/dispatch/entry_points.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def __new__(cls, group):
+        """Ensure only two instances of entry points exist, one for parameter sets and the other for models"""
+        if EntryPoint._instances < 2:
+            cls.instance = super().__new__(cls)
+        return cls.instance
+
+    def get_class(self, key) -> Callable:
+        """Return the class without instantiating it"""
+        return self._load_entry_point(key)
```
agriyakhetarpal: Maybe we should keep this as private?

agriyakhetarpal (approved):
Thanks, @santacodes! It looks good to me, though I'll request that others should also take a look at it before we merge.

src/pybamm/dispatch/entry_points.py
```diff
@@ -0,0 +1,165 @@
+import importlib.metadata
+import sys
+import textwrap
+import warnings
+from collections.abc import Mapping
+from typing import Callable
```
Saransh-cpp:
```suggestion
from collections.abc import Callable, Mapping
```

src/pybamm/dispatch/entry_points.py
```diff
@@ -0,0 +1,165 @@
...
+            for entry_point in self.get_entries(self.group):
+                self._all_entries[entry_point.name] = entry_point
+
+    @staticmethod
+    def get_entries(group_name):
+        """Wrapper for the importlib version logic"""
+        if sys.version_info < (3, 10):  # pragma: no cover
+            return importlib.metadata.entry_points()[group_name]
+        else:
+            return importlib.metadata.entry_points(group=group_name)
```
Saransh-cpp: Hopefully #5052 will be merged first -
```suggestion
            return importlib.metadata.entry_points(group=group_name)
```

src/pybamm/dispatch/entry_points.py
```diff
@@ -0,0 +1,165 @@
...
+parameter_sets = EntryPoint(group="pybamm_parameter_sets")
+
+#: Singleton Instance of :class:EntryPoint initialised with pybamm_models"""
+models = EntryPoint(group="pybamm_models")
+
+
+def Model(model: str, options=None, *args, **kwargs):  # doctest: +SKIP
+    """
+    Returns the loaded model object
+    Parameters
```
Saransh-cpp: Nitpick
```suggestion
    Returns the loaded model object

    Parameters
```

src/pybamm/dispatch/entry_points.py
```diff
@@ -0,0 +1,165 @@
...
+        Options to pass to the model constructor.
+    *args
+        Additional positional arguments to pass to the model constructor.
+    **kwargs
+        Additional keyword arguments to pass to the model constructor.
+    Returns
+    -------
+    pybamm.model
+        Model object of the initialised model.
+    Examples
```
Saransh-cpp: Ditto
```suggestion
        Additional keyword arguments to pass to the model constructor.

    Returns
    -------
    pybamm.model
        Model object of the initialised model.

    Examples
```

pyproject.toml
```diff
@@ -148,6 +148,10 @@ ECM_Example = "pybamm.input.parameters.ecm.example_set:get_parameter_values"
 MSMR_Example = "pybamm.input.parameters.lithium_ion.MSMR_example_set:get_parameter_values"
 Chayambuka2022 = "pybamm.input.parameters.sodium_ion.Chayambuka2022:get_parameter_values"
 
+[project.entry-points."pybamm_models"]
+SPM = "pybamm.models.full_battery_models.lithium_ion.spm:SPM"
+DFN = "pybamm.models.full_battery_models.lithium_ion.dfn:DFN"
```
Saransh-cpp, line 146: Should this have all the models?

agriyakhetarpal:
> I have missed a couple of GSoC meetings, but will we be encouraging users to use the new API (deprecate the old model API/just encourage and not deprecate to keep backward compatibility)?
> 
> If we want users to use the new API, we should update our examples (which @medha-14 should be able to pick up in a new PR).

Hi @Saransh-cpp, filling you in: I think this is not the idea right now, as we would like to keep both public APIs as they will be used differently. The idea is that models added and validated in the PyBaMM repository will be added to `pybamm.lithium_ion`, `pybamm.lead_acid` and other namespaces, and the `pybamm.Model()` API is for generic/third-party models that users can load through entry points. We will be sticking to the same API for the in-house models for now. Some follow-ups here for @medha-14 will be to:
- create an example package using pybamm-cookie, preferably in the PyBaMM GitHub organisation, with an external model (simpler ones such as the reservoir model, or the heat equation, or so on) and an external parameter set to go along with it, both exposed as entry points
- write an example notebook for the PyBaMM repo, link to it in the third-party parameters page (see also #4406), which installs that example package and teaches users how to use this API
- update the pybamm-cookie repository to use these `pybamm.Model()` APIs when this PR is included in the next PyBaMM release, and coordinate with @santacodes to create a new release for pybamm-cookie when done

- In the meantime, the more extended discussion on how to better distribute models (i.e., as an alternative to installable Python packages that expose entry points for the PyBaMM models group): whether through Zarr stores, object storage (S3 or elsewhere), as JSON + exposed via a HTTP REST API, etc., can continue through the rest of the GSoC timeline. There is an attempt in #5056 to serialise non-discretised PyBaMM models, if I recall correctly.

Saransh-cpp:
Thanks for the detailed explanation, @agriyakhetarpal! Everything is very clear now.

pyproject.toml
santacodes: @Saransh-cpp I have added all the `lithium_ion_models` as of now. Others can be added at the next meeting once we decide on the model naming convention. I think @medha-14 can take that up in a separate PR.

src/pybamm/dispatch/entry_points.py
```diff
@@ -0,0 +1,166 @@
...
+    def _load_entry_point(self, key) -> Callable:
+        """Check that ``key`` is a registered ``parameter_sets`` or ``models` ,
+        and return the entry point for the parameter set/model, loading it needed."""
+        if key not in self._all_entries:
+            raise KeyError(f"Unknown parameter set or model: {key}")
+        param_set = self._all_entries[key]
+        try:
+            param_set = self._all_entries[key] = param_set.load()
+        except AttributeError:
+            pass
```
Saransh-cpp, line 88: Should this be handled?

src/pybamm/dispatch/entry_points.py
```diff
@@ -0,0 +1,166 @@
...
+                name in _all_entries and group == "pybamm_parameter_sets"
+            ):  # pragma: no cover
+                msg = (
+                    f"Parameter sets should be called directly by their name ({name}), "
+                    f"instead of via pybamm.parameter_sets (pybamm.parameter_sets.{name})."
+                )
+                warnings.warn(msg, DeprecationWarning, stacklevel=2)
+                return name
+        except AttributeError:
+            pass  # Handle the attribute error normally
```
Saransh-cpp: Ditto

src/pybamm/dispatch/entry_points.py
```diff
@@ -0,0 +1,166 @@
...
+    pybamm.model
+        Model object of the initialised model.
+
+    Examples
+    --------
+    Listing available models:
+        >>> import pybamm
+        >>> list(pybamm.dispatch.models) # doctest: +ELLIPSIS
+        ['DFN', 'MPM', 'MSMR', 'NewmanTobias', 'SPM', 'SPMe', 'Yang2017']
+        >>> pybamm.Model('SPM') # doctest: +SKIP
```
Saransh-cpp: Are +SKIP and +ELLIPSIS required given that the doctests are ignored class-wide above?

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Features
 
+- Unified entry points for models and parameter sets in `pybamm.dispatch.entry_points`, models can now be initialised through `pybamm.Model('model_name)` ([#4490](https://github.com/pybamm-team/PyBaMM/pull/4490))
```
agriyakhetarpal: I think we should not document this in this form, as this isn't battle-tested yet and we don't want users to use it right away until we have working examples we can follow (that @medha-14 will be working on). How about:

```suggestion
- Introduced entry points for models, similar to parameter sets, and moved entry point handling to `pybamm.dispatch.entry_points`. There is now experimental (⚠️) support for loading third-party models outside of the PyBaMM framework via `pybamm.Model("model_name")`. This API is currently unstable until further notice and may be subject to change without warning. ([#4490](https://github.com/pybamm-team/PyBaMM/pull/4490))
```

src/pybamm/dispatch/entry_points.py
```diff
@@ -0,0 +1,166 @@
...
+    and parameter sets that have been registered as entry points.
+
+    Access via :py:data:`pybamm.parameter_sets` for parameter sets - provides access
+    to all registered battery parameter sets (e.g., 'Chen2020', 'Ai2020') that can
+    be used to parameterise battery models.
+
+    Access via :py:data:`pybamm.Model` for models - provides access to all registered
+    battery models (e.g., 'SPM', 'DFN').
+
+    Note: This feature is in its experimental phase.
```
agriyakhetarpal: How about this?

```suggestion
    .. attention::

        This feature is currently experimental.
```

agriyakhetarpal (dismissed):

agriyakhetarpal (dismissed):

src/pybamm/dispatch/entry_points.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+#: Singleton Instance of :class:EntryPoint initialised with pybamm_parameter_sets"""
+parameter_sets = EntryPoint(group="pybamm_parameter_sets")
+
+#: Singleton Instance of :class:EntryPoint initialised with pybamm_models"""
+models = EntryPoint(group="pybamm_models")
+
+
+def Model(model: str, options=None, *args, **kwargs):  # doctest: +SKIP
```
kratman: Can we change the name of this function so that it looks less like a class?

src/pybamm/dispatch/entry_points.py
```diff
@@ -0,0 +1,168 @@
...
+        return self._load_entry_point(key)()
+
+    def _load_entry_point(self, key) -> Callable:
+        """Check that ``key`` is a registered ``parameter_sets`` or ``models` ,
+        and return the entry point for the parameter set/model, loading it needed."""
+        if key not in self._all_entries:
+            raise KeyError(f"Unknown parameter set or model: {key}")
+        param_set = self._all_entries[key]
+        try:
+            param_set = self._all_entries[key] = param_set.load()
```
Saransh-cpp: Given that this operates both on parameter sets and models?
```suggestion
        entry_point = self._all_entries[key]
        try:
            entry_point = self._all_entries[key] = entry_point.load()
```

src/pybamm/dispatch/entry_points.py
```diff
@@ -0,0 +1,168 @@
...
+        except AttributeError:
+            pass
+        return param_set
```
Saransh-cpp:
```suggestion
        return entry_point
```

src/pybamm/dispatch/entry_points.py
```diff
@@ -0,0 +1,168 @@
...
+            # Only show deprecation warning for parameter sets, not models
+            if (
+                name in _all_entries and group == "pybamm_parameter_sets"
+            ):  # pragma: no cover
+                msg = (
+                    f"Parameter sets should be called directly by their name ({name}), "
+                    f"instead of via pybamm.parameter_sets (pybamm.parameter_sets.{name})."
+                )
+                warnings.warn(msg, DeprecationWarning, stacklevel=2)
+                return name
```
kratman: This is probably old enough to remove

src/pybamm/dispatch/entry_points.py
```diff
@@ -0,0 +1,168 @@
...
+        except AttributeError:
+            pass  # Handle the attribute error normally
```
kratman: This looks more like it ignores the error than handles it normally

src/pybamm/dispatch/entry_points.py
Saransh-cpp: @santacodes please feel free to correct me, but I think that is to match the `ParameterValues` syntax. Making a `Model` class would be redundant if a function can do the job.

kratman (changes requested):

medha-14:
> Slightly deviating from my previous response in [#4490 (comment)](https://github.com/pybamm-team/PyBaMM/pull/4490#issuecomment-2975814188) here:
> 
> This PR is going to be merged soon enough as it is nearing its completion, but it would be nicer if we could find out any underlying issues with the implementation and iterate faster. I propose that we should not wait until a PyBaMM release with this to test changes downstream, and we should start a PR in the https://github.com/pybamm-team/pybamm-cookie repository (which can be kept as draft until a PyBaMM release comes out) which installs PyBaMM from this branch from @santacodes's fork and uses these APIs to instantiate external models and parameter sets. It would give us a chance to test out these changes and see if there's anything else we require in this PR or as a follow-up to it.
> 
> @medha-14, do you think you could get started on this?

Sure, I'll get started on it right away.

src/pybamm/dispatch/entry_points.py
santacodes: Yes, it is to match the `ParameterValues` syntax, plus model namespace is already consumed and cannot be overloaded. Also it was already discussed in one of the discussions earlier and we settled on this convention for model entry point.

src/pybamm/dispatch/entry_points.py
santacodes: Oops, Yeah it should be, for some reason I forgot to handle it.

src/pybamm/dispatch/entry_points.py
santacodes: I added them because for some reason the doctests were still failing even after I added them, I will probably remove them both as they don't seem to be making any difference being here.

src/pybamm/dispatch/entry_points.py
santacodes: @Saransh-cpp, the old `parameter_sets.py` which contains the implementation of `parameter_sets` entry point did not have any error handling for this function and the one @kratman mentioned below, so I kept it intact, though I will just add a print log now explaining the `AttributeError`.

src/pybamm/dispatch/entry_points.py
santacodes: Removed

src/pybamm/dispatch/entry_points.py
santacodes: Okay now I get why the exception had `pass` statement rather than error log. The load function can be only called for `models` and not `parameter_sets` because models are class entry points and can be returned as initialised objects and parameter_sets are dicts and cannot be loaded into objects.

santacodes: I'll leave it as is because there is no point of logging it as it is a part of functionality and not an actual error because we are trying to load the `entry_point` first to check if it is a model but `parameter_set` would cause an exception as it cannot be loaded hence the original variable `entry_point` which contains the `dict` will be returned.

src/pybamm/dispatch/entry_points.py
santacodes: Removed the class-wide SKIP

Saransh-cpp (dismissed):
Perfect, thanks for the amazing work on this, @santacodes! :)

(sorry for taking too long to review)

agriyakhetarpal:
Short comment for now: won't removing it introduce a breaking change, @santacodes? We should still expose `pybamm.parameter_sets` until the next few releases.

agriyakhetarpal (approved):

Saransh-cpp (approved):

agriyakhetarpal (approved):

src/pybamm/dispatch/entry_points.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+#: Singleton Instance of :class:EntryPoint initialised with pybamm_parameter_sets"""
+parameter_sets = EntryPoint(group="pybamm_parameter_sets")
+
+#: Singleton Instance of :class:EntryPoint initialised with pybamm_models"""
+models = EntryPoint(group="pybamm_models")
+
+
+def Model(model: str, options=None, *args, **kwargs):
```
valentinsulzer: Just do `*args, **kwargs`, no options. Not all models will have options as an argument

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Features
 
+- Introduced entry points for models, similar to parameter sets, and moved entry point handling to `pybamm.dispatch.entry_points`. There is now experimental (⚠️) support for loading third-party models outside of the PyBaMM framework via `pybamm.Model("model_name")`. This API is currently unstable until further notice and may be subject to change without warning. ([#4490](https://github.com/pybamm-team/PyBaMM/pull/4490))
```
kratman: Nitpick
```suggestion
- Introduced entry points for models, similar to parameter sets, and moved entry point handling to `pybamm.dispatch.entry_points`. There is now experimental support for loading third-party models outside of the PyBaMM framework via `pybamm.Model("model_name")`. This API is currently unstable until further notice and may be subject to change without warning. ([#4490](https://github.com/pybamm-team/PyBaMM/pull/4490))
```

src/pybamm/dispatch/entry_points.py
santacodes: @valentinsulzer removed it, although I am not sure why `readthedocs` build is failing. Reading the logs, I think it is unrelated to my changes.

agriyakhetarpal (dismissed):

valentinsulzer (approved):

PR #4494, opened 2024-10-06
Removes `param = self.param` to use `self.param` directly 

### LINKED ISSUE #4484: Remove `param = self.param` and use `self.param` directly instead
if these are the same, then why not just use self.param instead?

_Originally posted by @kratman in https://github.com/pybamm-team/PyBaMM/pull/4451#discussion_r1767514786_
            

There are 63 occurrences of `param = self.param` so I thought this would be better as separate issue.

medha-14:
# Description

Used regex accompanied with some manual checking and have removed `param = self.param`  and have modified the code to use `self.param` directly.
Fixes #4484 

## Type of change

# Key checklist:

## Further checks:


valentinsulzer:
I think this change makes the code harder to read

brosaplanella:
> I think this change makes the code harder to read

For reference, see discussion in #4484

src/pybamm/models/full_battery_models/lithium_ion/electrode_soh.py
```diff
@@ -276,7 +276,7 @@ class ElectrodeSOHSolver:
     ----------
     parameter_values : :class:`pybamm.ParameterValues.Parameters`
         The parameters of the simulation
-    param : :class:`pybamm.LithiumIonParameters`, optional
+    self.param : :class:`pybamm.LithiumIonParameters`, optional
```
kratman, line 279: The argument name does not change
```suggestion
    param : :class:`pybamm.LithiumIonParameters`, optional
```

src/pybamm/models/submodels/interface/kinetics/diffusion_limited.py
```diff
@@ -12,7 +12,7 @@ class DiffusionLimited(BaseInterface):
 
     Parameters
     ----------
-    param :
+    self.param :
```
kratman, line 15: The input argument does not change
```suggestion
    param :
```

kratman (commented):
Couple mistakes

kratman (approved):
Looks good now.

To address the readability concerns from this change, I think the best approach is that we should move towards breaking up the long functions into much smaller ones. That way we can be clear that we are using class memory while also making the code easier to read. This is outside the scope of this PR though

PR #4501, opened 2024-10-08
feat: add discrete time sum expression tree node

### LINKED ISSUE #4485: add `DiscreteTimeSum` subclass of `pybamm.Symbol`
### Description

This would add a new unary operator in a pybamm expression tree that would represent a discrete sum over time. This would be a discrete version of the already existing `pybamm.ExplicitTimeIntegral` class.

### Motivation

This would give model developers and users the ability to add "sum of squares" type variables to a pybamm model that would calculate the difference between, for example, a solution variable and a user-provided dataset. This would be useful for implementing parameter inference using pybamm

### Possible Implementation

Similar to the `ExplicitTimeIntegral`, this would only be evaluated in the `Solution` class (see `_update_variable` function). I would propose that the sum is done over the list of time points in the solution (rather than have a separate list of points in the expression tree node), that way a user can specify the time-points via `t_interp`.

I'm a bit unsure how to allow users to provide their data to the expression. Perhaps they would wrap it in an Interpolant:

```python
data = pybamm.Interpolant(data_t, data_y, pybamm.t)
model.variable["data_comparison"] = pybamm.DiscreteTimeSum((model.variable["Voltage"] - data)**2)
```

### Additional context

see https://github.com/pybop-team/PyBOP/issues/513

martinjrobins:
# Description

Adds two new subclasses to `pybamm.Symbol`:
1. `pybamm.DiscreteTimeData` which is an interpolant for holding time series data
2. `pybamm.DiscreteTimeSum` which is a unary operator representing a sum of its child evaluation over a set of data time points. Its child must have exactly one `pybamm.DiscreteTimeData` object. 

When `pybamm.DiscreteTimeSum` is evaluated normally it returns the evaluation of its child. If it is evaluated as part of a `pybamm.Solution` then it interpolates this solution onto the data time points and sums these over the time axis

This also updates the evaluation of `pybamm.ExplicitTimeIntegral`. Previously this performed a trapezoidal integral over all the time points passed to the `__call__` function of `pybamm.ProcessedVariable`. Now it performs the integral over the solution time points, which is more consistent with its definition and always returns the same value irrespective of the time values that the user passes to `__call__`. In the future this integral should be updated to use the gradient information of the solution if available.

Fixes #4485

## Type of change

# Key checklist:

## Further checks:


tests/unit/test_solvers/test_solution.py
```diff
@@ -420,3 +420,43 @@ def test_solution_evals_with_inputs(self):
...
+        model.variables["data_comparison"] = data_comparison
+
+        solver = solver_class()
+        for a in [0.5, 1.0, 2.0]:
+            sol = solver.solve(model, t_eval=t_eval, inputs={"a": a})
+            y_sol = np.exp(-a * data_times)
+            expected = np.sum((y_sol - data_values) ** 2)
+            np.testing.assert_array_almost_equal(
+                sol["data_comparison"](t=sol.t), expected, decimal=2
+            )
```
MarcBerliner, line 462: Should this be
```python
sol["data_comparison"](t=data_times)
```

MarcBerliner (commented):
Looks great! Thanks for also separating out and simplifying the `ExplicitTimeIntegral`. The changelog needs updating as well.

tests/unit/test_solvers/test_solution.py
martinjrobins: that probably should just be `sol["data_comparison"]()` as the result is independent of t, I'll change that

MarcBerliner (approved):

PR #4507, opened 2024-10-11
Add "voltage as a state" option

rtimms:
# Description
Adds an option "voltage as a state" that can be "false" (default) or "true". If "true" adds an explicit algebraic equation for the voltage. This can be used to reformulate the SPM(e) as a DAE rather than and ODE. This increases the solve time but evaluating the voltage as an explicit state rather than as a function of other state variables can be faster if the time eval is dense.

```python3
import pybamm
import numpy as np

t = np.linspace(0, 3600, 100000)
for value in ["true", "false"]:
    model = pybamm.lithium_ion.SPMe({"voltage as a state": value})
    sim = pybamm.Simulation(model, solver=pybamm.IDAKLUSolver())
    sim.solve([0, 3600])
    print("voltage as a state:", value)
    solve_time = sim.solution.solve_time
    print("solve time:", solve_time)
    timer = pybamm.Timer()
    sim.solution["Voltage [V]"](t)
    eval_time = timer.time()
    print("eval time:", eval_time)
    print("total time:", solve_time + eval_time, "\n")
```

returns 
```
voltage as a state: true
solve time: 3.293 ms
eval time: 8.836 ms
total time: 12.129 ms 

voltage as a state: false
solve time: 2.311 ms
eval time: 687.892 ms
total time: 690.204 ms 
```

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/src/pybamm/models/full_battery_models/base_battery_model.py
+++ b/src/pybamm/models/full_battery_models/base_battery_model.py
@@ -210,6 +210,9 @@ class BatteryModelOptions(pybamm.FuzzyDict):
                 solve an algebraic equation for it. Default is "false", unless "SEI film
                 resistance" is distributed in which case it is automatically set to
                 "true".
+            * "voltage as a state" : str
+                Whether to make a state for the voltage and solve an algebraic equation
+                for it. Default is "false".
             * "working electrode" : str
                 Can be "both" (default) for a standard battery or "positive" for a
                 half-cell where the negative electrode is replaced with a lithium metal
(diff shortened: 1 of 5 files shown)


valentinsulzer (approved):
Looks good, thanks. Those timings are crazy - is that even with the latest changes from Marc?

rtimms:
Yep. Evaluating all those `asinh` terms is slow if you have a large `t_eval`.

kratman (approved):

PR #4516, opened 2024-10-15
Composite model for particle mechanics

mohammedasher:
# Description

Implementing the composite model for particle mechanics: The phase in the current particle mechanics implementation is set to 'primary' by default. This is being updated to support independent phases (e.g., primary/secondary such as Gr/Si). To accommodate these changes, the associated files in full_battery_models, interface, SEI, and particle modules have been updated, along with modifications to the Chen2020 parameter set.

This is part of the work by [Bonkile et al. ](https://www.sciencedirect.com/science/article/pii/S2352152X23020066)

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:


mohammedasher:
@brosaplanella Unable to request specific reviewer in draft mode.

brosaplanella:
Hi Asher! The style tests are failing, this is because you have some defined variables/function that are not used. This is an issue it can't be automatically fixed by the pre-commit bot, so you will need to do it manually. You can find what's causing the issue if you check the failing test (https://github.com/pybamm-team/PyBaMM/actions/runs/11345431722/job/31553629641?pr=4516) and scroll down (you need to start reading at line 46).

mohammedasher:
Hi @brosaplanella,
I've addressed the style issues, and now the remaining problem is with the unit tests. Although multiple tests are failing, they all stem from two types of errors:
1. AttributeError: 'set' object has no attribute 'values'
2. AttributeError: 'ReactionDriven' object has no attribute 'phase_name'
Could you advise on how I should proceed with resolving these? Thank you!

brosaplanella:
Hi! The latter is because the ReactionDriven porosity model (`reaction_driven_porosity.py`) is not available for composite materials, that's why it breaks. I need to dig a bit more regarding the first issue.

BTW, it's better to not force push. I believe the issue you encountered is that the pre-commit bot made some style changes. In those cases you have to pull first and then push.

DrSOKane:
> Hi! The latter is because the ReactionDriven porosity model (`reaction_driven_porosity.py`) is not available for composite materials, that's why it breaks.

I made a PR that fixes this, which Sunil is already using (he requested it). Go to [this branch of my fork of PyBaMM](https://github.com/DrSOKane/PyBaMM/tree/composite-porosity) and merge the changes into your branch.

mohmdasher:
Thank you @DrSOKane; merging reaction_driven_porosity from your branch and significant reduced the number of errors. 
The errors now are:
- AttributeError: 'set' object has no attribute 'values'
- AttributeError: 'ReactionDriven' object has no attribute 'phase_name' 
       Can this be re-scoped to self.phase_name??
- pybamm.expression_tree.exceptions.ModelError: Missing variable for submodel 'porosity': 'Negative total SEI thick...' 
- Failed: DID NOT RAISE <class 'pybamm.expression_tree.exceptions.OptionError'>

.gitignore
```diff
@@ -138,3 +138,5 @@ results/
 
 # tests
 test_callback.log
+PublicPyBaMM_Composite/
+PyBaMM/
```
brosaplanella, line 142: These two are repository specific, I would suggest to remove them

src/pybamm/models/submodels/porosity/reaction_driven_porosity.py
```diff
@@ -37,7 +38,9 @@ def get_coupled_variables(self, variables):
                 L_pl_k = variables[f"{Domain} lithium plating thickness [m]"]
                 L_dead_k = variables[f"{Domain} dead lithium thickness [m]"]
                 L_sei_cr_k = variables[f"{Domain} total SEI on cracks thickness [m]"]
-                roughness_k = variables[f"{Domain} electrode roughness ratio"]
+                roughness_k = variables[
+                    f"{Domain} electrode {self.phase_name}roughness ratio"
```
brosaplanella, line 42: The issue with the `phase_name` is that this class does not take a `phase` as argument (see for example how `loss_active_material.py` does it. However, I don't understand how this model works: you have different values for different phases for the roughness, but the SEI on cracks thickness is only one. Shouldn't you just have one roughness ratio overall?

brosaplanella (commented):
Some comments that should help fix the bugs. Regarding the one about the "Negative total SEI thickness [m]" I have not been able to find where the issue is. Maybe @DrSOKane knows, as he has been working with the model recently.

src/pybamm/models/submodels/porosity/reaction_driven_porosity.py
mohammedasher: @DrSOKane could you please advise on how I can go about with introducing phases in the reaction-driven-porosity model? Most of the remaining errors are stemming from this file. Thank you.

DrSOKane: Hi @mohammedasher, you need to move the roughness ratio inside the loop `for phase in phases:`. Something like this should work:

`roughness_k = variables[
    f{Domain} total {phase_name}SEI on cracks thickness [m]"
]`

Then, delete the reference to `roughness_k` outside the loop (line 36 in my version). `phase_name` is defined at the start of the loop `for phase in phases:` so that gets around the problems Ferran rightly pointed out.

mohammedasher: Hi @DrSOKane....I will define a new `for phase in phases:` loop; apart from the `roughness_k` variable, `L_sei_k = variables[f"{Domain} total SEI thickness [m]"] ` also runs into error when the domain is negative (negative electrode). I tried `L_sei_k = variables[f"{Phase}: {Domain} total SEI thickness [m]"]', but it did not help.

DrSOKane: @mohammedasher the code you need is [here](https://github.com/DrSOKane/PyBaMM/blob/composite-porosity/src/pybamm/models/submodels/porosity/reaction_driven_porosity.py). Uncomment lines 63-65 and deleted line 36, and it should work.

kratman:
@mohammedasher Are you still working on this?

mohammedasher:
> @mohammedasher Are you still working on this?

@kratman Yes, I'm working on this...

rtimms:
@aabills does this overlap with what your did / are doing?

aabills:
I think this is mostly closed by #4698, though its hard to tell if I got everything.

aabills:
@mohammedasher Can you check and see if what you were intending to fix here works on pybamm 25.1.1?

mohammedasher:
> @mohammedasher Can you check and see if what you were intending to fix here works on pybamm 25.1.1?

Hi @aabills, I think your work on #4698 addresses this PR i.e. to make particle mechanics work for composite electrodes; as a next step I wanted to include thermal model, LLI and LAM for composite electrodes. I see that the changes you have made should address the thermal part as well, can you please confirm?

Thank you.

aabills:
Thermal models shouldn't be affected by the composite model because thermal model occurs at the electrode domain, while the composite model is at the particle domain, so temperature can just be broadcast to the composite parts of the models.

Thanks for your attempted contribution anyway! Sorry, we normally wouldn't fly through it when we have someone trying to contribute, but we had an urgent need to get this working.

PR #4518, opened 2024-10-16
Updated np.ndarray to numpy.typing.NDArray everywhere by writing a script to do the changes automatically

### LINKED ISSUE #4512: Use `numpy.typing.NDArray` instead of `np.ndarray`
I noticed that pybamm uses `np.ndarray` everywhere, but we should use [numpy.typing.NDArray](https://numpy.org/devdocs/reference/typing.html#numpy.typing.NDArray) wherever possible (everywhere). Ideally they should be swappable without any errors. Feel free to close this issue if there was a reason to specifically use `np.ndarray`.

sanjayh-2022:
# Description

Fixes #4512

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/benchmarks/different_model_options.py
+++ b/benchmarks/different_model_options.py
@@ -33,8 +33,8 @@ def build_model(parameter, model_, option, value):
 class SolveModel:
     solver: pybamm.BaseSolver
     model: pybamm.BaseModel
-    t_eval: np.ndarray
-    t_interp: np.ndarray | None
+    t_eval: numpy.typing.NDArray
+    t_interp: numpy.typing.NDArray | None
 
     def solve_setup(self, parameter, model_, option, value, solver_class):
         import importlib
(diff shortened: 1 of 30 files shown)


arjxn-py (changes requested):
Hi @sanjayh-2022, thanks for starting this.

Just a heads up, You didn't actually needed to write a script to do these changes automatically as you can always use VS code search & replace using `Ctrl`+`Shift`+`F`.

Another thing that you might want to replace `numpy.typing.NDArray` with `np.typing.NDArray` everywhere, since we `import numpy as np`

benchmarks/different_model_options.py
```diff
@@ -33,8 +33,8 @@ def build_model(parameter, model_, option, value):
 class SolveModel:
     solver: pybamm.BaseSolver
     model: pybamm.BaseModel
-    t_eval: np.ndarray
-    t_interp: np.ndarray | None
+    t_eval: numpy.typing.NDArray
```
Saransh-cpp, line 36: We are importing `numpy` as `np` so this would obviously not work. In fact, we should `import numpy.typing as npt` and use `npt.NDArray` everywhere.

Saransh-cpp (changes requested):

PR #4520, opened 2024-10-16
Removing the macos-12 runner

kratman:
# Description

The macos-12 runner is being deprecated in December. This switches to the macos-13 runner

# Key checklist:

## Further checks:

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -40,7 +40,7 @@ def set_iree_state():
             # iree-compiler is currently only available as a wheel on macOS 13 (or
             # higher) and Python version 3.11
             mac_ver = int(platform.mac_ver()[0].split(".")[0])
-            if (not sys.version_info[:2] == (3, 11)) or mac_ver < 13:
+            if (not sys.version_info[:2] == (3, 11)) or mac_ver < 14:
                 warnings.warn(
                     (
                         "IREE is only supported on MacOS 13 (or higher) and Python"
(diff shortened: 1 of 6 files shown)


agriyakhetarpal (approved):

kratman:
Converting to a draft while I take a look at the failure. It does appear to be repeatable

agriyakhetarpal:
It looks like it's coming from the MLIR stuff, since Python 3.11 on macOS is the only Python version where it's enabled

kratman:
@agriyakhetarpal The version of IREE used before was a nightly build not a stable build, so I will try using the stable one instead

noxfile.py
```diff
@@ -40,7 +40,7 @@ def set_iree_state():
             # iree-compiler is currently only available as a wheel on macOS 13 (or
             # higher) and Python version 3.11
             mac_ver = int(platform.mac_ver()[0].split(".")[0])
-            if (not sys.version_info[:2] == (3, 11)) or mac_ver < 13:
+            if (not sys.version_info[:2] == (3, 11)) or mac_ver < 14:
```
kratman, line 43: Temporary fix to make this work. I will create an issue for it

kratman:
@agriyakhetarpal Let me know if this temporary fix is acceptable. I created this issue for solving the MacOS-13 issue #4521

agriyakhetarpal:
Yes, I think we can't do much about this, so this is acceptable for me

kratman:
> Ah, I didn't realise my review was dismissed. I'm personally not a fan of this setting (we experimented with it and later removed it in the `pybamm-cookie` repo), since the "Require conversation resolution before merging" option worked well.

Except I have seen a bunch of PRs get merged with real changes that never got reviewed due to the lack of having this setting. It is a necessary setting for a good project

agriyakhetarpal:
Sure, let's keep this in that case, then!

PR #4526, opened 2024-10-18
changed np.nparray to np.typing.NDArray

### LINKED ISSUE #4512: Use `numpy.typing.NDArray` instead of `np.ndarray`

sanjayh-2022:
# Description

Fixes #4512

## Type of change

# Key checklist:

## Further checks:


agriyakhetarpal (changes requested):
Thanks, @sanjayh-2022! Could you instead the NumPy typing submodule separately using the `import numpy.typing as npt` statement in addition to `import np` and use `npt.NDArray` here instead, like @Saransh-cpp has [previously suggested](https://github.com/pybamm-team/PyBaMM/pull/4518#discussion_r1802556405)?

src/pybamm/solvers/idaklu_jax.py
```diff
@@ -274,18 +274,18 @@ def f_isolated(*args, **kwargs):
...
         self,
-        t: np.ndarray = None,
+        t: np.typing.NDArray = None,
         inputs: Union[dict, None] = None,
         output_variables: Union[list[str], None] = None,
     ):
         """Helper function to compute the gradient of a jaxified expression
 
-        Returns a numeric (np.ndarray) object (not a JAX expression).
+        Returns a numeric (np.typing.NDArray) object (not a JAX expression).
```
kratman, line 283: I think in comments we would probably want to keep it np.ndarray instead of the typing object. My understanding is that the np.typing module is just for type hints.

sanjayh-2022:
I have made the required changes. please review my changes and let me know if any changes are required

arjxn-py (changes requested):
Thanks @sanjayh-2022, I'd once again have to request changes as you've to also be careful about the order of imports in a file. It'd be nicer to `import numpy.typing as npt` not on the top of the file but the line after `import numpy as np`

agriyakhetarpal:
You can fix the pre-commit CI and the style failures by installing `pre-commit` locally, please read through our Contributing guide: https://docs.pybamm.org/en/latest/source/user_guide/contributing.html#installing-and-using-pre-commit. You may also use `nox -s pre-commit` to do the same thing.

In particular, the following lines:

```python
pip install pre-commit
pre-commit install
```

will run the style checks before any code is committed, and will hence prevent you from pushing code that fails these style checks.

sanjayh-2022:
i have added  import numpy.typing as npt after import numpy as np. Please review the changes

src/pybamm/experiment/step/base_step.py
```diff
@@ -74,7 +76,7 @@ def __init__(
         self.input_duration = duration
         self.input_value = value
         # Check if drive cycle
-        is_drive_cycle = isinstance(value, np.ndarray)
+        is_drive_cycle = isinstance(value, npt.NDArray)
```
Saransh-cpp, line 79: This should not be changed as we are checking if `value` is an instance of `np.ndarray`. We could have changed it if we were checking the type of the `value`, but we don't want that.

src/pybamm/discretisations/discretisation.py
```diff
@@ -1012,7 +1014,7 @@ def check_initial_conditions(self, model):
         # Individual
         for var, eqn in model.initial_conditions.items():
             ic_eval = eqn.evaluate(t=0, inputs="shape test")
-            if not isinstance(ic_eval, np.ndarray):
+            if not isinstance(ic_eval, npt.NDArray):
```
Saransh-cpp, line 1017: Ditto

src/pybamm/experiment/step/base_step.py
```diff
@@ -260,7 +262,7 @@ def default_duration(self, value):
         Default duration for the step is one day (24 hours) or the duration of the
         drive cycle
         """
-        if isinstance(value, np.ndarray):
+        if isinstance(value, npt.NDArray):
```
Saransh-cpp, line 265: Ditto

src/pybamm/expression_tree/array.py
```diff
@@ -165,7 +167,7 @@ def to_json(self):
         Method to serialise an Array object into JSON.
         """
 
-        if isinstance(self.entries, np.ndarray):
+        if isinstance(self.entries, npt.NDArray):
```
Saransh-cpp, line 170: Ditto

src/pybamm/expression_tree/binary_operators.py
```diff
@@ -22,13 +24,13 @@ def _preprocess_binary(
 ) -> tuple[pybamm.Symbol, pybamm.Symbol]:
     if isinstance(left, (float, int, np.number)):
         left = pybamm.Scalar(left)
-    elif isinstance(left, np.ndarray):
+    elif isinstance(left, npt.NDArray):
```
Saransh-cpp, line 27: For all `isinstance` calls basically

Saransh-cpp (changes requested):
Thanks, @sanjayh-2022! Please see my comments below.

Saransh-cpp (changes requested):

kratman:
@sanjayh-2022 Are you still working on this?

agriyakhetarpal:
Marking as a draft for now, thanks for your work here so far!

kratman:
There has been no activity since the end of October, closing this for now

Issue #4524, opened 2024-10-18
[Bug]: IDAKLU Solver incompatible with OKane2022 NE OCP Interpolation


tomjholland:
### PyBaMM Version

24.9

### Python Version

3.11

### Describe the bug

Casadi error is returned when using the IDAKLU solver with the `OKane2022` parameter set. Appears to be related to the interpolation of the negative electrode OCP, as it is searching for `libcasadi_interpolatnt_bspline.dylib` and setting the OCP to the functional form in `Chen2020` eliminates the error. 

This is a separate issue, but I will tag #4365 as another issue caused by the interpolation of discrete points of NE OCP in this parameter set.

### Steps to Reproduce

```python
exp = pybamm.Experiment([ "Discharge at 1C until 2.5 V"])
model = pybamm.lithium_ion.DFN()
params = pybamm.ParameterValues("OKane2022")
solver = pybamm.IDAKLUSolver()
sim = pybamm.Simulation(model, experiment=exp, parameter_values=params, solver=solver)
sim.solve()
```
Produces the Runtime error:
```
RuntimeError: .../casadi/core/casadi_os.cpp:166: Assertion "handle!=nullptr" failed:
PluginInterface::load_plugin: Cannot load shared library 'libcasadi_interpolant_bspline.dylib': 
   (
    Searched directories: 1. casadipath from GlobalOptions
                          2. CASADIPATH env var
                          3. PATH env var (Windows)
                          4. LD_LIBRARY_PATH env var (Linux)
                          5. DYLD_LIBRARY_PATH env var (osx)
    A library may be 'not found' even if the file exists:
          * library is not compatible (different compiler/bitness)
          * the dependencies are not found
   )
  Tried '' :
[...]

Issue #4521, opened 2024-10-17
Jax-IREE not working on MacOS-13


kratman:
While removing the deprecated MacOS-12 runners for GitHub actions, I ran into issues with IREE. The comments suggest that it should work with MacOS version 13, but the tests fail. It seems to work fine one MacOS-14 and Linux.

As a temporary fix, I suggested changed the version in PR #4520
```python
            if (not sys.version_info[:2] == (3, 11)) or mac_ver < 14:
                warnings.warn(
                    (
                        "IREE is only supported on MacOS 13 (or higher) and Python"
                        "version 3.11. Setting PYBAMM_IDAKLU_EXPR_IREE=OFF."
                    ),
                    stacklevel=2,
                )
                return "OFF"
```

A few other things I noticed while looking into the failures:
- Updating Jax, JaxLib, and IREE requires updating versions in multiple places which is not easy to maintain. This should be centralized somehow
    - jax versions in `utils.py` and `pyproject.toml`
    - IREE compiler versions are set in both `noxfile.py` and `pyproject.toml`
- Newer versions of Jax/JaxLib have removed support for Python 3.9, so updating Jax and IREE could be difficult until we remove support for python 3.9

agriyakhetarpal:
cc @jsbrittain who could help us with this. I had a macOS device running macOS 13, but I couldn't reproduce the crash when running the relevant tests from the test suite locally (ran them a few times). I updated it to macOS 14 today, so I don't know if we can reliably reproduce this outside of a CI runner.

jsbrittain:
@agriyakhetarpal just a very quick comment for the time being: as noted above, there are multiple components that need to be upgraded together, but this has to be done carefully. I actually took a look at this a couple of months back in an effort to get rid of the demotion code (IREE only runs at 32-bit precision at the present). Notes from that investigation are here and may prove helpful: https://github.com/jsbrittain/PyBaMM/issues/4

jsbrittain:
Some follow-up exploratory notes. I tested upgrading to the latest Jax/Jaxlib (0.4.34) and corresponding iree-compiler (1036) with relaxed version constraints on everything apart from Windows. I don't have a macOS-13 machine but using gh-runners produced a slightly different set of results to your previous run (see https://github.com/jsbrittain/PyBaMM/actions/runs/11391403398/job/31695386575?pr=5):

ubuntu
3.9 jax incompatible
3.10-3.12 operations cancelled

macos-13
[...]

PR #4528, opened 2024-10-18
Fix CasADI path and disable IREE for MacOS

kratman:
# Description

Quick fixes for the issues with CasADI and IREE.

Related #4524
Related #4521 

## Type of change

The CasADI fix should work for MacOS and Linux builds, but does not fix Windows issues.

# Key checklist:

## Further checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,8 +21,8 @@
 
 ## Bug Fixes
 
-- Disabled IREE on MacOS due to compatibility issues and added the CasADI 
-  path to the environment to resolve issues on MacOS and Linux. Windows 
+- Disabled IREE on MacOS due to compatibility issues and added the CasADI
+  path to the environment to resolve issues on MacOS and Linux. Windows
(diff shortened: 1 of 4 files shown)


kratman:
I will deal with the lychee issues later, since that is not connected to this PR. We have entirely too many links that break

CHANGELOG.md
```diff
@@ -21,6 +21,9 @@
 
 ## Bug Fixes
 
+- Disabled IREE on MacOS due to compatibility issues and added the CasADI
+  path to the environment to resolve issues on MacOS and Linux. Windows
+  users may still experience issues with interpolation. ([#4528](https://github.com/pybamm-team/PyBaMM/pull/4528))
```
MarcBerliner, line 26: It's not a big deal, but removing `IREE` should be under breaking changes, right?

MarcBerliner (approved):

CHANGELOG.md
kratman: I put it there because it was probably not used. It was only enabled for a single python version, so unlikely that anyone will be using it. I will put the full removal of IREE under breaking changes when I do that work.

PR #4534, opened 2024-10-21
Allow using IDAKLU(output_variables=...) with Experiments

### LINKED ISSUE #4386: Make IDAKLU output variables compatible with Experiments
This was initially reported as a bug (see report below) but it turns out it is not an implemented feature yet. It seems the main issue is that when a new experiment step is initialised with the previous step solution, that one is empty so the relevant variables are not available. I believe that amending the `last_state` so it stores the whole state vector even when output variables are used should fix this.

### PyBaMM Version

develop

### Python Version

3.11.6

### Describe the bug

I am running a model with the `IDAKLUSolver` and some `output_variables`. These include, amongst others, the `"Discharge capacity [A.h]"`, however PyBaMM complains that that variable is not found.

### Steps to Reproduce

```python3
import pybamm

pybamm.set_logging_level("NOTICE")

output_variables = ["Discharge capacity [A.h]", "Time [s]", "Current [A]", "Voltage [V]"]
solver = pybamm.IDAKLUSolver(output_variables=output_variables)

    [
        (
            "Charge at 1C until 4.2 V",
            "Hold at 4.2 V until C/50",
            "Rest for 1 hour",
        )
    ]
)

sim = pybamm.Simulation(
    model,
    experiment=experiment,
    solver=solver,
)

sol = sim.solve()
```

### Relevant log output

```shell
  File "/home/brosaplanella/PyBaMM/src/pybamm/models/base_model.py", line 808, in set_initial_conditions_from
    final_state = solution[var.name]
                  ~~~~~~~~^^^^^^^^^^
  File "/home/brosaplanella/PyBaMM/src/pybamm/solvers/solution.py", line 547, in __getitem__
    self.update(key)
  File "/home/brosaplanella/PyBaMM/src/pybamm/solvers/solution.py", line 481, in update
    raise KeyError(
KeyError: "Cannot process variable 'Discharge capacity [A.h]' as it was not part of the solve. Please re-run the solve with `output_variables` set to include this variable."

  File "/home/brosaplanella/battery_modelling/degradation-data/examples/mwe.py", line 29, in <module>
    sol = sim.solve()
          ^^^^^^^^^^^
  File "/home/brosaplanella/PyBaMM/src/pybamm/simulation.py", line 702, in solve
    step_solution = solver.step(
                    ^^^^^^^^^^^^
  File "/home/brosaplanella/PyBaMM/src/pybamm/solvers/base_solver.py", line 1222, in step
    _, concatenated_initial_conditions = model.set_initial_conditions_from(
                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/brosaplanella/PyBaMM/src/pybamm/models/base_model.py", line 810, in set_initial_conditions_from
    raise pybamm.ModelError(
pybamm.expression_tree.exceptions.ModelError: To update a model from a solution, each variable in model.initial_conditions must appear in the solution with the same key as the variable name. In the solution provided, 'Cannot process variable 'Discharge capacity [A.h]' as it was not part of the solve. Please re-run the solve with `output_variables` set to include this variable.' was not found.
```

pipliggins:
# Description

Allows users to use the `output_variables` option with the IDAKLU solver within an experiment.

Fixes #4386

## Type of change

# Key checklist:

## Further checks:


src/pybamm/solvers/solution.py
```diff
@@ -460,9 +460,20 @@ def first_state(self):
         else:
             all_yps = self.all_yps[0][:, :1]
 
+        # check if all_ys is empty (i.e. if IDAKLU solver + output_variables used)
+        if isinstance(self.all_ys[-1], casadi.DM):
+            empty_ys = False
+        else:
+            empty_ys = self.all_ys[0].shape[0] == 0
+
```
MarcBerliner, line 468: Can we pass this information from the solver and make it a `Solution` property? It would be good to check the root cause instead of a consequence.

src/pybamm/solvers/solution.py
```diff
@@ -460,9 +460,20 @@ def first_state(self):
...
         new_sol = Solution(
             self.all_ts[0][:1],
-            self.all_ys[0][:, :1],
+            # Get first state from initial conditions if all_ys is empty
+            (
+                self.all_ys[0][:, :1]
+                if not empty_ys
+                else self.all_models[0].y0full.reshape(-1, 1)
+            ),
```
MarcBerliner, line 476: The pre-commit formatting makes this a bit hard to read. Can we move this into an if statement above?

src/pybamm/solvers/solution.py
```diff
@@ -500,9 +511,19 @@ def last_state(self):
...
+            empty_ys = self.all_ys[-1].shape[0] == 0
+
         new_sol = Solution(
             self.all_ts[-1][-1:],
-            self.all_ys[-1][:, -1:],
+            (
+                self.all_ys[-1][:, -1:]
+                if not empty_ys
+                else self.y_event.reshape(len(self.y_event), 1)
+            ),
```
MarcBerliner, line 526: ^same

src/pybamm/solvers/solution.py
```diff
@@ -962,6 +983,19 @@ def __add__(self, other):
...
         new_sol._sub_solutions = self.sub_solutions + other.sub_solutions
 
+        # update variables which were derived at the solver stage
+        if other._variables and all(
+            isinstance(v, pybamm.ProcessedVariableComputed)
+            for v in other._variables.values()
+        ):
+            if not self._variables:
+                new_sol._variables = other._variables.copy()
+            elif self._variables:
```
MarcBerliner, line 993:
```suggestion
            else:
```

tests/integration/test_solvers/test_idaklu.py
```diff
@@ -152,3 +152,54 @@ def test_interpolation(self):
...
+                    "Discharge capacity [A.h]",  # 0D variables
+                    "Time [s]",
+                    "Current [A]",
+                    "Voltage [V]",
+                    "Pressure [Pa]",  # 1D variable
+                    "Positive particle effective diffusivity [m2.s-1]",  # 2D variable
+                ]
+                solver = pybamm.IDAKLUSolver(output_variables=output_variables)
+            else:
+                solver = pybamm.IDAKLUSolver()
```
MarcBerliner, line 173:
```suggestion
            else:
                output_variables = None
            solver = pybamm.IDAKLUSolver(output_variables=output_variables)
```

MarcBerliner (changes requested):
Looks good @pipliggins, just a few minor changes. The changelog needs updating as well

src/pybamm/solvers/solution.py
pipliggins: Added a 'variables_returned' flag to the Solution class to use for this

PR #4541, opened 2024-10-23
Fix indexing bug with infeasible experiments for `IDAKLUSolver`

### LINKED ISSUE #4496: [Bug]: IDAKLUSolver output variables
### PyBaMM Version

24.9

### Python Version

3.11

### Describe the bug

Running the same experiment but with cycles tagged differently can make the `IDAKLUSolver` return a solution without all the variables. 

@MarcBerliner @martinjrobins 

### Steps to Reproduce

```

# works
experiment1 = pybamm.Experiment(
    [
        (
            "Rest for 1 hour",
            "Charge at C/5 until 4.2 V",
            "Hold at 4.2 V until C/20",
        ),
    ]
)
sim1 = pybamm.Simulation(
    model, parameter_values=parameter_values, experiment=experiment1, solver=solver
)
sim1.solve()
sim1.plot()

# fails
experiment2 = pybamm.Experiment(
    [
        ("Rest for 1 hour"),
        (
            "Charge at C/5 until 4.2 V",
            "Hold at 4.2 V until C/20",
        ),
    ]
)
sim2 = pybamm.Simulation(
    model, parameter_values=parameter_values, experiment=experiment2, solver=solver
)
sim2.solve()
sim2.plot()

```

### Relevant log output

KeyError: "Cannot process variable 'Negative particle surface concentration [mol.m-3]' as it was not part of the solve. Please re-run the solve with `output_variables` set to include this variable."

MarcBerliner:
# Description

@pipliggins figured out why infeasible experiments were failing with the `IDAKLUSolver`:
> I checked this with #4534 to confirm it's not an `output_variables` specific problem - it's not, but related. 
> 
> Having the voltage limit violation at the start of a cycle step results in the final solution containing an empty array in the middle of `all_ys`. The assumption leading to the output_variables error was that the only time an empty numpy array would be present in `all_ys` is if the solver isn’t returning the full state vector, as with IDAKLU(output_variables=true), so an false error is being flagged. I fixed that by checking for an empty `t` array too in `_update_variables()`, i.e.:
> ```
>       for i, (model, ts, ys, inputs, var_pybamm) in enumerate(
>           zip(self.all_models, self.all_ts, self.all_ys, self.all_inputs, vars_pybamm)
>       ):
>           if (
>               ys.size == 0
>               and var_pybamm.has_symbol_of_classes(
>                   pybamm.expression_tree.state_vector.StateVector
>               )
>               and not ts.size == 0
>           ):
> ```
> But using this I get a different error related to the hermite interpolation. Turning off hermite interpolation in the solver (and with the output_variables fix) the example runs as expected. I suspect the interpolation issue is also being caused by an unexpected empty array in the state vector, but haven’t dug that far. 

 _Originally posted by @pipliggins in [#4496](https://github.com/pybamm-team/PyBaMM/issues/4496#issuecomment-2432607221)_

I added additional checking in the `ProcessedVariable` class to ignore empty sub-solutions.

Fixes #4496

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/src/pybamm/solvers/processed_variable.py
+++ b/src/pybamm/solvers/processed_variable.py
@@ -133,19 +133,24 @@ class ProcessedVariable:
         ys = self.all_ys
         yps = self.all_yps
         inputs = self.all_inputs_casadi
+
+        # Remove all empty ts
+        idxs_nonempty = np.where([ti.size > 0 for ti in ts])[0]
+
         # Find the indices of the time points to observe
         if full_range:
-            idxs = range(len(ts))
+            idxs = idxs_nonempty
         else:
-            idxs = _find_ts_indices(ts, t)
+            ts_nonempty = [ts[idx] for idx in idxs_nonempty]
+            idxs_subset = _find_ts_indices(ts_nonempty, t)
+            idxs = idxs_nonempty[idxs_subset]
 
-        if isinstance(idxs, list):
-            # Extract the time points and inputs
-            ts = [ts[idx] for idx in idxs]
-            ys = [ys[idx] for idx in idxs]
-            if self.hermite_interpolation:
-                yps = [yps[idx] for idx in idxs]
-            inputs = [self.all_inputs_casadi[idx] for idx in idxs]
+        # Extract the time points and inputs
+        ts = [ts[idx] for idx in idxs]
+        ys = [ys[idx] for idx in idxs]
+        if self.hermite_interpolation:
+            yps = [yps[idx] for idx in idxs]
+        inputs = [self.all_inputs_casadi[idx] for idx in idxs]
 
         is_f_contiguous = _is_f_contiguous(ys)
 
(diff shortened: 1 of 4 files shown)


src/pybamm/solvers/processed_variable.py
```diff
@@ -133,19 +133,24 @@ def _setup_cpp_inputs(self, t, full_range):
...
         yps = self.all_yps
         inputs = self.all_inputs_casadi
+
+        # Remove all empty ts
+        idxs_nonempty = np.where([ti.size > 0 for ti in ts])[0]
+
         # Find the indices of the time points to observe
         if full_range:
-            idxs = range(len(ts))
+            idxs = idxs_nonempty
```
martinjrobins, line 142: do we really need this branch of the if statement? why not call it `idx` to begin with?

martinjrobins (changes requested):
Thanks @MarcBerliner and @pipliggins. just one comment that will hopefully simplify the logic

src/pybamm/solvers/processed_variable.py
MarcBerliner: I think I was just looking to avoid redefining `idxs = idxs[idxs_subset]`, but the previous way is probably just more confusing

martinjrobins (approved):
Looks good 👍

PR #4544, opened 2024-10-28
Using `tempfile` for test_symbol_visualise to remove flaky test.

prady0t:
# Description

Using tempfile avoids test failures when running `assert os.path.exists()`


Saransh-cpp (commented):
Why not use the [tmp_path fixture](https://docs.pytest.org/en/stable/how-to/tmp_path.html#the-tmp-path-fixture)? Sorry, I might have missed something.

kratman (changes requested):
The windows failures are caused by the tempfile behavior being a bit different on Windows.

For a bunch of the tests, we do something like this:
```
with tempfile.NamedTemporaryFile(
            suffix=filename, delete=False, mode="w"
        ) as tmp:
```
where the delete=False is what makes it work on windows.

agriyakhetarpal:
Either the Windows fix or using the `tmp_path` fixture sounds good to me. Since the latter is more tied in with `pytest`, I would suggest trying it first.

prady0t:
Thanks for all the suggestions. I've updated the PR

agriyakhetarpal (approved):

PR #4562, opened 2024-11-04
Throw an error when concatenations have different number of children

### LINKED ISSUE #4561: [Bug]: Binary operators on concatenations assume that each concatenation has the same number of orphans
### PyBaMM Version

develop

### Python Version

3.12

### Describe the bug

If a binary operator, such as `+`, is invoked with 2 concatenations (i.e. `isinstance(left, pybamm.Concatenation) and isinstance(right, pybamm.Concatenation) == True`), and the concatenations have a different number of orphans, the "extra orphans" are dropped. This can be easily seen [here](https://github.com/aabills/PyBaMM/blob/8ce4a20090fee6bae2251d54fd28b4b0d49abdfb/src/pybamm/expression_tree/binary_operators.py#L859).

This is causing reaction heating to not be present in DFN half-cell models (see comments in #4557)

### Steps to Reproduce

no MWE yet, but should be self evident in the code.

### Relevant log output

_No response_

aabills:
# Description

Fixes #4561 

## Type of change

# Key checklist:

## Further checks:


aabills:
Tests are going to fail on this due to an unrelated issue that I am working on concurrently, so opening this for now and once we merge the other one we can merge this one.

aabills:
Hi all, this PR now disables thermal half cell models. Previously, thermal half cell models were giving incorrect results. I am working on a long-term fix for this problem, but for now, I think it's better to throw an error than give incorrect results. Specifically, reaction heating from both electrodes has been ignored for at least the last few PyBaMM releases (and likely all time).

src/pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -990,6 +990,16 @@ def options(self, extra_options):
                 f"must use surface formulation to solve {self!s} with hydrolysis"
             )
 
+        if isinstance(self, pybamm.lithium_ion.DFN):
+            if options["working electrode"] == "positive":
+                if (
+                    options["thermal"] == "lumped"
```
kratman, line 996: Should this be something like checking if thermal is defined at all for the half cell? This would make me think that non-lumped models still work

aabills: Other thermal models are disabled in the options checking (https://github.com/pybamm-team/PyBaMM/blob/0e0936a2ce27b9ce908958a33f79895f149ea232/src/pybamm/models/full_battery_models/base_battery_model.py#L608). This would be pretty unwieldy to do here because I am trying to only disable the DFN case, which I don't think I can check in the options checker.

kratman (dismissed):
Seems fine as a temporary fix for the bug

src/pybamm/models/submodels/thermal/base_thermal.py
```diff
@@ -88,6 +88,32 @@ def _get_standard_coupled_variables(self, variables):
...
+            T_p = variables["Positive electrode temperature [K]"]
+            tor_s = variables["Separator electrolyte transport efficiency"]
+            tor_p = variables["Positive electrolyte transport efficiency"]
+            i_e_s = (self.param.kappa_e(c_e_s, T_s) * tor_s) * (
+                self.param.chiRT_over_Fc(c_e_s, T_s) * grad_c_e_s - grad_phi_e_s
+            )
+            i_e_p = (self.param.kappa_e(c_e_p, T_p) * tor_p) * (
+                self.param.chiRT_over_Fc(c_e_p, T_p) * grad_c_e_p - grad_phi_e_p
+            )
+            i_e = pybamm.concatenation(i_e_s, i_e_p)
```
kratman, line 115: Could we move this to a function instead?

aabills: yep good call, done

kratman:
@aabills Can you update the bug fix section of the change log

src/pybamm/models/submodels/thermal/base_thermal.py
```diff
@@ -88,6 +88,12 @@ def _get_standard_coupled_variables(self, variables):
         # TODO: change full stefan-maxwell conductivity so that i_e is always
         # a Concatenation
         i_e = variables["Electrolyte current density [A.m-2]"]
+        # Special case for half cell -- i_e has to be a concatenation for this to work due to a mismatch with Q_ohm, so we make a new i_e which is a concatenation.
+        if (not isinstance(i_e, pybamm.Concatenation)) and (
+            self.options.electrode_types["negative"] == "planar"
+        ):
```
kratman, line 94: Just for clarification, what happens outside of the `self.options.electrode_types["negative"] == "planar"` case? Are there still issues with other electrodes or electrode types? If so is there an issue for it somewhere?

aabills: In that case, Q_rxn is a concatenation with domains that match Q_ohm, which means everything matches up and works ( I just double checked this).

kratman (dismissed):
Looks good to me now. I can't vouch for the physics, but I don't see any issues with the code.

aabills:
> @aabills Can you update the bug fix section of the change log

Done, I realized I haven't done that for the other PR's either so I did all 3.

CHANGELOG.md
```diff
@@ -21,6 +22,9 @@
 - Removed the `start_step_offset` setting and disabled minimum `dt` warnings for drive cycles with the (`IDAKLUSolver`). ([#4416](https://github.com/pybamm-team/PyBaMM/pull/4416))
 
 ## Bug Fixes
+- Added error for binary operators on two concatenations with different numbers of children. Previously, the extra children were dropped. ([#4562](https://github.com/pybamm-team/PyBaMM/pull/4562))
+- Fixed bug where Q_rxn was dropped from the total heating term in half-cell models. ([#4562](https://github.com/pybamm-team/PyBaMM/pull/4562))
```
kratman, line 26: Since these are the same PR, can we just combine the descriptions

kratman:
```suggestion
- Added error for binary operators on two concatenations with different numbers of children. Previously, the extra children were dropped. Also fixed bug where Q_rxn was dropped from the total heating term in half-cell models. ([#4562](https://github.com/pybamm-team/PyBaMM/pull/4562))
```

kratman (approved):

PR #4563, opened 2024-11-04
rewrite full conductivity submodel

### LINKED ISSUE #4324: [Bug]: Half cell negative reaction heating set to 0
### PyBaMM Version

24.5

### Python Version

3.12

### Describe the bug

For some reason reaction heating for the half-cell models is set to 0. As a result, the half-cell models are under calculating heat source (which is important for thermal modeling).

### Steps to Reproduce

Just look at the code: https://github.com/pybamm-team/PyBaMM/blob/fa68ddc0496853b2abf8ef4ef3cdbca88615befc/pybamm/models/submodels/thermal/base_thermal.py#L148 

### Relevant log output

_No response_

aabills:
# Description

Fixes #4324 
Enables #4562 

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/src/pybamm/models/submodels/electrolyte_conductivity/base_electrolyte_conductivity.py
+++ b/src/pybamm/models/submodels/electrolyte_conductivity/base_electrolyte_conductivity.py
@@ -94,9 +94,9 @@ class BaseElectrolyteConductivity(pybamm.BaseSubModel):
                 "separator",
                 "positive electrode",
             ]:
-                i_e_n, _, i_e_p = i_e.orphans
+                i_e_n, i_e_s, i_e_p = i_e.orphans
             elif self.options.whole_cell_domains == ["separator", "positive electrode"]:
-                _, i_e_p = i_e.orphans
+                i_e_s, i_e_p = i_e.orphans
                 i_e_n = None
 
             if i_e_n is not None:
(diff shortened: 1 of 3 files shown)


aabills:
Coverage could be affected here because there are catches for a specific case that this PR eliminates

valentinsulzer:
I'm slightly surprised this works without adding boundary conditions for the internal boundaries but is that already being done by another part of the code?

aabills:
Not explicitly. We do check for accuracy somewhere in tests right? I don't want to merge this without confirming that it is correct.

aabills:
oh yeah, there we go, failing integration tests big time

brosaplanella:
> I'm slightly surprised this works without adding boundary conditions for the internal boundaries but is that already being done by another part of the code?

IIRC this is done in the base class, but only for the concatenated variable (the `composite_conductivity.py` I have been tinkering with didn't have BCs either).

aabills:
> IIRC this is done in the base class, but only for the concatenated variable (the composite_conductivity.py I have been tinkering with didn't have BCs either).

It appears that the code automatically sets Neumann internal boundary conditions for concatenated variables which is not always correct (at least , so that will need to be corrected and I'm guessing will fix this: https://github.com/pybamm-team/PyBaMM/blob/f05cae2ecda531cde049ccfc69d470cdb845bf98/src/pybamm/discretisations/discretisation.py#L339

To handle this, I propose that we should have another dictionary, `internal_boundary_conditions`,  which we check during the code above and default to what the code currently does.

PR #4565, opened 2024-11-05
Cleanup formatting in error messages

kratman:
# Description

I noticed some of the longer f-strings printed out with large indents. This PR is to clean that up a bit. There are still some triple quoted strings that could be cleaned up, but these are the easy ones to catch.

# Key checklist:

## Further checks:

### DIFF
--- a/src/pybamm/discretisations/discretisation.py
+++ b/src/pybamm/discretisations/discretisation.py
@@ -500,7 +500,7 @@ class Discretisation:
 
         if domain != "current collector":
             raise pybamm.ModelError(
-                "Boundary conditions can only be applied on the tabs in the domain"
+                "Boundary conditions can only be applied on the tabs in the domain "
                 f"'current collector', but {symbol} has domain {domain}"
             )
         # Replace keys with "left" and "right" as appropriate for 1D meshes
--- a/src/pybamm/discretisations/discretisation.py
+++ b/src/pybamm/discretisations/discretisation.py
@@ -500,8 +500,8 @@ class Discretisation:
 
         if domain != "current collector":
             raise pybamm.ModelError(
-                f"""Boundary conditions can only be applied on the tabs in the domain
-            'current collector', but {symbol} has domain {domain}"""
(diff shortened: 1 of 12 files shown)


src/pybamm/discretisations/discretisation.py
```diff
@@ -500,8 +500,8 @@ def check_tab_conditions(self, symbol, bcs):
 
         if domain != "current collector":
             raise pybamm.ModelError(
-                f"""Boundary conditions can only be applied on the tabs in the domain
-            'current collector', but {symbol} has domain {domain}"""
+                "Boundary conditions can only be applied on the tabs in the domain"
```
kratman, line 503:
```suggestion
                "Boundary conditions can only be applied on the tabs in the domain "
```

src/pybamm/expression_tree/operations/convert_to_casadi.py
```diff
@@ -231,8 +231,6 @@ def _convert(self, symbol, t, y, y_dot, inputs):
 
         else:
             raise TypeError(
-                f"""
-                Cannot convert symbol of type '{type(symbol)}' to CasADi. Symbols must all be
-                'linear algebra' at this stage.
-                """
+                f"Cannot convert symbol of type '{type(symbol)}' to CasADi. Symbols must all be"
```
kratman, line 234:
```suggestion
                f"Cannot convert symbol of type '{type(symbol)}' to CasADi. Symbols must all be "
```

agriyakhetarpal:
Out of lazy curiosity, is there a pre-commit hook that can clean things like these?

kratman (commented):
Fixing two mistakes

kratman:
> Out of lazy curiosity, is there a pre-commit hook that can clean things like these?

Probably, these just annoyed me when I was debugging something and it only took me a couple minutes to do it. I know there are a few `"some text {} {}".replace()` and `"some text {} {}".format()` cases that are trickier to fix.

agriyakhetarpal (approved):

PR #4572, opened 2024-11-07
bug: use direct casadi bspline function for 1D & 2D cubic interp

### LINKED ISSUE #4570: bug: use casadi `Function.bspline` interface to avoid plugin system
We've had problems in the past with the casadi plugin system (see https://github.com/pybamm-team/PyBaMM/issues/3783). The solution in #3783 was to swap to using the direct interface to the casadi linear interpolation function, avoiding the plugin system. However we didn't find a solution to the cubic interpolant.

We might be able to avoid the plugin system for cubic interpolation using the casadi `Function.bspline` function (https://web.casadi.org/python-api/#function), going to try this out and see if this is the case.

martinjrobins:
# Description

Instead of using the casadi plugin system, use direct bspline function for 1D and 2D cubic interpolation.

I didn't implement 3D cubic interpolation because (a) scipy has no functions for bspline interpolation in 3D, and (b) we only test 3D linear interpolation so didn't think it was neccessary.

Fixes #4570

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/src/pybamm/expression_tree/operations/convert_to_casadi.py
+++ b/src/pybamm/expression_tree/operations/convert_to_casadi.py
@@ -7,6 +7,7 @@ import pybamm
 import casadi
 import numpy as np
 from scipy import special
+from scipy import interpolate
 
 
 class CasadiConverter:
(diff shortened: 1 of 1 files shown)


kratman:
@martinjrobins Thanks, I am testing it now before I review

PR #4582, opened 2024-11-12
Update baseline for CasADi 3.6.7 bump

agriyakhetarpal:
# Description

I updated https://github.com/pybamm-team/casadi-vcpkg-registry/ to build CasADi 3.6.7, and this PR updates the baseline to account for that change – so that we build our extension module against CasADi 3.6.7.

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ requires = [
     "setuptools>=64",
     "wheel",
     # On Windows, use the CasADi vcpkg registry and CMake bundled from MSVC
-    "casadi>=3.6.6; platform_system!='Windows'",
+    "casadi>=3.6.7; platform_system!='Windows'",
     # Note: the version of CasADi as a build-time dependency should be matched
     # across platforms, so updates to its minimum version here should be accompanied
     # by a version bump in https://github.com/pybamm-team/casadi-vcpkg-registry.
@@ -37,7 +37,7 @@ classifiers = [
 dependencies = [
     "numpy>=1.23.5,<2.0.0",
     "scipy>=1.11.4",
-    "casadi>=3.6.6",
+    "casadi>=3.6.7",
     "xarray>=2022.6.0",
     "anytree>=2.8.0",
     "sympy>=1.12",
--- a/vcpkg-configuration.json
+++ b/vcpkg-configuration.json
@@ -13,7 +13,7 @@
     {
       "kind": "git",
       "repository": "https://github.com/pybamm-team/casadi-vcpkg-registry.git",
-      "baseline": "1cb93f2fb71be26c874db724940ef8e604ee558e",
+      "baseline": "e4b797736790af90de505e0296b07e87719cb1a6",
       "packages": ["casadi"]
     }
   ]


agriyakhetarpal:
Triggered a build from this branch on my fork here: https://github.com/agriyakhetarpal/PyBaMM/actions/runs/11805727556. If this passes, we should be good to go. I've marked it as a draft until then.

@kratman, this should be included in the release

pyproject.toml
```diff
@@ -37,7 +37,7 @@ classifiers = [
 dependencies = [
     "numpy>=1.23.5,<2.0.0",
     "scipy>=1.11.4",
-    "casadi>=3.6.6",
+    "casadi>=3.6.7",
```
kratman, line 40: Since we are linking to the C++ code, shouldn't this be a hard pin at 3.6.7? I looked at this before, but seems like a good time to do it

agriyakhetarpal: Yes, it should be. The problem is that if CasADi 3.7.X is released and we don't update it later, the mismatch will break things at runtime across all platforms (because CasADi libs are not compatible across minor versions, as we noticed in #3100). So I don't recommend doing this right now, a better time would be when we switch the build system and move away from `vcpkg`. Not pinning like it is currently would break just Windows, and would allow macOS and Linux builds from source to work.

The proper solution is:
- either, pin CasADi for both build-time and runtime, for which `conda-forge` arguably offers a better method with `{{ pin_compatible }}`. I'd like to do that in a separate PR after carefully considering it because it would be a non-weak constraint for users.
- or, not use the CasADi Python package as a build-time dependency but build it as its own extension module (like we do for SuiteSparse and SUNDIALS).

agriyakhetarpal:
The Windows build and tests are passing, but I'm unsure (and haven't followed along) what's up with the Linux wheels – it looks like the solver crashes at runtime. https://github.com/agriyakhetarpal/PyBaMM/actions/runs/11805727556/job/32890103930. That failure is unrelated to this change, though.

kratman:
> The Windows build and tests are passing, but I'm unsure (and haven't followed along) what's up with the Linux wheels – it looks like the solver crashes at runtime. https://github.com/agriyakhetarpal/PyBaMM/actions/runs/11805727556/job/32890103930. That failure is unrelated to this change, though.

I triggered the same changes on my pybammsolvers repo to see if I get the same failures

agriyakhetarpal:
Looks like it's coming from here:

https://github.com/pybamm-team/PyBaMM/blob/cb503633c7c920c4d78d4ad62c60091539188ad4/src/pybamm/config.py#L84-L97

kratman:
> Looks like it's coming from here

Are you working on that one here, or should I start working on it so that it does not block the release

agriyakhetarpal:
> > Looks like it's coming from here
> 
> Are you working on that one here, or should I start working on it so that it does not block the release

Yes, I'm looking into it as we speak and I have a fix, but I'll test it in CI first and check if it works.

kratman:
Since the crash is unrelated, is anything left on this one?

agriyakhetarpal:
Nothing else, this is ready to go

kratman (approved):

PR #4583, opened 2024-11-13
Fix IDAKLU solver crash when running in multiple threads + when telemetry is prompted for

agriyakhetarpal:
# Description

This came up as an unrelated issue in #4582 when testing the wheels and apparently stems from #4441 where a thread was being used to check for the input. This caused the Linux wheel tests to fail: https://github.com/agriyakhetarpal/PyBaMM/actions/runs/11805727556/job/32890103930. This PR redoes that with a multi-platform approach and avoids the use of threads altogether by checking for different environments: Windows via `msvcrt` and Linux/macOS via `termios`. 

~A _possible_ corner case for Jupyter Notebooks (running in VS Code versus Jupyter Notebook/JupyterLab) has also been fixed.~ I've resorted to a more rudimentary fix using the previous method for now, since this fix was erroneous and did not work intermittently.

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -254,9 +254,10 @@ def run_tests(session):
     session.install("setuptools", silent=False)
     session.install("-e", ".[all,dev,jax]", silent=False)
     specific_test_files = session.posargs if session.posargs else []
-    session.run(
-        "python", "-m", "pytest", *specific_test_files, "-m", "unit or integration"
-    )
+    if specific_test_files:
+        session.run("python", "-m", "pytest", *specific_test_files)
+    else:
+        session.run("python", "-m", "pytest", "-m", "unit or integration")
 
 
 @nox.session(name="docs")
--- a/noxfile.py
+++ b/noxfile.py
@@ -253,11 +253,12 @@ def run_tests(session):
     set_environment_variables(PYBAMM_ENV, session=session)
     session.install("setuptools", silent=False)
     session.install("-e", ".[all,dev,jax]", silent=False)
-    specific_test_files = session.posargs if session.posargs else []
-    if specific_test_files:
-        session.run("python", "-m", "pytest", *specific_test_files)
-    else:
-        session.run("python", "-m", "pytest", "-m", "unit or integration")
+    session.run(
+        "python",
+        "-m",
+        "pytest",
+        *(session.posargs if session.posargs else ["-m", "unit or integration"]),
+    )
 
 
 @nox.session(name="docs")
--- a/src/pybamm/config.py
+++ b/src/pybamm/config.py
@@ -53,8 +53,6 @@ def get_input_or_timeout(timeout):  # pragma: no cover
     Cross-platform input with timeout, using various methods depending on the
     environment. Works in Jupyter notebooks, Windows, and Unix-like systems.
 
-    Returns:
-
     Args:
         timeout (float): Timeout in seconds
 
@@ -70,69 +68,27 @@ def get_input_or_timeout(timeout):  # pragma: no cover
     if not (sys.stdin.isatty() or is_notebook()):
         return None, True
 
-    # 1. special handling for Jupyter notebooks
+    # 1. Handling for Jupyter notebooks. This is a simplified
+    # implementation in comparison to widgets because notebooks
+    # are usually run interactively, and we don't need to worry
+    # about the event loop. The input is disabled when running
+    # tests due to the environment variable set.
     if is_notebook():  # pragma: no cover
         try:
-            from ipywidgets import widgets
-            from IPython.display import display, clear_output
+            from IPython.display import clear_output
 
-            # Create buttons for yes/no
-            yes_button = widgets.Button(description="Yes")
-            no_button = widgets.Button(description="No")
-            output = widgets.Output()
+            user_input = input("Do you want to enable telemetry? (Y/n): ")
+            clear_output()
 
-            # Variable to store the result
-            result = {"value": None, "set": False}
+            if user_input.lower() in ["y", "yes", ""]:
+                print("Telemetry enabled.")
+            elif user_input.lower() in ["n", "no"]:
+                print("Telemetry disabled.")
 
-            def on_yes_clicked(b):
-                with output:
-                    result["value"] = "yes"
-                    result["set"] = True
-                    clear_output()
-                    print("Telemetry enabled.")
-
-            def on_no_clicked(b):
-                with output:
-                    result["value"] = "no"
-                    result["set"] = True
-                    clear_output()
-                    print("Telemetry disabled.")
-
-            yes_button.on_click(on_yes_clicked)
-            no_button.on_click(on_no_clicked)
-
-            # Display the buttons
-            print("Do you want to enable telemetry?")
-            display(widgets.HBox([yes_button, no_button]))
-            display(output)
-
-            # Wait for button click or timeout
-            start_time = time.time()
-            while not result["set"] and (time.time() - start_time < timeout):
-                time.sleep(0.05)
-
-            if not result["set"]:
-                with output:
-                    clear_output()
-                    print(
-                        "Timeout reached or negative input. Defaulting to not enabling telemetry."
-                    )
-                return None, True
-
-            return result["value"], False
+            return user_input, False
 
         except Exception:  # pragma: no cover
-            # Fallback to regular input for Jupyter environments where widgets
-            # aren't available. This should be quite rare at this point but is
-            # included for completeness.
-            try:
-                from IPython.display import clear_output
-
-                user_input = input("Do you want to enable telemetry? (Y/n): ")
-                clear_output()
-                return user_input, False
-            except Exception:
-                return None, True
+            return None, True
 
     # 2. Windows-specific handling
     if sys.platform == "win32":
@@ -155,6 +111,7 @@ def get_input_or_timeout(timeout):  # pragma: no cover
             return None, True
         except Exception:
             return None, True
+
     # 3. POSIX-like systems will need to use termios
     else:  # pragma: no cover
         try:
--- a/src/pybamm/config.py
+++ b/src/pybamm/config.py
@@ -71,7 +71,7 @@ def get_input_or_timeout(timeout):  # pragma: no cover
         return None, True
 
     # 1. special handling for Jupyter notebooks
-    if is_notebook():
+    if is_notebook():  # pragma: no cover
         try:
             from ipywidgets import widgets
             from IPython.display import display, clear_output
@@ -114,12 +114,14 @@ def get_input_or_timeout(timeout):  # pragma: no cover
             if not result["set"]:
                 with output:
                     clear_output()
-                    print("Timeout reached or negative input. Defaulting to not enabling telemetry.")
+                    print(
+                        "Timeout reached or negative input. Defaulting to not enabling telemetry."
+                    )
                 return None, True
 
             return result["value"], False
 
-        except Exception:
+        except Exception:  # pragma: no cover
             # Fallback to regular input for Jupyter environments where widgets
             # aren't available. This should be quite rare at this point but is
             # included for completeness.
@@ -153,8 +155,8 @@ def get_input_or_timeout(timeout):  # pragma: no cover
             return None, True
         except Exception:
             return None, True
-    # POSIX-like systems will need to use termios
-    else:
+    # 3. POSIX-like systems will need to use termios
+    else:  # pragma: no cover
         try:
             import termios
             import tty
@@ -190,7 +192,7 @@ def get_input_or_timeout(timeout):  # pragma: no cover
                 sys.stdout.write("\n")
                 sys.stdout.flush()
 
-        except Exception:
+        except Exception:  # pragma: no cover
             return None, True
 
     return None, True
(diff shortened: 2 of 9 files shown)


noxfile.py
```diff
@@ -254,9 +254,10 @@ def run_tests(session):
     session.install("setuptools", silent=False)
     session.install("-e", ".[all,dev,jax]", silent=False)
     specific_test_files = session.posargs if session.posargs else []
```
kratman, line 256: couldn't this be simplified with this:
```suggestion
    specific_test_files = session.posargs if session.posargs else ["unit or integration"]
```

agriyakhetarpal:
This approach might seem overkill but this is what came to mind first and seems like the most robust way to me, so I'd like to test this a bit more – I've converted this to a draft for now. I triggered a wheel build here, let's see if it passes: https://github.com/agriyakhetarpal/PyBaMM/actions/runs/11808256997

I've confirmed the fix locally with a fresh PyBaMM installation (i.e., with the `pybamm/config.yml` file not present in my Application Support  folder) every time:

- inside a terminal
- in an IPython shell
- in a notebook running in JupyterLab
- in a notebook running in VS Code

across the following scenarios:
- telemetry disabled
- telemetry enabled (which should be the same thing)
- a tim out (no input was given)

noxfile.py
agriyakhetarpal: No, I don't think so – for that we'll need the `-m` flag, but we exclude that when invoking a test file directly

agriyakhetarpal:
The buttons aren't working in the notebooks, actually – I'm quite sure they were working when I was testing earlier, so I guess I made some additional changes which made them break. I'll debug later in the day, and I'll consider dropping the buttons in favour of a simple input in case I don't get them to work.

agriyakhetarpal:
I switched to a more basic implementation for Jupyter notebooks, and I've seen bug reports about how inputs don't work properly with JupyterLab <4, but I don't think (too) many people are using version 3 now, so we should be good.

agriyakhetarpal:
Additionally, I tested on Google Colab by installing in editable mode from my branch – we should be ready to go with this. Suggestions on cleaning up the code are welcome.

agriyakhetarpal:
Linux wheel builds passing: https://github.com/agriyakhetarpal/PyBaMM/actions/runs/11819747790/job/32930388878

kratman:
Ok I will take a close look at this in an hour or so

src/pybamm/config.py
```diff
@@ -45,20 +48,120 @@ def is_running_tests():  # pragma: no cover
...
+        timeout (float): Timeout in seconds
 
-    Returns
-    -------
-    bool
-        True if the user opts in, False otherwise.
+    Returns:
+        tuple: A tuple containing:
+            - str: The user input if received before the timeout, or None if the timeout was reached.
+            - bool: True if the timeout was reached, False otherwise.
```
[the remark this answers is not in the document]
agriyakhetarpal: > empty string and None evaluate to False in if statements and other strings would be True

Missed that, this is why I shouldn't write code in the middle of the night :)

src/pybamm/config.py
```diff
@@ -45,20 +48,125 @@ def is_running_tests():  # pragma: no cover
...
+        try:
+            from IPython.display import clear_output
+
+            user_input = input("Do you want to enable telemetry? (Y/n): ")
+            clear_output()
+
+            return user_input
+
+        except Exception:  # pragma: no cover
+            return None
```
kratman, line 91: Not a major issue, but this has no timeout so people notebooks will just hang if they don't realize this popped up

src/pybamm/config.py
```diff
@@ -45,20 +48,125 @@ def is_running_tests():  # pragma: no cover
...
+                tty.setraw(sys.stdin.fileno())
+
+                sys.stdout.write("Do you want to enable telemetry? (Y/n): ")
+                sys.stdout.flush()
+
+                input_chars = []
+                start_time = time.time()
+
+                while time.time() - start_time < timeout:
+                    rlist, _, _ = select.select([sys.stdin], [], [], 0.1)
```
kratman, line 135: If we are already using select, do we need the loop? The timeout isn't very long

src/pybamm/config.py
```diff
@@ -45,20 +48,125 @@ def is_running_tests():  # pragma: no cover
...
-        True if the user opts in, False otherwise.
+    # 3. POSIX-like systems will need to use termios
+    else:  # pragma: no cover
+        try:
+            import termios
+            import tty
+            import select
+
+            # Save terminal settings for later
+            old_settings = termios.tcgetattr(sys.stdin)
```
kratman, line 123: It seems excessive that we need to change the terminal settings for this

src/pybamm/config.py
```diff
@@ -45,20 +48,125 @@ def is_running_tests():  # pragma: no cover
...
+    """
+    Ask the user if they want to opt in to telemetry.
     """
+    # Check for telemetry disable flag first
+    if os.getenv("PYBAMM_DISABLE_TELEMETRY", "false").lower() != "false":
+        return False
+
+    # Skip telemetry prompt in non-interactive environments
+    if not (sys.stdin.isatty() or is_notebook()):
+        return False
```
kratman, line 168: These checks are also in the `get_input_or_timeout()` function, do we need to check that twice?

src/pybamm/config.py
```diff
@@ -69,44 +177,28 @@ def ask_user_opt_in(timeout=10):
...
+            return True
+        elif user_input.lower() in ["n", "no"]:
+            print("Telemetry disabled.")
             return False
+        else:
+            print("Invalid input. Please enter 'Y/y' for yes or 'n/N' for no:")
+            user_input = get_input_or_timeout(timeout)
+            if user_input is None:
+                print("\nTimeout reached. Defaulting to not enabling telemetry.")
+                return False
```
kratman, line 201: I know I brought it up before, but if we just check for the yes options, we can skip having an infinite loop

src/pybamm/config.py
```diff
@@ -69,44 +177,28 @@ def ask_user_opt_in(timeout=10):
...
-                print("\nTelemetry enabled.\n")
-                return True
-            elif answer[0] in ["no", "n"]:
-                print("\nTelemetry disabled.\n")
-                return False
-            else:
-                print("\nInvalid input. Please enter 'yes/y' for yes or 'no/n' for no.")
-        else:
-            print("\nTimeout reached. Defaulting to not enabling telemetry.")
+        elif user_input.lower() in ["y", "yes"]:
```
kratman, line 190: We should be able to simplify this
```suggestion
        if user_input.lower() in ["y", "yes", ""]:
```

src/pybamm/config.py
```diff
@@ -69,44 +177,28 @@ def ask_user_opt_in(timeout=10):
...
+    if user_input is None:
+        print("\nTimeout reached. Defaulting to not enabling telemetry.")
+        return False
 
     while True:
-        if time.time() - time_start > timeout:
-            print("\nTimeout reached. Defaulting to not enabling telemetry.")
+        if user_input == "":  # Empty input should mean a no
+            print("Telemetry enabled.\n")
             return False
```
kratman, line 189: Merged with the one below
```suggestion
```

src/pybamm/config.py
```diff
@@ -118,7 +210,7 @@ def generate():
         return
 
     # Ask the user if they want to opt in to telemetry
-    opt_in = ask_user_opt_in()
+    opt_in = ask_user_opt_in(timeout=10)
```
kratman, line 213: Default is already 10
```suggestion
    opt_in = ask_user_opt_in()
```

kratman (changes requested):

src/pybamm/config.py
agriyakhetarpal: Yes, this is intentional. I opine that notebooks are meant to be used interactively by default – I expect this to be a one-time issue, since importing PyBaMM and choosing "yes" or "no" once would mean that the config gets saved and this never comes up again (until the config file gets deleted, of course).

src/pybamm/config.py
```diff
@@ -45,20 +48,125 @@ def is_running_tests():  # pragma: no cover
...
+    """
+    Ask the user if they want to opt in to telemetry.
+    """
+    # Check for telemetry disable flag first
+    if os.getenv("PYBAMM_DISABLE_TELEMETRY", "false").lower() != "false":
+        return False
+
+    # Skip telemetry prompt in non-interactive environments
+    if not (sys.stdin.isatty() or is_notebook()):
+        return False
```
agriyakhetarpal, line 168:
```suggestion
```

src/pybamm/config.py
agriyakhetarpal: If a person presses "Enter" in haste, that enables telemetry for them (because of your suggestion that I applied above), which might not be what they want – I'm trying to reduce the chance of that happening

src/pybamm/config.py
agriyakhetarpal: Removed

src/pybamm/config.py
agriyakhetarpal: So, should I keep it as "yes" if Enter is pressed or "no"? I feel the latter is the better option

src/pybamm/config.py
agriyakhetarpal: What this allows you to do is that you don't get new indentation when you write to the terminal again. I wrapped this in a try-finally block based on the Python docs: https://docs.python.org/3/library/termios.html#example.

So, by saving the attributes, I'm trying to disable the fact that `sys.stdout.write("\n")` will not return the cursor to the leftmost position (because we are in raw mode and we need to do that manually instead – for which I want to avoid).

We end up with this, otherwise:

```
❯ python -c "import pybamm; print(pybamm.IDAKLUSolver())"
PyBaMM can collect usage data and send it to the PyBaMM team to help us improve the software.
We do not collect any sensitive information such as models, parameters, or simulation results - only information on which parts of the code are being used and how frequently.
This is entirely optional and does not impact the functionality of PyBaMM.
For more information, see https://docs.pybamm.org/en/latest/source/user_guide/index.html#telemetry
Do you want to enable telemetry? (Y/n): n

                                         Telemetry disabled.
                                                            <pybamm.solvers.idaklu_solver.IDAKLUSolver object at 0x102de3ad0>
                                                                                                                             %
```

and this would break one's terminal until one restarts it – a more robust solution could be to use `TCSAFLUSH` instead of `TCSADRAIN`, I'll switch to that.

agriyakhetarpal:
Continuing from my question above: though I pushed e4859dcff274d584ce31b31cb33e3217f302799f which fixes the failing test, it also changes the behaviour on pressing "Enter" to enable telemetry – which, to me, does not seem like good UX because it can cause an inadvertent "yes" from a user. It might be better to be explicit rather than implicit: either allow only a "yes" or a "no" in the response and ask them to enter their response again if it's empty, or let "Enter" disable telemetry rather than enabling it). I understand that it would be good for us to get a reasonable amount of telemetry data from as many users as we can, but we should also be conscientious with our approach. @valentinsulzer, do you have any thoughts? I might have missed this when reviewing #4441. That said, if prompts in other applications do regard pressing Enter/an empty input as a "yes", then this approach is completely fine with me.

kratman:
Crashes should be resolved. I could not reproduce any of the notebook issues

agriyakhetarpal:
I'd say that this PR could still be useful because it avoids using `threading` altogether for telemetry. Could you please re-review it? I was under the assumption that #4591 would go in after this PR went in.

kratman:
@agriyakhetarpal I built the wheel with develop and did some testing on Windows.

A few things I noticed:
1. The tests in CI and wheel build all pass: https://github.com/kratman/PyBaMM/pull/5
2. Some tests fail if I try to run the tests locally with the wheel, but the same tests fail when I repeat the process for the v24.9.0 release. The failing tests all appear to be from Jax
3. Forcing posthog to be active during tests does show posthog in some of the tracebacks, but they appear to be some of the same tests that I saw failing in (2). 
4. When I forced posthog to run, I also disabled the `pybamm.config.generate()` function, so any remaining problems are likely in the internals of posthog, not from the use of threading while generating the config.
5. The timeout for creating the config is likely too short. With the slow import of a fresh PyBaMM install, it is not hard to miss the window for a 10 second opt-in.

The fact that tests are failing locally on windows bothers me, but it looks like some of this was already present and not caught by all of our tests. It is also possible that the ~45 failing Jax tests are due to issues with my Windows setup since the tests pass in CI. I won't know for sure until I do additional testing for the release.

I will continue investigating the failures on windows, but I don't think removing the threading or adding extra logic to text output is needed at the moment. The changes to the installation guide, noxfile.py, and the pyproject.toml look good.

Next steps:
- Try to trigger crashes in notebooks and scripts instead of tests. This will rule out any threading issues related to `pytest`
- Compare my Windows dev environment to the CI dev environment to see if the failing Jax tests are from a solvable issue.

agriyakhetarpal:
Just a quick comment for now; I'll get back to the rest in a while: @kratman, if you can't reproduce the Windows crashes in CI, could you also check how old the machine that you are using locally is? Some JAX operations might crash or simply not work on very old machines because of the lack of AVX512 instructions, so maybe this is indeed just something that is for your setup and should be fine for slightly newer Windows computers.

kratman:
> Some JAX operations might crash or simply not work on very old machines because of the lack of AVX512 instructions, so maybe this is indeed just something that is for your setup and should be fine for slightly newer Windows computers.

This is not a particularly old machine. It is an i5-1235U (2022), which is the Alder Lake group of chips that do not have AVX-512, it does have AVX and AVX2 though

kratman:
Can we close this now? The threading issues have not reappeared for telemetry

agriyakhetarpal:
Thanks, sure, we can close. I'll keep my branch in case we ever need it.

PR #4585, opened 2024-11-13
Remove IREE

### LINKED ISSUE #4521: Jax-IREE not working on MacOS-13

kratman:
# Description

Removes the IREE solver. The IREE solver only supports single precision and is difficult to update. It might be re-added in the future when it is more stable.

Fixes #4521

# Key checklist:

## Further checks:


kratman:
I am expecting some stuff to fail and for this branch to require additional changes

tests/shared.py
```diff
@@ -336,7 +336,7 @@ def no_internet_connection():
         conn = socket.create_connection((host, 80), 2)
         conn.close()
         return False
-    except socket.gaierror:
+    except (socket.gaierror, TimeoutError):
```
kratman, line 339: Unrelated, but this popped up while I was testing. It looked like a random network error

kratman:
@jsbrittain Can you look at this too please?

.github/workflows/test_on_push.yml
```diff
@@ -7,7 +7,6 @@ on:
 env:
   FORCE_COLOR: 3
   PYBAMM_IDAKLU_EXPR_CASADI: ON
-  PYBAMM_IDAKLU_EXPR_IREE: ON
```
martinjrobins, line 10: Can we please leave in the IREE solver code and just turn off this flag so it is not tested in the CI? Its going to be a lot more difficult to re-enable this in the future if the code is deleted.

martinjrobins (changes requested):

.github/workflows/test_on_push.yml
kratman: I have two main concerns with disabling it this way:
- Turning it off in the tests would cause the code to be untested, and untested code tends to decay. It would not be guaranteed to work when re-enabled
- IREE requires a specific version of Jax to operate. This blocks us from updating the Jax version cleanly. For instance we need to update the Jax version to support 3.13, and that would mean installing a different version whenever you wanted to use IREE. This is possible, but adds a lot of extra complexity that would not be fully tested if it is disabled by default

martinjrobins: I don't like untested code either, but I don't see how deleting the code from the repository makes it easier to re-enable the IREE functionality later on? If anything it will make it more difficult. I agree that the IREE code will decay if untested, but its going to decay faster if it only exists in an old commit. I think we should remove it from any documentation and slap an "experimental, internal use only and not guarenteed to work" so that no users try to use it. But the IREE backend is still our best plan for a GPU-compatible backend for PyBaMM, so I'd like to keep it so that we can (a) test any new functionality that IREE implements for double precision, and (b) see what can be achieved numerically with only single precision. If we turn if off, it won't require any additional workload to keep it maintained going forward (other than by the devs actually working on the IREE solver).

docs/source/user_guide/installation/gnu-linux-mac.rst
```diff
@@ -101,21 +101,6 @@ Users can install ``jax`` and ``jaxlib`` to use the Jax solver.
 
 The ``pip install "pybamm[jax]"`` command automatically downloads and installs ``pybamm`` and the compatible versions of ``jax`` and ``jaxlib`` on your system.
```
kratman, line 103:
```suggestion

PyBaMM's full `conda-forge distribution <index.rst#installation>`_ (``pybamm``) includes ``jax`` and ``jaxlib`` by default.
```

docs/source/user_guide/installation/index.rst
```diff
@@ -35,8 +35,7 @@ Optional solvers
 
 The following solvers are optionally available:
 
-*  `jax <https://jax.readthedocs.io/en/latest/notebooks/quickstart.html>`_ -based solver, see :ref:`optional-jaxsolver` .
-*  `IREE <https://iree.dev/>`_ (`MLIR <https://mlir.llvm.org/>`_) support, see :ref:`optional-iree-mlir-support`.
+*  `jax <https://jax.readthedocs.io/en/latest/notebooks/quickstart.html>`_ -based solver, see `Optional - JaxSolver <gnu-linux-mac.html#optional-jaxsolver>`_.
```
kratman, line 38:
```suggestion
*  jax <https://jax.readthedocs.io/en/latest/notebooks/quickstart.html>`_ -based solver, see :ref:`optional-jaxsolver` .
```

docs/source/user_guide/installation/index.rst
```diff
@@ -35,8 +35,7 @@ Optional solvers
 
 The following solvers are optionally available:
 
-*  `jax <https://jax.readthedocs.io/en/latest/notebooks/quickstart.html>`_ -based solver, see :ref:`optional-jaxsolver` .
-*  `IREE <https://iree.dev/>`_ (`MLIR <https://mlir.llvm.org/>`_) support, see :ref:`optional-iree-mlir-support`.
+*  jax <https://jax.readthedocs.io/en/latest/notebooks/quickstart.html>`_ -based solver, see :ref:`optional-jaxsolver` .
```
kratman, line 38:
```suggestion
*  `jax <https://jax.readthedocs.io/en/latest/notebooks/quickstart.html>`_ -based solver, see :ref:`optional-jaxsolver` .
```

PR #4591, opened 2024-11-15
Fix telemetry issues in wheel build and unit tests

### LINKED ISSUE #4588: Fortnightly build for wheels failed
The build is failing with the following logs - https://github.com/pybamm-team/PyBaMM/actions/runs/11854441290

kratman:
# Description

Fixes the wheel build and updates the telemetry code to make sure it does not run during tests

Fixes #4588

## Type of change

# Key checklist:

## Further checks:


.github/workflows/publish_pypi.yml
```diff
@@ -116,6 +117,8 @@ jobs:
       - name: Build wheels on Linux
         run: pipx run cibuildwheel --output-dir wheelhouse
         env:
+          CIBW_ENVIRONMENT: >
+            PYBAMM_DISABLE_TELEMETRY="true"
```
kratman, line 121: As far as I can tell the environment variables were not making it into the CIBW environment. This disables telemetry in the wheel tests

.github/workflows/test_on_push.yml
```diff
@@ -132,7 +131,6 @@ jobs:
 
   # Skips IDAKLU module compilation for speedups, which is already tested in other jobs.
   run_doctests:
-    needs: style
```
kratman, line 135: If style fails then PR is updated with the pre-commit fixes and the old jobs get canceled. This just speeds up the start of the jobs a bit

noxfile.py
```diff
@@ -49,7 +49,6 @@ def set_iree_state():
     "IREE_INDEX_URL": os.getenv(
         "IREE_INDEX_URL", "https://iree.dev/pip-release-links.html"
     ),
-    "PYBAMM_DISABLE_TELEMETRY": "true",
```
kratman, line 52: As far as I could tell, this was not working. It seemed like telemetry was still running in tests.

src/pybamm/config.py
```diff
@@ -21,20 +29,18 @@ def is_running_tests():  # pragma: no cover
...
+        "CIRCLECI",
+        "JENKINS_URL",
+        "GITLAB_CI",
+    ]
     if any(var in os.environ for var in ci_env_vars):
         return True
 
-    # Check for common test runner names in command-line arguments
-    test_runners = ["pytest", "unittest", "nose", "trial", "nox", "tox"]
-    if any(runner in arg.lower() for arg in sys.argv for runner in test_runners):
```
kratman, line 35: This did not seem to work on Mac or Windows. I was seeing nox and pytest commands go right past this. I tested it by moving it to the top of the file and putting an assert below it.

src/pybamm/config.py
```diff
@@ -110,7 +116,7 @@ def get_input():  # pragma: no cover
 
 
 def generate():
-    if is_running_tests():
+    if is_running_tests() or check_opt_out():
```
kratman, line 119: Since `is_running_tests()` does not work in all CI cases, I put the opt-out check here to provide another escape

src/pybamm/telemetry.py
```diff
@@ -1,38 +1,45 @@
...
-        "python_version": sys.version,
-        "pybamm_version": pybamm.__version__,
-    }
+    if pybamm.config.check_opt_out():
+        disable()
+        return
 
     config = pybamm.config.read()
     if config:
-        if config["enable_telemetry"]:
```
kratman, line 36: This check is now in the `check_out_out()` function, so it is redundant here

tests/shared.py
```diff
@@ -336,12 +336,12 @@ def no_internet_connection():
         conn = socket.create_connection((host, 80), 2)
         conn.close()
         return False
-    except socket.gaierror:
+    except (socket.gaierror, TimeoutError):
```
kratman, line 339: Sometimes I get timeouts from the `no_internet_connection()` function. This makes that also count as "not connection", but since I was actively connected to the internet, I don't think this function is working as expected.

kratman (commented):
Explanations of the changes

kratman:
Coverage is easy to fix. I will do that before I mark it ready for review

kratman:
Wheel build passed, this should be ready for the release

valentinsulzer (approved):

PR #4599, opened 2024-11-21
Sync PR for release 24.11.0

kratman:
Syncing changelog updates and git history


valentinsulzer (approved):

kratman:
This will need to be merged instead of squashed. I will do that as soon as RTD is done

PR #4604, opened 2024-11-23
Sync Develop with 24.11.1

kratman:
Sync branch


kratman:
I will have to fix this another way due to the conflicts

Issue #4619, opened 2024-11-26
Make `calc_esoh` a model attribute


rtimms:
Currently the `Simulation` class uses the following logic to decide if it should calculate the state of health in a simulation 

```
    def _get_esoh_solver(self, calc_esoh):
        if (
            calc_esoh is False
            or isinstance(self._model, pybamm.lead_acid.BaseModel)
            or isinstance(self._model, pybamm.equivalent_circuit.Thevenin)
            or self._model.options["working electrode"] != "both"
        ):
            return None

        return pybamm.lithium_ion.ElectrodeSOHSolver(
            self._parameter_values, self._model.param, options=self._model.options
        )
```

It would be nice if `cacl_esoh` was a model attribute (`True` by default) so that when writing custom models you don't have to pass `calc_esoh=False` to the solver every time.

PR #4621, opened 2024-11-26
Summary variables calculated only when called

### LINKED ISSUE #4058: Calculate summary variables only when they are called
As discussed in #3995 

- create a `SummaryVariable` object that saves `solution.last_state` and uses it to calculate the esoh variables only when they are called (the same way as `Solution` calculates the variables only when they are called)
- allow users to pass in functions that calculate summary variables based on other variables

pipliggins:
# Description

Rather than calculating summary variables and storing them in every Solution class, a 'SummaryVariables' class is created which stores the first + last state of the solution, plus any relevant esoh parameters, so that the summary variables are only calculated and stored when they are called (similar to accessing standard variables). If an esoh variable is requested, they are all calculated and stored at once to reduce the number of calls to the esoh solver.

rather than viewing the list of summary variable names as
```
sol.summary_variables.keys()
```
you can now use
```
sol.summary_variables.all_variables
```
to view the names without calculating the variables.

Similarly, when plotting rather than getting the number of cycles as 
```
sol.summary_variables['Cycle number']
```
they are accessed as 
```
sol.summary_variables.cycle_number
```

Fixes #4058 

## Type of change

# Key checklist:

## Further checks:


src/pybamm/solvers/summary_variable.py
```diff
@@ -0,0 +1,175 @@
...
+        self.user_inputs = user_inputs or {}
+        self.esoh_solver = esoh_solver
+        self._variables = {}  # Store computed variables
+        self.cycle_number = None
+
+        model = solution.all_models[0]
+        self._possible_variables = model.summary_variables  # minus esoh variables
+        self._esoh_variables = []  # Store eSOH variable names
+
+        # Flag if eSOH calculations are needed
```
rtimms, line 44: related: https://github.com/pybamm-team/PyBaMM/issues/4619

src/pybamm/solvers/summary_variable.py
```diff
@@ -0,0 +1,182 @@
...
+            self.first_state = solution.first_state
+            self.last_state = solution.last_state
+            self.cycles = None
+
+    def _initialize_for_cycles(self, cycle_summary_variables: list[SummaryVariables]):
+        """Initialize attributes for when multiple cycles are provided."""
+        self.first_state = None
+        self.last_state = None
+        self.cycles = cycle_summary_variables
+        self.cycle_number = np.arange(1, len(self.cycles) + 1)
```
martinjrobins, line 66: I can see that at least `cycle_number` is used outside this class, can you please document the attributes that are meant to be public in the docstring?

martinjrobins (changes requested):
sorry for the delay in the review! This looks great, can you please add to the CHANGELOG under breaking changes, since this changes the api for the summary variables?

martinjrobins (approved):
looks good, thanks @pipliggins

PR #4634, opened 2024-12-02
Fix BPX bug when activation energies not provided

rtimms:
# Description
Fixes a bug where activation energies get set to `None` instead of defaulting to zero if they are not provided in the BPX file

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/src/pybamm/parameters/bpx.py
+++ b/src/pybamm/parameters/bpx.py
@@ -108,7 +108,7 @@ def _get_phase_names(domain):
         )
 
 
-def _bpx_to_param_dict(bpx: BPX) -> dict:
+def bpx_to_param_dict(bpx: BPX) -> dict:
     """
     Turns a BPX object in to a dictionary of parameters for PyBaMM
     """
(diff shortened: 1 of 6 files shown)


tests/unit/test_parameters/test_bpx.py
```diff
@@ -483,3 +483,19 @@ def test_bpx_user_defined(self):
...
+        ] = None
+        with tempfile.NamedTemporaryFile(
+            suffix="test.json", delete=False, mode="w"
+        ) as test_file:
+            json.dump(copy.copy(bpx_obj), test_file)
+            test_file.flush()
+            param = pybamm.ParameterValues.create_from_bpx(test_file.name)
+            assert (
+                param["Negative electrode diffusivity activation energy [J.mol-1]"]
+                == 0.0
```
kratman, line 500: Float equality is generally not a good thing to check against. Can we use an almost equal statement with a very small tolerance?

src/pybamm/parameters/bpx.py
```diff
@@ -222,6 +222,9 @@ def _bpx_to_param_dict(bpx: BPX) -> dict:
             domain.pre_name + "transport efficiency"
         ] ** (1.0 / 1.5)
 
+    def _get_activation_energy(var_name):
+        return pybamm_dict.get(var_name) or 0.0
```
kratman, line 226: @rtimms I moved this function above the comment about the functions requiring more than 1 variable. I also renamed the function parameter since it shadowed the name from outer scope

kratman (approved):
Looks good to me

PR #4635, opened 2024-12-02
BPX example

rtimms:
# Description

Add a BPX example in the example notebooks to replace [this repo](https://github.com/pybamm-team/bpx-example) which isn't tested or maintained. Includes changes from #4634 but made a separate PR since that one is a bug fix.

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:


docs/source/examples/notebooks/parameterization/nmc_pouch_cell_BPX.json
```diff
@@ -0,0 +1,86 @@
+{
```
rtimms, line 1: should this be moved to [pybamm-data](https://github.com/pybamm-team/pybamm-data)? how do you actually add things there?

kratman:
I made this a draft until #4634 is merged

docs/source/examples/notebooks/parameterization/nmc_pouch_cell_BPX.json
kratman: This is a pretty small file, so there is not a major issue with putting it in the repo. Having the data on pybamm-data will probably make it more compatible with the google-colab functionality in the documentation

src/pybamm/parameters/parameter_values.py
```diff
@@ -67,6 +67,7 @@ def __init__(self, values):
             for citation in self._dict_items["citations"]:
                 pybamm.citations.register(citation)
 
+    @staticmethod
     def _create_from_bpx(bpx, target_soc):
```
kratman, line 71: It looked like this should have been a static method when I was doing the merge and the tests are still passing, @rtimms can you double check this

docs/source/examples/notebooks/parameterization/bpx.ipynb
```diff
@@ -0,0 +1,335 @@
...
+     },
+     "execution_count": 5,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "# solve for 1 hour\n",
+    "sim.solve([0, 3600])\n",
+    "sim.plot()"
```
kratman, line 173:
```suggestion
    "_ = sim.plot()"
```

docs/source/examples/notebooks/parameterization/nmc_pouch_cell_BPX.json
agriyakhetarpal: > should this be moved to [pybamm-data](https://github.com/pybamm-team/pybamm-data)?

> This is a pretty small file, so there is not a major issue with putting it in the repo. Having the data on pybamm-data will probably make it more compatible with the google-colab functionality in the documentation

Yes, I think we should centralise everything under [`pybamm-data`](https://github.com/pybamm-team/pybamm-data/releases/tag/v1.0.0), so it would be nice to add this file there regardless of its small size (we do have a user-facing `DataLoader` class, after all, so it makes sense to showcase it more through more example notebooks).

> how do you actually add things there?

The procedure for adding things there is to first create a new v1.0.1 release and add all of the files from the previous release, add this file, and then bump the version in `pybamm_data.py`. We do this for reproducibility, such that the next PyBaMM release will retrieve the files from a different release URL (and this gives us the added benefit of better-segregated download analytics/usage stats from the GitHub API). If the changes to this file are final, I can do this later today if you'd like, @rtimms, or step back in case you may wish to do this yourself; please let me know!

rtimms: done, thanks!

kratman: @agriyakhetarpal Can you add documentation to the pybamm-data repo for how that is done?

docs/source/examples/notebooks/parameterization/bpx.ipynb
```diff
@@ -0,0 +1,342 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "data_loader = pybamm.DataLoader()\n",
+    "path_to_bpx_file = data_loader.get_data(\"nmc_pouch_cell_BPX.json\")"
```
kratman, line 49: We can make this one line so that it is easier for people to copy
```suggestion
    "path_to_bpx_file = pybamm.DataLoader().get_data(\"nmc_pouch_cell_BPX.json\")"
```

kratman (approved):

PR #4641, opened 2024-12-03
Add function for splitting units and add a test

### LINKED ISSUE #4608: [Bug]:  pybam.Experiment parse units error
### PyBaMM Version

24.9.0

### Python Version

python 3.10

### Describe the bug

in https://github.com/pybamm-team/PyBaMM/blob/develop/src/pybamm/experiment/step/base_step.py function def _convert_electric(value_string), if the value_string float value contains 'e' or 'E', the units incorrect parsed:

Python 3.12.3 (main, Nov  6 2024, 18:32:19) [GCC 13.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> value_string = "-1.1729285959021311e-05 W"
>>> unit = value_string.lstrip("0123456789.- ")
>>> unit
'e-05 W'
>>> unit = value_string.lstrip("0123456789eE.- ")
>>> unit
'W'
>>>

for python3, according to https://docs.python.org/3/library/functions.html#float, should accept 'e' or 'E' as well

### Steps to Reproduce

1.  value_string = "-1.1729285959021311e-05 W"
2.  unit = value_string.lstrip("0123456789.- ")

get incorrect unit 'e-05 W', it should be 'W'

### Relevant log output

```shell

```

kratman:
# Description

Small patch to handle exponents in units. There is probably more that can be done to improve the value/unit parsing, but this fixes the bug for now.

Fixes #4608

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/src/pybamm/experiment/step/base_step.py
+++ b/src/pybamm/experiment/step/base_step.py
@@ -1,6 +1,4 @@
-#
-# Private classes and functions for experiment steps
-#
+from __future__ import annotations
 import pybamm
 import numpy as np
 from datetime import datetime
@@ -492,6 +490,10 @@ _type_to_units = {
 }
 
 
+def get_unit_from(a_string: str) -> str:
+    return a_string.lstrip("0123456789.-eE ")
+
+
 def _convert_time_to_seconds(time_and_units):
     """Convert a time in seconds, minutes or hours to a time in seconds"""
     if time_and_units is None:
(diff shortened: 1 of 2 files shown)


src/pybamm/experiment/step/base_step.py
```diff
@@ -492,6 +490,10 @@ def set_up(self, new_model, new_parameter_values):
 }
 
 
+def get_unit_from(a_string: str) -> str:
+    return a_string.lstrip("0123456789.-eE ")
```
kratman, line 494: Adding "eE" is what  patches the bug, the function was split to make it easier to test

PR #4650, opened 2024-12-09
Modified `search` to take in multiple strings

### LINKED ISSUE #4629: Allow dictionary search to take multiple substrings
### Description

Currently, the `search` method for dictionaries (in `util.py`) only takes a string. I would be useful to be able to provide a list of strings so it checks for all of them.

### Motivation

It would allow for more specific searches. For example, let's say I want a given overpotential in the negative electrode but can't recall the name. I could search for "[V]" and "negative" to whittle down the list.

### Possible Implementation

For the `search` method it would be fairly straightforward, we just need to iterate over the list and check all strings are in it.

For the `get_best_matches` (which is used if the search is unsuccesful) it is less obvious how to fix it as, from what I understand, the `difflib.get_close_matches` method does not allow for multiple strings to check.

### Additional context

_No response_

medha-14:
# Description

Fixes #4629 

## Type of change

# Key checklist:

## Further checks:


medha-14:
I have modified the `search` method to accept multiple strings. In cases where an exact match is not found, I pass the concatenated string of the multiple inputs to the `get_close_matches` method.Will this approach suffice for what we are trying to achieve here ?

src/pybamm/util.py
```diff
@@ -102,41 +102,45 @@ def __getitem__(self, key):
...
+        dict_keys.sort()
         results = {}
 
-        # Check if any of the dict keys contain the key we are searching for
-        for k in keys:
-            if key in k.lower():
+        # Check if all search terms are present in each dictionary key
+        for k in dict_keys:
+            k_lower = k.lower()
+            if all(term in k_lower for term in keys_lower):
```
kratman, line 124:
```suggestion
        for k in known_keys:
            k_lower = k.lower()
            if all(term in k_lower for term in search_keys):
```

src/pybamm/util.py
```diff
@@ -102,41 +102,45 @@ def __getitem__(self, key):
...
                 results[k] = self[k]
 
         if results == {}:
-            # If no results, return best matches
-            best_matches = self.get_best_matches(key)
+            # No results: Suggest best matches using the concatenated string
+            concatenated_keys = " ".join(keys_lower)
+            best_matches = self.get_best_matches(concatenated_keys)
```
kratman, line 130:
```suggestion
            best_matches = self.get_best_matches(" ".join(search_keys))
```

kratman (changes requested):
I did not test the code changes locally, but they were just renamings.

Can you add a test with multiple keys in the search? It looks like you only fixed the test for the output formatting

src/pybamm/util.py
```diff
@@ -102,41 +102,45 @@ def __getitem__(self, key):
...
         else:
-            # Just print keys
-            print("\n".join(f"{k}" for k in results.keys()))
+            # Print only keys of matching results
+            print("\n".join(results.keys()))
 
-    def copy(self):
-        return FuzzyDict(super().copy())
+        def copy(self):
+            return FuzzyDict(super().copy())
```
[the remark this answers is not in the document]
medha-14: This was a mistake made by me. Thank you for pointing this out.

medha-14:
I have implemented the suggested changes and added tests for searching multiple strings as well.

brosaplanella (commented):
Looks good to me, just needs an entry to the CHANGELOG before merging

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Features
 - Added two more submodels (options) for the SEI: Lars von Kolzenberg (2020) model and Tunneling Limit model ([#4394](https://github.com/pybamm-team/PyBaMM/pull/4394))
+- Enhanced the `search` method to accept multiple search terms in the form of a string or a list. ([#4650](https://github.com/pybamm-team/PyBaMM/pull/4650))
```
agriyakhetarpal, line 5: Could you please move this to the top and add a line after "## Features" (we sort PRs in descending order)?

src/pybamm/util.py
```diff
@@ -102,38 +102,43 @@ def __getitem__(self, key):
...
-    def search(self, key, print_values=False):
+    def search(self, keys, print_values=False):
         """
-        Search dictionary for keys containing 'key'. If print_values is True, then
-        both the keys and values will be printed. Otherwise just the values will
-        be printed. If no results are found, the best matches are printed.
+        Search dictionary for keys containing all terms in 'keys'.
+        If print_values is True, both the keys and values will be printed.
+        Otherwise, just the keys will be printed. If no results are found,
+        the best matches are printed.
```
agriyakhetarpal, line 110: Let's add the inputs to the docstring as well:

```suggestion
        Search dictionary for keys containing all terms in 'keys'.
        If print_values is True, both the keys and values will be printed.
        Otherwise, just the keys will be printed. If no results are found,
        the best matches are printed.

        Parameters
        ----------
        keys : str or list of str
            Search term(s)

        print_values : bool, optional
            If True, print both keys and values. Otherwise, print only keys.
            Default is False.
```

agriyakhetarpal (commented):
Thanks, @medha-14! Happy to approve/merge after these suggestions.

agriyakhetarpal:
Actually, in the case of a partial match, would it be better to indicate that it is such? For example,

```python
model.variables.search(["NotAVariable", "concentration"])
```
now returns

```
No results for search using '['NotAVariable', 'concentration']'.
Best matches are ['Electrolyte concentration']
```

but we could say something like

```
Partial match for key 'concentration' in search using keys '['NotAVariable', 'concentration']'.
Best matches are ['Electrolyte concentration']
```

because we do have a match here for `"concentration"`, but not for `"NotAVariable"`.

medha-14:
In cases where some keys have an exact match while others only have a best match, what should the expected result be?

agriyakhetarpal:
> In cases where some keys have an exact match while others only have a best match, what should the expected result be? Should we only prioritize the exact matches in such cases or should we also have the best matches printed separately?

Do you mean the case where we have an exact match for $m$ keys and a partial/best match for $n - m$ keys? Yes, we should print both the exact matches and the best matches according to the keys. Could you share an example?

If I understood your question correctly, then an input as follows:

```python
model.variables.search(["Electrolyte concentration", "Electrolite concentration"])
```

should return something, in my opinion, like:
```
Results matched against 'Electrolyte concentration' in search:
Electrolyte concentration

Partial match for 'Electrolite concentration' in search:
Best matches are ['Electrolyte concentration', 'Electrode potential']
```

We can figure out the best way to display the output later. There is also a case to be made to say that this improvement to the search functionality to accept multiple strings means that the result is returned for only the string that does return a match, but we are not really implementing a search engine, so I feel it is acceptable to have all results for all input strings (as if we are looping over them in the search). @brosaplanella, what do you think?

medha-14:
For now i have modified the method to search for exact matches having all the `search_keys` if no such matches are found it gives search results for each term individually.

``` python
model.variables.search(["electrolyte", "concentration"])
```
Since both terms are present together in a single key, the result will be:
```python
 Results for 'Electrolyte concentration': ['Electrolyte concentration']
```

For the cases where there are no such matches it will iterate over each string individually and give results as such:
```python
model.variables.search(["RandomKey", "elecrtolyte concentration","electrolite"])
```
will give results as:
```pyhton
No matches found for 'RandomKey'.
Exact matches for 'electrolyte concentration': ['Electrolyte concentration [Molar]', 'Electrolyte concentration [mol.m-3]']
No exact matches found for 'electrolite'. Best matches are: ['Electrolyte potential [V]']

**Review by agriyakhetarpal — COMMENTED** [review]
Almost there. It's looking good, but I tested with a few adversarial inputs – we should raise a helpful error where applicable for them, and an associated test case that catches it as well:

For example, for:

```python
model.variables = {
    "Concentration [mol.m-3]": 0,
    "Surface concentration [mol.m-3]": 1,
    "Flux [mol.m-2.s-1]": 2,
}
```

`model.variables.search([""])` returns `Results for '': ['Concentration [mol.m-3]', 'Flux [mol.m-2.s-1]', 'Surface concentration [mol.m-3]']`, but it should ask the user to input a non-empty string instead.

Another case, here I tried the following (with the same `model.variables`):

```
In [4]: model.variables.search(["abcd", "concentr"])
No matches found for 'abcd'.
Exact matches for 'concentr': ['Concentration [mol.m-3]', 'Surface concentration [mol.m-3]']

In [5]: model.variables.search(["abcd", "concent"])
No matches found for 'abcd'.
Exact matches for 'concent': ['Concentration [mol.m-3]', 'Surface concentration [mol.m-3]']

In [6]: model.variables.search(["abcd", "concen"])
No matches found for 'abcd'.
Exact matches for 'concen': ['Concentration [mol.m-3]', 'Surface concentration [mol.m-3]']

In [7]: model.variables.search(["abcd", "conce"])
No matches found for 'abcd'.
Exact matches for 'conce': ['Concentration [mol.m-3]', 'Surface concentration [mol.m-3]']

In [8]: model.variables.search(["abcd", "conc"])
No matches found for 'abcd'.
Exact matches for 'conc': ['Concentration [mol.m-3]', 'Surface concentration [mol.m-3]']

In [9]: model.variables.search(["abcd", "con"])
No matches found for 'abcd'.
Exact matches for 'con': ['Concentration [mol.m-3]', 'Surface concentration [mol.m-3]']

In [10]: model.variables.search(["abcd", "co"])
No matches found for 'abcd'.
Exact matches for 'co': ['Concentration [mol.m-3]', 'Surface concentration [mol.m-3]']

In [11]: model.variables.search(["abcd", "c"])
No matches found for 'abcd'.
Exact matches for 'c': ['Concentration [mol.m-3]', 'Surface concentration [mol.m-3]']
```

and the last few of these don't really make a lot of sense. However, this case is wrong on the `main` branch as well, so it probably doesn't need to be addressed in this PR itself. The best approach would be to use `difflib` to determine what part of the search string corresponds to at least a significant value (maybe 50%?) of the search results. Please feel free to take this up in a follow-up PR if you'd like to.

So, I'm happy to approve once we manage the empty string case (even that is currently failing on the `main` branch, but it's quite easy to handle). Thanks for your work!

medha-14:
Thank you for the detailed review! I’ve modified the method to handle the empty string case . I’ve also added a tests to cover the same.Regarding the issue with partial matches and using `difflib` to refine the search logic, I’ll be taking that up in a separate PR.

src/pybamm/util.py
```diff
@@ -123,10 +123,18 @@ def search(self, keys, print_values=False):
...
+        if isinstance(keys, list) and all(not str(k).strip() for k in keys):
+            print("Error: Please provide a non-empty search term in the list.")
+            return
+
         if isinstance(keys, str):
-            search_keys = [keys]
+            search_keys = [keys.strip()]
         else:
-            search_keys = keys.copy()
+            search_keys = [k.strip() for k in keys if k.strip()]
```
agriyakhetarpal: We should perform these operations after creating a copy, i.e., leave `original_keys` intact, and then move these to line 140 and after.

agriyakhetarpal (commented):
Thanks! Looks like lines 167–168 show up as not covered in the CodeCov annotations, could you please take another look?

agriyakhetarpal (dismissed):
Thanks, @medha-14! This looks like it is ready to go now.

src/pybamm/util.py
```diff
@@ -108,38 +108,77 @@ def __getitem__(self, key):
...
+                partial_matches = difflib.get_close_matches(
+                    search_key, known_keys, n=5, cutoff=0.5
+                )
+
+                if partial_matches:
+                    print(
+                        f"No exact matches found for '{original_key}'. Best matches are: {partial_matches}"
+                    )
+                else:
+                    print(f"No matches found for '{original_key}'.")
```
arjxn-py, line 181: I think, you can condense this part in a method - 
```python
def _find_matches(self, search_key, known_keys):
    exact = [key for key in known_keys if search_key in key.lower()]
    partial = difflib.get_close_matches(search_key, known_keys, n=5, cutoff=0.5)
    return exact, partial
```

then maybe use it as - 
```python
exact_key_matches, partial_matches = self._find_matches(search_key, known_keys)
```

src/pybamm/util.py
```diff
@@ -108,38 +108,77 @@ def __getitem__(self, key):
...
         """
-        key_in = key
-        key = key_in.lower()
 
-        # Sort the keys so results are stored in alphabetical order
-        keys = list(self.keys())
-        keys.sort()
-        results = {}
+        if not keys or (isinstance(keys, str) and not keys.strip()):
+            print("Error: Please provide a non-empty search term.")
```
arjxn-py, line 128: Question: How about we raise actual `ValueError` here? Since we're anyways exiting after this

src/pybamm/util.py
```diff
@@ -108,38 +108,77 @@ def __getitem__(self, key):
...
+            return
+        if isinstance(keys, list) and all(not str(k).strip() for k in keys):
+            print("Error: Please provide a non-empty search term in the list.")
```
arjxn-py, line 131: Ditto

arjxn-py (approved):
Awesome Work @medha-14, just a small question and a comment for your help. Otherwise it looks good :)

src/pybamm/util.py
agriyakhetarpal: I skipped it earlier because we didn't have a return value in the function. On second thought, yes, raising an error instead of `print("Error: <...>")` sounds good to me.

Saransh-cpp: This should be a `TypeError`. The wording can also be better -
```py
err = f"keys must be a list or list of strings; got {type(keys)}"
raise TypeError(err)
```

Saransh-cpp (changes requested):

src/pybamm/util.py
agriyakhetarpal: Ah yes, thanks for catching that! The test also needs to be updated accordingly to catch this error.

src/pybamm/util.py
```diff
@@ -108,38 +108,100 @@ def __getitem__(self, key):
...
+            The term to search for in the keys.
+        known_keys : list of str
+            The list of known dictionary keys to search within.
+
+        """
+        exact = [key for key in known_keys if search_key in key.lower()]
+        partial = difflib.get_close_matches(search_key, known_keys, n=5, cutoff=0.5)
+        return exact, partial
+
+    def search(self, keys, print_values=False):
```
agriyakhetarpal, line 127: Also, let's add annotations so that we can aid type checkers later on:

```suggestion
    def search(self, keys: str | list[str], print_values: bool = False):
```

and you can add `from __future__ import annotations` at the top of this file.

agriyakhetarpal (dismissed):

agriyakhetarpal (approved):

src/pybamm/util.py
```diff
@@ -108,38 +109,100 @@ def __getitem__(self, key):
...
+            if all(not str(k).strip() for k in keys):
+                raise ValueError(
+                    "The 'keys' list cannot contain only empty or whitespace strings."
+                )
+
+            if not all(isinstance(k, str) for k in keys):
+                raise TypeError("All elements in the 'keys' list must be strings.")
+
+            original_keys = keys
+            search_keys = [k.strip().lower() for k in keys if k.strip()]
```
Saransh-cpp, line 167: Why the `if` condition here?

src/pybamm/util.py
```diff
@@ -108,38 +109,100 @@ def __getitem__(self, key):
...
+                if print_values:
+                    for match in exact_key_matches:
+                        print(f"{match} -> {self[match]}")
+            else:
+                if partial_matches:
+                    print(
+                        f"No exact matches found for '{original_key}'. Best matches are: {partial_matches}"
+                    )
+                else:
+                    print(f"No matches found for '{original_key}'.")
```
Saransh-cpp, line 205:
```suggestion
                    print(f"No matches found for '{original_key}'")
```

src/pybamm/util.py
```diff
@@ -108,38 +109,100 @@ def __getitem__(self, key):
...
+            )
 
-        # Check if any of the dict keys contain the key we are searching for
-        for k in keys:
-            if key in k.lower():
-                results[k] = self[k]
+        if isinstance(keys, str):
+            if not keys.strip():
+                raise ValueError(
+                    "The search term cannot be an empty or whitespace-only string."
```
Saransh-cpp, line 152: To make everything consistent
```suggestion
                    "The search term cannot be an empty or whitespace-only string"
```

src/pybamm/util.py
```diff
@@ -108,38 +109,100 @@ def __getitem__(self, key):
...
+            original_keys = [keys]
+            search_keys = [keys.strip().lower()]
 
-        if results == {}:
-            # If no results, return best matches
-            best_matches = self.get_best_matches(key)
+        elif isinstance(keys, list):
+            if all(not str(k).strip() for k in keys):
+                raise ValueError(
+                    "The 'keys' list cannot contain only empty or whitespace strings."
```
Saransh-cpp, line 160:
```suggestion
                    "The 'keys' list cannot contain only empty or whitespace strings"
```

src/pybamm/util.py
```diff
@@ -108,38 +109,100 @@ def __getitem__(self, key):
                 f"'{key}' not found. Best matches are {best_matches}"
             ) from error
 
-    def search(self, key, print_values=False):
+    def _find_matches(self, search_key, known_keys):
```
Saransh-cpp, line 112: While we are at it :)
```suggestion
    def _find_matches(self, search_key: str, known_keys: list[str]):
```

src/pybamm/util.py
```diff
@@ -108,38 +109,100 @@ def __getitem__(self, key):
...
-        key = key_in.lower()
 
-        # Sort the keys so results are stored in alphabetical order
-        keys = list(self.keys())
-        keys.sort()
-        results = {}
+        if not isinstance(keys, (str, list)):
+            raise TypeError(
+                f" 'keys' must be a string or a list of strings, got {type(keys)}"
+            )
```
Saransh-cpp, line 147: I would suggest doing this for all the error messages. See [flake8-errmsg's explanation](https://github.com/henryiii/flake8-errmsg?tab=readme-ov-file#intro) for more context.
```suggestion
            msg = f"'keys' must be a string or a list of strings, got {type(keys)}"
            raise TypeError(msg)
```

Saransh-cpp (approved):
Thanks, @medha-14! This looks amazing!

Most of the suggestions below are "good practices" and they should be applied to the PR. Approving this as I am on holiday from tomorrow, and I don't want my review to block the merge :)

kratman:
@medha-14 Thanks for working on this, I will re-review after you finish @Saransh-cpp's suggestions

src/pybamm/util.py
medha-14: This condition was added to exclude empty strings or strings consisting only of whitespaces from `search_keys`, making the search more efficient and focused on the  meaningful terms only.

Saransh-cpp: Sorry, I should have added more details in -

You can either do -
```py
[k.strip().lower() for k in keys if k.strip()]
```
which skips empty/whitespace strings but also calls `strip()` in an `if` condition for each element of the list (but we save `lower()` calls as it is not called on empty/whitespace strings).


or -
```py
[k.strip().lower() for k in keys]
```
which calls `lower()` on every string (but we save calling `strip()` in an `if` condition for each element of the list).

I hope this makes sense. Both the expressions will give you the same result, but both the expressions call `strip()` and `lower()` on different number of elements. My questions is, which one of these is faster? If you have already experimented with the speed then I am okay with having the `if-strip()` implementation (could you please attach the results here), but if you haven't, then you should generate a large dummy string dataset (not necessarily pybamm oriented, maybe [Faker](https://faker.readthedocs.io/en/master/) can help) and [`timeit`](https://docs.python.org/3/library/timeit.html) both the implementations.

Remember - ["Our mental image of what code should be faster or slower is often wrong, or doesn't make much difference. The only way to really improve code performance is empirically, through measurements."](https://github-pages.ucl.ac.uk/rsd-engineeringcourse/ch08performance/015mandels.html)

I would have experimented with this but I am low on time for the next few weeks 😬 

Thank you again for taking this up!

agriyakhetarpal: > My questions is, which one of these is faster?

I don't have a benchmark to corroborate either, but I am sure the second one, i.e.,

```python
[k.strip().lower() for k in keys]
```

would be faster – because the first one calls `strip()` twice for non-empty strings, once per condition and once per comprehension. For the second one, the `lower()` operation on empty/whitespace strings is very fast since there's nothing to convert.

src/pybamm/util.py
```diff
@@ -108,38 +109,98 @@ def __getitem__(self, key):
...
+                    "The search term cannot be an empty or whitespace-only string"
+                )
+            original_keys = [keys]
+            search_keys = [keys.strip().lower()]
+
+        elif isinstance(keys, list):
+            if all(not str(k).strip() for k in keys):
+                raise ValueError(
+                    "The 'keys' list cannot contain only empty or whitespace strings"
+                )
```
agriyakhetarpal, line 162:
```suggestion
                msg = "The 'keys' list cannot contain only empty or whitespace strings"
                raise ValueError(msg)
```

src/pybamm/util.py
agriyakhetarpal: > My question is, which one of these is faster?

I don't have a benchmark to corroborate either, but I am sure the second one, i.e.,

```python
[k.strip().lower() for k in keys]
```

would be faster – because the first one calls `strip()` twice for non-empty strings, once per condition and once per comprehension. For the second one, the `lower()` operation on empty/whitespace strings is very fast since there's nothing to convert.

agriyakhetarpal (dismissed):

src/pybamm/util.py
Saransh-cpp: That was my initial thought too 👀

medha-14: Thank you for discussing this ,I understand the trade-offs now. Though I haven't benchmarked the two implementations yet,  I will work on it and share the results so we can proceed accordingly.

src/pybamm/util.py
medha-14: I considered making this change earlier but assumed that this format of writing error messages was preferred for f-strings. Thanks for pointing it out! I'll implement this now to maintain consistency in the code.

src/pybamm/util.py
medha-14: I worked on the benchmark 

- taking a set of 5000 random strings
- number of white spaces being a random number between 1 and 50.
- number of words in the keys also being random number between 1 and 50.
- shuffling the words and the whitespaces.
- getting the avergae result for 10,000 such cases

And this was the result:

```python
  [k.strip().lower() for k in keys if k.strip()] 
  ```

Screenshot from 2024-12-19 19-15-11

Screenshot from 2024-12-19 19-14-53

```python 
 [k.strip().lower() for k in keys] 
 ```

Screenshot from 2024-12-19 19-15-03

Screenshot from 2024-12-19 19-14-42

Comment (agriyakhetarpal) [reply src/pybamm/util.py:165]: Interesting! I seem to have the opposite result for random data containing a random sample of arbitrary-length strings of ASCII characters and non-empty whitespace strings. Here are the results from a script I wrote (10,000 iterations for each size):

```
Benchmark Results:
--------------------------------------------------
Method 1 is  [k.strip().lower() for k in keys if k.strip()]
Method 2 is  [k.strip().lower() for k in keys]
Size       Method 1 (ms)   Method 2 (ms)   Difference
--------------------------------------------------
100        0.006           0.004           Method 2 is 35.6% faster
1000       0.061           0.039           Method 2 is 56.1% faster
10000      0.591           0.442           Method 2 is 33.8% faster
100000     6.464           4.882           Method 2 is 32.4% faster
```

medha-14: In this case, the assumption is that all words in the dataset have whitespaces to strip, which explains the performance differences observed. However, in my benchmark, I considered a scenario where only a random subset (around 50%) of the words have whitespaces, while the rest are plain words without any whitespaces. I also included empty strings, which are filtered out by the condition in the code, as originally intended, before being included in the final search keys.

agriyakhetarpal: Ah, I see. I tried it again by adding empty strings in my benchmarking script – so that it contains 20% empty strings, 20% strings with whitespaces, and 60% of them with potential padding, in no particular order for higher entropy, and the second one was still faster.

I would say that this experiment segues into a separate discussion, which is not in line with real-life usage of the functionality – as we can notice, both methods are quite fast (and the difference is unnoticeable) for search queries that are not too large (no one would search for a thousand terms at once).

What this boils down to is that while calling `lower()` is probably slow (and redundant) on empty strings, and we are raising an error on their inclusion in the search term, there is a possibility of non-empty strings in the search keys being imperfect, such as `"Electrolyte concentration "` (note the extra whitespace at the end), where stripping is arguably good to do so that our search terms are refined. So, I wouldn't worry about speed, and I agree with your assessment in https://github.com/pybamm-team/PyBaMM/pull/4650#discussion_r1891170142 – we can keep this as is.

medha-14: Thank you for the detailed insights! I agree that the performance difference in both cases is negligible, especially for practical use cases. Let’s proceed with the if condition, as it is now.

agriyakhetarpal (approved):

Saransh-cpp (approved):

kratman (approved):

kratman:
I will merge this after the tests pass

PR #4652, opened 2024-12-09
Drop dependence on `wheel` at build time

agriyakhetarpal:
# Description

From https://github.com/conda-forge/pybamm-feedstock/pull/70. This allows us to simplify a few things, though it is only temporary until the time we migrate from the current system.

cc: @Saransh-cpp

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,6 @@
 [build-system]
 requires = [
-    "setuptools>=64",
-    "wheel",
+    "setuptools>=70.1.1",
     # On Windows, use the CasADi vcpkg registry and CMake bundled from MSVC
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -1,7 +1,6 @@
 [build-system]
 requires = [
-    "setuptools>=64",
-    "wheel",
+    "setuptools>=70.1.1",
```
agriyakhetarpal, line 3: See https://setuptools.pypa.io/en/stable/history.html#id105

Saransh-cpp (approved):
Thanks, @agriyakhetarpal! I will update this on conda when a PyPI release is made.

PR #4685, opened 2024-12-17
Temporary pin for BPX

kratman:
# Description

There are some breaking changes for BPX coming, so I am pinning this version until they are sorted out. It still may be a problem in the releases since the pin was a minimum before this change.

## Type of change

# Key checklist:

## Further checks:


kratman:
This will be improved next release

PR #4689, opened 2024-12-18
Adding list of missing dependencies to docs

### LINKED ISSUE #4631: Review dependencies in documentation
I can see that a few PyBaMM dependencies are missing from - https://docs.pybamm.org/en/stable/source/user_guide/installation/index.html.

The dependencies in `pyproject.toml` and documentation should match.

prady0t:
# Description

Adding the missing dependencies to docs.

Fixes #4631

### DIFF
--- a/docs/source/user_guide/installation/index.rst
+++ b/docs/source/user_guide/installation/index.rst
@@ -63,6 +63,8 @@ Package                                                             Minimum supp
 `pandas <https://pypi.org/project/pandas/>`__                       1.5.0
 `pooch <https://www.fatiando.org/pooch/>`__                         1.8.1
 `posthog <https://posthog.com/>`__                                  3.6.5
+`pyyaml <https://pyyaml.org/>`__
+`platformdirs <https://platformdirs.readthedocs.io/en/latest/>`__
 =================================================================== ==========================
 
 .. _install.optional_dependencies:
@@ -102,6 +104,7 @@ Installable with ``pip install "pybamm[docs]"``
 Dependency                                                                                        Minimum Version    pip extra          Notes
 ================================================================================================= ================== ================== =======================================================================
 `sphinx <https://www.sphinx-doc.org/en/master/>`__                                                \-                 docs               Sphinx makes it easy to create intelligent and beautiful documentation.
+`sphinx_rtd_theme <https://pypi.org/project/sphinx-rtd-theme/>`__                                 \-                 docs               This Sphinx theme provides a great reader experience for documentation.
 `pydata-sphinx-theme <https://pydata-sphinx-theme.readthedocs.io/en/stable/>`__                   \-                 docs               A clean, Bootstrap-based Sphinx theme.
 `sphinx_design <https://sphinx-design.readthedocs.io/en/latest/>`__                               \-                 docs               A sphinx extension for designing.
 `sphinx-copybutton <https://sphinx-copybutton.readthedocs.io/en/latest/>`__                       \-                 docs               To copy codeblocks.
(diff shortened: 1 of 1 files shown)


agriyakhetarpal (approved):
Thanks! While this is good for now, we should think of ways to update this file periodically, because we do forget about these updates quite a lot. Do you have any ideas?

Saransh-cpp (approved):

prady0t:
One this we can do is to create a script that parses both `.rst` and `.toml` file and compare the contents inside respective fields and run it as a nox session everywhere (in CI, locally every time before push etc.). Are there ways sphinx can be updated according to pyproject file? (there might be extensions which can be useful)

agriyakhetarpal:
No, a `nox` session/CI job sounds overkill to me, especially with the number of jobs we have already. However, one way is to create [custom, mini `pre-commit` hooks](https://github.com/conda/ceps/blob/389f21d9912660ac04f669ee4db1007f69ca24f4/.pre-commit-config.yaml#L25-L63) so that the user can be reminded if they add a dependency to `pyproject.toml` but don't update the documentation.

The other way – where we have the task of updating the documentation file, yes, could you convert it from reST to Markdown first? Tables are much simpler in the latter. Also, we already support Markdown via MyST and have the user guide in Markdown already, so at least for non-API documentation, we have a strong case for moving everything to Markdown. Pandoc or `rst2myst` should do this for you already. The next part would be to dynamically create the table of dependencies by parsing the `pyproject.toml` file, which you can do via some `jinja` and Python (see https://www.ericholscher.com/blog/2016/jul/25/integrating-jinja-rst-sphinx/), which we use to add in the ["bundled" parameter sets on the Parameter Sets page](https://docs.pybamm.org/en/stable/source/api/parameters/parameter_sets.html#bundled-parameter-sets):

https://github.com/pybamm-team/PyBaMM/blob/f851d5dda1c9069ebe9c5cc071cb7f5ece7f6426/docs/conf.py#L459-L484

is what is used to populate

https://github.com/pybamm-team/PyBaMM/blob/f851d5dda1c9069ebe9c5cc071cb7f5ece7f6426/docs/source/api/parameters/parameter_sets.rst?plain=1#L85-L105

agriyakhetarpal:
Or, maybe more straightforward, you could attempt to write a custom Sphinx extension and register it with the reST-based build, similar to something like https://github.com/pyodide/pyodide/blob/b6ce33c42dfd97aa029e562c4ca9455ae363066c/docs/sphinx_pyodide/sphinx_pyodide/packages.py

But it depends on how simple you can make the solution to be – having something like this means it is just another script we must keep maintaining. If you can come up with something neat and easy to work with, that would be nice. I'd consider this fairly low-priority work, though, and something to do as an exercise over a few hours on a weekend, especially because our list of dependencies and their lower/upper bounds are not updated often.

prady0t:
A sphinx extension seems like a good idea. Should I open an issue for now related to this?

kratman:
I don't think we really need to do more for this. We have too much CI for this project already

prady0t:
We can then maybe just create a custom pre-commit hook as mentioned by @agriyakhetarpal above?

Saransh-cpp (commented):
I agree with @kratman here. We should leave it as it is and manually check the documentation once in a while. Writing a script or a pre-commit hook for this is an overkill and a waste of resources.

agriyakhetarpal:
Agreed. FWIW, just a simple [admonition](https://pydata-sphinx-theme.readthedocs.io/en/stable/examples/kitchen-sink/admonitions.html#warning) on that page that mentions something like "this list can be outdated, please check the `pyproject.toml` file" would be enough, IMO. Please feel free to add it if you would like.

PR #4702, opened 2024-12-20
fix and refactor fv integral code, add test of tertiary domain

aabills:


aabills:
this is currently failing integration tests when there is a quaternary dimension, investigating further, do not merge for now.

aabills:
Actually, it looks like the integration test failures are coming from the original issue-4056 branch, so should be gtg

PR #4707, opened 2024-12-23
Improved option checking for tuples

### LINKED ISSUE #3303: Better option checking

medha-14:
# Description

Fixes #3303

## Type of change

# Key checklist:

## Further checks:


medha-14:
I have implemented better option checking by creating an `ensure_tuple` method. This method automatically converts any string-based options into tuples, ensuring consistent handling of options.

src/pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -513,59 +546,71 @@ def __init__(self, extra_options):
                 )
 
         # Options not yet compatible with contact resistance
-        if options["contact resistance"] == "true":
-            if options["operating mode"] == "explicit power":
+        _ensure_tuple(options["contact resistance"])
+        if "true" in options["contact resistance"]:
+            _ensure_tuple(options["operating mode"])
+            if "explicit power" in options["operating mode"]:
```
agriyakhetarpal, line 552:
```suggestion
        _ensure_tuple(options["contact resistance"])
        _ensure_tuple(options["operating mode"])
        if "true" in options["contact resistance"]:
            if "explicit power" in options["operating mode"]:
```

Would it make sense to use your helper function outside of conditional statements, so that the option is ensured to be a tuple regardless of whether the condition is hit?

src/pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -513,59 +546,71 @@ def __init__(self, extra_options):
...
+            if "explicit resistance" in options["operating mode"]:
                 raise NotImplementedError(
                     "Contact resistance not yet supported for explicit resistance."
                 )
 
         # Options not yet compatible with particle-size distributions
-        if options["particle size"] == "distribution":
+        _ensure_tuple(options["particle size"])
+        if "distribution" in options["particle size"]:
+            _ensure_tuple(options["lithium plating porosity change"])
```
agriyakhetarpal, line 564:
```suggestion
        _ensure_tuple(options["particle size"])
        _ensure_tuple(options["lithium plating porosity change"])
        if "distribution" in options["particle size"]:
```

Same question here

src/pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -513,59 +546,71 @@ def __init__(self, extra_options):
...
             if options["lithium plating porosity change"] != "false":
                 raise NotImplementedError(
-                    "Lithium plating porosity change not yet supported for particle-size"
-                    " distributions."
+                    "Lithium plating porosity change not yet supported for particle-size distributions."
                 )
+
+            _ensure_tuple(options["heat of mixing"])
```
agriyakhetarpal, line 570: Reading them as they come, I feel it would be nicer for readability if all the `_ensure_tuple()` calls would come before conditional statements.

agriyakhetarpal (commented):
Thanks for starting this!

src/pybamm/models/full_battery_models/base_battery_model.py
arjxn-py: Yes, I agree. In that case, the nested conditions could also consolidated in a single line

src/pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -466,6 +490,10 @@ def __init__(self, extra_options):
         # Note: this check is currently performed on full cells, but is loosened for
         # half-cells where you must pass a tuple of options to only set MSMR models in
         # the working electrode
+        _ensure_tuple(options["open-circuit potential"])
+        _ensure_tuple(options["particle"])
+        _ensure_tuple(options["intercalation kinetics"])
```
prady0t, line 495:
```suggestion
        options["open-circuit potential"] = _ensure_tuple(options["open-circuit potential"])
       options["particle"] = _ensure_tuple(options["particle"])
        ptions["intercalation kinetics"] = _ensure_tuple(options["intercalation kinetics"])
```

src/pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -653,6 +699,7 @@ def __init__(self, extra_options):
 
         # Check options are valid
         for option, value in options.items():
+            _ensure_tuple(value)
             if isinstance(value, str) or option in [
```
prady0t, line 703: Shouldn't we check for if isinstance(value, tuple) instead?

src/pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -18,6 +18,27 @@ def represents_positive_integer(s):
...
+    Returns
+    -------
+    tuple
+        The input value as a tuple.
+
+    """
+    if isinstance(value, str):
+        return (value,)
+    elif isinstance(value, tuple):
+        return value
```
agriyakhetarpal, line 39: `if-elif` constructs should always end with an `else:` statement to catch unwarranted behaviour. But since we are the only users of this private method, it's probably fine here...

We could simplify this to something like (since we know the inputs and outputs and would never hit the `None` case):

```suggestion
    return (value,) if isinstance(value, str) else value if isinstance(value, tuple) else None
```

src/pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -513,63 +546,76 @@ def __init__(self, extra_options):
...
+            "lithium plating porosity change",
+            "heat of mixing",
+            "particle",
+            "particle mechanics",
+            "particle shape",
+            "SEI",
+            "stress-induced diffusion",
+            "thermal",
+        ]:
+            _ensure_tuple(options[option])
```
agriyakhetarpal, line 576:
```suggestion
        for _ in [
            "particle size",
            "lithium plating porosity change",
            "heat of mixing",
            "particle",
            "particle mechanics",
            "particle shape",
            "SEI",
            "stress-induced diffusion",
            "thermal",
        ]:
            _ensure_tuple(options[_])
```

As a convention, for variables that are not needed later in the code and are used only during the context of loops, we can use `_` to define them.

src/pybamm/models/full_battery_models/base_battery_model.py
prady0t: If true we must then rename the helper function to something like `_make_tuple` instead

agriyakhetarpal (commented):
This looks good to me functionality-wise, thanks. I have a few more suggestions. I'll let @rtimms advise on any other changes, since the issue was originally his idea.

src/pybamm/models/full_battery_models/base_battery_model.py
medha-14: Would returning none  be the best approach as , there can be cases where the values can be numeric or boolean, should those cases be handled as they come. eg the part @prady0t pointed out 

```python
for option, value in options.items():
            _ensure_tuple(value)
            if isinstance(value, str) or option in [
                "dimensionality",
                "operating mode",
            ]:  # some options accept non-strings
                value = (value,)
```
Should i remove `ensure_tuple` here and handle such cases seperately?

agriyakhetarpal: Ah, I think what happened was that @prady0t and I added in our reviews at the same time, and I didn't see his review (sorry!)

Yes, handling numeric or boolean options is needed – please feel free to ignore my suggestion in this case. Thanks!

src/pybamm/models/full_battery_models/base_battery_model.py
Saransh-cpp: Yep! This is not editing things in-place.

src/pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -18,6 +18,29 @@ def represents_positive_integer(s):
...
+    tuple or None
+        The input value as a tuple, or None if the input is not a string or tuple.
+    """
+    return (
+        (value,)
+        if isinstance(value, str)
+        else value
+        if isinstance(value, tuple)
+        else None
+    )
```
Saransh-cpp, line 41: Are there cases where `value` can be something other than `str` or `tuple`? If yes, then the type signature of this function is wrong. If not, `None` should not be returned. (It is always preferred to not have the return value of a function depend on an `if` condition, so removing the `None` would be the ideal case.)

Also, I would really prefer a nested `if-else` block here for readability.

Saransh-cpp (changes requested):

src/pybamm/models/full_battery_models/base_battery_model.py
Saransh-cpp: Ah, just saw @agriyakhetarpal's comment above.

Saransh-cpp: The type signature should still be fixed + it should be converted into an `if-else` block.

medha-14:
Would it be wrong to assume that all options passed can be expressed in two ways:

If a single value is passed for an option, it would look like this:

```python
options = {
    "working electrode": "both",
}
```
If multiple values are passed, it would be structured like this:

```python
options = {
    "SEI": ("none", "constant"),
}
```
Are these the only two possibilities?
or should we consider some different input formats as well?

medha-14:
Is it necessary to convert the options to tuples beforehand? The problem can be resolved simply by changing
```python
if options["particle size"] == "distribution"
```
to this
```python
if "distribution" in options["particle size"] :
```
This approach works for both string and tuple options. Please let me know if I am overlooking anything.

src/pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -624,7 +630,7 @@ def __init__(self, extra_options):
         if options["particle phases"] not in ["1", ("1", "1")]:
             if not (
                 options["surface form"] != "false"
```
Saransh-cpp, line 632:
```suggestion
                "false" not in options["surface form"]
```

src/pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -540,22 +539,29 @@ def __init__(self, extra_options):
...
+            if "spherical" not in options["particle shape"]:
                 raise NotImplementedError(
                     "Particle shape must be 'spherical' for particle-size distribution"
                     " submodels."
                 )
-            if options["stress-induced diffusion"] == "true":
+            if "none" not in options["SEI"]:
+                raise NotImplementedError(
+                    "SEI submodels do not yet support particle-size distributions."
+                )
```
Saransh-cpp, line 563: Where did this block appear from?

src/pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -445,8 +444,8 @@ def __init__(self, extra_options):
             default_options["SEI"] = "none"
         # The "SEI" option will still be overridden by extra_options if provided
 
-        options = pybamm.FuzzyDict(default_options)
         # any extra options overwrite the default options
+        options = pybamm.FuzzyDict(default_options)
```
Saransh-cpp, line 448: I think this was moved down by accident.

Saransh-cpp (changes requested):
Thanks, @medha-14! Sorry for taking too long to review. See my comments below. it would also be nice to have some tests that check for both string and sequence-like values for `options`.

src/pybamm/models/full_battery_models/base_battery_model.py
medha-14: Sorry, Iwas testing some things and accidentally left that block in.

medha-14:
Sorry for taking too long to respond, I’ve added the requested tests and fixed the failing ones.

PR #4732, opened 2025-01-02
Misc. cleanup and fixes for broken tests

kratman:
# Description

I split these changes out of another branch to keep it cleaner. Most of the changes are to remove static analysis warnings.

Edit: Also fixed the broken tests on develop

# Key checklist:

## Further checks:

### DIFF
--- a/tests/unit/test_models/test_full_battery_models/test_base_battery_model.py
+++ b/tests/unit/test_models/test_full_battery_models/test_base_battery_model.py
@@ -430,7 +430,7 @@ class TestBaseBatteryModel:
         assert isinstance(model.default_solver, pybamm.CasadiSolver)
 
         # check that default_solver gives you a new solver, not an internal object
-        _ = model.default_solver
+        solver = model.default_solver
         solver = pybamm.BaseModel()
         assert isinstance(model.default_solver, pybamm.CasadiSolver)
         assert isinstance(solver, pybamm.BaseModel)
--- a/tests/unit/test_models/test_full_battery_models/test_base_battery_model.py
+++ b/tests/unit/test_models/test_full_battery_models/test_base_battery_model.py
@@ -430,7 +430,7 @@ class TestBaseBatteryModel:
         assert isinstance(model.default_solver, pybamm.CasadiSolver)
 
         # check that default_solver gives you a new solver, not an internal object
-        solver = model.default_solver
+        _ = model.default_solver
         solver = pybamm.BaseModel()
         assert isinstance(model.default_solver, pybamm.CasadiSolver)
         assert isinstance(solver, pybamm.BaseModel)
--- a/tests/unit/test_simulation.py
+++ b/tests/unit/test_simulation.py
@@ -1,4 +1,3 @@
-import tempfile
 import pybamm
 import numpy as np
 import pandas as pd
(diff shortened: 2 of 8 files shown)


tests/unit/test_simulation.py
```diff
@@ -482,27 +481,26 @@ def test_save_load(self):
...
-        assert (
-            "graphite_LGM50_electrolyte_exchange_current_density_Chen2020"
-            == pkl_obj.parameter_values[
-                "Negative electrode exchange-current density [A.m-2]"
-            ].__name__
-        )
-        os.remove(filename)
+        with tempfile.NamedTemporaryFile() as tmp:
+            filename = tmp.name
+            print(filename)
```
kratman, line 486:
```suggestion
```

tests/unit/test_simulation.py
```diff
@@ -482,27 +481,25 @@ def test_save_load(self):
...
-            raise excep
-
-        assert (
-            "graphite_LGM50_electrolyte_exchange_current_density_Chen2020"
-            == pkl_obj.parameter_values[
-                "Negative electrode exchange-current density [A.m-2]"
-            ].__name__
-        )
-        os.remove(filename)
+        with tempfile.NamedTemporaryFile() as tmp:
```
kratman, line 484:
```suggestion
        with tempfile.NamedTemporaryFile(delete=False) as tmp:
```

tests/unit/test_simulation.py
```diff
@@ -482,27 +481,25 @@ def test_save_load(self):
...
-            raise excep
-
-        assert (
-            "graphite_LGM50_electrolyte_exchange_current_density_Chen2020"
-            == pkl_obj.parameter_values[
-                "Negative electrode exchange-current density [A.m-2]"
-            ].__name__
-        )
-        os.remove(filename)
+        with tempfile.NamedTemporaryFile(delete=False) as tmp:
```
Saransh-cpp, line 484: Could you please use [pytest's `tmp_path` fixture](https://docs.pytest.org/en/stable/how-to/tmp_path.html) here instead?

tests/unit/test_models/test_full_battery_models/test_base_battery_model.py
```diff
@@ -430,7 +430,7 @@ def test_default_solver(self):
         assert isinstance(model.default_solver, pybamm.CasadiSolver)
 
         # check that default_solver gives you a new solver, not an internal object
-        solver = model.default_solver
+        _ = model.default_solver
         solver = pybamm.BaseModel()
         assert isinstance(model.default_solver, pybamm.CasadiSolver)
         assert isinstance(solver, pybamm.BaseModel)
```
Saransh-cpp, line 436: I could be wrong, but I think the point of this test was to check if the `solver` variable is being overwritten correctly? In that case we can ignore the error that the linter is throwing here; otherwise, this looks good!

PS: I don't really get what this test is trying to do.

Saransh-cpp (dismissed):
Thanks, @kratman! See a couple of quick comments below -

tests/unit/test_models/test_full_battery_models/test_base_battery_model.py
```diff
@@ -430,7 +430,7 @@ def test_default_solver(self):
         assert isinstance(model.default_solver, pybamm.CasadiSolver)
 
         # check that default_solver gives you a new solver, not an internal object
-        solver = model.default_solver
+        _ = model.default_solver
```
kratman, line 433:
```suggestion
        solver = model.default_solver
```

kratman:
This should fix the broken tests on develop as well

kratman:
Missed the integration tests, working on them now

kratman:
Most of the test failures on develop were from precision. Scipy updated and something else changed with string format stuff

brosaplanella (approved):
Looks good, thanks!

PR #4745, opened 2025-01-10
Use ```deprecation``` package for deprecation warnings #2028

### LINKED ISSUE #2028: Use `deprecation` package for deprecation warnings

RohitP2005:
# Description

This pull request has added astropy's deprecation wrapper for a function's arguement and the fucntion itself and has added ```-W``` flag to ensure the deprecation warnings are treated as error

Fixes #2028 

## Type of change

- [x] New feature (non-breaking change that adds functionality)
- [x] Deprecation (non-breaking change that deprecates functionality with warnings)

# Key checklist:

- [x] No style issues: `$ pre-commit run --all-files` (or `$ nox -s pre-commit`) (see [CONTRIBUTING.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CONTRIBUTING.md#installing-and-using-pre-commit) for how to set this up to run automatically when committing locally)
- [x] All tests pass: `$ python -m pytest -W error` (or `$ nox -s tests`) (ensure warnings are treated as errors to catch deprecation notices)

You can run integration tests, unit tests, and doctests together at once using `$ nox -s quick`.

## Further checks:

- [ ] Code is commented, particularly in hard-to-understand areas, including deprecation warnings and helper functions.
- [ ] Tests added for any new functionality or to prove that the deprecation warnings are effective.
- [x] Confirm that any borrowed code from Astropy is appropriately attributed according to its license.

### DIFF
--- a/src/pybamm/solvers/solution.py
+++ b/src/pybamm/solvers/solution.py
@@ -1130,6 +1130,10 @@ def make_cycle_solution(
         sum_sols.all_yps,
         sum_sols.variables_returned,
     )
+
+    if sum_sols.variables_returned:
+        cycle_solution._variables = sum_sols._variables
+
     cycle_solution._all_inputs_casadi = sum_sols.all_inputs_casadi
     cycle_solution._sub_solutions = sum_sols.sub_solutions
 
--- a/tests/integration/test_solvers/test_idaklu.py
+++ b/tests/integration/test_solvers/test_idaklu.py
@@ -133,23 +133,23 @@ class TestIDAKLUSolver:
         t_combined = np.concatenate((sol3.t, t_interp_dense))
         t_combined = np.unique(t_combined)
         t_combined.sort()
-        np.testing.assert_array_almost_equal(sol3.t, t_combined)
+        np.testing.assert_allclose(sol3.t, t_combined)
 
         # 4. sparse t_eval + sparse t_interp
         sol4 = solver.solve(
             model_disc, t_eval_sparse, t_interp=t_interp_sparse, inputs=inputs
         )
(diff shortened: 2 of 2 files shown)


utils/decorators.py
Saransh-cpp, line 1: Thanks for picking this up, @RohitP2005! The decorators should be adjusted to work with PyBaMM (and the current warnings should be replaced with the newly added decorators).

The current implementation added by you will not work as it uses Astropy's internal exception classes. Things should not be simply copied.

The file can also be renamed to something like `deprecation_utils`, but we can look into that once the decorators are up and running.

Saransh-cpp (changes requested):

utils/decorators.py
kratman: It is also a lot of code just for improving a couple warnings. We should make sure to only include what we need

RohitP2005: Yeah @Saransh-cpp , understood  I will start working on it.

RohitP2005:
I thought @kratman was right about having too much code for handling a few warnings. So, I wrote fresh decorators and exceptions to streamline and reuse them in the future. I integrated these into the existing functions and renamed them. The previous deprecation warnings are still in place and will be removed in the next commit once this is sorted out. I need help verifying if my implementations are correct. Kindly guide me on how to further address this issue.

agriyakhetarpal:
Thanks for working on this, @RohitP2005 – we would be happy to check your implementation in your new PR.

PR #4746, opened 2025-01-11
geometric param example

aabills:
# Description

adds an example for using a geometric input parameter

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:


kratman:
Seems like the docs failure is because the notebook was not added to the TOC in the documentation

docs/source/examples/notebooks/parameterization/geometric_parameters.ipynb
kratman, line 1: Overall I think it is fine, but it might benefit from a bit more discussion, especially towards the end. The example ends with a statement along the lines of "This is 10 times faster", but it could be nice to reiterate that the speedup is due to avoiding the recreation of certain classes. I know you mentioned it in the beginning, but the example just seems to end suddenly

aabills: I added that. I don't really think too much discussion is needed here tbh, I think most people who are advanced enough to actually use this understand the difference (people will have to rewrite the DFN model to use this anyway). The purpose of this example is to show how to use the symbolic mesh, not really intended as an intro.

kratman (approved):

PR #4748, opened 2025-01-12
Deprecation warning and function cleanup before 25.1

kratman:
# Description

Checking for outdated deprecation warnings before the 25.1 release. I did some other minor cleanup seen along the way as well.

# Key checklist:

## Further checks:

### DIFF
--- a/src/pybamm/models/base_model.py
+++ b/src/pybamm/models/base_model.py
@@ -1151,7 +1151,7 @@ class BaseModel:
 
         Parameters
         ----------
-        parameter_name : str
+        symbol_name : str
         """
         # Should we deprecate this? Not really sure how it's used?
 
--- a/src/pybamm/parameters/parameter_sets.py
+++ b/src/pybamm/parameters/parameter_sets.py
@@ -1,5 +1,4 @@
 import sys
-import warnings
 import importlib.metadata
 import textwrap
 from collections.abc import Mapping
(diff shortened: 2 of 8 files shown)


src/pybamm/parameters/parameter_sets.py
```diff
@@ -79,22 +78,6 @@ def get_docstring(self, key):
...
-            # this file now return the name as a string, which will load the same
-            # parameter set as before when passed to `ParameterValues`
-            if name in self:
-                msg = (
-                    f"Parameter sets should be called directly by their name ({name}), "
-                    f"instead of via pybamm.parameter_sets (pybamm.parameter_sets.{name})."
-                )
-                warnings.warn(msg, DeprecationWarning, stacklevel=2)
-                return name
-            raise error
```
kratman, line 96: This warning was a couple years old, so I removed the warning. It was used in an example, but I fixed that as well.

src/pybamm/util.py
```diff
@@ -23,23 +22,6 @@ def root_dir():
...
-    try:
-        # Get the latest git commit hash
-        return str(
-            subprocess.check_output(["git", "describe", "--tags"], cwd=root_dir())
-            .strip()
-            .decode()
-        )
-    except subprocess.CalledProcessError:  # pragma: no cover
-        # Not a git repository so just return the version number
-        return f"v{pybamm.__version__}"
```
kratman, line 40: This function only seemed to be used in tests, and does not seem like it belong in PyBaMM

src/pybamm/models/base_model.py
```diff
@@ -1151,7 +1151,7 @@ def info(self, symbol_name):
 
         Parameters
         ----------
-        parameter_name : str
+        symbol_name : str
```
kratman, line 1154: This variable name did not match what was in the function signature

kratman (commented):
Explanations

Saransh-cpp (approved):
Thanks, @kratman!

PR #4751, opened 2025-01-13
Remove some missed stuff from the build

kratman:
# Description

Fix an issue with the release process

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/.github/workflows/publish_pypi.yml
+++ b/.github/workflows/publish_pypi.yml
@@ -34,4 +34,3 @@ jobs:
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
           packages-dir: deploy/
-
--- a/.github/workflows/publish_pypi.yml
+++ b/.github/workflows/publish_pypi.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Upload package
         uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
         with:
-          name: wheels
+          name: distributions
           path: deploy/
           if-no-files-found: error
 
--- a/.github/workflows/publish_pypi.yml
+++ b/.github/workflows/publish_pypi.yml
@@ -82,7 +82,7 @@ jobs:
           CIBW_TEST_COMMAND: |
             python -m pytest -m cibw {project}/tests/unit
       - name: Upload Windows wheels
-        uses: actions/upload-artifact@6f51ac03b9356f520e9adb1b1b7802705f340c2b # v4.5.0
+        uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
         with:
           name: wheels_windows
           path: ./wheelhouse/*.whl
@@ -114,7 +114,7 @@ jobs:
             python -m pytest -m cibw {project}/tests/unit
 
       - name: Upload wheels for Linux
-        uses: actions/upload-artifact@6f51ac03b9356f520e9adb1b1b7802705f340c2b # v4.5.0
+        uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
         with:
           name: wheels_manylinux
           path: ./wheelhouse/*.whl
@@ -166,7 +166,7 @@ jobs:
             python -m pytest -m cibw {project}/tests/unit
 
       - name: Upload wheels for macOS (amd64, arm64)
-        uses: actions/upload-artifact@6f51ac03b9356f520e9adb1b1b7802705f340c2b # v4.5.0
+        uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
         with:
           name: wheels_${{ matrix.os }}
           path: ./wheelhouse/*.whl
@@ -186,7 +186,7 @@ jobs:
         run: pipx run build --sdist
 
       - name: Upload SDist
-        uses: actions/upload-artifact@6f51ac03b9356f520e9adb1b1b7802705f340c2b # v4.5.0
+        uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
         with:
           name: sdist
           path: ./dist/*.tar.gz
--- a/.github/workflows/publish_pypi.yml
+++ b/.github/workflows/publish_pypi.yml
@@ -8,7 +8,6 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Set up Python
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       - uses: actions/setup-python@0b93645e9fea7318ecaed2b359559ac225c90a2b # v5.3.0
         with:
--- a/.github/workflows/publish_pypi.yml
+++ b/.github/workflows/publish_pypi.yml
@@ -48,14 +48,6 @@ jobs:
           with open(output_file, "a", encoding="utf-8") as output_stream:
               output_stream.write(f"count={num_cpus}\n")
 
-      - name: Cache packages installed through vcpkg on Windows
-        uses: actions/cache@1bd1e32a3bdc45362d1e726936510720a7c30a57 # v4.2.0
-        env:
-          cache-name: vckpg_binary_cache
-        with:
-          path: C:\Users\runneradmin\AppData\Local\vcpkg\archives
-          key: ${{ runner.os }}-build-VS2022-${{ env.cache-name }}-${{ hashFiles('vcpkg*.json') }}
-
       # Enable tmate debugging of manually-triggered workflows if the input option was provided
       - name: Setup tmate session
         uses: mxschmitt/action-tmate@e5c7151931ca95bad1c6f4190c730ecf8c7dde48 # v3.19
@@ -66,10 +58,6 @@ jobs:
         env:
           CIBW_ENVIRONMENT: >
             PYBAMM_DISABLE_TELEMETRY="true"
-            PYBAMM_USE_VCPKG=ON
-            VCPKG_ROOT_DIR=C:\vcpkg
-            VCPKG_DEFAULT_TRIPLET=x64-windows-static-md
-            VCPKG_FEATURE_FLAGS=manifests,registries
             CMAKE_GENERATOR="Visual Studio 17 2022"
             CMAKE_GENERATOR_PLATFORM=x64
             CMAKE_BUILD_PARALLEL_LEVEL=${{ steps.get_num_cores.outputs.count }}
--- a/.github/workflows/publish_pypi.yml
+++ b/.github/workflows/publish_pypi.yml
@@ -2,239 +2,36 @@ name: Build and publish package to PyPI
 on:
   release:
     types: [published]
-  schedule:
-    # Run at 10 am UTC on day-of-month 1 and 15.
-    - cron: "0 10 1,15 * *"
-  workflow_dispatch:
-    inputs:
-      target:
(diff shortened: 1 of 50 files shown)


kratman:
@agriyakhetarpal Can you review this?

agriyakhetarpal (changes requested):
While the cleanup is good to have, I'd also like to point out that the `publish_pypi.yml` workflow will have to be revised completely now, as `cibuildwheel` does not support and fails on building pure Python wheels. We can simply use `python -m build` (i.e., just copying BPX's workflow should be fine). I'm happy to approve this PR once done, thanks!

Saransh-cpp:
Please feel free to ping me when this is ready for review!

.github/workflows/publish_pypi.yml
```diff
@@ -2,251 +2,36 @@ name: Build and publish package to PyPI
...
         with:
-          path: artifacts
-          merge-multiple: true
+          python-version: ${{ matrix.python-version }}
 
-      - name: Sanity check downloaded artifacts
-        run: ls -lA artifacts/
+      - name: Build wheel
+        run: |
+          pip wheel . --no-deps -w deploy
```
agriyakhetarpal, line 24:
```suggestion
      - name: Build wheel
        run: |
          pipx run build --outdir deploy
```

`pypa/build` is now more widely used and will become the canonical build frontend likely at some time in the future

.github/workflows/publish_pypi.yml
```diff
@@ -2,251 +2,36 @@ name: Build and publish package to PyPI
...
-      name: pypi
-      url: https://pypi.org/p/pybamm
+        python-version: ["3.12"]
     permissions:
       id-token: write
 
     steps:
-      - name: Download all artifacts
-        uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16 # v4.1.8
+      - uses: actions/checkout@v3
```
agriyakhetarpal, line 16:
```suggestion
      - uses: actions/checkout@v3
```

Could you bump to v4.2.2 and pin to the hash? v3 is quite old by now.

.github/workflows/publish_pypi.yml
```diff
@@ -2,251 +2,36 @@ name: Build and publish package to PyPI
...
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
```
agriyakhetarpal, line 18: Likewise – setup-python is also at v5. Could you pin it the commit hash as well?

.github/workflows/publish_pypi.yml
```diff
@@ -2,251 +2,36 @@ name: Build and publish package to PyPI
...
-          packages-dir: artifacts/
+          name: wheels
+          path: deploy/
+          if-no-files-found: error
 
-      - name: Publish to TestPyPI
-        if: github.event.inputs.target == 'testpypi'
-        uses: pypa/gh-action-pypi-publish@67339c736fd9354cd4f8cb0b744f2b82a74b5c70 # v1.12.3
+      - name: Publish on PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
```
agriyakhetarpal, line 34: It is strongly recommended to split the publishing action to a new job that `needs:` the build job for security reasons, and add the `id-token:` permission to just the publishing job: https://github.com/pypa/gh-action-pypi-publish#trusted-publishing and https://docs.pypi.org/trusted-publishers/security-model/#provider-specific-considerations. Also, we'll need an `environment:` for the publishing job.

.github/workflows/publish_pypi.yml
```diff
@@ -2,251 +2,35 @@ name: Build and publish package to PyPI
...
-      build_manylinux_wheels,
-      build_macos_wheels,
-      build_windows_wheels,
-      build_sdist
-    ]
-    runs-on: ubuntu-latest
-    environment:
-      name: pypi
-      url: https://pypi.org/p/pybamm
+        python-version: ["3.12"]
```
agriyakhetarpal, line 11: Since there's just one Python version, the matrix and the strategy can be dropped

agriyakhetarpal (changes requested):
Thanks, just a few more things left

kratman:
@agriyakhetarpal I put all the versions back and split the job

.github/workflows/publish_pypi.yml
```diff
@@ -2,251 +2,38 @@ name: Build and publish package to PyPI
...
+        run: |
+          pipx run build --outdir deploy
 
-      - name: Upload SDist
+      - name: Upload package
         uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
         with:
-          name: sdist
-          path: ./dist/*.tar.gz
+          name: wheels
```
Saransh-cpp, line 24: Nitpick, but -
1. Only 1 wheel is built now
2. SDist is built and uploaded too
```suggestion
          name: distributions
```

.github/workflows/publish_pypi.yml
```diff
@@ -2,251 +2,38 @@ name: Build and publish package to PyPI
...
-    # This job is only of value to PyBaMM and would always be skipped in forks
-    if: github.event_name != 'schedule' && github.repository == 'pybamm-team/PyBaMM'
-    name: Upload package to PyPI
-    needs: [
-      build_manylinux_wheels,
-      build_macos_wheels,
-      build_windows_wheels,
-      build_sdist
-    ]
+  Publish:
```
Saransh-cpp, line 28: Nitpick, but conventionally the job names are in `snake_case`

Saransh-cpp (changes requested):
Thanks, @kratman! See my comments below -

.github/workflows/publish_pypi.yml
```diff
@@ -2,251 +2,38 @@ name: Build and publish package to PyPI
...
 
     steps:
-      - name: Download all artifacts
-        uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16 # v4.1.8
-        with:
-          path: artifacts
-          merge-multiple: true
-
-      - name: Sanity check downloaded artifacts
-        run: ls -lA artifacts/
```
Saransh-cpp, line 220: I would prefer keeping this too

.github/workflows/publish_pypi.yml
kratman: Yeah it was a wheel before the previous requested changes

.github/workflows/publish_pypi.yml
```diff
@@ -2,251 +2,38 @@ name: Build and publish package to PyPI
...
-      build_macos_wheels,
-      build_windows_wheels,
-      build_sdist
-    ]
+  Publish:
+    needs: Build
     runs-on: ubuntu-latest
-    environment:
-      name: pypi
-      url: https://pypi.org/p/pybamm
```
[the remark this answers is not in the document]
kratman: I did not need that on other workflows that deployed to PyPi. The one I copied here before the last round of changes was taken from a working deployment I had setup for another project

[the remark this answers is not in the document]
kratman: I did put the environment back, the URL is not needed unless you are using test-pypi

https://docs.pypi.org/trusted-publishers/using-a-publisher/

.github/workflows/publish_pypi.yml
kratman: Actually I checked, and that command does make a wheel

.github/workflows/publish_pypi.yml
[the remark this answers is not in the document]
Saransh-cpp: Ah that's nice and neat! TIL

.github/workflows/publish_pypi.yml
```diff
@@ -2,251 +2,38 @@ name: Build and publish package to PyPI
...
           python-version: 3.12
 
-      - name: Build SDist
-        run: pipx run build --sdist
+      - name: Build wheel
+        run: |
+          pipx run build --outdir deploy
 
-      - name: Upload SDist
+      - name: Upload package
```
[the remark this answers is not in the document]
kratman: how is it not a package?

.github/workflows/publish_pypi.yml
```diff
@@ -2,251 +2,38 @@ name: Build and publish package to PyPI
...
+    environment: pypi
     permissions:
       id-token: write
 
     steps:
-      - name: Download all artifacts
-        uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16 # v4.1.8
-        with:
-          path: artifacts
-          merge-multiple: true
```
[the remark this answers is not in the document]
kratman: I am putting it back. The pervious version did not need it because it was a single job

.github/workflows/publish_pypi.yml
```diff
@@ -2,251 +2,38 @@ name: Build and publish package to PyPI
...
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       - uses: actions/setup-python@0b93645e9fea7318ecaed2b359559ac225c90a2b # v5.3.0
         with:
           python-version: 3.12
 
-      - name: Build SDist
-        run: pipx run build --sdist
+      - name: Build wheel
+        run: |
+          pipx run build --outdir deploy
```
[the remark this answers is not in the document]
kratman: Still builds wheels, but I made the pipx part a single line

.github/workflows/publish_pypi.yml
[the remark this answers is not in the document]
kratman: This was the original that I used when I updated it: https://github.com/ionworks/Ionwizard/blob/main/.github/workflows/release.yml

.github/workflows/publish_pypi.yml
Saransh-cpp: I meant that it builds both wheel + SDist. PyBaMM used to build the wheel and the SDist as two different steps before the C++ part was removed (IIRC). Both the distributions (wheel and SDist) are uploaded to PyPI and users can choose which one they want to use (wheel is the default one). It would be wrong to call it "wheels" given that the artifact has a source distribution too (I hope I am not doing something wrong, but that is how it should be).

kratman: Don't think the name matters but I will change it

.github/workflows/publish_pypi.yml
kratman: It does not seem like this really does any good. The build step prints them, and if this returns nothing it would not cause the job to stop

.github/workflows/publish_pypi.yml
```diff
@@ -2,251 +2,38 @@ name: Build and publish package to PyPI
 on:
   release:
     types: [published]
```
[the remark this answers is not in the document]
kratman: The workflow dispatch was more useful when testpypi was used, but without that it does not seem very useful

kratman:
@Saransh-cpp, @agriyakhetarpal Is this good now? I can't do the release until this is merged

.github/workflows/publish_pypi.yml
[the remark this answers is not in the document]
agriyakhetarpal: @Saransh-cpp is technically right, because `python -m build` builds both distributions – the sdist (source distribution), and the wheel (binary distribution). But let's not be pedantic, no worries

agriyakhetarpal (approved):

.github/workflows/publish_pypi.yml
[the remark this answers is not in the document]
Saransh-cpp: Yeah I would have preferred updating the name here and in the wheel job. I was being pedantic yes, but right now it is misguiding/wrong.

@kratman could you please also not resolve conversations and let the author mark it as resolved? Otherwise it defeats the purpose of adding a comment. I know this was blocking the release, but open-source reviews can be slow sometimes.

PR #4755, opened 2025-01-14
Release v25.1.0

kratman:
# Description

Release branch for v25.1.0

This must be merged instead of squashed.


Saransh-cpp (commented):
Hi @kratman, was this branch and the PR created from a fork by mistake? I would not advise merging things from forks to main (even trusted forks) as things can go wrong. We should ideally bump the version in develop and then merge develop into main so that everything is carried out safely.

kratman:
> Hi @kratman, was this branch and the PR created from a fork by mistake? I would not advise merging things from forks to main (even trusted forks) as things can go wrong. We should ideally bump the version in develop and then merge develop into main so that everything is carried out safely.

I use a single repo, for both remotes. I just pushed to the wrong remote. I will recreated it

PR #4759, opened 2025-01-15
Syncing v25.1.0 and develop

kratman:
Sync PR from the release.

This has to be merged, not squashed


valentinsulzer (approved):

Saransh-cpp (commented):
`main` and `develop` are not actually synced, the https://github.com/pybamm-team/PyBaMM/commit/8b335890750a68b2c84ecfd9028c19b08ba01f5e commit is not in `main` but is in `develop`. I am not sure if that was intentional or missed (as the release branch was created before the scheduled test fix was merged).

This means that the version update on main and develop have different number of commits. This also means that in the current release system, the release and sync PRs should be merged one after another, without anything going into develop in between.

In this case it turned out not to be harmful (just messed up git history a bit) as the extra commit was fixing scheduled tests, but it can be harmful in the future. Hence, my previous suggestion of updating the version number in develop (through a PR) first and merging develop into main for the release. Creating 2 additional branches for release and sync opens up doors to mess up the git history / miss commits.

There was also the instance where the release PR was opened through a fork instead of the main repository, which could have gone unnoticed, but that won't happen if we are merging develop.

Good software engineering practices but can be ignored! (ideally should not be 😞)

PR #4770, opened 2025-01-16
add operator to experiment terminations

aabills:
# Description

Adds operator for experiment terminations. While this won't functionally change many experiments, it enables experiments to more closely mirror cell testing protocols.

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:


src/pybamm/discretisations/discretisation.py
```diff
@@ -726,6 +726,8 @@ def process_dict(self, var_eqn_dict, ics=False):
         """
         new_var_eqn_dict = {}
         for eqn_key, eqn in var_eqn_dict.items():
+            if eqn_key == "Negative electrode open-circuit potential [V]":
+                print(eqn_key)
```
valentinsulzer, line 730: needs to be removed

tests/unit/test_experiments/test_experiment_steps.py
```diff
@@ -307,3 +307,28 @@ def custom_step_voltage(variables):
 
         with pytest.raises(ValueError, match="control must be"):
             pybamm.step.CustomStepImplicit(custom_step_voltage, control="bla")
+
+    def test_steps_with_operators(self):
+        # voltage
+        step = pybamm.step.voltage(1, duration=3600)
+        termination_lt_4_1 = pybamm.step.VoltageTermination(4.1, operator="<")
+        termination_gt_4_1 = pybamm.step.VoltageTermination(4.1, operator=">")
```
valentinsulzer, line 315: we could use operator overloading and have the interface be `pybamm.step.termination.Voltage > 4.1` (overload the `__gt__` operator to return `pybamm.step.VoltageTermination(4.1, operator="<")`

valentinsulzer (commented):
Good idea including this, do most cyclers ask for a direction or do they figure it out automatically?

src/pybamm/discretisations/discretisation.py
aabills: that's weird, that shouldn't have come from this pr

aabills:
> Good idea including this, do most cyclers ask for a direction or do they figure it out automatically?

Neware and Maccor, the only ones I've dealt with, have "charge" or "discharge" built into the steps and they also have these limits specified with operators.

The reason for this is so that they can go to a "safer" state in case of a short or something. This of course is not relevant for pybamm, but there's not a great way to filter the safety limits without catching the other limits.

src/pybamm/experiment/step/step_termination.py
```diff
@@ -67,9 +70,19 @@ def get_event(self, variables, step):
...
+        operator = self.operator
+        if operator is None:
+            expr = abs(variables["Current [A]"]) - self.value
+            string = f"abs(Current [A]) < {self.value} [A] [experiment]"
+        elif operator == ">":
+            expr = self.value - variables["Current [A]"]
+            string = f"Current [A] > {self.value} [A] [experiment]"
+        elif operator == "<":
+            expr = variables["Current [A]"] - self.value
+            string = f"Current [A] < {self.value} [A] [experiment]"
```
kratman, line 82: It is good to end if-elif chains with an else of some sort so that future devs know nothing was accidentally left off

aabills: In this case it won't be possible to reach the `else`, so what do you recommend doing about coverage

aabills: I guess we could make the None an else

src/pybamm/experiment/step/step_termination.py
```diff
@@ -89,24 +102,48 @@ def get_event(self, variables, step):
...
+                operator = "<"
+            else:
+                # No event for rest steps
+                return None
+
+        if operator == ">":
             sign = -1
-        elif direction == "Discharge":
+        elif operator == "<":
             sign = 1
```
kratman, line 119:
```suggestion
        else:
            # operator can only be "<" or ">"
            sign = 1
```

src/pybamm/experiment/step/step_termination.py
```diff
@@ -14,8 +14,11 @@ class BaseTermination:
         The value at which the event is triggered
     """
 
-    def __init__(self, value):
+    def __init__(self, value, operator=None):
         self.value = value
+        if operator not in ["<", ">", None]:
+            raise ValueError(f"Invalid operator: {operator}")
+        self.operator = operator
```
kratman, line 21: This could be a good case for using an enum. Then arbitrary stings can't be passed in and type checkers can see the error

src/pybamm/experiment/step/base_step.py
```diff
@@ -69,6 +76,10 @@ def __init__(
         description=None,
         direction=None,
```
kratman, line 77: This would work with `from __future__ import annotations` at the top.
```suggestion
        direction: Direction | None = None,
```

src/pybamm/experiment/step/base_step.py
```diff
@@ -69,6 +76,10 @@ def __init__(
...
     ):
+        if direction is not None and direction not in Direction:
```
kratman, line 79: Something like this could fix the failing tests
```suggestion
        if direction is not None and not isinstance(direction, Direction):
```

aabills: this doesn't work

aabills:
this idaklu error doesn't look related to this pr

kratman:
> this idaklu error doesn't look related to this pr

Yeah it is a flaky test, not sure why that happens, but I have not really been able to reproduce it locally. Re-triggering it should make it pass. It is on my to-do list

valentinsulzer (approved):

kratman (approved):

PR #4788, opened 2025-01-21
Replace CChem link with attribution to John Newman

kratman:
# Description

Linkchecker broke on this website. So instead of using the link, I changed the footnote to reference the program author and program name

# Key checklist:

## Further checks:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -422,17 +422,12 @@ wherever code is called that uses that citation (for example, in functions or in
 
 ### Installation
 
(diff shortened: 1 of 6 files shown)


kratman:
The other failure appears to be a bitbucket outage. Probably just temporary

kratman:
I replaced the bitbucket links with text explanation, and removed some links to stuff about CMake builds

aabills (approved):
to be honest, I'd really like to see a better citation, but I think it will be quite difficult to figure out

kratman:
> to be honest, I'd really like to see a better citation, but I think it will be quite difficult to figure out

Yeah I tried to find the original paper for it, but the earliest I found was a document for how to use it. Everything after that were application papers from what I could see

kratman:
I am guessing it was a link because previous devs had issues trying to find the best paper for citing it

PR #4813, opened 2025-02-03
Minor updates to degradation example

DrSOKane:
# Description

People are still setting `"calculate discharge energy"` to true based on my example notebook, so I changed it. I also emphasized the 1 minute period for the 1C discharge step (compared to the 5 minute period for other steps) because that has also caused confusion before.

If anyone has other ideas for how to improve the notebook while still being true to [the paper it's based on](https://pubs.rsc.org/en/content/articlelanding/2022/cp/d2cp00417h), I am happy to hear them.


rtimms:
I would recommend switching to the `IDAKLU` solver. It will be much faster and you won’t see all those solver warnings.

DrSOKane:
> I would recommend switching to the `IDAKLU` solver. It will be much faster and you won’t see all those solver warnings.

Done :)

kratman (approved):

PR #4820, opened 2025-02-06
Strip notebook output and switch to testing notebooks with docs

kratman:
# Description

Changes:
- Strips notebook output for easier review. Pre-commit will remove notebook output so that review are easier and unmodified notebooks will not show up as changes in PRs.
- Switches from nbmake and nbsphinx to myst_nb
- Runs notebooks in documentation instead of in the example notebooks job

Breaking changes:
- myst_nb is not compatible with the prolog message used for Google Colab links in the notebooks. I think we should probably just remove the Google Colab parts, but that is not done here.

# Key checklist:

## Further checks:


kratman:
The hold-up on this one is a timeout in the read-the-docs build from running all the notebooks. I have a workaround for it, but have not pushed it yet

PR #4825, opened 2025-02-07
Creates a 'calc_esoh' property in the BaseModel

### LINKED ISSUE #4619: Make `calc_esoh` a model attribute

pipliggins:
Creates a 'cal_esoh' property in the model class, set to False for the BaseModel, and True for the lithium ion BaseModel.
Allows the logic for whether eSOH variables should be calculated to be stored in the model, while allowing `Simulation.solve()` to override the model setting (i.e. set calc_esoh to False) as is the current behaviour.

Fixes #4619

### DIFF
--- a/src/pybamm/models/base_model.py
+++ b/src/pybamm/models/base_model.py
@@ -431,7 +431,7 @@ class BaseModel:
 
     @property
     def calc_esoh(self):
-        "Whether to include eSOH variables in the summary variables."
+        """Whether to include eSOH variables in the summary variables."""
         return self._calc_esoh
 
     def get_parameter_info(self, by_submodel=False):
--- a/src/pybamm/models/base_model.py
+++ b/src/pybamm/models/base_model.py
@@ -80,6 +80,9 @@ class BaseModel:
         self.is_discretised = False
         self.y_slices = None
 
+        # Non-lithium ion models shouldn't calculate eSOH parameters
+        self._calc_esoh = False
+
     @classmethod
     def deserialise(cls, properties: dict):
         """
@@ -425,6 +428,11 @@ class BaseModel:
         if self._input_parameters is None:
             self._input_parameters = self._find_symbols(pybamm.InputParameter)
         return self._input_parameters
+
+    @property
+    def calc_esoh(self):
+        "Whether to include eSOH variables in the summary variables."
+        return self._calc_esoh
 
     def get_parameter_info(self, by_submodel=False):
         """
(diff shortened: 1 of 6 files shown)


src/pybamm/models/base_model.py
```diff
@@ -426,6 +429,11 @@ def input_parameters(self):
             self._input_parameters = self._find_symbols(pybamm.InputParameter)
         return self._input_parameters
 
+    @property
+    def calc_esoh(self):
+        "Whether to include eSOH variables in the summary variables."
```
kratman, line 434: Minor but doc strings should be triple quotes
```suggestion
        """Whether to include eSOH variables in the summary variables."""
```

src/pybamm/models/full_battery_models/lithium_ion/base_lithium_ion_model.py
```diff
@@ -93,6 +96,24 @@ def default_quick_plot_variables(self):
                 "Voltage [V]",
             ]
 
+    @property
+    def calc_esoh(self):
+        "Whether to include eSOH variables in the summary variables."
```
kratman, line 101:
```suggestion
        """Whether to include eSOH variables in the summary variables."""
```

src/pybamm/models/full_battery_models/lithium_ion/base_lithium_ion_model.py
```diff
@@ -93,6 +96,24 @@ def default_quick_plot_variables(self):
...
+        if (
+            self.options["particle phases"] not in ["1", ("1", "1")]
+            or self.options["working electrode"] != "both"
+        ):
+            self._calc_esoh = False
+        return self._calc_esoh
+
+    @calc_esoh.setter
+    def calc_esoh(self, value: bool):
+        "Set whether to include eSOH variables in the summary variables."
```
kratman, line 111:
```suggestion
        """Set whether to include eSOH variables in the summary variables."""
```

kratman (commented):
Everything else looks fine

kratman (approved):

PR #4826, opened 2025-02-07
Added InputParameter support in PyBamm experiments

### LINKED ISSUE #4799: Use InputParameter in pybamm experiment
### Description

Allow using input parameters in a pybamm experiment, e.g. 
`pybamm.step.current(pybamm.InputParameter("I_app"), termination="2.5V")`

### Motivation

Firstly, this allows faster sweeping over different conditions

More importantly, we would be able to set up simulations to re-use steps instead of re-discretizing each step. Currently, it does avoid re-discretizing steps if they're exactly the same, but there are some edge cases where it incorrectly thinks steps are different. It also rediscretizes steps even if they have the same structure but different values, which should not be necessary (e.g. "Discharge at 1C until 2.5V" and "Charge at 2C until 4.2V" have the same model structure and should not need to be discretized twice).

### Possible Implementation

Create relevant input parameters when building the simulation experiment, then pass in the appropriate values as inputs at solve time

### Additional context

_No response_

Rishab87:
# Description

Added InputParameter support in experiments, now you can pass them something like this:
```python
current_step = pybamm.step.current(1, input_parameters=pybamm.InputParameter("I_app"), termination="2.5 V")
    input_parameters={"I_app": 1},
    termination="2.5 V",
    operating_conditions=[current_step],
)
```

Fixes #4799 

## Type of change

# Key checklist:

## Further checks:


valentinsulzer (changes requested):
Thanks for taking this on but this isn't quite what I had in mind, the interface should be:
```
step = pybamm.step.current(pybamm.InputParameter("I_app"), termination="2.5 V")
experiment = pybamm.Experiment([step])
sim = pybamm.Simulation(..., experiment=experiment)
sim.solve(inputs={"I_app": 1})
```

Rishab87:
made changes, now user can pass input params like this:
```python
step = pybamm.step.current(pybamm.InputParameter("I_app"), termination="2.5 V")
experiment = pybamm.Experiment([step])
sim = pybamm.Simulation(..., experiment=experiment)
sim.solve(inputs={"I_app": 1})
```

Rishab87:
Just a follow up anything else I need to change?

src/pybamm/experiment/step/base_step.py
```diff
@@ -71,6 +71,7 @@ def __init__(
         start_time=None,
         description=None,
         direction: str | None = None,
+        input_parameters: dict[str, any] | None = None,
```
MarcBerliner, line 74: The `input_parameters` do not need to be an input to `BaseStep` since they'll come in through `value`

src/pybamm/experiment/step/base_step.py
```diff
@@ -79,8 +80,8 @@ def __init__(
                 f"Invalid direction: {direction}. Must be one of {potential_directions}"
             )
         self.input_duration = duration
-        self.input_duration = duration
         self.input_value = value
+        self.input_parameters = input_parameters if input_parameters is not None else {}
```
MarcBerliner, line 84: Same as above, we don't need to store the `input_parameters`

src/pybamm/experiment/step/base_step.py
```diff
@@ -236,7 +237,7 @@ def basic_repr(self):
         and temperature, which are the variables involved in processing the model. Also
         used for hashing.
         """
-        return f"Step({self.hash_args})"
+        return f"{self.__class__.__name__}:value={self.input_value},duration={self.input_duration},termination={self.termination}"
```
MarcBerliner, line 240: Why did this change?

src/pybamm/experiment/step/base_step.py
```diff
@@ -393,7 +394,14 @@ def value_based_charge_or_discharge(self):
...
             inpt = {"start time": 0}
-            init_curr = self.value.evaluate(t=0, inputs=inpt).flatten()[0]
+            merged_inputs = {}
+            if hasattr(self, "input_parameters"):
+                merged_inputs.update(self.input_parameters)
+            merged_inputs.update(inpt)
+            if isinstance(self.value, pybamm.InputParameter):
+                init_curr = 0
+            else:
+                init_curr = self.value.evaluate(t=0, inputs=merged_inputs).flatten()[0]
```
MarcBerliner, line 404: This block of code will always register steps with `InputParameter` values as `"rest"` steps because of
```python
if isinstance(self.value, pybamm.InputParameter):
    init_curr = 0
```
This part is a bit tricky because the direction is really indeterminate until we pass the `inputs` dict to `sim.solve()`. The major place that uses the `step.direction` is `pybamm.step.VoltageTermination.get_event()`. Any thoughts on the best way to handle this @aabills?

MarcBerliner (changes requested):
@Rishab87 thanks for putting this PR together! I left a few comments

src/pybamm/experiment/step/base_step.py
Rishab87: yeah sorry for that

src/pybamm/experiment/step/base_step.py
aabills: I think so, you should double check though.

src/pybamm/experiment/step/base_step.py
```diff
@@ -70,6 +70,7 @@ def __init__(
         tags=None,
         start_time=None,
         description=None,
+        operator=None,
```
aabills, line 73: Operator needs to go into the termination, not the step.

aabills (changes requested):

aabills:
The `operator` is only a property of the termination function. For example, if the termination function is "2.5 V", whether that means that the step terminates when the battery goes above 2.5 V or below 2.5 V is not well defined. The workaround for this in the past was that we could detect whether the battery was charging or discharging based on the current, and to assume that the termination would be hit if the battery goes above the voltage during charge or below during discharge. This is actually the only reason we ever needed to know the direction anyway. In the case that it is an input parameter, we can no longer calculate whether the battery is charging or discharging a priori, so instead, we have to provide an operator to decide the event. 

So for this PR, we need to ensure that any termination has an operator, because we cannot make the assumption. We probably want to keep the string parsing available, so a nice syntax would be "{operator} {value} {unit}", e.g. "< 2.5 V". But the operator **must** be in the termination, not the step.

src/pybamm/experiment/step/base_step.py
Rishab87: double checked it

src/pybamm/experiment/step/base_step.py
Rishab87: yeah got it

src/pybamm/experiment/step/base_step.py
```diff
@@ -389,8 +392,10 @@ def update_model_events(self, new_model):
     def value_based_charge_or_discharge(self):
         """
         Determine whether the step is a charge or discharge step based on the value of the
-        step
+        step. If an operator is provided, the step direction is not used, so we return None.
         """
+        if isinstance(self.value, pybamm.InputParameter):
+            return None
         if isinstance(self.value, pybamm.Symbol):
```
aabills, line 399: I think this misses the case that `self.value` is a function of an input parameter, i.e. I_coeff*t, where I_coeff is an input parameter. We should handle that case.

aabills, line 399: Make it:

```
if cond:
    do something
elif cond:
    do something else
else:
    do something else
```

for readability

src/pybamm/experiment/step/base_step.py
```diff
@@ -582,3 +587,21 @@ def _convert_electric(value_string):
...
+    term_str = term_str.strip()
+    operator = None
+    remaining = term_str
+    # Check if the string starts with '<' or '>'
+    if term_str and term_str[0] in ("<", ">"):
+        operator = term_str[0]
+        remaining = term_str[1:].strip()
+    remaining = remaining.replace(" ", "")
+    typ, val = _convert_electric(remaining)
+    if isinstance(value, pybamm.InputParameter) and operator is None:
```
aabills, line 603: See above about functions of input parameters.

src/pybamm/experiment/step/base_step.py
Rishab87: added this case

Rishab87: done

src/pybamm/experiment/step/base_step.py
Rishab87: yeah handled callable here too

Rishab87:
@aabills thanks for the review I've added the callable case and also added few test cases accordingly to ensure full coverage

src/pybamm/experiment/step/base_step.py
```diff
@@ -389,9 +392,19 @@ def update_model_events(self, new_model):
     def value_based_charge_or_discharge(self):
         """
         Determine whether the step is a charge or discharge step based on the value of the
-        step
+        step. If an operator is provided, the step direction is not used, so we return None.
         """
-        if isinstance(self.value, pybamm.Symbol):
+        if isinstance(self.value, pybamm.InputParameter):
+            return None
+        elif callable(self.value):
```
aabills, line 399: close. This is my fault for being a bit unclear.

The case I'm trying to catch here is not that it's a callable function, but that it's a `pybamm.Symbol` with an input parameter as a child. See https://docs.pybamm.org/en/latest/source/examples/notebooks/expression_tree/expression-tree.html for information on the pybamm expression tree. I think we want to use the following code to check that:
```
leaves = self.value.post_order(filter=lambda node: len(node.children) == 0)
contains_input_parameter = any(isinstance(leaves, pybamm.InputParameter))
```

aabills (changes requested):
So close! Just one more change and this is good from my side

src/pybamm/experiment/step/base_step.py
Rishab87: done added a check to see if pybamm symbol contains input params

Rishab87:
@aabills I've changed the code and added a check to see if symbol has input params

src/pybamm/experiment/step/base_step.py
aabills: perfect, thanks

aabills:
Looks good! Fix coverage and you're good to go from my side. @MarcBerliner @valentinsulzer anything else from y'all?

aabills:
Oh and don't forget to add something to the CHANGELOG.md

aabills:
Also fix this failing integration test.

valentinsulzer (approved):
looks good to me

aabills (approved):

Rishab87:
@aabills added my PR to changelog and tests for full coverage. Also fixed the integration test

MarcBerliner (approved):
Thanks @Rishab87! Looks great

PR #4827, opened 2025-02-07
Enable cylindrical and platelet particle shape options

### LINKED ISSUE #1072: Non-spherical particles
**Summary**
 Include other geometries for electrode particles

 **Motivation**
 Tomography and SEM experiments agree that graphite particles are not spherical.

**Additional context**
I have devised a model for graphite particles that are coin-shaped, with (de)intercalation at the edge 
of the coin, as opposed to spherical, but have not had a chance to test it yet.

d-cogswell:
# Description

Enables `cylindrical` and `platelet` as model options for `particle shape`. A 2-tuple can also now be provided to assign different particle shapes at the anode and cathode. Cylindrical particles are discretized in cylindrical polar coordinates and given an area/volume ratio of 2/R. Platelet particles are discretized in cartesian coordinates and given an area/volume ratio of 1/R. 

The coordinate change is applied to copies of `r_n` and `r_p` to avoid the bad programming practice of modifying global variables in `pybamm.standard_spatial_vars`. This requires a minimal number of code changes while also insuring that changes made during the creation of a model are not accidentally inherited by subsequent models.

The model options documentation for `particle shape` has been updated.

Fixes #1072

## Type of change

# Key checklist:

## Further checks:


DrSOKane (approved):
In terms of the coding, this is flawless.

In terms of how useful this upgrade will be, I am not sure. While supervising a summer project with @katiezzzzz, I discovered that GITT measurements of diffusivity presuppose the particle to have a particular shape (usually spherical), which limits the utility of this feature. @brosaplanella I believe you looked into this but not sure if you ever published your findings.

Despite this, I would like to thank @d-cogswell for the code and strongly recommend its inclusion in PyBaMM. The question around diffusivity could be the subject of future work, by yourself of another PyBaMM user who can now hit the ground running.

valentinsulzer (commented):
I'm not sure this would work as is since `standard_spatial_vars.r_n` will still be used in other places with the spherical coordinates (e.g. in the geometry, and in `r_average`). Have you tested that this works?

valentinsulzer:
See #4857

d-cogswell:
> I'm not sure this would work as is since `standard_spatial_vars.r_n` will still be used in other places with the spherical coordinates (e.g. in the geometry, and in `r_average`). Have you tested that this works?

AFter

d-cogswell:
I'm working on a modification that uses factory functions for `r_n` and `r_p` instead of global variables.

valentinsulzer:
I think it would be easier to just do #4857

d-cogswell:
It looks like #4857 is also relying on global variables. I think that means that if you create a model, and then then create a second model with different coordinates, the first model is also changed?

d-cogswell:
I agree, #4857 is a better solution. With my approach, there isn't a good solution for the calls to `pybamm.standard_spatial_vars` from the Mesh classes.

PR #4834, opened 2025-02-11
rename CrateTermination to CRateTermination because the step is CRate

aabills:
# Description

I noticed today that to use a step for c-rate you call `pybamm.step.CRate` but to create a termination you use `pybamm.step.CrateTermination`. This is inconsistent,  so this pr fixes it.

Fixes # (issue)

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,12 @@
 
 ## Features
 
-- Deprecated `CrateTermination` and renamed it to `CRateTermination`. ([#4834](https://github.com/pybamm-team/PyBaMM/pull/4834))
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Features
 
+- Deprecated `CrateTermination` and renamed it to `CRateTermination`. ([#4834](https://github.com/pybamm-team/PyBaMM/pull/4834))
```
agriyakhetarpal, line 5: I think this should be under "Breaking Changes".

agriyakhetarpal (commented):
Thanks, @aabills! Could you please add a test for the warning as the coverage has apparently dropped?

PR #4839, opened 2025-02-12
add option to skip experiment steps

aabills:
# Description

Adds an option to skip experiment steps. This is useful in CCCV if you start with too high of an initial SOC to run the CC step.

Fixes # (issue)

## Type of change

# Important checks:

### DIFF
--- a/src/pybamm/experiment/step/base_step.py
+++ b/src/pybamm/experiment/step/base_step.py
@@ -55,6 +55,9 @@ class BaseStep:
         A description of the step.
     direction : str, optional
         The direction of the step, e.g. "Charge" or "Discharge" or "Rest".
+    skip_ok : bool, optional
+        If True, the step will be skipped if it is infeasible at the initial conditions.
+        Default is False.
     """
 
     def __init__(
--- a/src/pybamm/experiment/step/base_step.py
+++ b/src/pybamm/experiment/step/base_step.py
@@ -68,6 +68,7 @@ class BaseStep:
         start_time=None,
         description=None,
         direction: str | None = None,
+        skip_ok: bool = False,
     ):
         potential_directions = ["charge", "discharge", "rest", None]
         if direction not in potential_directions:
@@ -77,6 +78,7 @@ class BaseStep:
         self.input_duration = duration
         self.input_duration = duration
         self.input_value = value
+        self.skip_ok = skip_ok
         # Check if drive cycle
         is_drive_cycle = isinstance(value, np.ndarray)
         is_python_function = callable(value)
@@ -213,6 +215,7 @@ class BaseStep:
             start_time=self.start_time,
             description=self.description,
             direction=self.direction,
+            skip_ok=self.skip_ok,
         )
 
     def __str__(self):
--- a/src/pybamm/experiment/step/base_step.py
+++ b/src/pybamm/experiment/step/base_step.py
@@ -57,7 +57,7 @@ class BaseStep:
         The direction of the step, e.g. "Charge" or "Discharge" or "Rest".
     skip_ok : bool, optional
         If True, the step will be skipped if it is infeasible at the initial conditions.
-        Default is False.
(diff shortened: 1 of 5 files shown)


src/pybamm/experiment/step/base_step.py
```diff
@@ -68,6 +68,7 @@ def __init__(
         start_time=None,
         description=None,
         direction: str | None = None,
+        skip_ok: bool = False,
```
MarcBerliner, line 71: Can you add docstrings

src/pybamm/simulation.py
```diff
@@ -877,19 +878,28 @@ def solve(
...
-                                "])"
-                            )
+                            if step.skip_ok:
+                                pybamm.logger.warning(
+                                    f"Step '{step_str}' is infeasible at initial conditions, but skip_ok is True. Skipping step."
+                                )
+                                continue
+                            else:
+                                raise pybamm.SolverError(
+                                    f"Step '{step_str}' is infeasible "
```
MarcBerliner, line 893: Can you edit this message to mention that adding `skip_ok = True` to the experiment step will let the simulation continue without error

src/pybamm/simulation.py
```diff
@@ -877,19 +878,28 @@ def solve(
...
+                                    "round brackets should be used to indicate this, "
+                                    "e.g.:\n pybamm.Experiment([(\n"
+                                    "\tDischarge at C/5 for 10 hours or until 3.3 V,\n"
+                                    "\tCharge at 1 A until 4.1 V,\n"
+                                    "\tHold at 4.1 V until 10 mA\n"
+                                    "])"
+                                )
                         else:
                             this_cycle = self.experiment.cycles[cycle_num - 1]
                             raise pybamm.SolverError(
```
MarcBerliner, line 905: If all experiments in a cycle have `skip_ok = True`, should we still raise this error?

MarcBerliner (changes requested):
Looks good, just some minor comments.

src/pybamm/simulation.py
aabills: I kinda think we should raise an error in that case, but maybe we should raise a slightly different error. What do you think?

MarcBerliner: Yeah, it's not obvious to me how we should handle this. If every step in a cycle is skipped it suggests something is probably wrong, but on the other hand, it is what they're asking for. I think raising a slightly different error message is reasonable

aabills: I think we can figure that out in conjunction with #4840. This is intended as more of a patch

MarcBerliner (dismissed):

src/pybamm/experiment/step/base_step.py
```diff
@@ -55,6 +55,9 @@ class BaseStep:
         A description of the step.
     direction : str, optional
         The direction of the step, e.g. "Charge" or "Discharge" or "Rest".
+    skip_ok : bool, optional
+        If True, the step will be skipped if it is infeasible at the initial conditions.
+        Default is False.
```
valentinsulzer, line 60: Do we actually need this option or can it just always be True?

valentinsulzer: Ignore where the comment is - meant this as more of a general comment

aabills: my instinct is to not change things unless its very obvious, but if you think we should eliminate it or make it default true that would be fine with me.

valentinsulzer: let's do default true

kratman:
@aabills I made this a draft so that it was not accidentally merged while you address Valentin's comments

aabills:
Thanks yeah this is going to take a bit longer because it substantially changes behavior in a few tests.

valentinsulzer (approved):

PR #3494, opened 2023-11-02
Improve reliability of simulations when using `CasadiSolver`

agriyakhetarpal:
# Description

Closes #3415; wraps and calls a fixed random seed set to the name of the `Simulation` class when using the `CasadiSolver` or `CasadiAlgebraicSolver` solvers for a model, which are susceptible to the initial conditions and sometimes fail randomly. This situation has been fixed for the integration tests and for the benchmarks earlier, this PR ensures that it will be fixed for general-purpose use-cases of the `Simulation` class.

Setting a random seed has been shown to bring the failure rate to zero on repeated executions of the half-cell models Jupyter notebook: https://github.com/agriyakhetarpal/PyBaMM/actions/runs/6709408022, https://github.com/agriyakhetarpal/PyBaMM/actions/runs/6729021051.

## Type of change

Please add a line in the relevant section of [CHANGELOG.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CHANGELOG.md) to document the change (include PR #) - note reverse order of PR #s. If necessary, also add to the list of breaking changes.

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (back-end change that speeds up the code)
- [x] Bug fix (non-breaking change which fixes an issue)

# Key checklist:

- [x] No style issues: `$ pre-commit run` (or `$ nox -s pre-commit`) (see [CONTRIBUTING.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CONTRIBUTING.md#installing-and-using-pre-commit) for how to set this up to run automatically when committing locally, in just two lines of code)
- [x] All tests pass: `$ python run-tests.py --all` (or `$ nox -s tests`)
- [x] The documentation builds: `$ python run-tests.py --doctest` (or `$ nox -s doctests`)

You can run integration tests, unit tests, and doctests together at once, using `$ python run-tests.py --quick` (or `$ nox -s quick`).

## Further checks:

- [x] Code is commented, particularly in hard-to-understand areas
- [ ] Tests added that prove fix is effective or that feature works

PR #4847, opened 2025-02-13
Fix a bug where the usage of the CasADi solver breaks reproducibility

### LINKED ISSUE #3415: Scheduled tests failing while running example tests on python 3.8 (Ubuntu)

agriyakhetarpal:
# Description

This is an issue that I highlighted during our presence at the PyBaMM conference around a week ago: in #3494 (which closed #3415), I introduced a bug wherein a NumPy random seed was set when using the `CasadiSolver` to solve a simulation. It is discouraged to add a random seed in library/module code – it breaks reproducibility by overriding randomness that may have been set by user-level code in researchers' scripts and simulations (it is acceptable for test code, though).

As the PR was supposed to fix stochastic behaviour for the hall-cell model, I have moved the random seed to the notebook instead which used to fail at that time in the repository's development. Usually, this should be present for all notebooks, but one should be enough for now to keep the diff minimal.

Another potential fix is to look at the `globals()` and `locals()` to find if a seed has been set already and use that, but that seems to be a bit over-engineered.

## Type of change

## Important checks

## Additional context

[Stop Using NumPy's Global Random Seed](https://arc.net/l/quote/fnaijdko) – while the article mentions using `np.random.default_rng()` is a better way, `np.random.seed()` should be sufficient for now. The rationale is that we don't want to be one of those packages that resets the global random seed set by a user at the top of their script (or by another library that sets it).

### DIFF
--- a/docs/source/examples/notebooks/models/half-cell.ipynb
+++ b/docs/source/examples/notebooks/models/half-cell.ipynb
@@ -39,6 +39,26 @@
   },
   {
    "cell_type": "markdown",
+   "id": "2b2e62ad",
+   "metadata": {},
+   "source": [
+    "First, we introduce a random seed to ensure reproducibility of the results before running the simulation:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "7f30409c",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import numpy as np\n",
+    "\n",
+    "np.random.seed(42)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
    "id": "f372aa29",
    "metadata": {},
    "source": [
--- a/src/pybamm/simulation.py
+++ b/src/pybamm/simulation.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 import pickle
 import pybamm
 import numpy as np
-import hashlib
 import warnings
 from functools import lru_cache
 from datetime import timedelta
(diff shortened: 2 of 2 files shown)


agriyakhetarpal:
This is a user-facing change as it fixes is/was a slight bug in the solver functionality, but I don't think the fix warrants adding a CHANGELOG entry. What do you think, @kratman?

kratman:
I have not looked at the code yet, but from your description I don't think it would need one

Saransh-cpp (approved):
I remember this. Thanks for the fix, @agriyakhetarpal!

agriyakhetarpal:
I re-triggered the failing Windows test, as it looks like a flake that has been showing up a bit more often recently:

`FAILED tests/unit/test_solvers/test_idaklu_solver.py::TestIDAKLUSolver::test_multiple_inputs - ValueError: Unknown exception`

kratman:
> I re-triggered the failing Windows test, as it looks like a flake that has been showing up a bit more often recently:
> 
> `FAILED tests/unit/test_solvers/test_idaklu_solver.py::TestIDAKLUSolver::test_multiple_inputs - ValueError: Unknown exception`

Yeah I have not fully debugged that issue yet. It started appearing after I enabled the IDAKLU-Jax tests on windows. The issue was probably there for a while, but not caught. I am hoping I end up tracking it down as I keep working on the PyBaMM solvers build/test issues

Issue #4833, opened 2025-02-10
Deprecate number of electrodes in parallel to make a cell


aabills:
The "Number of electrodes connected in parallel to make a cell" parameter is confusing and not always used properly. For example, it is not used to calculate the area of the current collector: https://github.com/pybamm-team/PyBaMM/blob/74192198058fd914afd9130dbc201fcc673fab20/src/pybamm/parameters/geometric_parameters.py#L44, which is used to calculate the "total lithium in the electrolyte": https://github.com/pybamm-team/PyBaMM/blob/74192198058fd914afd9130dbc201fcc673fab20/src/pybamm/parameters/lithium_ion_parameters.py#L116

We could go through and try and find and fix all of these, or we could just deprecate this parameter. I think we should do the latter.

valentinsulzer:
Happy to remove this in favor of just scaling the electrode area properly. How about the number of cells in series? It only really applies to lead-acid batteries (to get to 12V when the OCV is ~2V)

aabills:
Should we just switch to "Electrode area [m2]" instead?

ejfdickinson:
I think that here the best development would be disentangling the parameterisation of the electrical model (relation of cell current to 1D macroscopic current density), and any parameterisation of the actual layout of the cell, for thermal modelling or other purposes.

Even if you can count your electrodes and their coatings, you need to make some assumption about what the 'out-of-plane' area of the SPM/SPMe/DFN should be, because negative electrode, positive electrode and separator all have different areas and while basically only the facing area counts, there can be edge effects.. So I'd recommend a simple `"Electrode area [m2]"` for intensive -> extensive scaling purposes, such that `Cell current [A] = Electrode geometric current density [A.m-2] * Electrode area [m2]`, and have any other more complicated parameters (height, width, number of layers, shape, funny stuff) exclusively for more complicated purposes.

valentinsulzer:
> It's not obvious to new PyBaMM users how to account for multiple layers in a cell. For example, Chen et al. and the papers based on it (O'Kane et al., O'Regan et al.) handle the double-coated electrode by doubling the electrode width, but other papers use different conventions.

I think removing the "number of electrodes in parallel" will help with this, as we can make it clear that the electrode width/area corresponds to a single anode/separator/cathode layer

[...]

PR #4802, opened 2025-01-26
Switch to hatch-vcs for versioning

vidipsingh:
# Description
This PR implements `hatch-vcs` for versioning, enabling automatic versioning based on Git tags.

Related to: #4742 

## Type of change
- [x] New feature (non-breaking change which adds functionality)
- [ ] Optimization (back-end change that speeds up the code)
- [ ] Bug fix (non-breaking change which fixes an issue)

# Key checklist:

- [x] No style issues: `$ pre-commit run` (or `$ nox -s pre-commit`) (see [CONTRIBUTING.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CONTRIBUTING.md#installing-and-using-pre-commit) for how to set this up to run automatically when committing locally, in just two lines of code)
- [x] All tests pass: `$ python -m pytest` (or `$ nox -s tests`)
- [x] The documentation builds: `$ python -m pytest --doctest-plus src` (or `$ nox -s doctests`)

You can run integration tests, unit tests, and doctests together at once, using `$ nox -s quick`.

## Further checks:

- [ ] Code is commented, particularly in hard-to-understand areas
- [ ] Tests added that prove fix is effective or that feature works

PR #4854, opened 2025-02-15
Deprecate number of electrodes in parallel parameter

vidipsingh:
# Description

This PR deprecates the `n_electrodes_parallel` parameter due to its inconsistent usage across the codebase. Area calculations in the thermal models have been updated to use the total effective area instead.

Fixes: #4833 

## Type of change

# Important checks:


src/pybamm/parameters/electrical_parameters.py
```diff
@@ -37,7 +37,7 @@ def _set_parameters(self):
             "Current function [A]", {"Time [s]": pybamm.t}
         )
         self.current_density_with_time = self.current_with_time / (
-            self.n_electrodes_parallel * self.geo.A_cc
+            self.n_electrodes_pair * self.geo.A_cc
```
valentinsulzer, line 40: why is this kept here?

valentinsulzer (commented):
Need to decide what to do about Ai2020 and Sulzer2019 - perhaps just scaling the electrode width. Also this will need a clear deprecation message (we could check specifically for when people give a value that is not equal to 1 and raise a warning in that case)

src/pybamm/parameters/electrical_parameters.py
vidipsingh: @valentinsulzer should I completely remove this, or is there another way I should handle this? I wasn't sure how to approach this, so I changed the name from `n_electrodes_parallel` to `n_electrodes_pair`. Would appreciate your guidance on the right approach here.

vidipsingh:
> Need to decide what to do about Ai2020 and Sulzer2019 - perhaps just scaling the electrode width. Also this will need a clear deprecation message (we could check specifically for when people give a value that is not equal to 1 and raise a warning in that case)

@valentinsulzer Thanks for the suggestion! For Ai2020 and Sulzer2019, I'll look into scaling the electrode width as you mentioned.

Regarding the deprecation message, I can add a check for when the `n_electrodes_parallel` value is not equal to 1 and raise a warning to inform users about the change. Could you please guide me on where exactly these messages should be added in the code?

DrSOKane:
I see why you want to do this but it can be a bit of a shock to people unfamiliar new to PyBaMM. I was just mentoring someone who was struggling to get the correct capacity and didn't realize that they needed to double the electrode length (compared to what it was in the paper) to account for it being double-sided.

aabills:
Thanks for taking this on, @vidipsingh . I can't help but notice that there are a bunch of changes made here that are unrelated to this issue. Perhaps they are related to #4802 ?  It would be good to start this from a clean branch to ensure that nothing extraneous is considered here.

aabills:
> I see why you want to do this but it can be a bit of a shock to people new to PyBaMM. I was just mentoring someone who was struggling to get the correct capacity and didn't realize that they needed to double the electrode length (compared to what it was in the paper) to account for it being double-sided, which implies the authors of that paper did it a different way than Chen et al.

My argument is just that we should either make everything consistent with this parameter, or deprecate it, of which I thought deprecation made more sense. Maybe we should debate this at the original issue, #4854

vidipsingh:
> Thanks for taking this on, @vidipsingh . I can't help but notice that there are a bunch of changes made here that are unrelated to this issue. Perhaps they are related to #4802 ? It would be good to start this from a clean branch to ensure that nothing extraneous is considered here.

Thanks for pointing that out, @aabills. I did start from a clean branch, so I'm not entirely sure how the changes from that PR ended up here. What would you suggest I do at this point? If I start from a new branch, I'll likely need to close this PR and raise a new one.

aabills:
> > Thanks for taking this on, @vidipsingh . I can't help but notice that there are a bunch of changes made here that are unrelated to this issue. Perhaps they are related to #4802 ? It would be good to start this from a clean branch to ensure that nothing extraneous is considered here.
> 
> Thanks for pointing that out, @aabills. I did start from a clean branch, so I'm not entirely sure how the changes from that PR ended up here. What would you suggest I do at this point? If I start from a new branch, I'll likely need to close this PR and raise a new one.

Seems reasonable to me, make sure you check out discussion in #4833 first.

vidipsingh:
> > > Thanks for taking this on, @vidipsingh . I can't help but notice that there are a bunch of changes made here that are unrelated to this issue. Perhaps they are related to #4802 ? It would be good to start this from a clean branch to ensure that nothing extraneous is considered here.
> > 
> > 
> > Thanks for pointing that out, @aabills. I did start from a clean branch, so I'm not entirely sure how the changes from that PR ended up here. What would you suggest I do at this point? If I start from a new branch, I'll likely need to close this PR and raise a new one.
> 
> Seems reasonable to me, make sure you check out discussion in #4833 first.

Sure, will check that.

PR #4869, opened 2025-02-21
Change fundamental SEI variable from thickness to concentration: attempt 2

### LINKED ISSUE #3006: [Bug]: LLI due to SEI incorrect if loss of active material is enabled

DrSOKane:
# Description

Previously, lithium was not conserved if both SEI growth and loss of active material were enabled. To fix this, the fundamental variable to be solved for is now SEI concentration, not thickness. This has the added bonus of simplifying the SEI on cracks model.

I closed #3171 and made a new one simply because PyBaMM has changed so much since then!

One major difference between #3171 and this PR is that this time around, I also deleted the parameter `"Initial SEI thickness [m]"` and replaced it with `"Initial SEI concentration [mol.m-3]"`. I did this partly to make the initial conditions less complicated, and partly because it's easier to measure the amount of lithium consumed by formation than it is to measure or even estimate the initial thickness. I assumed a value of 38.34 [mol.m-3], based on Ramadass _et al._ (2004) and Safari _et al._ (2009). The exception is `"OKane2022"`, for which I used the exact value (to 4 significant figures anyway!)

Fixes #3006


DrSOKane:
The reason the tests are failing is because none of the half-cell parameter sets have a value for "Initial SEI concentration [mol.m-2]" on the lithium metal electrode. [Liu et al. (2017)](https://iopscience.iop.org/article/10.1149/2.0381709jes/meta) propose values for this and other SEI parameters on Li metal electrodes. How do you feel about me going ahead using those and citing the paper?

rtimms:
In general, I am against adding parameters that aren't in the original paper, but since we already added a bunch of extra degradation parameters in these files it is OK to add this once too. Going forward we should be more strict on this.

rtimms:
@DrSOKane is this ready to be reviewed?

DrSOKane:
> @DrSOKane is this ready to be reviewed?

Yes please! All the commits since removing the reference to Liu and Lu are just updated from develop branch

CHANGELOG.md
```diff
@@ -81,6 +81,8 @@
 
 ## Breaking changes
 
+- Changed fundamental variable for all SEI models from thickness to concentration ([#4869](https://github.com/pybamm-team/PyBaMM/pull/4869))
+- Replaced parameter `"Initial SEI thickness [m]"` with `"Initial SEI concentration [mol.m-3]"` ([#4869](https://github.com/pybamm-team/PyBaMM/pull/4869))
```
kratman: Just a note for later, this needs to be moved up to a current release

src/pybamm/input/parameters/lithium_ion/Ecker2015.py
```diff
@@ -368,15 +368,15 @@ def SEI_limited_dead_lithium_OKane2022(L_sei):
...
+        SEI thickness [m]
     Returns
     -------
     :class:`pybamm.Symbol`
         Dead lithium decay rate [s-1]
     """
 
     gamma_0 = pybamm.Parameter("Dead lithium decay constant [s-1]")
-    L_sei_0 = pybamm.Parameter("Initial SEI thickness [m]")
+    L_sei_0 = pybamm.Scalar(5e-9)
```
rtimms: Is this consistent if the initial SEI concentration changes, or should it be derived from that?

src/pybamm/input/parameters/lithium_ion/OKane2022.py
```diff
@@ -79,15 +79,15 @@ def SEI_limited_dead_lithium_OKane2022(L_sei):
...
+        SEI thickness [m]
     Returns
     -------
     :class:`pybamm.Symbol`
         Dead lithium decay rate [s-1]
     """
 
     gamma_0 = pybamm.Parameter("Dead lithium decay constant [s-1]")
-    L_sei_0 = pybamm.Parameter("Initial SEI thickness [m]")
+    L_sei_0 = pybamm.Scalar(5e-9)
```
rtimms: as above

src/pybamm/models/submodels/interface/sei/sei_growth.py
```diff
@@ -217,56 +223,74 @@ def get_coupled_variables(self, variables):
...
-                dl_cr = variables[f"X-averaged {domain} particle cracking rate [m.s-1]"]
+                roughness = variables[f"X-averaged {domain} electrode roughness ratio"]
             else:
-                l_cr = variables[f"{Domain} particle crack length [m]"]
-                dl_cr = variables[f"{Domain} particle cracking rate [m.s-1]"]
-            spreading = dl_cr / l_cr * (self.phase_param.L_sei_crack_0 - L_sei)
-        else:
-            spreading = 0
+                roughness = variables[f"{Domain} electrode roughness ratio"]
+            a *= roughness - 1  # Replace surface area with crack area
```
rtimms, line 274: can you add a comment explaining this calculation?

rtimms (changes requested):
thanks @DrSOKane looks good, just a few comments

src/pybamm/input/parameters/lithium_ion/Ecker2015.py
DrSOKane: It's a tricky one, because the thickness and concentration are coupled via the surface area to volume ratio, which is different for every cell. For every cell except OKane2022, I derive the initial concentration to be consistent with that in the original Safari et al. (2009) paper. For OKane2022, the SEI is explicitly included, so I chose a value that results in a thickness of 5 nm exactly.

Safari et al. do not consider dead lithium, so I thought it more appropriate to stick with the value from O'Kane et al. Unfortunately, my dead lithium model has never been validated, so a factor of two difference is really not going to matter! The function is only there at all as proof of concept for pure graphite, as opposed to graphite+SiOx.

src/pybamm/input/parameters/lithium_ion/OKane2022.py
DrSOKane: It was difficult deciding whether to make the initial condition a concentration or thickness. Most of the academic literature reports thickness, with [Li et al. 2025](https://www.nature.com/articles/s41467-025-57968-3) being a notable exception. But for industry, concentration will be better if - and this is a big if - the formation losses are known, because the formation loss *is* the initial value.

src/pybamm/models/submodels/interface/sei/sei_growth.py
DrSOKane: Done

rtimms (approved):

PR #4887, opened 2025-03-02
Fixed mypy sp check guidelines

### LINKED ISSUE #3489: Check adherence with `sp-repo-review` guidelines

Rishab87:
# Description
Fixing mypy sp check guidelines

### Some major changes I did:

- In `src/pybamm/expression_tree/binary_operators.py` I've replaced `self.__ class __` in line 131 with new_instance because earlier when we were using  ```self.__ class __``` it showed a third arg was not getting passed:
  ```python
  error: Missing positional argument "right_child" in call to "BinaryOperator"  [call-arg]
  ```
  but this function was always getting called from instance of its child classes which don't need to pass 3 arguments, so i 
  thought it was better to make a new_instance method which can be overrided in child classes
  
- In `src/pybamm/expression_tree/broadcasts.py` I've forced child classes of `Broadcast` to implement `_unary_new_copy` because earlier we were using `self.broadcast_domain` in this function in Broadcast class but it does not have any attribute `self.broadcast_domain`, this function was only getting called by instance of their child classes which has self.broadcast_domain property.
  
- In `tests/unit/test_expression_tree/test_operations/test_copy.py` in `test_symbol_create_copy_new_children` test I've added an extra argument `perform_simplifications=False` to `create_copy` method to ensure 100% coverage of new_instance methods.

Related to #3489

## Type of change

# Important checks:


docs/conf.py
agriyakhetarpal, line 1: I think we can exclude this file from the Mypy checks. It is not a part of PyBaMM's runtime, and maintaining typing rules here seems tedious.

agriyakhetarpal (commented):
Thanks, I've triggered the workflows! I haven't reviewed all the files, because I understand that the PR is still a draft and you have yet to complete the changes.

Checking https://learn.scientific-python.org/development/guides/repo-review/?repo=pybamm-team/pybamm, a few more rules where we have problems have popped up. So we shouldn't mark this PR as one that closes the linked issue. Could you please replace the "Fixes" keyword in the PR description with "Related to"?

It is also possible to add `sp-repo-review` as a pre-commit hook so that new rules will be checked on every commit. I think we should investigate that and add it if running the checks is not too slow. I'm also curious to hear what @Saransh-cpp has to say about this.

Saransh-cpp:
It would be great to have repo-review as a pre-commit hook! @agriyakhetarpal maybe you can create a new independent issue for that?

Rishab87:
I've changed the PR description to related to and I also think it would be great to have a pre-commit hook for ```sp-check-guidelines```, I'll investigate its speed after this PR is done

Rishab87:
I've fixed the remaining errors too

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 - Moved concentration inside x-averaged when calculating LLI due to LAM variables ([#4858](https://github.com/pybamm-team/PyBaMM/pull/4858))
 - Fixed a bug that caused the variable `"Loss of lithium due to {domain} lithium plating"`to have the domain `"current collector"` (should not have any domain at all) if the `"x-average side reactions"` option was set to `"true"`. ([#4844](https://github.com/pybamm-team/PyBaMM/pull/4844))
 - Fixed interpolation bug in `pybamm.QuickPlot` with spatial variables. ([#4841](https://github.com/pybamm-team/PyBaMM/pull/4841))
+- Fixed mypy sp check guidelines ([#4887](https://github.com/pybamm-team/PyBaMM/pull/4887))
```
Saransh-cpp, line 24: Not a user facing change, so this should not be required.
```suggestion
```

pyproject.toml
```diff
@@ -252,6 +252,10 @@ concurrency = ["multiprocessing"]
 ignore_missing_imports = true
 allow_redefinition = true
 disable_error_code = ["call-overload", "operator"]
+strict = false
```
Saransh-cpp, line 255: Given how big this PR is, it would actually be better to spit it into multiple PRs, each one adding a new config option in pyproject.toml.

Saransh-cpp (changes requested):
Thanks for taking this up, @Rishab87! See my comment below about splitting this PR.

pyproject.toml
Rishab87: ok so should I go ahead then and make a PR on one of the config first? or edit this one accordingly?

Saransh-cpp: You can keep this PR for one config, and add other configs in subsequent PRs.

Rishab87: ok I'll create seperate PRs for different configs and then keep this one for the end, I think that would be faster for me

Rishab87: I noticed many of the ```warn_unreachable``` error depends on errors from ```enable_error_code``` config, they're related to each other and their are total of 77 errors out of which 57 are from ```enable_error_code``` so I think creating a seperate PR would still be almost as big as this one, so should I still proceed with it?

agriyakhetarpal: Yes, I think anything that reduces the diff and keeps this PR scoped to a specific change (or a few of them) would be great. Thanks for investigating!

examples/scripts/SPM_compare_particle_grid.py
```diff
@@ -48,7 +49,7 @@
     disc.process_model(model)
 
 # solve model
-solutions = [None] * len(models)
+solutions: Any = [None] * len(models)
```
kratman, line 52: Shouldn't this be something like `list[Any]`?

Rishab87: yess, I'll change that

pyproject.toml
Rishab87: I've created the PR with just ```enable_error_code``` config: https://github.com/pybamm-team/PyBaMM/pull/4891, sorry for the delay lab tests going on

src/pybamm/expression_tree/binary_operators.py
```diff
@@ -224,6 +226,9 @@ def __init__(
         """See :meth:`pybamm.BinaryOperator.__init__()`."""
         super().__init__("**", left, right)
 
+    def _new_instance(self, left: pybamm.Symbol, right: pybamm.Symbol) -> pybamm.Symbol:
+        return Power(left, right)
+
```
agriyakhetarpal, line 231: Why do we need new `_new_instance` methods? Could you document this in an inline comment somewhere, or add it in the PR description (better)?

agriyakhetarpal (commented):
Thanks, @Rishab87! It is not fully clear from some of the changes as to why they were made, could you please highlight them and tell us what was fixed where? Some changes also seem to change the logic in a few places; documenting it is important.

src/pybamm/expression_tree/binary_operators.py
Rishab87: I was getting 
```python
error: Left operand of "and" is always true  [redundant-expr]
```
but I think now this should not be modified, this seems weird that its not showing this for right symbol though its also going through the same checks above

src/pybamm/expression_tree/binary_operators.py
```diff
@@ -33,8 +32,8 @@ def _preprocess_binary(
             raise ValueError("right must be a 1D array")
         right = pybamm.Vector(right)
 
-    # Check both left and right are pybamm Symbols
-    if not (isinstance(left, pybamm.Symbol) and isinstance(right, pybamm.Symbol)):
+    # Check right is pybamm Symbol
+    if not isinstance(right, pybamm.Symbol):
```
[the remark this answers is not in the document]
Rishab87: I was getting 
```python
error: Left operand of "and" is always true  [redundant-expr]
```
but I think now this should not be modified, this seems weird that its not showing this for right symbol though its also going through the same checks above

src/pybamm/expression_tree/binary_operators.py
Rishab87: I've added new_instance because earlier when we were using  __ class __ it showed a third arg was not getting passed:
```python
error: Missing positional argument "right_child" in call to "BinaryOperator"  [call-arg]
```
but this function was always getting called from child classes of BinaryOperator which passed 3 arguments, so i thought it was better to make a new_instance method which can be overirded in child classes, when trying to pass self.name in __ class __ method it should we passed 4 args as the function was getting called from child classes always and they always passed 3 args

so should  I update the PR description or think of some other approach?

examples/scripts/heat_equation.py
```diff
@@ -106,7 +106,7 @@ def T_exact(x, t):
 # Plot ------------------------------------------------------------------------
 x_nodes = mesh["rod"].nodes  # numerical gridpoints
 xx = np.linspace(0, 2, 101)  # fine mesh to plot exact solution
-plot_times = np.linspace(0, 1, 5)
+plot_times: np.ndarray = np.linspace(0, 1, 5)
```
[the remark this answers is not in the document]
Rishab87: changed this

src/pybamm/expression_tree/binary_operators.py
Rishab87: And for broadcasts.py self.broadcast_domain was not defined in parent class they were params of child classes and _unary_new_copy function was getting called from child class instance always so I just overrided this function in child classes, apart from these changes all other changes are mostly type changes

Rishab87:
@agriyakhetarpal Thanks for the review,  I've left comments for the major changes, should I write these major changes in PR description too?

agriyakhetarpal:
> @agriyakhetarpal Thanks for the review,  I've left comments for the major changes, should I write these major changes in PR description too?

Yes, please do!

Rishab87:
@agriyakhetarpal I've updated the PR description

PR #4891, opened 2025-03-04
Added mypy ```enable_error_code``` sp check guideline

Rishab87:
# Description

Added mypy ```enable_error_code``` sp check guideline

Related to #3489, #4887

## Type of change

# Important checks:

### DIFF
--- a/examples/scripts/SPMe_step.py
+++ b/examples/scripts/SPMe_step.py
@@ -43,13 +43,14 @@ while time < end_time:
     time += dt
 
 # plot
-time_in_seconds = solution["Time [s]"].entries
-step_time_in_seconds = step_solution["Time [s]"].entries
-voltage = solution["Voltage [V]"].entries
-step_voltage = step_solution["Voltage [V]"].entries
-plt.plot(time_in_seconds, voltage, "b-", label="SPMe (continuous solve)")
-plt.plot(step_time_in_seconds, step_voltage, "ro", label="SPMe (stepped solve)")
-plt.xlabel(r"$t$")
-plt.ylabel("Voltage [V]")
-plt.legend()
-plt.show()
+if step_solution is not None:
+    time_in_seconds = solution["Time [s]"].entries
+    step_time_in_seconds = step_solution["Time [s]"].entries
+    voltage = solution["Voltage [V]"].entries
+    step_voltage = step_solution["Voltage [V]"].entries
+    plt.plot(time_in_seconds, voltage, "b-", label="SPMe (continuous solve)")
+    plt.plot(step_time_in_seconds, step_voltage, "ro", label="SPMe (stepped solve)")
+    plt.xlabel(r"$t$")
+    plt.ylabel("Voltage [V]")
+    plt.legend()
+    plt.show()
--- a/examples/scripts/heat_equation.py
+++ b/examples/scripts/heat_equation.py
@@ -106,7 +106,7 @@ def T_exact(x, t):
 # Plot ------------------------------------------------------------------------
 x_nodes = mesh["rod"].nodes  # numerical gridpoints
 xx = np.linspace(0, 2, 101)  # fine mesh to plot exact solution
-plot_times = np.linspace(0, 1, 5)
+plot_times: np.ndarray = np.linspace(0, 1, 5)
 
 plt.figure(figsize=(15, 8))
 cmap = plt.get_cmap("inferno")
--- a/examples/scripts/minimal_example_of_lookup_tables.py
+++ b/examples/scripts/minimal_example_of_lookup_tables.py
@@ -34,12 +34,12 @@ df["T"] = df["T"] + 273.15
 D_s_n_data = process_2D("Negative particle diffusivity [m2.s-1]", df)
 
 
-def D_s_n(sto, T):
+def D_s_n_func(sto, T):
     name, (x, y) = D_s_n_data
     return pybamm.Interpolant(x, y, [T, sto], name)
 
 
-parameter_values["Negative particle diffusivity [m2.s-1]"] = D_s_n
+parameter_values["Negative particle diffusivity [m2.s-1]"] = D_s_n_func
 
 k_n = parameter_values["Negative electrode exchange-current density [A.m-2]"]
 
--- a/src/pybamm/experiment/step/base_step.py
+++ b/src/pybamm/experiment/step/base_step.py
@@ -140,7 +140,7 @@ class BaseStep:
             self.value = pybamm.Interpolant(
                 t,
                 y,
-                pybamm.t - pybamm.InputParameter("start time"),
+                pybamm.t - pybamm.InputParameter("start time"),  # type: ignore[arg-type]
                 name="Drive Cycle",
             )
             self.period = np.diff(t).min()
--- a/src/pybamm/expression_tree/binary_operators.py
+++ b/src/pybamm/expression_tree/binary_operators.py
@@ -2,7 +2,6 @@
 # Binary operator classes
 #
 from __future__ import annotations
-import numbers
 
 import numpy as np
 import sympy
@@ -33,8 +32,8 @@ def _preprocess_binary(
             raise ValueError("right must be a 1D array")
         right = pybamm.Vector(right)
 
-    # Check both left and right are pybamm Symbols
-    if not (isinstance(left, pybamm.Symbol) and isinstance(right, pybamm.Symbol)):
+    # Check right is pybamm Symbol
+    if not isinstance(right, pybamm.Symbol):
         raise NotImplementedError(
             f"BinaryOperator not implemented for symbols of type {type(left)} and {type(right)}"
         )
@@ -113,6 +112,9 @@ class BinaryOperator(pybamm.Symbol):
             right_str = f"{self.right!s}"
         return f"{left_str} {self.name} {right_str}"
 
+    def _new_instance(self, left: pybamm.Symbol, right: pybamm.Symbol) -> pybamm.Symbol:
+        return self.__class__(self.name, left, right)  # pragma: no cover
+
     def create_copy(
         self,
         new_children: list[pybamm.Symbol] | None = None,
@@ -127,7 +129,7 @@ class BinaryOperator(pybamm.Symbol):
         children = self._children_for_copying(new_children)
 
         if not perform_simplifications:
-            out = self.__class__(children[0], children[1])
+            out = self._new_instance(children[0], children[1])
         else:
             # creates a new instance using the overloaded binary operator to perform
             # additional simplifications, rather than just calling the constructor
@@ -224,6 +226,9 @@ class Power(BinaryOperator):
         """See :meth:`pybamm.BinaryOperator.__init__()`."""
         super().__init__("**", left, right)
 
+    def _new_instance(self, left: pybamm.Symbol, right: pybamm.Symbol) -> pybamm.Symbol:
+        return Power(left, right)
+
     def _diff(self, variable: pybamm.Symbol):
         """See :meth:`pybamm.Symbol._diff()`."""
         # apply chain rule and power rule
@@ -273,6 +278,9 @@ class Addition(BinaryOperator):
         """See :meth:`pybamm.BinaryOperator.__init__()`."""
         super().__init__("+", left, right)
 
+    def _new_instance(self, left: pybamm.Symbol, right: pybamm.Symbol) -> pybamm.Symbol:
+        return Addition(left, right)
+
     def _diff(self, variable: pybamm.Symbol):
         """See :meth:`pybamm.Symbol._diff()`."""
         return self.left.diff(variable) + self.right.diff(variable)
@@ -299,6 +307,9 @@ class Subtraction(BinaryOperator):
         """See :meth:`pybamm.BinaryOperator.__init__()`."""
 
         super().__init__("-", left, right)
+
+    def _new_instance(self, left: pybamm.Symbol, right: pybamm.Symbol) -> pybamm.Symbol:
+        return Subtraction(left, right)
 
     def _diff(self, variable: pybamm.Symbol):
         """See :meth:`pybamm.Symbol._diff()`."""
@@ -328,6 +339,9 @@ class Multiplication(BinaryOperator):
         """See :meth:`pybamm.BinaryOperator.__init__()`."""
 
         super().__init__("*", left, right)
+
+    def _new_instance(self, left: pybamm.Symbol, right: pybamm.Symbol) -> pybamm.Symbol:
+        return Multiplication(left, right)
 
     def _diff(self, variable: pybamm.Symbol):
         """See :meth:`pybamm.Symbol._diff()`."""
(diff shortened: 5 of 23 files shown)


Saransh-cpp (changes requested):
Thanks, @Rishab87! Could you please merge develop and fix the tests?

Rishab87:
@Saransh-cpp I've merged the main branch and all the tests passes locally, its weird that I haven't changed anything related to serialization still it fails, can you try re-running it? From what I'm able to interpret is failure is most likely due to a timing issue with how the filename is generated

Saransh-cpp (changes requested):
Thanks, @Rishab87! The tests pass now. Could you please comment why each of the change was carried out?

tests/unit/test_solvers/test_solution.py
```diff
@@ -255,7 +255,7 @@ def test_copy_with_computed_variables(self):
 
         assert (
             sol1._variables[k] == sol2._variables[k] for k in sol1._variables.keys()
-        )
+        ) is not None
```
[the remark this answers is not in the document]
Rishab87: aahh sorry I missed that, changed it now

examples/scripts/minimal_example_of_lookup_tables.py
```diff
@@ -34,12 +34,12 @@ def process_2D(name, data):
 D_s_n_data = process_2D("Negative particle diffusivity [m2.s-1]", df)
 
 
-def D_s_n(sto, T):
+def D_s_n_func(sto, T):
```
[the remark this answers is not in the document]
Rishab87: because their was a variable in this file with name D_s_n so mypy gave error function and variable both have same name

src/pybamm/expression_tree/binary_operators.py
```diff
@@ -114,6 +113,9 @@ def __str__(self):
             right_str = f"{self.right!s}"
         return f"{left_str} {self.name} {right_str}"
 
+    def _new_instance(self, left: pybamm.Symbol, right: pybamm.Symbol) -> pybamm.Symbol:
```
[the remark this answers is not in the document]
Rishab87: I've replaced `self.__ class __`  with new_instance because earlier when we were using  ```self.__ class __``` it showed a third arg was not getting passed:
  ```python
  error: Missing positional argument "right_child" in call to "BinaryOperator"  [call-arg]
  ```
  but this function was always getting called from instance of its child classes which don't need to pass 3 arguments, so i 
  thought it was better to make a new_instance method which can be overrided in child classes
  
  I've already added this in the PR description of previous sp-check-guidelines PR, should I add it here too? Or follow some different approach

Rishab87:
@Saransh-cpp Thanks for the review, I've replied to the comments and changed the test

examples/scripts/minimal_example_of_lookup_tables.py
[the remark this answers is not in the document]
Saransh-cpp: I think the motive behind redefining the variable here was to show how one can pass both a constant value and a function as a parameter value. I would say we should revert this and not run mypy on examples at all (see my comment below).

src/pybamm/util.py
```diff
@@ -152,7 +152,7 @@ def search(
             Default is 0.4
         """
 
-        if not isinstance(keys, (str, list)) or not all(
+        if not isinstance(keys, (str, list)) or not all(  # type: ignore[redundant-expr]
             isinstance(k, str) for k in keys
```
Saransh-cpp, line 156: Can we simplify this expression instead of ignoring the warning?

src/pybamm/telemetry.py
```diff
@@ -20,7 +21,7 @@ def capture(**kwargs):  # pragma: no cover
         project_api_key="phc_acTt7KxmvBsAxaE0NyRd5WfJyNxGvBq1U9HnlQSztmb",
         host="https://us.i.posthog.com",
     )
-    _posthog.log.setLevel("CRITICAL")
+    cast(Posthog, _posthog).log.setLevel("CRITICAL")
```
Saransh-cpp, line 24: Why is this `cast` required? I think `Posthog` should return an object of the type `Posthog`?

src/pybamm/solvers/summary_variable.py
```diff
@@ -123,7 +128,7 @@ def __getitem__(self, key: str) -> float | list[float]:
             # return it if it exists
             return self._variables[key]
         elif key == "Cycle number":
-            return self.cycle_number
+            return cast(list[float], self.cycle_number.tolist())
```
Saransh-cpp, line 131: Don't `cast` here. Remember, `tolist()` is an expensive operation as it converts a NumPy `array` (which is C underneath) to a Python `list`. You should instead update the function's signature to include `npt.NDArray` as a possible return type.

src/pybamm/solvers/summary_variable.py
```diff
@@ -69,7 +72,7 @@ def _initialize_for_cycles(self, cycle_summary_variables: list[SummaryVariables]
         self.first_state = None
         self.last_state = None
         self.cycles = cycle_summary_variables
-        self.cycle_number = np.arange(1, len(self.cycles) + 1)
+        self.cycle_number = np.arange(1, len(self.cycles) + 1, dtype=float)
```
Saransh-cpp, line 75: Switching from int to float, or float32 to float64 comes with a memory cost.
```suggestion
        self.cycle_number = np.arange(1, len(self.cycles) + 1)
```

src/pybamm/solvers/summary_variable.py
```diff
@@ -81,25 +84,27 @@ def all_variables(self) -> list[str]:
         Return names of all possible summary variables, including eSOH variables
          if appropriate.
         """
-        try:
+        if self._all_variables is not None:
```
Saransh-cpp, line 87: `None` is a value. This block is checking if the variable even exists; hence, the new logic will not work. Something like this should work, but please check -
```suggestion
        if getattr(self, "_all_variables", None) is not None:
```

src/pybamm/solvers/summary_variable.py
```diff
@@ -81,25 +84,27 @@ def all_variables(self) -> list[str]:
...
+        return self._all_variables
 
     @property
     def esoh_variables(self) -> list[str] | None:
         """Return names of all eSOH variables."""
-        if self.calc_esoh and self._esoh_variables is None:
+        if (
+            self.esoh_solver is not None
+            and self.calc_esoh
+            and self._esoh_variables is None
```
Saransh-cpp, line 106: Why was this added?

examples/scripts/SPMe_step.py
```diff
@@ -43,13 +43,14 @@
...
-step_time_in_seconds = step_solution["Time [s]"].entries
-voltage = solution["Voltage [V]"].entries
-step_voltage = step_solution["Voltage [V]"].entries
-plt.plot(time_in_seconds, voltage, "b-", label="SPMe (continuous solve)")
-plt.plot(step_time_in_seconds, step_voltage, "ro", label="SPMe (stepped solve)")
-plt.xlabel(r"$t$")
-plt.ylabel("Voltage [V]")
-plt.legend()
-plt.show()
+if step_solution is not None:
```
Saransh-cpp, line 46: Should be reverted

examples/scripts/heat_equation.py
```diff
@@ -106,7 +106,7 @@ def T_exact(x, t):
 # Plot ------------------------------------------------------------------------
 x_nodes = mesh["rod"].nodes  # numerical gridpoints
 xx = np.linspace(0, 2, 101)  # fine mesh to plot exact solution
-plot_times = np.linspace(0, 1, 5)
+plot_times: np.ndarray = np.linspace(0, 1, 5)
```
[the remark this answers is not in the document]
Saransh-cpp: Actually, should be reverted

src/pybamm/experiment/step/base_step.py
```diff
@@ -140,7 +140,7 @@ def __init__(
             self.value = pybamm.Interpolant(
                 t,
                 y,
-                pybamm.t - pybamm.InputParameter("start time"),
+                pybamm.t - pybamm.InputParameter("start time"),  # type: ignore[arg-type]
```
Saransh-cpp, line 143: What is the type of this expression? Isn't it `pybamm.Symbol`?

src/pybamm/expression_tree/broadcasts.py
```diff
@@ -78,8 +78,7 @@ def _from_json(cls, snippet):
         )
 
     def _unary_new_copy(self, child: pybamm.Symbol, perform_simplifications=True):
-        """See :meth:`pybamm.UnaryOperator._unary_new_copy()`."""
-        return self.__class__(child, self.broadcast_domain)
+        pass  # pragma: no cover
```
Saransh-cpp, line 81: Why was this removed?

src/pybamm/expression_tree/broadcasts.py
```diff
@@ -191,6 +190,10 @@ def reduce_one_dimension(self):
         """Reduce the broadcast by one dimension."""
         return self.orphans[0]
 
+    def _unary_new_copy(self, child: pybamm.Symbol, perform_simplifications=True):
```
Saransh-cpp, line 193: Why are these functions being repeated?

src/pybamm/expression_tree/concatenations.py
```diff
@@ -526,7 +526,7 @@ def intersect(s1: str, s2: str):
     return intersect.lstrip().rstrip()
 
 
-def simplified_concatenation(*children, name: Optional[str] = None):
+def simplified_concatenation(*children, name=None):
```
Saransh-cpp, line 529:
```suggestion
def simplified_concatenation(*children, name: str | None = None):
```

src/pybamm/models/base_model.py
```diff
@@ -77,6 +77,8 @@ def __init__(self, name="Unnamed model"):
         self.use_jacobian = True
         self.convert_to_format = "casadi"
 
+        self.calculate_sensitivities = []
```
Saransh-cpp, line 80: Where is this variable coming from?

src/pybamm/solvers/base_solver.py
```diff
@@ -1122,7 +1122,7 @@ def _set_sens_initial_conditions_from(
         """
 
         ninputs = len(model.calculate_sensitivities)
-        initial_conditions = tuple([] for _ in range(ninputs))
+        initial_conditions: tuple = tuple([] for _ in range(ninputs))
```
Saransh-cpp, line 1125: Why do we need the explicit `tuple`?

src/pybamm/solvers/idaklu_jax.py
```diff
@@ -259,7 +259,7 @@ def f_isolated(*args, **kwargs):
 
     def jax_value(
         self,
-        t: npt.NDArray = None,
+        t: Union[npt.NDArray, None] = None,
```
Saransh-cpp, line 262:
```suggestion
        t: npt.NDArray | None = None,
```

Saransh-cpp: Could you also fix the types below?

src/pybamm/solvers/idaklu_jax.py
```diff
@@ -292,7 +292,7 @@ def jax_value(
 
     def jax_grad(
         self,
-        t: npt.NDArray = None,
+        t: Union[npt.NDArray, None] = None,
```
Saransh-cpp, line 295: Ditto

src/pybamm/solvers/processed_variable_time_integral.py
```diff
@@ -7,7 +7,7 @@
 @dataclass
 class ProcessedVariableTimeIntegral:
     method: Literal["discrete", "continuous"]
-    initial_condition: npt.NDArray
+    initial_condition: Union[npt.NDArray, float]
     discrete_times: Optional[npt.NDArray]
```
Saransh-cpp, line 11: Ditto

tests/unit/test_expression_tree/test_operations/test_copy.py
```diff
@@ -79,6 +79,7 @@ def test_symbol_create_copy_new_children(self):
                 a * b,
                 a / b,
                 a**b,
+                b % a,
```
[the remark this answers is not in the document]
Rishab87: to ensure coverage, as told in my earlier comment too above:
> I've replaced self.__ class __ with new_instance because earlier when we were using self.__ class __ it showed a third arg was not getting passed:
error: Missing positional argument "right_child" in call to "BinaryOperator"  [call-arg]
but this function was always getting called from instance of its child classes which don't need to pass 3 arguments, so i thought it was better to make a new_instance method which can be overided in child classes

due to this many new_instance functions have been added for modulo and other operations, we can use pragma no cover too instead of adding this in test

src/pybamm/telemetry.py
Rishab87: mypy infers _posthog as `_posthog: MockTelemetry | Posthog` so it throws: 
```python 
src/pybamm/telemetry.py:24: error: "MockTelemetry" has no attribute "log"  [attr-defined]
```

What I found in docs was mypy reads the file all at once. At the top level it sees both assignments to` _posthog` one from `MockTelemetry()` and one from `Posthog(...)` and so that's why it's inferring this.

src/pybamm/solvers/summary_variable.py
Rishab87: mypy still thinks none can be returned so I think it should be something like this:
```python
all_vars = getattr(self, "_all_variables", None)
if all_vars is not None:   
  ```

File: src/pybamm/solvers/summary_variable.py (hunk 9 above)
Comment (Rishab87) [reply src/pybamm/solvers/summary_variable.py:106]: earlier it gave a mypy error I dont exactly remember now, but it does not give any error now so reverting it

File: src/pybamm/expression_tree/broadcasts.py (hunk 13 above)
Comment (Rishab87) [reply src/pybamm/expression_tree/broadcasts.py:81]: In `src/pybamm/expression_tree/broadcasts.py` I've forced child classes of `Broadcast` to implement `_unary_new_copy` because earlier we were using `self.broadcast_domain` in this function in Broadcast class but it does not have any attribute `self.broadcast_domain`, this function was only getting called by instance of their child classes which has self.broadcast_domain property.

File: src/pybamm/expression_tree/broadcasts.py (hunk 14 above)
Comment (Rishab87) [reply src/pybamm/expression_tree/broadcasts.py:193]: As mentioned above now they need to be overrided in child classes

File: src/pybamm/experiment/step/base_step.py (hunk 12 above)
Comment (Rishab87) [reply src/pybamm/experiment/step/base_step.py:143]: yup you're right but it is is not a `Sequence[Symbol]` or a `Time`, so mypy throws error that's why I've added it inside a list

File: src/pybamm/solvers/summary_variable.py (hunk 9 above)
Comment (Rishab87) [reply src/pybamm/solvers/summary_variable.py:106]: Because I was getting this:
```python
src/pybamm/solvers/summary_variable.py:106: error: Item "None" of "ElectrodeSOHSolver | None" has no attribute "_get_electrode_soh_sims_full"  [union-attr]
```

src/pybamm/models/base_model.py
Rishab87: In base solver got this error:
```python
src/pybamm/solvers/base_solver.py:1124: error: "BaseModel" has no attribute "calculate_sensitivities"  [attr-defined]
```

that's why I added this

src/pybamm/solvers/base_solver.py
Rishab87: Because mypy was unable to infer it:
```python
src/pybamm/solvers/base_solver.py:1125: error: Need type annotation for "initial_conditions"  [var-annotated]
```

I think mypy has difficulty inferring that the tuple contains lists of a particular type

Rishab87:
@Saransh-cpp , I've made the changes and replied to the comments, can you have a look?

Issue #4884, opened 2025-02-28
Unify hysteresis models and add heat generation terms


rtimms:
### Description

We now have three options for modelling open-circuit potential hysteresis in PyBaMM: 1) "current sigmoid"; 2) "Wycisk"; and 3) "Axen", all of which have slightly different implementations.

Both "Wycisk" and "Axen" are single-state hysteresis models which contain a single ODE for a hysteresis state variable, `h`, of the form `dh/dt = K(1-sgn(i))h`, but again with slightly different implementations. The current sigmoid model is a zero-state model (the OCP just jumps between branches depending on the sign of the applied current), and doesn't track the variable `h`

We propose we unify these models by (see https://github.com/pybamm-team/PyBaMM/issues/4332#issuecomment-2296943415):

- Making sure all hysteresis models use the state variable `-1<h<1` (in the current sigmoid model `h` will just jump from -1 to 1)

- Add a base one-state model that solves a general one-state equation:

$\frac{\partial h}{\partial t} = \gamma\cdot\left(\frac{i_\mathrm{surf}a_\mathrm{vol}}{\phi_\mathrm{act}Fc_\mathrm{max}}\right)\cdot \left(1 - \mathrm{sgn}(i_\mathrm{surf})h\right)$

for $i_\mathrm{surf}$ the interfacial phase-specific current density (anodic current density positive).

- Allow a scalar input or arbitrary functional input for $\gamma$, **or** allow $\gamma$ to be computed using the specific Wycisk formula with parameters $K$, $n$ and the OCP:

$\gamma = K \cdot \left(\frac{\frac{\partial q_\mathrm{vol}}{\partial U}}{\left.\frac{\partial q_\mathrm{vol}}{\partial U}\right|_\mathrm{ref}}\right)^{-n}$

where the evaluation at a reference point (implicitly at $x_\mathrm{Li}=x_\mathrm{Li,ref}$ which in turn defines $K$ as the decay rate at the reference stoichiometry) is to ensure that we take an arbitrary power only of a unitless quantity.

- Allow general (particle-wise) initialisation of the hysteresis state, not necessarily = 0.

Unifying the models will also allow us to more reasonably include hysteresis heating (see https://github.com/About-Energy-OpenSource/AEPyBaMM/blob/main/src/aepybamm/pybamm_tools.py#L91).

ejfdickinson:
@rtimms Looks good.

[...]

PR #4893, opened 2025-03-05
Issue 4884 unify hysteresis

### LINKED ISSUE #5021: [Bug]:
### PyBaMM Version

25.4.2

### Python Version

3.12.7

### Describe the bug

I'm seeing a weird behaviour when applying current pulses to an SPMe model with anode hysteresis. There is no multiphase, just 2 anode branches I made up:
 
Image

I'm charging the cell at 1C in steps of 10%, then resting it for 10 minutes. 

## Using Wycisk and the algebraic surface form

```python
model = pybamm.lithium_ion.SPMe(
    {
        "working electrode": "both",
        "surface form": "algebraic",
        "contact resistance": "false",
        "open-circuit potential": ("Wycisk", "single")
    }
)
```

Image

First of all, the h state (which is initialised at -1) changes when the current is 0, which should not happen under the Wycisk ODE. Also, there are some bumps in the anode OCP (`sol["Battery negative electrode bulk open-circuit potential [V]"].entries`):

Image

This was using decay rate 0.1 and switching factor 0. Increasing the decay rate affects the way the h-state changes in those regions with 0 current:

Image

## Surface form false

This way the h state no longer changes, but the artifacts in the OCP are still there:

Image

## Using current sigmoid

No more OCP artifacts, here the curves match well during the relaxations and we also get the expected jumps due to the current.

Image

### Steps to Reproduce

```python
import numpy as np
import pybamm

def graphite_ocp(sto):
    return (graphite_ocp_lit(sto) + graphite_ocp_delit(sto)) / 2

def graphite_ocp_delit(sto):
    u_eq = (
        1.9793 * np.exp(-15.3631 * (sto))
        + 0.2482
        - 0.0709 * np.tanh(29.8538 * (sto - 0.234))
        - 0.04478 * np.tanh(14.9159 * (sto - 0.3769))
        - 0.0205 * np.tanh(30.4444 * (sto - 0.7103))
        - 0.0205 * np.tanh(30.4444 * (sto - 0.9103))
    )

    return u_eq

def graphite_ocp_lit(sto):
    u_eq = (
        1.9793 * np.exp(-50.3631 * sto)
        + 0.2482
        - 0.0909 * np.tanh(35.8538 * (sto - 0.0834))
        - 0.04478 * np.tanh(14.9159 * (sto - 0.1769))
        - 0.0105 * np.tanh(30.4444 * (sto - 0.6103))
        - 0.0105 * np.tanh(30.4444 * (sto - 0.4103))
    )

    return u_eq


parameter_values.update({
    "Initial hysteresis state in negative electrode": -1,
    "Negative particle hysteresis decay rate": 1,
    "Negative particle hysteresis switching factor": 0,
    "Negative electrode OCP [V]": graphite_ocp,
    "Negative electrode lithiation OCP [V]": graphite_ocp_lit,
    "Negative electrode delithiation OCP [V]": graphite_ocp_delit,
}, check_already_exists=False)


model = pybamm.lithium_ion.SPMe(
    {
        "working electrode": "both",
        "surface form": "false",
        "contact resistance": "false",
        "open-circuit potential": ("Wycisk", "single")
    }
)
model.insert_reference_electrode()


loop = [
    f"Charge at 1C for 6 minutes",
    "Rest for 10 minutes"
]

exp = pybamm.Experiment(["Rest for 1 second"] + loop * 9, 
period = '1 second')
sim = pybamm.Simulation(model, parameter_values=parameter_values, experiment=exp, solver=pybamm.IDAKLUSolver())
sol = sim.solve(initial_soc=0)


fig, axs = plt.subplots(1, 2, figsize=(18, 6), width_ratios=[1, 2])

axs[0].plot(sol["Time [s]"].entries, sol["X-averaged negative electrode hysteresis state"].entries)

ax = axs[0].twinx()
ax.plot(sol["Time [s]"].entries, sol["Current [A]"].entries)

axs[1].plot(sol["Time [s]"].entries,
            sol["Negative electrode 3E potential [V]"].entries)  
axs[1].plot(sol["Time [s]"].entries,
            sol["Battery negative electrode bulk open-circuit potential [V]"].entries)
```

### Relevant log output

```shell

```

### LINKED ISSUE #5038: [Bug]: Wycisk model differential capacity variable error
### PyBaMM Version

25.4.2

### Python Version

3.12.7

### Describe the bug

One bug and one discussion question really. 

The bug: I'm unable to access the differential capacity variable after simulating a model with the Wycisk option:

```python
sol["Negative electrode secondary differential capacity [A.s.V-1]"]
# or sol["Negative electrode differential capacity [A.s.V-1]"] for single phase anode behaves the same
```
```
The numpy boolean negative, the `-` operator, is not supported, use the `~` operator or the logical_not function instead.
```

I tried to do some debugging by looking at the `WyciskOpenCircuitPotential` class. Printing `dQdU` from where it's defined at:

```python
dU = self.phase_param.U(sto_surf, T_bulk).diff(sto_surf)
dQdU = Q_mag / dU
```
prints something that's probably right at the model initialisation:

```
0.0002777777777777778 * yz-average(x-average(Secondary: Negative electrode active material volume fraction)) * Negative electrode thickness [m] * Electrode width [m] * Electrode height [m] * Secondary: Maximum concentration in negative electrode [mol.m-3] * Faraday constant [C.mol-1] 
/ 
(Secondary: Negative electrode OCP [V] + (yz-average(Ambient temperature [K]) - Reference temperature [K]) * Secondary: Negative electrode OCP entropic change [V.K-1] + 1e-06 * (-(1e-10 <= minimum(boundary value(X-averaged negative secondary particle concentration [mol.m-3]) / Secondary: Maximum concentration in negative electrode [mol.m-3], 0.9999999999)) * (boundary value(X-averaged negative secondary particle concentration [mol.m-3]) / Secondary: Maximum concentration in negative electrode [mol.m-3] <= 0.9999999999) / ((maximum(minimum(boundary value(X-averaged negative secondary particle concentration [mol.m-3]) / Secondary: Maximum concentration in negative electrode [mol.m-3], 0.9999999999), 1e-10)) ** 2.0) + -(1e-10 <= minimum(boundary value(X-averaged negative secondary particle concentration [mol.m-3]) / Secondary: Maximum concentration in negative electrode [mol.m-3], 0.9999999999)) * (boundary value(X-averaged negative secondary particle concentration [mol.m-3]) / Secondary: Maximum concentration in negative electrode [mol.m-3] <= 0.9999999999) / ((-1.0 + maximum(minimum(boundary value(X-averaged negative secondary particle concentration [mol.m-3]) / Secondary: Maximum concentration in negative electrode [mol.m-3], 0.9999999999), 1e-10)) ** 2.0)))
```

Or at least the numerator is for sure correct, and I'll trust the denominator too 😅.

Later, in the `set_rhs()`, I see we're only extracting a component from this variable:
```python
dQdU = dQdU.orphans[0]
```
and using that inside the ODE parameter. Printing this new `dQdU` here just gives the numerator from above, i.e. the phase capacity:
`
0.0002777777777777778 * yz-average(x-average(Secondary: Negative electrode active material volume fraction)) * Negative electrode thickness [m] * Electrode width [m] * Electrode height [m] * Secondary: Maximum concentration in negative electrode [mol.m-3] * Faraday constant [C.mol-1]
`

So the question: are we not removing the stoichiometry dependency of the $k(z)$ rate by using this? What is the purpose of the `orphans[0]` line in `set_rhs()`?

I tried commenting it, hoping to put back the full dQdU in the ODE. If I do that, I get exactly the same error message as above, but this time it's when running `sim.solve()`. 

### Steps to Reproduce

```python
import pybamm

model = pybamm.lithium_ion.SPMe({
    "working electrode": "both",
    "open-circuit potential": (("single", "Wycisk"), "single"),
    "particle phases": ("2", "1"),
})

parameters_DCHS = pybamm.ParameterValues("Chen2020_composite")
lithiation_ocp = parameters_DCHS["Secondary: Negative electrode lithiation OCP [V]"]
delithiation_ocp = parameters_DCHS["Secondary: Negative electrode delithiation OCP [V]"]

def ocp_avg(sto):
    return (lithiation_ocp(sto) + delithiation_ocp(sto)) / 2

parameters_DCHS.update({
    "Secondary: Initial hysteresis state in negative electrode": 1,
    "Secondary: Negative particle hysteresis decay rate": 1,
    "Secondary: Negative particle hysteresis switching factor": 2,
    "Secondary: Negative electrode OCP [V]": ocp_avg,
}, check_already_exists=False)


    "Rest for 1 second",
    "Discharge at 1C for 3 minutes"
])
sim = pybamm.Simulation(model, parameter_values=parameters_DCHS, experiment=experiment, solver=pybamm.IDAKLUSolver())
sol = sim.solve(calc_esoh=False)

sol["Negative electrode secondary differential capacity [A.s.V-1]"].entries
```

### Relevant log output

```shell

```

### LINKED ISSUE #5081: [Bug]: Simplifications cause heavisides to evaluate as booleans
### PyBaMM Version

develop

### Python Version

3.11

### Describe the bug

Simplifications cause Heaviside functions to evaluate as booleans when not multiplied by something else. This makes it not possible to have a zoh as described in #3446 with a value of 1

### Steps to Reproduce

```
import pybamm
import numpy as np
pybamm.set_logging_level("DEBUG")

# Works
param.update({"Current function [A]": pybamm.Scalar(1.000001) * (pybamm.t<10.0) + pybamm.Scalar(-1.000001) * (pybamm.t>=10.0)})
sim = pybamm.Simulation(model, solver=solver, parameter_values=param)
sol = sim.solve(np.array([0.0, 10.0, 20.0]))

# Errors
param.update({"Current function [A]": pybamm.Scalar(1) * (pybamm.t<10.0) + pybamm.Scalar(-1) * (pybamm.t>=10.0)})
sim = pybamm.Simulation(model, solver=solver, parameter_values=param)
sol = sim.solve(np.array([0.0, 10.0, 20.0]))
```

```


### Relevant log output

```shell
2025-06-25 14:13:28.785 - [DEBUG] discretisation.process_dict(738): Discretise 'Electrode current density [A.m-2]'
  File "/Users/abills/pybamm/pybamm/src/pybamm/discretisations/discretisation.py", line 769, in process_symbol
    return self._discretised_symbols[symbol]
           ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
KeyError: Concatenation(-0x2b7cf2d7ff82eb9f, concatenation, children=['-58.56620185738528 * grad(Negative electrode potential [V])', 'broadcast(0.0)', '-5.856620185738528 * grad(Positive electrode potential [V])'], domains={'primary': ['negative electrode', 'separator', 'positive electrode'], 'secondary': ['current collector']})

During handling of the above exception, another exception occurred:

  File "/Users/abills/pybamm/pybamm/test.py", line 19, in <module>
    sol = sim.solve(np.array([0.0, 10.0, 20.0]))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/abills/pybamm/pybamm/src/pybamm/simulation.py", line 474, in solve
    self.build(initial_soc=initial_soc, inputs=inputs)
  File "/Users/abills/pybamm/pybamm/src/pybamm/simulation.py", line 334, in build
    self._built_model = self._disc.process_model(
                        ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/abills/pybamm/pybamm/src/pybamm/discretisations/discretisation.py", line 213, in process_model
    model_disc.variables = self.process_dict(pre_processed_variables)
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/abills/pybamm/pybamm/src/pybamm/discretisations/discretisation.py", line 739, in process_dict
    processed_eqn = self.process_symbol(eqn)
                    ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/abills/pybamm/pybamm/src/pybamm/discretisations/discretisation.py", line 771, in process_symbol
    discretised_symbol = self._process_symbol(symbol)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/abills/pybamm/pybamm/src/pybamm/discretisations/discretisation.py", line 983, in _process_symbol
    new_symbol = spatial_method.concatenation(new_children)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/abills/pybamm/pybamm/src/pybamm/spatial_methods/finite_volume.py", line 1234, in concatenation
    child_size = child.size
                 ^^^^^^^^^^
  File "/Users/abills/miniconda3/envs/pybamm/lib/python3.11/functools.py", line 1001, in __get__
    val = self.func(instance)
          ^^^^^^^^^^^^^^^^^^^
  File "/Users/abills/pybamm/pybamm/src/pybamm/expression_tree/symbol.py", line 1019, in size
    return np.prod(self.shape)
                   ^^^^^^^^^^
  File "/Users/abills/miniconda3/envs/pybamm/lib/python3.11/functools.py", line 1001, in __get__
    val = self.func(instance)
          ^^^^^^^^^^^^^^^^^^^
  File "/Users/abills/pybamm/pybamm/src/pybamm/expression_tree/symbol.py", line 1030, in shape
    evaluated_self = self.evaluate(0, y, y, inputs="shape test")
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/abills/pybamm/pybamm/src/pybamm/expression_tree/binary_operators.py", line 162, in evaluate
    right = self.right.evaluate(t, y, y_dot, inputs)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/abills/pybamm/pybamm/src/pybamm/expression_tree/binary_operators.py", line 162, in evaluate
    right = self.right.evaluate(t, y, y_dot, inputs)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/abills/pybamm/pybamm/src/pybamm/expression_tree/binary_operators.py", line 162, in evaluate
    right = self.right.evaluate(t, y, y_dot, inputs)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/abills/pybamm/pybamm/src/pybamm/expression_tree/binary_operators.py", line 163, in evaluate
    return self._binary_evaluate(left, right)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/abills/pybamm/pybamm/src/pybamm/expression_tree/binary_operators.py", line 314, in _binary_evaluate
    return left - right
           ~~~~~^~~~~~~
TypeError: numpy boolean subtract, the `-` operator, is not supported, use the bitwise_xor, the `^` operator, or the logical_xor function instead.
```

rtimms:
# Description
Fixes #5038
Fixes #5081
Fixes #5021

Unifies the hysteresis models in PyBaMM and adds heat generation due to hysteresis. In particular:

- Adds a new `BaseHysteresisOpenCircuitPotential` class that sets variables for the lithiation and delithiation OCP, as well as the average (simply the mean), and the hysteresis voltage (`H = U_lith - U_delith`).
- Makes sure all models, even `CurrentSigmoidOpenCircuitPotential`, are expressed in terms of a hysteresis state variable `-1<h<1`
- Allows the initial hysteresis state to be a function of position through the electrode 
- Allows the hysteresis decay rates of the Axen and Wycisk models to be a function of stoichiometry and temperature 
- Adds a heat source term in each active material phase `Q_hys = i_vol * (U - U_eq)` where `i_vol` is the volumetric interfacial current density, `U` is the OCP (i.e. includes hysteresis), and `U_eq` is the "equilibrium OCP" 
- Renames the  notebook `differential-capacity-hysteresis-state.ipynb` to`hysteresis-state-models.ipynb` and extends it to include a comparison of all of the existing hysteresis models

Related #4884 

What this doesn't do from #4884:

- Reformulate the Wycisk model to use $\gamma = K \cdot \left(\frac{\frac{\partial q_\mathrm{vol}}{\partial U}}{\left.\frac{\partial q_\mathrm{vol}}{\partial U}\right|_\mathrm{ref}}\right)^{-n}$. This would be breaking as it requires rescaling parameters. Is this a breaking change worth making? Changes to parameters have, in the past, been problematic/easy for users to miss.
- Express all the models simply as $\frac{\partial h}{\partial t} = \gamma\cdot\left(\frac{i_\mathrm{surf}a_\mathrm{vol}}{\phi_\mathrm{act}Fc_\mathrm{max}}\right)\cdot \left(1 - \mathrm{sgn}(i_\mathrm{surf})h\right)$ using the base class, and then specify `\gamma` in the subclasses. This would be an easy change to make, but I'm not sure it is much harder to just specify the whole `rhs` in the base class and it makes the code more readable (even if there is some repetition). The way it currently is, you see the whole ODE in the `set_rhs` method and don't need to chase back to the base class. Happy to take input on this.

## Breaking
- Need to explicitly give the equilibrium, delithiation, and lithiation OCPs when using a hysteresis model. E.g. you must provide all three of "Negative electrode OCP [V]", "Negative electrode delithiation OCP [V]", and "Negative electrode lithiation OCP [V]"

## Type of change

# Important checks:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,7 @@ ci:
 
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: "v0.12.1"
+    rev: "v0.12.2"
     hooks:
       - id: ruff
         args: [--fix, --show-fixes]
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -374,7 +374,10 @@ PyBaMM is documented in several ways.
 
 First and foremost, every method and every class should have a [docstring](https://www.python.org/dev/peps/pep-0257/) that describes in plain terms what it does, and what the expected input and output is.
 
-These docstrings can be fairly simple, but can also make use of [reStructuredText](http://docutils.sourceforge.net/docs/user/rst/quickref.html), a markup language designed specifically for writing [technical documentation](https://en.wikipedia.org/wiki/ReStructuredText). For example, you can link to other classes and methods by writing ``:class:`pybamm.Model` `` and ``:meth:`run()` ``.
+These docstrings can be fairly simple, but can also make use of
+`reStructuredText`, a markup language designed specifically for writing
+technical documentation. For example, you can link to other classes and methods
+by writing ``:class:`pybamm.Model` `` and ``:meth:`run()` ``.
 
 In addition, we write a (very) small bit of documentation in separate reStructuredText files in the `docs` directory. Most of what these files do is simply import docstrings from the source code. But they also do things like add tables and indexes. If you've added a new class to a module, search the `docs` directory for that module's `.rst` file and add your class (in alphabetical order) to its index. If you've added a whole new module, copy-paste another module's file and add a link to your new file in the appropriate `index.rst` file.
 


rtimms:
@ejfdickinson would appreciate your feedback on this

ejfdickinson:
@rtimms I'll look over it.

Relating to #4884 incomplete items currently excluding :

- If the breaking change to introduce a reference value is unappetising (fair enough), I recommend that we consider this a "specification bug fix" where we redefine from

  $\left(\frac{\partial q_\mathrm{vol}}{\partial U}\right)^{-n}$

  to

  $\left(\frac{\frac{\partial q_\mathrm{vol}}{\partial U}}{C_\mathrm{ref,vol}}\right)^{-n}$

  where $C_\mathrm{ref,vol}$ is (implicitly) hardcoded = 1 F/m3. Then the equation has legitimate unit syntax but the implementation in code doesn't need altering.
  
- I'd be fine to avoid inheriting an equation using $\gamma$ from the base class.

- In my view we should prefer a solution with future-proofed flexibility for the true equilibrium OCP to be unequal to the mean of the two branches. Not doing this now is just storing another breaking change for the future, isn't it? From my opinion and in terms of my use cases, I'd be ok to suck up a breaking change that impacts the lumped heat source. Of course, it'd be good to check this against a general policy.

ejfdickinson (commented):
In addition to specific comments on the docs, one additional comment:

- Should we use this as an opportunity to move from arbitrary author-name definitions to more descriptive inputs options, e.g.

`"Axen"` -> `"one-state"`
`"Wycisk"` -> `"one-state differential capacity"`

and similarly for module names? Although they may be the original reports, I don't think that "Axen" or "Wycisk" are really that intuitively descriptive for what are actually quite general hysteresis descriptions.

src/pybamm/models/submodels/interface/open_circuit_potential/wycisk_ocp.py
```diff
@@ -161,17 +93,12 @@ def set_rhs(self, variables):
...
+        x = self.phase_param.hysteresis_switch
 
-        dhdt = (
-            K * (i_surf / (Q_cell * (dQdU**K_x))) * (1 - pybamm.sign(i_surf) * h)
-        )  #! current is backwards for a halfcell
-        self.rhs[h] = dhdt
+        i_surf_sign = pybamm.sign(i_surf)
+        signed_h = 1 - i_surf_sign * h
+        gamma = Gamma * (1 / dQdU**x)
+        dhdt = gamma * i_surf / Q_cell * signed_h
```
Daniel-Nicolae23: How about using this in the ODE:

$$\left( \frac{1}{U_{100} - U_{0}} \frac{\partial U}{\partial x} \right)^n$$

This way it's unitless and without any newly introduced reference variable. And it removes the need to use the `Q_mag` from lines 50-60, which is an extensive parameter, so it makes the decay behave differently for differently-sized models having the same `Gamma`.
It would probably be correct to call this model "differential voltage" rather than "differential capacity", but it would do the same job.

rtimms: my main reason for not wanting to change the equation is that it requires modifying previously provided parameters to retain the same behaviour, and as a result may lead to an unexpected change in behaviour. my takeaway from this is that we should recommend the "Axen" model and leave the awkward Wycisk units

Daniel-Nicolae23: Fair enough, I think they've been equivalent this whole time anyway due to #5038.

docs/source/examples/notebooks/models/hysteresis-state-models.ipynb
```diff
@@ -0,0 +1,351 @@
...
+    "\n",
+    "Warning: in this implementation, the units of both $\\gamma$ and $\\Gamma$ are unusual. The parameter $\\gamma$ has units $\\text{m}^2$ and $\\Gamma$ has units $\\text{m}^2\\text{V}^{x}\\text{A}^{-x}\\text{h}^{-x}$. Also note that the units of $Q$ are $\\text{A}\\text{h}$, but the time $t$ is expressed in seconds.\n",
+    "\n",
+    "### Current Sigmoid (Ai)\n",
+    "This approach uses a sigmoid function to switch between the delithiation and lithiation branches of the open-circuit potential, depending on the sign of the current. The original paper gives the following expression for the open-circuit potential:\n",
+    "\n",
+    "$$ U = \\text{sigmoid}\\left(-\\frac{KI}{Q_{\\text{cell}}}\\right) U_{\\text{delith}} + \\text{sigmoid}\\left(\\frac{KI}{Q_{\\text{cell}}}\\right) U_{\\text{lith}}\n",
+    "$$\n",
+    "\n",
+    "Where $K$ is a fitting parameter (in the current implementation this is hard-coded to 100, as in the original publication), $I$ is the cell current, and $Q_{\\text{cell}}$ is the cell capacity. To simplify the comparison with the other models, we can rewrite this expression in terms of the hysteresis state variable $h$ given by\n",
```
aabills, line 75: didn't look at this all that closely but I think it's important here to make the things which are functions of time functions of time in the documentation

src/pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -349,6 +349,23 @@ def __init__(self, extra_options):
         }
         extra_options = extra_options or {}
 
+        # Catch renamed hysteresis options
+        ocp_option = extra_options.get("open-circuit potential")
+        if (isinstance(ocp_option, str) and ocp_option == "Wycisk") or (
+            isinstance(ocp_option, tuple) and "Wycisk" in ocp_option
+        ):
+            raise pybamm.OptionError(
```
aabills: is there a good reason to throw an error rather than a deprecation warning here? This could be less breaking

src/pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -349,6 +349,23 @@ def __init__(self, extra_options):
...
+                "The 'Wycisk' open-circuit potential model has been renamed. "
+                "Use 'one-state differential capacity hysteresis' instead."
+            )
+        if (isinstance(ocp_option, str) and ocp_option == "Axen") or (
+            isinstance(ocp_option, tuple) and "Axen" in ocp_option
+        ):
+            raise pybamm.OptionError(
```
aabills: here too

src/pybamm/models/submodels/interface/open_circuit_potential/single_ocp.py
```diff
@@ -12,39 +12,24 @@ def get_coupled_variables(self, variables):
...
             dUdT = self.phase_param.dUdT(sto_surf)
 
-            # Bulk OCP is from the average SOC and temperature
             sto_bulk = variables[f"{Domain} electrode {phase_name}stoichiometry"]
             T_bulk = pybamm.xyzs_average(T)
             ocp_bulk = self.phase_param.U(sto_bulk, T_bulk)
+
+            variables.update(
+                {
+                    f"{Domain} electrode {phase_name}equilibrium open-circuit potential [V]": ocp_surf,
```
aabills, line 26: I think it's going to be pretty confusing for users to have this variable...

src/pybamm/models/submodels/interface/open_circuit_potential/one_state_differential_capacity_hysteresis_ocp.py
```diff
@@ -0,0 +1,129 @@
...
+
+        i_surf_sign = pybamm.sign(i_surf)
+        signed_h = (1 - i_surf_sign * h) / 2
+        gamma = Gamma * (1 / dQdU**x)
+        dhdt = gamma * i_surf / Q_cell * signed_h
+
+        self.rhs[h] = dhdt
+
+
+class WyciskOpenCircuitPotential(
```
aabills: these can't really be accessed without a lot of work right? What's the point in keeping it?

src/pybamm/models/submodels/thermal/base_thermal.py
```diff
@@ -146,11 +150,27 @@ def _get_standard_coupled_variables(self, variables):
             # Reversible electrochemical heating
             dUdT_p = variables[f"Positive electrode {phase}entropic change [V.K-1]"]
             Q_rev_p += a_j_p * T_p * dUdT_p
+            # Hysteresis electrochemical heating
+            if ocp_option == "single":
```
aabills, line 154: I think this should cite something, it's not necessarily obvious to most engineers that it should be included.

src/pybamm/models/submodels/interface/open_circuit_potential/one_state_differential_capacity_hysteresis_ocp.py
aabills: What if this implemented the old model, since the new model is slightly different? And then remove the option errors. That would make this a non-breaking change.

docs/source/examples/notebooks/models/hysteresis-state-models.ipynb
valentinsulzer, line 1: unfinished sentence "We add some"

valentinsulzer: Could also do with a sentence at the end describing the results

valentinsulzer (commented):
Looks good, a couple of minor comments for the notebook.
I also prefer the previous names (Axen/Wycisk) over the new names which are a real mouthful, but not a strong opinion

kratman:
I am looking into the docs failure now, probably something pretty minor

kratman:
> I am looking into the docs failure now, probably something pretty minor

I pushed a temporary fix for the failing citation so that review could continue, I will push a  in a real fix once I get it working again

kratman:
Docs are fixed now. I also updated the description to make sure all tickets get closed

src/pybamm/models/submodels/interface/open_circuit_potential/single_ocp.py
rtimms: yes, but you need it to compute the hysteresis source term

src/pybamm/models/submodels/interface/open_circuit_potential/one_state_differential_capacity_hysteresis_ocp.py
rtimms: `Axen` and `OneStateHysteresisOpenCircuitPotential` are the same (except a factor of 2, but happy to revert that as it is annoyingly breaking).

`Wycisk` and `OneStateDifferentialCapacityHysteresisOpenCircuitPotential`, except this PR fixes #5038, so currently `Wycisk` is wrong on main.

Agree that the classes should go as they are basically inaccessible.

rtimms:
> I also prefer the previous names (Axen/Wycisk) over the new names which are a real mouthful, but not a strong opinion

The argument for them is that they at least infer what the model is from the option. I don't think the option names are so much of a mouthful ("one-state hysteresis"), but I agree the classes could be shortened.

aabills (approved):
Thanks!

PR #4915, opened 2025-03-13
feat: change default solver to idaklu

### LINKED ISSUE #4914: change default solver to idaklu
### Description

We're planning on changing the default solver to idaklu (from casadi), but I couldn't find an issue for this, so creating one now

#3910

### Motivation

most development is focused on idaklu solver, which is usually faster than casadi

### Possible Implementation

_No response_

### Additional context

_No response_

martinjrobins:
# Description

Fixes #4914

*Note:* as part of this change I've had to relax a lot of the test tolerances, due to the different default tolerances that we have for the IDAKLU solver. If users have existing code that uses the default solver and tolerances, this might fail suddenly when they upgrade. We have to think about how to handle this

## Type of change

# Key checklist:

## Further checks:


pipliggins (approved):
Looks good to me aside from the missing coverage + a changelog entry :)

Would adding a temporary warning on the base solver if the tolerances aren't the default values work for letting users know about the change?

martinjrobins:
> Looks good to me aside from the missing coverage + a changelog entry :)
> 
> Would adding a temporary warning on the base solver if the tolerances aren't the default values work for letting users know about the change?

good idea, I've added a warning if users use the default solver. We can remove it after a couple of releases

MarcBerliner:
@martinjrobins one minor request: can we also update the readme to use the `IDAKLUSolver` instead of casadi?

martinjrobins:
> @martinjrobins one minor request: can we also update the readme to use the `IDAKLUSolver` instead of casadi?

good point, added this now

brosaplanella:
I am a bit confused, did this make it in the release or not?

agriyakhetarpal:
No, it didn't make it to the release, @brosaplanella

brosaplanella:
Then we need to update the CHANGELOG, as it has included under the 25.4 section.

Issue #4925, opened 2025-03-20
Review ruff rules


Saransh-cpp:
I think we never looked at the ruff rules after it was added to the codebase. Some of these are very useful and should be selected.

https://github.com/pybamm-team/PyBaMM/blob/10a404eb2f90d263b3bc78e1704f9e881b1d9923/pyproject.toml#L172-L212

For someone looking at this issue: feel free to pick up a rule and discuss why it should be included/excluded in this thread before making a PR.

Aswinr24:
I do feel I (isort), ICN can be included to standardize imports, PT since we use pytest for our testing suite and recently had the addition of fixtures as well (PT020, PT001, PT003, PT001) and C4 to get rid of unnecessary generators and list comprehensions maybe?

Saransh-cpp:
Yes! PT and I definitely ✅ 

Should be added in separate PRs tho.

Aswinr24:
I ran ruff check --fix . for the isort rules, and it changed nearly 400+ files. Should I include all in a single pr ?
and for the Pytest Rule it suggests decent amount of changes to be done, I guess seperate PRs for each should work?

Saransh-cpp:
I'd say let's go ahead with the pytest rule and discuss isort more in this thread.

I hope the pytest rule is not asking us to remove classes?

Aswinr24:
The errors from pytest rule are mostly around ` pytest.raises()` being too broad without a match parameter (PT011), the `pytest.raises` block containing multiple statements under its context (PT012) - should be limited to a single statement
and `pytest.mark.parameterize` must have a tuple for the initial parameters if it has a sequence of strings (PT006)

agriyakhetarpal:
Thanks, @Aswinr24. All three of those would be pretty nice to fix!

Aswinr24:
Yes will work on fixing them!

mxxn-ai:
Following up on @Aswinr24's suggestion earlier in this thread to add **ICN** for standardizing imports, I would like to take this one on, since `I` (isort) and `PT` are already being handled.

Running ruff 0.15.16 across the whole repository, ICN reports 6 violations, all of type `ICN001`. Here is exactly what I plan to do with each one.

**Four imports I will change to the standard alias** (three in one file, one in another):

[...]

PR #4928, opened 2025-03-23
chore: enable pytest ruff rule and apply fixes

Aswinr24:
# Description

This PR enables additional Ruff linting rules and applies fixes to the codebase as suggested by Ruff. The changes include:

Enabling PT (pytest-style) rules in pyproject.toml.

fixes issues raised by ruff linter which included:

  - PT006: Update @pytest.mark.parametrize to use a tuple for multiple parameters instead of a list or comma-separated strings.
  - PT011: Ensure `pytest.raises()` is called with a `match` parameter to avoid catching overly broad exceptions.
  - PT012: Ensure `pytest.raises()` context managers contains only a single statement.

Related to #4925 

## Type of change

# Key checklist:

## Further checks:


tests/unit/test_solvers/test_idaklu_solver.py
```diff
@@ -689,7 +689,7 @@ def test_failures(self):
         solver = pybamm.IDAKLUSolver()
 
         t_eval = [0, 3]
-        with pytest.raises(ValueError):
+        with pytest.raises(ValueError, match="std::exception"):
```
Saransh-cpp, line 692: Ditto

Saransh-cpp (changes requested):
Thanks, @Aswinr24! Where did you get the `match="string"` strings from, as most of them are wrong? Is it from an older version of PyBaMM?

Please see my comments below to fix the tests.

tests/unit/test_solvers/test_idaklu_solver.py
```diff
@@ -828,7 +828,7 @@ def test_setup_options(self):
                             soln.y, soln_base.y, rtol=1e-5, atol=1e-4
                         )
                     else:
-                        with pytest.raises(ValueError):
+                        with pytest.raises(ValueError):  # noqa: PT011
```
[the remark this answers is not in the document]
Aswinr24: The error messages were inconsistent (e.g., sometimes "Unknown linear solver..." otherwise "Unknown jacobian type..."). So I had to ignore PT011 here or can I use match="Unknown" as the common pattern? It seems to be the only consistent part across the errors.

tests/unit/test_solvers/test_idaklu_solver.py
```diff
@@ -894,7 +894,7 @@ def test_solver_options(self):
             options = {option: options_fail[option]}
             solver = pybamm.IDAKLUSolver(options=options)
 
-            with pytest.raises(ValueError):
+            with pytest.raises(ValueError, match="std::exception"):
```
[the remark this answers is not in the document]
Aswinr24: The test is matching only "std::exception" maybe since that seems to be the generic error message propagated from the C++ solver, what do you suggest we do here?

tests/unit/test_solvers/test_idaklu_solver.py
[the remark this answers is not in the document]
Saransh-cpp: Sounds good!

tests/unit/test_solvers/test_idaklu_solver.py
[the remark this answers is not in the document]
Saransh-cpp: Thanks!

tests/unit/test_solvers/test_jax_solver.py
```diff
@@ -218,7 +218,7 @@ def test_get_solve(self):
         disc.process_model(model)
 
         # test that another method string gives error
-        with pytest.raises(ValueError):
+        with pytest.raises(ValueError, match="Invalid solver method"):
```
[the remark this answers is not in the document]
Aswinr24: Sorry for the oversight, I hadn't installed PyBaMM with JAX solver hence these tests were being skipped and I missed updating the match strings. I will correct them now with the proper error messages.

Saransh-cpp (changes requested):
Thanks, @Aswinr24! This looks good now. Could you please resolve the conflicts?

Aswinr24:
Yes thanks!, would resolve the conflicts and push it

Aswinr24:
Was thinking to open the PR to enable isort rule along with its suggested fixes as well  
Though it does suggest changes in a lot of files, most of them are at the top of the files which shouldn't cause any  issues, Do you suggest to go ahead with this @Saransh-cpp ?

tests/unit/test_solvers/test_idaklu_solver.py
Saransh-cpp, line 692: The error raised is different on Windows and UNIX. Can we not have `match=` here (ignore the rule) and add a comment for future reference -
```py
# raises `std::exception` on UNIX and `IDA failed with flag -22` on Windows
with pytest.raises(ValueError)  # noqa: ...
    ...
```

tests/unit/test_solvers/test_idaklu_solver.py
Saransh-cpp, line 897: Ditto!

Saransh-cpp (changes requested):
Thanks, @Aswinr24! Please see my comments below -

tests/unit/test_solvers/test_idaklu_solver.py
Aswinr24: Sure, Will add this comment and ignore the rule here

Saransh-cpp:
pre-commit.ci run

Saransh-cpp (approved):
Looks good now, thanks, @Aswinr24!

Aswinr24:
Thanks!

tests/unit/test_expression_tree/test_operations/test_convert_to_casadi.py
```diff
@@ -205,7 +205,7 @@ def test_interpolation(self):
         data4 = 2 * x4  # np.tile(2 * x3, (10, 1)).T
         with pytest.raises(ValueError, match="Invalid dimension of x"):
             interp = pybamm.Interpolant(x4_, data4, y4, interpolator="linear")
-            interp_casadi = interp.to_casadi(y=casadi_y)
+        interp_casadi = interp.to_casadi(y=casadi_y)
```
kratman, line 209: This part looks off to me. If the raise value error happens in the `pybamm.Interpolant()` line then the `to_casadi()` line does nothing

tests/unit/test_models/test_full_battery_models/test_equivalent_circuit/test_thevenin.py
```diff
@@ -46,8 +46,8 @@ def test_changing_number_of_rcs(self):
         model = pybamm.equivalent_circuit.Thevenin(options=options)
         model.check_well_posedness()
 
-        with pytest.raises(pybamm.OptionError, match="natural numbers"):
-            options = {"number of rc elements": -1}
+        options = {"number of rc elements": -1}
+        with pytest.raises(pybamm.OptionError, match="natural numbers"):  # noqa: PT012
```
kratman, line 50: What is being suppressed here?

kratman (changes requested):

PR #4944, opened 2025-03-31
Ensure stress-induced diffusion is only enabled on electrodes that have particle mechanics

### LINKED ISSUE #4943: [Bug]: If particle mechanics is enabled on one electrode, stress-induced diffusion is enabled on both
### PyBaMM Version

develop branch

### Python Version

3.10.12

### Describe the bug

If the `"particle mechanics" option` is enabled on either electrode, the `"stress-induced diffusion"` option is enabled on both, unless otherwise specified. See the example below, where the voltage is much lower if stress-induced diffusion is explicitly disabled in the positive electrode.

I can add some code to `base_battery_model` to fix this, but I'm aware that code has to handle a lot of complex cases already so not sure how to proceed...

### Steps to Reproduce

import pybamm
model1 = pybamm.lithium_ion.DFN({"particle mechanics": ("swelling only", "none")})
exp = pybamm.Experiment(["Discharge at 1C until 2.5 V"])
sim1 = pybamm.Simulation(model1, parameter_values=param, experiment=exp)
sol1 = sim1.solve()
model2 = pybamm.lithium_ion.DFN({"particle mechanics": ("none", "swelling only")})
sim2 = pybamm.Simulation(model2, parameter_values=param, experiment=exp)
sol2 = sim2.solve()
model3 = pybamm.lithium_ion.DFN(
    {
        "stress-induced diffusion": ("true", "false"),
    }
)
sim3 = pybamm.Simulation(model3, parameter_values=param, experiment=exp)
sol3 = sim3.solve()
model4 = pybamm.lithium_ion.DFN(
    {
        "particle mechanics": ("none", "swelling only"),
        "stress-induced diffusion": ("false", "true"),
    }
)
sim4 = pybamm.Simulation(model4, parameter_values=param, experiment=exp)
sol4 = sim4.solve()
Q1 = sol1["Discharge capacity [A.h]"].data
V1 = sol1["Voltage [V]"].data
Q2 = sol2["Discharge capacity [A.h]"].data
V2 = sol2["Voltage [V]"].data
Q3 = sol3["Discharge capacity [A.h]"].data
V3 = sol3["Voltage [V]"].data
Q4 = sol4["Discharge capacity [A.h]"].data
V4 = sol4["Voltage [V]"].data
plt.figure()
plt.plot(Q1,V1,label="model1")
plt.plot(Q2,V2,label="model2",linestyle="dashed")
plt.plot(Q3,V3,label="model3",linestyle="dotted")
plt.plot(Q4,V4,label="model4",linestyle="dashdot")
plt.legend()
plt.show()

### Relevant log output

```shell

```

DrSOKane:
# Description

If particle mechanics is enabled on one electrode, stress-induced diffusion is enabled on both by default. I added some code to base_battery_model to catch most instances where this might happen.

Fixes #4943

### DIFF
--- a/src/pybamm/models/full_battery_models/base_battery_model.py
+++ b/src/pybamm/models/full_battery_models/base_battery_model.py
@@ -425,6 +425,21 @@ class BatteryModelOptions(pybamm.FuzzyDict):
             and default_options["particle mechanics"] == "none"
         ):
             default_options["stress-induced diffusion"] = "false"
+        elif (
+            mechanics_option == ("none", "swelling only")
+            and default_options["particle mechanics"] == "none"
+        ):
+            default_options["stress-induced diffusion"] = ("false", "true")
+        elif (
+            mechanics_option == ("swelling only", "none")
+            and default_options["particle mechanics"] == "none"
+        ):
+            default_options["stress-induced diffusion"] = ("true", "false")
+        elif (
+            mechanics_option == ("swelling and cracking", "none")
+            and default_options["particle mechanics"] == "none"
+        ):
+            default_options["stress-induced diffusion"] = ("true", "false")
         else:
             default_options["stress-induced diffusion"] = "true"
         # The "stress-induced diffusion" option will still be overridden by


rtimms:
My opinion is that we should remove all of these default options that change based on other options and instead users should explicitly choose all the options they want. If any options require other options to be enabled then we should catch that and raise an error. This would be breaking, but I think options silently changing has a lot of potential to introduce error.

aabills:
> My opinion is that we should remove all of these default options that change based on other options and instead users should explicitly choose all the options they want. If any options require other options to be enabled then we should catch that and raise an error. This would be breaking, but I think options silently changing has a lot of potential to introduce error.

I agree, it's always bugged me that stress automatically turns on stress-enhanced diffusion. I think most parameterizations are done assuming that this isn't true, so I don't think it should be the default.

PR #4950, opened 2025-04-02
Release v25.4.0

kratman:
Release branch from develop->main.

Note: This PR must be merged not squashed


MarcBerliner (approved):

Saransh-cpp (commented):
FYI: This will put the conda-forge distribution (plus ARM64 and Docker+ARM64 outage) 3 releases behind. Users on these platforms are still running older versions of PyBaMM. I really think the infrastructure issues of `pybammsolvers` should be prioritised.

An easy fix right now is to make it an optional dependency till the infrastructure is ready.

PR #4964, opened 2025-04-09
Update docstring in Experiment

### LINKED ISSUE #4963: [Bug]: Issue with Experiment temperature
### PyBaMM Version

25.4

### Python Version

3.11

### Describe the bug

The Experiment class says it takes temperature in Celsius:
```
    temperature: float, optional
        The ambient air temperature in degrees Celsius at which to run the experiment.
        Default is None whereby the ambient temperature is taken from the parameter set.
        This value is overwritten if the temperature is specified in a step.
```

However, the steps take it in Kelvin (or string)
```
    temperature : float or string, optional
        The temperature of the step. If a float, the value is in Kelvin. If a string,
        the value should be a valid temperature string, e.g. "25 oC".
```

Given that the Experiment class uses the method `_convert_temperature_to_kelvin(temperature)` from the step, they should be consistent. Moreover, that method explicitly says if a double/int assume it is in Kelvin.

TLDR: it is an issue with the docs only (I think).

### Steps to Reproduce

It all arose when @AlexandruPascu was running an Experiment with `temperature=5` and the `Okane2022` parameter set and it broke down (thanks for flagging this up!).

A better way to check is
```
experiment = pybamm.Experiment(["Discharge at 1C until 2.5 V"], temperature=5)
print(experiment.temperature)
print(experiment.steps)
```

They both return 5, even though it should be 278.5 according to the Experiment docs.

### Relevant log output

```shell

```

brosaplanella:
# Description

The docstring in Experiment was wrong. I think there is no further bug (just updating the docstring) but please double check. Not sure it needs a CHANGELOG entry as technically it is just a documentation update.

Related to #4963

## Type of change

# Important checks:


agriyakhetarpal (changes requested):
Hi @brosaplanella, I think this change is incorrect? The type annotation in `temperature: float | None = None` doesn't accept strings, and we pass this attribute down to `pybamm.experiment._convert_temperature_to_kelvin` which only accepts `int`s or `floats`. That type annotation is wrong, too, but that's a separate issue. We could make `_convert_temperature_to_kelvin` accept strings, but that seems to be too relaxed – just `float`s would be better.

agriyakhetarpal:
I think the issue in #4963 is correct, but we do need to enforce our Mypy rules to catch things like this.

brosaplanella:
> Hi @brosaplanella, I think this change is incorrect? The type annotation in `temperature: float | None = None` doesn't accept strings, and we pass this attribute down to `pybamm.experiment._convert_temperature_to_kelvin` which only accepts `int`s or `float`s. That type annotation is wrong, too, but that's a separate issue. We could make `_convert_temperature_to_kelvin` accept strings, but that seems to be too relaxed – IMO, just `float`s would be better.

Hi! Note that the `_convert_temperature_to_kelvin` in `Experiment` actually calls it from `base_step` (https://github.com/pybamm-team/PyBaMM/blob/bc5f030146746fd4562e35dd186160b59900a047/src/pybamm/experiment/step/base_step.py#L542), which allows both for strings (with units) and float/int (assuming Kelvin). Therefore, the documentation should match that of the `base_step`.

I am not sure what's the best way forward so I didn't want to address this in this PR.

agriyakhetarpal (approved):
Ah, alright! I'm okay with the docs-only change in that case, but maybe we can leave #4963 open as we should clean up the inputs/outputs.

agriyakhetarpal:
Thanks, @brosaplanella!

PR #4968, opened 2025-04-10
Added "use lumped thermal capacity" option in lumped thermal model.

### LINKED ISSUE #4612: Create "Properly Lumped" thermal model
### Description

This would implement an alternative battery model where the heat capacity is given as a lumped parameter rather than for each cell component.

### Motivation

While the lumped thermal model in PyBaMM is physically informative and accurate, it poses a problem for practicing battery modelers who most of the time do not have detailed thermal characterization of each component of the cell.

### Possible Implementation

Rather than implementing a new model, I propose that we just add an option "use lumped thermal capacity" which is default "false", and just alter the existing lumped submodel code.

### Additional context

_No response_

Rishab87:
# Description

Added "use lumped thermal capacity" option in lumped thermal model so that we can simulate properly for those who do not have detailed thermal characterization of each component of the cell.

Fixes #4612 

## Type of change

# Important checks:


src/pybamm/input/parameters/lithium_ion/Ai2020.py
```diff
@@ -670,6 +670,7 @@ def get_parameter_values():
         "Initial concentration in negative electrode [mol.m-3]": 24108.0,
         "Initial concentration in positive electrode [mol.m-3]": 21725.0,
         "Initial temperature [K]": 298.15,
+        "Cell heat capacity [J.K-1.m-3]": 2.5e6,
```
aabills, line 673: In general, we try not to add parameters that were not in the original paper to the parameter sets, so could you delete this from all the parameter sets?

src/pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -337,6 +341,7 @@ def __init__(self, extra_options):
             "voltage as a state": ["false", "true"],
             "working electrode": ["both", "positive"],
             "x-average side reactions": ["false", "true"],
+            "use lumped capacity": ["false", "true"],
```
aabills, line 344: Let's rename this to "use lumped thermal capacity" for clarity, given that batteries also have an charge capacity.

tests/unit/test_models/test_submodels/test_thermal_lumped.py
```diff
@@ -0,0 +1,39 @@
...
+
+        var = sim._built_model.variables[
+            "Volume-averaged effective heat capacity [J.K-1.m-3]"
+        ]
+
+        t_eval = np.array([0])
+        y_eval = np.zeros((sim._built_model.len_rhs, 1))
+        heat_capacity_eval = var.evaluate(t_eval, y_eval)
+
+        assert abs(heat_capacity_eval - lumped_value) < lumped_value * 1e-10
```
aabills, line 39: I'm not sure that this is a great test. What is this trying to test exactly?

aabills: Also you can use np.assert_almost_equal

src/pybamm/parameters/lithium_ion_parameters.py
```diff
@@ -56,6 +56,7 @@ def _set_parameters(self):
         self.h_total = self.therm.h_total
         self.rho_c_p_eff = self.therm.rho_c_p_eff
         self.lambda_eff = self.therm.lambda_eff
+        self.cell_heat_capacity = self.therm.cell_heat_capacity
```
aabills, line 59: This is the same thing as rho_c_p_eff right? So why make a new parameter for it?

src/pybamm/parameters/thermal_parameters.py
```diff
@@ -40,6 +40,9 @@ def _set_parameters(self):
         # Initial temperature
         self.T_init = pybamm.Parameter("Initial temperature [K]")
 
+        # Lumped model heat capacity (volumetric)
+        self.cell_heat_capacity = pybamm.Parameter("Cell heat capacity [J.K-1.m-3]")
```
aabills, line 44: Do we want this to be volumetric? @rtimms I think we discussed this but can't remember what we decided.

aabills (changes requested):
Thanks @Rishab87 !

Overall it looks really great, there are just a few small issues. 

I'd also like to see an integration test where we ensure that the equation is doing what it is supposed to be doing, maybe by back calculating the lumped capacity and comparing it with the standard thermal model.

src/pybamm/parameters/lithium_ion_parameters.py
Rishab87: I think rho_c_p_eff calculates the effective volumetric heat capacity of the entire battery cell by combining the heat capacities of each component but cell_heat_capacity is the lumped heat capacity of the whole cell

src/pybamm/parameters/thermal_parameters.py
Rishab87: I think this should be volumetric as the current `rho_c_p_eff(T)` function returns a volumetric heat capacity and
`self.rhs = {T_vol_av: (Q_vol_av + Q_cr_vol_av + Q_cool_vol_av) / rho_c_p_eff_av}`.

Here, the Q terms are in [W.m-3]. When divided by the volumetric heat capacity [J.K-1.m-3], we get the temperature change rate [K.s-1].

So for consistency I think it should be volumetric.

tests/unit/test_models/test_submodels/test_thermal_lumped.py
Rishab87: I've changed this unit test a bit, it checks whether the heat capacity parameter is being used correctly when the option is enabled.

Rishab87:
@aabills thanks for the review!
I've added an integration test and made changes accordingly.

src/pybamm/parameters/thermal_parameters.py
rtimms: Yes I think volumetric makes sense since this parameter is a direct drop in for `rho_c_p_eff`

src/pybamm/models/full_battery_models/base_battery_model.py
```diff
@@ -224,6 +224,10 @@ class BatteryModelOptions(pybamm.FuzzyDict):
                 the respective porosity change) over the x-axis in Single Particle
                 Models, can be "false" or "true". Default is "false" for SPMe and
                 "true" for SPM.
```
rtimms, line 226: Can you add a check that raises an `OptionError` for incompatible options? There are some examples of this further down in this file.

src/pybamm/models/submodels/thermal/lumped.py
```diff
@@ -64,6 +67,14 @@ def get_coupled_variables(self, variables):
             Q_cr_W = pybamm.Scalar(0)
             Q_cr_vol_av = Q_cr_W
 
+        # Add lumped heat capacity if option is enabled
```
rtimms, line 70: It might be cleaner to have this option check in the base class rather than overwriting the variable for the  lumped model.

aabills:
Hey @Rishab87 I'll re-review once you've implemented Rob's changes

src/pybamm/models/full_battery_models/base_battery_model.py
Rishab87: ohh yeah, missed that adding

src/pybamm/models/submodels/thermal/lumped.py
Rishab87: right

Rishab87:
@rtimms, thanks for the review
I've made those changes.

@aabills it's ready to be re-reviewed

src/pybamm/parameters/lithium_ion_parameters.py
aabills: yeah I think it makes sense to keep it separate after further consideration.

aabills (changes requested):
Ok at this point it looks great substantively. The only thing is, the testing in unit/ doesn't really match what we did for the other models. Could you please add it to `unit/test_models/test_full_battery_models/base_lithium_ion_tests.py` and `base_lithium_ion_half_cell_tests.py`? You should be able to do both the option error and the check well posedness there. And then move the parameter test from unittests to test_thermal_parameters, which is in `test_lithium_ion_parameters.py`

Thanks so much!

Rishab87:
@aabills, shifted those tests and changed them accordingly

PR #4969, opened 2025-04-10
Update the termination state when skip_ok is enabled

### LINKED ISSUE #4961: [Bug]: skip_step option should return termination condition
### PyBaMM Version

v25.4.0

### Python Version

3.12.7

### Describe the bug

It would be helpful if the new `skip_step` option (#4839) could return empty solutions with the correct `termination` condition.

### Steps to Reproduce

Here is an example of an experiment with two steps. The second step cannot be initialized because the current constraint is violated. However, the termination returned is "final time" from the first step.
```
import pybamm
    [
        "Rest for 10 min",
        "Hold at 3.8V for 1h until > 5A",
    ]
)
sim = pybamm.Simulation(
    model,
    experiment=experiment,
)
solution = sim.solve()
print(f"Termination: {solution.termination}")
```

### Relevant log output

```shell

```

d-cogswell:
# Description

Update the termination state when `skip_ok` is enabled. The termination state is now updated when a step fails to initialize because of a constraint.

Fixes #4961 

## Type of change

# Important checks:


aabills (changes requested):
Thanks Dan! It looks great, could you add a test?

d-cogswell:
> Thanks Dan! It looks great, could you add a test?

@aabills how about if we wait to add a regression test until the terminations are standardized? The termination string will likely change, right now it is just "Event exceeded in initial conditions".

aabills (approved):

PR #4971, opened 2025-04-11
Automate source wrap

### LINKED ISSUE #1410: Automate source terms in FEM
When you solve an ODE (pointwise) using the FEM the variable on the lhs is multiplied by the mass matrix upon discretisation. If you wrap the rhs in `pybamm.source` then it also gets multiplied by the mass matrix so everything is fine. However, it would be easy to miss this. We should either NOT multiply the lhs by the mass matrix OR automate adding `pybamm.source` where it is needed. 

See this example 
```python3
import pybamm


class Model(pybamm.lithium_ion.BaseModel):
    def __init__(self, name="2+1D test"):
        super().__init__({"dimensionality": 2}, name)

        i = pybamm.PrimaryBroadcast(1, "current collector")
        a = pybamm.Variable("a", domain="current collector")
        b = pybamm.Variable(
            "b",
            domain="negative electrode",
        )
        c = pybamm.Variable("c", domain="current collector")

        self.rhs[a] = i
        self.rhs[b] = pybamm.PrimaryBroadcast(i, "negative electrode")
        self.rhs[c] = pybamm.source(i, c)

        self.initial_conditions[a] = pybamm.Scalar(0)
        self.initial_conditions[b] = pybamm.Scalar(0)
        self.initial_conditions[c] = pybamm.Scalar(0)

        self.variables = {
            "a": a,
            "b": b,
            "c": c,
            "b_av": pybamm.x_average(b),
            "true": i * pybamm.t,
        }

    def new_empty_copy(self):
        return pybamm.BaseModel.new_empty_copy(self)


model = Model()
var_pts = {
    var.x_n: 3,
    var.x_s: 3,
    var.x_p: 3,
    var.r_n: 3,
    var.r_p: 3,
    var.y: 3,
    var.z: 3,
}
sim = pybamm.Simulation(model, var_pts=var_pts)
sim.solve([0, 3600])
sim.plot(["a", "b_av", "c", "true"])
```

`a` gives the wrong result, but `b` and `c` are correct (because `b` has domain negative electrode it is discretised by FVM and because `c` uses `source`)

ArivoliR:
# Description

Fixes #1410 

## Type of change

# Important checks:

### DIFF
--- a/bandit.yml
+++ b/bandit.yml
@@ -1,2 +1,2 @@
 # To ignore false flagging of assert statements in tests by Codacy.
-skips: ['B101']
+skips: ["B101"]
--- a/src/pybamm/models/base_model.py
+++ b/src/pybamm/models/base_model.py
@@ -157,7 +157,10 @@ class BaseModel:
 
     @rhs.setter
     def rhs(self, rhs):
-        self._rhs = EquationDict("rhs", rhs)
+        wrapped_rhs = {}
+        for variable, expression in rhs.items():
+            wrapped_rhs[variable] = self.wrap_source(expression, variable)
+        self._rhs = EquationDict("rhs", wrapped_rhs)
 
     @property
     def algebraic(self):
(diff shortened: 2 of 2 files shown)


rtimms:
Thanks taking this on! For it to work properly we need to wrap all the terms that aren’t a spatial operator in source, eg see https://github.com/pybamm-team/PyBaMM/blob/9415db5037ca5978a91723ff9a64c3c44d85e0e0/src/pybamm/models/submodels/thermal/pouch_cell/pouch_cell_2D_current_collectors.py#L89

I’m not sure this is what your solution does. Can you add a test to show that `pybamm.laplacian(a) + b` gets turned into `pybamm.laplacian(a) + pybamm.source(b)` to show this works in the simplest case? Then we can go from there.

ArivoliR:
Okay I've made a huge mistake clearly, will try to figure out if there's a way to handle all these cases by just modifying the wrap_source function that I've made. Have my final exams going on, will continue my attempt at solving this issue in a few weeks

PR #4974, opened 2025-04-15
refactor: migrate citations.py from pybtex to bibtexparser

### LINKED ISSUE #3648: Switch to an active citations parser dependency to replace `pybtex`

gurukiran7:
# Description

This PR refactors `pybamm/citations.py` to fully remove the legacy `pybtex` dependency and migrate all citation management to `bibtexparser` (v2). The new implementation:

- Uses bibtexparser for reading, parsing, and validating BibTeX entries
- Updates the `register` method to handle both known keys and raw BibTeX strings
- Adds improved error handling and type validation
- Implements custom citation string formatting to closely match the previous output
- Issues warnings when overwriting existing citations with different content

Fixes #3648

# Important checks:

### DIFF
--- a/src/pybamm/citations.py
+++ b/src/pybamm/citations.py
@@ -2,6 +2,11 @@ import pybamm
 import os
 import warnings
 from sys import _getframe
+from bibtexparser import BibDatabase, UndefinedString
+from bibtexparser import BibTexParser
+from bibtexparser import BibTexWriter
+from bibtexparser import load as bibtex_load
+from bibtexparser.model import Entry
 from pybamm.util import import_optional_dependency
 
 
@@ -21,24 +26,17 @@ class Citations:
     """
 
     _module_import_error = False
-    # Set of citation keys that have been registered
-    _papers_to_cite: set
-    # Set of unknown citations to parse with pybtex
-    _unknown_citations: set
-    # Dict mapping citation tags for use when registering citations
-    _citation_tags: dict
+    
+    _all_citations: dict = {}
 
     def __init__(self):
         self._check_for_bibtex()
-        # Dict mapping citations keys to BibTex entries
-        self._all_citations: dict[str, str] = dict()
-
         self.read_citations()
         self._reset()
 
     def _check_for_bibtex(self):
         try:
-            import_optional_dependency("pybtex")
+            import_optional_dependency("bibtexparser")
         except ModuleNotFoundError:
             self._module_import_error = True
 
@@ -46,7 +44,7 @@ class Citations:
         """Reset citations to default only (only for testing purposes)"""
         self._papers_to_cite = set()
         self._unknown_citations = set()
-        self._citation_tags = dict()
+        self._citation_tags = {}
         # Register the PyBaMM paper and the NumPy paper
         self.register("Sulzer2021")
         self.register("Harris2020")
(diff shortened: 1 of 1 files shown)


agriyakhetarpal:
Thanks for starting this, @gurukiran7! To fix the tests, you'll need to add `bibtexparser` as in the `[cite]` set of optional dependencies in `pyproject.toml`. You may also run the tests locally to see if they pass and ease your development, as we'll need to trigger the workflows every time you push a commit: https://docs.pybamm.org/en/latest/source/user_guide/contributing.html#testing.

gurukiran7:
@agriyakhetarpal I've updated the tests to align with the new citations.py implementation. At the moment, only a few tests are passing I'm still working on the rest. I want your feedback on the implementation so far

agriyakhetarpal:
Yes, that's alright, but the CI logs still don't reflect it, so it is difficult to see all the failures. I pulled your branch to my machine, and I see the same error: `ImportError: cannot import name 'BibDatabase' from 'bibtexparser'`. I think there might be a few commits on your local branch you haven't pushed?

https://github.com/search?q=repo:sciunto-org/python-bibtexparser+%22bibdatabase%22&type=code also reports no results, so I'm not sure where you're getting that API from.

It's not in the documentation either: https://bibtexparser.readthedocs.io/en/main/search.html?q=bibdatabase

agriyakhetarpal:
Yes, that's alright, but the CI logs still don't reflect it – it is hence difficult to see all the failures. I pulled your branch to my machine, and I see the same error: `ImportError: cannot import name 'BibDatabase' from 'bibtexparser'`. I think there might be a few commits on your local branch you haven't pushed?

https://github.com/search?q=repo:sciunto-org/python-bibtexparser+%22bibdatabase%22&type=code also reports no results, so I'm not sure where you're getting that attribute from.

It's not in the documentation either: https://bibtexparser.readthedocs.io/en/main/search.html?q=bibdatabase

gurukiran7:
Thanks for pointing that out! the ImportError stems from my use of BibDatabase, which was part of bibtexparser v1 but has been removed in v2. I mistakenly mixed v1-style imports with the newer v2 API. I’ve now updated the code to use Entry from bibtexparser.model instead and removed all references to BibDatabase. That should fix the import issue  I’ll push the changes shortly.

gurukiran7:
Hi @agriyakhetarpal , I've pushed updated code, and most methods and their tests are passing as expected However, there's an issue with the register method, causing tests like test_citations and test_overwrite_citation to fail. Despite debugging, I couldn't resolve  the issue, as it seems related to how citations are added to _all_citations and _papers_to_cite.
could u pls review the register method and suggest a fix?

agriyakhetarpal:
Thanks! Could you please run the whole test suite using `nox` directly? Running `pytest -k test_citations` is fine locally, but we're running all the tests in CI and that's different from running these tests in specific as you may see.

Also, you would need to fix the `style` job to allow the test suite to run.

src/pybamm/citations.py
```diff
@@ -2,6 +2,10 @@
 import os
 import warnings
 from sys import _getframe
+import bibtexparser
+from bibtexparser import load as bibtex_load
+from bibtexparser import loads as bibtex_loads
+from bibtexparser.library import Library
```
kratman, line 8: I believe all the citation stuff was optional before, is there a reason it is not optional now?

gurukiran7:
@agriyakhetarpal I've pushed the latest updated code addressing most of the test failures and feedback

 Refactored the register method and _add_citation to support bibtexparser v2+
Made citation handling optional using import_optional_dependency
 Cleaned up all formatting/style issues (ruff-format, pre-commit now passes)
  if you have any suggestions on expected behavior or test intentn pls let me know

gurukiran7:
@agriyakhetarpal , @kratman , I've pushed the latest code please review when you get the chance

src/pybamm/citations.py
gurukiran7: updated to optional

PR #4982, opened 2025-04-17
Update `AlgebraicSolver` tolerances + default `ESOH` solver method

MarcBerliner:
# Description

Previously, the `AlgebraicSolver` would only declare success if the absolute tolerance of the residuals were below the `tol` threshold. However, the scipy solvers would frequently terminate if other thresholds were hit, such as `xtol` or `gtol`, and `AlgebraicSolver` may erroneously throw an error even after a successful solve

This PR:
- tightens `xtol` and `gtol` from `1e-8` to `1e-12` to reflect the fact that we desire `all(abs(f(x)) < tol)`
- updates the success criterion to terminate either if the solver successfully exits or `all(abs(f(x)) < tol)` is satisfied
- also changes the default solver in `ElectrodeSOHHalfCell` to trust-region instead of a minimizer

## Type of change

# Important checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,14 @@
 # [Unreleased](https://github.com/pybamm-team/PyBaMM/)
 
+## Bug fixes
+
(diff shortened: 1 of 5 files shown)


kratman (approved):
I think this is fine as long as the tests are passing

PR #4996, opened 2025-04-25
Update docs, examples and tests to use `IDAKLUSolver`

MarcBerliner:
# Description

Following the default solver change to `IDAKLUSolver`, this PR changes the examples, docs, and main tests to use the `IDAKLUSolver`. The one exception is the lead acid models, which do not work with the `IDAKLUSolver`. 

## Type of change

# Important checks:


docs/source/examples/notebooks/getting_started/tutorial-8-solver-options.ipynb
```diff
@@ -89,26 +77,26 @@
...
    "metadata": {},
    "outputs": [
     {
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "Safe mode solve time: 137.215 ms\n",
-      "Fast mode solve time: 92.051 ms\n"
+      "Safe mode solve time: 16.418 ms\n",
+      "Fast mode solve time: 83.043 ms\n"
```
kratman, line 88: This does not seem to match what is above anymore. Since the old one was about the safe/fast modes of casadi

docs/source/examples/notebooks/initialize-model-with-solution.ipynb
```diff
@@ -89,7 +89,9 @@
     "param = model.default_parameter_values\n",
     "current_interpolant = pybamm.Interpolant(drive_cycle[:, 0], drive_cycle[:, 1], pybamm.t)\n",
     "# set drive cycle\n",
-    "param[\"Current function [A]\"] = current_interpolant"
+    "param[\"Current function [A]\"] = current_interpolant\n",
+    "# set upper voltage cut-off high to avoid the cutoff\n",
+    "param[\"Upper voltage cut-off [V]\"] = 4.3"
```
kratman, line 94: Is this a symptom of some other issue? Seems weird to have to change the parameter set

kratman (commented):
Partial review. I will finish the reset later

docs/source/examples/notebooks/solvers/dae-solver.ipynb
```diff
@@ -297,7 +297,7 @@
     "disc.process_model(model)\n",
     "print(f\"y0_guess={model.concatenated_initial_conditions.evaluate().flatten()}\")\n",
     "\n",
-    "dae_solver = pybamm.CasadiSolver(mode=\"safe\")\n",
+    "dae_solver = pybamm.IDAKLUSolver()\n",
```
kratman, line 300: Can we remove these as well

kratman (commented):
Minor stuff, most of the rest looks good

docs/source/examples/notebooks/solvers/dae-solver.ipynb
MarcBerliner: I think we should keep this one, because they intentionally use the solver interface to explain how it works under the hood

docs/source/examples/notebooks/initialize-model-with-solution.ipynb
MarcBerliner: I modified the example and removed the line changing the upper voltage cutoff

kratman (changes requested):
One last change, then I will approve

PR #5006, opened 2025-05-06
Process Fucntion refactor to Enforce return_jacp_stacked=True

### LINKED ISSUE #4970: refactor: always return_jacp_stacked
The `process` function in `base_solver.py` is used to either take the sensitivities wrt each parameter in turn, or take the sensitivities wrt the parameter vector (if true). The later is neccessary to calculate online sensitivities if `output_variables` is set.

This refactor proposes to always have return_jacp_stacked=True, in order to simplify the code that needs to deal with each case individually.

Might have to do this anyway for https://github.com/pybamm-team/PyBaMM/issues/4547, but I'll just put this issue in place for a reminder to do this.

RohitP2005:
# Description

Refactored the process function in `base_solver.py`to simplify its logic and improve maintainability. 

Fixes #4970

## Type of change

 Refactor (non-breaking change that improves code structure) to the process function in the base solver to enforce `return_jacp_stacked=True `.

# Important checks:

### DIFF
--- a/src/pybamm/solvers/base_solver.py
+++ b/src/pybamm/solvers/base_solver.py
@@ -1512,8 +1512,6 @@ def process(
         function evaluators created will have this base name
     use_jacobian: bool, optional
         whether to return Jacobian functions
-    return_jacp_stacked: bool, optional
-        returns Jacobian function wrt stacked parameters instead of jacp
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


RohitP2005:
@martinjrobins 
I've refactored the process function in base_solver.py to simplify the logic  Could you please take a look at the changes? If everything looks good, we can go ahead and check the CI tests.

Thanks!

src/pybamm/solvers/base_solver.py
```diff
@@ -1562,103 +1550,55 @@ def report(string):
...
             )
             jacp = func.get_sensitivities()
-        if use_jacobian:
-            report(f"Calculating jacobian for {name} using jax")
-            jac = func.get_jacobian()
-            jac_action = func.get_jacobian_action()
-        else:
-            jac = None
-            jac_action = None
+        jac = func.get_jacobian() if use_jacobian else None
```
martinjrobins, line 1553: I think the single if statement is less complex then multiple, plus you are removing the `report`, so I'd disagree with this change

src/pybamm/solvers/base_solver.py
```diff
@@ -1562,103 +1550,55 @@ def report(string):
...
+            raise pybamm.SolverError(
                 "Sensitivies are no longer supported for the python "
                 "evaluator. Please use `convert_to_format = 'casadi'`, or `jax` "
                 "to calculate sensitivities."
             )
-
-        else:
-            jacp = None
-
-        if use_jacobian:
```
martinjrobins, line 1587: same with this one

src/pybamm/solvers/base_solver.py
```diff
@@ -1562,103 +1550,55 @@ def report(string):
...
         casadi_expression = symbol.to_casadi(t_casadi, y_casadi, inputs=p_casadi)
-        # Add sensitivity vectors to the rhs and algebraic equations
         jacp = None
 
         if model.calculate_sensitivities:
-            report(
-                f"Calculating sensitivities for {name} with respect "
-                f"to parameters {model.calculate_sensitivities} using "
-                "CasADi"
+            report(f"Calculating sensitivities for {name} using CasADi")
```
martinjrobins, line 1584: its useful to have the printout of `model.calculate_sensitivities` here, so keep this

src/pybamm/solvers/base_solver.py
```diff
@@ -1562,103 +1550,55 @@ def report(string):
...
+            jacp = casadi.Function(
+                f"d{name}_dp",
+                [t_casadi, y_casadi, p_casadi_stacked],
+                [casadi.jacobian(casadi_expression, p_casadi_stacked)],
             )
-            # Compute derivate wrt p-stacked (can be passed to solver to
-            # compute sensitivities online)
-            if return_jacp_stacked:
```
martinjrobins, line 1621: removing this if statement is only a small part of this change, you will also need to change the behaviour of the solvers that depend on this functionality (I believe, but I'll run the tests so we can check)

martinjrobins (changes requested):
Hi @RohitP2005, I think there is much more to do on this (than just removing the arg. Have you run the tests locally?

RohitP2005:
> Hi @RohitP2005, I think there is much more to do on this (than just removing the arg. Have you run the tests locally?

Yes , I ran the tests locally and there are errors that needs to addressed.
we fix those issues and make the requested changes.

PR #5008, opened 2025-05-09
bug: fix sensitivity calculation for discrete time sum

### LINKED ISSUE #5007: [Bug]: pybamm.DiscreteTimeSum sensitivities not being summed over time
### PyBaMM Version

develop

### Python Version

3.10

### Describe the bug

When calculating the sensitivities for the `pybamm.DiscreteTimeSum`, the sum over time isn't being performed, giving a sensitivity vector of the same dimension as the number of time points. This is unsuprising since this is not currently being tested either, so tests and a fix are needed

### Steps to Reproduce

create a model with a variable using `pybamm.DiscreteTimeSum`, then calculate the sensitivities and check the dimension of the result

### Relevant log output

```shell

```

martinjrobins:
# Description

Fixes #5007

## Type of change

# Important checks:


src/pybamm/solvers/processed_variable.py
```diff
@@ -34,11 +36,13 @@ class ProcessedVariable:
 
     def __init__(
         self,
+        variable: str,
         base_variables,
         base_variables_casadi,
         solution,
         time_integral: Optional[pybamm.ProcessedVariableTimeIntegral] = None,
     ):
+        self._variable_name = variable
```
MarcBerliner, line 45: We use the word `variable` a lot in this class. Can we change it to `name` throughout instead?
```suggestion
    def __init__(
        self,
        name: str,
        base_variables,
        base_variables_casadi,
        solution,
        time_integral: Optional[pybamm.ProcessedVariableTimeIntegral] = None,
    ):
        self._name = variable
```

MarcBerliner (changes requested):
Thanks Martin! Looks good, just some minor things. For consistency, can you add the `name` argument/property to the other processed variable classes?

Question (non-blocking):
- If hermite interpolation is available, can we treat `dvar_dy_eval`, `dy_dp`, and `dvar_dp_eval` as distinct process variables, hermite interpolate them onto `self.time_integral.discrete_times`, and then compute the discrete sum?

src/pybamm/solvers/processed_variable.py
martinjrobins: yup, will do

martinjrobins:
> * If hermite interpolation is available, can we treat `dvar_dy_eval`, `dy_dp`, and `dvar_dp_eval` as distinct process variables, hermite interpolate them onto `self.time_integral.discrete_times`, and then compute the discrete sum?

Yea, that would be nicer, we'd need their time derivatives though I think? I think these can be extracted during integration using `IDAGetSensDky`, we'd have to do that and return them as part of the solution.

src/pybamm/solvers/processed_variable.py
```diff
@@ -458,6 +462,26 @@ def initialise_sensitivity_explicit_forward(self):
...
+            # post fix for discrete time integral won't give correct result
+            # if ts are not equal to the discrete times. Raise error
+            # in this case
+            if (
+                self.time_integral is not None
+                and self.time_integral.method == "discrete"
+            ):
+                if not (
+                    len(ts) == len(self.time_integral.discrete_times)
+                    and np.all(np.abs(ts - self.time_integral.discrete_times) < 1e-10)
```
BradyPlanden, line 475: Seems like a good spot to use allclose 

```suggestion
                    and np.allclose(ts, self.time_integral.discrete_times, atol=1e-10))
```

MarcBerliner (approved):
Thanks!

PR #5037, opened 2025-05-30
fix: DiscreteTimeSum sensitivity calculation dimensions

### LINKED ISSUE #5036: [Bug]: DiscreteTimeSum sensitivities error for input dimensions > 1
### PyBaMM Version

25.6

### Python Version

3.12

### Describe the bug

Variables created as DiscreteTimeSum will have dimension errors during processing due to the numpy summation located here:https://github.com/pybamm-team/PyBaMM/blob/e1c0c87064f3e08f9fe09c00af72b57150c3ae9a/src/pybamm/solvers/processed_variable.py#L527

### Steps to Reproduce

Compute sensitivities on a DiscreteTimeSum variable with input dimensions > 1

### Relevant log output

```shell

```

BradyPlanden:
# Description

Fixes #5036

## Type of change

# Important checks:

### DIFF
--- a/src/pybamm/solvers/processed_variable.py
+++ b/src/pybamm/solvers/processed_variable.py
@@ -466,10 +466,7 @@ class ProcessedVariable:
             # post fix for discrete time integral won't give correct result
             # if ts are not equal to the discrete times. Raise error
             # in this case
-            if (
-                self.time_integral is not None
-                and self.time_integral.method == "discrete"
-            ):
+            if self._is_discrete_time_method():
                 if not (
                     len(ts) == len(self.time_integral.discrete_times)
                     and np.allclose(ts, self.time_integral.discrete_times, atol=1e-10)
@@ -485,6 +482,8 @@ class ProcessedVariable:
             all_S_var.append(S_var)
 
         S_var = casadi.vertcat(*all_S_var)
+        if self._is_discrete_time_method():
+            S_var = S_var.T
         sensitivities = {"all": S_var}
 
         # Add the individual sensitivity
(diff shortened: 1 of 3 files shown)


src/pybamm/solvers/processed_variable.py
```diff
@@ -485,6 +482,8 @@ def initialise_sensitivity_explicit_forward(self):
             all_S_var.append(S_var)
 
         S_var = casadi.vertcat(*all_S_var)
+        if self._is_discrete_time_method():
+            S_var = S_var.T
```
martinjrobins: its unclear here why this is needed, can we add a comment?

martinjrobins (changes requested):
Thanks @BradyPlanden. Would you mind also adding some comments here on the dimensions of the sensitivity variables and why the transpose is needed?

src/pybamm/solvers/processed_variable.py
BradyPlanden: Converting the `casadi.verticat` to `np.vstack` fixed the dimension issue in the below loop. It appears the casadi DM object isn't as flexible as `np.ndarray` when slicing.

noxfile.py
```diff
@@ -6,7 +6,7 @@
 
 # Options to modify nox behaviour
 nox.options.default_venv_backend = "uv|virtualenv"
-nox.options.reuse_existing_virtualenvs = True
+nox.options.reuse_existing_virtualenvs = False
```
martinjrobins: why did this change?

martinjrobins (dismissed):
yay, don't have to stick `full()` everywhere in the tests :)

noxfile.py
BradyPlanden: Because I accidentally committed it — oops! 😄

PR #5044, opened 2025-06-03
feat: generalise pybamm.DiscreteTimeSum so that it can be inside an expression

### LINKED ISSUE #5039: DiscreteTimeSum as child node
### Description

At the moment the DiscreteTimeSum is restricted in its use. You can only use it as the root node in an expression tree, because in Solution/ProcessedVariable only the root node is looked at when detecting this type of node (along with the integral node).

This issue proposes to modify the processing so that you can embed the DiscreteTimeSum as a child node, for example a sum like the Minkowski sum which raises the sum to the power of 1/p

$$
( \sum_i f(x_i) )^1/p
$$$

### Motivation

_No response_

### Possible Implementation

_No response_

### Additional context

_No response_

martinjrobins:
# Description

Fixes #5039

## Type of change

# Important checks:


kratman:
@martinjrobins The unrelated CI issues should be at least temporarily.

tests/unit/test_solvers/test_solution.py
```diff
@@ -498,10 +498,45 @@ def test_solution_evals_with_inputs(self):
...
+
+    _solver_classes = [
+        (pybamm.CasadiSolver, False),
+        (pybamm.IDAKLUSolver, False),
+        (pybamm.CasadiSolver, True),
+        (pybamm.IDAKLUSolver, True),
+    ]
+
+    @pytest.mark.parametrize("solver_class,use_post_sum", _solver_classes)
+    def test_discrete_data_sum(self, solver_class, use_post_sum):
```
BradyPlanden, line 541: Dropping a comment to test against an SPM or similar model. When checking on the PyBOP side this test ran, but the Minkowski with an SPM produces a N-dimensional array instead of the summed node. See https://github.com/pybop-team/PyBOP/blob/3ff1d5155bbb37fc0e1a889117341336b91ddbbd/tests/unit/test_cost.py#L77 for reference.

martinjrobins: this was due to a bug if the nstates of the model was different to the nstates of the sum node, fixed now and test modified to uncover this. Can you see if this works now @BradyPlanden?

BradyPlanden: Thanks @martinjrobins - I can confirm it's fixed!

BradyPlanden (approved):
LGTM - thanks @martinjrobins!

PR #5048, opened 2025-06-04
Remove pytest pin

kratman:
# Description

This removes the pin on PyTest and fixes any remaining issues with the coverage in CI

## Type of change

# Important checks:


pyproject.toml
```diff
@@ -258,6 +258,9 @@ log_date_format = "%Y-%m-%d %H:%M:%S"
 [tool.coverage.run]
 source = ["src/pybamm"]
 concurrency = ["multiprocessing"]
+# Parse warnings have existed for a while, but became failures in pytest 8.4.0
+disable_warnings = ["couldnt-parse"]
```
kratman, line 262: I don't like doing this, but I checked and it does appear that these errors still appear in older versions of pytest. The difference is that the handling of warnings/errors changed

PR #5052, opened 2025-06-09
Deprecate python 3.9 support 

kratman:
# Description

This is earlier than necessary (~October), but it makes the switch to numpy 2.0 easier.

Related: #4183 

## Type of change

# Important checks:

### DIFF
--- a/src/pybamm/discretisations/discretisation.py
+++ b/src/pybamm/discretisations/discretisation.py
@@ -385,7 +385,7 @@ class Discretisation:
                 if first_child not in bc_keys:
                     internal_bcs.update({first_child: {"left": lbc, "right": rbc}})
 
-                for current_child, next_child in itertools.pairwise(children):
+                for current_child, next_child in itertools.pairwise(children[1:]):
                     lbc = rbc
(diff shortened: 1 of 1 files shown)


src/pybamm/discretisations/discretisation.py
```diff
@@ -384,7 +385,7 @@ def boundary_gradient(left_symbol, right_symbol):
                 if first_child not in bc_keys:
                     internal_bcs.update({first_child: {"left": lbc, "right": rbc}})
 
-                for current_child, next_child in zip(children[1:-1], children[2:]):
+                for current_child, next_child in itertools.pairwise(children):
```
kratman: This automatic change did not seem to break the tests, but appears to be incorrect
```suggestion
                for current_child, next_child in itertools.pairwise(children[1:]):
```

agriyakhetarpal (approved):
Thanks! I'm okay with dropping support for Python 3.9 (I'm coming from @Saransh-cpp's comment at #4490), but I would delegate the decision to others.

brosaplanella (approved):
Looks good to me, thanks!

PR #5056, opened 2025-06-12
[GSOC 2025] Third Party Model Dispatching

medha-14:
# Description

Fixes # (issue)

## Type of change

# Important checks:

### DIFF
--- a/src/pybamm/__init__.py
+++ b/src/pybamm/__init__.py
@@ -61,7 +61,7 @@ from .expression_tree.operations.evaluate_python import JaxCooMatrix
 from .expression_tree.operations.jacobian import Jacobian
 from .expression_tree.operations.convert_to_casadi import CasadiConverter
 from .expression_tree.operations.unpack_symbols import SymbolUnpacker
-
+from .expression_tree.operations.serialise import Serialise
 # Model classes
 from .models.base_model import BaseModel
 from .models.event import Event
@@ -149,8 +149,6 @@ from .meshes.scikit_fem_submeshes import (
 
 # Serialisation
 from .models.base_model import load_model
-from .models.base_model import load_custom_model
-
 # Spatial Methods
 from .spatial_methods.spatial_method import SpatialMethod
 from .spatial_methods.zero_dimensional_method import ZeroDimensionalSpatialMethod
--- a/src/pybamm/__init__.py
+++ b/src/pybamm/__init__.py
@@ -149,6 +149,7 @@ from .meshes.scikit_fem_submeshes import (
 
 # Serialisation
 from .models.base_model import load_model
+from .models.base_model import load_custom_model
 
 # Spatial Methods
 from .spatial_methods.spatial_method import SpatialMethod
--- a/src/pybamm/expression_tree/operations/serialise.py
+++ b/src/pybamm/expression_tree/operations/serialise.py
@@ -143,6 +143,51 @@ class Serialise:
         with open(filename + ".json", "w") as f:
             json.dump(model_json, f)
 
+    def save_custom_model(
+        self,
+        model: pybamm.BaseModel,
+        mesh: pybamm.Mesh | None = None,
+        variables: pybamm.FuzzyDict | None = None,
+        filename: str | None = None,
+    ):
+        if model.is_discretised is False:
+            raise NotImplementedError(
+                "PyBaMM can only serialise a discretised, ready-to-solve model."
+            )
+
+        model_json = {
+            "pybamm_version": pybamm.__version__,
+            "name": model.name,
+            "options": model.options,
+            "bounds": [bound.tolist() for bound in model.bounds],  # type: ignore[attr-defined]
+            "concatenated_rhs": self._SymbolEncoder().default(model._concatenated_rhs),
+            "concatenated_algebraic": self._SymbolEncoder().default(
+                model._concatenated_algebraic
+            ),
+            "concatenated_initial_conditions": self._SymbolEncoder().default(
+                model._concatenated_initial_conditions
+            ),
+            "events": [self._SymbolEncoder().default(event) for event in model.events],
+            "mass_matrix": self._SymbolEncoder().default(model.mass_matrix),
+            "mass_matrix_inv": self._SymbolEncoder().default(model.mass_matrix_inv),
+        }
+
+        if mesh:
+            model_json["mesh"] = self._MeshEncoder().default(mesh)
+
+        if variables:
+            if model._geometry:
+                model_json["geometry"] = self._deconstruct_pybamm_dicts(model._geometry)
+            model_json["variables"] = {
+                k: self._SymbolEncoder().default(v) for k, v in dict(variables).items()
+            }
+
+        if filename is None:
+            filename = model.name + "_" + datetime.now().strftime("%Y_%m_%d-%p%I_%M")
+
+        with open(filename + ".json", "w") as f:
+            json.dump(model_json, f)
+
     def load_model(
         self, filename: str, battery_model: pybamm.BaseModel | None = None
     ) -> pybamm.BaseModel:
@@ -234,6 +279,76 @@ class Serialise:
             The PyBaMM battery model to use has not been provided.
             """
         )
+
+    def load_custom_model(
+        self,
+        filename: str,
+        battery_model: pybamm.BaseModel,
+    ) -> pybamm.BaseModel:
+        """
+        Loads a discretised custom PyBaMM model from a JSON file.
+
+        Requires the user to pass the correct model class (`battery_model`) since
+        the saved file does not contain any class metadata.
+
+        Parameters
+        ----------
+        filename: str
+            Path to the JSON file containing the serialised model.
+        battery_model: :class:`pybamm.BaseModel`
+            An *instance* of the model class to be populated with the data.
+
+        Returns
+        -------
+        :class:`pybamm.BaseModel`
+            A reconstructed PyBaMM model ready to solve.
+        """
+        with open(filename) as f:
+            model_data = json.load(f)
+
+        recon_model_dict = {
+            "name": model_data["name"],
+            "options": self._convert_options(model_data["options"]),
+            "bounds": tuple(np.array(bound) for bound in model_data["bounds"]),
+            "concatenated_rhs": self._reconstruct_expression_tree(
+                model_data["concatenated_rhs"]
+            ),
+            "concatenated_algebraic": self._reconstruct_expression_tree(
+                model_data["concatenated_algebraic"]
+            ),
+            "concatenated_initial_conditions": self._reconstruct_expression_tree(
+                model_data["concatenated_initial_conditions"]
+            ),
+            "events": [
+                self._reconstruct_expression_tree(event)
+                for event in model_data["events"]
+            ],
+            "mass_matrix": self._reconstruct_expression_tree(model_data["mass_matrix"]),
+            "mass_matrix_inv": self._reconstruct_expression_tree(
+                model_data["mass_matrix_inv"]
+            ),
+        }
+
+        recon_model_dict["geometry"] = (
+            self._reconstruct_pybamm_dict(model_data["geometry"])
+            if "geometry" in model_data
+            else None
+        )
+
+        recon_model_dict["mesh"] = (
+            self._reconstruct_mesh(model_data["mesh"]) if "mesh" in model_data else None
+        )
+
+        recon_model_dict["variables"] = (
+            {
+                k: self._reconstruct_expression_tree(v)
+                for k, v in model_data["variables"].items()
+            }
+            if "variables" in model_data
+            else None
+        )
+
+        return battery_model.deserialise(recon_model_dict)
 
     # Helper functions
 
--- a/src/pybamm/expression_tree/operations/serialise.py
+++ b/src/pybamm/expression_tree/operations/serialise.py
@@ -6,6 +6,9 @@ import json
 import importlib
 import numpy as np
 import re
+from enum import Enum
+import inspect
+import numbers
 
 
 class Serialise:
@@ -143,51 +146,6 @@ class Serialise:
         with open(filename + ".json", "w") as f:
             json.dump(model_json, f)
 
-    def save_custom_model(
-        self,
-        model: pybamm.BaseModel,
-        mesh: pybamm.Mesh | None = None,
-        variables: pybamm.FuzzyDict | None = None,
-        filename: str | None = None,
-    ):
-        if model.is_discretised is False:
-            raise NotImplementedError(
-                "PyBaMM can only serialise a discretised, ready-to-solve model."
-            )
-
-        model_json = {
-            "pybamm_version": pybamm.__version__,
-            "name": model.name,
-            "options": model.options,
-            "bounds": [bound.tolist() for bound in model.bounds],  # type: ignore[attr-defined]
-            "concatenated_rhs": self._SymbolEncoder().default(model._concatenated_rhs),
-            "concatenated_algebraic": self._SymbolEncoder().default(
-                model._concatenated_algebraic
-            ),
-            "concatenated_initial_conditions": self._SymbolEncoder().default(
-                model._concatenated_initial_conditions
-            ),
-            "events": [self._SymbolEncoder().default(event) for event in model.events],
-            "mass_matrix": self._SymbolEncoder().default(model.mass_matrix),
-            "mass_matrix_inv": self._SymbolEncoder().default(model.mass_matrix_inv),
-        }
-
-        if mesh:
-            model_json["mesh"] = self._MeshEncoder().default(mesh)
-
-        if variables:
-            if model._geometry:
-                model_json["geometry"] = self._deconstruct_pybamm_dicts(model._geometry)
-            model_json["variables"] = {
-                k: self._SymbolEncoder().default(v) for k, v in dict(variables).items()
-            }
-
-        if filename is None:
-            filename = model.name + "_" + datetime.now().strftime("%Y_%m_%d-%p%I_%M")
-
-        with open(filename + ".json", "w") as f:
-            json.dump(model_json, f)
-
     def load_model(
         self, filename: str, battery_model: pybamm.BaseModel | None = None
     ) -> pybamm.BaseModel:
@@ -279,6 +237,69 @@ class Serialise:
             The PyBaMM battery model to use has not been provided.
             """
         )
+
+    def _json_encoder(obj):
+        if isinstance(obj, pybamm.Scalar):
+            return obj.evaluate()
+
+        if isinstance(obj, (np.integer, np.floating)):
+            return obj.item()
+        if isinstance(obj, np.ndarray):
+            return obj.tolist()
+        if isinstance(obj, Enum):
+            return obj.name
+        raise TypeError(
+            f"Object of type {obj.__class__.__name__} is not JSON serializable"
+        )
+
+    @staticmethod
+    def save_custom_model(model, param_values, filename=None):
+        """
+        Save the custom PyBaMM model and parameters to a JSON file.
+        """
+        model_json = {
+            "pybamm_version": pybamm.__version__,
+            "name": model.name,
+            "options": model.options,
+            "rhs": {
+                str(k): Serialise.convert_symbol_to_json(v)
+                for k, v in model.rhs.items()
+            },
+            "algebraic": {
+                str(k): Serialise.convert_symbol_to_json(v)
+                for k, v in model.algebraic.items()
+            },
+            "initial_conditions": {
+                str(k): Serialise.convert_symbol_to_json(v)
+                for k, v in model.initial_conditions.items()
+            },
+            "boundary_conditions": {
+                str(var): {
+                    side: [Serialise.convert_symbol_to_json(expr), btype]
+                    for side, (expr, btype) in conds.items()
+                }
+                for var, conds in model.boundary_conditions.items()
+            },
+            "events": [
+                {
+                    "name": event.name,
+                    "expression": Serialise.convert_symbol_to_json(event.expression),
+                    "event_type": event.event_type,
+                }
+                for event in model.events
+            ],
+            "parameters": Serialise.convert_parameter_values_to_json(param_values),
+            "variables": {
+                str(name): Serialise.convert_symbol_to_json(expr)
+                for name, expr in model.variables.items()
+            },
+        }
+
+        if filename is None:
+            filename = model.name + "_" + datetime.now().strftime("%Y_%m_%d-%I_%M_%p")
+
+        with open(filename + ".json", "w") as f:
+            json.dump(model_json, f, indent=2, default=Serialise._json_encoder)
 
     def load_custom_model(
         self,
@@ -512,3 +533,217 @@ class Serialise:
             return tuple(self._convert_options(item) for item in d)
         else:
             return d
+
+    @staticmethod
+    def convert_parameter_values_to_json(parameter_values, filename=None):
+        """
+        Converts a ParameterValues object to a JSON-serializable dictionary and optionally
+        saves it to a file.
+
+        Parameters
+        ----------
+        parameter_values : ParameterValues
+            The ParameterValues object to convert
+
+        filename : str, optional
+            The filename to save the JSON file to. If not provided, the dictionary is
+            not saved.
+        """
+        parameter_values_dict = {}
+
+        for k, v in parameter_values.items():
+            if callable(v):
+                parameter_values_dict[k] = Serialise.convert_symbol_to_json(
+                    Serialise.convert_function_to_symbolic_expression(v, k)
+                )
+            else:
+                parameter_values_dict[k] = Serialise.convert_symbol_to_json(v)
+
+        if filename is not None:
+            with open(filename, "w") as f:
+                json.dump(parameter_values_dict, f, indent=2)
+
+        return parameter_values_dict
+
+    @staticmethod
+    def convert_symbol_to_json(symbol):
+        """
+        Converts a PyBaMM symbolic expression to a JSON-serializable dictionary
+
+        Parameters
+        ----------
+        symbol : pybamm.Symbol
+            The PyBaMM symbolic expression to convert
+
+        Returns
+        -------
+        dict
+            The JSON-serializable dictionary
+        """
+        if isinstance(symbol, numbers.Number | list):
+            return symbol
+        elif isinstance(symbol, pybamm.Time):
+            return {"type": "Time"}
+        elif isinstance(symbol, pybamm.Parameter):
+            # Parameters are stored with their type and name
+            return {"type": "Parameter", "name": symbol.name}
+        elif isinstance(symbol, pybamm.Scalar):
+            # Scalar values are stored with their numerical value
+            return {"type": "Scalar", "value": symbol.value}
+        elif isinstance(symbol, pybamm.BinaryOperator | pybamm.UnaryOperator):
+            # Operators (like +, -, *, /) are stored with their type and operands
+            return {
+                "type": symbol.__class__.__name__,
+                "children": [
+                    Serialise.convert_symbol_to_json(c) for c in symbol.children
+                ],
+            }
+        elif isinstance(symbol, pybamm.SpecificFunction):
+            if symbol.__class__ == pybamm.SpecificFunction:
+                raise NotImplementedError("SpecificFunction is not supported")
+            else:
+                # Subclasses of SpecificFunction (e.g. Exp, Sin, etc.) can be reconstructed
+                # from only the children
+                return {
+                    "type": symbol.__class__.__name__,
+                    "children": [
+                        Serialise.convert_symbol_to_json(c) for c in symbol.children
+                    ],
+                }
+        elif isinstance(symbol, pybamm.Interpolant):
+            return {
+                "type": symbol.__class__.__name__,
+                "x": [x.tolist() for x in symbol.x],
+                "y": symbol.y.tolist(),
+                "children": [
+                    Serialise.convert_symbol_to_json(c) for c in symbol.children
+                ],
+                "name": symbol.name,
+                "interpolator": symbol.interpolator,
+                "entries_string": symbol.entries_string,
+            }
+        elif isinstance(symbol, pybamm.FunctionParameter):
+            input_names = symbol.input_names
+            inputs = {
+                input_names[i]: Serialise.convert_symbol_to_json(symbol.orphans[i])
+                for i in range(len(input_names))
+            }
+            diff_variable = symbol.diff_variable
+            if diff_variable is not None:
+                diff_variable = Serialise.convert_symbol_to_json(diff_variable)
+            return {
+                "type": symbol.__class__.__name__,
+                "inputs": inputs,
+                "diff_variable": diff_variable,
+                "name": symbol.name,
+            }
+        elif isinstance(symbol, pybamm.Variable):
+            return {
+                "type": "Variable",
+                "name": symbol.name,
+                "domain": symbol.domain,
+                "bounds": symbol.bounds,
+            }
+        elif isinstance(symbol, pybamm.SpatialVariable):
+            return {
+                "type": "SpatialVariable",
+                "name": symbol.name,
+                "domain": symbol.domain,
+                "coord_sys": symbol.coord_sys,
+            }
+
+        else:
+            raise ValueError(
+                f"Error processing '{symbol.name}'. Unknown symbol type: {type(symbol)}"
+            )
+
+    @staticmethod
+    def convert_function_to_symbolic_expression(func, name=None):
+        """
+        Converts a Python function to a PyBaMM symbolic expression
+
+        Parameters
+        ----------
+        func : callable
+            The Python function to convert
+
+        name : str, optional
+            The name of the function to use in the symbolic expression. If not provided,
+            the name of the function is used.
+
+        Returns
+        -------
+        pybamm.Symbol
+            The PyBaMM symbolic expression
+        """
+        # First, determine the number of inputs the function takes
+        num_inputs = len(inspect.signature(func).parameters)
+
+        # Create symbolic parameters for each input argument
+        # For myfun(x, y), this creates ["myfun.input_0", "myfun.input_1"]
+        func_name = name or func.__name__
+        sym_inputs = [
+            pybamm.Parameter(f"{func_name}.input_{i}") for i in range(num_inputs)
+        ]
+
+        # Evaluate the function with symbolic inputs to get symbolic expression
+        sym_output = func(*sym_inputs)
+
+        return sym_output
+
+    @staticmethod
+    def convert_symbol_from_json(json_data):
+        """
+        Recursively converts a JSON dictionary back into PyBaMM symbolic expressions
+
+        Parameters
+        ----------
+        json_data : dict
+            Dictionary containing the serialized PyBaMM expression
+
+        Returns
+        -------
+        pybamm.Symbol
+            The reconstructed PyBaMM symbolic expression
+        """
+        if isinstance(json_data, numbers.Number | list):
+            return json_data
+        elif json_data["type"] == "Parameter":
+            # Convert stored parameters back to PyBaMM Parameter objects
+            return pybamm.Parameter(json_data["name"])
+        elif json_data["type"] == "Scalar":
+            # Convert stored numerical values back to PyBaMM Scalar objects
+            return pybamm.Scalar(json_data["value"])
+        elif json_data["type"] == "Interpolant":
+            return pybamm.Interpolant(
+                [np.array(x) for x in json_data["x"]],
+                np.array(json_data["y"]),
+                [Serialise.convert_symbol_from_json(c) for c in json_data["children"]],
+                name=json_data["name"],
+                interpolator=json_data["interpolator"],
+                entries_string=json_data["entries_string"],
+            )
+        elif json_data["type"] == "FunctionParameter":
+            diff_variable = json_data["diff_variable"]
+            if diff_variable is not None:
+                diff_variable = Serialise.convert_symbol_from_json(diff_variable)
+            return pybamm.FunctionParameter(
+                json_data["name"],
+                {
+                    k: Serialise.convert_symbol_from_json(v)
+                    for k, v in json_data["inputs"].items()
+                },
+                diff_variable=diff_variable,
+            )
+        elif json_data["type"] in [
+            "Variable",
+            "StateVector",
+            "InputParameter",
+            "BoundaryValue",
+        ]:
+            return getattr(pybamm, json_data["type"])(json_data["name"])
+        else:
+            children = json_data.get("children", [])
+            return getattr(pybamm, json_data["type"])(
+                *[Serialise.convert_symbol_from_json(c) for c in children]
+            )
--- a/src/pybamm/models/base_model.py
+++ b/src/pybamm/models/base_model.py
@@ -1524,22 +1524,6 @@ def load_model(filename, battery_model: BaseModel | None = None):
     return Serialise().load_model(filename, battery_model)
 
 
-def load_custom_model(filename, battery_model: BaseModel | None = None):
-    """
-    Load in a saved model from a JSON file
-
-    Parameters
-    ----------
-    filename: str
-        Path to the JSON file containing the serialised model file
-    battery_model: :class: pybamm.BaseBatteryModel, optional
-            PyBaMM model to be created (e.g. pybamm.lithium_ion.SPM), which will
-            override any model names within the file. If None, the function will look
-            for the saved object path, present if the original model came from PyBaMM.
-    """
-    return Serialise().load_custom_model(filename, battery_model)
-
-
 # helper functions for finding symbols
 def find_symbol_in_tree(tree, name):
     if name == tree.name:
--- a/src/pybamm/models/base_model.py
+++ b/src/pybamm/models/base_model.py
@@ -1524,6 +1524,22 @@ def load_model(filename, battery_model: BaseModel | None = None):
     return Serialise().load_model(filename, battery_model)
 
 
+def load_custom_model(filename, battery_model: BaseModel | None = None):
+    """
+    Load in a saved model from a JSON file
+
+    Parameters
+    ----------
+    filename: str
+        Path to the JSON file containing the serialised model file
+    battery_model: :class: pybamm.BaseBatteryModel, optional
+            PyBaMM model to be created (e.g. pybamm.lithium_ion.SPM), which will
+            override any model names within the file. If None, the function will look
+            for the saved object path, present if the original model came from PyBaMM.
+    """
+    return Serialise().load_custom_model(filename, battery_model)
+
+
 # helper functions for finding symbols
 def find_symbol_in_tree(tree, name):
     if name == tree.name:
--- a/src/pybamm/simulation.py
+++ b/src/pybamm/simulation.py
@@ -1237,6 +1237,34 @@ class Simulation:
                 """
             )
 
+    def save_custom_model(
+        self,
+        filename: str | None = None,
+        mesh: bool = False,
+        variables: bool = False,
+    ):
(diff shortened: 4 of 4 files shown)


agriyakhetarpal:
Oh, okay, that makes sense – thanks, @Saransh-cpp!

medha-14:
I have added tests for the PR, the serialisation and de-serialisation of function parameters still needs to be fixed but apart from that it is working.

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -237,6 +239,149 @@ def load_model(
...
+            return obj.name
+        raise TypeError(
+            f"Object of type {obj.__class__.__name__} is not JSON serializable"
+        )
+
+    @staticmethod
+    def save_custom_model(model, filename=None):
+        """
+        Save the custom PyBaMM model and parameters to a JSON file.
+        """
```
santacodes: Could you elaborate more about this staticmethod in this docstring, maybe even add an example since it is user-facing and it would be easier to document this?

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -237,6 +239,149 @@ def load_model(
...
+                    Serialise.convert_symbol_to_json(v),
+                ]
+                for k, v in model.algebraic.items()
+            ],
+            "initial_conditions": [
+                (
+                    Serialise.convert_symbol_to_json(k),
+                    Serialise.convert_symbol_to_json(v),
+                )
+                for k, v in getattr(model, "initial_conditions", {}).items()
```
santacodes: It would be better to change the variable names for `k` and `v` that iterates over the `getattr()` result to something that gives semantic clarity.

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -237,6 +239,149 @@ def load_model(
...
+
+        model = battery_model if battery_model is not None else pybamm.BaseModel()
+        model.name = model_data["name"]
+
+        all_keys = (
+            [k for k, _ in model_data["rhs"]]
+            + [k for k, _ in model_data["initial_conditions"]]
+            + [k for k, _ in model_data["algebraic"]]
+            + [var for var, _ in model_data["boundary_conditions"]]
+        )
```
santacodes: Extension of the above comment

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -237,6 +239,149 @@ def load_model(
...
+            symbol_map[
+                str(Serialise.convert_symbol_from_json(k))
+            ]: Serialise.convert_symbol_from_json(v)
+            for k, v in model_data["algebraic"]
+        }
+        model.initial_conditions = {
+            symbol_map[
+                str(Serialise.convert_symbol_from_json(k))
+            ]: Serialise.convert_symbol_from_json(v)
+            for k, v in model_data["initial_conditions"]
```
santacodes: Ditto

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -399,3 +544,272 @@ def _convert_options(self, d):
...
+        Parameters
+        ----------
+        symbol : pybamm.Symbol
+            The PyBaMM symbolic expression to convert
+
+        Returns
+        -------
+        dict
+            The JSON-serializable dictionary
+        """
```
santacodes: This static method is widely used throughout the implementation, and I would encourage you to elaborate more about it in this docstring and add an example.

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -399,3 +544,272 @@ def _convert_options(self, d):
...
+        Parameters
+        ----------
+        json_data : dict
+            Dictionary containing the serialized PyBaMM expression
+
+        Returns
+        -------
+        pybamm.Symbol
+            The reconstructed PyBaMM symbolic expression
+        """
```
santacodes: Same as the above comment

tests/unit/test_serialisation/test_custom_models.py
```diff
@@ -0,0 +1,226 @@
...
+    assert isinstance(loaded_model.rhs, dict)
+    assert next(iter(loaded_model.rhs.keys())).name == "a"
+    assert next(iter(loaded_model.rhs.values())).name == "b"
+
+
+def test_plotting_serialised_models():
+    models = [BasicSPM(), BasicDFN()]
+    filenames = ["spm", "dfn"]
+
+    for model, name in zip(models, filenames, strict=False):
```
Saransh-cpp: Given that we know that there are 2 models and 2 filenames
```suggestion
    for model, name in zip(models, filenames, strict=True):
```

tests/unit/test_serialisation/test_custom_models.py
```diff
@@ -0,0 +1,226 @@
...
+    # save model
+    Serialise.save_custom_model(model, filename="test_model")
+
+    # check json exists
+    assert os.path.exists("test_model.json")
+
+    # saving with defualt filename
+    Serialise().save_custom_model(model)
+    filename = "test_model_" + datetime.now().strftime("%Y_%m_%d-%p%I_%M") + ".json"
+    assert os.path.exists(filename)
```
Saransh-cpp: This test looks flaky. I guess it would be possible for the function to save a model at say `12:23:36` and the `filename` variable to be initialised at `12:23:37`. Could you instead check using regex? The filename should match the characters followed be `%Y_%m_%d-%p%I_%M` pattern.

tests/unit/test_serialisation/test_custom_models.py
Saransh-cpp, line 1: All of these tests should technically go into -

https://github.com/pybamm-team/PyBaMM/blob/1a39a994298a90fbcdd4d5dfea5958428740f150/tests/unit/test_serialisation/test_serialisation.py#L89

src/pybamm/__init__.py
```diff
@@ -57,7 +57,7 @@
 from .expression_tree.operations.jacobian import Jacobian
 from .expression_tree.operations.convert_to_casadi import CasadiConverter
 from .expression_tree.operations.unpack_symbols import SymbolUnpacker
-
+from .expression_tree.operations.serialise import Serialise
```
Saransh-cpp, line 61:
```suggestion
from .expression_tree.operations.serialise import Serialise

```

src/pybamm/__init__.py
```diff
@@ -72,7 +72,6 @@
 from .models.full_battery_models import lithium_ion
 from .models.full_battery_models import equivalent_circuit
 from .models.full_battery_models import sodium_ion
```
Saransh-cpp:
```suggestion
from .models.full_battery_models import sodium_ion

```

src/pybamm/__init__.py
```diff
@@ -150,7 +149,6 @@
 
 # Serialisation
 from .models.base_model import load_model
```
Saransh-cpp:
```suggestion
from .models.base_model import load_model

```

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -237,6 +239,205 @@ def load_model(
             """
         )
 
+    def _json_encoder(obj):
+        if isinstance(obj, pybamm.Symbol):  # or SymbolBase
```
Saransh-cpp: What is `SymbolBase` here?

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -237,6 +239,205 @@ def load_model(
...
+        filename : str, optional
+            The desired name of the JSON file. If not provided, a name will be
+            generated from the model name and current datetime.
+
+        Example
+        -------
+        >>> MyModel = MyCustomModelClass()
+        >>> MyModel.build_model()
+        >>> Serialise.save_custom_model(MyModel, "my_custom_model")
+        # Saves 'my_custom_model.json' in the current directory
```
Saransh-cpp: The doctests must be entirely runnable (in a sense that if a user copies these exactly as it is, they should run). This is why the doctests are failing.

So here, you'll have to define what `MyCustomModelClass` is and so on (better to use an existing PyBaMM class and leave a comment saying it generalises to custom classes as well).

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -237,6 +239,205 @@ def load_model(
...
+        -------
+        :class:`pybamm.BaseModel` or subclass
+            The reconstructed symbolic PyBaMM model.
+
+        Example
+        -------
+        >>> loaded_model= Serialise.load_custom_model("my_model.json", battery_model=pybamm.lithium_ion.BaseModel())
+        >>> sim = pybamm.Simulation(loaded_model)
+        >>> sim.solve([0, 3600])
+        >>> sim.plot()
```
Saransh-cpp: Ditto! No need to plot in doctests.

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -399,3 +601,296 @@ def _convert_options(self, d):
...
+        dict
+            A JSON-compatible representation of the input.
+
+        Examples
+        --------
+        >>> Serialise.convert_symbol_to_json(pybamm.Scalar(5))
+        {'type': 'Scalar', 'value': 5}
+
+        >>> Serialise.convert_symbol_to_json(pybamm.Variable("c"))
+        {'type': 'Variable', 'name': 'c', 'domains': {}, 'bounds': [None, None]}
```
Saransh-cpp: Ditto

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -399,3 +601,296 @@ def _convert_options(self, d):
...
+        Returns
+        -------
+        pybamm.Symbol or primitive
+            The reconstructed PyBaMM symbolic expression or a primitive (float, int, bool).
+
+        Examples
+        --------
+        >>> json_expr = {'type': 'Scalar', 'value': 42}
+        >>> Serialise.convert_symbol_from_json(json_expr)
+        pybamm.Scalar(42)
```
Saransh-cpp: Ditto. The output style is right, and it should be present in the other doctests as well.

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -237,6 +239,208 @@ def load_model(
...
+
+        Example
+        -------
+        >>> import pybamm
+        >>> model = pybamm.lithium_ion.BasicDFN()
+        >>> from pybamm.expression_tree.operations.serialise import Serialise
+        >>> Serialise.save_custom_model(model, "basicdfn_model")
+        # Saves 'basicdfn_model.json' in the current directory
+
+        # Note: 'save_custom_model' also works for user-defined models.
```
Saransh-cpp:
```suggestion
```

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -237,6 +239,208 @@ def load_model(
...
+        :class:`pybamm.BaseModel` or subclass
+            The reconstructed symbolic PyBaMM model.
+
+        Example
+        -------
+        >>> import pybamm
+        >>> model = pybamm.lithium_ion.BasicDFN()
+        >>> from pybamm.expression_tree.operations.serialise import Serialise
+        >>> Serialise.save_custom_model(model, "basicdfn_model")
+        >>> loaded_model= Serialise.load_custom_model("basicdfn_model.json", battery_model=pybamm.lithium_ion.BaseModel())
```
Saransh-cpp:
```suggestion
        >>> loaded_model = Serialise.load_custom_model("basicdfn_model.json", battery_model=pybamm.lithium_ion.BaseModel())
```

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -399,3 +604,315 @@ def _convert_options(self, d):
...
+        dict
+            A JSON-compatible representation of the input.
+
+        Examples
+        --------
+        >>> import pybamm
+        >>> from pybamm.expression_tree.operations.serialise import Serialise
+        >>> s = pybamm.Scalar(5)
+        >>> Serialise.convert_symbol_to_json(s)
+        {'type': 'Scalar', 'value': 5}
```
Saransh-cpp: Is this 5.0 (float value) expected?
```suggestion
        {'type': 'Scalar', 'value': 5.0}
```

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -399,3 +604,315 @@ def _convert_options(self, d):
...
+        pybamm.Symbol or primitive
+            The reconstructed PyBaMM symbolic expression or a primitive (float, int, bool).
+
+        Examples
+        --------
+        >>> import pybamm
+        >>> from pybamm.expression_tree.operations.serialise import Serialise
+        >>> json_expr = {'type': 'Scalar', 'value': 42}
+        >>> Serialise.convert_symbol_from_json(json_expr)
+        pybamm.Scalar(42)
```
Saransh-cpp: Should probably ignore this as a doctest because of the memory address
```suggestion
        Scalar(0x1aa1e910749e9f89, 42.0, children=[], domains={})
```

Saransh-cpp (changes requested):
You can run the doctests locally using -
```bash
nox -s doctests
```
😉

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -237,6 +239,208 @@ def load_model(
             """
         )
 
+    def _json_encoder(obj):
```
agriyakhetarpal, line 247:
```suggestion
    @staticmethod
    def _json_encoder(obj):
```

What do you think?

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -237,6 +239,208 @@ def load_model(
...
+        if isinstance(obj, Enum):
+            return obj.name
+        elif isinstance(obj, np.ndarray):
+            return obj.tolist()
+        elif isinstance(obj, (np.float64 | np.float32)):
+            return float(obj)
+        elif isinstance(obj, (np.int64 | np.int32)):
+            return int(obj)
+        else:
+            raise TypeError(f"Object of type {type(obj)} is not JSON serializable.")
```
agriyakhetarpal: I think these are too specific. How about the following base types?

```suggestion
        if isinstance(obj, Enum):
            return obj.name
        elif isinstance(obj, np.ndarray):
            return obj.tolist()
        elif isinstance(obj, np.floating):
            return float(obj)
        elif isinstance(obj, np.integer):
            return int(obj)
        else:
            raise TypeError(f"Object of type {type(obj)} is not JSON serializable.")
```

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -237,6 +239,208 @@ def load_model(
...
+                for event in getattr(model, "events", [])
+            ],
+            "variables": {
+                str(variable_name): Serialise.convert_symbol_to_json(expression)
+                for variable_name, expression in getattr(model, "variables", {}).items()
+            },
+        }
+
+        if filename is None:
+            filename = model.name + "_" + datetime.now().strftime("%Y_%m_%d-%p%I_%M")
```
agriyakhetarpal: I think you should also sanitise the filename to remove invalid characters, which could prevent the file from being saved. This can be handled with a regular expression.

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -237,6 +239,208 @@ def load_model(
...
+        >>> from pybamm.expression_tree.operations.serialise import Serialise
+        >>> Serialise.save_custom_model(model, "basicdfn_model")
+        # Saves 'basicdfn_model.json' in the current directory
+
+        # Note: 'save_custom_model' also works for user-defined models.
+
+        """
+        model_json = {
+            "pybamm_version": pybamm.__version__,
+            "name": getattr(model, "name", "unnamed_model"),
```
agriyakhetarpal: Another thing to include here, as a forward-thinking architecture decision: decoupled from the PyBaMM version, please add a `schema_version` – so that if/when we change the JSON structure later on, we can increment the version here (for reproducibility too).

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -399,3 +604,315 @@ def _convert_options(self, d):
...
+            )
+        return json_dict
+
+    @staticmethod
+    def convert_symbol_from_json(json_data):
+        """
+        Recursively reconstructs a PyBaMM symbolic expression from a JSON dictionary.
+
+        Supports all major PyBaMM symbol types, including Scalars, Variables, Parameters,
+        Operators, FunctionParameters, Broadcasts, Interpolants.
```
agriyakhetarpal: You can refer to these like `pybamm.Scalar`, `pybamm.Variable`, etc. in the reST format, so that they will be linked in the documentation :) You may use the RTD PR previews to test that out.

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -399,3 +604,315 @@ def _convert_options(self, d):
...
+                    "domains",
+                    {
+                        "primary": json_data.get("domain", []),
+                        "secondary": [],
+                        "tertiary": [],
+                        "quaternary": [],
+                    },
+                ),
+                bounds=bounds,
+            )
```
agriyakhetarpal: So I think there is some inconsistency here. In this code path, you handle this domain format for primary up to quaternary domains. But in the serialisation path, you always write `"domains": {...}` (see the `"domains": symbol.domains` line). Thus, is the following line unnecessary?

```suggestion
        elif symbol_type == "Variable":
            bounds = tuple(
                Serialise.convert_symbol_from_json(b)
                for b in json_data.get("bounds", [-float("inf"), float("inf")])
            )
            return pybamm.Variable(
                json_data["name"],
                domains=json_data["domains"]
                bounds=bounds,
            )
```

because you never get this as a product of the serialisation? You already do this for the `SpatialVariable` case just below. Please feel free to correct me if I'm wrong!

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -237,6 +239,208 @@ def load_model(
...
+                    Serialise.convert_symbol_to_json(rhs_expression),
+                )
+                for variable, rhs_expression in getattr(model, "rhs", {}).items()
+            ],
+            "algebraic": [
+                [
+                    Serialise.convert_symbol_to_json(variable),
+                    Serialise.convert_symbol_to_json(algebraic_expression),
+                ]
+                for variable, algebraic_expression in model.algebraic.items()
```
agriyakhetarpal: What is the reason to use a `list` here instead of a `tuple` (like you do for `rhs` above)?

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -237,6 +239,208 @@ def load_model(
...
+            [lhs_json for lhs_json, _ in model_data["rhs"]]
+            + [lhs_json for lhs_json, _ in model_data["initial_conditions"]]
+            + [lhs_json for lhs_json, _ in model_data["algebraic"]]
+            + [variable_json for variable_json, _ in model_data["boundary_conditions"]]
+        )
+
+        symbol_map = {}
+        for variable_json in all_variable_keys:
+            symbol = Serialise.convert_symbol_from_json(variable_json)
+            symbol_map[str(symbol)] = symbol
```
agriyakhetarpal: Won't this cause collisions if two different symbols have the same string representation? For example,

```python
var1 = pybamm.Variable("x", bounds=(0, 1))
var2 = pybamm.Variable("x", bounds=(0, 2))
```

will have the same string representation. It will also propagate to complex expressions:

```python
expr1 = var1 + pybamm.Scalar(1)
expr2 = var2 + pybamm.Scalar(1)
```

Here, `expr1` and `expr2` will have the same string representation, too. This means that you could get the wrong variable when you reconstruct the model from the symbol map (below these lines).

I would suggest creating a `_create_symbol_key` helper that represents the `domains`, `bounds`, and `coord_sys` (for `SpatialVariable`s), etc. in the key in a dict, and then you can do something like

```python
model.rhs = {
    symbol_map[_create_symbol_key(lhs_json)]: 
    # ...
    }
```

afterwards.

agriyakhetarpal (changes requested):
Going great, @medha-14! Thanks to @santacodes and @Saransh-cpp for the existing reviews; please resolve their comments first. Could you also rename the PR title to more accurately reflect its added functionality?

src/pybamm/expression_tree/operations/serialise.py
medha-14: Yes, that's how they’re stored internally. I checked by comparing, and pybamm.Scalar(5) prints as 5.0.

src/pybamm/expression_tree/operations/serialise.py
medha-14: I was initially using lists throughout, but I missed updating this particular one when I switched to tuples for consistency. Thanks for pointing it out!

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -237,6 +241,228 @@ def load_model(
...
+            "variables": {
+                str(variable_name): Serialise.convert_symbol_to_json(expression)
+                for variable_name, expression in getattr(model, "variables", {}).items()
+            },
+        }
+
+        if filename is None:
+            filename = model.name + "_" + datetime.now().strftime("%Y_%m_%d-%p%I_%M")
+
+        filename = re.sub(r'[<>:"/\\|?*\x00-\x1F]', "", filename)
```
agriyakhetarpal: Thanks for improving this. Two things here:
- we should use the 24-hour format instead of the 12-hour format (easier to read and more _computer_-ish)
- the regex removes invalid characters, but it doesn't replace them with safe ones.

I have a suggestion as follows. Please feel free to apply it or iterate on it further:

```suggestion
        if filename is None:
            safe_name = re.sub(r'[^\w\-_.]', '_', model.name or 'unnamed_model')
            timestamp = datetime.now().strftime("%Y_%m_%d_%H_%M_%S")
            filename = f"{safe_name}_{timestamp}"
            
        else:
            # Clean provided filename
            filename = re.sub(r'[<>:"/\\|?*\x00-\x1F]', '_', filename)
```

Then, please also update the test case to match this. Thanks!

tests/unit/test_serialisation/test_serialisation.py
```diff
@@ -591,3 +594,339 @@ def test_serialised_model_plotting(self):
...
+        Serialise.save_custom_model(model, filename="test_model")
+
+        # check json exists
+        assert os.path.exists("test_model.json")
+
+        # saving with defualt filename
+        Serialise().save_custom_model(model)
+        pattern = r"test_model_\d{4}_\d{2}_\d{2}-(AM|PM)\d{2}_\d{2}\.json"
+        matched = [f for f in os.listdir(".") if re.fullmatch(pattern, f)]
+        assert matched
```
agriyakhetarpal: I think this should be:

```suggestion
        # saving with default filename
        Serialise().save_custom_model(model)
        pattern = r"test_model_\d{4}_\d{2}_\d{2}_\d{2}_\d{2}_\d{2}\.json"
        matched = [f for f in os.listdir(".") if re.fullmatch(pattern, f)]
        assert matched
```

tests/unit/test_serialisation/test_serialisation.py
```diff
@@ -591,3 +594,339 @@ def test_serialised_model_plotting(self):
...
+
+        json1 = Serialise.convert_symbol_to_json(var1)
+        json2 = Serialise.convert_symbol_to_json(var2)
+
+        key1 = Serialise._create_symbol_key(json1)
+        key2 = Serialise._create_symbol_key(json2)
+
+        assert isinstance(key1, str)
+        assert isinstance(key2, str)
+        assert key1 != key2
```
agriyakhetarpal, line 681: Superb, thanks for adding a test for this as well. Could you add another test, such as something like `test_variable_domain_collision_prevention` (or a better name of your liking), which initialises two variables with different domains, and similarly test that the symbol keys they create are unique? Just covering the gaps in testing.

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -237,6 +241,228 @@ def load_model(
...
+        >>> import pybamm
+        >>> model = pybamm.lithium_ion.BasicDFN()
+        >>> from pybamm.expression_tree.operations.serialise import Serialise
+        >>> Serialise.save_custom_model(model, "basicdfn_model")
+
+        """
+        SCHEMA_VERSION = "1.0"
+        model_json = {
+            "pybamm_version": pybamm.__version__,
+            "schema_version": SCHEMA_VERSION,
```
agriyakhetarpal: Nice, thanks for adding this. Could you add two tests for this? (Not sure why this wasn't caught in our coverage)
- a test that loads a JSON file with an unsupported schema version and checks that a `ValueError` is raised
- a test that checks that that the saved JSON files do contain the schema version

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -237,6 +241,228 @@ def load_model(
...
+        """
+        with open(filename) as file:
+            model_data = json.load(file)
+
+        schema_version = model_data.get("schema_version", SUPPORTED_SCHEMA_VERSION)
+        if schema_version != SUPPORTED_SCHEMA_VERSION:
+            raise ValueError(
+                f"Unsupported schema version: {schema_version}. "
+                f"Expected: {SUPPORTED_SCHEMA_VERSION}"
+            )
```
agriyakhetarpal: i.e., tests for this code path

agriyakhetarpal (commented):
Great work, @medha-14! This is close, and thanks for iterating on my feedback. I will also request @valentinsulzer and co. to review to cover that I may have missed during my review.

tests/unit/test_serialisation/test_serialisation.py
medha-14: Already done something like this :)

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -237,6 +241,233 @@ def load_model(
...
+        Example
+        -------
+        >>> import pybamm
+        >>> model = pybamm.lithium_ion.BasicDFN()
+        >>> from pybamm.expression_tree.operations.serialise import Serialise
+        >>> Serialise.save_custom_model(model, "basicdfn_model")
+
+        """
+        SCHEMA_VERSION = "1.0"
+        model_json = {
```
Saransh-cpp: Given that models with `FunctionParameters` cannot be serialised at the moment, it would be good to add an error here if the model contains one. Something like -

```py
if <check if the model has a function parameter>:
    raise NotImplementedError("serialising models with function parameters is not implemented")
```

agriyakhetarpal: Additionally, it would be beneficial to include an inline comment in the code (or better, a comment in this PR, with an inline comment linking to it) with the technical details, so that we can recall why this feature is not implemented.

agriyakhetarpal: Though, seeing that the serialisation of `FunctionParameter`s is working as highlighted in the code and associated tests, is there still a problem in the serialisation of models that contain them?

medha-14: Yes, models with FunctionParameter are serialising correctly in the current implementation.
However, the issue arises when you attempt to Serialise a ParameterValues object that includes an actual Python callable as a value for a parameter (rather than a erialisable symbolic expression or a float).

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -237,6 +241,312 @@ def load_model(
...
+            }
+
+            if filename is None:
+                safe_name = re.sub(r"[^\w\-_.]", "_", model.name or "unnamed_model")
+                timestamp = datetime.now().strftime("%Y_%m_%d_%H_%M_%S")
+                filename = f"{safe_name}_{timestamp}.json"
+            else:
+                # Clean provided filename and ensure .json
+                base = re.sub(r'[<>:"/\\|?*\x00-\x1F]', "_", filename)
+                filename = base if base.endswith(".json") else f"{base}.json"
```
agriyakhetarpal: We should be explicit and ask users to add `.json` to the filename when they save, rather than handling it on our own.

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -399,3 +710,299 @@ def _convert_options(self, d):
...
         else:
             return d
+
+    @staticmethod
+    def convert_symbol_to_json(symbol):
+        """
+        Recursively converts a PyBaMM symbolic expression into a JSON-serializable format.
+
+        Supports most PyBaMM symbol types, including scalars, variables, parameters,
+        operators, broadcasts, and interpolants.
```
agriyakhetarpal:
```suggestion
        Supports all major PyBaMM symbol types, including :class:`pybamm.Scalar`, 
        :class:`pybamm.Variable`, :class:`pybamm.Parameter`, :class:`pybamm.Operator`,
        :class:`pybamm.FunctionParameter`, :class:`pybamm.Broadcast`, and
        :class:`pybamm.Interpolant`.
```

(from the `convert_symbol_from_json` docstring)

tests/unit/test_serialisation/test_serialisation.py
```diff
@@ -591,3 +598,749 @@ def test_serialised_model_plotting(self):
...
+        loaded_model = Serialise.load_custom_model(
+            f"{filename}.json", battery_model=pybamm.lithium_ion.BaseModel()
+        )
+
+        try:
+            assert hasattr(loaded_model, "schema_version")
+            assert loaded_model.schema_version == SUPPORTED_SCHEMA_VERSION
+        finally:
+            # Clean up
+            os.remove(f"{filename}.json")
```
agriyakhetarpal: The file save and cleanup can be better handled with `pytest` and its fixtures – https://docs.pytest.org/en/stable/how-to/tmp_path.html#how-to-use-temporary-directories-and-files-in-tests (and similarly for other tests too)

tests/unit/test_serialisation/test_serialisation.py
```diff
@@ -591,3 +598,749 @@ def test_serialised_model_plotting(self):
...
+            "events": [],
+            "variables": {},
+        }
+        file = "model2.json"
+
+        with open(file, "w") as f:
+            json.dump(model_json, f)
+
+        with pytest.raises(ValueError) as e:
+            Serialise.load_custom_model(str(file))
```
agriyakhetarpal: Could you please also add a `match` argument to `pytest.raises()` everywhere so that it is easier to see the error message associated with this `ValueError`?

santacodes (approved):

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -237,6 +242,329 @@ def load_model(
...
+            return float(obj)
+        elif isinstance(obj, np.integer):
+            return int(obj)
+        else:
+            raise TypeError(f"Object of type {type(obj)} is not JSON serializable.")
+
+    @staticmethod
+    def save_custom_model(model, filename=None):
+        """
+        Saves a custom (non-discretised) PyBaMM model to a JSON file. Works for user defined models that are subclasses of BaseModel.
```
agriyakhetarpal, line 264:
```suggestion
        Saves a custom (non-discretised) PyBaMM model to a JSON file. Works for user-defined models that subclass :class:`pybamm.BaseModel`.
```

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -237,6 +242,329 @@ def load_model(
...
+                ],
+                "variables": {
+                    str(variable_name): Serialise.convert_symbol_to_json(expression)
+                    for variable_name, expression in getattr(
+                        model, "variables", {}
+                    ).items()
+                },
+            }
+
+            # Final JSON structure with schema and version info at top level
```
agriyakhetarpal:
```suggestion
```

This comment looks like it is redundant, as the JSON structure is easy to read already. Could we remove it?

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -237,6 +242,329 @@ def load_model(
...
+        algebraic equations, initial and boundary conditions, events, and variables.
+        Returns a fully symbolic model ready for further processing or discretisation.
+
+        Parameters
+        ----------
+        filename : str
+            Path to the JSON file containing the saved model.
+        battery_model : :class:`pybamm.BaseModel`, optional
+            An optional existing model instance to populate. If not provided, a new
+            :class:`pybamm.BaseModel` is created.
```
agriyakhetarpal: We previously discussed that we wouldn't want the `battery_model` to be an argument for `load_custom_model` as it would be redundant, right? i.e., users should do `loaded_model = Serialise.load_custom_model("basicdfn_model.json")` and it would already be inferred from their code that they're working with a custom model that derives from the `BasicDFN` class.

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -237,6 +242,329 @@ def load_model(
...
+        >>> loaded_model = Serialise.load_custom_model("basicdfn_model.json", battery_model=pybamm.lithium_ion.BaseModel())
+
+        """
+        try:
+            with open(filename) as file:
+                data = json.load(file)
+        except FileNotFoundError as err:
+            raise FileNotFoundError(f"Could not find file: {filename}") from err
+        except json.JSONDecodeError as e:
+            raise ValueError(f"Invalid JSON in file '{filename}': {e!s}") from e
```
agriyakhetarpal: I think this can benefit from a more specific error rather than a `ValueError`, such as `InvalidModelJSONError` (or a better phrase if you have one). The error message could reflect something like `f"The model defined in the file {filename} contains invalid JSON"`. You may define custom exceptions here: https://github.com/pybamm-team/PyBaMM/blob/cd9b533b7ac99b9be9281a5390bb5954b3e48d30/src/pybamm/expression_tree/exceptions.py.

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -360,6 +688,7 @@ def _reconstruct_pybamm_dict(self, obj: dict):
         {"rod":
             {SpatialVariable(name="spat_var"): {"min":0.0, "max":2.0} }
             }
+
```
agriyakhetarpal, line 720:
```suggestion
```

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -237,6 +242,329 @@ def load_model(
...
+    @staticmethod
+    def _create_symbol_key(symbol_json: dict) -> str:
+        """
+        Given the JSON‐dict for a symbol, return a unique, hashable key.
+        We just sort the dict keys and dump to a string.
+        """
+        return json.dumps(symbol_json, sort_keys=True)
+
+    @staticmethod
+    def load_custom_model(filename, battery_model=None):
```
agriyakhetarpal: Could you add type annotations here?

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -237,6 +242,329 @@ def load_model(
...
+            return obj.tolist()
+        elif isinstance(obj, np.floating):
+            return float(obj)
+        elif isinstance(obj, np.integer):
+            return int(obj)
+        else:
+            raise TypeError(f"Object of type {type(obj)} is not JSON serializable.")
+
+    @staticmethod
+    def save_custom_model(model, filename=None):
```
agriyakhetarpal: Could you add type annotations here?

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -399,3 +728,301 @@ def _convert_options(self, d):
             return tuple(self._convert_options(item) for item in d)
         else:
             return d
+
+    @staticmethod
+    def convert_symbol_to_json(symbol):
```
agriyakhetarpal: Could we please have better typing here?

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -399,3 +728,301 @@ def _convert_options(self, d):
...
+                json_dict["name"] = symbol.name
+
+        else:
+            raise ValueError(
+                f"Error processing '{symbol.name}'. Unknown symbol type: {type(symbol)}"
+            )
+        return json_dict
+
+    @staticmethod
+    def convert_symbol_from_json(json_data):
```
agriyakhetarpal: Same as above; this can benefit from better typing. :)

tests/unit/test_serialisation/test_serialisation.py
```diff
@@ -591,3 +598,739 @@ def test_serialised_model_plotting(self):
...
+        assert symbol2.broadcast_domain == ["negative electrode"]
+        assert isinstance(symbol2.orphans[0], pybamm.Scalar)
+        assert symbol2.orphans[0].value == 42
+
+    def test_interpolant_serialisation(self):
+        x = np.linspace(0, 1, 5)
+        y = np.array([0, 1, 4, 9, 16])
+        child = pybamm.Variable("z")
+        interp = pybamm.Interpolant(
+            x, y, child, name="test_interplot", interpolator="linear"
```
agriyakhetarpal:
```suggestion
            x, y, child, name="test_interpolant", interpolator="linear"
```

Perhaps this was a typo? 😄

tests/unit/test_serialisation/test_serialisation.py
```diff
@@ -591,3 +598,739 @@ def test_serialised_model_plotting(self):
...
+        file_path = tmp_path / "model.json"
+
+        # Write JSON to the temporary file
+        with open(file_path, "w") as f:
+            json.dump(unhandled_schema_json, f)
+
+        # Assert that loading this model raises a ValueError
+        with pytest.raises(ValueError, match="Unsupported schema version: 9.9"):
+            Serialise.load_custom_model(file_path, battery_model=pybamm.BaseModel())
+
```
agriyakhetarpal:
```suggestion
        with open(file_path, "w") as f:
            json.dump(unhandled_schema_json, f)

        with pytest.raises(ValueError, match="Unsupported schema version: 9.9"):
            Serialise.load_custom_model(file_path, battery_model=pybamm.BaseModel())

```

tests/unit/test_serialisation/test_serialisation.py
```diff
@@ -591,3 +598,739 @@ def test_serialised_model_plotting(self):
...
+            Serialise.save_custom_model(model, filename=str(path))
+
+            # Load the model
+            loaded_model = Serialise.load_custom_model(
+                str(path), battery_model=pybamm.lithium_ion.BaseModel()
+            )
+
+            sim = pybamm.Simulation(loaded_model)
+            sim.solve([0, 3600])
+            sim.plot(show_plot=False)
```
agriyakhetarpal: You can parameterise this test with `pytest` instead of using a loop, like this:


```suggestion
    @pytest.mark.parametrize(
        "model, filename",
        [
            (BasicSPM(), "basic_spm.json"),
            (BasicDFN(), "basic_dfn.json"),
            (pybamm.lithium_ion.SPM(), "spm.json"),
            (pybamm.lithium_ion.DFN(), "dfn.json"),
        ],
        ids=["basic_spm", "basic_dfn", "spm", "dfn"]
    )
    def test_plotting_serialised_models(self, model, filename, tmp_path):
        path = tmp_path / filename

        Serialise.save_custom_model(model, filename=str(path))

        loaded_model = Serialise.load_custom_model(
            str(path), battery_model=pybamm.lithium_ion.BaseModel()
        )

        sim = pybamm.Simulation(loaded_model)
        sim.solve([0, 3600])
        sim.plot(show_plot=False)
```

agriyakhetarpal (commented):
Great work! I only have some nits and one functional change to request.

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -399,3 +761,303 @@ def _convert_options(self, d):
...
+        >>> Serialise.convert_symbol_to_json(s)
+        {'type': 'Scalar', 'value': 5.0}
+
+        >>> v = pybamm.Variable("c")
+        >>> Serialise.convert_symbol_to_json(v)
+        {'type': 'Variable', 'name': 'c', 'domains': {'primary': [], 'secondary': [], 'tertiary': [], 'quaternary': []}, 'bounds': [{'type': 'Scalar', 'value': -inf}, {'type': 'Scalar', 'value': inf}]}
+        """
+
+        if isinstance(symbol, numbers.Number | list):
+            return symbol
```
agriyakhetarpal, line 797: The function accepts `numbers.Number` and `list` objects as well, could you please update the type hint as well? i.e., see my other comment

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -399,3 +761,303 @@ def _convert_options(self, d):
             return tuple(self._convert_options(item) for item in d)
         else:
             return d
+
+    @staticmethod
+    def convert_symbol_to_json(symbol: pybamm.Symbol) -> dict[str, Any]:
```
agriyakhetarpal:
```suggestion
    def convert_symbol_to_json(symbol: pybamm.Symbol | numbers.Number | list) -> dict[str, Any] | numbers.Number | list:
```

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -237,6 +243,362 @@ def load_model(
...
+        if missing:
+            raise KeyError(f"Missing required model sections: {missing}")
+
+        battery_model = model_data.get("base_class")
+        if (
+            not battery_model
+            or battery_model.isspace()
+            or battery_model == "pybamm.BaseModel"
+        ):
+            base_cls = pybamm.BaseModel
```
agriyakhetarpal:
```suggestion
        if not battery_model or battery_model.strip() == "pybamm.BaseModel":
            base_cls = pybamm.BaseModel
```

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -399,3 +761,303 @@ def _convert_options(self, d):
...
+                diff_variable=diff_variable,
+            )
+        elif symbol_type == "PrimaryBroadcast":
+            child = Serialise.convert_symbol_from_json(json_data["children"][0])
+            domain = json_data["broadcast_domain"]
+            return pybamm.PrimaryBroadcast(child, domain)
+        elif symbol_type == "FullBroadcast":
+            child = Serialise.convert_symbol_from_json(json_data["children"][0])
+            domains = json_data["domains"]
+            return pybamm.FullBroadcast(child, "broadcast", domains)
```
agriyakhetarpal: `broadcast` is hardcoded here, but `pybamm.FullBroadcast` also accepts a bunch of other parameters: https://docs.pybamm.org/en/stable/source/api/expression_tree/broadcasts.html#pybamm.FullBroadcast

If you exclude the `broadcast_domain` and `broadcast_dmains` arguments, we wouldn't be able to specify which specific domain(s) the value is being broadcast to. Should we save `"broadcast_domain": symbol.broadcast_domain`, and similarly the rest of the args at the time of serialisation, and use keyword arguments here just to be safe?

medha-14: In the SPM model and maybe others
```
symbol = FullBroadcast(0x15a3e0c02229aaca, concatenation, children=['Initial concentration in electrolyte [mol.m-3]'], domains={'primary': ['negative electrode', 'separator', 'positive electrode'], 'secondary': ['current collector']})
```
`broadcast_domains` is stored as `domains`, so when saving, `broadcast_domains` is not recognized.

agriyakhetarpal: Hmm, given that the SPM round-trip is working as per the tests, let's leave this for now. I checked and we normalise this internally anyway, which might be the reason for what you're seeing: https://github.com/pybamm-team/PyBaMM/blob/123aff7998b3687c5fa9839ab8331d948d5b2f88/src/pybamm/expression_tree/broadcasts.py#L484

Still, let's change this to `pybamm.FullBroadcast(child, broadcast_domains=domains)` so that we can be explicit about what's being serialised.

Saransh-cpp (approved):
Should be ready to merge once @valentinsulzer approves it 🚀

valentinsulzer (approved):

PR #5070, opened 2025-06-20
Add constant symbol for named constants

rtimms:
# Description

Adds a new `Constant` symbol for named constants. This is a subclass of  `pybamm.Scalar` and is used to represent named constants such as the gas constant. This avoids constants being simplified out when printing expressions (but they are still simplified when evaluated). For example, `pybamm.constants.R + 2` should print `2+R` and evaluate to `10.314462618`. Counterintuitively, the "is_constant" method is set to False, so that named constants are not simplified when constructed. This is maybe a little confusing, but this is the cleanest way of avoiding simplifications without checking for instances of `Constant` everywhere. 

Fixes # (issue)

## Type of change

# Important checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Features
 
+- Added a `Constant` symbol for named constants. This is a subclass of `Scalar` and is used to represent named constants such as the gas constant. This avoids constants being simplified out when constructing expressions. ([#5070](https://github.com/pybamm-team/PyBaMM/pull/5070))
 - Generalise `pybamm.DiscreteTimeSum` to allow it to be embedded in other expressions ([#5044](https://github.com/pybamm-team/PyBaMM/pull/5044))
 - Adds `all` key-value pair to `output_variables` sensitivity dictionaries, accessible through `solution[var].sensitivities['all']`. Aligns shape with conventional solution sensitivities object. ([#5067](https://github.com/pybamm-team/PyBaMM/pull/5067))
 


rtimms:
Yes, good point, I think this is what happens since they get simplified on evaluation, but I'll add a test to confirm this behaviour

src/pybamm/expression_tree/scalar.py
```diff
@@ -112,3 +112,33 @@ def to_json(self):
...
 
         return json_dict
+
+
+class Constant(Scalar):
+    """
+    A node in the expression tree representing a named constant value. This is a subclass of
+    :class:`pybamm.Scalar` and is used to represent named constants such as :math:`R`, :math:`F`,
+    and :math:`k_b`. Counterintuitively, the "is_constant" method is set to False, so that
+    named constants are not simplified when evaluated.
```
rtimms:
```suggestion
    named constants are not simplified when constructed.
```

src/pybamm/expression_tree/scalar.py
```diff
@@ -112,3 +112,33 @@ def to_json(self):
...
+    """
+
+    def __init__(self, value: Numeric, name: str):
+        super().__init__(value, name)
+
+    def is_constant(self) -> Literal[True]:
+        """
+        The value is constant, but setting this to False makes sure we retain named constants
+        in the expression tree. For example, `pybamm.constants.R + 2` should not be
+        simplified to `10.314462618`. See :meth:`pybamm.Symbol.is_constant()`.
```
rtimms:
```suggestion
        simplified to `10.314462618` until after the expression is evaluated. See :meth:`pybamm.Symbol.is_constant()`.
```

PR #5083, opened 2025-06-27
Update numpy to 2.0

kratman:
# Description

This should allow us to use `numpy>2.0.0` with PyBaMM.

Note: As far as I know everything in here still works with `numpy<2.0.0`, however, we probably don't want to continue maintaining compatibility with older versions. When I attempted this the first time, I had to make a few code changes for some numpy versions between 2.0.0 and 2.10.0, but those issues seem to be gone now

## Type of change

# Important checks:

### DIFF
--- a/.github/workflows/run_periodic_tests.yml
+++ b/.github/workflows/run_periodic_tests.yml
@@ -30,7 +30,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-latest, macos-13, macos-14, windows-latest ]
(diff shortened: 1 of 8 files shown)


kratman:
@BradyPlanden I had to change the multiprocessing to spawn, otherwise the SciPy solver would crash. I could not reproduce the issue locally with my Mac (same OS version, same python versions). It might be something in one of the backend libraries for numerical operations. The failing test was `test_model_solver_multiple_inputs_happy_path()` if you are interested in digging further.

PR #5084, opened 2025-06-27
Update spawn context for multiprocessing

### LINKED ISSUE #3987: BaseSolver.solve() multiprocessing is broken on system's defaulting to spawn context for windows
Follow up to #3975

> It's worth noting that this only fixes https://github.com/pybamm-team/PyBaMM/issues/3974 for darwin systems and protects linux systems from the default context change in Python 3.14. Windows systems will still require the __main__ guard discussed in https://github.com/pybamm-team/PyBaMM/issues/3974.

kratman:
# Description

Fixes the spawning issue from my attempts to update to numpy 2.0.0 and switches to macos-latest

Closes #3987

## Type of change

# Important checks:

### DIFF
--- a/.github/workflows/run_periodic_tests.yml
+++ b/.github/workflows/run_periodic_tests.yml
@@ -30,7 +30,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-latest, macos-13, macos-14, windows-latest ]
+        os: [ ubuntu-latest, macos-13, macos-latest, windows-latest ]
         python-version: ["3.10", "3.11", "3.12"]
     name: Tests (${{ matrix.os }} / Python ${{ matrix.python-version }})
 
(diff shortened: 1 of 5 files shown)


kratman:
@BradyPlanden I could not reproduce the original issue when testing this, and fork() on MacOS seems to be problematic with numpy 2.0.0 and the ScipySolver

If you can find a new example where this fails, then we can revisit it, but for now MacOS should stay with the spawn method for multiprocessing

BradyPlanden:
Thanks @kratman! I haven't had time to investigate further; however, I did check that this doesn't affect the multiprocessing setup in PyBOP, and I can confirm that dependent packages don't appear to be affected. I don't think this is surprising, though, given the difference in how we multiprocess. I suspect the reason the __main__ defending issue is cropping up in the CI is due to the change to the default solver to IDAKLU, as multiprocessing occurs within [the solver](https://github.com/pybamm-team/PyBaMM/blob/eac0b55709d7d89239768ff915e37a61e7e36e5f/src/pybamm/solvers/base_solver.py#L859).

MarcBerliner (approved):

PR #5098, opened 2025-07-07
bug: enforce sensitivity shape consistency

martinjrobins:
# Description

This fixes an inconsistency in the shape of the value and sensitivities of a discrete time sum or explicit time integral depending on if output_variables was set or not

## Type of change

# Important checks:

### DIFF
--- a/src/pybamm/solvers/idaklu_solver.py
+++ b/src/pybamm/solvers/idaklu_solver.py
@@ -660,8 +660,13 @@ class IDAKLUSolver(pybamm.BaseSolver):
 
             # handle any time integral variables
             if var in self._time_integral_vars:
+                # time integral variables should all be 1D
+                if data.ndim > 1 and data.shape[1] > 1:
+                    raise pybamm.SolverError(
+                        f"Time integral variable {var} should be 1D, but got shape {data.shape}"
+                    )
                 tiv = self._time_integral_vars[var]
-                data = tiv.postfix(data, sol.t, inputs_dict)
+                data = tiv.postfix(data.reshape(-1), sol.t, inputs_dict)
                 time_indep = True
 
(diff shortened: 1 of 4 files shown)


src/pybamm/solvers/idaklu_solver.py
```diff
@@ -675,22 +677,26 @@ def _post_process_solution(self, sol, model, integration_time, inputs_dict):
...
-                    sens_data = sol.yS[:, start_idx:end_idx, param_idx]
-                    if var in self._time_integral_vars:
-                        tiv = self._time_integral_vars[var]
-                        sens_data = tiv.postfix_sensitivities(
-                            var, data, sol.t, inputs_dict, sens_data
-                        )
-                    newsol[var].add_sensitivity(
-                        param,
-                        [sens_data],
+                if var_length != math.prod(var_shape):
```
MarcBerliner: This check determines if it's a sparse array by checking the length of the non-zero elements, but it fails if the sparse array is filled out. Does the following code still work in that case?

I think it might be clearer if `_get_variable_info` returned the shape and an `is_sparse` flag. The `var_length` name is a bit confusing since it's the length of the non-zero elements

martinjrobins: yes, it will still work. All casadi matrices are sparse so `is_sparse` will always be true. But the bug only occurs when the number of non-zeros is != the product of the shape (i.e. the CCS matrix has implicit zeros). Maybe instead of `var_length` I can use `var_nnz`?

MarcBerliner: Ok understood. In that case, let's just change the name to `var_nnz`

MarcBerliner (approved):
Thanks!

PR #5107, opened 2025-07-09
Override boundary value and provide constant extrapolation

aabills:
# Description

Allows for constant extrapolation of boundary values (i.e. using the value of the last node) and allows the user to override the spatial method's extrapolation method. This is useful for geometric input parameters, if one wants to calculate the harmonic mean of the diffusivity at the interface. A follow up PR (separating because they are pretty unrelated) will finish this capability by creating a node in the expression tree to retrieve the mesh size at the boundary.

Fixes # (issue)

## Type of change

# Important checks:


MarcBerliner (approved):
Thanks!

PR #5134, opened 2025-07-28
[GSOC 2025] Saving and Loading Parameter Sets

medha-14:
# Description

## Type of change

# Important checks:


agriyakhetarpal:
Hi @medha-14, I left a note on your other PR about how you can help us access a better diff: https://github.com/pybamm-team/PyBaMM/pull/5137#issuecomment-3152131640

cc: @Saransh-cpp @santacodes @valentinsulzer for reviews

tests/unit/test_serialisation/test_serialisation.py
```diff
@@ -591,3 +598,787 @@ def test_serialised_model_plotting(self):
...
+        param3 = pybamm.ParameterValues(param_dict)
+
+        param3.update({"Electrode height [m]": "[input]"})
+
+        # Load the model
+        model2 = Serialise.load_custom_model(
+            "my_model.json", battery_model=pybamm.lithium_ion.BaseModel()
+        )
+
+        # ✅ Process the model AFTER update() so the necessary .input_* keys get created
```
santacodes, line 1377: @medha-14 Could you clean up the comments and refine the test functions more, seems like this was generated through an LLM, not against it, I just want to be sure that we keep the code readable and works as intended if you haven't tested it. I think you can add a short `docstring` above each test function instead of comments. Also, if you see any test functions that are irrelevant or not needed, feel free to remove them.

santacodes (changes requested):
@medha-14 left a note for you to work on, I will review it once again after my suggestion is addressed.

tests/unit/test_serialisation/test_serialisation.py
agriyakhetarpal: I have the same suggestion here. I am also not against the use of AI models to generate code, but you must understand what the generated code does both from a microscopic level (say, what's happening in a specific test) and from a macroscopic one (what feature is being implemented, does the API match what we need, are we reusing functions we've previously defined, whether or not the AI system being used is confabulating code or not, etc.).

PR #5137, opened 2025-07-29
[GSoC 2025] Load model JSON files from URLs

medha-14:
# Description

This PR is based on #5056  and contains changes that build upon it.
Once #5056 is merged, the diff for this PR will automatically update to show only the relevant changes.

## Type of change

# Important checks:


medha-14:
I've updated the entry point so that models can now be loaded directly from a URL. This is how it is working at the moment.
```
import pybamm
url = "https://raw.githubusercontent.com/medha-14/model_json/refs/heads/main/dfn.json"
model =  pybamm.Model(url = url,battery_model=pybamm.lithium_ion.BaseModel())
sim.solve([0, 3600])
sim.plot(show_plot=False)
```
Please let me know if any part of this approach needs changing or improvisation.

Saransh-cpp:
@medha-14 can you change the base branch here? Would be much easier to review

src/pybamm/dispatch/entry_points.py
```diff
@@ -109,7 +115,28 @@ def __getattribute__(self, name):
...
+    cache_dir.mkdir(parents=True, exist_ok=True)
+    file_hash = hashlib.md5(url.encode()).hexdigest()
+    return cache_dir / f"{file_hash}.json"
+
+
+def clear_model_cache():
+    cache_dir = Path.home() / ".pybamm_cache" / "pybamm" / "models"
+    if cache_dir.exists():
+        for file in cache_dir.glob("*.json"):
+            file.unlink()
```
agriyakhetarpal: This should be handled using `platformdirs.user_cache_dir`: https://platformdirs.readthedocs.io/en/latest/api.html#cache-directory

agriyakhetarpal:
> @medha-14 can you change the base branch here? Would be much easier to review

Hi @medha-14, could you please create a PR within your fork from this link? https://github.com/medha-14/PyBaMM/compare/GSoC...medha-14:PyBaMM:load_json_from_url. I see a better diff there. We could add comments on that PR until #5056 is merged. Otherwise, I think it is on the right track, modulo that we need to use `platformdirs` and we need tests.

src/pybamm/dispatch/entry_points.py
```diff
@@ -1,6 +1,16 @@
...
+import urllib.request
 from collections.abc import Callable, Mapping
+from pathlib import Path
+
+from platformdirs import user_cache_dir
+
+from pybamm.expression_tree.operations.serialise import Serialise
+
+APP_NAME = "pybamm"
+APP_AUTHOR = "pybamm"
```
agriyakhetarpal: I think we can keep just `APP_NAME` and drop `APP_AUTHOR`, and we can set `Path(user_cache_dir(APP_NAME)) / "models"` as a constant here instead.

src/pybamm/dispatch/entry_points.py
```diff
@@ -109,7 +119,35 @@ def __getattribute__(self, name):
...
+def _get_cache_dir() -> Path:
+    cache_dir = Path(user_cache_dir(APP_NAME, APP_AUTHOR)) / "models"
+    cache_dir.mkdir(parents=True, exist_ok=True)
+    return cache_dir
+
+
+def get_cache_path(url: str) -> Path:
+    cache_dir = _get_cache_dir()
+    file_hash = hashlib.md5(url.encode()).hexdigest()
+    return cache_dir / f"{file_hash}.json"
```
agriyakhetarpal, line 132: Why MD5 and not SHA-256? :)

src/pybamm/dispatch/entry_points.py
```diff
@@ -109,7 +119,35 @@ def __getattribute__(self, name):
...
+
+
+def clear_model_cache() -> None:
+    cache_dir = _get_cache_dir()
+    for file in cache_dir.glob("*.json"):
+        try:
+            file.unlink()
+        except Exception as e:
+            # Optional: log error instead of failing silently
```
agriyakhetarpal:
```suggestion
```

src/pybamm/dispatch/entry_points.py
```diff
@@ -109,7 +119,35 @@ def __getattribute__(self, name):
...
+    file_hash = hashlib.md5(url.encode()).hexdigest()
+    return cache_dir / f"{file_hash}.json"
+
+
+def clear_model_cache() -> None:
+    cache_dir = _get_cache_dir()
+    for file in cache_dir.glob("*.json"):
+        try:
+            file.unlink()
+        except Exception as e:
```
agriyakhetarpal: Note, we also store some of PyBaMM's data files in the cache dir using a [`pooch`](https://github.com/fatiando/pooch) registry: https://github.com/pybamm-team/PyBaMM/blob/a1aa02c2ca108816b69294ee0287cf6729308424/src/pybamm/pybamm_data.py#L126-L139

I wonder if we could reuse some of the code here, because clearing the cache directory of JSON files could have unintended side effects. We do have JSON files there: https://github.com/pybamm-team/pybamm-data/releases/tag/v1.0.1

Could we perhaps rewrite the PR to use `pooch` instead? That will also provide safer defaults than using `urllib.request.urlretrieve()` directly, and we could rely on it for the checksum/caching bits to see if we need to download the model JSON again or not.

src/pybamm/dispatch/entry_points.py
```diff
@@ -109,7 +119,35 @@ def __getattribute__(self, name):
...
+            print(f"Could not delete {file}: {e}")
+
+
+def Model(
+    model=None,
+    url=None,
+    force_download=False,
+    *args,
+    **kwargs,
+):
```
agriyakhetarpal: I think we can have better typing here.

agriyakhetarpal (changes requested):

src/pybamm/dispatch/entry_points.py
medha-14: I have set it to SHA-256.

PR #5174, opened 2025-08-26
add model option override

aabills:
# Description

Allows for use of custom submodels by not erroring when a custom model option is detected.

Fixes # (issue)

## Type of change

# Important checks:


valentinsulzer (commented):
I don't get this, will it automatically use the custom thermal submodel in the example? Or if not what's stopping you from creating the model first using the default options (with build=False), then updating the submodel manually before building?

aabills:
It's for custom full models, not necessarily submodels, it's fine, I can hack it on the other end, it's not worth arguing about.

PR #5191, opened 2025-09-13
Fixes input parameters being used in initial conditions

### LINKED ISSUE #5154: [Bug]: input parameters that affect the initial condition not handled
### PyBaMM Version

develop

### Python Version

3.10

### Describe the bug

If an input parameter is used in the initial conditions of the model, then passing in a list of input parameter sets doesn't calculate the correct initial conditions for each set (only the first set is used)

### Steps to Reproduce

1. setup a model with an input parameter used in the initial condition
2. solve it with multiple input params and check the solution, the initial conditions will be identical for each 

### Relevant log output

```shell

```

pipliggins:
# Description

Input parameters should not be able to be used as initial conditions (see https://github.com/pybamm-team/PyBaMM/pull/1261), but a bug was present where if more than one input parameter was provided but not all were used as initial conditions, the error warning could be bypassed.

Fixes #5154


pipliggins:
> Does this still allow for the case where there is an `InputParameter` in the initial conditions but there is only one solver? I.e. there is just a dict of inputs not a list of dicts. That case should still work IMO.

Still works; going to edit this PR actually to make it a new feature that a list of input parameters can be provided and the inputs will be distributed correctly (useful for PyBoP).

src/pybamm/solvers/base_solver.py
```diff
@@ -505,8 +517,9 @@ def heaviside_event(symbol, expr):
                 )
             )
 
+        # TODO: still needs to be fixed
```
martinjrobins: is this TODO still valid? If so what needs to be fixed?

src/pybamm/solvers/base_solver.py
```diff
@@ -543,7 +556,7 @@ def modulo_event(symbol, expr, num_events):
                 )
 
         def modulo_t_discon(symbol, expr, num_events):
-            value = expr.evaluate(0, model.y0.full(), inputs=inputs)
+            value = expr.evaluate(0, model.y0_list[0].full(), inputs=inputs)
```
martinjrobins: is this anything to worry about? Presumably the input parameters can modify the times of the discontinuity, but we only consider the first set of input parameters.

tests/integration/test_solvers/test_idaklu.py
```diff
@@ -212,51 +212,39 @@ def test_with_experiments(self):
...
 
         t_eval = np.array([0, 1])
         solver = pybamm.IDAKLUSolver()
 
-        base_sol = solver.solve(model, t_eval)
-        y0_base = base_sol.y[:, 0]
-
-        ics = make_ics(y0_base)
-        inputs = [{}] * len(ics)
+        inputs = [{"Current function [A]": value} for value in [0.1, 2.0]]
```
martinjrobins: surely all the initial conditions for this simulation would be identical, I don't think current function changes the initial conditions does it?

martinjrobins (changes requested):
I still need to finish looking through the changes in the tests, but I'll do that tomorrow, just a couple of questions for now

src/pybamm/solvers/base_solver.py
pipliggins: Relates to the below `model.y0_list[0]` call

src/pybamm/solvers/base_solver.py
pipliggins: Decided that this does raise an issue, input parameters could theoretically result in different t_eval points for each input set, if it affects the timing of discontinuity events. Will check t_eval for each y0/input set, and if they are not the same raise an error as unsupported behaviour for now.

pipliggins:
Closing as superseded by #5311

Issue #4539, opened 2024-10-22
Support Python 3.13


agriyakhetarpal:
Now that Python 3.13 stable is out, we can try this everywhere and see if we can add support.

A checklist is as follows:

- [ ] Add Python 3.13 trove classifier and update the rest of the metadata in `pyproject.toml`
- [ ] Add 3.13 in most places in our CI
- [ ] Add a CHANGELOG entry

melanopsis:
Is there any timeline for this and numpy v2?

agriyakhetarpal:
> Is there any timeline for this and numpy v2?

Hi! NumPy v2 compatibility at build time is discussed in #4183, and #5052 is the first step. https://github.com/pybamm-team/pybammsolvers will then need to publish binaries compatible with NumPy v2. #4552 has also been inactive for a while now, but it also requires updating NumPy to v2 and reworking the IREE code.

I can't say there is a timeline here, but the issues have been ironed out and it's possible that it might get done by the next few releases, give or take.

PR #5203, opened 2025-09-18
Remove deprecated JAX funtions

DavidMStraub:
# Description

This PR attempts to make PyBaMM compatible with JAX 0.6.0 and 0.7.0 by removing several deprecated or removed functions (see [JAX changelog](https://docs.jax.dev/en/latest/changelog.html)).

For me, unit and integration tests all pass locally with the following versions installed

- JAX 0.7.0
- Numpy 2.3.0
- Python 3.13.3

Consequently, I also allow Python 3.13 and require Numpy 2.0. Related: #4539 

Disclosure: I used Github Copilot with Claude Sonnet 4.0 to help make these changes.

## Type of change

**-> TODO if/when this is approved**

# Important checks:

**Since I had to remove the Python <3.13 pin in `pybammsolvers` as well, I didn't use nox but manually ran:**

```bash
pytest tests/unit
pytest tests/integration
pytest --doctest-plus src
```
All passed successfully.

I had to change one doc string in `serialise.py` to account for the fact that Numpy 2.0 prints scalars differently (but the actual type does not change, so I don't think we need to change anything at runtime).

### DIFF
--- a/tests/unit/test_solvers/test_jax_bdf_solver.py
+++ b/tests/unit/test_solvers/test_jax_bdf_solver.py
@@ -179,32 +179,32 @@ class TestJaxBDFSolver:
     def test_split_list(self):
         """Test the split_list utility function."""
         from pybamm.solvers.jax_bdf_solver import split_list
-        
+
         # Test case 1: Empty indices should return the original list
         original_list = [1, 2, 3, 4, 5]
         result = split_list(original_list, [])
         assert result == [[1, 2, 3, 4, 5]]
-        
+
         # Test case 2: Single index should split at that point
         result = split_list([1, 2, 3, 4, 5], [3])
         assert result == [[1, 2, 3], [4, 5]]
-        
+
         # Test case 3: Multiple indices should create multiple sublists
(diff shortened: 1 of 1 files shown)


agriyakhetarpal:
Great, thanks for working on this @DavidMStraub! We'll need a new version of https://github.com/pybamm-team/pybammsolvers to support Python 3.13 first, after which we can bump its version here.

DavidMStraub:
CI won't work as `pybammsolvers` needs to be unpinned as well...

DavidMStraub:
Almost done now.

I had to pin Jax to <0.7.0 for the moment because with 0.7.2, CI failed with Python 3.11 and 3.12. I guess we can fix that as a next step.

The coverage reduced because an `ImportError` that is only relevant for older Jax versions is not tested. But it *is* actually tested in reality because it is used on `macos-13` which uses on older Jax version.

MarcBerliner:
Thanks @DavidMStraub, this looks good to me! Can you please add a small test for the `split_list` coverage? I'm happy to approve if you also want to bump the jax upper bound

PR #5214, opened 2025-09-24
Volume fractions and porosity that vary in y and z

DrSOKane:
Added support for porosity and active material volume fractions that vary in y and z, not just x. It appears to work for the z coordinate (`"dimensionality": 1`) but the y coordinate appears to get replaced with the `StateVector` between experiment steps. I tried changing all instances of `y` in `lithium_ion_parameters.py` to `y_cc` but that didn't change anything.

Added an example `graded_pouch.py` to show that it works in one dimension.


.github/CODEOWNERS
```diff
@@ -1,2 +0,0 @@
-# Automatically request reviews from maintainers
```
rtimms, line 1: can you add this file back?

rtimms (changes requested):
thanks @SOK-Elysia looks good except re-adding CODEOWNERS

PR #5215, opened 2025-09-29
allow custom initial state

aabills:
# Description

Allows for custom implementation of set_initial_state. Open to other ways of doing this.

Fixes # (issue)

## Type of change

# Important checks:

### DIFF
--- a/src/pybamm/parameters/parameter_values.py
+++ b/src/pybamm/parameters/parameter_values.py
@@ -65,10 +65,15 @@ class ParameterValues:
                     f"'{values}' is not a valid parameter set. Parameter set must be one of:\n{valid_sets}"
                 )
 
-        if chemistry == "ecm":
-            self._set_initial_state = pybamm.equivalent_circuit.set_initial_state
-        else:
-            self._set_initial_state = pybamm.lithium_ion.set_initial_state
+        if "Initial state function" not in self._dict_items:
+            if chemistry == "ecm":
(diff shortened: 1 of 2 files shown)


valentinsulzer (changes requested):
I don't really like putting it as a parameter because it's a different "type" of thing, it seems more natural to make it a method of a custom subclass of `ParameterValues`

aabills:
> I don't really like putting it as a parameter because it's a different "type" of thing, it seems more natural to make it a method of a custom subclass of `ParameterValues`

That's fine with me, I left it as that because that's where we keep "chemistry", which is used for this and only this. Shall I remove the "chemistry" parameter as well?

valentinsulzer:
Yes let's remove chemistry as well



On Mon, Sep 29, 2025 at 4:56 PM, Alec Bills ***@***.******@***.***>> wrote:
[https://avatars.githubusercontent.com/u/48105066?s=20&v=4]aabills left a comment (pybamm-team/PyBaMM#5215)<https://github.com/pybamm-team/PyBaMM/pull/5215#issuecomment-3349024044>

I don't really like putting it as a parameter because it's a different "type" of thing, it seems more natural to make it a method of a custom subclass of ParameterValues

That's fine with me, I left it as that because that's where we keep "chemistry", which is used for this and only this. Shall I remove the "chemistry" parameter as well?

aabills:
Just pointing out that's a breaking change

PR #5229, opened 2025-10-15
Plot voltage components for composite electrodes

### LINKED ISSUE #4434: Plot Voltage Components for Composite Electrodes

Daniel-Nicolae23:
# Description
Adding an option to choose which phase of a blended electrode is used for plotting overpotentials in `sol.plot_voltage_components()`.

Fixes #4434

## Type of change

New feature (non-breaking change which adds functionality)

# Important checks:

### DIFF
--- a/tests/unit/test_plotting/test_plot_voltage_components.py
+++ b/tests/unit/test_plotting/test_plot_voltage_components.py
@@ -14,13 +14,12 @@ def solved_simulations():
     sim = pybamm.Simulation(model)
     sol = sim.solve([0, 3600])
 
-    model_composite = pybamm.lithium_ion.SPM({
-        "particle phases": ("2", "1")
-    })
+    model_composite = pybamm.lithium_ion.SPM({"particle phases": ("2", "1")})
     params = pybamm.ParameterValues("Chen2020_composite")
     sim_composite = pybamm.Simulation(model_composite, parameter_values=params)
     sol_composite = sim_composite.solve([0, 3600])
     return sim, sol, sim_composite, sol_composite
+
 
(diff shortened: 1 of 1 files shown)


DrSOKane (approved):
The feature has clearly been shown to work and does not appear to conflict with any possible model options. Approved subject to tests passing.

PR #5239, opened 2025-10-24
Update discussion_autoresponder.yml

### LINKED ISSUE #5186: Discussion autoresponder is down
See the logs at https://github.com/pybamm-team/PyBaMM/actions/runs/17429381866/job/49483948686; I think the commit pointing to the tag changed, so we should also ensure that the action is safe and maintained upstream, otherwise switch to a maintained one or use a short JS script with `@actions/github-script` instead.

We should fix this soon, as people are now being inclined to post to GitHub Discussions and they are no longer being redirected to Discourse.

yuvikaaaaaa:
# Description

This PR replaces the broken GitHub Action wesleyscholl/discussion-auto-responder with a safe, maintained solution using @actions/github-script.

Fixes #5186 

## Type of change

 - Bug fix (workflow was broken due to unmaintained action)
 - Improvement (replaced with a safer, controlled script)

# Important checks:

 - No style issues: nox -s pre-commit
 - All tests pass: nox -s tests
 - Documentation builds: nox -s doctests
 - Code is commented for clarity
 - Tests/workflows verified to function as expected

### DIFF
--- a/.github/workflows/discussion_autoresponder.yml
+++ b/.github/workflows/discussion_autoresponder.yml
@@ -4,21 +4,42 @@ on:
   discussion:
     types: [created]
 
-
-permissions: {}
+permissions:
+  discussions: write
+  contents: read
 
 jobs:
   autorespond:
     name: Autorespond to New Discussions
     runs-on: ubuntu-latest
-    permissions:
-      discussions: write
-      contents: read
 
     steps:
-      - name: Run Discussion Autoresponder
-        uses: wesleyscholl/discussion-auto-responder@b1a3c1b9a1e3d1b1a3c1b9a1e3d1b1a3c1b9a1e3 # v1.0.8
+      - name: Redirect New Discussions to Discourse
+        uses: actions/github-script@v7
         with:
-          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/discussion_autoresponder.yml
```diff
@@ -4,21 +4,42 @@ on:
   discussion:
     types: [created]
 
-
-permissions: {}
+permissions:
+  discussions: write
+  contents: read
```
agriyakhetarpal, line 9: Hmm, could you please explain this change? It would be better to leave the permissions blank here (like it was previously), and add the permissions back to the `autorespond` job.

.github/workflows/discussion_autoresponder.yml
```diff
@@ -4,21 +4,42 @@ on:
...
+            // Add a comment
+            await github.rest.issues.createComment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: discussion.number,
+              body: message
+            });
+
+            // Wait 1 second to ensure comment is posted
+            await new Promise(resolve => setTimeout(resolve, 1000));
```
agriyakhetarpal, line 37: I get what you're trying to do here, but it is not necessary here. We don't need to create a promise and wait until it's resolved, as we can just do `await github.rest.discussions.createComment(...);` and the comment will be posted immediately.

PR #5240, opened 2025-10-25
feat(solvers): add max_wall_time for solver timeout handling

### LINKED ISSUE #5161: Add walltime timeout to solvers
### Description

Allow simulations to be terminated after a given amount of running time. This would add a `max_wall_time` argument or similar to the `solve()` method of a `Simulation`.

### Motivation

Cluster-based workflows often have walltime limits for jobs. This means if a pybamm simulation exceeds this walltime the cluster terminates the process leaving no results for the user. It would be better if the user could set a walltime for pybamm lower than the cluster walltime to enable the simulation to exit gracefully before the job ends.

### Possible Implementation

This appears to be relatively simple to implement directly in the `BaseSolver` class. This is demonstrated by the example below where I have subclassed the `IDAKLUSolver` class:
```python
import time
import pybamm

class TimeoutSolver(pybamm.IDAKLUSolver):
    def __init__(self, timeout_seconds, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.timeout_seconds = timeout_seconds
        self._start_time = None

    def _ensure_timer_started(self):
        """Make sure we have a start time set before checking timeout."""
        if self._start_time is None:
            self._start_time = time.time()

    def solve(self, *args, **kwargs):
        self._ensure_timer_started()
        return super().solve(*args, **kwargs)

    def step(self, *args, **kwargs):
        self._ensure_timer_started()
        if time.time() - self._start_time > self.timeout_seconds:
            raise pybamm.SolverError("Wallclock timeout reached")
        return super().step(*args, **kwargs)


    ("Discharge at 1C until 2.5V", "Rest for 1 hour", 
     "Charge at 1C until 4.2V", "Rest for 1 hour")
]*100) 

sim = pybamm.Simulation(model, solver=TimeoutSolver(timeout_seconds=5), experiment=experiment)
start = time.time()
solution = sim.solve()
print("5s walltime: ", len(solution.t), "timesteps", "Elapsed time:", time.time() - start)

sim = pybamm.Simulation(model, solver=pybamm.IDAKLUSolver(), experiment=experiment)
start = time.time()
solution = sim.solve()
print("Unlimited walltime: ", len(solution.t), "timesteps", "Elapsed time:", time.time() - start)
```
Returns:
```bash
2025-08-13 19:27:38.198 - [ERROR] callbacks.on_experiment_error(233): Simulation error: Wallclock timeout reached
5s walltime:  16794 timesteps Elapsed time: 5.46578574180603
Unlimited walltime:  49412 timesteps Elapsed time: 14.546168088912964
Unlimited walltime:  49412 timesteps Elapsed time: 14.546168088912964
```
This works in `step()` as I have used an experiment. I think like the `save_at_cycles` option, this would only be feasible when running with an experiment, as some form of looping is required for the if statement to be triggered.

### Additional context

_No response_

r0hansaxena:
# Description

This PR introduces a max_wall_time parameter to the BaseSolver class, with support extended to child classes like CasadiSolver and IDAKLUSolver. The parameter enables wall-clock timeout handling to stop solver steps if they exceed a set time limit, avoiding prolonged hangs. It is optional (default None, meaning no limit) and specified in seconds. When enabled, the solver records the start time and checks elapsed time in BaseSolver.step() using time.time(). If the limit is reached, it raises a SolverError ("Wall time limit exceeded during step") and retains partial Solution objects for analysis. Users can set it during solver creation (e.g., CasadiSolver(max_wall_time=3600)) or pass it via **kwargs in Simulation.solve(). In experiment runs, the _solve_steps() method re-raises timeout errors for clear handling, while logging others as standard.​

No specific issue is resolved; this is a new enhancement for solver stability.

Fixes #5161

## Type of change

This is an enhancement adding timeout functionality to solvers and simulations.

max_wall_time parameter to BaseSolver.__init__() (propagated to child solvers CasadiSolver and IDAKLUSolver), enabling optional wall-clock timeout checks in BaseSolver.step() using time.time(); raises SolverError on exceedance while preserving partial Solution objects for analysis; supports setting via Simulation.solve(**kwargs) and clean re-raising in Simulation._solve_steps() for experiments

# Important checks:

- [x] Tests added that prove fix is effective or that feature works (new `TestWallTimeTimeout` class with 3 tests: normal SPM run without timeout, timeout via solver init on long experiment, timeout via `solve()` kwarg—all pass with `pytest -v`)

### DIFF
--- a/src/pybamm/simulation.py
+++ b/src/pybamm/simulation.py
@@ -427,6 +427,8 @@ class Simulation:
             Initial State of Charge (SOC) for the simulation. Must be between 0 and 1.
             If given, overwrites the initial concentrations provided in the parameter
             set.
+        direction : str, optional
+            Direction of the solve ("forward" or "backward"). Defaults to "forward".
         callbacks : list of callbacks, optional
             A list of callbacks to be called at each time step. Each callback must
             implement all the methods defined in :class:`pybamm.callbacks.BaseCallback`.
@@ -434,9 +436,15 @@ class Simulation:
             Whether to show a progress bar for cycling. If true, shows a progress bar
             for cycles. Has no effect when not used with an experiment.
             Default is False.
+        inputs : dict, optional
+            Dictionary of input values to override model defaults. If None, uses built-in
+            values from the parameter set.
         t_interp : None, list or ndarray, optional
             The times (in seconds) at which to interpolate the solution. Defaults to None.
             Only valid for solvers that support intra-solve interpolation (`IDAKLUSolver`).
+        initial_conditions : dict, optional
+            Dictionary of initial conditions for the variables. If None, the initial
+            conditions are inferred from the model.
         max_wall_time : float, optional
             Maximum wall-clock time (in seconds) for the entire solve process. If
             exceeded during solver steps, raises a :class:`pybamm.SolverError` while


src/pybamm/solvers/base_solver.py
```diff
@@ -795,6 +799,11 @@ def solve(
         """
         pybamm.logger.info(f"Start solving {model.name} with {self.name}")
 
+        import time
```
BradyPlanden: Let's move the import to the top of the file with the others.

src/pybamm/solvers/base_solver.py
BradyPlanden, line 1: Instead of providing `max_wall_time` as a construction argument, it would be more useful to have it as a runtime argument to `solver.solve()`, allowing users to modify at runtime, instead needing to re-construct or modify a setter.

src/pybamm/simulation.py
```diff
@@ -380,6 +381,7 @@ def solve(
         inputs=None,
         t_interp=None,
         initial_conditions=None,
+        max_wall_time=None,
```
BradyPlanden, line 384: This will need an entry in the docstring for the `solve` method, with corresponding units

PR #5244, opened 2025-10-27
Serialization refactor

aabills:
# Description

Fixes # (issue)

## Type of change

# Important checks:

### DIFF
--- a/src/pybamm/plotting/plot_voltage_components.py
+++ b/src/pybamm/plotting/plot_voltage_components.py
@@ -110,6 +110,12 @@ def plot_voltage_components(
             "Ohmic negative electrode overpotential",
             "Ohmic positive electrode overpotential",
         ]
+    # Only add the contact overpotential label if its values are not (numerically) all zero
+    if not np.allclose(
+        solution["Contact overpotential [V]"].entries, 0, atol=1e-12, equal_nan=True
+    ):
+        overpotentials.append("Contact overpotential [V]")
+        labels.append("Contact overpotential")
 
     # Plot
     # Initialise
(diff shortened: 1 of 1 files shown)


valentinsulzer (approved):

valentinsulzer (approved):

pyproject.toml
```diff
@@ -29,6 +29,7 @@ classifiers = [
 ]
 dependencies = [
     "pybammsolvers>=0.3.0,<0.4.0",
+    "black",
```
agriyakhetarpal, line 32: We support `ruff` already which should subsume `black`. But you want to keep it, could you please add it to `[dev]` instead?

aabills: Black here is used to format python source code, which is not supported by ruff: https://github.com/astral-sh/ruff/discussions/8539 

I also don't think it should be in `dev`, since it's required for serialization. I think that would be quite confusing for users.

agriyakhetarpal: Ah, got it – yeah, then this makes sense to be here. Resolving!

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -33,13 +33,30 @@ def to_source(self):
...
+                isinstance(child, pybamm.Parameter) and child.name not in self.func_args
+            ):
                 child.name = f'Parameter("{child.name}")'
 
-        src += f"    return {expression.to_equation()}\n"
-        return src
+        src += f"    return {expression.to_equation()}"
+
+        formatted_src = black.format_str(src, mode=black.FileMode())
+        return formatted_src
```
agriyakhetarpal, line 59: Non-blocking for now, but if you want to explore how to avoid `black` later on, I would recommend implementing just line-length management and indentation support, which should be simple. My suggestion aligns with the fact that this is intended to be read by machines, not by humans.

aabills: I need this PR kind of urgently, but would be happy to approve a later PR that changes this.

agriyakhetarpal: Yes, of course!

Issue #3131, opened 2023-07-10
Assign a spacial temperature vector as initial temperatrue field for DFN model


xyyrly:
### Description

Hi team,
I try to set an initial temperature distribution in x dimension for DFN but now it looks only a scalar value can be set for the whole domain. Could it be fixed?   Thanks!

tommaull:
I'll take this one

xyyrly:
Thank you very much!

tommaull:
xyyrly - assuming you were using the xfull temperature model?

xyyrly:
Hi tommaull,

I am using a new thermal model to calculate electrolyte temperature along x dimension and particle temperature along r dimension,  kind of like DFN model.

tommaull:
Can you write out the steps to replicate with existing models?

xyyrly:
Oh sorry I was confused. I think here what I hope to realize is to assign a spatical temperature distribution to DFN model without subthermal model. Is it possible?

tommaull:
I think a thermal model is required. My approach was going to be to add something like: 

def set_initial_conditions(self, variables):
        T = variables["Cell temperature [K]"]

        if isinstance(self.param.T, (list, np.ndarray)):
            if len(self.param.T) == 3:
                # Get sizes of negative, separator, and positive domains
                sizes = {
                    "negative electrode": self.mesh["negative electrode"].npts,
                    "separator": self.mesh["separator"].npts,
                    "positive electrode": self.mesh["positive electrode"].npts,
                }

                # Check if sizes match the length of T_init
                if all(size == len(T_init) for size in sizes.values()):
                    self.initial_conditions = {
                        "negative electrode": self.param.T_init[
                            : sizes["negative electrode"]
                        ],
                        "separator": self.param.T_init[
                            sizes["negative electrode"] : sizes["negative electrode"]
                            + sizes["separator"]
                        ],
                        "positive electrode": self.param.T_init[
                            sizes["negative electrode"] + sizes["separator"] :
                        ],
                    }
                else:
                    raise ValueError("T_init vector size does not match domain sizes")
            else:
                raise ValueError("T_init should be a vector of length 3")
        else:
            self.initial_conditions = {T: np.full_like(T, self.param.T_init)}

rtimms:
[...]

PR #5248, opened 2025-10-29
Fix: correct temperature initialisation in lumped and x_full thermal …

swastim01:
### Description

While exploring the thermal submodels and related discussions around spatial initialization, I noticed that the lumped thermal model’s temperature variable and initial condition setup did not fully align with spatial models when a nonuniform `T_init` was specified.

This PR introduces small but necessary fixes to make the model more consistent:

- Added the `"current collector"` domain to `T_vol_av` in `lumped.py`.
- Removed the redundant `PrimaryBroadcast` of `T_vol_av`.
- Updated the initial condition to compute the x-averaged initial temperature using `pybamm.x_average(self.param.T_init)`, ensuring correct initialization from spatial fields.
- Minor adjustments in `x_full.py` for domain consistency.

These minimal changes ensure that the lumped model behaves consistently with the spatial ones while retaining its simpler formulation.

I went through related discussions in earlier thermal PRs and issues to understand the context, but I’m open to feedback from maintainers in case any aspect needs deeper alignment with the broader thermal framework.

### **Fixes**  
Fixes: #3131
discussion used for reference:  #3151

### Changelog

Add the following under **Unreleased → Bug Fixes** in [`CHANGELOG.md`](https://github.com/pybamm-team/PyBaMM/blob/develop/CHANGELOG.md):

```md
- Fixed domain handling and temperature initialization in the lumped thermal model to ensure consistent behavior with spatial `T_init`. ([#5248](https://github.com/pybamm-team/PyBaMM/pull/5248))
````

---

### Important Checks

* [x] No style issues: `nox -s pre-commit`
* [x] All tests pass: `nox -s tests`
* [x] Documentation builds: `nox -s doctests`
* [x] Code is commented for hard-to-understand areas
* [ ] Tests added or verified to prove fix effectiveness

---

### Notes

This PR is a small consistency fix based on observations during exploration of the thermal submodels.
I’m open to feedback or suggestions from maintainers for further improvements.

### DIFF
--- a/src/pybamm/models/submodels/thermal/lumped.py
+++ b/src/pybamm/models/submodels/thermal/lumped.py
@@ -27,11 +27,10 @@ class Lumped(BaseThermal):
     def get_fundamental_variables(self):
         T_vol_av = pybamm.Variable(
             "Volume-averaged cell temperature [K]",
-            domain=["current collector"],
             scale=self.param.T_ref,
             print_name="T_av",
         )
-        T_x_av = T_vol_av
+        T_x_av = pybamm.PrimaryBroadcast(T_vol_av, ["current collector"])
         T_dict = {
             "negative current collector": T_x_av,
             "positive current collector": T_x_av,
@@ -90,11 +89,7 @@ class Lumped(BaseThermal):
 
     def set_initial_conditions(self, variables):
         T_vol_av = variables["Volume-averaged cell temperature [K]"]
-
-        # Compute x-averaged initial temperature from spatial field T_init(x)
         T_init_av = pybamm.x_average(self.param.T_init)
-        T_init_av = pybamm.PrimaryBroadcast(T_init_av, T_vol_av.domain)
-
         self.initial_conditions = {T_vol_av: T_init_av}
 
     def calculate_Q_cr_W(self, current, contact_resistance):

### CODE REVIEW

File: src/pybamm/models/submodels/thermal/lumped.py [hunk 1]
```diff
@@ -27,10 +27,11 @@ def __init__(self, param, options=None, x_average=False):
     def get_fundamental_variables(self):
         T_vol_av = pybamm.Variable(
             "Volume-averaged cell temperature [K]",
+            domain=["current collector"],
```
rtimms: thanks! the initial conditions fix looks good, but this variable shouldn't have a domain since it is just a scalar. for example, if comparing with models with higher-dimensional current collector domains, this will lead to errors since the current collector domain may have some length

rtimms (changes requested):

src/pybamm/models/submodels/thermal/lumped.py
swastim01: Thanks for the clarification! I have removed the domain assignment so the variable remains a scalar and doesn’t cause shape mismatches with models having higher-dimensional current collector domains. Please let me know if any further changes/improvements can be made.

edit: just pushed the update.

rtimms: thanks! LGTM

PR #5261, opened 2025-11-04
Fix/discussion autoresponder 5186

### LINKED ISSUE #5186: Discussion autoresponder is down

swastim01:
## Description

Replaced the unmaintained `wesleyscholl/discussion-auto-responder` GitHub Action with a safe, maintained implementation using `actions/github-script`.
This fixes the broken Discussion Autoresponder workflow that was preventing automatic replies to new discussions.

### Fixes:
Fixes #5186

### Type of change

* [x] Bug fix (non-breaking change which fixes an issue)
* [x] Workflow improvement

Added changelog entry:
`- Fixed Discussion Autoresponder workflow by replacing the unmaintained action with a maintained github-script implementation ([#5261](https://github.com/pybamm-team/PyBaMM/pull/5261))`

### Important checks

* [x] No style issues: `nox -s pre-commit`
* [x] All tests pass: `nox -s tests`
* [x] The documentation builds: `nox -s doctests`
* [x] Code is commented for clarity
* [x] Workflow verified to function as expected

### DIFF
--- a/.github/workflows/discussion_autoresponder.yml
+++ b/.github/workflows/discussion_autoresponder.yml
@@ -22,12 +22,17 @@ jobs:
             const discussion = context.payload.discussion;
             const message = "Hi! We have now moved our discussions to [Discourse](https://pybamm.discourse.group/). Please post your question there.";
 
-            await github.request('POST /repos/{owner}/{repo}/discussions/{discussion_number}/comments', {
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              discussion_number: discussion.number,
+            const query = `
+              mutation($discussionId: ID!, $body: String!) {
+                addDiscussionComment(input: {discussionId: $discussionId, body: $body}) {
+                  comment {
+                    id
+                  }
+                }
+              }
+            `;
+
+            await github.graphql(query, {
+              discussionId: discussion.node_id,
               body: message,
-              headers: {
-                'X-GitHub-Api-Version': '2022-11-28',
-              },
             });


CHANGELOG.md
```diff
@@ -22,6 +22,7 @@
 
 ## Bug fixes
 
+- Fixed Discussion Autoresponder workflow by replacing the unmaintained action with a maintained github-script implementation ([#5261](https://github.com/pybamm-team/PyBaMM/pull/5261))
```
agriyakhetarpal: Since this is not a user-facing change, but rather a repository-level one, let's remove this note:


```suggestion
```

.github/workflows/discussion_autoresponder.yml
```diff
@@ -17,8 +16,15 @@ jobs:
...
+          script: |
+            const discussion = context.payload.discussion;
+            const message = `Hi! We have now moved our discussions to [Discourse](https://pybamm.discourse.group/). Please post your question there.`;
+
+            await github.rest.discussions.createComment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              discussion_number: discussion.number,
+              body: message,
+            });
```
agriyakhetarpal, line 38: Looks good! Could you please enable Discussions on your fork, merge this PR into your fork's main branch, post a dummy discussion, and verify that this is working as intended?  An alternative is to point me to a different repository that's doing something similar and showing that it's working for them. Please feel free to choose either option!

swastim01: Thanks for your review

I tested the updated workflow on my fork and it triggers correctly on new discussions but fails with a `HttpError: Not Found`. I think this might be due to limited GITHUB_TOKEN permissions on forks, which prevent access to the Discussions API.

While testing, I also found that `github.rest.discussions.createComment` isn’t available in the Actions runtime and throws a `TypeError`. I’ve therefore updated the workflow to use:

```
github.request('POST /repos/{owner}/{repo}/discussions/{discussion_number}/comments')
```

This correctly calls the Discussions API. The workflow runs and triggers as expected, and should function properly once merged into the main PyBaMM repository, where the token will have the required Discussions permissions.

agriyakhetarpal: Hmm, I think sending this POST request is fine. However, I don't think GITHUB_TOKEN is the issue, though, as the workflow would run in the context of your fork and your fork's Discussions. Perhaps we're missing a `Accept: 'application/vnd.github.echo-preview+json'` header here?

I'd recommend trying that first. If that doesn't work, how about we use the GraphQL API instead? Here are the relevant docs: https://docs.github.com/en/graphql/guides/using-the-graphql-api-for-discussions.

Please let me know if you need any help!

swastim01: Thank you for the suggestion! I tried adding the Accept: `application/vnd.github.echo-preview+json` header, but the REST call still returned a `HttpError: Not Found`.
I’ve now switched the workflow to use the GraphQL API (addDiscussionComment mutation) as you suggested, and it works correctly on my fork, the autoresponder now triggers and posts comments as expected.
image

agriyakhetarpal (approved):
Perfect, thanks, @swastim01!

Issue #4502, opened 2024-10-09
Use `pytest`'s fixtures to reduce repetition across tests


prady0t:
Pytest offers various features, one of them being the use of [fixtures](https://docs.pytest.org/en/6.2.x/fixture.html). This helps us better manage the tests and setup preconditions. Using the `pytest.mark.parameterize` fixture helps us give multiple inputs to test functions. 
Learn more : https://docs.pytest.org/en/stable/how-to/parametrize.html.

We can take for example

```py
        assert isinstance(processed[1][0][0], np.ndarray)
        assert isinstance(processed[1][0][1], np.ndarray)
        assert isinstance(processed[1][1], np.ndarray)
```
inside a fixture and work up things accordingly.

prady0t:
@agriyakhetarpal We can put this for hacktoberfest.

agriyakhetarpal:
I added the labels – additionally, could you please link to some code examples (a few low-hanging fruits) where you've noticed that this is the case, so that it's easier for potential contributors to know what to start with?

Aswinr24:
Hey would love to pick up this issue !
I understand that we must set up fixtures to handle multiple cases here and thereby consolidating tests for different inputs into a single, reusable function, therefore making the code more concise.
Can I start working on this ?

arjxn-py:
Sure @Aswinr24, feel free to work on this and also don't hesitate to let us know if you need help at any point :)

Aswinr24:
Could You Please Review my PR and suggest if any changes are required

Aswinr24:
Hi @agriyakhetarpal, hope you're doing well!
I've opened this PR as part of a series where I aim to parameterize tests wherever possible, following your suggestions. Could you please review this PR and let me know if any changes are needed?

Thank you!

swastim01:
Hi! I noticed some PRs already linked to this issue, is there still scope for additional refactoring using pytest fixtures in other test modules? I’d be happy to help with that.

agriyakhetarpal:
Hi, @swastim01! Indeed, there is still scope for additional refactoring. There is #4401 that @prady0t previously started, but it wasn't touched for a while and was later closed. The idea is that we can move most of our `shared.py` test harnesses into `conftest.py`. I am open to hearing ideas on how this could be done. I suppose you could draw some inspiration from that PR in a PR of your own (while including @prady0t's previous commits, of course!)

mxxn-ai:
[...]

PR #5294, opened 2025-11-26
Feat/pytest fixtures parameters

swastim01:
# Description

Refactor `test_process_parameter_data` by introducing a pytest fixture to remove repeated ndarray checks.

contributes to issue: #4502 

## Type of change

# Important checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,7 +45,6 @@
 
 ## Bug fixes
 
-- Fixed Discussion Autoresponder workflow by replacing the unmaintained action with a maintained github-script implementation ([#5261](https://github.com/pybamm-team/PyBaMM/pull/5261))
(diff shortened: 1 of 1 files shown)


agriyakhetarpal:
Hi @swastim01, thanks – this is clearer now. I’m hesitant to merge these changes for similar reasons I mentioned in your other PR #5295. This change splits the code into two files, and I find it makes navigation a bit harder and the code less readable. The current `assert isinstance(<...>, np.ndarray)` statements are more straightforward.

My aim is not to discourage you from contributing, so please don't take this as such – rather, it would be great if we could align closer with how pytest fixtures are used in other open source projects.

It might be useful to adapt the test mesh generation functions in `tests/shared.py` as fixtures, since they’re reused frequently in our tests. Also, that file should be converted into `tests/conftest.py`. Here's an example to consider: https://github.com/pybamm-team/PyBaMM/blob/eecf5e5fa925dc472a10fbdab94e90438811f1f3/tests/shared.py#L123-L218

Thanks again for your work!

PR #5295, opened 2025-11-26
Feat/pytest fixtures parameter values

swastim01:
# Description

Refactor test_parameter_values by adding simple_param fixture to reduce duplication.

Contribution to #4502

## Type of change

# Important checks:

### DIFF
--- a/tests/unit/test_parameters/test_parameter_values.py
+++ b/tests/unit/test_parameters/test_parameter_values.py
@@ -41,10 +41,9 @@ class TestParameterValues:
         with pytest.raises(ValueError, match="'Junk' is not a valid parameter set."):
             pybamm.ParameterValues("Junk")
 
-    def test_repr(self):
-        param = pybamm.ParameterValues({"a": 1})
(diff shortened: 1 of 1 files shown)


agriyakhetarpal:
Hi @swastim01! Thanks for putting this together. The idea for the PR is correct, but I am leaning against this change, since we'd want this fixture used in several more places (that's the purpose of a fixture, after all – reusability!). We're using it in only two tests here, so we are not worried about repetition.

My judgement for introducing a new fixture would be that it should address boilerplate code that exists as a precursor to several tests, and that writing it would extract such boilerplate code for reuse. It is alright to have fixtures with little purpose or limited scope, but this one is a bit too trivial in that regard, as you can see.

I'll proceed to close this PR for now, but I would be happy to review more PRs from you. Please feel free to tag me on the PRs, or preferably ping me on Slack, as I may not see them coming.

Also, FYI, you can write the fixture inline in the test file at the top, too, besides a separate `conftest.py` file 😉

PR #5308, opened 2025-12-05
Observe custom variables after solving + faster simulation build time

MarcBerliner:
# Description

Adds a few features:
- All values in the `model.variables` dict are now lazily evaluated as a post-processing step, which reduces build time by ~**2.25x**. The total time to solve the first simulation is now **~1.5x** faster.
- Adds the ability to lazily observe custom pybamm expressions after solving the simulation. E.g.,
```python
import pybamm

sol = sim.solve([0, 1000])

soc = model.variables["Discharge capacity [A.h]"] / model.variables["Nominal cell capacity [A.h]"]

sol.observe(soc)
# <pybamm.solvers.processed_variable.ProcessedVariable0D at 0x104606c90>
```
which supports all variables. This requires that the solver does not have`output_variables`. 

## Type of change

# Important checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Reduced the time to build `Simulation`s by creating a post-processing step for variables. ([#5308](https://github.com/pybamm-team/PyBaMM/pull/5308))
 - Adds the ability to observe custom variables from a `Solution` object using `Solution.observe(symbol)`. ([#5308](https://github.com/pybamm-team/PyBaMM/pull/5308))
+- Added inverse kinetics for linear kinetics. ([#5303](https://github.com/pybamm-team/PyBaMM/pull/5303))
 - Adds `silence_sundials_errors` IDAKLU solver option with `default=False` to match historical output. ([#5290](https://github.com/pybamm-team/PyBaMM/pull/5290))
 
 ## Bug fixes
--- a/docs/source/api/models/submodels/interface/kinetics/inverse_kinetics/base_inverse.rst
+++ b/docs/source/api/models/submodels/interface/kinetics/inverse_kinetics/base_inverse.rst
@@ -1,0 +1,5 @@
+Base Inverse Kinetics
+=====================
+
+.. autoclass:: pybamm.kinetics.BaseInverseKinetics
+    :members:
--- a/docs/source/api/models/submodels/interface/kinetics/inverse_kinetics/index.rst
+++ b/docs/source/api/models/submodels/interface/kinetics/inverse_kinetics/index.rst
@@ -3,4 +3,6 @@ Inverse Kinetics
 
 .. toctree::
 
+ base_inverse
  inverse_butler_volmer
+ inverse_linear
--- a/docs/source/api/models/submodels/interface/kinetics/inverse_kinetics/inverse_linear.rst
+++ b/docs/source/api/models/submodels/interface/kinetics/inverse_kinetics/inverse_linear.rst
@@ -1,0 +1,5 @@
+Inverse Linear
+==============
+
+.. autoclass:: pybamm.kinetics.InverseLinear
+    :members:
--- a/docs/source/examples/notebooks/models/using-submodels.ipynb
+++ b/docs/source/examples/notebooks/models/using-submodels.ipynb
@@ -27,6 +27,9 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
+      "\n",
+      "\u001b[1m[\u001b[0m\u001b[34;49mnotice\u001b[0m\u001b[1;39;49m]\u001b[0m\u001b[39;49m A new release of pip is available: \u001b[0m\u001b[31;49m25.1.1\u001b[0m\u001b[39;49m -> \u001b[0m\u001b[32;49m25.3\u001b[0m\n",
+      "\u001b[1m[\u001b[0m\u001b[34;49mnotice\u001b[0m\u001b[1;39;49m]\u001b[0m\u001b[39;49m To update, run: \u001b[0m\u001b[32;49mpip install --upgrade pip\u001b[0m\n",
       "Note: you may need to restart the kernel to use updated packages.\n"
      ]
     }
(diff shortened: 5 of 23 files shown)


src/pybamm/expression_tree/operations/replace_symbols.py
```diff
@@ -0,0 +1,248 @@
+import pybamm
```
MarcBerliner: `SymbolReplacer` and its test are copied from https://github.com/pybamm-team/pybamm-eis/blob/c88e134b28c50070dbf7b904c020c16b4b73e1d9/pybammeis/utils.py

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -435,6 +435,13 @@ def serialise_custom_model(model: pybamm.BaseModel) -> dict:
                 str(variable_name): convert_symbol_to_json(expression)
                 for variable_name, expression in getattr(model, "variables", {}).items()
             },
+            "parameter_values": (
+                Serialise._serialise_parameter_values(
+                    getattr(model, "_parameter_values", None)
+                )
+                if getattr(model, "_parameter_values", None) is not None
+                else None
+            ),
```
valentinsulzer: I don't think adding parameter values as an attribute to the model is the right pattern, I could see this causing some bugs. It would be better to have a "parameterized model" object which is model + parameter values.

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -1549,6 +1571,73 @@ def _convert_options(self, d):
         else:
             return d
 
+    @staticmethod
+    def _serialise_parameter_values(
```
valentinsulzer: why not use `ParameterValues.to_json()` ?

src/pybamm/expression_tree/operations/serialise.py
valentinsulzer: Is this just here so we can access parameter values when processing a variable in observe? In that case could the parameter values be an optional argument of Solution instead (and throw an error in observe if not provided)?

src/pybamm/expression_tree/operations/replace_symbols.py
MarcBerliner: Edit: removed

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -358,13 +375,17 @@ def serialise_custom_model(model: pybamm.BaseModel) -> dict:
...
+            raise ValueError("Cannot serialise a built model.")
+
         required_attrs = [
             "rhs",
             "algebraic",
             "initial_conditions",
             "boundary_conditions",
             "events",
             "variables",
+            "_solution_observable",
```
valentinsulzer: why does `solution_observable` matter if the model has not been built yet?

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -1571,73 +1574,6 @@ def _convert_options(self, d):
         else:
             return d
 
-    @staticmethod
```
valentinsulzer: why was this removed?

src/pybamm/plotting/quick_plot.py
```diff
@@ -222,6 +225,7 @@ def t_sample(sol):
         output_variable_tuples = []
         self.variable_limits = {}
         for variable_list in output_variables:
+            print(variable_list)
```
valentinsulzer: remove?

src/pybamm/solvers/base_solver.py
```diff
@@ -1608,21 +1610,63 @@ def _set_up_model_inputs(model, inputs):
...
+                inputs_in_model[name] = value
+
+        if solution_is_observable and missing_inputs:
+            missing_inputs_str = ", ".join(f"'{n}'" for n in missing_inputs)
+            pybamm.logger.warning(
+                f"No value provided for input: [{missing_inputs_str}]. "
+                "The `Solution` can no longer be observed with `solution.observe(symbol)`. "
+                "To fix this, provide all input parameter values to `simulation.solve(..., inputs=inputs)` "
+                "or `simulation.step(..., inputs=inputs)`.",
+            )
```
valentinsulzer, line 1666: Is this a situation you have encountered before? It's potentially a bit too "clever" - I think it's ok to error if an input parameter is missing even if it's only in the variables

tests/integration/test_solvers/test_solution.py
```diff
@@ -49,3 +50,46 @@ def test_append(self):
...
+            # Check exact equality
+            np.testing.assert_array_equal(out_unobservable, out_observable)
+
+        # check that observe works with input parameters that are not part of the model
+        out_true = sol_observable["Current [A]"].data * inputs["dummy"] ** 2
+        symbol = pybamm.Parameter("dummy") ** 2 * model.variables["Current [A]"]
+        np.testing.assert_allclose(
+            sol_observable.observe(symbol).data,
+            out_true,
+        )
```
valentinsulzer: it would make more sense to me to pass any new input parameters to `observe`, i.e. `sol.observe(symbol, inputs=inputs)`, is that possible?

src/pybamm/expression_tree/operations/serialise.py
valentinsulzer: sorry must have reviewed an outdated version

tests/integration/test_solvers/test_solution.py
MarcBerliner: I think that's a related but different thing. This test checks that an input parameter, which is included in the `parameter_values` dict but not in `model.rhs`/`model.algebraic`/`model.initial_conditions`, can still be observed. If you tried to pass a truly new input parameter to `observe`, it would error. I'll update the comment to be clearer

src/pybamm/solvers/base_solver.py
MarcBerliner: I'd also like to raise an error, but essentially, we need this for cases where the input parameters are unknown at `solve` time, which may occur when setting initial states. I changed the `DUMMY_INPUT_PARAMETER_VALUE` to `np.nan` to make dummy parameter values more clearly incorrect

valentinsulzer (approved):

PR #4981, opened 2025-04-17
Added an option for multiple initial conditions in IDAKLU solver

Rishab87:
# Description

Added an option for multiple initial conditions in IDAKLU solver

Fixes #3713 

## Type of change

Please add a line in the relevant section of [CHANGELOG.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CHANGELOG.md) to document the change (include PR #)

# Important checks:

Please confirm the following before marking the PR as ready for review:
- [x] No style issues: `nox -s pre-commit`
- [x] All tests pass: `nox -s tests`
- [x] The documentation builds: `nox -s doctests`
- [x] Code is commented for hard-to-understand areas
- [x] Tests added that prove fix is effective or that feature works

PR #5311, opened 2025-12-12
Lists of input parameters in initial conditions

### LINKED ISSUE #5131: IDAKLU solver JAX: add option for multiple initial conditions
### Description

Following issue #3713, it would be great if we allow specifying multiple initial conditions and solving in parallel using the IDAKLU JAX solver.

### Motivation

Solving many simulations in parallel with different initial conditions is neccessary for implementing Unscented Kalman Filter or Particle Filter based parameter estimation methods, such as those currently being implemented in PyBOP (https://github.com/pybop-team/PyBOP/issues/12). The IDAKLU JAX solver currently has functionality for running multiple simulations with different parameters, but not for different initial conditions

### Possible Implementation

The solver could have an API for overriding the model y0 which allows for passing in either a column vector (single simulation) or a 2D array (n simulations where n is the number of columns). You can refer to #4981.

### Additional context

_No response_

### LINKED ISSUE #5154: [Bug]: input parameters that affect the initial condition not handled

pipliggins:
Re-implementation of #5191 for clearer implementation history

Allows input parameters used as initial conditions to be included within lists of inputs to the solver by converting model.y0 to model.y0_list, and calculating one initial state per input parameter set.

This rolls back a lot of the changes made by #4981 as the functionality is added to all the solvers; the `initial_conditions` argument added for IDAKLU has been removed in favour of allowing all inputs to be included within `inputs`.

Fixes #5154
Fixes #5131


martinjrobins (dismissed):
this looks great, thanks @pipliggins, all good to merge from me, although there was a failing test runner?

martinjrobins (dismissed):
looks like the only CI errors are links, so happy to merge

src/pybamm/solvers/base_solver.py
```diff
@@ -603,8 +568,11 @@ def modulo_t_discon(symbol, expr, num_events):
                 k = 20
                 # address numpy 1.25 deprecation warning: array should have
                 # ndim=0 before conversion
+                # note: assumes that the sign for all batches is the same
                 init_sign = float(
-                    np.sign(event.evaluate(0, model.y0.full(), inputs=inputs)).item()
+                    np.sign(
+                        event.evaluate(0, model.y0_list[0].full(), inputs=inputs)
```
BradyPlanden: Should inputs be a dict here? I.e.
```suggestion
                        event.evaluate(0, model.y0_list[0].full(), inputs=inputs[0])
```

BradyPlanden (approved):

pipliggins:
Contributes to #4087

PR #5315, opened 2025-12-15
Restructure idaklu jax tests

pipliggins:
Restructures the `test_idaklu_jax` file to use pytest fixtures rather than running code during test collection.

Also clears the cached solve after every test.


MarcBerliner (approved):
Looks good! Thanks

PR #5326, opened 2025-12-24
Coupled variable serialization

valentinsulzer:
# Description

Update serialization to work with CoupledVariable.

Fixes # (issue)

## Type of change

# Important checks:


examples/scripts/coupled_variable_example.py
```diff
@@ -0,0 +1,103 @@
...
+                raise ValueError(f"Variable '{depends_on_name}' not found in model")
+            elif hasattr(expr, "children") and expr.children:
+                new_children = [replace_coupled_vars(c) for c in expr.children]
+                return expr.create_copy(new_children=new_children)
+            return expr
+
+        new_var_expr = replace_coupled_vars(new_var_expr)
+
+        # Add the new variable to the model
+        model.variables[new_var_name] = new_var_expr
```
aabills: I think this function should happen somewhere in pybamm, it shouldn't be the users responsibility to define this.

aabills (changes requested):
can you pls add a test of serialization

examples/scripts/coupled_variable_example.py
valentinsulzer: moved it

examples/scripts/coupled_variable_example.py
```diff
@@ -0,0 +1,63 @@
...
+    # Dictionary of new variables to add (this would come from external source)
+    coupled_vars_dict = {
+        "Double voltage [V]": serialized,
+    }
+
+    # Add the coupled variables to the model
+    add_variables_from_dict(model, coupled_vars_dict)
+
+    print(f"\nCoupled variables: {model.list_coupled_variables()}")
+    print(f"Double voltage expression: {model.variables['Double voltage [V]']}")
```
aabills, line 44: why do we have to serialize here?

valentinsulzer: this example is for the case where you want to be able to serialize the expression that contains the CoupledVariable. If you don't need to serialize then you can just use `model.variables`

aabills (approved):
I still don't understand why you wouldn't just add the variable the normal way and then serialize the whole model, but it's fine, I don't see anything wrong with this per say

PR #5334, opened 2026-01-02
Fix coupled variable

aabills:
# Description

Improves usage of coupled variable. Adds option to resolve before discretization [defaults to False due to high computational expense]. Otherwise resolves lazily in `process_and_register_variable`

Fixes # (issue)

## Type of change

# Important checks:

### DIFF
--- a/src/pybamm/expression_tree/coupled_variable.py
+++ b/src/pybamm/expression_tree/coupled_variable.py
@@ -19,19 +19,6 @@ class CoupledVariable(pybamm.Symbol):
     domain : iterable of str, optional
         List of domains that this coupled variable is valid over.
 
-    Example
-    -------
-    >>> # In submodel 1: reference a variable defined elsewhere
-    >>> temperature = pybamm.CoupledVariable("Temperature", domain="positive electrode")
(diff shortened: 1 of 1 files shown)


src/pybamm/discretisations/discretisation.py
```diff
@@ -48,6 +51,7 @@ def __init__(
         spatial_methods=None,
         check_model=True,
         remove_independent_variables_from_rhs=False,
+        resolve_coupled_variables=False,
```
valentinsulzer, line 54: Is there a reason we'd need to set this to False? Or at least can the default be True?

aabills: I think it's an unnecessary computational expense to traverse the full expression tree. It's only needed if there is a coupled variable somewhere in rhs, algebraic, initial_conditions, or boundary_conditions.  If in the future we start using coupled variable more, we can default it to true, but in the meantime, I think its superfluous.

aabills: to be clear, the classic "double voltage" example works fine whether this is true or false

valentinsulzer (approved):

PR #5347, opened 2026-01-13
Improve composite eSOH

rtimms:
# Description

The current `ElectrodeSOHComposite` is not very robust and often gives solver failures. This adds a "split solve" method that first solves the standard eSOH model using the "primary" OCV curves and total capacities, then solves that all the OCP must be equal in all phases in the electrode at equilibrium. The `set_initial_state` method now tries to solve the full eSOH model. If it fails, it uses the "split" model (which is only approximately correct) to obtain initial conditions, then retries solving the full eSOH model. This improves the robustness considerably based on the test cases I have run.

This PR also ensures the `_0` and `_100` stoichiometry variables are defined consistently when a model contains hysteresis. Specifically, it implements the same definition as used in BPX:  "Stoichiometry limits (x_0, x_100, y_0, y_100) are evaluated in a temperature-independent manner (at the reference temperature, ignoring entropy effects), so that the SOC range is not temperature-dependent. In the presence of a hysteresis model, equilibration in the charging direction is assumed for 100% SOC (charging OCP branch), and equilibration in the discharging direction is assumed for 0% SOC (discharging OCP branch)."

Fixes # (issue)

## Type of change

# Important checks:


src/pybamm/models/full_battery_models/lithium_ion/electrode_soh.py
```diff
@@ -7,7 +7,7 @@
 
 import pybamm
 
-from .util import _get_lithiation_delithiation
+from .util import _get_equilibrium_direction, _get_lithiation_delithiation
```
BradyPlanden: Should these be public methods? We seem to use them in a couple areas across the codebase.

src/pybamm/models/full_battery_models/lithium_ion/electrode_soh_composite.py
```diff
@@ -3,10 +3,58 @@
...
+    Create options dict with only primary phase OCP settings.
+
+    When composite options have tuple OCP settings like
+    (("single", "hysteresis"), "single"), the non-composite model needs only the
+    primary (first) element for each electrode to avoid incorrectly detecting
+    hysteresis from the secondary phase.
+
+    Example transformation:
+        Input: {"open-circuit potential": (("single", "hysteresis"), "single")}
+        Output: {"open-circuit potential": ("single", "single")}
```
BradyPlanden: Can these be doctest format

src/pybamm/models/full_battery_models/lithium_ion/electrode_soh_composite.py
```diff
@@ -331,6 +514,315 @@ def __init__(
     def default_solver(self):
         return pybamm.AlgebraicSolver(method="lsq")
 
+    @staticmethod
+    def solve_split(
+        initial_value,
```
BradyPlanden: It would be great to add type-hints across the changes

src/pybamm/models/full_battery_models/lithium_ion/electrode_soh_composite.py
```diff
@@ -341,68 +833,114 @@ def get_initial_stoichiometries_composite(
...
+        return ElectrodeSOHComposite.solve_full(
+            initial_value,
+            parameter_values,
+            direction=direction,
+            param=param,
+            options=options,
+            tol=tol,
+            inputs=inputs,
+        )
+    except Exception as first_error:
```
BradyPlanden: Could we catch the expected error instead of the raw Exception?

BradyPlanden (commented):
Just a few general comments

src/pybamm/models/full_battery_models/lithium_ion/electrode_soh_composite.py
```diff
@@ -331,78 +534,446 @@ def __init__(
...
+        )
+
+        if initial_conditions is not None:
+            sim.build()
+            sim.built_model.set_initial_conditions_from(
+                initial_conditions, inputs=all_inputs
+            )
+            sol = sim.solve([0, 1], inputs=all_inputs)
+        else:
+            sol = sim.solve([0, 1], inputs=all_inputs)
```
MarcBerliner: Small thing, but I noticed that the algebraic solver is doing double the work while I was working on a different PR
```suggestion
        if initial_conditions is not None:
            sim.build()
            sim.built_model.set_initial_conditions_from(
                initial_conditions, inputs=all_inputs
            )
        sol = sim.solve([0], inputs=all_inputs)
```

MarcBerliner (changes requested):

src/pybamm/models/full_battery_models/lithium_ion/electrode_soh_composite.py
rtimms: thanks!

BradyPlanden (approved):
Looks good to me - thanks!

MarcBerliner (approved):

PR #4829, opened 2025-02-08
chore: switch to `np.testing.assert_allclose` for better precision in unit tests

Aswinr24:
# Description

This PR refactors the testing methods in the codebase(unit tests directory) by replacing the`np.testing.assert_array_almost_equal` method with `np.testing.assert_allclose` for better precision and flexibility in comparison.

Related to #4488 

## Type of change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (back-end change that speeds up the code)
- [ ] Bug fix (non-breaking change which fixes an issue)

# Key checklist:

- [ ] No style issues: `$ pre-commit run` (or `$ nox -s pre-commit`) (see [CONTRIBUTING.md](https://github.com/pybamm-team/PyBaMM/blob/develop/CONTRIBUTING.md#installing-and-using-pre-commit) for how to set this up to run automatically when committing locally, in just two lines of code)
- [x] All tests pass: `$ python -m pytest` (or `$ nox -s tests`)
- [ ] The documentation builds: `$ python -m pytest --doctest-plus src` (or `$ nox -s doctests`)

You can run integration tests, unit tests, and doctests together at once, using `$ nox -s quick`.

## Further checks:

- [ ] Code is commented, particularly in hard-to-understand areas
- [x] Tests added that prove fix is effective or that feature works

PR #5362, opened 2026-01-22
Switch to `npt.assert_allclose` in unit tests

### LINKED ISSUE #4488: Use `numpy.testing.assert_allclose` for more consistent floating-point comparisons
Currently, we use `np.testing.assert_array_almost_equal` in 404 places across 41 files, but [the NumPy authors no longer recommend it](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_almost_equal.html#numpy.testing.assert_array_almost_equal). We should switch to the recommended [`np.testing.assert_allclose`](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html#numpy.testing.assert_allclose) testing method where we can (which can be in one PR or multiple PRs), across all our tests.

Please note that there are a few differences in the number of arguments both functions take and in the default tolerance(s) set for them, so it's not a quick swap and some failures could be expected – but it should still be easy for someone to take it up in parts.

pipliggins:
Re-implements some of the changes made in #4829 which seem to have since been reverted (probably during merge commits).

Fixes #4488

### DIFF
--- a/tests/unit/test_experiments/test_experiment.py
+++ b/tests/unit/test_experiments/test_experiment.py
@@ -239,7 +239,7 @@ class TestExperiment:
 
         current = solution["Current [A]"].entries
 
-        assert np.allclose(current, 1, atol=1e-3)
+        np.testing.assert_allclose(current, 1, atol=1e-3)
 
     def test_current_step_raises_error_without_operator_with_input_parameters(self):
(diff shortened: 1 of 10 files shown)


MarcBerliner (approved):
Thanks!

PR #5364, opened 2026-01-26
Improve time-to-first-solve

BradyPlanden:
# Description

The aim of this PR is to improve the time-to-first-solve performance of cold and warm starts. To achieve this, a lazy import structure closely following [SPEC001](https://scientific-python.org/specs/spec-0001/) from Scientific-Python has been added for internal submodules, and the management of external imports has been updated.

Here is the performance difference on my M4 Pro machine for time-to-first-solve of the standard DFN.

| Metric | Main | Local | Speedup |
|--------|------|-------|---------|
| Warm start import | 0.667s | 0.336s | **1.98x** |
| Warm start total | 0.787s | 0.500s | **1.57x** |
| Cold start import | 3.289s | 1.849s | **1.78x** |
| Cold start total | 3.461s | 2.246s | **1.54x** |
| Fresh env total | 38.041s | 18.398s | **2.07x** |

There should not be any breaking changes to the submodule API for end-users. However, the documentation does require updating full paths. 

An alternative implementation was presented in https://github.com/pybamm-team/PyBaMM/pull/3732; however, this approach greatly simplifies the added code and shouldn't cause massive challenges to developers.

Fixes # (issue)

## Type of change

# Important checks:


.github/workflows/test_on_push.yml
```diff
@@ -40,6 +40,12 @@ jobs:
       - name: Check style
         run: uvx pre-commit run -a
 
+      - name: Check stub file is up to date
```
agriyakhetarpal, line 43: Could you please add this into the `.pre-commit-config.yaml` file directly as a pre-commit hook? This way we won't need to add another step here and it will be covered by `uvx pre-commit run -a` above:

```yaml
- id: stub-file-validation
  name: Check that the stub file is up to date
  entry: python scripts/generate_pyi_stub.py
  language: python
  files: ^src/pybamm/*.py$
  pass_filenames: true
```

Something like this should work (please validate the regex here).

scripts/generate_pyi_stub.py
```diff
@@ -0,0 +1,172 @@
...
+import argparse
+import importlib
+import sys
+from pathlib import Path
+
+from pybamm._lazy_config import EAGER_IMPORTS, LAZY_IMPORTS, SUBMODULE_ALIASES
+
+# Add src to path for imports
+REPO_ROOT = Path(__file__).parent.parent
+sys.path.insert(0, str(REPO_ROOT / "src"))
```
agriyakhetarpal, line 34: This should happen before the `pybamm._lazy_config` import line so it picks up the latest code, right? Could you reorder them?

src/pybamm/citations.py
```diff
@@ -32,7 +32,9 @@ class Citations:
     def __init__(self):
         self._check_for_bibtex()
         # Dict mapping citations keys to BibTex entries
-        self._all_citations: dict[str, str] = dict()
+        self._all_citations: dict = dict()
```
agriyakhetarpal, line 35: I think this should be `dict[str, Entry]`, so this is a slight regression in type annotations

src/pybamm/__init__.pyi
```diff
@@ -0,0 +1,524 @@
...
+
+# batch_study
+from .batch_study import BatchStudy as BatchStudy
+
+# pybamm_data
+from .pybamm_data import DataLoader as DataLoader
+
+# dispatch
+from .dispatch import parameter_sets as parameter_sets
+from .dispatch import models as models
```
agriyakhetarpal, line 490: This was not in the `__init__.py` file, could you please double check?

src/pybamm/telemetry.py
```diff
@@ -1,47 +1,61 @@
...
+_posthog = None
+_disabled = False
+
 
-class MockTelemetry:
-    def __init__(self):
-        self.disabled = True
+def _get_posthog():
+    """Lazily initialize the posthog client on first use."""
+    global _posthog, _disabled
```
agriyakhetarpal, line 12: This looks like it would induce a race pretty easily. Two threads calling `_get_posthog()` simultaneously would both see `_posthog` is `None` and both attempt initialisation. This is the same for `_jax_configured` and `get_jax`. Could you add a lock here? We previously had an issue about something like this back when telemetry was recently implemented and we are working around it by `PYBAMM_DISABLE_TELEMETRY`. Also, it doesn't look like the thread safety tests are reaching these.

tests/unit/test_lazy_imports.py
```diff
@@ -0,0 +1,300 @@
...
+        """Test that get_jax configures JAX for float64 precision."""
+        import pybamm
+
+        jax = pybamm.get_jax()
+        # After calling get_jax, x64 should be enabled (unless on Metal)
+        platform = jax.lib.xla_bridge.get_backend().platform.casefold()
+        if platform != "metal":
+            assert jax.config.x64_enabled, "JAX x64 should be enabled after get_jax()"
+
+    @pytest.mark.skipif(not __import__("pybamm").has_jax(), reason="JAX not installed")
```
agriyakhetarpal, line 247: This (and the rest of the JAX tests) can simply use `pytest.importorskip()`: https://docs.pytest.org/en/stable/reference/reference.html#pytest-importorskip

(otherwise we import `pybamm` and `has_jax` which will slow down the test collection)

scripts/generate_pyi_stub.py
```diff
@@ -0,0 +1,172 @@
...
+        "",
+    ]
+
+    # Eager imports
+    for module_path, attrs in EAGER_IMPORTS.items():
+        module_name = module_path.split(".")[-1]
+        lines.append(f"# {module_name}")
+        for attr in attrs:
+            lines.append(f"from {module_path} import {attr} as {attr}")
+        lines.append("")
```
agriyakhetarpal, line 56: We can make the order deterministic and that'll help reduce PR diffs a bit:
```suggestion
    for module_path in sorted(EAGER_IMPORTS):
        attrs = EAGER_IMPORTS[module_path]
        module_name = module_path.split(".")[-1]
        lines.append(f"# {module_name}")
        for attr in attrs:
            lines.append(f"from {module_path} import {attr} as {attr}")
        lines.append("")
```

scripts/generate_pyi_stub.py
```diff
@@ -0,0 +1,172 @@
...
+    # Lazy imports
+    lines.append("# LAZILY LOADED (via lazy_loader stub mechanism)")
+    lines.append("")
+
+    for module_path, attrs in LAZY_IMPORTS.items():
+        module_name = module_path.split(".")[-1]
+        lines.append(f"# {module_name}")
+        for attr in attrs:
+            lines.append(f"from {module_path} import {attr} as {attr}")
+        lines.append("")
```
agriyakhetarpal, line 72: Same here:

```suggestion
    for module_path in sorted(LAZY_IMPORTS):
        attrs = LAZY_IMPORTS[module_path]
        module_name = module_path.split(".")[-1]
        lines.append(f"# {module_name}")
        for attr in attrs:
            lines.append(f"from {module_path} import {attr} as {attr}")
        lines.append("")
```

agriyakhetarpal (commented):
Hi @BradyPlanden, wow, this is a lot of code for optimising our import time! At the same time, the numbers speak for themselves, so it's nice to see this. I've added a few comments below. It is very difficult to do a code review this large in one sitting, so I'm sure I might have missed something though.

Also, I notice that there are still a few eager imports that you noted as intentional. That kind of undermines the lazy import strategy and (I haven't tracked this) undoes some of the improvements. I wonder if we could do a warm-up on the first solve when we create a simulation instead of those?

A few more things:
- The `.pyi` files must be distributed both in the sdist and the wheels; for `hatchling`, these are not handled by default, but I do not remember closely. We will need to add them to `[tool.hatch.build.targets.sdist]` and so on.
- Considering the size of this change, could you please add something about it to the contributing guide as appropriate?

PR #5368, opened 2026-01-28
Migrate dev and docs dependencies to dependency-groups

### LINKED ISSUE #5327: Use dependency groups
This came up in #5195. We should use the modern [dependency groups](https://peps.python.org/pep-0735/) instead of `optional-dependencies`.

AlessioBugetti:
Move docs and dev dependencies from project.optional-dependencies to dependency-groups per PEP 735. These are development-only tools, not user-facing package features.

# Description

This PR migrates development and documentation dependencies from `project.optional-dependencies` to the new `dependency-groups` section as specified in [PEP 735](https://peps.python.org/pep-0735/).
See also [readthedocs' documentation](https://docs.readthedocs.com/platform/latest/build-customization.html#install-dependencies-from-dependency-groups).

What changed:
- Moved `docs` dependencies to `[dependency-groups]`
- Moved `dev` dependencies to `[dependency-groups]`
- Kept user-facing optional dependencies (`examples`, `plot`, `cite`, `bpx`, `tqdm`, `jax`, `all`) in `[project.optional-dependencies]`
Fixes #5327 

## Type of change

# Important checks:

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -21,6 +21,14 @@ build:
   tools:
     python: "3.13"
   jobs:
+    pre_create_environment:
+      - asdf plugin add uv
+      - asdf install uv latest
+      - asdf global uv latest
+    create_environment:
+      - uv venv "${READTHEDOCS_VIRTUALENV_PATH}"
+    install:
+      - UV_PROJECT_ENVIRONMENT="${READTHEDOCS_VIRTUALENV_PATH}" uv sync --group docs --extra all
     pre_build:
       - python -m sphinx -b linkcheck -D linkcheck_timeout=20 docs/ ./linkcheck_output
     # Unshallow the git clone otherwise this may cause issues with Sphinx extensions
@@ -37,12 +45,3 @@ build:
       - cd docs/build/ && latexmk -r latexmkrc -pdf -f -dvi- -ps- -jobname=pybamm -interaction=nonstopmode || true
       - test -f docs/build/pybamm.pdf && echo "pybamm.pdf exists. Copying source file to $READTHEDOCS_OUTPUT/pdf/."
       - cp "docs/build/pybamm.pdf" $READTHEDOCS_OUTPUT/pdf/
-
-# Optionally declare the Python requirements required to build your docs
-python:
-  install:
-    - method: pip
-      path: .
-      extra_requirements:
-        - docs
-        - all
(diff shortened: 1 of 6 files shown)


.readthedocs.yaml
```diff
@@ -21,6 +21,10 @@ build:
   tools:
     python: "3.13"
   jobs:
+    install:
+      - pip install --upgrade pip
```
BradyPlanden: Can we use `uv` for python management here, it would align better with the other pybamm workflows.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Features
 
+- Migrated `docs` and `dev` dependencies from `project.optional-dependencies` to `dependency-groups` per PEP 735. ([#5368](https://github.com/pybamm-team/PyBaMM/pull/5368))
```
BradyPlanden: This entry should be listed under 'breaking changes', you can copy a previous release sub-header if there isn't one for breaking changes in the unreleased section

BradyPlanden (changes requested):
Thanks @AlessioBugetti. This seems like a good change, I just have few minor comments. As this would be a breaking change for maintainers which use pip's optional `[]` installation syntax. I.e `pip install pybamm[dev]`, I'd like to hear from others before proceeding (cc @agriyakhetarpal @Saransh-cpp @brosaplanella @MarcBerliner @rtimms @aabills). We will also need to update the installation documentation and the `noxfile` for these changes.

.readthedocs.yaml
```diff
@@ -21,6 +21,10 @@ build:
   tools:
     python: "3.13"
   jobs:
+    install:
+      - pip install --upgrade uv
+      - uv pip install ".[all]"
+      - uv pip install --group docs
```
agriyakhetarpal: Let's follow what the Read The Docs developers are suggesting (I don't mean to say that this isn't fine, but I think using their recommendations would be the path that offers the least number of surprises): https://docs.readthedocs.com/platform/stable/build-customization.html#install-dependencies-with-uv

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ as initial conditions. ([#5311](https://github.com/pybamm-team/PyBaMM/pull/5311)
 
 ## Bug fixes
 
+## Breaking changes
```
agriyakhetarpal, line 13:
```suggestion
## Breaking changes

```

agriyakhetarpal (commented):
Hi @AlessioBugetti, thanks for working on this. LGTM! Please proceed to make the changes to the installation section as well.

@BradyPlanden, this is a breaking change that will matter more to developers rather than users, as these optional dependencies were never supposed to be exposed to users; just that Python packaging standards did not support them at the time.

agriyakhetarpal (approved):
LGTM, thanks @AlessioBugetti!

BradyPlanden (approved):
LGTM -- thanks @AlessioBugetti

PR #5371, opened 2026-02-02
Improve numerical stability of exchange current density and OCPs

MarcBerliner:
# Description

Exchange current densities include `sqrt(c_e)`, `sqrt(c_s)`, and `sqrt(c_max - c_s)` terms that blow up when concentrations -> 0/c_max. Inverse BV kinetics do `arcsinh(j/j0)`, which also blows up when `j0` is small or causes divide by zero errors for `j0=0`. These cause solver failures in edge cases

This PR adds regularised versions of two expressions:
- `RegPower(x, a)`: approximates `|x|^a * sign(x)` but stays linear near zero (based on modelica's `regPow`)
- `Arcsinh2(a, b)`: computes `arcsinh(a/b)` while avoiding blowup/division by zero errors as `b -> 0`

Major changes:
- The exchange current density now automatically replaces all `Sqrt`/`Power` -> `RegPower` to make them more numerically stable at small concentrations
- The barrier function for the OCP is changed from `1e-6/sto` to a more complicated logarithmic expression, whose parameters were tuned to provide a barrier of 1 mV at `sto = 0.001` and 1000 mV at `sto = 0`. This is a significant from the previous version because we no longer have an asymptote as sto -> 0. Some calculations that relied on this assumption, like the ESOH solver, will have slightly different answers in this extreme limit. Most usecases will not change

## Type of change

# Important checks:


examples/scripts/3d_examples/dried_out_pouch.py
```diff
@@ -27,7 +27,8 @@ def eps_n(x, y, z):
 
 
 param.update({"Negative electrode porosity": eps_n})
-var_pts = {"x_n": 8, "x_s": 8, "x_p": 8, "r_n": 8, "r_p": 8, "y": 24, "z": 24}
+# Increase y and z for higher resolution
```
BradyPlanden, line 30: Should this be 'decrease' given the diff is from 24 -> 10?

src/pybamm/expression_tree/binary_operators.py
```diff
@@ -1744,3 +1797,267 @@ def tensor_product(
         A TensorProduct node representing the outer product.
     """
     return TensorProduct(simplify_if_constant(left), simplify_if_constant(right))
+
+
+class RegPower(BinaryOperator):
```
BradyPlanden: I think RegPower could be refactored to inherit from `Function(base, exponent, scale)` (like `Arcsinh2`), then `scale` could be a child node in the expression tree. This would allows us to remove some of the special processing that occurs for this attribute (i.e. in parameter substitution, jacobian, etc.).

src/pybamm/expression_tree/functions.py
```diff
@@ -329,6 +329,154 @@ def arcsinh(child: pybamm.Symbol):
...
+def arcsinh2(
+    a: pybamm.Symbol | float,
+    b: pybamm.Symbol | float,
+    eps: float | None = None,
+) -> pybamm.Symbol:
+    """
+    Two-argument arcsinh function for arcsinh(a/b) that avoids division by zero
+    by adding a small regularisation term to the denominator.
+
+    Computes sign(a) * sign(b) * (log(|a| + hypot(|a|, b_eff)) - log(b_eff))
```
BradyPlanden: I believe this should be: `arcsinh(a / b_eff) where b_eff = sign(b) * hypot(b, eps)`

src/pybamm/parameters/lithium_ion_parameters.py
```diff
@@ -886,3 +932,59 @@ def beta_LAM(self, T, direction=None):
...
+    c = -7.7e-4
+
+    # Clamp sto to avoid exp overflow (float64 max is ~exp(709))
+    max_safe_exp = 700
+    sto_limit = c - max_safe_exp / b
+
+    # For sto >= sto_limit: use log(1 + exp(...))
+    # For sto < sto_limit: continue linearly with slope -b -- this is an exact floating
+    # point match for large sto values.
+    out = np.log(1 + np.exp(-b * (np.maximum(sto_limit, sto) - c))) + -b * (
```
BradyPlanden: Should these be pybamm operators instead of numpy?

BradyPlanden (commented):
Thanks @MarcBerliner - this is looking good. Just a few questions / comments

examples/scripts/3d_examples/dried_out_pouch.py
MarcBerliner: This was intentional, the comment is for users if they want more accuracy

src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -87,6 +87,9 @@ def default(self, node: dict):
                         node.initial_condition
                     )
 
+                if hasattr(node, "scale"):
+                    node_dict["scale"] = self.default(node.scale)
+
```
BradyPlanden: I think this is now dead code:
```suggestion
```

BradyPlanden (dismissed):
Looks great, thanks @MarcBerliner! Just one non-blocking comment

src/pybamm/expression_tree/operations/serialise.py
MarcBerliner: Good catch, I missed this one

BradyPlanden (approved):

PR #5379, opened 2026-02-13
Fix file handle leaks in COMSOL comparison scripts

taovinci0:
## Problem

Two example scripts in `examples/scripts/compare_comsol/` open JSON files but never explicitly close them, which can lead to resource leaks.

## Solution

Use context managers (with statements) to ensure files are properly closed after reading.

## Changes

- `compare_comsol_DFN.py`: Wrap `json.load(open(...))` in a `with` statement
- `discharge_curve.py`: Wrap `json.load(open(...))` in a `with` statement

This follows Python best practices and prevents potential resource leaks in example code that users might copy.


MarcBerliner (approved):
Thanks @taovinci0!

PR #5391, opened 2026-02-24
Remove mass matrix inverse and dense matrix temporaries

MarcBerliner:
# Description

These caused a memory bottleneck for large models

## Type of change

# Important checks:

### DIFF
--- a/src/pybamm/models/base_model.py
+++ b/src/pybamm/models/base_model.py
@@ -149,7 +149,6 @@ class BaseModel:
         instance.bounds = properties["bounds"]
         instance.events = properties["events"]
         instance.mass_matrix = properties["mass_matrix"]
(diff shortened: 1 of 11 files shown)


src/pybamm/models/base_model.py
BradyPlanden, line 1: I think we can cleanup the setter/getter and private attribute for `mass_matrix_inv` as well

BradyPlanden (commented):
Looks good, just one comment. Thanks!

BradyPlanden (dismissed):

src/pybamm/models/base_model.py
MarcBerliner: Yes good catch

BradyPlanden (approved):

PR #5394, opened 2026-02-25
add roundtrip serialization tests

valentinsulzer:
# Description

Tests that parameter values are correct after full roundtrip. Threw lots of edge cases at it, no bugs found so far.

Fixes # (issue)

## Type of change

# Important checks:


tests/unit/test_parameters/test_parameter_values_serialisation.py
```diff
@@ -0,0 +1,670 @@
...
+
+def _assert_evaluate_equal(pv_original, pv_loaded, name, inputs):
+    """Assert that a FunctionParameter evaluates identically
+    before and after roundtrip."""
+    fp = pybamm.FunctionParameter(name, inputs)
+    orig = pv_original.process_symbol(fp).evaluate()
+    loaded = pv_loaded.process_symbol(fp).evaluate()
+    assert not np.any(np.isnan(orig)), f"'{name}' original has NaN"
+    np.testing.assert_allclose(orig, loaded)
+
```
MarcBerliner: I think we should check for exact floating point equivalence here and elsewhere for safety

MarcBerliner (changes requested):
Otherwise LGTM

tests/unit/test_parameters/test_parameter_values_serialisation.py
valentinsulzer: updated, except for cases with functions / solvers where we get some machine error

MarcBerliner: All the tests that fail with `_assert_evaluate_equal` are good first candidates for looking into these serialization issues. I think we shouldn't see any machine error even if they go through a solver

MarcBerliner (approved):

PR #5404, opened 2026-03-03
Fix MSMR exchange current density regression

MarcBerliner:
# Description

See https://github.com/pybamm-team/PyBaMM/pull/5371#issuecomment-3991168734

Now getting the same answer as pre-#5371. I believe the first potential difference figure has a different end condition, so the second figure is not 1:1

msmr_results

msmr_results_phi

cc @DrSOKane 

## Type of change

# Important checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ as initial conditions. ([#5311](https://github.com/pybamm-team/PyBaMM/pull/5311)
 
 ## Bug fixes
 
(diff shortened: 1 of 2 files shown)


DrSOKane (approved):
Thank you very much Marc!! I tried another case from the same paper and pre-5371 behaviour was restored for that as well.

Why does the fix work? I assume it's because linear behaviour as `x_j` approaches 0 is not a valid assumption for small `w_j`, which graphite and LFP have.

Approved :)

PR #5411, opened 2026-03-10
Preserve custom variables and events in built-in model to_config

rtimms:
## Summary
- When users modify a built-in model (e.g. `model.variables["My var"] = ...` or `model.events = []`), `to_config()` now detects changes by diffing against a fresh reference model and includes them as optional override keys in the compact config format
- `from_config()` applies overrides after constructing the fresh built-in model
- New public methods: `serialise_builtin_overrides()` and `apply_builtin_overrides()` on `BaseModel`
- Unmodified models produce identical configs to before (no extra keys)

### Config format (when overrides are present)
```json
{
    "type": "SPM",
    "module": "lithium_ion",
    "options": {},
    "custom_variables": {"My variable": "<serialised expression>"},
    "removed_variables": ["Some variable"],
    "events": []
}
```

### Limitations
- Only tracks added/removed variable **keys**; overwriting an existing variable's expression is not detected (documented in docstring)
- Compression flag does not apply to override expressions (they are small)

## Test plan
- [x] 7 new tests in `TestBaseModelToConfig`: extra variable, cleared events, removed variable, added event, combined changes, unmodified no-op, JSON serializable
- [x] All 37 existing serialization tests pass
- [x] Downstream ionworkspipeline and backend changes tested separately

### DIFF
--- a/src/pybamm/models/base_model.py
+++ b/src/pybamm/models/base_model.py
@@ -2050,7 +2050,7 @@ class BaseModel:
         options).  Any differences are written into *model_config* using
         the following optional keys:
 
-        ``extra_variables``
+        ``custom_variables``
             ``dict[str, json]`` – variables the user *added* (keys not
             present in the reference model).  Each value is the symbolic
             expression serialised via
@@ -2094,7 +2094,7 @@ class BaseModel:
         removed = ref_keys - current_keys
 
         if added:
-            model_config["extra_variables"] = {
+            model_config["custom_variables"] = {
                 name: convert_symbol_to_json(self.variables[name])
                 for name in sorted(added)
             }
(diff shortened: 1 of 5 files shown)


valentinsulzer (dismissed):

src/pybamm/models/base_model.py
```diff
@@ -2041,6 +2041,82 @@ def _find_builtin_module(cls):
...
+
+        if (
+            len(self.events) != len(ref.events)
+            or current_event_names != ref_event_names
+        ):
+            model_config["events"] = [
+                {
+                    "name": event.name,
+                    "expression": convert_symbol_to_json(event.expression),
+                    "event_type": event.event_type,
```
BradyPlanden: This has a round-trip issue when serialising the raw enum. 
```suggestion
                    "event_type": event.event_type.value,
```

src/pybamm/models/base_model.py
```diff
@@ -2212,11 +2295,58 @@ def from_config(config: str | dict) -> BaseModel:
...
+                model.variables.pop(name, None)
+
+        # --- Events override ---
+        events_data = data.get("events")
+        if events_data is not None:
+            model.events = [
+                pybamm.Event(
+                    e["name"],
+                    convert_symbol_from_json(e["expression"]),
+                    e["event_type"],
```
BradyPlanden: And the corresponding fix:

```suggestion
                    EventType(e["event_type"]) if isinstance(e["event_type"], int) else e["event_type"],
```

BradyPlanden (changes requested):
This looks good, just one change request

BradyPlanden (approved):

PR #5417, opened 2026-03-17
Porosity change with particle size distributions

DrSOKane:
# Description

Porosity change due to SEI and lithium plating was previously done in terms of thickness. This had the advantage of being able to exclude the initial SEI thickness from the porosity change, but the drawback of rendering porosity change incompatible with particle size distributions.

By deleting the correction for initial SEI thickness, I was able to refactor the porosity change in terms of concentration. To achieve this, I modified the `BaseSEI` class so that the SEI concentration is always a size-averaged variable, as is already done with plating. This is physically justified because the SEI growth rate is a function of the surface potential difference, SEI thickness and various input parameters, none of which vary in the particle size domain.

## Type of change

Removal of the correction for initial thickness makes this a breaking change.

# Important checks:


docs/source/examples/notebooks/models/coupled-degradation.ipynb
```diff
@@ -94,10 +94,25 @@
...
   {
    "cell_type": "code",
-   "execution_count": 4,
+   "execution_count": 8,
    "id": "23cab5d3",
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "ename": "TypeError",
```
rtimms: @MarcBerliner is this warning new? and do you know why docs passed despite this?

DrSOKane: The CI generates new outputs so this output may have been overwritten in the CI. Will look at it later

MarcBerliner: I believe these diffs are generated by the user. This should work for a fresh install since we lower-bound pybammsolvers to `v0.6.0`. Maybe this notebook was evaluated using an older version of pybammsolvers?

DrSOKane: I was having some strange issues and put it down to VSCode's Jupyter extension having bugs when used with virtual environments

rtimms: @DrSOKane I had the same problem, a fresh install should fix it.

PR #5422, opened 2026-03-21
Single model for experiment

### LINKED ISSUE #5407: feat: experimental steps use the same model
### Description

At the moment the equations of the model vary depending on the current experimental step (i.e. current controlled or voltage controlled). It would be advantageous to only have a single model for all steps:
1. efficiency: you need to build each model, and then the solver needs to convert each model to casadi. During solve you need to map the final state from each model to the initial conditions of the next. this would all be eliminated if we only have a single model for each step.
2. sensitivities: also need to map the sensitivities from one step to the next. This is complicated by the fact that some variables are created and removed from one step to the other, so sensitivities cannot be tracked for these variables.

Would it be possible to formulate the models so that the model does not change from one step to the next? E.g. have all variables in the state (current, voltage, etc), and the input current or input voltage is done via observer-like equations with a given relaxation parameter `r`, e.g. `d_current / dt = r * (current - observed_current)`

### Motivation

_No response_

### Possible Implementation

_No response_

### Additional context

_No response_

martinjrobins:
# Description

fixes #5407

This PR adds a unified experiment-model path that reuses one processed model and one solver across compatible experiment steps, instead of building one model per step.

The unified path combines per-step control and termination behavior into shared experiment-wide expressions, and switches the active step through solve-time one-hot inputs.

For step weights $w_i$, step control residuals $g_i(x)$, and step-local termination aggregates $h_i(x)$, the unified model uses:

$$
\sum_i w_i g_i(x) = 0
$$

and

$$
\sum_i w_i h_i(x) = 0
$$

where exactly one $w_i$ is active for each solve step.
UPDATE: the sum over one-hot weights has been replaced by a conditional expression over a single "step index" input, see discussion below

Examples of the control residuals include:

$$
g_i(x) =
\begin{cases}
I - I_{\mathrm{set}} & \text{current step} \\
V - V_{\mathrm{set}} & \text{voltage step} \\
VI - P_{\mathrm{set}} & \text{power step} \\
V - R_{\mathrm{set}} I & \text{resistance step}
\end{cases}
$$

For steps with multiple terminations $f_{ij}(x)$, the step-local aggregate is formed as a nested minimum so that any active termination can stop the step:

$$
h_i(x) = \min_j f_{ij}(x)
$$

The legacy per-step build remains available, and `Simulation` now supports:
- `experiment_model_mode="unified"`
- `experiment_model_mode="legacy"` (default)

Unsupported cases raise clearly in `unified` mode.

## Benchmarks

Benchmark script:
- `scripts/benchmark_unified_experiment.py`

Scenarios:
- `event_driven_cccv`
  - `Charge at C/3 until 4.1 V`
  - `Hold at 4.1 V until C/20`
- `mixed_control`
  - `Discharge at C/20 for 20 minutes`
  - `Charge at 1 A for 10 minutes`
  - `Hold at 4.1 V for 10 minutes`
  - `Discharge at 2 W for 10 minutes`
  - `Discharge at 4 Ohm for 10 minutes`
- `start_time_padding`
  - `Current(1, duration=20 minutes, start_time=2023-01-01 08:00:00)`
  - `Rest(duration=20 minutes, start_time=2023-01-01 09:00:00)`

Models:
- `SPM`
- `DFN`

The script was run on both this branch and `main`. `main` legacy timings were effectively the same as this branch’s legacy timings, so the comparison is consistent.

### Current branch results

| Model | Scenario | Legacy Total (s) | Unified Total (s) | Unified / Legacy |
| --- | --- | ---: | ---: | ---: |
| SPM | event_driven_cccv | 0.335 | 0.314 | 0.94x |
| SPM | mixed_control | 0.481 | 0.155 | 0.32x |
| SPM | start_time_padding | 0.244 | 0.105 | 0.43x |
| DFN | event_driven_cccv | 0.616 | 0.438 | 0.71x |
| DFN | mixed_control | 1.165 | 0.320 | 0.27x |
| DFN | start_time_padding | 0.736 | 0.295 | 0.40x |

### `main` legacy comparison

| Model | Scenario | This branch legacy (s) | `main` legacy (s) |
| --- | --- | ---: | ---: |
| SPM | event_driven_cccv | 0.335 | 0.334 |
| SPM | mixed_control | 0.481 | 0.488 |
| SPM | start_time_padding | 0.244 | 0.241 |
| DFN | event_driven_cccv | 0.616 | 0.618 |
| DFN | mixed_control | 1.165 | 1.181 |
| DFN | start_time_padding | 0.736 | 0.738 |

On this machine, the unified path is faster in all benchmarked cases, with the largest gains in mixed-control and start-time experiments.

### DIFF
--- a/scripts/benchmark_unified_experiment.py
+++ b/scripts/benchmark_unified_experiment.py
@@ -23,6 +23,7 @@ class Scenario:
     description: str
     experiment_factory: callable
     solve_kwargs: dict
+    solve_inputs_list: tuple[dict, ...] = ({}, {}, {})
 
 
 def make_event_driven_cccv_experiment():
@@ -62,6 +63,24 @@ def make_start_time_padding_experiment():
     )
 
 
+def make_input_parameter_repeat_experiment():
+    s = pybamm.step.string
+    return pybamm.Experiment(
+        [
+            (
+                pybamm.step.current(
+                    pybamm.InputParameter("I_app"),
+                    duration=20 * 60,
+                ),
+                s("Charge at 1 A for 10 minutes"),
+                s("Hold at 4.1 V for 10 minutes"),
+                "Discharge at 2 W for 10 minutes",
+                "Discharge at 4 Ohm for 10 minutes",
+            )
+        ]
+    )
+
+
 SCENARIOS = [
     Scenario(
         name="event_driven_cccv",
@@ -81,10 +100,25 @@ SCENARIOS = [
         experiment_factory=make_start_time_padding_experiment,
         solve_kwargs={"calc_esoh": False},
     ),
+    Scenario(
+        name="input_parameter_repeated_solves",
+        description=(
+            "Mixed-control cycle repeated on one built simulation with varying "
+            "input current in the first step"
+        ),
+        experiment_factory=make_input_parameter_repeat_experiment,
+        solve_kwargs={"calc_esoh": False},
+        solve_inputs_list=(
+            {"I_app": 0.5},
+            {"I_app": 1.0},
+            {"I_app": 1.5},
+        ),
+    ),
 ]
 
 MODEL_FACTORIES = {
     "SPM": pybamm.lithium_ion.SPM,
+    "SPMe": pybamm.lithium_ion.SPMe,
     "DFN": pybamm.lithium_ion.DFN,
 }
 
@@ -167,16 +201,27 @@ def run_case(model_name, scenario, mode):
     t0 = time.perf_counter()
     sim.build_for_experiment()
     t1 = time.perf_counter()
-    solution = sim.solve(**scenario.solve_kwargs)
+    solutions = []
+    solve_step_s = []
+    for solve_inputs in scenario.solve_inputs_list:
+        solve_kwargs = dict(scenario.solve_kwargs)
+        if solve_inputs:
+            solve_kwargs["inputs"] = solve_inputs
+        solve_t0 = time.perf_counter()
+        solutions.append(sim.solve(**solve_kwargs))
+        solve_step_s.append(time.perf_counter() - solve_t0)
     t2 = time.perf_counter()
+    solution = solutions[-1]
 
     return {
         "build_s": t1 - t0,
-        "solve_s": t2 - t1,
+        "solve_s": sum(solve_step_s),
+        "solve_step_s": solve_step_s,
         "total_s": t2 - t0,
+        "solve_count": len(solutions),
         "termination": solution.termination,
         "uses_unified_model": getattr(sim, "_experiment_uses_unified_model", False),
-        "solution": solution,
+        "solutions": solutions,
         "status": "ok",
     }
 
@@ -254,7 +299,15 @@ def assert_solutions_match(legacy_solution, unified_solution):
 def verify_scenario_pair(model_name, scenario):
     legacy_result = run_case(model_name, scenario, "legacy")
     unified_result = run_case(model_name, scenario, "unified")
-    assert_solutions_match(legacy_result["solution"], unified_result["solution"])
+    if len(legacy_result["solutions"]) != len(unified_result["solutions"]):
+        _raise_mismatch(
+            "number of repeated solves differs: "
+            f"{len(legacy_result['solutions'])} != {len(unified_result['solutions'])}"
+        )
+    for legacy_solution, unified_solution in zip(
+        legacy_result["solutions"], unified_result["solutions"], strict=True
+    ):
+        assert_solutions_match(legacy_solution, unified_solution)
     gc.collect()
 
 
@@ -266,7 +319,9 @@ def unsupported_result(model_name, scenario, mode, reason):
         "mode": mode,
         "build_s": None,
         "solve_s": None,
+        "solve_step_s": [None] * len(scenario.solve_inputs_list),
         "total_s": None,
+        "solve_count": len(scenario.solve_inputs_list),
         "termination": "n/a",
         "uses_unified_model": False,
         "status": reason,
@@ -278,6 +333,7 @@ def main():
     scenarios = [scenario for scenario in SCENARIOS if scenario.name in args.scenarios]
     results = []
     supported_modes = [mode for mode in MODES if mode_is_supported(mode)]
+    max_solve_count = max(len(scenario.solve_inputs_list) for scenario in scenarios)
 
     print(
         f"Running {len(scenarios)} scenario(s) across {len(args.models)} model(s), "
@@ -320,7 +376,17 @@ def main():
                         "mode": mode,
                         "build_s": median([sample["build_s"] for sample in samples]),
                         "solve_s": median([sample["solve_s"] for sample in samples]),
+                        "solve_step_s": [
+                            median(
+                                [
+                                    sample["solve_step_s"][solve_index]
+                                    for sample in samples
+                                ]
+                            )
+                            for solve_index in range(samples[-1]["solve_count"])
+                        ],
                         "total_s": median([sample["total_s"] for sample in samples]),
+                        "solve_count": samples[-1]["solve_count"],
                         "termination": samples[-1]["termination"],
                         "uses_unified_model": samples[-1]["uses_unified_model"],
                         "status": "ok",
@@ -332,15 +398,35 @@ def main():
             json.dump(results, handle, indent=2)
 
     print("\n## Raw Results")
-    print(
-        "| Model | Scenario | Mode | Build (s) | Solve (s) | Total (s) | Termination |"
+    solve_headers = " | ".join(
+        f"Solve {solve_index} (s)" for solve_index in range(1, max_solve_count + 1)
     )
-    print("| --- | --- | --- | ---: | ---: | ---: | --- |")
+    print(
+        "| Model | Scenario | Mode | Solves | Build (s) | "
+        f"{solve_headers} | Solve Total (s) | Total (s) | Termination |"
+    )
+    print(
+        "| --- | --- | --- | ---: | ---: | "
+        + " | ".join(["---:"] * max_solve_count)
+        + " | ---: | ---: | --- |"
+    )
     for row in results:
+        solve_step_displays = [
+            (
+                format_seconds(row["solve_step_s"][solve_index])
+                if solve_index < len(row["solve_step_s"])
+                and row["solve_step_s"][solve_index] is not None
+                else "n/a"
+            )
+            for solve_index in range(max_solve_count)
+        ]
         print(
             "| "
             f"{row['model']} | {row['scenario']} | {row['mode']} | "
+            f"{row['solve_count']} | "
             f"{format_seconds(row['build_s']) if row['build_s'] is not None else 'n/a'} | "
+            + " | ".join(solve_step_displays)
+            + " | "
             f"{format_seconds(row['solve_s']) if row['solve_s'] is not None else 'n/a'} | "
             f"{format_seconds(row['total_s']) if row['total_s'] is not None else 'n/a'} | "
             f"{row['termination'] if row['status'] == 'ok' else row['status']} |"
--- a/scripts/benchmark_unified_experiment.py
+++ b/scripts/benchmark_unified_experiment.py
@@ -1,0 +1,397 @@
+#!/usr/bin/env python3
+
+import argparse
+import gc
+import inspect
+import json
+import os
+import statistics
+import time
+from dataclasses import dataclass
+from datetime import datetime
+
+os.environ.setdefault("OPENBLAS_NUM_THREADS", "1")
+
+import numpy as np
+
+import pybamm
+
+
+@dataclass(frozen=True)
+class Scenario:
+    name: str
+    description: str
+    experiment_factory: callable
+    solve_kwargs: dict
+
+
+def make_event_driven_cccv_experiment():
+    return pybamm.Experiment(
+        [("Charge at C/3 until 4.1 V", "Hold at 4.1 V until C/20")]
+    )
+
+
+def make_mixed_control_experiment():
+    s = pybamm.step.string
+    return pybamm.Experiment(
+        [
+            (
+                s("Discharge at C/20 for 20 minutes"),
+                s("Charge at 1 A for 10 minutes"),
+                s("Hold at 4.1 V for 10 minutes"),
+                "Discharge at 2 W for 10 minutes",
+                "Discharge at 4 Ohm for 10 minutes",
+            )
+        ]
+    )
+
+
+def make_start_time_padding_experiment():
+    return pybamm.Experiment(
+        [
+            pybamm.step.Current(
+                1,
+                duration=20 * 60,
+                start_time=datetime(2023, 1, 1, 8, 0, 0),
+            ),
+            pybamm.step.rest(
+                duration=20 * 60,
+                start_time=datetime(2023, 1, 1, 9, 0, 0),
+            ),
+        ]
+    )
+
+
+SCENARIOS = [
+    Scenario(
+        name="event_driven_cccv",
+        description="Charge to voltage cutoff, then hold until current cutoff",
+        experiment_factory=make_event_driven_cccv_experiment,
+        solve_kwargs={"initial_soc": 0.2, "calc_esoh": False},
+    ),
+    Scenario(
+        name="mixed_control",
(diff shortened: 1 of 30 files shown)


scripts/benchmark_unified_experiment.py
```diff
@@ -0,0 +1,397 @@
+#!/usr/bin/env python3
```
MarcBerliner, line 1: Can you add timings for the `SPMe` and benchmark repeated solves with input parameters?

src/pybamm/simulation.py
```diff
@@ -213,24 +251,373 @@ def _update_experiment_models_for_capacity(self, inputs, solve_kwargs=None):
...
+        )
+        if current_variable is None:
+            return built_model.build_initial_state_mapper(built_model)
+        current_slice = built_model.y_slices[current_variable][0]
+        if current_slice.stop - current_slice.start != 1:
+            raise pybamm.ModelError(
+                "Unified experiment self-mapper expects a scalar current-control state."
+            )
+
+        current_guess = pybamm.Scalar(0)
```
MarcBerliner: I think we should just use the current from the end of the previous step as the initial guess for the next one. If we reset the initial guess, then we're perturbing the initial guess state vector farther away from a known solution

src/pybamm/simulation.py
```diff
@@ -213,24 +251,373 @@ def _update_experiment_models_for_capacity(self, inputs, solve_kwargs=None):
...
+        self.experiment_unique_steps_to_model = {
+            self._experiment_unified_model_key: processed_model,
+        }
+        # Keep the legacy lookup keys alive even though compatible steps now share one
+        # processed model instance.
+        for step in self.experiment.unique_steps:
+            self.experiment_unique_steps_to_model[step.basic_repr()] = processed_model
+
+    def _build_unified_experiment_state_mapper(self, built_model):
+        """
```
MarcBerliner: Why is this not handled by the generic experiment state mapper that maps itself onto itself?

src/pybamm/simulation.py
```diff
@@ -213,24 +251,373 @@ def _update_experiment_models_for_capacity(self, inputs, solve_kwargs=None):
...
+        if step_solution.t_event is not None:
+            t = float(np.asarray(step_solution.t_event).reshape(-1)[0])
+        else:
+            t = float(step_solution.t[-1])
+
+        class SolutionVariables(dict):
+            def __missing__(self, key):
+                value = step_solution[key](t)
+                self[key] = value
+                return value
```
MarcBerliner, line 489: Can we make this a non-local class

src/pybamm/experiment/step/base_step.py
```diff
@@ -401,20 +401,53 @@ def process_model(
...
         for term in self.termination:
-            event = term.get_event(new_model.variables, self)
+            event = term.get_event(variables, self)
             if event is not None:
-                new_model.events.append(event)
+                events.append(event)
+        return events
+
+    def get_combined_termination_expression(self, variables):
+        events = self.get_termination_events(variables)
```
MarcBerliner, line 417: We could use a casadi conditional here, since its codegen supports switch statements. My concern with the combined event termination expression is that it's potentially no longer continuous and differentiable if multiple events trigger within the same interval, which will negatively impact the root finding. It creates an implementation difference between events in no experiment/legacy experiment/unified experiment, and becomes another step to trace back which event actually fired. Eventually, if we want to hook up non-terminating events into the solver, we'd need that additional layer to figure out which events triggered anyway, so it might be worth building in that separation now

src/pybamm/models/submodels/external_circuit/function_control_external_circuit.py
```diff
@@ -129,16 +129,34 @@ def __init__(self, param, options, control="algebraic"):
...
+            self.weighted_control_residual,
+            options,
+            control="algebraic",
+        )
+
+    def weighted_control_residual(self, variables):
+        expression = pybamm.Scalar(0)
+        for weight_name, builder in self.step_control_builders:
+            expression += pybamm.InputParameter(weight_name) * builder(variables)
+        return expression
```
MarcBerliner: I also think we should use a conditional here. There are some situations where computing the voltage expression is very expensive, and it'd also let us hook up drive cycles to the unified model architecture without incurring an additional cost for evaluating the interpolant

We can also use a single input parameter for the active expression (e.g., we can check for step 0 using `-0.5 < step_input_parameter < 0.5`, step 1 using `0.5 < step_input_parameter < 1.5`, etc.)

src/pybamm/simulation.py
```diff
@@ -213,24 +251,373 @@ def _update_experiment_models_for_capacity(self, inputs, solve_kwargs=None):
...
+            if (
+                isinstance(step, pybamm.step.CustomStepImplicit)
+                and step.control != "algebraic"
+            ):
+                return ["CustomStepImplicit with differential control is not supported"]
+            if issubclass(step.__class__, pybamm.experiment.step.BaseStepImplicit):
+                has_implicit_step = True
+            elif not issubclass(
+                step.__class__, pybamm.experiment.step.BaseStepExplicit
+            ):
```
MarcBerliner: Can we access this as a property?

src/pybamm/simulation.py
```diff
@@ -213,24 +251,373 @@ def _update_experiment_models_for_capacity(self, inputs, solve_kwargs=None):
...
+                "'legacy', or 'per-step'"
+            ) from err
+
+    def _get_unified_experiment_model_blockers(self):
+        if self.experiment is None:
+            return ["no experiment is attached to the simulation"]
+
+        has_implicit_step = False
+        for step in self.experiment.steps:
+            if getattr(step, "is_drive_cycle", False):
```
MarcBerliner: Do we need `getattr`?

src/pybamm/simulation.py
```diff
@@ -213,24 +251,373 @@ def _update_experiment_models_for_capacity(self, inputs, solve_kwargs=None):
...
+            ):
+                return ["CustomStepImplicit with differential control is not supported"]
+            if issubclass(step.__class__, pybamm.experiment.step.BaseStepImplicit):
+                has_implicit_step = True
+            elif not issubclass(
+                step.__class__, pybamm.experiment.step.BaseStepExplicit
+            ):
+                return [f"unsupported experiment step type '{step.__class__.__name__}'"]
+
+        if not has_implicit_step and getattr(self._solver, "ode_solver", False):
```
MarcBerliner: Likewise `ode_solver` is defined for `BaseSolver`

src/pybamm/simulation.py
```diff
@@ -954,14 +1398,23 @@ def solve(
...
+                                self._experiment_padding_rest_weight_name(),
+                                current_solution.t[-1],
+                                temperature,
+                            )
+                        else:
+                            inputs = {
+                                **user_inputs,
+                                "Ambient temperature [K]": temperature,
+                                "start time": current_solution.t[-1],
+                            }
```
MarcBerliner: Can we remove the duplicate setters for ambient temperature and start time here and build unified experiment

src/pybamm/simulation.py
```diff
@@ -213,24 +251,373 @@ def _update_experiment_models_for_capacity(self, inputs, solve_kwargs=None):
...
+                self._experiment_step_weight_input_names,
+                self.experiment.steps,
+                strict=True,
+            )
+        ]
+        if self._experiment_includes_padding_rest:
+            step_control_builders.append(
+                (
+                    self._experiment_padding_rest_weight_name(),
+                    lambda variables: variables["Current [A]"],
```
MarcBerliner: Can we use maybe a partial or small class instead of the lambda for easier serialization and traceability?

src/pybamm/simulation.py
```diff
@@ -213,24 +251,373 @@ def _update_experiment_models_for_capacity(self, inputs, solve_kwargs=None):
...
+                combined_termination_expression,
+            )
+        )
+        pybamm.step.BaseStep.update_voltage_safety_events(new_model)
+
+        processed_model = new_parameter_values.process_model(
+            new_model,
+            inplace=False,
+            delayed_variable_processing=True,
+        )
```
MarcBerliner, line 413: We copy the model at the top, so can we use `inplace=True`?

src/pybamm/simulation.py
```diff
@@ -213,24 +251,373 @@ def _update_experiment_models_for_capacity(self, inputs, solve_kwargs=None):
...
+            elif not issubclass(
+                step.__class__, pybamm.experiment.step.BaseStepExplicit
+            ):
+                return [f"unsupported experiment step type '{step.__class__.__name__}'"]
+
+        if not has_implicit_step and getattr(self._solver, "ode_solver", False):
+            return [
+                "all-explicit experiments require a DAE-capable solver when using "
+                "the unified experiment model"
+            ]
```
MarcBerliner: Is this error message correct? Explicit steps can work with ODEs, but implicit cannot

src/pybamm/simulation.py
```diff
@@ -289,6 +676,31 @@ def _set_up_and_parameterise_experiment(self, solve_kwargs=None):
...
+            return
+        if self._experiment_model_mode == "auto" and not blockers:
+            self._set_up_unified_experiment_model(parameter_values)
+            return
+        if self._experiment_model_mode == "auto" and blockers:
+            pybamm.logger.debug(
+                "Falling back to per-step experiment models: %s",
+                "; ".join(blockers),
+            )
+
```
MarcBerliner, line 615: Can we simplify these if statements

MarcBerliner (changes requested):
Thanks Martin, the build time improvements are a nice win. I left some comments on a few areas I think we should sort out before this is ready to merge

src/pybamm/simulation.py
martinjrobins: yea, that is what I had originally, but ran into problems with the steps that set a constant current (e.g. the rest step which sets current = 0), if the current in the previous step was far away from that set in the current step, then the solver had problems satisfying the new algebraic equation

src/pybamm/simulation.py
martinjrobins: see comment above, if the current is being set to a constant in the next step, you need to set this for the initial condition of the next step.

src/pybamm/experiment/step/base_step.py
martinjrobins: What do you mean by using a switch statement, what are you switching between? At the moment the combined step termination expression is:

$$
\sum_i w_i \min_j f_{ij}(x) = 0
$$

with $i$ is the step number and $f_{ij}$ are all the termination conditions in step $i$. Within any given step only one of the weights is 1, the rest are 0, so all termination events from other steps are zeroed out and not included within the combined expression. 

I agree its not differentiable due to the $\min$ function (it is continuous within an experimental step), I originally used a product, but this caused difficulties if there were multiple $f_{ij} < 0$ due to the minus values cancelling out. But I don't think a switch statement would fix this as its essentially the same as the nested minimum.

src/pybamm/models/submodels/external_circuit/function_control_external_circuit.py
martinjrobins: ahhhh ok, you mean switching as an alternative to the weights? That makes sense, yes could certainly do this. Do we have a switch statement in the expression tree yet? if not I'll add it...

src/pybamm/experiment/step/base_step.py
martinjrobins: sorry, I see you answer my question below! Yea I'll look into the using the switch statement instead of the weights

MarcBerliner: Yeah sorry, I should've been clearer. I meant doing something like this pseudocode so we can preserve the vectorized events
```python
step_input = 0
# initialize to inf for different length events
out[:]= np.inf
match step_input:
    case 0:
        f_events_0(out, y, yp, t, inputs)
    case 1:
        f_events_1(out, y, yp, t, inputs)
    ...
```

src/pybamm/models/submodels/external_circuit/function_control_external_circuit.py
MarcBerliner: We don't have anything like this at the moment. I believe casadi's `sx` ifelse creates codegen that evaluates both paths of the true/false branch, but `mx` creates a proper if/else or switch statement in the C code

src/pybamm/simulation.py
martinjrobins: nope, fixing...

src/pybamm/simulation.py
MarcBerliner: I'm surprised that rest would cause a failure. The current residual is linear in current and on initialization, the first newton iteration should only update the current value since all other residuals are zero. If this is failing, then I think we should reasonably consider it a rootfinder bug. Do you have an example of this failing without the custom initial guess?

martinjrobins: I believe it was the lithium-plating-composite notebook example that triggered the problem. Once I've dealt with the other changes I'll circle back to this and see if I can reproduce the original problem.

src/pybamm/simulation.py
martinjrobins: I played around with this a bit, but I think it is more readable where it is. Moving it away from the function only obscures what is going on, it spreads it out over a wider area when it is more readable all together.

src/pybamm/simulation.py
martinjrobins: yea, that should be an error if an ode solver is used with the unified experiment model (since its always a dae)

src/pybamm/simulation.py
martinjrobins: yea it was the lithium-plating-composite experiment, it fails to find a consistent initial condition at the start of the rest step

src/pybamm/experiment/step/base_step.py
martinjrobins: ok, I've added a conditional node to the expression tree and it seems to work well :)

scripts/benchmark_unified_experiment.py
martinjrobins: good point on the repeated solves. These indicate that the timings are not as straightforward as I hoped. The unified model is slower on the repeated solves (for one of the SPM scenarios it takes almost twice as long). For the particular simulations that I've picked for the benchmark this is often negligable in comparison to the improvements to build time, but presumably there would be situations where the run-time was significantly longer. You can see the updated script, each scenario is run 3 times, and I've added new scenarios for SPMe and an input parameter

martinjrobins: the results are in a separate comment below, but this brings up the question of which mode should be the default, unified or legacy? The fastest depends on the specific ratio of build-time to run-time

src/pybamm/simulation.py
MarcBerliner: I did some digging, and this is related to https://github.com/casadi/casadi/issues/4280. In this notebook, neither the legacy nor the unified formats are converging to the appropriate root tolerance because both terminate early due to the step tolerance. I have a PR in the works that will fix this, but we can leave this be for now

scripts/benchmark_unified_experiment.py
MarcBerliner: I think I lean towards keeping legacy the default for now. From an implementation POV, unified is the clear winner. But from a runtime perspective, I think it's a tough decision. Unified does great at decreasing build time, but it looks like the additional complexity adds non-trivial solve time, especially for `SPM`/`SPMe` with repeated solves and for long-running experiments. I don't expect the unified solve time to ever be faster than legacy for `SPM` and `SPMe` since it adds an additional algebraic equation during CC, but my guess is that there are still some performance improvements left. My vote would be for keeping legacy as default for now, continuing to iterate on unified performance, and making the switch to unified when we're comfortable with the solve time

src/pybamm/simulation.py
martinjrobins: ok, I can remove the mapper bits then, the example notebook will still be failing though

scripts/benchmark_unified_experiment.py
martinjrobins: yea, I'm happy with keeping legacy the default for now and working to improve the unified performance over time.

src/pybamm/simulation.py
martinjrobins: actually, if we make the legacy path as default the notebook won't be failing after all ;) So no worries!

MarcBerliner (approved):
Thanks @martinjrobins!

PR #5427, opened 2026-03-24
Add zizmor config to ignore secrets-outside-env findings

BradyPlanden:
# Description

Suppresses secrets-outside-env warnings for workflows that use repo-level secrets without dedicated GitHub environments.

Fixes # (issue)

## Type of change

# Important checks:

### DIFF
--- a/zizmor.yml
+++ b/zizmor.yml
@@ -1,4 +1,6 @@
 rules:
+  # These workflows use repo-level secrets without GitHub Environments.
+  # Environments add unnecessary friction for these non-deployment workflows.
   secrets-outside-env:
     ignore:
       - docker.yml
--- a/zizmor.yml
+++ b/zizmor.yml
@@ -1,0 +1,9 @@
+rules:
+  secrets-outside-env:
+    ignore:
+      - docker.yml
+      - periodic_benchmarks.yml
+      - run_benchmarks_over_history.yml
+      - run_periodic_tests.yml
+      - test_on_push.yml
+      - validation_benchmarks.yml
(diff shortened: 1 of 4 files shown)


zizmor.yml
```diff
@@ -0,0 +1,11 @@
+rules:
+  # These workflows use repo-level secrets without GitHub Environments.
+  # Environments add unnecessary friction for these non-deployment workflows.
```
Saransh-cpp: Could you expand on this? What friction is being added? I looked at the previous PR and it wasn't clear to me why the workflows were automatically cancelled?

Ideally we should have environments in these workflows as well, as they are technically "deployment" workflows (deploying things on docker/codecov/...). But yes, if things don't work then we can skip it.

Saransh-cpp (commented):
Thanks for looking into this, @BradyPlanden! Please see my comment below -

zizmor.yml
BradyPlanden: The `test_on_push` workflow would add a lot of noise to the repository/slack so I migrated them together to the `zizmor.yml` config. There's probably a middle ground where the daily/manual workflows are defined with environments but `test_on_push` is ignored.  Thoughts?

Saransh-cpp: Ahh, I just saw the notifications on Slack; that is indeed annoying. Your proposal sounds good. We can have the manual/on release/daily workflows with environment and test_on_push without. We can even have the daily workflows without environment to further bring down the noise; whichever you prefer!

BradyPlanden: Sounds good, I've just split them and added environments to the manual dispatch ones.

Saransh-cpp (approved):
Looks good, thanks!

PR #5436, opened 2026-04-07
Refactor casadi conversion

MarcBerliner:
# Description

Refactors the casadi conversion functionality to live in the `Symbol` subclasses, removing the monolithic `CasadiConverter`

Also improves the build/eval time for the pchip interpolator:

#### Pchip build time (µs/call)

| N | Before | After | Speedup |
|--:|-------:|------:|--------:|
| 10 | 3.6 ms | 311 µs | **12x** |
| 100 | 33.5 ms | 347 µs | **97x** |
| 500 | 159.3 ms | 439 µs | **362x** |
| 1000 | 316.8 ms | 566 µs | **560x** |

#### Pchip evaluation time (batch=100, uniform grid, µs/eval)

| N | Before | After | Speedup |
|--:|-------:|------:|--------:|
| 10 | 54.9 | 11.2 | **5x** |
| 100 | 433.5 | 10.6 | **41x** |
| 500 | 2139.4 | 11.7 | **183x** |
| 1000 | 4400.7 | 12.4 | **355x** |

Fixes # (issue)

## Type of change

# Important checks:


src/pybamm/expression_tree/functions.py
```diff
@@ -152,6 +153,30 @@ def _evaluate_for_shape(self):
     def _function_evaluate(self, evaluated_children):
         return self.function(*evaluated_children)
 
+    def _casadi_evaluate(self, *converted_children):
+        """CasADi analog of :meth:`_function_evaluate`. Override in subclasses where the
+        CasADi function differs from the numpy one."""
+        if self.function.__name__.startswith("elementwise_grad_of_"):
```
BradyPlanden: Are the "elementwise_grad_of" functions produced anywhere? I couldn't find them. Seems like they might be dead code leftover from autograd.

src/pybamm/expression_tree/symbol.py
```diff
@@ -980,13 +980,74 @@ def to_casadi(
...
+            casadi expressions. When converting multiple related expressions over the
+            same (t, y, y_dot, inputs), pass a single ``{}`` created by the caller so
+            that shared subgraph nodes are only converted once. The cache is never
+            stored on the symbol itself — its lifetime is controlled by the caller.
+
+        Returns
+        -------
+        :class:`casadi.MX`
+        """
+        pybamm.citations.register("Andersson2019")
```
BradyPlanden: I believe we will call the citation register on every node that hits the `to_casadi condition`. Can we either cache it or move the citation to an object construction?

BradyPlanden (changes requested):
This is looking good, thanks!

src/pybamm/expression_tree/functions.py
MarcBerliner: I'm not where this came from, but tests pass after removing it so should be safe to remove

src/pybamm/expression_tree/symbol.py
MarcBerliner: Good catch, I refactored this and added a test to make sure the child nodes don't retrigger it

BradyPlanden (approved):

PR #5439, opened 2026-04-09
Improve composite model performance and sparsity

MarcBerliner:
# Description

Improved performance of the composite models by removing dense blocks in the Jacobian. Also adds a jacobian sparsity helper in `IDAKLUSolver`

Benchmark: composite SPMe (`particle_phases=(2, 1)`), 1C discharge 3600s

| Metric | `main` | This PR | Change |
|:-------|-------:|--------:|-------:|
| **surface form = algebraic** | | | |
| System size | 124x124 | 127x127 | +3 algebraic equations |
| Jacobian nnz | 3,129 | 866 | **-72%** |
| Sparsity | 79.7% | 94.6% |  |
| Median solve time | 10.7 ms | 6.6 ms | **-38%** |
| Max voltage diff | -- | 3.78e-12 V |  |
| **surface form = differential** | | | |
| System size | 124x124 | 127x127 | +3 algebraic equations |
| Jacobian nnz | 3,129 | 866 | **-72%** |
| Sparsity | 79.7% | 94.6% | |
| Median solve time | 12.6 ms | 9.7 ms | **-23%** |
| Max voltage diff | -- | 1.70e-12 V |  |

sparsity_before_after

## Type of change

# Important checks:


src/pybamm/solvers/idaklu_solver.py
```diff
@@ -1074,3 +1075,76 @@ def reduce_solution(
...
+        if setup is None:
+            raise pybamm.SolverError("Solver not set up. Call set_up() first.")
+        indptr = setup["jac_times_cjmass_colptrs"]
+        indices = setup["jac_times_cjmass_rowvals"]
+        nnz = setup["jac_times_cjmass_nnz"]
+        n = setup["number_of_states"]
+        data = np.ones(nnz)
+        return csc_matrix((data, indices, indptr), shape=(n, n))
+
+    def spy(self, ax=None, *, show_plot=None, **kwargs):
```
BradyPlanden, line 1097: Nice - this is helpful! I believe it's currently reporting IDA's sparsity of `J - cj*M` - it would be good to make that clear in the docstring or update to report only the model jacobian sparsity.

src/pybamm/solvers/idaklu_solver.py
```diff
@@ -1074,3 +1075,76 @@ def reduce_solution(
...
+
+        ax.spy(J, **kwargs)
+
+        ax.set_xlabel("State index")
+        ax.set_ylabel("Equation index")
+
+        info = f"{nnz} nnz, {sparsity:.2f}% sparse"
+        info += f"\n{n} states: {n_rhs} differential and {n_alg} algebraic"
+        ax.set_title(info, fontsize=10)
+        fig.tight_layout()
```
BradyPlanden, line 1145: nit: this `tight_layout()` can cause an error if existing Axes are passed into the method.

src/pybamm/models/submodels/electrolyte_conductivity/surface_potential_form/composite_surface_form_conductivity.py
BradyPlanden, line 1: I dug further into the `full_diffusion` `c_e_av` change and benchmarked it against the old `Full` behaviour. For composite SPMe this is a clear win: Jacobian nnz drops from `3320 -> 866`, and median solve time drops from `8.24 ms -> 6.00 ms`.

  For DFN surface-form cases, though, it looks like a regression:
  - DFN (`surface form="algebraic"`): `902 -> 903` states, `3376 -> 3437` nnz, `14.78 -> 14.94 ms` median solve time, `74.6 -> 87.6 ms` median setup time
  - Composite DFN (`particle phases=("2","1"), surface form="algebraic"`): `1302 -> 1303` states, `4656 -> 4717` nnz, `17.70 -> 19.20 ms` median solve time, `94.0 -> 114.6 ms` median setup time

  That matches the code path: `full_diffusion.Full` now adds `X-averaged electrolyte concentration [mol.m-3]` for all surface-form models, but the DFN surface-form conductivity path never consumes that state, so DFN pays for an extra algebraic variable and denser Jacobian without getting the sparsity benefit.

  I think it would make sense to scope this optimization to the composite/integrated conductivity users that actually benefit from `c_e_av`, rather than applying it to all  `Full` electrolyte diffusion surface-form cases, thoughts?

BradyPlanden (approved):
This is looking good!

packages/pybamm/src/pybamm/solvers/idaklu_solver.py
```diff
@@ -1074,3 +1075,76 @@ def reduce_solution(
...
+        if setup is None:
+            raise pybamm.SolverError("Solver not set up. Call set_up() first.")
+        indptr = setup["jac_times_cjmass_colptrs"]
+        indices = setup["jac_times_cjmass_rowvals"]
+        nnz = setup["jac_times_cjmass_nnz"]
+        n = setup["number_of_states"]
+        data = np.ones(nnz)
+        return csc_matrix((data, indices, indptr), shape=(n, n))
+
+    def spy(self, ax=None, *, show_plot=None, **kwargs):
```
MarcBerliner: Updated the docstring: it reports the pattern of `J - cj*M` that IDA factorizes (the union of the model Jacobian and mass-matrix patterns), rather than the model Jacobian alone.

packages/pybamm/src/pybamm/solvers/idaklu_solver.py
```diff
@@ -1074,3 +1075,76 @@ def reduce_solution(
...
+
+        ax.spy(J, **kwargs)
+
+        ax.set_xlabel("State index")
+        ax.set_ylabel("Equation index")
+
+        info = f"{nnz} nnz, {sparsity:.2f}% sparse"
+        info += f"\n{n} states: {n_rhs} differential and {n_alg} algebraic"
+        ax.set_title(info, fontsize=10)
+        fig.tight_layout()
```
MarcBerliner: Good catch. It was actually a `NameError`, since `fig` only existed when `spy()` created the figure. `tight_layout()` is now only called in that case, and `test_spy` covers passing an existing Axes.

packages/pybamm/src/pybamm/models/submodels/electrolyte_conductivity/surface_potential_form/composite_surface_form_conductivity.py
MarcBerliner: Agreed, thanks for digging into this. The x-averaged concentration state is now opt-in via `electrolyte_diffusion.Full(..., x_average_state=True)`, which only SPMe passes (4cc9604f2). The full conductivity submodel never reads it, so DFN and lead-acid Full no longer carry it. There's a DFN unit test checking it stays an expression there.

`main` vs this branch, 1 h discharge with `IDAKLUSolver` (Chen2020 / Chen2020_composite). Solve is the median of 30 re-solves, setup (`build` + `set_up`) the median of 5, each the median of 3 interleaved rounds:

| Case | States | Jacobian nnz | Solve (ms) | Setup (ms) |
|---|---|---|---|---|
| SPM (control) | 42 → 42 | 118 → 118 | 0.62 → 0.61 | 27 → 27 |
| SPMe | 102 → 102 | 296 → 296 | 2.05 → 2.12 | 67 → 65 |
| SPMe, `surface form="algebraic"` | 104 → 107 | 2968 → 743 | 9.57 → 4.96 | 131 → 104 |
| Composite SPMe | 124 → 127 | 3129 → 866 | 9.41 → 6.81 | 160 → 131 |
| DFN, `surface form="algebraic"` | 902 → 902 | 3376 → 3376 | 15.44 → 15.07 | 161 → 162 |
| Composite DFN, `surface form="algebraic"` | 1302 → 1302 | 4656 → 4656 | 18.77 → 18.31 | 227 → 208 |

DFN is structurally identical to `main` now, and the timing differences there are within noise. Final voltages match `main` to ≤2e-13 V everywhere except composite SPMe, which moves by ~1e-5 V because the new averaged states are solved to tolerance rather than evaluated exactly.

BradyPlanden (approved):
Nice!

PR #5389, opened 2026-02-24
Model to json

rtimms:
# Description

## Summary

Adds `to_config()` / `from_config()` methods to `BaseModel`, `Experiment`, and `BaseSolver`, enabling round-trip serialization of models, experiments, and solvers via a simple config dict.

### Changes

**`BaseModel`** (`src/pybamm/models/base_model.py`)
- `to_config()` — wraps the existing `to_json()` output in `{"type": "custom", "model": ...}`
- `from_config()` — accepts a wrapped config dict, a raw model dict, or a path to a JSON file
- `_write_json_to_file()` — private static helper shared by `to_json` and `to_config` to remove duplication

**`Experiment`** (`src/pybamm/experiment/experiment.py`)
- `to_config()` / `from_config()` — round-trips an experiment as `{"cycles": [[{step}, ...], ...]}`; handles all step types (current, voltage, power, c-rate, rest), `InputParameter` values, and terminations

**`BaseSolver`** (`src/pybamm/solvers/base_solver.py`)
- `to_config()` / `from_config()` — round-trips any solver via `inspect.signature`; handles `CompositeSolver` recursively

**`Serialise`** (`src/pybamm/expression_tree/operations/serialise.py`)
- `serialise_experiment` / `deserialise_experiment`
- `serialise_solver` / `deserialise_solver`
- `_to_json_safe` — numpy-safe JSON conversion helper

**Tests** (`tests/unit/test_serialisation/test_serialisation.py`, `tests/unit/test_models/test_model_to_json.py`)
- `TestExperimentSerialization` — all step types, multi-cycle, `InputParameter`, terminations
- `TestSolverSerialization` — `ScipySolver`, `CasadiSolver`, `CompositeSolver`
- `TestBaseModelToJson` — `BaseModel` round-trip, `to_config` / `from_config`

Fixes # (issue)

## Type of change

Please add a line in the relevant section of [CHANGELOG.md](https://github.com/pybamm-team/PyBaMM/blob/main/CHANGELOG.md) to document the change (include PR #)

# Important checks:

Please confirm the following before marking the PR as ready for review:
- No style issues: `nox -s pre-commit`
- All tests pass: `nox -s tests`
- The documentation builds: `nox -s doctests`
- Code is commented for hard-to-understand areas
- Tests added that prove fix is effective or that feature works

PR #5441, opened 2026-04-09
Fix load_custom_model when base class package is unavailable

rtimms:
## Summary
- `Serialise.load_custom_model` now falls back to `pybamm.BaseModel` (with a warning) when the recorded `base_class` can't be re-imported, instead of raising `ImportError`. This restores the pre-#5389 contract and makes models authored in third-party packages loadable by anyone, since the model content is stored fully symbolically.
- Consolidated the fallback logic: the `"builtins.object"` sentinel now lives in the top-level guard alongside `""` and `"pybamm.BaseModel"`, removing a dead branch inside the `except` clause (importing `builtins` never raises).
- Updated `test_import_base_class_non_builtin_object` to assert the new warn-and-fallback behaviour, and simplified its setup to round-trip through `save_custom_model` + a single JSON field edit rather than hand-rolling the model JSON.

## Context
Reported by a user loading a `HalfCellGITTModel` defined in package on a machine without that package installed.

Before #5389, `load_custom_model(filename, battery_model=None)` defaulted to `pybamm.BaseModel()` and never tried to import anything. #5389 started storing `f"{cls.__module__}.{cls.__name__}"` and `importlib.import_module`-ing it at load time, with no fallback for non-pybamm subclasses. This PR keeps the opportunistic re-import (so subclass behaviour still round-trips when the package *is* installed) but degrades gracefully when it isn't.

## Test plan
- [x] `pytest tests/unit/test_serialisation/test_serialisation.py -k "custom_model or base_class"` — 8 passed
- [ ] CI

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ## Bug fixes
 
+- Fixed `load_custom_model` raising `ImportError` when the saved base class lives in a package not installed in the loading environment. The loader now falls back to `pybamm.BaseModel` with a warning. ([#5441](https://github.com/pybamm-team/PyBaMM/pull/5441))
 - Fixed serialisation bug in 2D finite volume discretisation. ([#5434](https://github.com/pybamm-team/PyBaMM/pull/5434))
 
 ## Breaking changes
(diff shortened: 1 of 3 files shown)


BradyPlanden (commented):
Looks good, could you also add a changelog entry

BradyPlanden (approved):

PR #5448, opened 2026-04-17
Add ahead-of-time casadi compilation

MarcBerliner:
# Description

Adds an alternative codegen path to the casadi functions, which enables ahead-of-time compilatiom with o3 flags. For non-trivial functions, this is generally faster than casadi's default virtual machine method at the expense of slightly slower build times. This is mainly built for situations where you repeatedly solve the simulation and can accept the longer build times

| Model | Workload | Default VM (on main) [ms] | AOT compiled [ms] | speedup |
| --- | --- | ---: | ---: | ---: |
| SPM | Repeated `sim.solve()` | 3.15 | 2.55 | 1.24x |
| SPM | Repeated `sol["Voltage [V]"]` (10k pts) | 6.89 | 2.54 | 2.71x |
| SPMe | Repeated `sim.solve()` | 4.79 | 3.64 | 1.32x |
| SPMe | Repeated `sol["Voltage [V]"]` (10k pts) | 43.31 | 8.50 | 5.09x |
| DFN | Repeated `sim.solve()` | 34.28 | 27.79 | 1.23x |
| DFN | Repeated `sol["Voltage [V]"]` (10k pts) | 0.92 | 1.28 | 0.72x |

Also includes some simplification passes on symbolic expressions to make them easier to optimize during casadi codegen

## Type of change

# Important checks:


src/pybamm/expression_tree/averages.py
```diff
@@ -197,8 +191,7 @@ def x_average(symbol: pybamm.Symbol) -> pybamm.Symbol:
             }
             return pybamm.FullBroadcast(symbol.orphans[0], broadcast_domains=domains)
         elif isinstance(symbol, pybamm.FullBroadcast) and all(
-            dom in ["negative electrode", "separator", "positive electrode"]
-            for dom in symbol.tertiary_domain
+            dom in X_DOMAINS for dom in symbol.tertiary_domain
```
aabills: This seems a bit out of place in this PR.

For another thing, I'm not wild about having anything like "negative electrode" hardcoded in the expression tree, though I realize that this PR is just refactoring it, not introducing it.

src/pybamm/expression_tree/averages.py
```diff
@@ -227,10 +220,27 @@ def x_average(symbol: pybamm.Symbol) -> pybamm.Symbol:
...
+def _is_x_constant(symbol: pybamm.Symbol) -> bool:
+    # True if no Variable/SpatialVariable leaf carries an electrode primary
+    # domain. Classifies ``PrimaryBroadcast(c, electrode)`` with x-independent
+    # ``c`` as x-constant even though the broadcast has an electrode domain.
+    for node in symbol.pre_order():
+        if isinstance(node, pybamm.Variable | pybamm.SpatialVariable) and any(
+            dom in X_DOMAINS for dom in node.domain
+        ):
+            return False
+    return True
```
aabills: can this be done for y and z too?

src/pybamm/solvers/idaklu_solver.py
```diff
@@ -50,6 +51,10 @@ class IDAKLUSolver(pybamm.BaseSolver):
             options = {
                 # Print statistics of the solver after every solve
                 "print_stats": False,
+                # Casadi function backend: "vm" (default, in-process virtual
+                # machine) or "aot" (compile to a shared library via the
+                # system C compiler; see :mod:`pybamm.codegen.compilation`).
+                "compilation": "vm",
```
aabills: should we rename this to something more intuitive? Like just "compile": "true", "false"

aabills (changes requested):

src/pybamm/expression_tree/averages.py
MarcBerliner: Yeah, I put the averaging in this PR is because the SPMe voltage expression uses a ton of integral averages, whose evaluations were super slow because of the convoluted symbolic expression tree. I took a stab at simplifying some of the averaging code and extending to y/z domains, but I'm not familiar enough with the meshes to fully remove the hardcoded x domains

aabills (approved):
Lgtm

PR #5451, opened 2026-04-21
Fix DomainError for hysteresis/MSMR OCP with particle size distribution and thermal

rtimms:
## Summary

Combining a hysteresis / MSMR / single open-circuit-potential submodel with `"particle size": "distribution"` and a non-isothermal `thermal` option previously raised a `DomainError` when the thermal submodel built its hysteresis-heating term.

### MWE

```python
import pybamm
options = {
    "open-circuit potential": "one-state hysteresis",
    "particle size": "distribution",
    "surface form": "algebraic",
    "thermal": "lumped",
}
pybamm.lithium_ion.SPM(options)
# pybamm.expression_tree.exceptions.DomainError: children must have same or empty domains,
#   not ['positive electrode'] and ['positive particle size']
```

### Root cause

`"Positive electrode open-circuit potential [V]"` is correctly size-averaged onto `['positive electrode']` inside `BaseOpenCircuitPotential._get_standard_ocp_variables`. But `"Positive electrode equilibrium open-circuit potential [V]"` and `"Positive electrode OCP hysteresis [V]"` were published directly from expressions still on `['positive particle size']` under PSD. `base_thermal._get_standard_coupled_variables` then computed `V_hys = U - U_eq`, which triggered the mismatch (subtraction for hysteresis; concatenation for MSMR).

A latent bug was also fixed: `U_eq_x_av` / `H_x_av` were assigned the same expression as `U_eq` / `H` — never x-averaged.

### Fix

- `base_hysteresis_ocp.py`: under PSD, publish distribution-named variables at the particle-size shape and size-average `U_eq` / `H` before publishing the electrode-level names. Properly x-average the `_x_av` versions.
- `single_ocp.py` and `msmr_ocp.py`: after `_get_standard_ocp_variables` has produced the properly shaped main OCP variables, reuse them for the equilibrium OCP (which is equal to the OCP when there is no hysteresis). This avoids duplicating shape logic and keeps domains consistent.

### Tests

Unit regression tests added in `base_lithium_ion_tests.py` (inherited by SPM / SPMe / DFN / Newman-Tobias) and `test_mpm.py`:
- `test_well_posed_psd_hysteresis_thermal`
- `test_well_posed_psd_single_ocp_thermal` (also pins `equilibrium OCP primary == ['positive electrode']`)
- `test_well_posed_psd_msmr_thermal` (skipped on Newman-Tobias — pre-existing, unrelated DomainError in MSMR Butler-Volmer kinetics)
- `test_one_state_hysteresis_ocp_lumped_thermal` (MPM)

Integration regression test added (SPM / SPMe / DFN via base class):
- `test_psd_hysteresis_thermal` — full solve via `StandardModelTest.test_all(skip_output_tests=True)` with Chen2020 + `get_size_distribution_parameters` and hysteresis overrides.

No integration test for MSMR + PSD + thermal: `MSMR_Example` is an OCP-only parameter set and does not ship the thermal / geometry parameters a lumped thermal run needs. The unit test already pins the fix.

## Test plan

- [x] MWE builds cleanly after fix.
- [x] New unit tests pass across SPM, SPMe, DFN, MPM.
- [x] New integration tests pass across SPM, SPMe, DFN.
- [x] Full lithium-ion unit suite: 828 passed, 12 skipped (all pre-existing skips), 0 failures.
- [x] Existing MPM hysteresis integration tests still green.


MarcBerliner (dismissed):
Looks good, thanks!

BradyPlanden (approved):

PR #5459, opened 2026-04-23
Add pybammsolvers nonlinear solver

MarcBerliner:
# Description

Replaces `CasadiAlgebraicSolver` as the default nonlinear solver. This fixes an issue where the algebraic solver could terminate early when hitting step change tolerances but not the desired absolute tolerance. The new `NonlinearSolver` is tightly integrated with sundials to satisfy the same error tolerance and avoid expensive calls to `IDACalcIC` where only analgebraic sub-solver is necessary

## Type of change

# Important checks:


MarcBerliner:
@martinjrobins on my machine, this fixes the `lithium-plating-composite.ipynb` initialization issue with unified experiments that you mentioned in https://github.com/pybamm-team/PyBaMM/pull/5422#discussion_r2983552952

pyproject.toml
```diff
@@ -27,7 +27,7 @@ classifiers = [
     "Topic :: Scientific/Engineering",
 ]
 dependencies = [
-    "pybammsolvers>=0.6.0,<0.7.0",
```
BradyPlanden, line 30: issue: I think it's worth removing the upper pin on pybammsolvers.  I just realised we haven't been using `0.7.0` since it wasn't updated here. It probably better to fail loudly, then to silently forget about it.

src/pybamm/solvers/idaklu_solver.py
```diff
@@ -405,6 +455,22 @@ def set_up(self, model, inputs=None, t_eval=None, ics_only=False):
...
+            alg_res_fn = getattr(model, "algebraic_eval", None)
+            jac_alg_fn = getattr(model, "jac_algebraic_eval", None)
+        else:
+            alg_res_fn = None
+            jac_alg_fn = None
+
+        if alg_res_fn is None or not callable(getattr(alg_res_fn, "serialize", None)):
+            alg_res_fn = casadi.Function("empty_alg_res", [], [])
+        if jac_alg_fn is None or not callable(getattr(jac_alg_fn, "serialize", None)):
+            jac_alg_fn = casadi.Function("empty_alg_jac", [], [])
```
BradyPlanden: suggestion (non-blocking): I believe this can be compressed into a single if-else statement + a helper.

tests/unit/test_experiments/test_simulation_with_experiment.py
```diff
@@ -864,35 +865,53 @@ def test_run_experiment_cccv_solvers(self):
         experiment_2step = pybamm.Experiment(
             [
                 (
-                    "Discharge at C/20 for 1 hour",
-                    "Charge at 1 A until 4.1 V",
-                    "Hold at 4.1 V until C/2",
-                    "Discharge at 2 W for 1 hour",
+                    "Discharge at C/10 for 1 hour",
```
BradyPlanden, line 868: question: was there an issue with performance or coverage on the previous experiment?

BradyPlanden (commented):
This looks great - thanks @MarcBerliner! Just a few comments/questions

pyproject.toml
Saransh-cpp: Blocking https://github.com/conda-forge/pybamm-feedstock/issues/65 - can only put PyBaMM on conda-forge back if a release version depends on v0.7.0+, and surprisingly we don't use pybammsolvers v0.7.0?

tests/unit/test_experiments/test_simulation_with_experiment.py
MarcBerliner: The previous version of this test was ostensibly a CC-CV test (which is tested elsewhere), but is really a 1:1 comparison check between casadi and IDAKLU in disguise. The state-dependent termination conditions caused slightly different end times from solver to solver, and the array checks were frequently flaky and had very loose tolerances. Checking the y values is more reliable than just checking a select few states/outputs, which we can now do with tighter tols

pyproject.toml
MarcBerliner: I would selfishly like to have some kind of upper pin on pybammsolvers (could only be for major releases like `<1") because we effectively use pybammsolvers as an internal dependency, and changing any API in pybammsolvers is a huge headache if we have to consider backward compatibility across multiple pybamm versions

Saransh-cpp:
Hi @MarcBerliner, given that this is related to https://github.com/pybamm-team/pybammsolvers/pull/99. Could you please check - https://github.com/conda-forge/pybammsolvers-feedstock/pull/8#pullrequestreview-4197669873:

> https://github.com/pybamm-team/pybammsolvers/blob/b8628c230916df6bad1dade003259bd72f63bda1/CMakeLists.txt#L233-L238
> 
> pybammsolvers started using internal SUNDIALS headers that are not installed by the SUNDIALS by default (and subsequently, by the distribution on conda-forge - https://sundials.readthedocs.io/en/v6.6.1/idas/Introduction_link.html#changes-in-v3-1-0.
> 
> We have two options:
>
> 1. Back to square one - build SUNDIALS in our CI (which I really don't want, but it might be the only fix available)
> 2. The release notes linked above say "The implementation header file idas_impl.h is no longer installed. This means users who are directly manipulating the IDAMem structure will need to update their code to use IDAS’s public API." which means that we can use IDAS' public API to achieve what we achieve using the internal headers?

More specifically the second point, is it possible to add your feature (to PyBaMM) without using the internal headers? Would help me in fixing the `v0.8.0` build on conda-forge. Thanks!

MarcBerliner:
> More specifically the second point, is it possible to add your feature (to PyBaMM) without using the internal headers? Would help me in fixing the `v0.8.0` build on conda-forge. Thanks!

@Saransh-cpp I think we can do this. I can work on this today as a followup in pybammsolvers/bump in pybamm

pyproject.toml
MarcBerliner: Let's keep this pin setup for now and discuss this topic at the next pybamm dev meeting

BradyPlanden (approved):
Thanks!

PR #5478, opened 2026-04-30
Fix PotentialPair1plus1D/2plus1D normalisation for parallel electrodes (#5467)

### LINKED ISSUE #5467: [Bug]: PotentialPair1plus1D not normalising correctly parallel cells
### PyBaMM Version

26.4.1

### Python Version

3.12.12

### Describe the bug

Class `PotentialPair1plus1D` , i.e. `options={"current collector": "potential pair", "dimensionality": 1,},` does not account for number of cells in parallel when computing current density. This leads to incorrect capacity calculations.

Using same parameters with conventional P2D model and standard options returns correct capacity.

### Steps to Reproduce

[1D_pouch_bug.py](https://github.com/user-attachments/files/27137503/1D_pouch_bug.py)

### Relevant log output

```shell
Base DFN - 4.11 x 0.414 x 1 - Discharge capacity: 40.83695999999995 A.h
Base DFN - 4.11 x 0.207 x 2 - Discharge capacity: 40.83695999999995 A.h
1D pouch cell DFN - 4.11 x 0.414 x 1 - Discharge capacity: 40.83696000000003 A.h
2026-04-27 20:23:32.272 - [WARNING] callbacks.on_experiment_infeasible_event(254): 

        Experiment is infeasible: 'event: Minimum voltage [V]' was triggered during 'Discharge at 0.01C for 100 hour'. The returned solution only contains up to step 1 of cycle 1. 
1D pouch cell DFN - 4.11 x 0.207 x 2 - Discharge capacity: 22.3637893208887 A.h
```

kamal20122012:
## Summary
- The 1+1D and 2+1D `PotentialPair` current-collector submodels apply a Neumann BC at the positive tab of a single modelled electrode pair, but the boundary condition was computed from the **total cell current** (`applied_current_density * A_cc`) rather than the **per-electrode-pair current**. Because `A_cc` already includes `n_electrodes_parallel`, the gradient of `phi_s_cp` at the tab was overstated by that factor whenever more than one electrode pair was connected in parallel — voltage drops in the current collector were inflated, the lower voltage cutoff was hit prematurely, and the reported discharge capacity was approximately `1 / n_electrodes_parallel` of the correct value.
- Fix: divide by `n_electrodes_parallel` so the BC uses the current per modelled electrode pair. No effect for the default `n=1` case.
- Added an integration regression test that compares discharge capacity between an equivalent wide single-electrode configuration and a two-parallel-electrode configuration with the same total current-collector area; capacities now match.

Fixes #5467

## Test plan
- [x] Reproduction script from the issue: parallel config now reports **40.84 A·h** (matches wide single-electrode case) instead of 22.36 A·h.
- [x] `pytest tests/integration/test_models/test_full_battery_models/test_lithium_ion/test_dfn.py::TestDFN::test_potential_pair_1plus1D_parallel_electrodes` (passes; verified to fail on pre-fix code)
- [x] `pytest tests/unit/test_models/test_full_battery_models/test_lithium_ion -k "1plus1D or 2plus1D or potential_pair"` — 50 passed
- [x] `pre-commit run --files <changed files>` — all hooks pass (ruff-check, ruff-format, blacken-docs, file-checks, sp-repo-review)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 - Added `NonlinearSolver` as the default nonlinear solver, which replaces `CasadiAlgebraicSolver`. `IDAKLUSolver` now computes the initial conditions in C++ by default. ([#5459](https://github.com/pybamm-team/PyBaMM/pull/5459))
 
+## Bug fixes
+
+- Fixed `PotentialPair1plus1D` and `PotentialPair2plus1D` current collector models so that the positive-tab Neumann boundary condition uses the per-electrode current rather than the total cell current. Previously, configurations with `"Number of electrodes connected in parallel to make a cell"` greater than 1 hit voltage cutoffs prematurely and reported reduced discharge capacity. ([#5467](https://github.com/pybamm-team/PyBaMM/issues/5467))
+
 # [v26.4.1](https://github.com/pybamm-team/PyBaMM/tree/v26.4.1) - 2026-04-24
 
 ## Bug fixes
--- a/src/pybamm/models/submodels/current_collector/potential_pair.py
+++ b/src/pybamm/models/submodels/current_collector/potential_pair.py
@@ -77,13 +77,20 @@ class PotentialPair1plus1D(BasePotentialPair):
         phi_s_cp = variables["Positive current collector potential [V]"]
 
         applied_current_density = variables["Total current density [A.m-2]"]
-        total_current = applied_current_density * self.param.A_cc
+        # The 1+1D model represents a single electrode pair, so the current entering
+        # the modelled stack is the total cell current divided by the number of
+        # electrodes connected in parallel.
+        current_per_electrode_pair = (
+            applied_current_density * self.param.A_cc / self.param.n_electrodes_parallel
+        )
 
         # In the 1+1D model, the behaviour is averaged over the y-direction, so the
         # effective tab area is the cell width multiplied by the current collector
         # thickness
         positive_tab_area = self.param.L_y * self.param.p.L_cc
-        pos_tab_bc = -total_current / (self.param.p.sigma_cc * positive_tab_area)
+        pos_tab_bc = -current_per_electrode_pair / (
+            self.param.p.sigma_cc * positive_tab_area
+        )
 
         # Boundary condition needs to be on the variables that go into the Laplacian,
         # even though phi_s_cp isn't a pybamm.Variable object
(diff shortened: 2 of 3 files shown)


src/pybamm/models/submodels/current_collector/potential_pair.py
```diff
@@ -77,13 +77,20 @@ def set_boundary_conditions(self, variables):
         phi_s_cp = variables["Positive current collector potential [V]"]
 
         applied_current_density = variables["Total current density [A.m-2]"]
-        total_current = applied_current_density * self.param.A_cc
+        # The 1+1D model represents a single electrode pair, so the current entering
```
BradyPlanden, line 80: issue: please remove the 'how' comments across this PR.

tests/integration/test_models/test_full_battery_models/test_lithium_ion/test_dfn.py
```diff
@@ -60,6 +60,53 @@ def test_cycling_extreme_conditions(self):
 
         assert sol.termination == "final time"
 
+    def test_potential_pair_1plus1D_parallel_electrodes(self):
+        # Regression test for #5467: the 1+1D potential-pair current collector
+        # model represents a single electrode pair, so its boundary condition
+        # must use the per-electrode current. With multiple parallel electrodes,
+        # the cell discharge capacity should match that of a single wide electrode
+        # with the same total current-collector area.
+        model = pybamm.lithium_ion.DFN(
```
BradyPlanden, line 69: suggestion: let's move this to either the potential pair related tests or the SPM/SPMe suite. It doesn't need to be a DFN, so let's save the computation time.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 - Added `NonlinearSolver` as the default nonlinear solver, which replaces `CasadiAlgebraicSolver`. `IDAKLUSolver` now computes the initial conditions in C++ by default. ([#5459](https://github.com/pybamm-team/PyBaMM/pull/5459))
 
+## Bug fixes
+
+- Fixed `PotentialPair1plus1D` and `PotentialPair2plus1D` current collector models so that the positive-tab Neumann boundary condition uses the per-electrode current rather than the total cell current. Previously, configurations with `"Number of electrodes connected in parallel to make a cell"` greater than 1 hit voltage cutoffs prematurely and reported reduced discharge capacity. ([#5467](https://github.com/pybamm-team/PyBaMM/issues/5467))
```
BradyPlanden, line 9: issue: please update this to match the changelog convention of PR link.

BradyPlanden (changes requested):
Thanks @kamal20122012 - a few comments

PR #5484, opened 2026-05-01
fix: stack state mapper p using previous model inputs in experiment steps

aabills:
Simulation._build_experiment_state_mappers compiles mappers from previous_model.input_parameters; BaseSolver.step must pass the same p layout when calling the compiled CasADi mapper.

Fixes CasADi shape errors when the next step model has more InputParameters than the previous (e.g. scalar Current then piecewise Current with start time).

- Regression test: SPM + SEI input, scalar then array Current
- Align CV initial-guess test with previous-model p stacking

Made-with: Cursor

# Description

Fixes # (issue)

## Type of change

# Important checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 - Added `NonlinearSolver` as the default nonlinear solver, which replaces `CasadiAlgebraicSolver`. `IDAKLUSolver` now computes the initial conditions in C++ by default. ([#5459](https://github.com/pybamm-team/PyBaMM/pull/5459))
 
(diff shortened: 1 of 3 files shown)


martinjrobins (approved):
looks good, thanks @aabills! how does test_state_mapper_previous_step_input_layout_scalar_then_array_current work if you use the unified model?

MarcBerliner:
> how does test_state_mapper_previous_step_input_layout_scalar_then_array_current work if you use the unified model?

Unified fails to build since it doesn't support drive cycles

PR #5485, opened 2026-05-01
Fix load_custom_model to walk recorded MRO instead of dropping to BaseModel

rtimms:
## Description

`Serialise.load_custom_model` previously fell back to `pybamm.BaseModel` whenever the recorded `base_class` could not be imported (e.g. a user subclass defined in a package not installed on the loader). This is fragile: bare `pybamm.BaseModel` has an empty `default_spatial_methods`, so the loaded model carries the right symbolic equations but fails downstream at discretisation in a way that is hard to trace back to the silent fallback.

This PR records the full MRO at serialise time and walks it at load time, resolving to the closest importable ancestor (typically the lithium-ion `BaseModel`) so subclass defaults are preserved. `pybamm.BaseModel` is still the last-resort fallback if nothing in the MRO is importable. Legacy payloads written before this change continue to load.

## Type of change

## Key changes

- `serialise_custom_model` now writes a `base_class_mro` field — the model class's full ancestor chain, fully-qualified `module.ClassName` strings, with `object` and other `builtins` entries filtered out.
- `load_custom_model` delegates resolution to a new `Serialise._resolve_base_class` helper that tries `base_class` first, then walks `base_class_mro`, and only falls back to `pybamm.BaseModel` if nothing imports.
- New helper `Serialise._import_dotted_class` performs the dotted-path import.
- The fallback warning now identifies the ancestor used (or that the final fallback to `pybamm.BaseModel` was taken).
- Schema version is unchanged; `base_class_mro` is optional on read so older payloads still load.

## Why this matters

Custom models are commonly serialised on a developer machine and loaded on a separate backend that has only `pybamm` installed. Under the previous behaviour, a subclass of the lithium-ion `BaseModel` would silently load as `pybamm.BaseModel` and fail at discretisation with a confusing error about missing spatial methods on the `positive particle` domain. The new behaviour preserves the closest pybamm-provided ancestor and discretisation proceeds correctly.

## Testing

New tests in `tests/unit/test_serialisation/test_serialisation.py`:

- `test_serialise_records_base_class_mro` — the new field is present, includes `pybamm`-provided ancestors, and excludes `builtins`.
- `test_load_falls_back_to_mro_ancestor_when_base_class_unimportable` — payload pointing at an unimportable `base_class` resolves to the lithium-ion `BaseModel` via the MRO and the loaded model has `positive particle` in `default_spatial_methods`.
- `test_load_falls_back_to_base_model_when_no_mro_entry_importable` — when no MRO entry can be imported, `pybamm.BaseModel` is still used (legacy behaviour).
- `test_load_legacy_payload_without_base_class_mro` — payloads predating this change still load.

The existing `test_import_base_class_non_builtin_object` still passes: its corruption only mutates `base_class`, so the MRO walk finds the original `pybamm.models.base_model.BaseModel` entry.

All 178 tests in `tests/unit/test_serialisation/` pass locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 - Added `NonlinearSolver` as the default nonlinear solver, which replaces `CasadiAlgebraicSolver`. `IDAKLUSolver` now computes the initial conditions in C++ by default. ([#5459](https://github.com/pybamm-team/PyBaMM/pull/5459))
 
+## Bug fixes
+
+- `Serialise.load_custom_model` no longer silently drops to `pybamm.BaseModel` when the recorded `base_class` cannot be imported. `serialise_custom_model` now records the full ancestor chain in a new `base_class_mro` field, and the loader walks it to resolve to the closest importable ancestor (typically a pybamm-provided class such as the lithium-ion `BaseModel`). This preserves subclass defaults — including `default_spatial_methods` — so loaded models discretise correctly on machines that do not have the original subclass package installed. Falling back to `pybamm.BaseModel` is still the last resort if no MRO entry can be imported.
+
 # [v26.4.1](https://github.com/pybamm-team/PyBaMM/tree/v26.4.1) - 2026-04-24
 
 ## Bug fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ## Bug fixes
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 - Added `NonlinearSolver` as the default nonlinear solver, which replaces `CasadiAlgebraicSolver`. `IDAKLUSolver` now computes the initial conditions in C++ by default. ([#5459](https://github.com/pybamm-team/PyBaMM/pull/5459))
 
+## Bug fixes
+
+- `Serialise.load_custom_model` no longer silently drops to `pybamm.BaseModel` when the recorded `base_class` cannot be imported. `serialise_custom_model` now records the full ancestor chain in a new `base_class_mro` field, and the loader walks it to resolve to the closest importable ancestor (typically a pybamm-provided class such as the lithium-ion `BaseModel`). This preserves subclass defaults — including `default_spatial_methods` — so loaded models discretise correctly on machines that do not have the original subclass package installed. Falling back to `pybamm.BaseModel` is still the last resort if no MRO entry can be imported.
```
BradyPlanden: suggestion: Can we shrink this to the standard one line high-level comment

BradyPlanden (dismissed):

BradyPlanden (approved):

PR #5487, opened 2026-05-01
fix: EISSimulation.solve must forward inputs to build

rtimms:
## Summary

`EISSimulation.solve(initial_soc=..., inputs=...)` rebuilds the simulation for the requested SOC by calling `self.build(initial_soc=...)`, but it did not pass `inputs` through. Inside `set_initial_state`, the half-cell ESOH path evaluates `Q_init` (which depends on the active material volume fraction). When a parameter value is a symbolic expression that references an `InputParameter` — e.g. a linked-parameter expression `(1 - porosity) * pybamm.InputParameter("vf_solid")` for the active material volume fraction — evaluation with `inputs=None` raises:

```
ValueError: symbol must evaluate to a constant scalar or array
```

This patch forwards `inputs` to `build` so the ESOH path can substitute them.

## Change

Single line in `src/pybamm/simulation/eis_simulation.py`:

```diff
 if initial_soc is not None:
-    self.build(initial_soc=initial_soc)
+    self.build(initial_soc=initial_soc, inputs=inputs)
 elif self._built_model is None:
     self.build()
```

## Test plan

- [x] Added `test_initial_soc_forwards_inputs` to `tests/unit/test_eis_simulation.py` covering both numeric SOC and voltage-string SOC paths with a half-cell SPM whose active material volume fraction is `(1 - porosity) * pybamm.InputParameter("vf_solid")`. Fails on `main` with the exact `"symbol must evaluate to a constant scalar or array"` error; passes with the fix.
- [x] Existing EIS tests continue to pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ## Bug fixes
 
 - Fixed legacy experiment stepping: the precalculated state mapper is evaluated with the *previous* step model's input layout, not the next model's. Fixes CasADi mapper shape errors when transitioning from a scalar `Current` step to a piecewise (array) `Current` step with additional `InputParameter`s.
+- Fixed `EISSimulation.solve(initial_soc=..., inputs=...)` not forwarding `inputs` to `build`.
 
 # [v26.4.1](https://github.com/pybamm-team/PyBaMM/tree/v26.4.1) - 2026-04-24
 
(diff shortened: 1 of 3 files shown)


BradyPlanden (approved):
Ah, classic. Thanks!

PR #5498, opened 2026-05-06
Add generic JSON round-trip property tests for the Serialise API

rtimms:
## Summary

- Adds a parametrised JSON round-trip property test, one fixture per concrete subclass, covering **`BaseSolver`, `Experiment`, `ParameterValues`, `SpatialMethod`, `MeshGenerator`, `SubMesh`, built-in `BaseModel`, custom `BaseModel`, and every concrete `Symbol` type** (Scalar, Parameter, InputParameter, Variable, Time, SpatialVariable, binary/unary operators, broadcasts, Interpolant, FunctionParameter). Each test reads every `__init__`-derived attribute on the original and the round-tripped instance and compares them with bool-strict, float-approx, recurse-into-`BaseSolver` equality.
- Closes the structural gap behind the recent run of serialisation bugs (#5495 bool→int coercion, #5496 dropped per-step `temperature`, #5497 dropped nested `root_method`). All four bugs share one shape: a value reachable from a constructor argument was silently lost on JSON round-trip, and no test asserted "after round-trip, the same constructor-arg state survives".
- **The expanded test surfaced four more instances of the same bug class**, all fixed in this PR:
  1. `Serialise.load_custom_model` reconstructed `Event.event_type` as the bare string `"TERMINATION"` instead of `EventType.TERMINATION`, because the custom JSON encoder writes Enum values as their `.name` and the loader passed that string straight to `pybamm.Event.__init__` without converting it back.
  2. `Serialise.serialise_experiment` / `deserialise_experiment` dropped every top-level `pybamm.Experiment` argument other than `operating_conditions` — `period`, `temperature`, and `termination` were never written and never read, so a JSON round-trip silently reset them to defaults.
  3. The same dump/load pair dropped every per-step `BaseStep` argument other than `value` / `duration` / `terminations` / `temperature` — `period`, `tags`, `description`, `start_time`, `direction`, and `skip_ok` were lost. `start_time` is now encoded as ISO 8601 and decoded with `datetime.fromisoformat`.
  4. The `Resistance` step type was missing from the deserialiser's `step_func_map` (only present in the serialiser's `step_type_map`), so any Resistance step round-tripped to a `ValueError`.
- Going forward, adding a new init arg to a serialisable class means adding a fixture; the fixture either round-trips cleanly or the serialise path needs a new handler. Moves the maintenance burden from "remember to update three places in `serialise.py`" to "add a fixture when you add a feature".

## Validation

The property test was validated to catch each of the four known bugs by reverting the corresponding fix locally and confirming the test fails on the right fixture(s):

| Reverted fix | Fixtures that fail |
| --- | --- |
| #5495 (bool→int) | 5 solver fixtures (any solver with bool init args — IDAKLU options + CasadiSolver \`return_solution_if_failed_early\` etc.) |
| #5496 (dropped step \`temperature\`) | the temperature-bearing experiment fixture |
| #5497 (dropped \`root_method\`) | 5 solver fixtures (any solver with \`root_method\`) |
| EventType bug (this PR) | the events-bearing custom model fixture |
| Top-level Experiment kwargs (this PR) | the top-level-kwargs experiment fixture |
| Per-step BaseStep kwargs (this PR) | the per-step-kwargs experiment fixture |
| Resistance step type (this PR) | the resistance-step experiment fixture |

With all fixes in place: `46 passed`. Full serialisation suite: `233 passed`.

## Test plan

- [x] \`pytest tests/unit/test_serialisation/test_round_trip.py\` (46 passed)
- [x] \`pytest tests/unit/test_serialisation/\` (233 passed total — 187 existing + 46 new)
- [x] Locally reverted each known fix in turn, confirmed the corresponding fixtures fail with clear assertion messages, restored.
- [x] Confirmed each of the four newly-fixed bugs (EventType, top-level experiment kwargs, per-step BaseStep kwargs, Resistance) fails the relevant fixture before the fix and passes after.

## Follow-ups (separate PRs)

- Replace warn+continue with raise TypeError in \`Serialise.serialise_solver\` (after this PR's fixtures all pass on main, low-risk).
- Collapse \`_serialise_step\` / \`_parse_step\` into delegation through \`BaseStep.to_dict\`, removing the field-list duplication that caused #5496 and the new per-step drops.


tests/unit/test_serialisation/test_round_trip.py
BradyPlanden, line 1: suggestion (non-blocking): I'm having a bit of difficult following the code flow in this file. Could it be class based with fixtures located at the top?

BradyPlanden (dismissed):
Looks good, just one comment

rtimms:
Pushed 041c4864d addressing the five review comments.

1. **Omitted `duration` round-trips.** `_serialise_step` now uses `step.input_duration` and only emits `"duration"` when not `None`; `_parse_step` only forwards it when present. So a step constructed with `duration=None` (e.g. `current(1.0, termination="3.0 V")`) now comes back with `uses_default_duration=True`. New fixture covers this and `_STEP_ATTRS` was extended with `input_duration` and `uses_default_duration` so `_experiments_equal` actually verifies it.

2. **Strict `skip_ok` validation.** Replaced `bool(step_dict["skip_ok"])` with an `isinstance(value, bool)` check that raises `TypeError`. New `TestExperimentFromConfigValidation` covers `"False"` rejection plus JSON-bool round-trip in both directions.

3. **`CustomTermination` / custom steps refused cleanly.** Serialiser now raises `NotImplementedError` with a useful message for `CustomTermination`, `CustomStepExplicit`/`CustomStepImplicit`, and any unknown step / termination class, rather than crashing inside `term.value` or `step.input_value.tolist()`. `serialise_experiment` docstring spells the limitation out. New `TestExperimentSerialiseUnsupported` covers both shapes.

4. **Custom-model assertions tightened.** `TestCustomModelRoundTrip.test_preserves_rhs_and_events` now compares `{var.id: expr.id}` for both `rhs` and `initial_conditions`, and pairs events by name with `expression.id` equality. Fixtures now use multi-term expressions (`-c*c + 0.5`, `c - 0.5`) so a sign flip or expression mutation actually breaks the structural id.

5. **Canonical JSON output.** `_serialise_step` now skips `period` / `temperature` when they equal the experiment-level defaults (broadcast by `Experiment.process_steps`), drops empty `tags`, and drops computed `direction` for step types with `calculate_charge_or_discharge=True`. Top-level defaults appear once at the top of the config, not duplicated onto every step. Voltage steps still keep an explicit user-supplied `direction`. New `TestExperimentSerialiseCanonical` asserts the JSON shape.

Verified: `tests/unit/test_serialisation` (242 passed), `tests/unit/test_experiments` (135 passed), `tests/unit/test_simulation.py` (75 passed), and the new property-test fixtures all pass; pre-commit clean.

BradyPlanden (dismissed):
Looks good, thanks!

BradyPlanden (approved):

PR #5499, opened 2026-05-07
feat: add store_first_last solver kwarg for memory-light long experiments

rtimms:
## Summary

- Adds a `store_first_last: bool = False` kwarg to `BaseSolver` (forwarded through `IDAKLUSolver`, `CasadiSolver`, `ScipySolver`, `JaxSolver`).
- When `True`, only the first and last sample of each integration window are stored: one experiment step in `Simulation.solve`, or the full `[t_eval[0], t_eval[-1]]` window in plain `BaseSolver.solve`.
- Composes with `output_variables` for memory-light long ageing experiments whose post-processing only reads per-step first/last values (e.g. metrics built on per-cycle/per-step `First` / `Last` aggregators).
- Has effect on solvers that support intra-solve interpolation (`IDAKLUSolver`); other solvers warn once per instance and no-op.

The flag is intended for the case where post-processing only needs first/last state. With this flag set, intra-step interpolation falls back to linear across the whole step (IDAKLU's Hermite checkpoints are auto-disabled by the existing `t_interp` non-empty path), so it is **not** appropriate when post-processing queries an intra-step time (e.g. a `RelativeTime`-style sample).

## Implementation notes

- Centralised in `BaseSolver._store_first_last_endpoints`, called from both `solve()` (collapses `t_eval` so IDAKLU stops only at endpoints) and `process_t_interp()` (sets `t_interp` to the same endpoints).
- `BaseStep._setup_timestepping` short-circuits to `t_interp = [0, tf]` when the flag is set, overriding any per-step `period` and any caller-provided `t_interp`.
- Round-trips through `Serialise.serialise_solver` automatically (reflected from the `__init__` signature).

## Test plan

- [x] Plain `solver.solve(model, t_eval=...)` returns 2 stored samples on IDAKLU.
- [x] Each experiment step's sub-solution has exactly 2 stored samples; end-of-step values match a reference solve.
- [x] Per-step `period` and caller `t_interp` are overridden when the flag is on.
- [x] `output_variables` composes (variables-only × endpoints-only).
- [x] Non-IDAKLU solvers (Casadi) warn once per instance and integrate normally.
- [x] `to_config()` / `from_config()` round-trips for IDAKLU and Casadi.


src/pybamm/experiment/step/base_step.py
```diff
@@ -355,6 +355,10 @@ def _setup_timestepping(self, solver, tf, t_interp):
         else:
             t_eval = np.array([0, tf])
 
+        # `getattr` because some test/duck-typed solvers don't subclass BaseSolver.
```
BradyPlanden: suggestion: Can we instead update the tests/solvers to avoid having different functionality for each attr. I.e. t_interp, t_eval, etc. are guaranteed, let's make store_first_last as well.

src/pybamm/solvers/idaklu_solver.py
```diff
@@ -59,6 +59,18 @@ class IDAKLUSolver(pybamm.BaseSolver):
     output_variables : list[str], optional
         List of variables to calculate and return. If none are specified then
         the complete state vector is returned (can be very large) (default is [])
+    store_first_last : bool, optional
+        If True, only the first and last sample of each integration window are
+        stored (one experiment step in :meth:`pybamm.Simulation.solve`, or the
+        full ``[t_eval[0], t_eval[-1]]`` window in :meth:`solve`). Composes
```
BradyPlanden: docs: 'Composes with `output_variables`' is used in a few places on this PR. It's not clear initially if this refers to combined usage or if we enable `output_variables` when the user enables `store_first_last`. I think we should just remove the reference to `output_variables` in the parameter definition.

(non-blocking): we add an example using both options together to reduce memory allocation. Perhaps updating the [06-output-variables](https://github.com/pybamm-team/PyBaMM/blob/main/docs/source/examples/notebooks/performance/06-output-variables.ipynb) notebook

tests/unit/test_solvers/test_store_first_last.py
```diff
@@ -0,0 +1,158 @@
...
+import warnings
+
+import numpy as np
+import pytest
+
+import pybamm
+
+
+def _build_simple_dae_model():
+    """Tiny DAE used by base-solver-level tests (avoids battery-model overhead)."""
```
BradyPlanden:
```suggestion
    """Tiny DAE used by base-solver-level tests."""
```

src/pybamm/solvers/base_solver.py
```diff
@@ -71,6 +84,10 @@ def __init__(
         self.root_method = root_method
         self.extrap_tol = extrap_tol or -1e-10
         self.output_variables = [] if output_variables is None else output_variables
+        if not isinstance(store_first_last, bool):
```
BradyPlanden, line 86: suggestion (non-blocking): we add types to the init args and accept that users should read IDE suggested type. Alternatively, we could move to a `SolverOptions` dataclass/enum to avoid needing to check the types of the args explicitly individually.

src/pybamm/solvers/base_solver.py
```diff
@@ -908,7 +925,14 @@ def solve(
         if (np.diff(t_eval) < 0).any():
             raise pybamm.SolverError("t_eval must increase monotonically")
 
-        t_interp = self.process_t_interp(t_interp)
+        # The integrator stops at every t_eval point, so collapsing it is what
+        # actually reduces the stored output count in plain solve() (step() is
+        # already a 2-point window).
```
BradyPlanden:
```suggestion
```

tests/unit/test_solvers/test_store_first_last.py
```diff
@@ -0,0 +1,158 @@
...
+        solver = pybamm.IDAKLUSolver(store_first_last=True)
+        sim = pybamm.Simulation(model, experiment=experiment, solver=solver)
+        sol = sim.solve()
+
+        assert len(sol.sub_solutions) == 3
+        for sub in sol.sub_solutions:
+            assert sub.t.size == 2
+
+        # End-of-step values should still match a reference solve at a
+        # loose tolerance (Hermite is disabled here so this is approximate).
```
BradyPlanden: suggestion: This should be within the solver precision, since both case are provided as `t_eval` points, right?

BradyPlanden (changes requested):
This is looking good, just a few comments on docs / alignment.

src/pybamm/solvers/idaklu_solver.py
rtimms: I've removed this from the docstring and updated the notebook

tests/unit/test_solvers/test_store_first_last.py
rtimms: yes. fixed

src/pybamm/solvers/base_solver.py
rtimms: I've left this as-is here and made an issue to implement `SolverOptions`, see https://github.com/pybamm-team/PyBaMM/issues/5506

BradyPlanden (approved):
Nice one! The notebook results are great 🚀

PR #5502, opened 2026-05-08
Set closest_event_idx in IDAKLUSolver after IDA_ROOT_RETURN

rtimms:
## Summary

After a SUNDIALS root return, `BaseSolver.get_termination_reason` needs to label `solution.termination` with the specific event that fired. The casadi solver records this in `Solution.closest_event_idx` and `get_termination_reason` short-circuits. IDAKLU never set it, so the slow fallback path re-walked every TERMINATION event's symbolic expression on the Python side every step — dominating per-step allocations on long event-terminated cycling runs.

The combined-events root function (`rootfn`) is already built and serialised for the C++ wrapper. This PR keeps a Python-callable handle on it in `self._setup["rootfn_casadi"]` so `_post_process_solution` can evaluate it once at `(t_event, y_event, p_stacked)` and set `closest_event_idx = nanargmin(|values|)`. Round-trip through `__getstate__`/`__setstate__` is supported by deserialising from the same `rootfn_pkl` the C++ wrapper consumes.

Also extracts the duplicated inputs-dict-flattening pattern in `_integrate` into a small `_flatten_inputs` helper, reused at both call sites.

## Where the allocations were

memray on a 1000-cycle SPM (`output_variables` set, IDAKLUSolver, `save_at_cycles=N` so most cycles are discarded):

| Frame                                                       | Baseline | With fix |
|-------------------------------------------------------------|----------|----------|
| `StateVector._base_evaluate` (state_vector.py:299)          | 326,168  | 0        |
| `_matmul_vector` (scipy.sparse._compressed.py:395)          | 305,144  | 0        |

Both eliminated. Top-line:

| 1000-cycle SPM      | Wall    | Cumulative bytes | Cumulative allocs |
|---------------------|---------|------------------|-------------------|
| baseline            | 13.10 s | 1756 MB          | 4,431,707         |
| with fix            | 10.95 s | 1311 MB          | 3,978,320         |
| **saved**           | **16%** | **445 MB (25%)** | **453k (10%)**    |

## Test plan

- [x] New regression test `test_closest_event_idx_set_after_root_return` in `tests/unit/test_solvers/test_idaklu_solver.py` — asserts every event-terminated step has `closest_event_idx` populated AND that the index resolves to the same event name `step.termination` carries. Fails on `main` (always `None` for IDAKLU), passes after this change.
- [x] All 37 IDAKLU unit tests pass (was 36 + 1 new).
- [x] All 309 other solver unit tests pass.
- [x] All 9 `tests/memory/test_memory_leaks.py` tests pass.
- [x] All 212 simulation/experiment unit tests pass.


BradyPlanden (dismissed):
Nice find!

BradyPlanden (approved):

PR #5504, opened 2026-05-08
Defer ProcessedVariableComputed.initialise_* until first data read

rtimms:
## Summary

Long cycling experiments with `output_variables` and `save_at_cycles` build a fresh `ProcessedVariableComputed` per output, per step, per cycle — most cycles are then discarded by `save_at_cycles` without anyone reading the per-step PVCs. Building each PVC eagerly ran `np.concatenate / flatten / xr.DataArray` construction inside `__init__`, and `Solution.__add__` → `PVC._concat` re-ran the full build on every step-add. On a 500-cycle SPM run with 6 output_variables this was **38% of solve wall time**.

This PR splits dispatch from execution in `ProcessedVariableComputed.__init__`:
- `_resolve_initialise_method` returns `(method, dimensions)` for the variable's shape, without running anything.
- `_materialise()` runs the chosen `initialise_*` exactly once.
- `entries` and `_xr_data_array` become properties that materialise on first read; setters preserve back-compat with the existing `self.entries = X` writes inside each `initialise_*` method (no changes to those methods).

Per-step PVCs that nobody ever reads (the discarded-cycle case) now skip the heavy work entirely. PVCs whose data is actually read (e.g. cycle-summary boundary first/last states) materialise normally on the first access.

## Benchmarks

500-cycle SPM, 6 `output_variables`, `save_at_cycles=N`, `IDAKLUSolver`, measured against `main`:

| Metric                              | main      | with PR   | Delta       |
|-------------------------------------|-----------|-----------|-------------|
| solve wall time                     | 4.33 s    | 2.62 s    | **−39%**    |
| `PVC.__init__` total time           | 1.66 s    | 0.03 s    | −98%        |
| `PVC.__init__` % of solve           | 38.3%     | 1.2%      | —           |
| `PVC.__init__` call count           | 19,044    | 19,044    | unchanged   |
| cumulative allocations (memray)     | 1352 MB   | 1327 MB   | −2% / −131k allocs |

The `__init__` count is the same — every step still constructs PVCs — but each call is now nearly free (just records which method to run later). The heavy work only happens for PVCs whose data is actually read.

Cumulative byte savings via memray are modest because most of the headline number is import-time and C++ wheel boundary cost; the win is in eliminated CPU work (xarray init + numpy concatenate), which uses few but expensive allocations and dominated wall time.

## Test plan

- [x] New `tests/memory/test_memory_leaks.py::TestExperimentMemory::test_processed_variable_computed_initialise_is_lazy` — monkey-patches `unroll_0D` and asserts fewer than `num_cycles` calls during a 50-cycle solve. Eager init produces hundreds; lazy init produces a handful.
- [x] All 71 `test_processed_variable.py` tests pass.
- [x] All 309 other solver tests pass (excluding pre-existing JAX failures on main).
- [x] All 212 `test_simulation.py` + `test_experiments/` tests pass.
- [x] All 10 `tests/memory/` tests pass.
- [x] Pickle round-trip on a solved Solution preserves the lazy state correctly.


src/pybamm/solvers/processed_variable_computed.py
```diff
@@ -83,57 +83,81 @@ def __init__(
...
         ]:
-            self.initialise_3D()
-            return
+            return self.initialise_3D, 3
+
+        raise NotImplementedError(f"Shape not recognized for {base_var}")
+
+    def _materialise(self):
+        # Not thread-safe: concurrent first reads may run initialise twice.
+        # PVCs are constructed and read on the same thread in the standard
```
BradyPlanden: docs: can we remove 'PVC's' from the docs. I don't think it's a common acronym we use in the codebase.

src/pybamm/solvers/processed_variable_computed.py
```diff
@@ -83,57 +83,81 @@ def __init__(
...
+        if not self._initialised:
+            self._initialise_method()
+            self._initialised = True
+
+    @property
+    def entries(self):
+        self._materialise()
+        return self._entries
+
+    @entries.setter
```
BradyPlanden: suggestion: instead of using the `entries` setter to direct `self.entries` -> `self._entries`, let's just have each `initialise_*` write directly to `self._entries`. Then we can remove the setter completely.

src/pybamm/solvers/processed_variable_computed.py
```diff
@@ -83,57 +83,81 @@ def __init__(
...
+    def entries(self, value):
+        self._entries = value
+
+    @property
+    def _xr_data_array(self):
```
BradyPlanden, line 150: suggestion: same here, writing directly to `_xr_data_array_cache`

BradyPlanden (commented):
Nice find, looks good - just some comments on structuring the property/setters

BradyPlanden (approved):
Looks great - thanks!

Issue #5463, opened 2026-04-24
Support BPX v1.x


ejfdickinson:
### Description

Upgrade parameter import from BPX JSON format to [BPX v1.1](https://github.com/FaradayInstitution/BPX/releases/tag/v1.1.0), or later if released.

Affected modules:

- `pybamm/parameters/parameter_values.py`
- `pybamm/parameters/bpx.py`

### Motivation

BPX v1.1 was released in February. Since it brings a number of enabling and simplifying features, it would be timely to upgrade the PyBaMM BPX utilities to support this version.

Questions to maintainers (@rtimms):

- Is BPX v1.1 sufficiently stable and coherent for support, or are any further bug fixes required at the BPX side before support by PyBaMM is reasonable?
- Maintain backwards compatibility for BPX v0.x, or not?

BradyPlanden:
BPX v1.x support is added in #5469. Feel free to reopen if something has been missed!

ejfdickinson:
@BradyPlanden I found one issue (there may be more). You need to make some additions to `parameters.bpx._get_pybamm_name()` where you are missing key conversions for the newly added BPX hysteresis variables:

"OCP (lithiation) [V]" -> "lithiation OCP [V]"
"OCP (delithiation) [V]" -> "delithiation OCP [V]"
"OCP hysteresis decay constant" -> "hysteresis decay rate"

Also the decay rate has an obligatory swap "electrode" -> "particle" and would need to be placed on both the lithiation and delithiation branches (two parameters) for the PyBaMM "one-state hysteresis" model.

ejfdickinson:
Completed in 26.5 release

#5469
#5507

PR #5507, opened 2026-05-11
Fix BPX 1.1 hysteresis parameter mapping

rtimms:
## Summary

Follow-up to #5469 (which added BPX 1.x support). The newly-introduced hysteresis fields on BPX 1.1's `Particle` block weren't being translated by `parameters/bpx.py::_get_pybamm_name`, so:

- `OCP (lithiation) [V]` ended up as `Negative electrode OCP (lithiation) [V]` (PyBaMM expects `Negative electrode lithiation OCP [V]`)
- `OCP (delithiation) [V]` similarly
- `OCP hysteresis decay constant` ended up under its raw BPX alias on the electrode pre-name; PyBaMM's Axen one-state hysteresis OCP submodel (`OneStateHysteresisOpenCircuitPotential`) evaluates two FunctionParameters — `<Domain> particle lithiation hysteresis decay rate` and `<Domain> particle delithiation hysteresis decay rate` — both on the **particle** pre-name. The BPX scalar now fans out to both with the same value.

While auditing the State section I found three adjacent gaps and fixed them in the same PR:

- `Initial hysteresis state: Positive/Negative electrode` were extracted into `bpx.state` but never copied to the PyBaMM dict, so the FunctionParameter was always falling back to 0. Now mapped to `[<Phase>: ]Initial hysteresis state in <domain> electrode`. Handles both the scalar (single-phase) and the per-particle-phase dict (blended) shapes by routing the blended-ness decision through the existing `_get_phase_names` helper and matching BPX phase keys to PyBaMM `Primary:` / `Secondary:` by particle declaration order.
- `Total heat transfer coefficient [W.m-2.K-1]` from `bpx.state.thermal_environment` was being silently clobbered to 0 a few lines later by an unconditional `pybamm_dict.update(...)`. Replaced with `setdefault`.
- `target_soc` defaulted to `1.0` regardless of what the BPX file declared in `State.initial_conditions.initial_soc`. The default is now `None`, resolved to the BPX value when a State section is present and 1.0 otherwise.

Implementation note: `_get_pybamm_name` now returns `list[str]` instead of `str` to support the 1→2 fan-out for the decay constant. The two call sites in `_bpx_to_domain_param_dict` iterate the returned list. Missing/`None`-valued optional hysteresis fields are skipped so they don't pollute the parameter dict.

Refs https://github.com/pybamm-team/PyBaMM/issues/5463.

## Test plan

- [x] All existing BPX tests in `tests/unit/test_parameters/test_bpx.py` still pass (13/13)
- [x] New unit tests (8) added to the same file:
  - [x] Single-phase BPX hysteresis fields → expected PyBaMM names; legacy names not present
  - [x] Missing hysteresis fields are skipped, not added as `None`
  - [x] Blended-electrode hysteresis fields → phase-prefixed PyBaMM names
  - [x] Initial hysteresis state (scalar form)
  - [x] Initial hysteresis state (per-particle-phase dict form, blended)
  - [x] Heat transfer coefficient regression test (BPX-supplied value preserved)
  - [x] `target_soc=None` falls back to BPX `Initial state-of-charge`
  - [x] End-to-end: BPX-derived parameter set runs the Axen one-state hysteresis SPM through a Discharge experiment

## Notes for reviewer

- BPX models only the Axen one-state hysteresis OCP, so the decay constant intentionally does NOT fan out to the Wycisk differential-capacity variant's `<Domain> particle hysteresis decay rate` (without `lithiation`/`delithiation` prefix). Users on that submodel will still need to add it manually.
- BPX `State.degradation` (`LLI`, `LAM: Positive electrode`, `LAM: Negative electrode`) is still unhandled — it has no direct PyBaMM equivalent and is out of scope here. Worth a separate issue if/when there's demand.


rtimms:
@ejfdickinson could you please take a look and see if any issues remain with any BPX files you have?

ejfdickinson:
@rtimms Thanks, everything BPX-related is working for me for the test cases we have.

Notes:

1. There is an oddity around initial hysteresis state parameters being created for materials that don't have an attached hysteresis model, but this is inherited from the BPX limitations, right? I assume these parameters don't do anything, and you don't know what model is targeted.

2. I did one test case (full DFN) with a single-material (blended description) graphite-SiOx electrode but setting
`options={"open-circuit potential": ("one-state hysteresis", "single")}`
This fails at discretisation stage with `ModelError: model is overdetermined (extra algebraic keys)` from the well-posedness checks. Is this something you're familiar with? It seems like it should be a separate issue but let me know if you'd like a repeat case, if you don't have any similar parameter sets.

rtimms:
@ejfdickinson thanks for testing!

1. Yes, exactly. These will just be ignored if they don't show up in the model.
2. I can't reproduce this. Could you share an example?

ejfdickinson:
@rtimms I diagnosed the problem - it was due to some placeholder zeroes for electrolyte properties, so no issues here. Just the runtime error path had changed compared to expectations from previous versions.

Good to go for this PR as far as I can see! Thank you!

src/pybamm/parameters/parameter_values.py
Daniel-Nicolae23: Hi @rtimms, can we please have the initial concentrations calculated via ESOH and added in for blended electrodes too?

ejfdickinson: This would be great, but it's a known limitation also for beginning-of-life preceding the degradation content in BPX, so I'd ask this not be at the expense of delaying a PyBaMM release covering the schema support!

@Daniel-Nicolae23 @rtimms You are welcome to inspect some of the tools we use to implement this after BPX import in our AEPyBaMM toolkit. Currently the code supports this for negative electrode blends only, but we are also working on some general-purpose utilities to be released soon.

https://github.com/About-Energy-OpenSource/AEPyBaMM/blob/main/src/aepybamm/sci_tools.py

see `compute_lithiation_bounds` and `add_initial_concentrations`

rtimms: Since PyBaMM's ESOH handles setting the initial state for blends and/or hysteresis, I'm actually inclined to drop `target_soc` from the `create_from_bpx` method and have it explicitly set `initial_soc=1` so that you can compute the initial concentrations for the negative phases from `theta_max` and all positive phases from `theta_min`. This would then allow PyBaMM to do its total Li calculations (which it does using the initial concentrations), and initial state could then be set after the BPX imports. 

Thoughts?

ejfdickinson: @rtimms No concern for us because this offers better flexibility in how the PyBaMM object is configured for context.

The overarching issue (which AEPyBaMM was designed to address) is that if `create_from_bpx()` doesn't know what model its output `pybamm.ParameterValues` object is going to be paired with, it doesn't know exactly how to make self-consistent parameters.

Daniel-Nicolae23: @rtimms I agree, we tend to run many simulations with various starting points and so we use the `lithium_ion.set_initial_state` quite a lot anyways. What matters in the first place is just to have the complete set, as you said, so the calculations can run.

rtimms:
## Update: deprecate `target_soc`, default to SOC=1 for all electrodes

Pushed three commits on top of the existing branch (commits `bcb2e3431`, `748f19662`, `1ce9cc0ec`).

### Motivation

The original branch made `ParameterValues.create_from_bpx[_obj]` fall back to the BPX file's `State.Initial state-of-charge` when `target_soc` was not passed. With the new BPX 1.1 hysteresis support, that workflow gets awkward — ESOH in PyBaMM is the right place to set initial state for blends and/or hysteresis, and `bpx.get_electrode_concentrations` doesn't handle blended electrodes anyway (the previous code silently logged a warning and left initial concentrations unset, which then broke downstream total-Li calculations).

### Change

Single contract for `create_from_bpx` / `create_from_bpx_obj`:

- **`target_soc=None`** (new default): always returns parameter values with initial concentrations at 100% SOC. Each phase's initial concentration is computed in a loop over `_get_phase_names(...)` from `{phase}{pre_name}{maximum,minimum} stoichiometry` × `{phase}Maximum concentration in {domain} [mol.m-3]` keys already populated by `bpx_to_param_dict`. Works uniformly for non-blended and blended electrodes (negative phases at θ_max, positive at θ_min). The BPX `State.Initial state-of-charge` field is no longer auto-applied.
- **`target_soc=<value>`** (deprecated): emits `DeprecationWarning`. Non-blended preserves the previous behaviour (`bpx.get_electrode_concentrations(target_soc, bpx)`); blended now raises `NotImplementedError` (previously this silently produced incomplete parameter values).

Users wanting a non-1 initial SOC call `param.set_initial_state(soc)` after creation.

### Files

- `src/pybamm/parameters/parameter_values.py` — `create_from_bpx`, `create_from_bpx_obj` emit the deprecation warning (`_warn_target_soc_deprecation` helper, `stacklevel=3`); `_create_from_bpx` rewritten. Reuses the existing `negative_electrode` / `positive_electrode` `Domain` constants from `pybamm.parameters.bpx` rather than re-deriving the lowercase/capitalised electrode strings.
- `tests/unit/test_parameters/test_bpx.py` — replaces `test_bpx_target_soc_defaults_to_bpx_initial_soc` with `test_bpx_default_initial_concentrations_at_full_charge` (θ_max × c_max / θ_min × c_max assertions) and `test_bpx_target_soc_is_deprecated_but_preserves_behaviour` (asserts exact match against `bpx.get_electrode_concentrations`). Adds `test_bpx_blended_default_sets_per_phase_initial_concentrations` (Primary/Secondary keys at θ_min × c_max) and `test_bpx_blended_with_target_soc_raises`. Both blended tests share a small `_blended_bpx_obj()` helper that calls the existing `_blended_positive_electrode()` fixture. `test_bpx_soc_error` now also asserts `DeprecationWarning`. All 24 `test_bpx.py` tests pass; full `tests/unit/test_parameters/` suite (266) green.
- `CHANGELOG.md` — drops the now-superseded "defaults to BPX Initial state-of-charge" bullet that landed earlier on this branch; adds a `## Breaking changes` heading describing the deprecation, the SOC=1 default, the `NotImplementedError` for `target_soc` + blended, and the `set_initial_state` recommendation.

valentinsulzer (approved):

PR #5518, opened 2026-05-15
Fix sensitivities crash when experiment input is not a sensitivity target

### LINKED ISSUE #5517: [Bug]: calculate_sensitivities unable to ignore experiment input parameters
### PyBaMM Version

26.4.3

### Python Version

3.14

### Describe the bug

Besides `True`, the `calculate_sensitivities` kwarg from `sim.solve()` accepts an explicit list of input parameter strings such that sensitivities are only calculated for those. However, when using input parameters in the experiment and not including them in the list, I get an error at evaluation time:
```
pybamm/expression_tree/input_parameter.py:118
KeyError: "Input parameter 'current' not found"
```
If I include the experiment parameter, but not other regular input parameters, everything works fine.

### Steps to Reproduce

```python

# Add an input parameter for the model
params = model.default_parameter_values
params.update({
    "Positive particle diffusivity [m2.s-1]": "[input]"
})

# Experiment with input parameter
exp2 = pybamm.Experiment([
    pybamm.step.Current(
        pybamm.InputParameter("current"),
        duration=10
    )
])

sim = pybamm.Simulation(
    model, 
    experiment=exp2,
    parameter_values=params
)
sol = sim.solve(
    inputs={
        "current": 0.1,
        "Positive particle diffusivity [m2.s-1]": 1e-14
    },
    calculate_sensitivities=["Positive particle diffusivity [m2.s-1]"] # This doesn't successfully ignore the current
    # calculate_sensitivities=["current"] # This successfully ignores the diffusivity
)

# Evaluate
sol["Voltage [V]"].sensitivities
```

### Relevant log output

```shell

```

rtimms:
## Summary

Fixes #5517.

When `sim.solve(calculate_sensitivities=[...])` is passed a *subset* of input parameters, `ProcessedVariable.initialise_sensitivity_explicit_forward` built a casadi-conversion `inputs` dict containing only the sensitivity *targets* as symbolic `MX.sym` variables. If the variable's expression tree referenced any *other* input parameter — for example a `pybamm.InputParameter` used as the step current in an experiment — the casadi tree walk hit `InputParameter._base_evaluate`, failed to find the name in the dict, and raised:

```
KeyError: "Input parameter 'current' not found"
```

The fix merges the concrete `inputs` dict with the symbolic `p_casadi` (`{**inputs, **p_casadi}`) so non-target inputs are passed through as constants. The jacobian is still taken with respect to `p_casadi_stacked`, which contains only the sensitivity targets, so the output shape and values are unaffected in the existing working case.

## Test plan

- [x] New regression test `test_processed_variable_sensitivities_ignore_experiment_input` (the MWE from #5517) — asserts `sol["Voltage [V]"].sensitivities` returns the correct keys, all values finite, non-zero.
- [x] All 71 existing `tests/unit/test_solvers/test_processed_variable.py` tests still pass.
- [x] Nearby existing test `test_solve_with_sensitivity_list_ignores_internal_experiment_input` still passes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Bug fixes
 
-- Fixed `ProcessedVariable.sensitivities` raising `KeyError` when `calculate_sensitivities` was passed a subset of inputs and a non-target input parameter (e.g. a `pybamm.InputParameter` used in an experiment step) appeared in the variable's expression tree. ([#5517](https://github.com/pybamm-team/PyBaMM/issues/5517))
+- Fixed `ProcessedVariable.sensitivities` raising `KeyError` when `calculate_sensitivities` was passed a subset of inputs and a non-target input parameter (e.g. a `pybamm.InputParameter` used in an experiment step) appeared in the variable's expression tree. ([#5518](https://github.com/pybamm-team/PyBaMM/pull/5518))
 
 ## Features
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # [Unreleased](https://github.com/pybamm-team/PyBaMM/)
 
+## Bug fixes
+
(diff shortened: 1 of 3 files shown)


martinjrobins (approved):
looks great, thanks @rtimms

PR #5520, opened 2026-05-17
Add solver timeout

NicolaCourtier:
# Description

Adds a `timeout` option to `BaseSolver` to avoid system freezes during integration. These changes address long-standing issue #5161. The `timeout` option is set to `None` by default to avoid any breaking changes.

## Type of change

# Important checks:

### DIFF
--- a/src/pybamm/solvers/base_solver.py
+++ b/src/pybamm/solvers/base_solver.py
@@ -8,6 +8,7 @@ import warnings
 
 import casadi
 import numpy as np
+from pebble import ProcessPool
 
 import pybamm
 from pybamm import ParameterValues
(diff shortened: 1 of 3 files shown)


src/pybamm/solvers/base_solver.py
```diff
@@ -795,32 +800,23 @@ def _integrate(
 
         inputs_list = inputs_list or [{}]
 
-        ninputs = len(inputs_list)
-        if ninputs == 1:
```
martinjrobins, line 799: I suspect the pool creation is going to be a big overhead (most of the time people are using ninputs == 1). You should benchmark this to make the use of a pool and async isn't going to slow down simulations, especially for ninputs == 1.

martinjrobins (changes requested):

src/pybamm/solvers/base_solver.py
NicolaCourtier: Thanks Martin, good point... In my case, I had ninputs=1 and needed the timeout.

I'll leave this PR open in the hope that someone can take the idea and finish it off.

NicolaCourtier:
The `pebble` package seems to provide significantly better stability and timeout functionality than either the multiprocessing Pool or ProcessPoolExecutor.

Note that the `timeout` functionality added to `BaseSolver` is not applied to the `IDAKLUSolver` due to a pickling error, but the `IDAKLUSolver` provides alternative early termination `options`.

Sven-Mattus:
Thanks @NicolaCourtier for addressing this. This is definitely helpful for the community and parallel/HPC simulations as mentioned in the other linked issues. I know of many users who implemented timeout functions around their simulations themselves, this helps address this directly. Also good to focus on stability (e.g. with using pebble) and allowing multiple ways for multiprocessing.

PR #5524, opened 2026-05-19
Flux boundary conditions in 1D finite volume

### LINKED ISSUE #5523: Flux boundary conditions in 1D finite volume
### Description

Say we have a model with a governing equation of the following form:
```
N = some_expression(grad(c))
dcdt = -pybamm.div(N)
model.rhs = {c: dcdt}
```
To define a boundary condition given for the flux `N`, this currently has to be reformulated in terms of `grad(c)` as a Neumann boundary condition on `c`. We propose to add another type of boundary condition that allows to directly define a boundary condition for the flux `N`, possibly looking something like this:
```
model.boundary_conditions = {
    c: {
        "left": (pybamm.Scalar(0), "Neumann"),
        "right": (pybamm.Scalar(-2), ("Flux", N)),
    }
}
```

### Motivation

This will simplify the implementation of models with more complex Fluxes and corresponding boundary conditions.

### Possible Implementation

_No response_

### Additional context

_No response_

SarahRo:
# Description

This PR implements the ability to apply 'flux' boundary conditions for the 1D finite volume method. This is an alternative way to apply Neumann boundary conditions, where the expression for the boundary condition does not have to rearranged as an explicit expression for the gradient. Instead  a method call 'add_flux_values' is implemented that is called from the method implementing the divergence operator.

Consider the following problem
```
N = -D * pybamm.grad(c)
dcdt = - pybamm.div(N)
```
with the boundary condition `N=-2` at the right hand boundary.
Instead of defining the boundary condition as `grad(c) = 2/D` at the right hand boundary, we can now define it directly as `N = 2` as follows:

```
bcs = {u: {"left": (pybamm.Scalar(0), "Neumann"), "right": (pybamm.Scalar(-2): ("Flux", N))}
```
To identify the correct boundary condition from within the implementation of the divergence operator, the symbol inside of `pybamm.div` must be identical to the symbol `N` in the boundary condition definition. To guarantee that this is the case, we need to avoid that any simplifications change the symbol inside of `pybamm.div`. For this purpose a 'simplify' property was added to pybamm.Divergence that we need to explicitly set to `False` for flux boundary conditions:
```
dcdt = - pybamm.div(N, simplify=False)
```

The following changes have been implemented:
- An add_flux_values method has been added to the `FiniteVolume` class.
- A test 'TestFluxBoundaryConditions' as added to test_finite_volume.py
- Unit tests (test_flux_boundary_conditions.py) have been added.
- `add_flux_values` is called from within the `divergence` method.
- A method '_extrapolate_gradient_to_boundaries' is added to ensure gradient has correct dimensions. The extra entries will be substituted either with Neumann boundary conditions or flux boundary conditions.
- A property 'simplify' was added to the Divergence operator that is `True` by default. If set to `False` the divergence operator no longer commutes with negation so that its child symbol remains unchanged.

Fixes #5523 

## Type of change

# Important checks:


SarahRo:
Hi @aabills , @brosaplanella suggested you may have some suggestions on how best to implement this.

Some of the open questions are:
- Where in the code it is best to add the boundary condition (currently within `divergence` method, although I have also experimented with calling it from within the `_process_symbol` method in the `Discretisation` class (https://github.com/SarahRo/PyBaMM/tree/flux-boundary-conditions)
- How to ensure the boundary condition is added in the right place (currently this relies on identifying the symbol for the flux which requires some workarounds to avoid the symbol being changed due to simplifications of constants/negation).

brosaplanella:
@aabills, for context, my first thought when I asked you about that is that we could just override the boundary flux. I believe that would require the boundary condition to be imposed at the divergence level (where the flux comes in) rather than at the gradient level.

brosaplanella (commented):
Hi Sarah! The PR looks in the right direction, but tests need fixing. I have checked them all, but it seems quite a few are dimension mismatches. Also, I suspect you will need to add some unit tests for the flux boundary conditions, as the coverage is computed only from unit tests.

brosaplanella:
@aabills it would be great to have your review once the tests pass!

Issue #5263, opened 2025-11-06
[Bug]: SPM model SEI on cracks doesn't actually reduce capacity


lorenzplank:
### PyBaMM Version

25.6.1

### Python Version

3.12.11

### Describe the bug

I am running a coupled degradation simulation based on the https://docs.pybamm.org/en/latest/source/examples/notebooks/models/coupled-degradation.html examples page, but using both DFN and SPM models. My code sets "SEI on cracks": "true" for both models.

Both models report very simular values for "Loss of capacity to negative SEI on cracks [A.h]" and "Loss of capacity to negative SEI [A.h]", so the variable is present and populated for SPM as well as DFN.

<img width="576" height="432" alt="Image" src="https://github.com/user-attachments/assets/3afcea3d-9321-4271-8d22-d9e526b64b8b" />

However, when I compare the actual capacity loss (time integrated from the charge cycles) to the reported loss contributions:

For DFN, the time integrated capacity loss matches the sum of SEI and SEI on cracks loss.
For SPM, the time integrated capacity loss matches only the SEI loss, and does not seem to include the SEI on cracks loss, even though it is shown as a Loss contribution.

<img width="576" height="432" alt="Image" src="https://github.com/user-attachments/assets/bab30d6e-4f3f-4a53-9d7a-ca410a9fc6d1" />

### Steps to Reproduce

```
import pybamm
import matplotlib.pyplot as plt
import numpy as np

# This function integrates the current over time to extract the delivered charge capacity
# for each charge segment (current < 0). Used to compare actual cell capacity loss.
def extract_charge_capacities(solution: pybamm.Solution) -> np.ndarray:
    """
    Extract charge capacities (Ah) from a solution.
    """
    time = solution["Time [s]"].entries
    current = solution["Current [A]"].entries
    discharge_mask = current < 0
    edges = np.where(np.diff(discharge_mask.astype(int)) != 0)[0] + 1
    if discharge_mask[0]:
        edges = np.insert(edges, 0, 0)
    if discharge_mask[-1]:
        edges = np.append(edges, len(current))
    capacities = []
    for i in range(0, len(edges), 2):
        start = edges[i]
        end = edges[i + 1]
        t_seg = time[start:end]
        i_seg = -current[start:end]
        if len(t_seg) > 1:
            capacity_As = np.trapz(i_seg, t_seg)
            capacities.append(capacity_As / 3600)
    return np.asarray(capacities)

[...]

PR #5541, opened 2026-05-28
Fix: SPM/SPMe ignore SEI-on-cracks Li loss in capacity budget (#5263)

binggao1230:
## Problem

When ageing a `pybamm.lithium_ion.SPM` or `pybamm.lithium_ion.SPMe` with `{"SEI on cracks": "true"}` enabled, the reported `"Loss of capacity to negative SEI on cracks [A.h]"` was not actually being deducted from the cell's deliverable capacity. The standard `"Loss of capacity to negative SEI [A.h]"` *was* applied. DFN's surface-form formulation handled both correctly.

A 30-cycle OKane2022 ageing experiment exposes the gap clearly:

```
                            BEFORE fix:                       AFTER fix:
                  SEI       SEI cracks   sum     actual    actual/sum     actual/sum
DFN          0.00437       0.01483     0.01920  0.01479      0.77          0.77
SPM          0.00431       0.01468     0.01899  0.00375      0.20  ← bug   0.81  ← matches DFN
SPMe         0.00436       0.01480     0.01916  0.00400      0.21  ← bug   0.84  ← matches DFN
```

Originally reported in #5263 with `OKane2022` + the `SEI: "solvent-diffusion limited"` + `"particle mechanics": ("swelling and cracking", "swelling only")` + `"SEI on cracks": "true"` recipe from the [coupled-degradation example](https://docs.pybamm.org/en/latest/source/examples/notebooks/models/coupled-degradation.html).

## Root cause

`CurrentForInverseKinetics.get_coupled_variables` in `src/pybamm/models/submodels/interface/kinetics/inverse_kinetics/base_inverse.py` builds the main intercalation current density via

```python
j_tot = variables["X-averaged {domain} electrode total interfacial current density [A.m-2]"]
j_sei = variables["{Domain} electrode SEI interfacial current density [A.m-2]"]
j_stripping = variables["{Domain} electrode lithium plating interfacial current density [A.m-2]"]
j = j_tot - j_sei - j_stripping
```

This is how SPM / SPMe close the lithium budget: any current consumed by side reactions has to be taken out of the main intercalation flux into the particle, otherwise that lithium ends up "double-spent" — the SEI submodel adds an Li-loss bookkeeping term while the particle model still receives the full current.

`SEI on cracks` is a separate side reaction (`SEI_on_cracks` submodel hierarchy, registered when `"SEI on cracks": "true"`) but `j_sei_on_cracks` is never subtracted here. So SPM / SPMe lose lithium to SEI-on-cracks *only on the bookkeeping side*, not on the simulated reaction current — making the reported loss decorative.

DFN's surface-potential-form electrolyte doesn't have this asymmetry: every reaction independently contributes a current density and total mass balance is enforced through charge conservation in `div(i_e)`, so adding a new side reaction is automatically accounted for.

## Fix

```python
j_sei_cracks_key = (
    f"{Domain} electrode SEI on cracks interfacial current density [A.m-2]"
)
if j_sei_cracks_key in variables:
    j_sei_cracks = variables[j_sei_cracks_key]
    roughness = variables[f"{Domain} electrode roughness ratio"]
    j_sei_cracks_eff = j_sei_cracks * (roughness - 1)
else:
    j_sei_cracks_eff = pybamm.Scalar(0)
...
j = j_tot - j_sei - j_sei_cracks_eff - j_stripping
```

Two subtleties:

1. The lookup is conditional. With `"SEI on cracks": "false"` (the default), the SEI-on-cracks submodel hierarchy doesn't register the current-density variable; the `dict.get`-style guard keeps that path identical to before.

2. `j_sei_on_cracks` is expressed *per unit crack area* `a_cr = (roughness - 1) * a`, where `a` is the primary particle surface area to volume ratio. `j_tot` and `j_sei` are per unit primary surface area `a`. To compose them on the same basis the new term is multiplied by `(roughness - 1)`. This mirrors the conversion already used in `base_interface._get_standard_volumetric_current_density_variables` (line 326–334) where the "SEI on cracks" branch multiplies the volumetric current by exactly the same factor.

## Verification

`tests/unit/test_models/test_submodels/test_interface/test_sei_on_cracks_inverse_kinetics_5263.py` adds three regression tests:

1. **Static graph check** — inspects the SPM model's `"X-averaged negative electrode interfacial current density [A.m-2]"` symbol tree and asserts the SEI-on-cracks current density variable is in there. Catches a future regression that drops the term again without needing a full ageing run.
2. **SPM ageing test** — runs the 30-cycle OKane2022 ageing experiment and asserts `actual_capacity_loss / (sei_loss + cracks_loss) > 0.6` and `actual_capacity_loss / sei_loss > 1.5`. Pre-fix value was `0.20` (well below `0.6`) and `0.87` (well below `1.5`); post-fix is `0.81` / `2.24`.
3. **SPMe ageing test** — same but for `pybamm.lithium_ion.SPMe`. Pre-fix `0.21`; post-fix `0.84`.

```text
$ python -m pytest tests/unit/test_models/test_submodels/test_interface/test_sei_on_cracks_inverse_kinetics_5263.py -v
3 passed in 3.20s

$ python -m pytest tests/unit/test_models/test_submodels/test_interface tests/unit/test_models/test_full_battery_models/test_lithium_ion -q
663 passed, 12 skipped in 8.34s
```

No regressions. DFN behaviour is unchanged by construction (DFN doesn't go through `CurrentForInverseKinetics`).

## Scope

Three files:

- `src/pybamm/models/submodels/interface/kinetics/inverse_kinetics/base_inverse.py` — add the `j_sei_on_cracks` subtraction with roughness scaling.
- `tests/unit/test_models/test_submodels/test_interface/test_sei_on_cracks_inverse_kinetics_5263.py` — new regression suite.
- `CHANGELOG.md` — unreleased bug-fix entry.

No public API change. The fix is gated on the variable already being present, so models without `"SEI on cracks": "true"` are untouched.


klinders:
Could it be that this issue is also showing in the average electrolyte concentration?

I'm seeing decrease in electrolyte capacity when running the SPMe with particle cracking enabled.

**Result**
image

Code below
```
import pybamm

options = {
    "SEI": "ec reaction limited",
    # "SEI": "none",
    # "lithium plating": "none",
    "lithium plating porosity change": "true",
    "transport efficiency": "Bruggeman",
    "particle mechanics": ("swelling and cracking", "swelling only"),
    "SEI on cracks": "true",
    "loss of active material": "stress-driven",
}
models = {
    pybamm.lithium_ion.SPM(options=options),
    pybamm.lithium_ion.SPMe(options=options),
    pybamm.lithium_ion.DFN(options=options)
}

param.update({
    "SEI kinetic rate constant [m.s-1]":1e-16,
    "EC diffusivity [m2.s-1]":1e-21,    
})

    [(
        # "Rest for 1500s",
        "Discharge at 5A for 3600s",
        "Rest for 1000s",
        "Charge at 1A for 9000s",
        "Rest for 8299314s"
    )],
    temperature=273.15+45
)

sims = []
for model in models:
    sim = pybamm.Simulation(model, parameter_values=param, experiment=experiment)
    sim.solve()
    sims.append(sim)

pybamm.dynamic_plot(sims, ["X-averaged electrolyte concentration [mol.m-3]"])
```

I see this is due to the different definitions of j_tot in the model. Especially here:
https://github.com/pybamm-team/PyBaMM/blob/f82976ffe45448e79a3f11473327890134568d30/packages/pybamm/src/pybamm/models/submodels/interface/kinetics/inverse_kinetics/base_inverse.py#L142

Will your fix also tackle this issue? Or should I make a new issue?

binggao1230:
Yes, same root cause. `j` from `CurrentForInverseKinetics` also feeds `TotalInterfacialCurrent`'s "Sum of ... volumetric interfacial current densities [A.m-3]" (`total_interfacial_current.py`), which the electrolyte diffusion/conductivity submodels use as their source term. Pre-fix, that sum double-counts SEI-on-cracks current: once via the inflated main-reaction `j` (never subtracted it), and again via the reaction's own separately-tracked `a_j_SEI on cracks` term — a spurious extra sink, consistent with the concentration drop you're seeing. I traced this from source rather than reproducing your exact experiment; will run it to confirm before this merges. No need for a separate issue.

PR #5547, opened 2026-05-28
Fix #2484: include unsaved cycles in solve_time / integration_time

### LINKED ISSUE #2484: [Bug]: Solve time returns wrong value if not all cycles are saved
### PyBaMM Version

22.10

### Python Version

3.9

### Describe the bug

The `solution.solve_time.value` returns different values depending on whether all cycles are saved or only one. I suspect it only returns the time for the saved cycles.

### Steps to Reproduce

```python3
import pybamm

    [
        "Discharge at 1C until 2.5 V",
        "Charge at C/3 until 4.2 V",
        "Hold at 4.2 V until C/20",
    ] * 20
)

param = pybamm.ParameterValues("Chen2020")

sim = pybamm.Simulation(model, parameter_values=param, experiment=experiment)

sim.solve()
print(sim.solution.solve_time.value)

sim.solve(save_at_cycles=[1])
print(sim.solution.solve_time.value)
```

### Relevant log output

```shell
2.047022205000019
0.0252866289999929
```

gaoflow:
# Description

Fixes #2484.

`Simulation.solve(save_at_cycles=...)` reported `solution.solve_time` and `solution.integration_time` that only covered the cycles `save_at_cycles` happened to keep in `Solution.cycles`. On the OP's 20-cycle experiment with `save_at_cycles=[1]`, `solve_time.value` was ~10× smaller than the all-cycles-saved run even though the solver had done the full work:

```
2.047022205000019   # save_at_cycles default
0.0252866289999929  # save_at_cycles=[1]
```

## Root cause

In `simulation.solve_for_experiment`, the per-cycle data is appended to `self._solution` only when the cycle is actually saved:

```python
if cycle_solution is not None and (
    save_this_cycle or feasible is False or stop_experiment
):
    self._solution = self._solution + cycle_solution
```

`Solution.__add__` is the only place where `solve_time` and `integration_time` get accumulated. When the cycle skips this `+`, its timer values are discarded along with its data.

## Fix

Track per-cycle timers in dedicated running totals that don't depend on `save_this_cycle`:

```python
cycle_solve_time_total = pybamm.TimerTime(0)
cycle_integration_time_total = pybamm.TimerTime(0)
# ...
for cycle_num, cycle_length in enumerate(...):
    # inner step loop builds up `cycle_solution += step_solution`,
    # so its solve_time / integration_time already cover the cycle.
    if cycle_solution is not None and not isinstance(cycle_solution, pybamm.EmptySolution):
        if cycle_solution.solve_time is not None:
            cycle_solve_time_total = cycle_solve_time_total + cycle_solution.solve_time
        if cycle_solution.integration_time is not None:
            cycle_integration_time_total = cycle_integration_time_total + cycle_solution.integration_time
    # ... existing save_this_cycle logic unchanged ...

# After the loop, replace the timer attributes on the returned solution:
if self._solution is not None:
    base_solve = (starting_solution.solve_time or pybamm.TimerTime(0)) if starting_solution else pybamm.TimerTime(0)
    base_integration = ...
    self._solution.solve_time = base_solve + cycle_solve_time_total
    self._solution.integration_time = base_integration + cycle_integration_time_total
```

The accumulator is populated from the **inner-loop** `cycle_solution`, which is built up by `cycle_solution += step_solution` over each step — that variable carries the full per-cycle cost, even before `make_cycle_solution` decides whether to keep the data. The final timer assignment supersedes whatever `Solution.__add__` accumulated, so the totals correctly include both saved and unsaved cycles plus any `starting_solution` contribution.

## Verification

Original reproducer (scaled to 10 cycles for CI speed):

```
Before:                    Now:
all saved:        0.0214   all saved:        0.0214
save_at_cycles=[1]: 0.0025 save_at_cycles=[1]: 0.0205   ← within jitter of all-saved
```

## Tests

`tests/unit/test_experiments/test_solve_time_unsaved_cycles_2484.py` (+121 lines, 5 tests):

* `test_solve_time_with_save_at_cycles_list_includes_unsaved_cycles` — `save_at_cycles=[1]` total must match the all-saved total within CPU-jitter tolerance.
* `test_integration_time_with_save_at_cycles_list_includes_unsaved_cycles` — same invariant for `integration_time`.
* `test_solve_time_with_save_at_cycles_int_includes_unsaved_cycles` — `save_at_cycles=N` (every Nth) path.
* `test_solve_time_with_starting_solution_carries_over_and_accumulates` — combined `starting_solution` + `save_at_cycles=[1]` doesn't lose either history or increment.
* `test_solve_time_remains_positive_and_finite_when_no_cycles_kept` — even `save_at_cycles=[999]` (no cycle saved) reports a real, positive `solve_time`.

Tolerance is `rel=2.0` (factor of 3 either way) to ride out CPU jitter; the bug was a factor of ~10, so any regression is still caught.

All 5 new tests pass. Pre-existing test suites stay green:

* `tests/unit/test_experiments/test_simulation_with_experiment.py` — 68 passed.
* `tests/unit/test_solvers/test_solution.py` + `tests/unit/test_simulation.py` — 133 passed (+1 pre-existing matplotlib optional-dep skip).

`ruff check` clean.

## Type of change

- [ ] New feature
- [ ] Optimization
- [x] Bug fix

# Key checklist

- [x] `ruff check` passes on touched files
- [x] All tests pass for the touched suites
- [x] Regression tests added (5 tests, covering list/int `save_at_cycles`, `starting_solution`, and edge cases)
- [x] CHANGELOG entry under `# [Unreleased]` → `## Bug fixes`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 - `pybamm.citations.register` now names the citation the caller passed in when a BibTeX string fails to parse, instead of whichever entry the parser had reached. ([#5677](https://github.com/pybamm-team/PyBaMM/pull/5677))
 - Deserialising a parameter set whose interpolant specification is invalid now logs a warning naming the offending parameter, instead of printing the bare exception to stdout with no indication of which parameter fell back to zero. ([#5679](https://github.com/pybamm-team/PyBaMM/pull/5679))
 - A user-specified boundary tag that is absent from a mesh file now raises `GeometryError` instead of degrading to a warning and a mesh with no boundary groups. ([#5679](https://github.com/pybamm-team/PyBaMM/pull/5679))
+- `Simulation.solve` with `save_at_cycles` no longer under-reports `solve_time` and `integration_time`: the timers now account for every cycle the solver ran, not only the saved ones. ([#5547](https://github.com/pybamm-team/PyBaMM/pull/5547))
+- `RegulariseSqrtAndPower` no longer regularises state-independent bases, fixing corrupted small rate constants in exchange-current density functions. ([#5600](https://github.com/pybamm-team/PyBaMM/pull/5600))
 
 # [v26.7.1.0](https://github.com/pybamm-team/PyBaMM/tree/pybamm-v26.7.1.0) - 2026-07-22
 
--- a/packages/pybamm/src/pybamm/simulation/simulation.py
+++ b/packages/pybamm/src/pybamm/simulation/simulation.py
@@ -988,11 +988,9 @@ class Simulation(BaseSimulation):
         # folded solution below to match the old left-fold's final termination.
         last_termination = None
 
-        # Accumulate solve/integration time from every step, not just the saved
-        # cycles. The folded solution only spans saved cycles, so its timers
-        # would exclude cycles dropped by save_at_cycles (#2484); the loop is the
-        # single source of truth instead. Seed with the loop-entry solution so a
-        # starting_solution and any initial padding rest are included.
+        # Track timers across every cycle separately from the folded solution,
+        # which only spans saved cycles (#2484). Seed from the loop-entry
+        # solution so starting_solution + any initial padding rest are counted.
         solve_time = pybamm.TimerTime(0)
         integration_time = pybamm.TimerTime(0)
         if self._solution is not None and not isinstance(
--- a/packages/pybamm/src/pybamm/simulation/simulation.py
+++ b/packages/pybamm/src/pybamm/simulation/simulation.py
@@ -1185,7 +1185,8 @@ class Simulation(BaseSimulation):
                 if not isinstance(cycle_solution, pybamm.EmptySolution):
                     last_termination = cycle_solution.termination
 
-            # Fold in this cycle's timers whether or not the cycle is saved.
+            # Accumulate this cycle's timers even if the cycle isn't saved
+            # (#2484). Must run before make_cycle_solution reassigns steps.
             for step_solution in steps:
                 if isinstance(step_solution, pybamm.EmptySolution):
                     continue
--- a/packages/pybamm/src/pybamm/simulation/simulation.py
+++ b/packages/pybamm/src/pybamm/simulation/simulation.py
@@ -988,6 +988,21 @@ class Simulation(BaseSimulation):
         # folded solution below to match the old left-fold's final termination.
         last_termination = None
 
+        # Accumulate solve/integration time from every step, not just the saved
+        # cycles. The folded solution only spans saved cycles, so its timers
+        # would exclude cycles dropped by save_at_cycles (#2484); the loop is the
+        # single source of truth instead. Seed with the loop-entry solution so a
+        # starting_solution and any initial padding rest are included.
+        solve_time = pybamm.TimerTime(0)
+        integration_time = pybamm.TimerTime(0)
+        if self._solution is not None and not isinstance(
+            self._solution, pybamm.EmptySolution
+        ):
+            if self._solution.solve_time is not None:
+                solve_time = self._solution.solve_time
+            if self._solution.integration_time is not None:
+                integration_time = self._solution.integration_time
+
         for cycle_num, cycle_length in enumerate(
             cycle_lengths,
             start=1,
@@ -1173,6 +1188,15 @@ class Simulation(BaseSimulation):
                 if not isinstance(cycle_solution, pybamm.EmptySolution):
                     last_termination = cycle_solution.termination
 
+            # Fold in this cycle's timers whether or not the cycle is saved.
+            for step_solution in steps:
+                if isinstance(step_solution, pybamm.EmptySolution):
+                    continue
+                if step_solution.solve_time is not None:
+                    solve_time += step_solution.solve_time
+                if step_solution.integration_time is not None:
+                    integration_time += step_solution.integration_time
+
             if steps:
                 if all(
                     isinstance(s, pybamm.EmptySolution) for s in steps
@@ -1242,6 +1266,13 @@ class Simulation(BaseSimulation):
             self._solution = pybamm.Solution.from_sub_solutions(cross_cycle_segments)
             if last_termination is not None:
                 self._solution.termination = last_termination
+
+        # Overwrite the folded timers with the all-cycles accumulator (#2484).
+        if self._solution is not None and not isinstance(
+            self._solution, pybamm.EmptySolution
+        ):
+            self._solution.solve_time = solve_time
+            self._solution.integration_time = integration_time
 
         if self._solution is not None and len(all_cycle_solutions) > 0:
             self._solution.cycles = all_cycle_solutions
--- a/packages/pybamm/src/pybamm/simulation/simulation.py
+++ b/packages/pybamm/src/pybamm/simulation/simulation.py
@@ -988,9 +988,8 @@ class Simulation(BaseSimulation):
         # folded solution below to match the old left-fold's final termination.
         last_termination = None
 
-        # Track timers across every cycle separately from the folded solution,
-        # which only spans saved cycles (#2484). Seed from the loop-entry
-        # solution so starting_solution + any initial padding rest are counted.
+        # Folded solution only spans saved cycles; accumulate timers over every
+        # cycle separately, seeded from any prior solution to keep its time (#2484).
         solve_time = pybamm.TimerTime(0)
         integration_time = pybamm.TimerTime(0)
         if self._solution is not None and not isinstance(
--- a/packages/pybamm/tests/unit/test_experiments/test_simulation_with_experiment.py
+++ b/packages/pybamm/tests/unit/test_experiments/test_simulation_with_experiment.py
@@ -89,6 +89,62 @@ class TestSimulationExperiment:
         assert len(sol.cycles) == 5  # cycles wiring preserved
         assert sol.summary_variables is not None
 
+    def test_solve_time_includes_unsaved_cycles(self, monkeypatch):
+        # #2484: solve_time / integration_time must cover every cycle the solver
+        # ran, not just the ones kept by save_at_cycles. Pin the per-step timer
+        # so the totals are deterministic, then check they don't change when
+        # intermediate cycles are dropped.
+        monkeypatch.setattr(pybamm.Timer, "time", lambda self: pybamm.TimerTime(1.0))
+
+        def run(**kwargs):
+            exp = pybamm.Experiment(
+                [("Discharge at 1C until 3.0V", "Charge at 1C until 4.2V")] * 3,
+                period=600,
+            )
+            sim = pybamm.Simulation(pybamm.lithium_ion.SPM(), experiment=exp)
+            sim.solve(**kwargs)
+            return sim.solution
+
+        full = run()
+        saved = run(save_at_cycles=[1])
+
+        # Cycles 2 and 3 are dropped (stored as None) ...
+        assert any(cycle is None for cycle in saved.cycles)
+        assert all(cycle is not None for cycle in full.cycles)
+        # ... but the reported timers still account for all three cycles.
+        assert saved.solve_time.value == full.solve_time.value
+        assert saved.integration_time.value == full.integration_time.value
+        assert full.solve_time.value > 0
+        assert full.integration_time.value > 0
+
+    def test_solve_time_seeded_from_starting_solution(self, monkeypatch):
+        # #2484: resuming from a starting_solution must add the new cycles'
+        # timers on top of the carried-over total, independent of save_at_cycles.
+        monkeypatch.setattr(pybamm.Timer, "time", lambda self: pybamm.TimerTime(1.0))
+
+        def make_sim(n_cycles):
+            exp = pybamm.Experiment(
+                [("Discharge at 1C until 3.0V", "Charge at 1C until 4.2V")] * n_cycles,
+                period=600,
+            )
+            return pybamm.Simulation(pybamm.lithium_ion.SPM(), experiment=exp)
+
+        start = make_sim(2).solve()
+        start_time = start.solve_time.value
+
+        resumed_full = make_sim(3).solve(starting_solution=start)
+        resumed_saved = make_sim(3).solve(starting_solution=start, save_at_cycles=[1])
+
+        # A middle resume cycle is dropped from the saved run ...
+        assert any(cycle is None for cycle in resumed_saved.cycles)
+        # ... the carried-over total is preserved and added to, not dropped ...
+        assert resumed_full.solve_time.value > start_time
+        # ... and dropping cycles from the resume run leaves the total unchanged.
+        assert resumed_saved.solve_time.value == resumed_full.solve_time.value
+        assert resumed_saved.integration_time.value == (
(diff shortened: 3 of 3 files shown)


rtimms (changes requested):
@gaoflow thanks, the implementation looks good, can you please clean up the verbose LLM comments?

binggao1230:
Thanks! I've trimmed the verbose comments down to short one-liners (the per-cycle timer accumulation block, the in-loop accumulate, and the final timing override), shortened the changelog entry to match the surrounding style, and removed an unused sim in the list test. The logic is unchanged and the 5 regression tests still pass locally.

src/pybamm/simulation/simulation.py
```diff
@@ -843,6 +843,17 @@ def solve(
...
         experiment_termination = None
 
+        # Running totals of per-cycle ``solve_time`` /
+        # ``integration_time``. Tracked separately from
+        # ``self._solution`` so that cycles excluded from
+        # ``self._solution`` by ``save_at_cycles`` still
+        # contribute to the reported timings. Without this,
+        # ``Solution.__add__`` only sees the saved cycles and
+        # ``sim.solution.solve_time`` under-reports the work
+        # actually done (issue #2484).
```
rtimms: Condense, e.g.:
```python
# Per-cycle solve/integration time, tracked separately from
# self._solution so cycles dropped by save_at_cycles still count
# towards the reported timings (#2484).
```

src/pybamm/simulation/simulation.py
```diff
@@ -1072,6 +1083,25 @@ def solve(
                 else:
                     idx += 1
 
+            # Always accumulate this cycle's timer values, even
+            # when the cycle data itself is not appended to
+            # ``self._solution``. ``cycle_solution`` here is the
+            # local in-loop accumulator built up by ``+=``
+            # over each step, so its ``solve_time`` /
+            # ``integration_time`` already reflect the full
+            # per-cycle cost. See #2484.
```
rtimms: Condense to ~2 lines, e.g. `# Accumulate this cycle's timers even if the cycle isn't saved (#2484).`

Worth keeping a one-line note that this must run *before* `cycle_solution` is reassigned by `make_cycle_solution` below (line ~1114) — that's the load-bearing ordering constraint, so it's the one thing actually worth a comment.

src/pybamm/simulation/simulation.py
```diff
@@ -1148,6 +1178,33 @@ def solve(
...
 
+        # Replace ``solve_time`` / ``integration_time`` on the
+        # returned solution with the cycle-aggregated totals
+        # (plus any starting-solution contribution), so the
+        # reported timings cover the entire experiment rather
+        # than only the cycles ``save_at_cycles`` happened to
+        # include. ``Solution.__add__`` already accumulates
+        # timers from saved cycles, but the unsaved ones never
+        # touch ``self._solution`` and would otherwise be
+        # silently dropped — see #2484.
```
rtimms: Condense, e.g. `# Apply cycle-aggregated timers (plus starting_solution) so timings cover unsaved cycles too (#2484).`

src/pybamm/simulation/simulation.py
```diff
@@ -1148,6 +1178,33 @@ def solve(
...
+        if self._solution is not None:
+            base_solve = (
```
rtimms: Minor edge case: `base_solve` only takes `starting_solution.solve_time`, but when a `starting_solution` with a `start_time` triggers the initial padding rest (lines ~859-893), that rest calls `run_padding_rest` and its solve time is folded into `self._solution` *before* this overwrite. Since the accumulator is only updated inside the cycle loop, that initial-rest time is now dropped from the total (the old `__add__`-based path counted it). Probably negligible, but flagging it.

tests/unit/test_experiments/test_solve_time_unsaved_cycles_2484.py
```diff
@@ -0,0 +1,129 @@
...
+    because cycles 2..N never made it into ``self._solution``. After
+    the fix the two paths must agree within a generous tolerance
+    that accounts for normal CPU jitter."""
+    sim_all = _make_sim()
+    t_all = _solve_time(sim_all)
+    sim_one = _make_sim()
+    t_one = _solve_time(sim_one, save_at_cycles=[1])
+    # Allow up to 3x variation either way for CPU jitter — the bug
+    # was a factor of ~10x under-report, so this still catches it.
+    assert t_one == pytest.approx(t_all, rel=2.0)
```
rtimms: This compares wall-clock `solve_time` between two separate solver runs — inherently non-deterministic and a likely source of intermittent CI failures despite the loose `rel=2.0`. Prefer a deterministic invariant: assert the reported total equals the sum of the per-cycle `solve_time`s (`sum(c.solve_time.value for c in sim.solution.cycles)` won't work for unsaved cycles, but you can monkeypatch `pybamm.TimerTime`/the solver timer to return fixed values and assert exact accumulation).

tests/unit/test_experiments/test_solve_time_unsaved_cycles_2484.py
```diff
@@ -0,0 +1,129 @@
...
+    # And it must definitely be more than the time for a single cycle.
+    sim_short = _make_sim()
+    sim_short.solve()
```
rtimms: `sim_short` is solved but its result is never used, and the comment on line 67 ("must definitely be more than the time for a single cycle") describes an assertion that isn't here. Looks like leftover scaffolding — drop these three lines or finish the assertion.

rtimms (commented):
Following up on my earlier review with specific spots. The implementation logic is correct (verified: `last_state` zeroes the timers so the carry-over case doesn't double-count, and the final overwrite supersedes `Solution.__add__` rather than adding to it, so saved cycles aren't counted twice). Main asks:

1. **Trim the comments.** The three explanatory blocks read like design notes — each is ~8 lines where 1–2 would do. Inline suggestions below.
2. **The timing-based tests are flaky by construction** (see comment on the `rel=2.0` assertion). Comparing wall-clock `solve_time` across two independent solver runs and asserting they agree within a factor of 3 will be noisy on CI. Consider asserting the structural invariant instead — e.g. that the `save_at_cycles=[1]` total equals the sum of per-cycle `solve_time`s, or monkeypatch the timer so the comparison is deterministic.
3. **Test file naming / placement.** A new file named after the issue number (`test_solve_time_unsaved_cycles_2484.py`) is non-idiomatic here; these would sit more naturally as a couple of cases in `test_simulation_with_experiment.py`. Also the docstrings are quite verbose.

rtimms:
Thanks for condensing the comments in `a7975bbd` 👍

Digging into this a bit more, I think there's a cleaner approach — but first a heads-up that's going to require some rework: **this branch is based on `main` from before #5551** (the "O(N) experiment accumulation via `Solution.from_sub_solutions`" change, merged June 1).

That PR removed the left-fold this fix is built around. The description's premise —

> `Solution.__add__` is the only place where `solve_time` and `integration_time` get accumulated

— is no longer true on current `main`. `simulation.solve_for_experiment` no longer does `self._solution = self._solution + cycle_solution`; it now collects saved cycles into a list and folds once:

```python
cross_cycle_segments.append(cycle_solution)          # only when the cycle is saved
...
self._solution = pybamm.Solution.from_sub_solutions(cross_cycle_segments)
```

and `from_sub_solutions` sums the timers over its segments. The bug you're fixing is still real on `main` (unsaved cycles → `cycle_solution = None` → never appended → excluded from the timer sum), but the patch here won't apply over the new code, so this'll need a rebase + re-implementation.

The upside: the rebase makes the fix cleaner and removes the part I think we both find awkward (tracking a parallel total, then overwriting the value the fold computed). The key point is that **timing can't come from the folded segments** — by design they only hold *saved* cycles — so the loop has to be the single source of truth. That means a single running accumulator, seeded from the starting solution and filled per-cycle from the step solutions (which are present regardless of `save_at_cycles`):

```python
# before the loop
solve_time = pybamm.TimerTime(0)
integration_time = pybamm.TimerTime(0)
if starting_solution is not None:
    solve_time = starting_solution.solve_time or solve_time
    integration_time = starting_solution.integration_time or integration_time

# in the loop, right after the cycle's `steps` are built — these step
# solutions carry the true solver cost whether or not the cycle is saved
for s in steps:
    if not isinstance(s, pybamm.EmptySolution):
        if s.solve_time is not None:
            solve_time += s.solve_time
        if s.integration_time is not None:
            integration_time += s.integration_time

# after the fold
if self._solution is not None:
    self._solution.solve_time = solve_time
    self._solution.integration_time = integration_time
```

One assignment after the loop is unavoidable (the returned `Solution` doesn't exist until `from_sub_solutions` runs), but it's now a single authoritative write off one accumulator rather than shadow-tracking + overwriting the fold's own result.

Two smaller things while you're in here:
- The timing-comparison tests (`solve_time` of two independent runs within `rel=2.0`) will be flaky on CI; a deterministic invariant would be more robust — see my inline comment on the earlier review.
- Once rebased, please double-check the new tests still exercise the path, since the internals they're implicitly pinning have changed.

Happy to push a rebased branch with this if it'd help give you a concrete diff to work from — just let me know.

binggao1230:
Thanks for digging in so thoroughly — and for verifying the carry-over/double-count logic.

You're right that #5551 pulls the rug out from under the premise here: with the left-fold gone and `from_sub_solutions` only summing over saved segments, timing can't come from the folded result anymore, so the loop has to be the single source of truth. Your single-accumulator version — seed from `starting_solution`, add each cycle's step-solution timers in the loop (they're present regardless of `save_at_cycles`), one authoritative write after the fold — is cleaner than the shadow-track-then-overwrite I had, and it also naturally fixes the `start_time`/padding-rest edge case you flagged, since the padding-rest steps get summed like any others rather than being left out of `base_solve`.

I'd gladly take you up on the offer to push the rebased branch — having the post-#5551 base as a concrete diff to build on would save a guessy rebase. On top of it I'll: reimplement with the single accumulator, swap the wall-clock comparisons for a deterministic invariant (assert the `save_at_cycles=[1]` total equals the sum of per-cycle/step `solve_time`s, no timing tolerance), and fold the cases into `test_simulation_with_experiment.py` with shorter docstrings rather than a separate issue-numbered file.

The `sim_short` leftover and the three comment blocks are already cleaned up in `a7975bbd`. Thanks again.

binggao1230:
Reworked on top of #5551. The loop is now the single source of truth for timing: a running \`solve_time\`/\`integration_time\` accumulator, incremented per cycle from the step solutions, with the folded solution's timers overwritten once after the fold — no more shadow-tracking-then-overwriting the fold's own result.

On the edge case you flagged: I seed the accumulator from the loop-entry solution rather than from `starting_solution` directly, so the initial padding rest (folded into `self._solution` before the loop) is included too.

Tests moved into `test_simulation_with_experiment.py` and made deterministic — they pin the per-step timer (`pybamm.Timer.time`) to a constant and assert the reported totals are invariant to `save_at_cycles`, instead of comparing wall-clock across runs. One covers a plain experiment, the other a resume from a `starting_solution` where a middle cycle is dropped.

One thing to flag: the branch targets the #5551 merge commit to keep the timing diff clean, so there's a one-line `CHANGELOG.md` conflict against the v26.6.0.0 release section — happy to rebase that away whenever the approach looks good to you.

binggao1230:
Thanks for the detailed pass. All three main asks are now in:

1. **Comments trimmed.** The three blocks are down to 1–3 lines each; kept the one load-bearing note that the accumulator is seeded from the loop-entry solution.
2. **Tests are deterministic now.** Dropped the wall-clock `rel=2.0` comparison entirely — the two tests `monkeypatch` the timer to a fixed value and assert exact accumulation (saved + unsaved cycles), so no CI flakiness.
3. **Placement.** Folded into `test_simulation_with_experiment.py` as `test_solve_time_includes_unsaved_cycles` / `test_solve_time_seeded_from_starting_solution`; the issue-numbered file (and the leftover `sim_short` scaffolding) is gone, docstrings shortened.

On your initial-padding-rest flag: that's covered — the accumulator seeds from `self._solution` at loop entry, which already has `starting_solution` plus any `run_padding_rest` folded in before the cycle loop, so that time is counted, not dropped.

One thing I can't clear from my side: there's a trivial `CHANGELOG.md` conflict (my Unreleased bullet vs. the v26.6.0.0 release reshuffle) — content is fine, it's purely the section move. Happy for you to take it on merge, or I'll rebase once I can.

binggao1230:
Thanks @rtimms. Cleaned up in d4247c768: trimmed the verbose comments in the timer-accumulation block and the two tests to concise one-liners, and replaced the per-assertion narration with short docstrings. No behaviour change; the two tests still pass and ruff is clean.

src/pybamm/simulation/simulation.py
binggao1230: Done, condensed to 1 line.

src/pybamm/simulation/simulation.py
binggao1230: Done, condensed to 2 lines.

src/pybamm/simulation/simulation.py
binggao1230: Done, condensed and added the ordering note (must run before `make_cycle_solution`).

tests/unit/test_experiments/test_solve_time_unsaved_cycles_2484.py
binggao1230: Done, tests now use `monkeypatch` for deterministic timing (`Timer.time` returns a fixed `TimerTime(1.0)`).

src/pybamm/simulation/simulation.py
binggao1230: The accumulator is seeded from `self._solution` after the initial rest runs (lines 996-1003), so the initial rest time is captured — the overwrite preserves it rather than dropping it. No change needed here.

tests/unit/test_experiments/test_solve_time_unsaved_cycles_2484.py
binggao1230: Done, the separate test file was removed and the tests rewritten in `test_simulation_with_experiment.py`.

binggao1230:
Thanks again for the thorough pass @rtimms. Current state addresses all the specific spots:

- **Comments** — the three accumulator/seed comments are now one/two-liners, and I kept the ordering note that the per-cycle accumulation must run before `make_cycle_solution` reassigns the steps.
- **Deterministic test** — the wall-clock comparison is gone. `test_solve_time_includes_unsaved_cycles` now pins `pybamm.Timer.time` to a constant so the totals are exact, and asserts `save_at_cycles` leaves `solve_time`/`integration_time` unchanged. The leftover `sim_short` and its dangling comment are removed.
- **Padding-rest edge case** — good catch. Since the rework, the loop seeds from `self._solution.solve_time`, and `Solution.__add__` already accumulates timers, so the resume-time padding rest (`current_solution += step_solution_with_rest`) folds into `self._solution` *before* the seed reads it — counted exactly once. I added `test_solve_time_counts_starting_solution_padding_rest` (2ebeed5ce) that proves it: with the timer pinned, a resume that triggers the padding rest reports exactly one extra unit of `solve_time` versus a no-gap resume.

All three `solve_time` tests pass locally. Ready for another look.

Issue #5548, opened 2026-05-28
Serialisation: 30 round-trip failures in convert_symbol_to_json / convert_symbol_from_json


BradyPlanden:
Six distinct round-trip failures in `convert_symbol_to_json` / `convert_symbol_from_json` (`src/pybamm/expression_tree/operations/serialise.py`).

## Common failure mode

Every bug shares the same shape. `convert_symbol_from_json` (`serialise.py:2235-2360`) ends with a generic fallback:

```python
return getattr(pybamm, json_data["type"])(
    *[convert_symbol_from_json(c) for c in json_data["children"]]
)
```

This works when the constructor takes only symbol children. It fails when the constructor needs non-Symbol positional args — a string side, a numpy array of data, a slice index — because those args were never serialised. `convert_symbol_to_json` (`serialise.py:2364-2520`) has its own symmetric gap: lacking an explicit case for the class, it falls into a generic fallback that writes only `{"type": ClassName, "children": [...]}`, dropping the constructor args that aren't children.

## Failures

Each repro below assumes:

```python
import pybamm
from pybamm.expression_tree.operations.serialise import (
    convert_symbol_from_json, convert_symbol_to_json,
)
```

- [ ] **`pybamm.BoundaryGradient` — drops `side`**

  ```python
  tree = pybamm.BoundaryGradient(pybamm.Time(), "left")
  restored = convert_symbol_from_json(convert_symbol_to_json(tree))
  # TypeError: BoundaryGradient.__init__() missing 1 required positional argument: 'side'
  ```
  There is an unmerged commit `5fecb74` that adds both to_json and from_json cases for `BoundaryGradient`.

- [ ] **`pybamm.BackwardIndefiniteIntegral` — drops `integration_variable`**

  `convert_symbol_from_json` has an explicit case for `IndefiniteIntegral` (`serialise.py:2329`) but not for the `Backward` subclass; the `isinstance(symbol, pybamm.IndefiniteIntegral)` check in `convert_symbol_to_json` (`serialise.py:2422`) doesn't fire either because `BackwardIndefiniteIntegral` isn't a subclass. The class falls into the generic fallback on both sides and the `integration_variable` arg is lost.

  ```python
  child = pybamm.Variable("u", domains={"primary": ["negative electrode"]})
  x = pybamm.SpatialVariable("x", domain=child.domain)
  tree = pybamm.BackwardIndefiniteIntegral(child, x)
[...]

PR #5549, opened 2026-05-29
Fix convert_symbol round-trip for symbols with non-child constructor args (#5548)

gaoflow:
## Description

Fixes the bulk of the round-trip failures catalogued in #5548.

The standalone `convert_symbol_to_json` / `convert_symbol_from_json` helpers in
`serialise.py` round-trip only a symbol's `children` (and `domains`). Symbols whose
constructors take **other** arguments therefore lost those arguments on round-trip:

- a numpy array of entries (`Array`/`Matrix`/`Vector`), a slice/int index (`Index`),
  `y_slices` (`StateVector`/`StateVectorDot`) → `TypeError`/`IndexError` on load;
- a side/direction/region string, a numeric position, an integration domain, or a
  sibling `Symbol` that is not a child (an initial condition, a distribution) →
  `TypeError` on load, or a silently *different* object;
- the three `*ToEdges` broadcasts additionally round-tripped back as their **parent
  class**, because `convert_symbol_to_json` matched them against the parent's
  `isinstance` branch and wrote the parent's `type`.

## Approach

Two complementary mechanisms, both routed through the existing dispatch chains:

1. **Protocol delegation.** Data/index nodes that already implement a complete
   `to_json` / `_from_json` protocol — `Index`, `Array` (+ `Matrix`/`Vector`), and
   `StateVectorBase` (+ `StateVector`/`StateVectorDot`) — now delegate to it instead of
   the generic child-only fallback. Listing the base classes covers their subclasses, so
   one definition keeps the `to_json` (class) and `convert_symbol_from_json` (type-name)
   sides in sync.

2. **A class-keyed registry** (`_extra_arg_serialisers`) records, per affected class, how
   to serialise the extra constructor arguments and how to rebuild the instance from its
   reconstructed children plus the JSON dict. Keying on the **exact class name** (rather
   than `isinstance`) is also what fixes the `*ToEdges` subclass-identity loss — these are
   dispatched here, before the `isinstance` chain, so the concrete subclass is preserved.

Classes covered by the registry: `PrimaryBroadcastToEdges`, `SecondaryBroadcastToEdges`,
`FullBroadcastToEdges`, `TertiaryBroadcast`, `TertiaryBroadcastToEdges`, `BoundaryMeshSize`,
`DeltaFunction`, `EvaluateAt`, `Magnitude`, `NodeToEdge2D`, `UpwindDownwind2D`,
`OneDimensionalIntegral`, `BoundaryIntegral`, `BackwardIndefiniteIntegral`,
`ExplicitTimeIntegral`, `SizeAverage`, `TensorField`, `VariableDot`, `SpatialVariableEdge`,
`DiscreteTimeData`.

## Scope / what is deferred

This addresses every failure in #5548 **except** the four already handled by the unmerged
commit `5fecb74` referenced in the issue — `BoundaryGradient` (`side`), `Variable`
(`scale`/`reference`), and `FullBroadcast` (`name`) — to avoid conflicting with that work.
Those can be merged independently or folded in if preferred.

## Verification

Each repro from the issue now round-trips with the correct class **and** a matching
structural `id` (or raises no error and preserves the relevant arg). New parametrized
fixtures in `tests/unit/test_serialisation/test_round_trip.py::TestSymbolRoundTrip` cover
all 20 newly-handled classes through a full `json.dumps`/`loads` cycle.

```
tests/unit/test_serialisation/  →  269 passed
TestSymbolRoundTrip             →  43 passed (was 23)
```

`ruff check` and `ruff format --check` are clean on both files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # [Unreleased](https://github.com/pybamm-team/PyBaMM/)
 
+## Bug fixes
+
+- Fixed `convert_symbol_to_json` / `convert_symbol_from_json` dropping non-child constructor arguments (numpy entries, slice/int index, `y_slices`, side/direction/region strings, a position, an integration domain, or a sibling `Symbol`), and silently rewriting the `*ToEdges` broadcasts to their parent class, on round-trip. ([#5549](https://github.com/pybamm-team/PyBaMM/pull/5549))
+
 # [v26.5.0](https://github.com/pybamm-team/PyBaMM/tree/v26.5.0) - 2026-05-27
 
 ## Breaking changes
--- a/src/pybamm/expression_tree/operations/serialise.py
+++ b/src/pybamm/expression_tree/operations/serialise.py
@@ -2232,6 +2232,163 @@ def convert_function_to_symbolic_expression(func, name=None):
     return ExpressionFunctionParameter(name, sym_output, func_name, func_args)
 
 
+# Symbols that already implement a complete ``to_json`` / ``_from_json`` round-trip
+# but whose constructors take non-Symbol positional arguments (a numpy array of
+# entries, a slice index, ``y_slices``...). The generic ``convert_symbol_*``
+# fallbacks only serialise ``children``, so these args were silently dropped,
+# causing ``TypeError``/``IndexError`` on load (see #5548). Delegate to the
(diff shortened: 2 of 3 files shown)


BradyPlanden:
Thanks @gaoflow, but I already have an implementation in mind.

PR #5390, opened 2026-02-24
Hermite knot reduction

MarcBerliner:
# Description

Companion PR to https://github.com/pybamm-team/pybammsolvers/pull/90

Adds an option to the `IDAKLUSolver` that dynamically reduces the size of the solution by removing points in the Hermite interpolant. Setting `hermite_reduction_factor > 1.0` compresses the solution size by introducing a small amount of error to the Hermite spline interpolant. A value of `2.0` roughly corresponds to a maximum 2x increase in error (practically the error is much smaller), while reducing the number of saved states by around 5-6x.

Example: DFN CC-CV charge showing the power. A value of `hermite_reduction_factor = 2` reduces the number of points by 5.4x, and introduces at most 0.9 mW of error compared to the original Hermite interpolant with a max error of ~5 mW near the CC-CV peak

image

## Type of change

Please add a line in the relevant section of [CHANGELOG.md](https://github.com/pybamm-team/PyBaMM/blob/main/CHANGELOG.md) to document the change (include PR #)

# Important checks:

Please confirm the following before marking the PR as ready for review:
- No style issues: `nox -s pre-commit`
- All tests pass: `nox -s tests`
- The documentation builds: `nox -s doctests`
- Code is commented for hard-to-understand areas
- Tests added that prove fix is effective or that feature works

PR #5550, opened 2026-05-29
perf: remove two O(N^2) slowdowns in Solution accumulation

BradyPlanden:
## Summary

Long experiment/ageing simulations rebuild a fresh `Solution` on every step append via `Solution.__add__`/`copy`. Two pieces of work in `Solution.__init__` re-scanned the entire accumulation on each append, making the total cost O(N^2) in step count. This fixes both.

## Root cause

A scaling profiler (run sim at N and 2N, compare per-function growth) surfaced two clusters scaling ~4x per doubling:

1. **Time-series validation.** `_ensure_sorted_t` / `_ensure_t_evals` ran on every construction. Added in #5390.
2. **Observability.** `Solution.observable` scanned every accumulated model on every construction.

Both are O(N) per append over the growing accumulation, so O(N^2) total. A microbenchmark of the validation scan extrapolates to ~160s at 16.5k steps, matching the original report.

## Fix

1. Gate the time-series validation behind a private `_validate_time_structure` flag. `__add__`/`copy` pass it `False` and instead re-validate only the newly joined boundary (self's last segment plus other's contribution) for both `all_ts` and `all_t_evals`. This is O(size of the appended piece) per add, so O(N) total, and is provably equivalent to the old full re-scan because each operand was already validated when it was built. `check_solution` keeps its original meaning (large-y check only), so the public contract and `CasadiSolver`'s existing `check_solution=False` call sites are unchanged.
2. Compute `Solution.observable` lazily on first access and cache it. It is only read at post-processing, never during accumulation.

## Benchmark

SPM experiment, `period=3600` (integration minimized to isolate the accumulation overhead). `gap` is wall minus reported `solve_time`.

| cycles | steps | unfixed wall | unfixed gap | fixed wall | fixed gap |
|------:|------:|------------:|-----------:|----------:|---------:|
| 200 | 800 | 0.83s | 0.41s | 0.63s | 0.21s |
| 400 | 1600 | 2.00s | 1.14s | 1.14s | 0.29s |
| 800 | 3200 | 5.73s | 4.01s | 2.20s | 0.52s |
| 1600 | 6400 | 19.26s | 15.71s | 4.35s | 0.97s |

Unfixed gap grows ~3.9x per doubling (O(N^2)); fixed grows ~1.85x (O(N)). At 1600 cycles, 4.4x faster wall and 16x less overhead. The win grows with size.

## Tests

- `_ensure_sorted_t`/`_ensure_t_evals` are called during `__add__` only on bounded boundary slices, never the full accumulation.
- Out-of-order and duplicate-junction merges still raise (the boundary re-scan preserves the strictly-increasing invariant; the duplicate-junction case is one the naive O(1) guard would have missed).
- `check_solution=False` still validates time structure and skips only the large-y check.
- `observable` is computed lazily and cached.
- Existing solution/experiment/CasADi/IDAKLU suites pass.

## Out of scope

A residual O(N^2) remains in `__add__` from list concatenation (`all_ts + other.all_ts` etc). It is pure pointer-copy, no longer dominant, and only matters far beyond current sim sizes. Removing it needs a batch-append restructure of the stepping loop, left as a follow-up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # [Unreleased](https://github.com/pybamm-team/PyBaMM/)
 
+## Optimizations
+
+- Fixed two O(N²) slowdowns in long experiment/ageing simulations where `Solution.__add__`/`copy` re-did whole-accumulation work on every step append: time-series validation now re-checks only the joined boundary, and `Solution.observable` is computed lazily. ([#5550](https://github.com/pybamm-team/PyBaMM/pull/5550))
+
 # [v26.5.0](https://github.com/pybamm-team/PyBaMM/tree/v26.5.0) - 2026-05-27
 
 ## Breaking changes
--- a/src/pybamm/solvers/solution.py
+++ b/src/pybamm/solvers/solution.py
@@ -256,10 +256,12 @@ class Solution(SolutionBase):
         variables_returned=False,
         check_solution=True,
         options=None,
+        _validate_time_structure=True,
     ):
         if not isinstance(all_ts, list):
             all_ts = [all_ts]
-        self._ensure_sorted_t(all_ts, "all_ts")
+        if _validate_time_structure:
+            self._ensure_sorted_t(all_ts, "all_ts")
         if not isinstance(all_ys, list):
             all_ys = [all_ys]
         if not isinstance(all_models, list):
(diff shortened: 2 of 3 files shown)


MarcBerliner (approved):
Looks great, thanks!

PR #5551, opened 2026-05-29
perf: O(N) experiment accumulation via Solution.from_sub_solutions

BradyPlanden:
## Summary
- Adds `Solution.from_sub_solutions`, an O(N) single-pass fold that is behaviourally equivalent to a left-fold of `Solution.__add__` (`functools.reduce(operator.add, ...)`), and routes both `make_cycle_solution` and the cross-cycle experiment accumulation through it. This removes the residual O(N^2) list concatenation flagged in #5550.
- Fixes an aliasing bug where `Solution.__add__` mutated the left operand's `_all_sensitivities` dict in place.
- Cross-cycle accumulation is now built once after the loop instead of via per-cycle `self._solution = self._solution + cycle_solution`; the skipped-step termination handling (`_check_infeasible_steps`) is preserved via a tracked `last_termination` applied to the folded solution.

Stacked on #5550 (base branch `fix/solution-append-on2`). Merge after #5550 lands.

### DIFF
--- a/src/pybamm/solvers/solution.py
+++ b/src/pybamm/solvers/solution.py
@@ -615,8 +615,7 @@ class Solution(SolutionBase):
             all_yps=all_yps,
             options=self.user_options,
         )
-        new_sol._all_inputs_stacked = self.all_inputs_stacked[:1]
-        new_sol._all_inputs_casadi = self.all_inputs_casadi[:1]
+        # stacked/casadi stay lazy; built from all_inputs[:1] on first access
         new_sol._sub_solutions = self.sub_solutions[:1]
 
         new_sol.solve_time = 0
@@ -660,8 +659,7 @@ class Solution(SolutionBase):
             all_yps=all_yps,
             options=self.user_options,
         )
-        new_sol._all_inputs_stacked = self.all_inputs_stacked[-1:]
-        new_sol._all_inputs_casadi = self.all_inputs_casadi[-1:]
+        # stacked/casadi stay lazy; built from all_inputs[-1:] on first access
         new_sol._sub_solutions = self.sub_solutions[-1:]
         new_sol.solve_time = 0
         new_sol.integration_time = 0
@@ -1192,7 +1190,12 @@ class Solution(SolutionBase):
             if s is not None and not isinstance(s, EmptySolution)
         ]
         if not sols:
-            return EmptySolution()
+            # all empty: match reduce(operator.add, ...) by copying the last empty
+            last_empty = next(
+                (s for s in reversed(sub_solutions) if isinstance(s, EmptySolution)),
+                None,
+            )
+            return last_empty.copy() if last_empty else EmptySolution()
         if len(sols) == 1:
             return sols[0].copy()
 
@@ -1216,8 +1219,7 @@ class Solution(SolutionBase):
         t_evals_present = all(s._all_t_evals is not None for s in segments)
 
         all_ts, all_ys, all_yps, all_t_evals = [], [], [], []
-        all_models, all_inputs = [], []
-        inputs_stacked, inputs_casadi, sub_sols = [], [], []
+        all_models, all_inputs, sub_sols = [], [], []
 
         for s, repeated in kept:
             first = slice(1, None) if repeated else slice(None)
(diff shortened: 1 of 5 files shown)


src/pybamm/solvers/solution.py
```diff
@@ -1177,6 +1178,122 @@ def __add__(self, other):
...
+            return EmptySolution()
+        if len(sols) == 1:
+            return sols[0].copy()
+
+        hermite = all(s.hermite_interpolation for s in sols)
+        t_evals_present = all(s._all_t_evals is not None for s in sols)
+
+        all_ts, all_ys, all_yps, all_t_evals = [], [], [], []
+        all_models, all_inputs = [], []
+        inputs_stacked, inputs_casadi, sub_sols = [], [], []
```
MarcBerliner: Can you pre-allocate these to `[None] * len(sub_solutions)`

src/pybamm/solvers/solution.py
```diff
@@ -1177,6 +1178,122 @@ def __add__(self, other):
     def __radd__(self, other):
         return self.__add__(other)
 
+    @classmethod
```
MarcBerliner, line 1181: There's a decent amount of overlap between this method and `__add__`. I think we should remove the `__add__` internals and always use this method, something like
```python
def __add__(self, other):
    return self.from_sub_solutions(self.sub_solutions + other.sub_solutions)`
```

src/pybamm/solvers/solution.py
```diff
@@ -1177,6 +1178,122 @@ def __add__(self, other):
...
+                all_yps.append(s.all_yps[0][:, first])
+                all_yps.extend(s.all_yps[1:])
+            if t_evals_present:
+                all_t_evals.append(s._all_t_evals[0][first])
+                all_t_evals.extend(s._all_t_evals[1:])
+            # bookkeeping lists
+            all_models.extend(s.all_models)
+            all_inputs.extend(s.all_inputs)
+            inputs_stacked.extend(s.all_inputs_stacked)
+            inputs_casadi.extend(s.all_inputs_casadi)
```
MarcBerliner: I think these two should be lazy. The casadi inputs especially are deceptively expensive to compute
```suggestion
```

src/pybamm/solvers/solution.py
```diff
@@ -1177,6 +1178,122 @@ def __add__(self, other):
...
+            options=options,
+            _validate_time_structure=False,
+        )
+
+        # validate the joined series once (O(N), not O(N^2))
+        new_sol._ensure_sorted_t(new_sol.all_ts, "all_ts")
+        if new_sol._all_t_evals is not new_sol.all_ts:
+            new_sol._ensure_t_evals(
+                all_ts=new_sol.all_ts, all_t_evals=new_sol._all_t_evals
+            )
```
MarcBerliner: Do we need the t checking here is we set `_validate_time_structure=True`?

src/pybamm/solvers/solution.py
```diff
@@ -1177,6 +1178,122 @@ def __add__(self, other):
...
+        instead of O(N^2): segment lists are concatenated once rather than
+        rebuilding a Solution per step. Inputs must already be valid.
+        """
+        sols = [
+            s
+            for s in sub_solutions
+            if s is not None and not isinstance(s, EmptySolution)
+        ]
+        if not sols:
+            return EmptySolution()
```
MarcBerliner: Does this need a potential termination attached as well?

MarcBerliner (changes requested):
Overall looks good, some comments

src/pybamm/solvers/solution.py
BradyPlanden: This pushes `__add__` back to O(N^2) accumulation unfortunately, as it re-folds the entire history on every addition. I've refactored in the spirit of this, with two small helpers to capture the subset that wasn't DRY.

src/pybamm/solvers/solution.py
BradyPlanden: This would break the `.extend()/.append()` accumulation used on these lists, which is already amortised `O(1)` in python. These can be larger than `[None] * len(sub_solutions)` due to nested sub-solutions, so we would need some additional logic to pre-allocate the correct sizing.

MarcBerliner (dismissed):
Thanks!

MarcBerliner (approved):

PR #5556, opened 2026-06-01
Allow electrode conductivity to be a function of stoichiometry

aabills:
## Description

Electrode electronic conductivity (`sigma`) can now be supplied as a function of **stoichiometry** (in addition to temperature) for all lithium-ion and sodium-ion models, following the same pattern already used for diffusivity `D` and open-circuit potential `U`.

Previously `sigma` was a function of temperature only. The `sigma` parameter method now accepts an optional stoichiometry argument; the electrode-ohm submodels look up the (surface) stoichiometry and feed it through. Constant or temperature-only conductivity parameter sets remain fully backwards compatible (the stoichiometry input is simply ignored), and lead-acid is unaffected.

### Changes

- `LithiumIonParameters.sigma(T, sto=None)` — when `sto` is given, it is clipped to `(tol, 1 - tol)` and passed as the `"{Domain} electrode stoichiometry"` `FunctionParameter` input. New `sigma__c_s` tolerance in `settings.py`.
- `LeadAcidParameters.sigma(T, sto=None)` — accepts (and ignores) `sto` for a uniform call signature with the shared submodels.
- Electrode-conductivity submodels updated to pass surface stoichiometry to `sigma`:
  - `electrode/ohm/full_ohm.py` (DFN, Newman-Tobias)
  - `electrode/ohm/composite_ohm.py` (SPMe; uses X-averaged surface stoichiometry)
  - `electrode/ohm/surface_form_ohm.py`
  - `electrolyte_conductivity/surface_potential_form/full_surface_form_conductivity.py`
  - shared helper `_get_electrode_stoichiometry` on `base_ohm` (falls back to the primary phase for two-phase electrodes, and to `None` when no particle is present — e.g. lead-acid, planar lithium metal).
- Standalone reference models updated too: `lithium_ion/basic_dfn.py`, `basic_dfn_composite.py`, `basic_dfn_2d.py`, and `sodium_ion/basic_dfn.py`.

## Tests

- Unit: parameter plumbing + clipping (`test_lithium_ion_parameters.py`), DFN single- and two-phase wiring (`test_dfn.py`).
- Integration (`test_stoichiometry_conductivity.py` + sodium `test_basic_models.py`): exercises every `sigma` path — `full_ohm`, `composite_ohm`, `surface_form_ohm`, full surface-form conductivity (differential + algebraic), two-phase composite, and all four basic models — asserting a stoichiometry-dependent conductivity both solves and changes the solution.
- Existing lead-acid integration tests confirm the `sto=None` path is unaffected.

## Test plan

- [x] New unit tests pass
- [x] New integration tests pass (all sigma paths)
- [x] Lead-acid integration tests pass (backwards-compatible None path)
- [x] `ruff check` / `ruff format` clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # [Unreleased](https://github.com/pybamm-team/PyBaMM/)
 
+## Features
+
+- Electrode electronic conductivity can now be specified as a function of stoichiometry (in addition to temperature) for all lithium-ion and sodium-ion models. ([#5558](https://github.com/pybamm-team/PyBaMM/pull/5558))
+
 ## Bug fixes
 
 - Fixed unified experiment mode using excessive memory and time for experiments with many cycles. ([#5554](https://github.com/pybamm-team/PyBaMM/pull/5554))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ## Breaking changes
 
-- Electrode electronic conductivity supplied as a function must now accept `(stoichiometry, temperature)`. A constant value is unaffected, but a conductivity function previously written as `f(temperature)` must be updated to `f(stoichiometry, temperature)`. ([#5556](https://github.com/pybamm-team/PyBaMM/pull/5556))
+- Electrode electronic conductivity supplied as a function must now accept `(stoichiometry, temperature)`. A constant value is unaffected, but a conductivity function previously written as `f(temperature)` must be updated to `f(stoichiometry, temperature)`; supplying a temperature-only function now raises a clear error pointing at the new signature. ([#5556](https://github.com/pybamm-team/PyBaMM/pull/5556))
 
 ## Bug fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 - Electrode electronic conductivity can now be specified as a function of stoichiometry (in addition to temperature) for all lithium-ion and sodium-ion models. ([#5556](https://github.com/pybamm-team/PyBaMM/pull/5556))
 
+## Breaking changes
+
+- Electrode electronic conductivity supplied as a function must now accept `(stoichiometry, temperature)`. A constant value is unaffected, but a conductivity function previously written as `f(temperature)` must be updated to `f(stoichiometry, temperature)`. ([#5556](https://github.com/pybamm-team/PyBaMM/pull/5556))
+
 ## Bug fixes
 
 - Fixed unified experiment mode using excessive memory and time for experiments with many cycles. ([#5554](https://github.com/pybamm-team/PyBaMM/pull/5554))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Features
 
-- Electrode electronic conductivity can now be specified as a function of stoichiometry (in addition to temperature) for all lithium-ion and sodium-ion models. ([#5558](https://github.com/pybamm-team/PyBaMM/pull/5558))
+- Electrode electronic conductivity can now be specified as a function of stoichiometry (in addition to temperature) for all lithium-ion and sodium-ion models. ([#5556](https://github.com/pybamm-team/PyBaMM/pull/5556))
 
 ## Bug fixes
 
--- a/src/pybamm/input/parameters/lithium_ion/ORegan2022.py
+++ b/src/pybamm/input/parameters/lithium_ion/ORegan2022.py
@@ -482,7 +482,7 @@ def graphite_LGM50_entropic_change_ORegan2022(sto):
     return dUdT
 
 
-def nmc_LGM50_electronic_conductivity_ORegan2022(T):
+def nmc_LGM50_electronic_conductivity_ORegan2022(sto, T):
     """
     Positive electrode electronic conductivity as a function of the temperature from
     [1].
@@ -495,6 +495,9 @@ def nmc_LGM50_electronic_conductivity_ORegan2022(T):
 
     Parameters
     ----------
+    sto: :class:`pybamm.Symbol`
+       Electrode surface stoichiometry (unused; conductivity depends on temperature
+       only, but the model supplies stoichiometry first)
     T: :class:`pybamm.Symbol`
        Dimensional temperature
 
--- a/src/pybamm/models/full_battery_models/lead_acid/basic_full.py
+++ b/src/pybamm/models/full_battery_models/lead_acid/basic_full.py
@@ -150,11 +150,11 @@ class BasicFull(BaseModel):
         # Current in the solid
         ######################
         i_s_n = (
-            -self.param.n.sigma(T)
+            -self.param.n.sigma(None, T)
             * (1 - eps_n) ** self.param.n.b_s
             * pybamm.grad(phi_s_n)
         )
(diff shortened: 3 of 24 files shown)


rtimms:
### Suggested change: clear error when `sigma` is passed as a function of temperature only

This PR makes electrode conductivity a function of `(sto, T)`. That's a breaking change for any parameter set whose conductivity is a `f(T)` function — but right now the failure is cryptic. With a temperature-only function the simulation dies with:

```
TypeError: <lambda>() takes 1 positional argument but 2 were given
```

which gives the user no hint that they need to add a stoichiometry argument. I'd suggest detecting this and raising a clear, actionable error. It fires **once** (the raise stops execution) and never for a scalar/constant conductivity (a `Scalar` is handled before the `callable` branch, so it never reaches the check).

The natural place is the callable branch in `ParameterSubstitutor._process_symbol`, where the user's function is actually invoked and its arity can be compared against the supplied inputs:

```python
elif callable(function_name):
    # otherwise evaluate the function to create a new PyBaMM object
    self._check_electrode_conductivity_signature(symbol, function_name)
    function = function_name(*new_children)
```

plus a small helper (and `import inspect` at the top of the file):

```python
@staticmethod
def _check_electrode_conductivity_signature(symbol, function_name):
    """
    Electrode conductivity may now be supplied as a function of
    (stoichiometry, temperature). If it is still supplied as a function of
    temperature only, raise a clear error rather than letting it fail later
    with a cryptic "takes 1 positional argument but 2 were given" TypeError.

    Only triggers for a callable conductivity that is being passed
    stoichiometry (``input_names`` has more than just temperature); a
    constant conductivity is handled elsewhere and never reaches here.
    """
    if not symbol.name.endswith("electrode conductivity [S.m-1]"):
        return
    # only temperature is supplied (e.g. lead-acid, planar lithium metal)
    if len(symbol.input_names) < 2:
        return
    try:
        params = inspect.signature(function_name).parameters.values()
    except (TypeError, ValueError):
        # can't introspect (e.g. some builtins) - let the call proceed
        return
    # a function with *args can take any number of positional arguments
    if any(p.kind == p.VAR_POSITIONAL for p in params):
        return
    n_positional = sum(
        p.kind in (p.POSITIONAL_ONLY, p.POSITIONAL_OR_KEYWORD) for p in params
    )
    if n_positional < len(symbol.input_names):
        raise TypeError(
            f"'{symbol.name}' must now be a function of "
            "(stoichiometry, temperature), e.g. `def sigma(sto, T): ...`. "
            "Update your temperature-only conductivity function `f(T)` to "
            "`f(sto, T)`; the stoichiometry argument may be ignored if the "
            "conductivity does not depend on it."
        )
```

Why this spot rather than `LithiumIonParameters.sigma`: at the point `sigma()` builds the `FunctionParameter` it only has the symbolic stoichiometry/temperature args, not the user-supplied callable — the substitutor is the layer where the actual value is resolved.

Behaviour I verified locally on this branch:

| conductivity supplied as | before | with suggestion |
|---|---|---|
| `lambda T: ...` (old style, Li-ion) | cryptic `TypeError` | clear error naming the fix |
| scalar / constant | solves | solves (check skipped) |
| `lambda sto, T: ...` (new style) | solves | solves |
| `lambda T: ...` for lead-acid / Li-metal (`sto=None`, 1 input) | solves | solves (check skipped) |

It might also be worth a small unit test that asserts the helpful error is raised for an old-style `f(T)` conductivity, so the breaking change is locked in.

rtimms:
### Proposed updated PR description (current one is stale)

The description below reflects what's actually in the diff (the file list had drifted — it omitted `basic_dfn_half_cell.py`, `li_metal.py`, `base_thermal.py`, and the `ORegan2022` parameter-set signature change) and folds in the clear-error behaviour from the suggestion comment.

---

## Description

Electrode electronic conductivity (`sigma`) can now be supplied as a function of **stoichiometry** (in addition to temperature) for all lithium-ion and sodium-ion models, following the same pattern already used for diffusivity `D` and open-circuit potential `U`.

Previously `sigma` was a function of temperature only. The `sigma` parameter method now takes `(sto, T)`; the electrode-conductivity submodels look up the (surface) stoichiometry and feed it through. **This is a breaking change**: a conductivity supplied as a function must now accept `(stoichiometry, temperature)`. Constant (scalar) conductivities are unaffected, and the no-particle paths (lead-acid, planar lithium metal) pass `sto=None` so they are unchanged.

## Changes

- `LithiumIonParameters.sigma(self, sto, T)` — when `sto` is not `None` it is clipped to `(tol, 1 - tol)` and passed as the `"{Domain} electrode stoichiometry"` `FunctionParameter` input. New `sigma__c_s` tolerance in `settings.py`.
- `LeadAcidParameters.sigma(self, sto, T)` — accepts (and ignores) `sto` for a uniform call signature with the shared submodels.
- Shared helper `_get_electrode_stoichiometry` on `base_ohm` (falls back to the primary phase for two-phase electrodes, and to `None` when no particle is present).
- Electrode-conductivity submodels updated to pass surface stoichiometry to `sigma`:
  - `electrode/ohm/full_ohm.py` (DFN, Newman-Tobias)
  - `electrode/ohm/composite_ohm.py` (SPMe; uses X-averaged surface stoichiometry)
  - `electrode/ohm/surface_form_ohm.py`
  - `electrode/ohm/base_ohm.py` (positive-electrode boundary condition)
  - `electrolyte_conductivity/surface_potential_form/full_surface_form_conductivity.py`
- No-particle / current-collector paths updated to pass `sto=None`:
  - `electrode/ohm/li_metal.py`, `thermal/base_thermal.py` (planar negative ohmic heating), lead-acid `basic_full.py`.
- Standalone reference models updated: `lithium_ion/basic_dfn.py`, `basic_dfn_composite.py`, `basic_dfn_2d.py`, `basic_dfn_half_cell.py`, and `sodium_ion/basic_dfn.py`.
- `input/parameters/lithium_ion/ORegan2022.py`: `nmc_LGM50_electronic_conductivity_ORegan2022(sto, T)` now takes the stoichiometry argument first (unused; conductivity depends on temperature only).
- A temperature-only conductivity function (`f(T)`) supplied to a lithium-ion / sodium-ion model raises a clear error telling the user to change it to `f(sto, T)`, instead of failing with a cryptic `TypeError: takes 1 positional argument but 2 were given`. _(pending the suggestion in review)_

## Tests

- Unit: parameter plumbing + clipping (`test_lithium_ion_parameters.py`), DFN single- and two-phase wiring (`test_dfn.py`), and the `ORegan2022` conductivity signature (`test_LGM50_ORegan2022.py`).
- Integration (`test_stoichiometry_conductivity.py` + sodium `test_basic_models.py`): exercises every `sigma` path — `full_ohm`, `composite_ohm`, `surface_form_ohm`, full surface-form conductivity (differential + algebraic), two-phase composite, and all four basic models — asserting a stoichiometry-dependent conductivity both solves and changes the solution.
- Existing lead-acid integration tests confirm the `sto=None` path is unaffected.
- _(suggested)_ a test asserting the helpful error is raised for an old-style `f(T)` conductivity.

## Test plan

- [x] New unit tests pass
- [x] New integration tests pass (all sigma paths)
- [x] Lead-acid integration tests pass (backwards-compatible `None` path)
- [x] `ruff check` / `ruff format` clean

rtimms (changes requested):
LGTM but I think we should add a catch for `sigma(T)` in the code so we can give better errors

rtimms (approved):

MarcBerliner (approved):

PR #5560, opened 2026-06-03
Serialisation kernel: safe-or-loud encode/decode and Symbol surface migration

### LINKED ISSUE #5548: Serialisation: 30 round-trip failures in convert_symbol_to_json / convert_symbol_from_json

BradyPlanden:
# Description

First of two stacked PRs replacing the lossy `convert_symbol_to_json` / `convert_symbol_from_json` type-switch with a single safe-or-loud serialisation kernel. Every value either round-trips correctly or raises `SerialisationError` at encode time; no code path silently drops a constructor argument.

This PR covers:

- **Legacy fixtures pinned first**: authentic pre-refactor files (compact `type` symbols, `class`+`module` submesh types) committed from `main` and regression-tested through the legacy read shim, so old files keep loading.
- **The kernel** (`expression_tree/operations/serialise_kernel.py`): encode/decode engine, codec registry, introspection-based `DefaultCodec`, explicit `to_json`/`_from_json` hooks, leaf codecs (ndarray, slice, tuple, inf/nan floats), `class_reference` for type references, and a read-only legacy normalisation shim. Trap tests lock in the safe-or-loud guard.
- **Symbol surface migration**: `convert_symbol_*` now route through the kernel and the central isinstance ladder is deleted. Hooks added for the ~20 classes that previously crashed or silently dropped arguments (spatial operators, broadcasts, `Index`, `Variable` scale/reference, concatenations, interpolants, and friends).
- `_KNOWN_FAILING` is empty: every #5548 repro round-trips, verified by dedicated repro tests and property tests over the strategy suite.

Fixes #5548

Stacked on `feat/serialisation-property-tests` (the property-test harness this builds on). A follow-up PR will target this branch to migrate the discretised-model surface and adjacent surfaces (solvers, experiments, parameter values, geometry, meshes) onto the same kernel, with cleanup, docs, and the CHANGELOG entry.

# Important checks:

### DIFF
--- a/src/pybamm/expression_tree/interpolant.py
+++ b/src/pybamm/expression_tree/interpolant.py
@@ -51,6 +51,9 @@ class Interpolant(pybamm.Function):
         the higher potential for errors in extrapolation.
     """
 
+    # entries_string is a derived hash cache of x/y, which are emitted by to_json.
+    _serialise_derived_params = frozenset({"entries_string"})
+
     def __init__(
         self,
         x: npt.NDArray[np.float64] | Sequence[npt.NDArray[np.float64]],
@@ -172,9 +175,9 @@ class Interpolant(pybamm.Function):
             np.array(snippet["y"]),
             snippet["children"],
             name=snippet["name"],
-            interpolator=snippet["interpolator"],
-            extrapolate=snippet["extrapolate"],
-            _num_derivatives=snippet["_num_derivatives"],
+            interpolator=snippet.get("interpolator", "linear"),
+            extrapolate=snippet.get("extrapolate", True),
+            _num_derivatives=snippet.get("_num_derivatives", 0),
         )
 
     @property
--- a/src/pybamm/expression_tree/interpolant.py
+++ b/src/pybamm/expression_tree/interpolant.py
@@ -406,7 +406,6 @@ class Interpolant(pybamm.Function):
 
         json_dict = {
             "name": self.name,
-            "id": self.id,
             "x": [x_item.tolist() for x_item in self.x],
             "y": self.y.tolist(),
             "interpolator": self.interpolator,
--- a/src/pybamm/expression_tree/interpolant.py
+++ b/src/pybamm/expression_tree/interpolant.py
@@ -175,7 +175,8 @@ class Interpolant(pybamm.Function):
             np.array(snippet["y"]),
             snippet["children"],
             name=snippet["name"],
-            interpolator=snippet.get("interpolator", "linear"),
+            interpolator=snippet["interpolator"],
+            # the legacy compact format omits these two; default like its decoder did
             extrapolate=snippet.get("extrapolate", True),
             _num_derivatives=snippet.get("_num_derivatives", 0),
         )
--- a/src/pybamm/expression_tree/operations/serialise_kernel.py
+++ b/src/pybamm/expression_tree/operations/serialise_kernel.py
@@ -8,6 +8,7 @@ raises :class:`SerialisationError` rather than silently dropping any value.
 from __future__ import annotations
 
 import importlib
+import inspect
 
 import numpy as np
 
@@ -151,9 +152,95 @@ def decode(node):
     raise SerialisationError(f"Cannot decode value of type {type(node)}")
 
 
-def _lookup_codec(cls: type):
-    return None  # replaced in Task 1.5
-
-
 def normalise_legacy(node: dict) -> dict:
     return node  # replaced in Task 1.6
+
+
+# ---------------------------------------------------------------------------
+# DefaultCodec
+# ---------------------------------------------------------------------------
+
+_CHILD_PARAMS = frozenset({"child", "children", "child_input", "left", "right"})
+_SKIP_PARAMS = frozenset({"self", "domain", "domains", "auxiliary_domains"})
+_MISSING = object()
+
+
+def _read_attr(obj, name):
+    for candidate in (name, f"_{name}"):
+        if hasattr(obj, candidate):
+            return getattr(obj, candidate)
+    return _MISSING
+
+
+class DefaultCodec:
+    """Introspection-based codec: derive fields from ``__init__``.
+
+    Captures every resolvable non-child parameter (including defaulted ones);
+    raises if a *required* parameter cannot be read from the instance.
+    """
+
+    def to_json(self, obj, encode) -> dict:
+        node: dict = {}
+        node["domains"] = obj.domains
+        sig = inspect.signature(type(obj).__init__)
+        # Validate and collect non-child params first so that missing-required-param
+        # errors surface before any encoding attempt (fail-fast, predictable order).
+        extra: dict = {}
+        for name, param in sig.parameters.items():
+            if name in _SKIP_PARAMS or name in _CHILD_PARAMS:
+                continue
+            if param.kind in (param.VAR_POSITIONAL, param.VAR_KEYWORD):
+                raise SerialisationError(
+                    f"{type(obj).__name__}.__init__ uses *args/**kwargs ({name}); "
+                    f"add a to_json/_from_json hook."
+                )
+            value = _read_attr(obj, name)
+            if value is _MISSING:
+                if param.default is inspect.Parameter.empty:
+                    raise SerialisationError(
+                        f"{type(obj).__name__}.__init__ requires {name!r}; no "
+                        f"attribute {name}/_{name} found -- add a to_json/_from_json "
+                        f"hook or register a codec."
+                    )
+                continue
+            extra[name] = value
+        # Encode children and collected params only after validation succeeds.
+        if obj.children:
+            node["children"] = [encode(c) for c in obj.children]
+        for name, value in extra.items():
+            node[name] = encode(value)
+        return node
+
+    def from_json(self, node, decode, cls):
+        children = [decode(c) for c in node.get("children", [])]
+        param_names = set(inspect.signature(cls.__init__).parameters)
+        # Forward only real __init__ params; ignore bookkeeping/legacy extras
+        # (e.g. the old switch's "name") rather than passing them as kwargs.
+        # "domains" is excluded here and handled separately below.
+        kwargs = {
+            key: decode(raw)
+            for key, raw in node.items()
+            if key not in (TAG, "children", "domains") and key in param_names
+        }
+        # to_json always writes node["domains"]; map it onto whichever param the
+        # signature accepts -- `domains` (full dict), `domain` (primary list), or
+        # neither (class infers domain from children). Prevents the silent
+        # domain-drop for `domain`-param classes like CoupledVariable.
+        domains = node.get("domains")
+        if domains:
+            if "domains" in param_names:
+                kwargs.setdefault("domains", domains)
+            elif "domain" in param_names:
+                kwargs.setdefault("domain", domains.get("primary") or None)
+        return cls(*children, **kwargs)
+
+
+_default_codec = DefaultCodec()
+
+
+def _lookup_codec(cls: type):
+    import pybamm  # local import to avoid circular dependency at module level
+
+    if issubclass(cls, pybamm.Symbol):
+        return _default_codec
+    return None  # extended in Task 1.5 with hook detection
--- a/src/pybamm/expression_tree/operations/serialise_kernel.py
+++ b/src/pybamm/expression_tree/operations/serialise_kernel.py
@@ -278,14 +278,14 @@ def _guarded_params(cls: type) -> tuple[str, ...]:
 
 
 def _is_structural(value) -> bool:
-    """True if value is (or contains) a Symbol, so it travels through children
-    and the hook need not emit it as a scalar field."""
+    """True if value is (or contains, at any nesting depth) a Symbol, so it
+    travels through children and the hook need not emit it as a scalar field."""
     if isinstance(value, pybamm.Symbol):
         return True
     if isinstance(value, (list, tuple)):
-        return any(isinstance(v, pybamm.Symbol) for v in value)
+        return any(_is_structural(v) for v in value)
     if isinstance(value, dict):
-        return any(isinstance(v, pybamm.Symbol) for v in value.values())
+        return any(_is_structural(v) for v in value.values())
     return False
 
 
--- a/src/pybamm/expression_tree/operations/serialise_kernel.py
+++ b/src/pybamm/expression_tree/operations/serialise_kernel.py
@@ -73,16 +73,87 @@ def _decode_leaf(node):
         return node
     tag = node.get(TAG)
     if tag == "builtins.float":
-        return _FLOAT_SENTINELS_INV[node["value"]]
+        value = node["value"]
+        if value not in _FLOAT_SENTINELS_INV:
+            raise SerialisationError(f"Unknown float sentinel: {value!r}")
+        return _FLOAT_SENTINELS_INV[value]
     if tag == "numpy.ndarray":
         return np.array(node["data"], dtype=node["dtype"])
     if tag == "builtins.slice":
         return slice(node["start"], node["stop"], node["step"])
     if tag == "builtins.tuple":
-        return tuple(decode(x) for x in node["items"])  # noqa: F821 - decode added in Task 1.3
+        return tuple(decode(x) for x in node["items"])
     raise SerialisationError(f"Not a leaf node: {tag!r}")
 
 
 _LEAF_TAGS = frozenset(
     {"builtins.float", "numpy.ndarray", "builtins.slice", "builtins.tuple"}
 )
+
+
+def _encode_class_ref(cls: type) -> dict:
+    return {TAG: "type", "class": _class_path(cls)}
+
+
+def encode(obj):
+    """Encode any serialisable object into a JSON-compatible structure.
+
+    Raises SerialisationError for objects with no codec (safe-or-loud).
+    """
+    if obj is None or isinstance(obj, (bool, int, str)):
+        return obj
+    if isinstance(obj, float):
+        return _encode_float(obj)
+    if isinstance(obj, np.ndarray):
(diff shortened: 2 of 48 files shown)


src/pybamm/expression_tree/interpolant.py
```diff
@@ -172,9 +175,9 @@ def _from_json(cls, snippet: dict):
             np.array(snippet["y"]),
             snippet["children"],
             name=snippet["name"],
-            interpolator=snippet["interpolator"],
-            extrapolate=snippet["extrapolate"],
-            _num_derivatives=snippet["_num_derivatives"],
+            interpolator=snippet.get("interpolator", "linear"),
+            extrapolate=snippet.get("extrapolate", True),
+            _num_derivatives=snippet.get("_num_derivatives", 0),
```
MarcBerliner: Can we make these fields explicit in `to_json`

src/pybamm/expression_tree/operations/serialise_kernel.py
```diff
@@ -0,0 +1,380 @@
...
+    Contract: only ``children`` are encoded/decoded by the kernel. Every other
+    field a hook emits must already be JSON-native (scalars, or lists/dicts of
+    them) -- it reaches ``_from_json`` undecoded. Route any Symbol-, ndarray-,
+    slice- or tuple-valued argument through ``children``, or have the hook
+    (de)serialise it itself in ``to_json``/``_from_json``.
+    """
+
+    def to_json(self, obj, encode) -> dict:
+        node = dict(obj.to_json())
+        node.pop("id", None)  # identity is not part of the round-trip contract
```
MarcBerliner, line 336: Is the `"id"` field ever used in `from_json`? If so, can we just remove them from all the `to_json`s?

BradyPlanden: I've removed `"id"` from `to_json` but left this pop to ensure historical format continue having `"id"` removed from the round-trip.

src/pybamm/expression_tree/interpolant.py
BradyPlanden: This over-extended a bit since interpolator is included in both the historical and current wire format. For the others, they are in the current `to_json` but need to be filled in for the historical format, which the default provides.

MarcBerliner (dismissed):
Thanks!

MarcBerliner (approved):

PR #5561, opened 2026-06-03
Serialisation kernel: migrate discretised-model, solver, and mesh surfaces

BradyPlanden:
# Description

Second of the stacked serialisation-kernel PRs, built on #5560. With the kernel and the Symbol surface in place, this PR migrates the remaining serialisation surfaces onto it and removes the bespoke encoders, so every surface shares the same safe-or-loud mechanism and one wire format.

This PR covers:

- **Discretised-model surface**: `serialise_model`/`load_model` now encode and decode through the kernel; `_SymbolEncoder`, `_MeshEncoder`, and the bespoke reconstructors are deleted. `Event`, `Mesh`, and `SubMesh` become kernel known bases with kernel-shaped hooks (expressions and sub-meshes carried via children). A read-only relocation pass maps legacy `py/object` nested fields onto the canonical shape, so pre-refactor discretised model files keep loading (regression-tested against an authentic legacy fixture; the fixture is git-ignored for size and the test skips if absent). SPM, SPMe, and DFN round-trip through save, load, and solve.
- **Solvers**: the warn-and-omit branch is replaced by a `SerialisationError` raise, extending safe-or-loud to solver serialisation. All shipped solver classes verified to round-trip; an audit found no solver currently omits anything.
- **Submesh and spatial-method types**: the bespoke `{class, module}` handling is unified onto the kernel's `class_reference` codec (both legacy shapes still read).
- **Coverage meta-test generalised** beyond Symbol to solver and submesh bases, so a new subclass of any known serialisable base without a strategy or explicit exemption fails CI.
- **Contributor doc** ("making a class serialisable") added to the API docs, plus the CHANGELOG entry for the whole kernel refactor.

# Important checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Features
 
 - Electrode electronic conductivity can now be specified as a function of stoichiometry (in addition to temperature) for all lithium-ion and sodium-ion models. ([#5556](https://github.com/pybamm-team/PyBaMM/pull/5556))
-- Unified PyBaMM serialisation onto a single safe-or-loud encode/decode kernel. Serialisation now either round-trips or raises `SerialisationError`, never silently dropping a field, across the expression tree, discretised models, meshes, solvers, experiments and parameter values. There is one canonical on-disk format, and files saved by older PyBaMM versions continue to load via backward-compatible readers. ([#5548](https://github.com/pybamm-team/PyBaMM/issues/5548))
+- Unified PyBaMM serialisation onto a single safe-or-loud encode/decode kernel. Serialisation now either round-trips or raises `SerialisationError`, never silently dropping a field, across the expression tree, discretised models, meshes, solvers, experiments and parameter values. There is one canonical on-disk format, and files saved by older PyBaMM versions continue to load via backward-compatible readers. ([#5560](https://github.com/pybamm-team/PyBaMM/pull/5560)) and ([#5561](https://github.com/pybamm-team/PyBaMM/pull/5561))
 
 ## Breaking changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Features
 
 - Electrode electronic conductivity can now be specified as a function of stoichiometry (in addition to temperature) for all lithium-ion and sodium-ion models. ([#5556](https://github.com/pybamm-team/PyBaMM/pull/5556))
+- Unified PyBaMM serialisation onto a single safe-or-loud encode/decode kernel. Serialisation now either round-trips or raises `SerialisationError`, never silently dropping a field, across the expression tree, discretised models, meshes, solvers, experiments and parameter values. There is one canonical on-disk format, and files saved by older PyBaMM versions continue to load via backward-compatible readers. ([#5548](https://github.com/pybamm-team/PyBaMM/issues/5548))
 
 ## Breaking changes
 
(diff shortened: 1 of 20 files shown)


rtimms (commented):
## Review notes (discretised-model, solver, mesh surfaces)

Solid, well-structured PR. The migration is faithful and the test coverage is strong (the strategy-coverage meta-tests and the legacy-fixture/relocation tests are the right safety nets). Verified locally:

- `tests/unit/test_serialisation/` — **1247 passed, 1 skipped** (the skip is the intentionally git-ignored legacy fixture)
- SPM save → load → **solve** round-trips
- Mesh/submesh round-trip preserves the exact filter the old `_MeshEncoder` used (single-domain, non-ghost) — no regression there
- Solver round-trip incl. `CompositeSolver` recursion

The reordering of `encode` (type/codec dispatch *before* container branches) is correct and necessary now that `Mesh` (a `dict` subclass) is a known base, and it's well-commented.

### Findings

**1. [Medium] `save_model(model, mesh)` now raises for any non-uniform / encode-only submesh — with a misleading message.**

Models meshed with `Exponential1DSubMesh`, `Chebyshev1DSubMesh`, `SpectralVolume1DSubMesh`, `UserSupplied*`, or any 2D non-uniform submesh now make `save_model` **raise** when a mesh is passed. Verified:

```
DFN + Exponential1DSubMesh particle mesh:
  serialise_model(model, mesh=...)  -> SerialisationError
  serialise_model(model)            -> OK
```

Two issues:
- **Behavioural change.** The `mesh` arg is documented as optional and "Not necessary to solve the model … required to use pybamm's plotting tools." Previously `save_model(model, mesh)` *succeeded* (the mesh just couldn't be reloaded). Users who pass a mesh purely for plotting now hit a hard failure on save. Defensible under "safe-or-loud" (it genuinely can't round-trip), but worth a conscious decision + changelog note.
- **The error message is inaccurate.** It reads *"…is a registered serialisable base but defines no `to_json`/`_from_json` hook."* These classes **do** inherit `to_json`; only `_from_json` is missing. The message should name the real cause (no `_from_json` ⇒ cannot round-trip), otherwise a contributor debugging this looks in the wrong place.
- **The exemption comment is also inaccurate.** In `tests/unit/test_serialisation/test_base_strategy_coverage.py`, `_SUBMESH_EXEMPT` labels these as *"encode-only: inherits to_json, no _from_json."* They are not encode-only — `encode()` raises on them. The comment should read "cannot serialise (no `_from_json`)".

Suggested resolution: keep raising (consistent with the PR's philosophy), but (a) fix the message in `serialise_kernel._lookup_codec` to point at the missing `_from_json`, and (b) correct the `_SUBMESH_EXEMPT` comments. If preserving the old "save succeeds, load fails" behaviour matters for plotting workflows, that's a larger design call.

**2. [Low] `_relocate_legacy_model_tree` overwrites `children` when both `sub_meshes` and `children` exist.** Not reachable for real legacy mesh nodes (they carry only `sub_meshes`), so currently harmless — just slightly fragile.

**3. [Low, pre-existing — out of scope] `_reconstruct_pybamm_dict` uses `k.lstrip("symbol_")`** (serialise.py). `str.lstrip` strips a *character set*, not a prefix, so it can over-strip leading characters of a symbol name. Predates this PR (not in the diff) but now sits on the legacy-geometry decode path; worth a follow-up issue.

**4. [Nit] CHANGELOG** entry double-cites `([#5560]) and ([#5561])` — reads a little awkwardly but fine.

### Bottom line
Approve-with-comments. Nothing blocks correctness of the common path (defaults round-trip and solve). Would ask for the message + comment fixes in finding **1** before merge, plus a one-line changelog note that `save_model(model, mesh)` is now strict about non-round-trippable submeshes.

rtimms (dismissed):

rtimms (approved):

PR #5573, opened 2026-06-05
Default 'voltage as a state' to 'true'

BradyPlanden:
# Description

Flips the "voltage as a state" model option default to "true": voltage is solved as an algebraic state and standard lithium-ion models become DAEs. Stacked on #5572; Targets 26.7.0.

Why: reading `Voltage [V]` from a solution becomes an O(1) state lookup instead of a post-solve expression evaluation.

Benchmarks (IDAKLU defaults, internal initialisation; median of 7 cold reps, total = build + solve + observe):

| Scenario | main | this PR |
| --- | --- | --- |
| SPM 1C discharge | 0.056 s | 0.051 s (-9%) |
| SPM 5-cycle experiment | 0.157 s | 0.123 s (-22%) |
| SPMe 1C discharge | 0.098 s | 0.092 s (-6%) |
| SPMe dense output (10k pts) | 0.128 s | 0.095 s (-26%) |
| SPMe 5-cycle experiment | 0.337 s | 0.280 s (-17%) |
| DFN 1C discharge | 0.141 s | 0.122 s (-13%) |
| DFN dense output (10k pts) | 0.161 s | 0.134 s (-17%) |
| DFN 5-cycle experiment | 0.503 s | 0.464 s (-8%) |

With in-solver `output_variables=["Voltage [V]"]` (10k points) the win is larger: on main, computing the voltage expression in-solver roughly doubles SPM/SPMe solve time, while with this PR it is a state slice:

| Model | main | this PR |
| --- | --- | --- |
| SPM | 0.064 s | 0.033 s (-48%) |
| SPMe | 0.127 s | 0.070 s (-45%) |
| DFN | 0.112 s | 0.106 s (-5%) |

Up to 73% faster for very dense output (100k points); 8-24% across cycle counts. DFN is already a DAE on main, so its gains come from the cheaper voltage reads (observe drops 9.5 ms to 2.4 ms on the experiment scenario); final voltages match main to 1e-4.

Migration: `{"voltage as a state": "false"}` restores the previous behaviour; combined with `{"surface form": "false"}` SPM/SPMe remain pure ODE models for ODE-only solvers (`ScipySolver`, `JaxSolver(method="RK45")`). The default `IDAKLUSolver`, `CasadiSolver` (all modes), and `JaxSolver(method="BDF")` handle the DAE form.

## Type of change

CHANGELOG entry under Breaking changes with migration note.

# Important checks:

- [x] All tests pass: `nox -s tests` (affected unit and integration suites verified locally, full matrix in CI)
- [ ] The documentation builds: `nox -s doctests`


martinjrobins:
great speedups :) How does this work? If you have a voltage expression g(x), how does the algebraic equation `0 = V - g(x)` end up being faster than just assigning `V = g(x)` at the end of the step, you still have to calculate g(x)...?

BradyPlanden:
Yeah, it's interesting since the intuition is that the cost of `g(x)` is constant and accumulated within the total solve, independent of where it's processed. The delta is that on main, `g(x`) sits in two hot paths:

  1. Post-solve, once per output point. `sol["Voltage [V]"]` evaluates the full discretised expression at every `t_interp` point via a large casadi graph: 272 nodes for SPM, 1489 for SPMe, including arcsinh2/log/exp from the inverse Butler-Volmer closure and the OCPs. With voltage-as-a-state the same read is a 2-node state slice. The cost scales with output density: at 100k points the SPMe observe is 269 ms on main vs 7.6 ms here, while solve only moves from 86 to 90 ms.
  2. In-solve, in the events. The four voltage cutoff events each embed `g(x)` and are checked at every internal step. With voltage-as-a-state they become 3-node expressions (V − cutoff). 

The algebraic formulation does add a cost as `g(x)` moves into the residual and IDA carries an extra algebraic state. Which is about ~5-10 ms for SPM/SPMe 1C discharge solves, but observe time improves 10-50 ms, and a 1C discharge takes ~140 internal steps vs 10k-100k output points, so the trade wins as soon as voltage is read densely.

_Edit: I was running some low output densities while typing this, and most of the observe cost is actually the expression tree to casadi conversion paid on first access when voltage is an expression (classic!). It's approx. a fixed ~13 ms (SPM) / ~22 ms (SPMe) per solution, nearly flat from 100 to 10k points on my machine._

martinjrobins:
Thanks for the info, I didn't realise about the events. Very good  catch, I never would have thought that adding an extra equation to solve would actually speed things up! :)

BradyPlanden:
Noting that continuous solves of rapidly alternating current profiles (drive-cycle style) **is a regression**. Posting the data so this merges with eyes open.

**Scenario:** 100 x 30 s alternating +-1C square-wave current via Interpolant (10 ms ramps, ~0.8% SOC window), single continuous solve, default IDAKLU. Totals are solve + cold voltage read, medians over 7 reps:</p>

Scenario | main | this PR | change
-- | -- | -- | --
SPM, model events on | 0.076 s | 0.118 s | +54%
SPM, events stripped | 0.046 s | 0.114 s | +146%
SPMe, events on | 0.195 s | 0.209 s | +7%
SPMe, events stripped | 0.122 s | 0.221 s | +81%


The same alternation done as a 200-step Experiment is fine (-16 to -23% with termination clauses) where solver restarts and consistent re-initialisation are not a problem. With output_variables=["Voltage [V]"] the gaps narrow ~10-30% (main pays in-solver expression eval) but do not close, e.g. SPM events-off is still +119%.

The crossover point is ~10-25 reversals for SPM, and ~25-50 for SPMe. Below that, this PR wins or holds parity; so single and few-pulse profiles (GITT-style per-solve) are wins.

The full parity escape hatch is just restoring the voltage to an expression and ensure the surface form is "false", i.e. <code>{"voltage as a state": "false", "surface form": "false"}</code> which gives 0.028 s / 2128 steps on the worst case, identical to main, and remains supported.

I think we should ship the flip with the trade-off disclosed in the CHANGELOG breaking entry and the option docstring. Given experiments, dense output, and output_variables all win by 8-73%, I think the trade is right, but flagging it clearly for opinions.

MarcBerliner (dismissed):

MarcBerliner (approved):

PR #5583, opened 2026-06-09
Fix CasADi conversion of differentiated Interpolant (#5582)

### LINKED ISSUE #5582: [Bug]: CasADI interpolant derivatives return the original function
### PyBaMM Version

main

### Python Version

3.11

### Describe the bug

If I understand correctly, when converting the derivative of an interpolant to CasADI, it returns the original interpolant.

### Steps to Reproduce

Run

```python3
        x = np.linspace(0, 1)
        y = pybamm.StateVector(slice(0, 2))
        casadi_y = casadi.MX.sym("y", 2)
        y_test = np.array([0.4, 0.6])

        # test derivative
        for interpolator in ["cubic", "pchip"]:
            interp = pybamm.Interpolant(x, x ** 2, y, interpolator=interpolator)
            interp_diff = interp.diff(y)
            interp_diff_casadi = interp_diff.to_casadi(y=casadi_y)
            f = casadi.Function("f", [casadi_y], [interp_diff_casadi])
            expected = interp_diff.evaluate(y=y_test)
            print("Interpolant: ", expected)
            print("Casadi: ", f(y_test))
```

The interpolant returns [0.8, 1.2], as expected, but CasADI still returns [0.16, 0.36].

### Relevant log output

```shell

```

MarcBerliner:
## Description

Fixes #5582.

The CasADi conversion of an `Interpolant` ignored `self._num_derivatives`, so a symbolically-differentiated `cubic`/`pchip` interpolant converted to the **original, un-differentiated** function instead of its derivative. scipy's `evaluate()` returned the correct derivative while the CasADi conversion returned the original — a silent correctness bug.

This produced wrong values wherever a tabulated (data-backed) function is differentiated and then evaluated through a CasADi/IDAKLU solver, e.g. the **heat-of-mixing** term and the **differential-capacity hysteresis OCP** with data-based open-circuit potentials, as well as any manual `interp.diff(...).to_casadi(...)`.

Note: the standard solver Jacobian is unaffected — it applies `casadi.jacobian` (AD) to the 0th-derivative interpolant rather than symbolically differentiating it first.

## Changes

- `_cubic_to_casadi`: differentiate the scipy b-spline `_num_derivatives` times before building the CasADi b-spline. A `cubic` differentiated three or more times yields a degree-0 (piecewise-constant) spline that CasADi mis-evaluates at the knots (returns `0`), so that case now raises a clear `NotImplementedError` rather than returning wrong values.
- `_pchip_to_casadi`: analytically differentiate the per-interval Horner coefficients (the stored `x_lo` is left untouched). Because pchip uses a Horner polynomial rather than a b-spline, its derivatives convert correctly at all orders.

## Testing

Added regression tests in `test_interpolant.py` (pytest fixtures + parametrization):
- `test_diff_to_casadi` — cubic/pchip × uniform/non-uniform grids, 1st and 2nd derivatives
- `test_diff_to_casadi_vector_valued` — vector-valued data
- `test_diff_to_casadi_pchip_third_derivative` — pchip 3rd derivative stays correct
- `test_diff_to_casadi_cubic_degree_zero_raises` — cubic 3rd derivative raises

Full `tests/unit/test_expression_tree/` suite passes (398 passed, 1 skipped).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ## Bug fixes
 
+- Fixed CasADi conversion of a differentiated `Interpolant`, which returned the original function. ([#5582](https://github.com/pybamm-team/PyBaMM/issues/5582))
 - Fixed the "explicit power" and "explicit resistance" operating modes, which failed to build with a `ModelError`. These modes now default "voltage as a state" to "true", which breaks the circular dependency between current and voltage (I = P/V), and raise a clear `OptionError` if it is explicitly disabled. ([#5572](https://github.com/pybamm-team/PyBaMM/pull/5572))
 - Fixed `latexify()` crashing with a `NotImplementedError` for models whose boundary conditions contain `boundary_gradient` nodes (e.g. DFN with `{"surface form": "algebraic"}`). ([#5572](https://github.com/pybamm-team/PyBaMM/pull/5572))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ## Bug fixes
 
-- Fixed CasADi conversion of a differentiated `Interpolant`, which returned the original function. ([#5582](https://github.com/pybamm-team/PyBaMM/issues/5582))
+- Fixed CasADi conversion of a differentiated `Interpolant`, which returned the original function. ([#5583](https://github.com/pybamm-team/PyBaMM/pull/5583))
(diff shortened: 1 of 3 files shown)


brosaplanella (dismissed):
Looks good to me, thanks Marc for addressing this so quickly!

MarcBerliner:
@brosaplanella I made one small performance change -- can I please get a re-approval?

brosaplanella (approved):

PR #5587, opened 2026-06-09
Fix two unified-experiment input bugs (Scalar ambient temperature, sensitivity leak)

MarcBerliner:
## Description

Two independent bug fixes for `experiment_model_mode="unified"`, where the experiment injects per-step values as solver input parameters.

### 1. `pybamm.Scalar` ambient-temperature value crashes the solve
The unified model reads ambient temperature as a per-step solver input. If the `"Ambient temperature [K]"` parameter value is a `pybamm.Scalar` (rather than a plain float), it was injected uncoerced, so the stacked input array became `dtype=object` and the casadi function call failed. Fixed by coercing a `pybamm.Scalar` to its float value before building the step inputs.

### 2. `calculate_sensitivities` leaks the experiment's internal control inputs
The experiment injects control inputs (`"Experiment step index"`, `"start time"`, ambient/initial temperature). When a user requested sensitivities — either `calculate_sensitivities=True` or a list that happened to include one of these — they leaked into the parameter Jacobian as if they were user sensitivity targets. Fixed by filtering the internal control inputs (`_INTERNAL_EXPERIMENT_INPUTS`) out of the requested sensitivities.

## Tests
- `test_unified_ambient_temperature_scalar_value`: a `Scalar(298.15)` ambient value now matches a plain `298.15` float solve.
- `test_solve_with_sensitivities_true_excludes_all_internal_inputs`: with a python-function step (so `"start time"` is genuinely a model input), `calculate_sensitivities=True` returns sensitivities only w.r.t. the user input, never the internal control inputs.

Both tests were verified to fail without the corresponding fix and pass with it. The full `test_simulation_with_experiment.py` suite passes (78 passed).

## Type of change


rtimms (approved):
thanks!

PR #5592, opened 2026-06-11
Consolidate PyBaMM and pybammsolvers into a monorepo (#5512)

### LINKED ISSUE #5512: Consolidate PyBaMM and PyBaMM-solvers into a single monorepo
### Summary

Proposal to merge `pybamm-solvers` back into the main PyBaMM repository as a monorepo, using a `packages/` layout with UV workspaces. The original reason for the split (build coupling) is no longer a blocker thanks to `scikit-build-core` and UV workspaces, and the two-repo setup is creating ongoing friction for development and version pinning.

### Background

Splitting solvers out into their own repo was the right call at the time because the C/C++ build couldn't be co-located cleanly with the pure-Python PyBaMM build. That constraint is gone:

- `scikit-build-core` lets each package build independently with its own native components.
- UV workspaces let us develop multiple packages from one checkout with editable installs and a shared lockfile.

In day-to-day work, the two-repo split now mostly causes pain:

- Version pins between PyBaMM and PyBaMM-solvers drift, and bumping them is a coordinated two-PR dance.
- Cross-repo changes (e.g. anything touching the Python ↔ solver boundary) can't be reviewed as a single atomic change.
- Releases are harder to align.

### Proposed structure

```
pybamm/
├── pyproject.toml          # top-level workspace config
├── packages/
│   ├── pybamm/             # current pybamm package
│   └── pybamm-solvers/     # merged in, history preserved
├── ...
```

- A top-level `pyproject.toml` declaring the UV workspace.
- Each subpackage keeps its own `pyproject.toml`, build config, and tests.
- UV workspace sync installs everything in editable mode for development.
- **Separate release workflows** triggered by different tag names (e.g. `pybamm-v*` vs `pybamm-solvers-v*`), so the two packages can continue to release independently on PyPI.
- Git history from `pybamm-solvers` is preserved on the merge (e.g. via `git subtree` or `git filter-repo`).

### Known issue: circular imports

There's an existing circular import between the two packages. The plan:

1. **First pass:** move the code into the monorepo without changing the import structure — just get the layout right.
2. **Cleanup pass:** untangle the circular dependency in follow-up branches. Agents could help here (Brady to lead).

### Dependencies / sequencing

- **Blocked on:** Agriya's `scikit-build-core` PR landing first. Once that's in, this work can start.
- Brady to drive the migration.

### Open questions

- Do we keep two independently-versioned packages on PyPI long-term, or does one of them eventually fold into the other?
- CI matrix — do we still test the two packages independently, or just the workspace as a whole?
- Documentation site layout — single docs build, or two?

cc: @BradyPlanden @martinjrobins @agriyakhetarpal

BradyPlanden:
## Aims

Consolidate PyBaMM and `pybammsolvers` into a single repository (closes #5512) so the two develop, build, and test together — while still releasing **independently** to PyPI. PyBaMM's existing commit history is preserved unchanged; `pybammsolvers` is brought in with its full history via a merge commit (not a copy).

## What changes for developers

- **Layout:** both packages now live under `packages/` — `packages/pybamm/` and `packages/pybamm-solvers/` — as a [UV workspace](https://docs.astral.sh/uv/concepts/projects/workspaces/). The repo root holds shared tooling (workspace `pyproject.toml`, `noxfile.py`, `uv.lock`, `docs/`, CI).
- **Dev setup:** `git submodule update --init` then `uv sync` builds both from source (solver via scikit-build-core). nox sessions and test paths now target `packages/pybamm/...`.
- **Versioning & releases:** independent, discriminated by tag namespace — `pybamm-v*` and `pybamm-solvers-v*` — each with its own publish workflow. See `RELEASE.md`.
- **CI:** path-filtered (dorny/paths-filter). PyBaMM's matrix now tests against the **in-repo** solver on every platform via cached `cibuildwheel` wheels (a reusable workflow shared with the solver release); a smoke job still checks the released PyPI solver to guard the `pip install` path.
- **No user-facing change:** `pip install pybamm` and its `pybammsolvers` dependency are unchanged.

## Merge with a merge commit — do NOT squash

This PR **must be merged with "Create a merge commit"**, not squash or rebase. `pybammsolvers`' history is brought in via an `--allow-unrelated-histories` merge; squashing would collapse those commits into one and destroy the solver's provenance. PyBaMM's own commit SHAs are unchanged by this PR.

## Verification

Clean-checkout gate passed locally: submodules resolve, `uv sync` clean, version scoped correctly to `pybamm-v*`, 49 solver + 3974 pybamm unit tests green, `git log --follow`/`blame` traverse the move, zizmor clean.

> **Note:** the new CI workflows are parse- and logic-validated locally; the cross-job wheel-artifact handoff, the Windows/vcpkg build, and the per-hash wheel cache only fully exercise on a real Actions run — expect to iterate here. Solver PR-time benchmarks are deferred (re-add gated on the solver path filter if wanted).


BradyPlanden:
Future note: we will need to change the PyPi publishing repository to ensure the workflows don't break

MarcBerliner (approved):

PR #5598, opened 2026-06-12
Fix load_custom_model leaving param inconsistent with restored options

aabills:
## Description

`Serialise.load_custom_model` reconstructs a model by calling `base_cls()` (which builds `param` from the **default** options — a single-phase `LithiumIonParameters`) and then restoring the serialised options. The `options` setter only updates the options dict; it does **not** rebuild the cached `param`. For a composite custom model this leaves `param` and `options` inconsistent:

- `options["particle phases"] == ("2", "1")` → pybamm dispatches to the **composite** electrode-SOH solver, but
- `param.n.prim.epsilon_s_av` still resolves to the **single-phase** name `"Negative electrode active material volume fraction"`.

So voltage-based initial-state setup (`set_initial_state("X V")`) on a round-tripped composite model raises:

```
ParameterNotFoundError: Parameter 'Negative electrode active material volume fraction' not found...
you may need to use Primary: Negative electrode active material volume fraction instead.
```

## Fix

After restoring the options in `load_custom_model`, rebuild `LithiumIonParameters` from those options so `param` matches `options`. Scoped to `pybamm.lithium_ion.BaseModel` (lead-acid `param` is options-independent; generic `BaseModel` has no `param`). Single-phase models are unaffected (rebuild is equivalent).

## Testing

- Added `test_custom_composite_model_roundtrip_rebuilds_param`: round-trips a composite (`"particle phases": ("2", "1")`) custom model through the generic-base-class load path and asserts the reloaded `param.n.prim.epsilon_s_av` uses the phase-prefixed `Primary:` name. Verified it fails without the fix and passes with it.
- Existing custom-model / options round-trip tests still pass.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/pybamm/expression_tree/operations/serialise.py
+++ b/src/pybamm/expression_tree/operations/serialise.py
@@ -1393,6 +1393,18 @@ class Serialise:
         opts = model_data.get("options", {})
         if opts is not None:
             model.options = Serialise._convert_options(opts)
+            # The options setter only updates the options dict; it does not
+            # rebuild the cached ``param`` object, which ``base_cls()`` built
+            # from the default options. For lithium-ion models this leaves the
+            # parameters inconsistent with the restored options -- e.g. a
+            # composite electrode ("particle phases": ("2", "1")) would keep
+            # the single-phase parameter names ("Negative electrode active
+            # material volume fraction") instead of the phase-prefixed ones
+            # ("Primary: Negative electrode active material volume fraction"),
+            # which breaks downstream electrode-SOH initialisation. Rebuild the
+            # parameter object so it matches the restored options.
+            if isinstance(model, pybamm.lithium_ion.BaseModel):
+                model.param = pybamm.LithiumIonParameters(model.options)
 
         all_variable_keys = (
             [lhs_json for lhs_json, _ in model_data["rhs"]]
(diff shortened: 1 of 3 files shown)


src/pybamm/expression_tree/operations/serialise.py
```diff
@@ -1393,6 +1393,18 @@ def load_custom_model(filename: str | dict) -> pybamm.BaseModel:
...
+            # from the default options. For lithium-ion models this leaves the
+            # parameters inconsistent with the restored options -- e.g. a
+            # composite electrode ("particle phases": ("2", "1")) would keep
+            # the single-phase parameter names ("Negative electrode active
+            # material volume fraction") instead of the phase-prefixed ones
+            # ("Primary: Negative electrode active material volume fraction"),
+            # which breaks downstream electrode-SOH initialisation. Rebuild the
+            # parameter object so it matches the restored options.
+            if isinstance(model, pybamm.lithium_ion.BaseModel):
+                model.param = pybamm.LithiumIonParameters(model.options)
```
valentinsulzer, line 1407: what happens if model is not a lithium ion model? should we gate?

valentinsulzer (changes requested):

src/pybamm/expression_tree/operations/serialise.py
aabills: > what happens if model is not a lithium ion model? should we gate?

(Replying as an AI agent on @aabills's behalf.)

It's already gated — the `isinstance(model, pybamm.lithium_ion.BaseModel)` check means the rebuild is skipped for any non-lithium-ion model. The gate is deliberate, and lithium-ion is the only built-in family that needs it:

- **lithium-ion**: `self.param = LithiumIonParameters(self.options)` — options-dependent (composite `particle phases` changes the parameter names, e.g. `Primary: ...`). This is the only family with the inconsistency, hence the fix.
- **lead-acid**: `self.param = LeadAcidParameters()` — constructor takes **no** options; parameters don't vary with options, so `base_cls()` already produces a consistent `param`. Nothing to fix.
- **equivalent circuit (Thevenin)**: `self.param = EcmParameters()` — likewise no options arg, already consistent.
- **generic `pybamm.BaseModel`**: has no options-dependent `param` (and the `isinstance` check avoids touching it).

I also intentionally avoided the more general `model.param = type(model.param)(model.options)`: that would raise `TypeError` for lead-acid/ECM, since `LeadAcidParameters()` / `EcmParameters()` don't accept an `options` argument.

One acknowledged edge case: a user-defined custom model with its own *options-dependent* parameter class wouldn't be rebuilt — but `load_custom_model` can't generically know how to reconstruct an arbitrary parameter class, and lithium-ion is the only built-in family where this matters. Happy to broaden the gate (e.g. lead-acid/ECM no-arg rebuilds, or a registry) if you'd prefer a more explicit/future-proof form.

valentinsulzer (approved):

aabills:
superseded by #5599

PR #5599, opened 2026-06-12
Harden serialisation round-trip tests + #5598

BradyPlanden:
# Description

Harden serialisation round-trip tests; fix param-rebuild altitude and MSMR option gaps

Follow-up to the tuple-options fix, addressing issues found while hardening the serialisation test suite.

Rebuild param at the right altitude. The previous fix rebuilt a loaded lithium-ion model's `param` with an `isinstance` branch inside `load_custom_model`. Replace it with a model-owned `_rebuild_param` hook (no-op on `BaseModel`, overridden by the lithium-ion base model, which now also uses it in `__init__`). `load_custom_model` calls the hook, so the options setter stays free of param concerns and the fix lives where the model owns its derived state (#5598).

Fix two MSMR option-validation gaps surfaced by option fuzzing. Both let an invalid option set crash later in parameter construction with "invalid literal for int() with base 10: 'none'" instead of raising a clean OptionError:
- the all-or-nothing MSMR check missed MSMR inside a per-electrode tuple (`("MSMR", "single")`) because it compared the whole value to "MSMR";
- an all-MSMR model left on the default "number of MSMR reactions" ("none") passed validation, since the registry only lists that placeholder.

Harden the serialisation test suite:
- generative fuzzing of valid option dicts through to_json/from_json, with the construction guard narrowed to (OptionError, NotImplementedError) so real bugs surface instead of being silently rejected;
- deterministic tuple-option and int-dimensionality round-trip coverage;
- discretised round-trip now asserts structural identity, solution equivalence and event preservation, plus a non-default tuple option.

Fixes # (issue)

## Type of change

# Important checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@
 
 - The "voltage as a state" model option now defaults to "true": voltage is solved as an algebraic state, making standard models DAEs. Reading `Voltage [V]` from a solution is now an O(1) state lookup instead of a post-solve expression evaluation (up to 73% faster end-to-end for dense output; 8-24% faster experiment cycling across SPM/SPMe/DFN; up to 48% faster with in-solver `output_variables=["Voltage [V]"]`). The default `IDAKLUSolver`, `CasadiSolver` (all modes), and `JaxSolver(method="BDF")` handle DAEs; ODE-only solvers (`ScipySolver`, `JaxSolver(method="RK45")`) require `{"voltage as a state": "false", "surface form": "false"}` for SPM/SPMe, which remains a supported configuration. Known trade-off: continuous solves of rapidly alternating current profiles (e.g. interpolant drive cycles with more than ~25 current reversals in one solve) can be slower, up to ~2x for SPM in the worst measured case; the legacy configuration above restores previous performance for these workloads. Experiment-driven cycling is unaffected. ([#5573](https://github.com/pybamm-team/PyBaMM/pull/5573))
 
+## Bug fixes
+
+- Fixed `Serialise.load_custom_model` leaving a loaded lithium-ion model's cached `param` built from default options instead of the restored ones, so a model loaded with non-default options (e.g. composite `"particle phases"`) had parameters inconsistent with its options. ([#5599](https://github.com/pybamm-team/PyBaMM/pull/5599))
+- Fixed `BatteryModelOptions` letting two invalid MSMR option sets pass validation and then crash during parameter construction with `invalid literal for int() with base 10: 'none'`; both now raise a clear `OptionError`. The all-or-nothing MSMR check now detects MSMR inside a per-electrode tuple (e.g. `"open-circuit potential": ("MSMR", "single")`), and an MSMR model must set `"number of MSMR reactions"` to a positive integer rather than the default `"none"`. ([#5599](https://github.com/pybamm-team/PyBaMM/pull/5599))
+
 # [v26.6.1.1](https://github.com/pybamm-team/PyBaMM/tree/v26.6.1.1) - 2026-06-11
 
 ## Bug fixes
--- a/src/pybamm/expression_tree/operations/serialise.py
+++ b/src/pybamm/expression_tree/operations/serialise.py
@@ -1386,13 +1386,12 @@ class Serialise:
         model = base_cls()
         model.name = model_data["name"]
         model.schema_version = schema_version
-        # Restore options so round-trip serialisation produces an equivalent
-        # model. A JSON round-trip turns tuple-valued options (e.g. "particle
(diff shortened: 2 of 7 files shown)


aabills (commented):
Review summary. This thread tracks concerns **@aabills** raised — including the MSMR edge case he suspected — with diagnosis from a review pass done on his behalf (agent findings marked below). The core fix is correct — the #5598 symptom is gone, verified by round-tripping a composite `("particle phases": ("2","1"))` model and confirming `param.n.prim.epsilon_s` resolves to the `Primary:`-prefixed name. The concerns are about *where* the fix lives and whether the tests actually guard it.

---

### 1. `_rebuild_param` having to be called is clunky — @aabills

Agreed, and it's a latent bug, not just style. The hook lives at the right altitude (the model owns its derived `param`), but it fires from *callers* of the options setter rather than the setter itself. The invariant "`param` matches `options`" then holds only if every present and future caller remembers the second call. Today there are two callers; any new one silently reintroduces #5598.

Verified the trap on this branch:
```python
m = pybamm.lithium_ion.SPM(build=False)
m.options = full_composite_opts
m.param.n.prim.epsilon_s.name
# -> "Negative electrode active material volume fraction"  (stale — no "Primary:")
```
The new `test_rebuild_param_syncs_param_with_reassigned_options` actually *documents* this broken intermediate state (it asserts the model is inconsistent until the hook is hand-called) rather than removing it.

**Suggestion:** call the hook from the options setter itself (the `self._options = options` line in `BatteryModelOptions`'s setter, plus the plain `BaseModel.options` setter). Then `lithium_ion.BaseModel.__init__` and `load_custom_model` both drop their explicit `_rebuild_param()` calls, and the inconsistent state becomes unrepresentable. Options are assigned once per construction (checked all call sites), so no extra rebuild cost.

### 2. Not sure this addresses the original concern in the tests — @aabills

Confirmed — the new param-consistency assertion is a tautology. `_assert_param_matches_options` does `dict(model.param.options) == dict(model.options)`, but after the rebuild `model.param.options is model.options` (same object — verified `True`). It's comparing an object to itself. It can only ever detect "the hook wasn't called at all"; it can never detect "the parameters are wrong." A refactor that broke phase-prefixed naming in `LithiumIonParameters` would keep this suite green.

The original #5598 test asserted the actual symptom — that the reloaded composite model's `param.n.prim.epsilon_s_av` uses the `Primary:` name — and that assertion didn't carry over here.

**Suggestion:** re-add the symptom-level assertion (check the resolved parameter name, not `options`-dict equality), and ideally one end-to-end `set_initial_state("X V")` smoke on a round-tripped composite model, since that voltage-based init was the real user-facing failure in #5598.

---

### 3. MSMR count check rejects a config that builds on `main` — @aabills (suspected the edge case), agent confirmed

The new `number of MSMR reactions` check flattens per-electrode structure and over-rejects. `_is_positive_integer_count` requires *every* entry of the count tuple to be a positive int whenever *any* electrode requests MSMR, so a mixed-electrode config is rejected:

```python
opts = {
    "open-circuit potential": ("MSMR", "single"),
    "particle": ("MSMR", "Fickian diffusion"),
    "intercalation kinetics": ("MSMR", "symmetric Butler-Volmer"),
    "number of MSMR reactions": ("3", "none"),   # positive electrode is not MSMR
    "surface form": "differential",
}
pybamm.lithium_ion.DFN(opts)   # builds AND is well-posed on main; raises OptionError on this branch
```

The `"none"` is the non-MSMR electrode's placeholder and is never `int()`-ed — the per-domain code only parses the count where that electrode is actually MSMR (`base_kinetics.py:86`, `msmr_diffusion.py:525`, `lithium_ion_parameters.py:877`). `_is_msmr` flattens the same way, so the all-or-nothing check also fires across electrodes that should be validated independently.

**Suggestion:** validate per electrode rather than over the flattened tuple. Note the existing `options.negative[opt]` / `options.positive[opt]` accessors can't be used here — they read from `self`, which isn't populated until `super().__init__(options.items())` at the end of validation — so resolve each electrode's value from the local `options` dict (scalar applies to both; 2-tuple is `(neg, pos)`). For each electrode: if any of OCP/particle/intercalation-kinetics is MSMR then all three must be (keep this gated to `working electrode == "both"` to preserve the half-cell loosening), and require a positive-integer count only for an electrode that is MSMR. That drops both `_is_msmr` and `_is_positive_integer_count`. Also add a *positive* test that mixed-electrode MSMR is accepted — the new MSMR tests only assert rejection.

---

Common thread on the tests (agent): the new suite is green but not guarding — tautology in (2), documented-not-fixed state in (1), and only the rejection half of MSMR covered in (3). The fix's correctness is real; the tests just wouldn't catch a regression of any of these.

aabills (approved):

PR #5600, opened 2026-06-12
Fix RegulariseSqrtAndPower corrupting state-independent bases

MarcBerliner:
## Description

`RegulariseSqrtAndPower` (the exchange-current density post-processor added in #5371) replaced **every** `Sqrt`/`Power` node with `RegPower`, including bases that are independent of the state. With the default `scale=1` and `delta=1e-3`, this corrupts values below the regularisation tolerance: a `pybamm.Parameter` rate constant of ~1e-9 raised to the power 0.5 evaluated **~470x too small** (`6.7e-5` → `1.4e-7`).

The result: solver failures (Newton/IDACalcIC diverges) or silently wrong overpotentials (~0.5 V instead of ~5 mV at C/20) for any user-supplied exchange-current density function that raises a `Parameter`/`InputParameter` to a fractional power — a common pattern for parameterised electrochemistry.

A plain Python float (`4.52e-9 ** 0.5`) is computed eagerly and never becomes a symbolic `Power` node, so this only bit symbolic parameter bases.

## Fix

Only replace a `Sqrt`/`Power` node with `RegPower` when its base **either** matches a registered scale **or** is state-dependent (contains a `Variable`, `StateVector`, or `Time`). State-independent bases have no derivative to regularise and are left unchanged.

This deliberately differs from "only regularise matched-scale bases": built-in parameter sets (`ORegan2022`, sodium-ion `Chayambuka2022`) use **normalised** bases like `(c_e / c_e_ref) ** (1 - alpha)` that match no registered scale. Keying off state-dependence keeps those regularised exactly as before (`RegPower` with default scale 1), so shipped parameter sets are unaffected. Manually-created `reg_power` nodes (MSMR `j0_j`) are `Function`s, not `Sqrt`/`Power`, so they are never touched.

| Base | Before | After |
|---|---|---|
| `Parameter("rka") ** 0.5` (~1e-9) | corrupt 470x | ✓ exact |
| `(c_e / c_e_ref) ** 0.5` (ORegan/Na) | ✓ RegPower s=1 | ✓ RegPower s=1 |
| `c_e ** 0.5` (matched scale) | ✓ RegPower s=c_e_init | ✓ unchanged |
| manual `reg_power(...)` (MSMR) | ✓ preserved | ✓ preserved |

## Testing

- Unit: 7 new tests in `test_regularise.py` (state-independent `Parameter`/`InputParameter` bases left alone; normalised state-dependent bases still regularised; matched-scale wiring pinned; manual `reg_power` preserved; full j0 pipeline produces 3 `RegPower` / 0 plain `Power` for Chen2020/ORegan2022/Ecker2015). Full unit suite: 4194 passed.
- Integration: li-ion `test_dfn` + `test_compare_outputs` + sodium-ion (64 passed). Parameter-based vs float-based DFN agree to ~2e-5 V; ORegan2022 `dj0/dc_e` bounded (0.317) down to `c_e=0`; MSMR DFN solves end-to-end.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # [Unreleased](https://github.com/pybamm-team/PyBaMM/)
 
+## Bug fixes
+
+- `RegulariseSqrtAndPower` no longer regularises state-independent bases, fixing corrupted small rate constants in exchange-current density functions. ([#XXXX](https://github.com/pybamm-team/PyBaMM/pull/XXXX))
+
 ## Breaking changes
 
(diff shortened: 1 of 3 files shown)


brosaplanella (approved):
Looks good to me, thanks Marc!

PR #5601, opened 2026-06-12
Preserve termination provenance in unified experiment mode

martin1cifuentes:
## Summary

This PR preserves termination provenance when running experiments in unified experiment mode.

In the unified path, multiple step controls can be collapsed into a shared model branch to reduce model-building work. That works well for execution, but when termination information is reported back to the user, some of the original experiment step context can be lost. The goal of this change is to keep that context available so termination reporting still reflects the relevant experiment step.

The underlying execution behaviour remains the same.

## What changed

The implementation now carries forward the experiment step context needed for termination reporting in unified mode, while remaining compatible with the existing branch collapsing behaviour. I also added test coverage around unified experiment termination reporting to make sure the original context is preserved.

## Why

Unified mode simplifies some of the internal execution mechanics, but termination reporting is part of the user facing behaviour. Even if different controls share the same execution path internally, the reported termination reason should still be traceable to the experiment step that produced it.

## Behaviour preservation

This is intended to be a behaviour preserving change focused on retaining provenance information.

Experiment execution and solver behaviour should remain unchanged. The only difference is that the termination context is preserved and reported correctly when using the unified path.

## Tests

Added and updated focused unit tests covering termination reporting in experiment simulations.

### DIFF
--- a/packages/pybamm/src/pybamm/experiment/step/base_step.py
+++ b/packages/pybamm/src/pybamm/experiment/step/base_step.py
@@ -472,7 +472,10 @@ class BaseStep:
         return events
 
     def get_combined_termination_expression(self, variables):
-        events = self.get_termination_events(variables)
+        return self._combine_termination_events(self.get_termination_events(variables))
+
+    @staticmethod
+    def _combine_termination_events(events):
         if not events:
             return pybamm.Scalar(1)
 
--- a/packages/pybamm/src/pybamm/simulation/simulation.py
+++ b/packages/pybamm/src/pybamm/simulation/simulation.py
@@ -470,7 +470,7 @@ class Simulation(BaseSimulation):
         active_step_index = inputs[self._STEP_INDEX_INPUT]
         if isinstance(active_step_index, np.ndarray):
             active_step_index = active_step_index.item()
-        return self._experiment_termination_event_names_by_index.get(active_step_index)
+        return self._experiment_termination_event_names_by_index[active_step_index]
 
     def _get_step_termination_event_candidates(self, step, model, inputs):
         candidates = []
@@ -483,9 +483,6 @@ class Simulation(BaseSimulation):
             return candidates
 
         source_event_names = self._get_unified_termination_event_names(inputs)
-        if source_event_names is None or len(source_event_names) != len(candidates):
-            return candidates
-
         # Prefer the names recorded when the unified branch was built; the events
         # below are reconstructed only to evaluate which source termination crossed.
         return [
--- a/packages/pybamm/src/pybamm/simulation/simulation.py
+++ b/packages/pybamm/src/pybamm/simulation/simulation.py
@@ -467,15 +467,9 @@ class Simulation(BaseSimulation):
         return value
 
     def _get_unified_termination_event_names(self, inputs):
-        active_step_index = inputs.get(self._STEP_INDEX_INPUT)
-        if active_step_index is None:
-            return None
-
-        try:
-            active_step_index = int(np.asarray(active_step_index).reshape(-1)[0])
-        except (IndexError, TypeError, ValueError):
-            return None
-
+        active_step_index = inputs[self._STEP_INDEX_INPUT]
+        if isinstance(active_step_index, np.ndarray):
+            active_step_index = active_step_index.item()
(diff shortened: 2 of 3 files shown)


src/pybamm/simulation/simulation.py
```diff
@@ -454,16 +466,46 @@ def __missing__(self, key):
...
+        active_step_index = inputs.get(self._STEP_INDEX_INPUT)
+        if active_step_index is None:
+            return None
+
+        try:
+            active_step_index = int(np.asarray(active_step_index).reshape(-1)[0])
+        except (IndexError, TypeError, ValueError):
+            return None
+
+        return self._experiment_termination_event_names_by_index.get(active_step_index)
```
MarcBerliner: I believe `_STEP_INDEX_INPUT` input parameter should always be a scalar parameter since we define it internally

src/pybamm/simulation/simulation.py
```diff
@@ -454,16 +466,46 @@ def __missing__(self, key):
...
+        candidates = []
+        for term in step.termination:
+            event = term.get_event(model.variables, step)
+            if event is not None:
+                candidates.append((event.name, event, term))
+
+        source_event_names = self._get_unified_termination_event_names(inputs)
+        if source_event_names is None or len(source_event_names) != len(candidates):
+            return candidates
+
```
MarcBerliner: It looks like some of the checks for `x is None` in this function and `_get_unified_termination_event_names` will only trigger if the experiment mode is unified. It'd be nice to use `Simulation`'s `self._experiment_uses_unified_model` property to make it clearer what branch we're testing, without relying on fallbacks

tests/unit/test_experiments/test_simulation_with_experiment.py
```diff
@@ -1143,6 +1143,46 @@ def test_run_multi_termination_step_unified_matches_legacy(self):
...
+        combined = f"event: {unified_sim._COMBINED_TERMINATION_EVENT}"
+        legacy_hold = legacy_sol.cycles[0].steps[1]
+        unified_hold = unified_sol.cycles[0].steps[1]
+
+        assert legacy_sol.termination == expected
+        assert unified_sol.termination == expected
+        assert legacy_hold.termination == expected
+        assert unified_hold.termination == expected
+        assert unified_sol.termination != combined
+        assert unified_hold.termination != combined
```
MarcBerliner: Is it now not possible to hit the `_COMBINED_TERMINATION_EVENT` event message?

MarcBerliner (commented):
Hi @martin1cifuentes, looking great so far! I'm making this a partial review since you mentioned expanding this PR's scope -- could you elaborate on what you're thinking of updating?

src/pybamm/simulation/simulation.py
martin1cifuentes: I was treating step-index input as a defensive external input boundary, but youre right. Since PyBaMM defines it internally for unified experiments, it shouldnt handle malformed internal input as a normal falback. Ill revise this.

src/pybamm/simulation/simulation.py
martin1cifuentes: Got it, thanks. I was using absence of provenance data as a proxy, but we have the _experiment_uses_unified_model to tell us what branch we are testing.

tests/unit/test_experiments/test_simulation_with_experiment.py
martin1cifuentes: The intended behaviour is that when the source step termination is resolvable, unified mode should project that concrete event name rather than the internal combined event name. Ill change this to a positive assertion against the expected termination so that the scope is clearer, thanks.

martin1cifuentes:
Thanks for the review @MarcBerliner. I checked the fallback path again and I believe this PR should stay focused on the resolvable provenance case rather than expanding it. 

I also checked a small follow up, where unified_branch_repr() includes custom terminations through their repr, but CustomTermination equality/hash distinguish the event function. So two same named custom termination with different functions can have the same unified branch identity even though they are semantically different. I suspect this can cause unified mode to collapse distinct branches that legacy mode would treat differently. Ill keep that separate from this PR, but it seems like a good focused follow up around unified branch identity.

PR #5612, opened 2026-06-16
Fix incomplete BPX import for interpolated-table OCP branches (#5610)

### LINKED ISSUE #5610: [Bug]: Incomplete BPX import with interpolated table for new parameters
### PyBaMM Version

26.6.1.1

### Python Version

3.12

### Describe the bug

For newly added BPX parameters in BPX v1.1 with the `FloatFunctionTable` type, processing to PyBaMM objects is incomplete when they contain interpolated tables. This is because the targets of the last processing stage are written manually and were not updated by #5463 #5507. This renders the `pybamm.ParameterValues` output of `create_from_bpx_[...]` unusable, because the parameters do not have an expression tree value.

Detailed issue:

- `parameter_values.bpx.process_float_function_table` will output a tuple `(name, (x, y))` when acting on a `bpx.InterpolatedTable`.
- These tuples are manually converted to valid PyBaMM expression tree objects using `functools.partial` acting on  `parameter_values.bpx._interpolant_func` as a boilerplate interpolant. The arguments to the interpolant are parameter specific.
- BPX v1.1 added two new parameters that support this type: `"OCP (lithiation) [V]"` and `"OCP (delithiation) [V]"`, for the two OCP branches.
- If present as interpolated tables, the tuples need converting, as is done for the main branch `"OCP [V]".

### Steps to Reproduce

No publicly available repeat available at short notice but I can create a dummy example if required.

**To fix:**

```python
# ocp
U = pybamm_dict[phase_domain_pre_name + "OCP [V]"]
if isinstance(U, tuple):
    pybamm_dict[phase_domain_pre_name + "OCP [V]"] = partial(
        _interpolant_func, name=U[0], x=U[1][0], y=U[1][1]
    )
```

could be replaced by something like

```python
# ocp branches
for branch in ["", "delithiation ", " lithiation "]:
  ocp_param = phase_domain_pre_name + branch + "OCP [V]"
  U = pybamm_dict.get(ocp_param, None)  # skip missing branches  
  if isinstance(U, tuple):
      pybamm_dict[ocp_param] = partial(
          _interpolant_func, name=U[0], x=U[1][0], y=U[1][1]
      )
```

### Relevant log output

```shell

```

rtimms:
Fixes #5610.

## Problem

BPX `FloatFunctionTable` fields are parsed into placeholder tuples `(name, (x, y))` that must be converted back into `pybamm.Interpolant` objects in a final processing stage. As the issue notes, the conversion targets were written manually and were not updated when #5463/#5507 added the new BPX v1.1 branches `OCP (lithiation) [V]` and `OCP (delithiation) [V]`. When supplied as interpolated tables, those parameters were left as raw tuples, making the resulting `ParameterValues` unusable (no expression-tree value).

## Fix

Rather than add two more entries to the hand-maintained list (which is the root cause), this generalises the last stage:

- Extract a small `_table_to_interpolant` helper that names the `(name, (x, y))` placeholder contract.
- Replace the per-parameter conversion blocks (OCP, entropic change, user-defined) with a single catch-all that converts **any** remaining placeholder tuple just before returning.
- Reuse the helper at the temperature-dependent sites (particle/electrolyte diffusivity, electrolyte conductivity), which stay explicit because they wrap the interpolant inside an Arrhenius dependence.

New 1:1 table parameters now need no hand-maintained conversion entry. The only tuples remaining in the dict at that point are interpolant placeholders (temperature-dependent params are already wrapped into partials above); this invariant is documented inline.

## Tests

Strengthened `test_bpx_hysteresis_names` to assert the lithiation/delithiation OCP branches are converted to usable callables rather than left as raw tuples (it previously only checked key presence). Existing `test_table_data` already covers the main OCP and entropic-change table paths. Full BPX suite (24 tests) passes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Bug fixes
 
+- Fixed BPX import leaving interpolated-table OCP hysteresis branches (`OCP (lithiation) [V]`, `OCP (delithiation) [V]`) as raw data tuples instead of usable expression-tree objects. Remaining `FloatFunctionTable` placeholders are now converted generically, so new BPX table parameters no longer need a hand-maintained conversion list. ([#5610](https://github.com/pybamm-team/PyBaMM/issues/5610))
 - `RegulariseSqrtAndPower` no longer regularises state-independent bases, fixing corrupted small rate constants in exchange-current density functions. ([#5600](https://github.com/pybamm-team/PyBaMM/pull/5600))
 
 ## Breaking changes
--- a/src/pybamm/parameters/bpx.py
+++ b/src/pybamm/parameters/bpx.py
@@ -18,6 +18,12 @@ def _interpolant_func(var, name, x, y):
     return pybamm.Interpolant(x, y, var, name=name, interpolator="linear")
 
 
+def _table_to_interpolant(value):
+    """Wrap a ``(name, (x, y))`` FloatFunctionTable placeholder in an interpolant."""
+    name, (x, y) = value
+    return partial(_interpolant_func, name=name, x=x, y=y)
+
+
 preamble = "from pybamm import exp, tanh, cosh\n\n"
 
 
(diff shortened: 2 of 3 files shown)


BradyPlanden (approved):
LGTM - thanks!

MarcBerliner (approved):
Thanks!

PR #5574, opened 2026-06-05
BPX v0.x backward compatibility (via bpx) + optional State support

rtimms:
## Description

Fixes #5571.

PyBaMM v26.5.0 upgraded `bpx` `0.5.0` → `1.1.0` (#5469), adopting the BPX **v1.x** schema (a top-level `State` block, with the initial/ambient temperature and initial electrolyte concentration moved out of `Parameterisation`). This was an **undocumented breaking change**: BPX v0.x files that loaded in 26.4 now raised a `ValidationError`.

Rather than carry a converter inside PyBaMM, the v0.x→v1.x conversion lives in **bpx itself** (the format library owns its own migration). bpx also made all `State` fields optional, so this PR adapts PyBaMM's BPX reader accordingly.

Both changes shipped in **bpx 1.1.1**, so this PR now depends only on a **released** pin and is ready to merge. (The releasable initial-state-of-charge handling was split into #5616 and has already merged into `main`.)

## What this PR does

1. **Bumps the lower `bpx` pin to `>=1.1.1,<1.2.0`** (`packages/pybamm/pyproject.toml`), the first release carrying the v0.x conversion + optional `State`.
2. **Relies on bpx for v0.x conversion.** `create_from_bpx` / `create_from_bpx_obj` just call `bpx.parse_bpx_file` / `parse_bpx_obj`, which auto-detect and convert legacy v0.x inputs (emitting a `UserWarning`). No PyBaMM-side converter. PyBaMM officially supports `bpx>=1`.
3. **Supports the now-optional BPX `State` section.** Any `State` field — or the whole section — may be absent:
   - Ambient/initial temperature default to the cell's **reference temperature**; the initial electrolyte concentration (the reference `c_e0` used to normalise the exchange-current density) defaults to **1000 mol.m⁻³**. Each substitution is logged (`pybamm.logger.info`) and summarised in a single `UserWarning`.
   - Initial hysteresis state and heat transfer coefficient are **opt-in**, so they are left unset for the model to default — no warning.

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] Bug fix (mitigates the undocumented 26.5.0 breaking change)

## Tests

[...]

PR #5616, opened 2026-06-16
Apply BPX State 'Initial state-of-charge' instead of ignoring it

rtimms:
## Description

`ParameterValues.create_from_bpx` / `create_from_bpx_obj` previously **discarded** the BPX `State` `Initial state-of-charge` and always initialised the cell at full charge. This PR makes them honor it.

The methods now seed the stoichiometric-limit full charge as before (negative phases at `theta_max`, positive at `theta_min`), then apply the BPX initial SOC via `set_initial_state` — so blended/composite electrodes are supported — after validating it lies in `[0, 1]`.

This PR also **retroactively documents** the `bpx` `0.5.0` → `1.1.0` upgrade (#5469) as a breaking change in the changelog: it adopted the BPX v1.x schema, so v0.x files stopped validating. (The backward-compatible *fix* for that — see #5574 below — is not yet releasable, so it is not referenced from this changelog entry.)

## Why this is split out

This is the subset of the BPX work on #5574 that runs against the **currently released** `bpx` (`>=1.1.0,<1.2.0`): every BPX v1.x file carries a required `State` `Initial state-of-charge`, so no unreleased `bpx` feature is needed.

The remaining BPX work — tolerating an absent/partial `State` section and converting legacy v0.x files — both require an as-yet-unreleased `bpx`. In released `bpx` 1.1.0 the `State` section is effectively mandatory for a full parameterisation (a model validator rejects its absence) and every `State` field is required, so neither behaviour can ship yet. Those stay on **#5574**, which is stacked on top of this PR and remains `[DO NOT MERGE]` until `bpx` cuts a release.

## Testing

- `tests/unit/test_parameters/test_bpx.py` adds coverage for the initial SOC being applied (single and blended electrodes) and for an out-of-range value raising.
- Validated the suite passes against the **released** `bpx==1.1.0`, not just the local dev checkout.


rtimms:
### Status

Releasable now against the currently-released `bpx` (`>=1.1.0,<1.2.0`) — no dependency on an unreleased `bpx`.

**Scope (intentionally narrow):**
- Honors the BPX `State` `Initial state-of-charge` via `set_initial_state` (blended-electrode support; validated to `[0, 1]`).
- Retroactively records the `bpx` 0.5.0→1.1.0 breaking change in the changelog.

**What is deliberately *not* here** (kept on the stacked #5574, both gated on the next `bpx` release): the v0.x→v1.x conversion and the optional-`State` handling. I confirmed against released `bpx==1.1.0` that those genuinely can't ship yet — the `State` section is effectively required and all its fields are mandatory there.

Latest `main` is merged in (incl. #5610), tests pass against released `bpx==1.1.0`, and `ruff` is clean.

BradyPlanden (approved):

PR #5624, opened 2026-06-17
Deprecate ScipySolver, CasadiSolver, CasadiAlgebraicSolver; make IDAKLUSolver the universal solver

MarcBerliner:
## Summary

Steers all users to `IDAKLUSolver` by deprecating the legacy solvers and removing them from the docs.

- **Deprecations** (`DeprecationWarning` + docstring `.. note::`): `ScipySolver`, `CasadiSolver` → use `IDAKLUSolver`; `CasadiAlgebraicSolver` → use `NonlinearSolver`. The `root_method="casadi"` string is also deprecated (→ `"nonlinear_solver"`); it still works but warns, and the redundant inner `CasadiAlgebraicSolver` warning is suppressed.
- **IDAKLU now solves every model.** Re-enabled the lead-acid `LOQS` Jacobian (the `use_jacobian = False` flag was a vestigial 2019 performance default, not a singularity workaround — confirmed by git archaeology + numerics; matches `CasadiSolver` to ~2e-6 V) and dropped its `CasadiSolver` default override, so `LOQS` inherits the global `IDAKLUSolver` default like every other ODE/DAE model.
- **Docs scrubbed of the deprecated solvers**: 18 example notebooks (code swapped to IDAKLU and re-executed for clean outputs), `public_api.rst`, the `03-pybamm-solvers` notebook (reworked from a Casadi-vs-IDAKLU comparison into an IDAKLU timing demo), and the example scripts. Deleted `compare_dae_solver.py`.
- **AGENTS.md**: always use `IDAKLUSolver`; don't pass `solver=` to `Simulation` without a concrete reason (leave it bare for the default); read solutions via the interpolating call `solution["Voltage [V]"](t)`, not `.data`/`.entries`.
- **Tests**: a `pytest.warns` test per deprecated class and for the `root_method="casadi"` kwarg.

The three classes are deprecated, not removed — they keep working this release. CHANGELOG entry added under Breaking changes.


docs/source/examples/notebooks/models/lithium-plating-composite.ipynb
BradyPlanden, line 1: It looks like there has been some path leakage here in the notebook. If it's easy to remove, that is probably worthwhile - I would have thought we would already scrub this away via pre-commit

BradyPlanden (approved):
Looks great, just one comment on path leakage. It's non-blocking, as we should add a pre-commit scrub for this anyway

docs/source/examples/notebooks/models/lithium-plating-composite.ipynb
MarcBerliner: Which leakage are you referring to? I see the `/Users/marcberliner/Documents/GitHub/PyBaMM/src/pybamm/simulation.py:12` which was deleted, but I don't see other ones

BradyPlanden: Ah, yep I didn't register that it was a deletion! Gtg

PR #5630, opened 2026-06-26
Re-implement benchmarks in pytest

pipliggins:
# Description

Moves PyBaMM's speed and peak memory tests away from asv and into the pytest ecosystem, using pytest-benchmark and pytest-memray.

Adds a new workflow to check that the speed benchmarks in a PR do not regress compared to main.

To discuss: What to do with [pybamm-bench](https://github.com/pybamm-team/pybamm-bench). This stores the asv benchmarks over time and provides a gh-pages website to display them. The workflow to populate the repo hasn't run since April 2024 so the site is fairly out of date.
One option is to use the https://github.com/benchmark-action/github-action-benchmark, which should be able to provide a similar workflow and display plots in a similar manner to asv, should we want to keep some record.


.github/workflows/benchmarks_pr.yml
```diff
@@ -0,0 +1,234 @@
+name: Benchmarks (PR)
```
martinjrobins, line 8: how about the old benchmark CI workflows? (think there is one in benchmark_on_push.yml), should these be removed

packages/pybamm/tests/benchmarks/README.md
```diff
@@ -0,0 +1,40 @@
...
+`--benchmark-compare-fail=mean:125%` exits with an error if any benchmark is more than 25% slower than the baseline.
+
+### Markers
+
+- `@pytest.mark.memory_bench` — memory benchmarks (run via `benchmark-memory`).
+
+### CI
+
+- **`benchmarks_pr.yml`** — triggers on PRs touching `packages/pybamm/`. Runs speed benchmarks against main HEAD (fails on >25% regression) and memory benchmarks with memray (fails if `limit_memory` thresholds are exceeded).
+- **`benchmarks_history.yml`** — triggers on push to `main`. Runs the full suite and publishes results to the `gh-pages` branch for time-series visualization via [github-action-benchmark](https://github.com/benchmark-action/github-action-benchmark).
```
martinjrobins: is this for the future?

packages/pybamm/tests/benchmarks/test_unit_benchmarks.py
```diff
@@ -0,0 +1,99 @@
...
+    benchmark.pedantic(_discretise, setup=setup, rounds=2000, iterations=1)
+
+
+def test_solve(benchmark):
+    R, model = _create_expression()
+    r, geometry = _parameterise(R, model)
+    _discretise(r, geometry, model)
+
+    def run():
+        solver = pybamm.ScipySolver()
```
martinjrobins: think this solver is deprecated now?

packages/pybamm/tests/benchmarks/test_model_options.py
```diff
@@ -0,0 +1,243 @@
+"""
+Benchmarks for different model option combinations.
+
+All tests here are marked `slow` — the full option sweep is large.
```
martinjrobins:
```suggestion
All tests here are marked `speed_bench` — the full option sweep is large.
```

martinjrobins (changes requested):
thanks @pipliggins, this looks good, a few questions below.  https://github.com/benchmark-action/github-action-benchmark looks really good, can you please add these plots? (good find, I think I might add this to diffsol too!)

packages/pybamm/tests/benchmarks/README.md
brosaplanella: Tagging @agriyakhetarpal as this would be nice to have as a tab on the website (not urgent, but if you can open an issue for someone to pick up that would be great!).

agriyakhetarpal (commented):
Thanks! I only did a quick pass just yet; it would be useful for us to set up these new benchmarks in such a way that we ensure they are as reliable as possible. In conda, we used CodSpeed (https://codspeed.io) until very recently – it is a good solution to eliminate statistical outliers from GitHub Actions runners and has a free tier. There is also https://bencher.dev, which is also free, which we've started using. I would recommend setting up either of those to go with this PR, for a start (I am not opposed to https://github.com/benchmark-action/github-action-benchmark at all, just that it might make sense to adopt these solutions in addition for their low variability).

We already have a 301 redirect set up for the GitHub Pages website at https://pybamm.org/benchmarks, so I can add that in the footer somewhere or change the link wherever our new location will be.

pipliggins:
> Thanks! I only did a quick pass just yet; it would be useful for us to set up these new benchmarks in such a way that we ensure they are as reliable as possible. In conda, we used CodSpeed (https://codspeed.io) until very recently – it is a good solution to eliminate statistical outliers from GitHub Actions runners and has a free tier. There is also https://bencher.dev, which is also free, which we've started using. I would recommend setting up either of those to go with this PR, for a start (I am not opposed to https://github.com/benchmark-action/github-action-benchmark at all, just that it might make sense to adopt these solutions in addition for their low variability).
> 
> We already have a 301 redirect set up for the GitHub Pages website at https://pybamm.org/benchmarks, so I can add that in the footer somewhere or change the link wherever our new location will be.

I saw Bencher come up a lot as I was googling around this, it looks good. We'd need an account associated with pybamm to 'claim' the project and view results - @brosaplanella is that something you could do? A potential issue with the free tier is it limits you to a 5m job timeout. On my mac it takes 4.30 - 5 mins to run the full suite, so we're likely to be close to/over that limit.

packages/pybamm/tests/benchmarks/README.md
pipliggins: Removed the `benchmarks_history.yml` item for now, as we're looking into Bencher

brosaplanella:
> I saw Bencher come up a lot as I was googling around this, it looks good. We'd need an account associated with pybamm to 'claim' the project and view results - @brosaplanella is that something you could do? A potential issue with the free tier is it limits you to a 5m job timeout. On my mac it takes 4.30 - 5 mins to run the full suite, so we're likely to be close to/over that limit.

I have created a Bencher account for PyBaMM and invited both of you (Pip and Agriya), let me know if you didn't receive the invites. We can try the free tier for now and, if needed, we can upgrade later on. There should be some funds in the PyBaMM account to pay for it.

agriyakhetarpal:
I received the invite and accepted it – thanks @brosaplanella!

pipliggins:
Thanks @brosaplanella - could you put the generated API key (think it's the one called "PyBaMM's key") in the repository secrets please?

brosaplanella:
Done, I labelled it as BENCHER

pipliggins:
Bencher Bare Metal (to avoid the github runner outliers) is now wired up.

One snag you'll see in a couple of the comments is that when running on Bencher's bare metal hardware, the full time-benchmarking suite exceeds the free tier's 5 minute limit. I've currently got around this by skipping the slower running tests in CI (using the `slow_bench` marker), just to prove everything works; obviously this isn't ideal going forward though. I also can't check some of the actions yet as the pytest benchmarks don't currently exist on main.

Potential options for Bencher going forward:

1. **Leave as is** - free, but a subset of the timing benchmarks won't be caught by PR checks.
2. **Paid tier**
3. **Split per module** - We could split the tests into a few groups/by module and run them separately to fit into the free tier time limit. The free tier allows only 1 concurrent job so they'd queue, and we'd get a separate report and check per module which might get messy.

I'd lean towards merging the PR as is, then making any further changes for Bencher, to get the pytest benchmarks into main and at least some checks running.

martinjrobins (dismissed):
looks great, thanks @pipliggins. I'm happy to merge as is for now, and follow up with a separate PR trimming down the benchmarks until we have a representative sample over the main parameter space we want to benchmark against. Perhaps we can speed things up by using a smaller t_final or a shorter experiment

martinjrobins (approved):

PR #5634, opened 2026-06-29
Fix internal link errors

pipliggins:
Fixes various errors in the [url checker](https://github.com/pybamm-team/PyBaMM/actions/workflows/lychee_url_checker.yml) for internal links which weren't updated during/after the monorepo merge.

The remaining error in `scripts/update_version.py` is because no pybamm-v{version} tags (the new naming convention after the monorepo merge) have been generated yet.


BradyPlanden (approved):
Thanks @pipliggins!

PR #5642, opened 2026-07-07
Fix deprecated "electrode diffusivity" alias silently clobbering "particle diffusivity"

rtimms:
## Description

Setting or fitting `"<phase> particle diffusivity [m2.s-1]"` could have **zero effect** on the solved result: perturbing it changed nothing, because the deprecated `"<phase> electrode diffusivity [m2.s-1]"` alias silently overrode it. This bites hardest with BPX, since `create_from_bpx` produced a set containing **both** names for the same quantity.

### Root cause

`ParameterValues.check_parameter_values` migrated the deprecated name to the current one by *adding* the new key without removing the deprecated one, and by unconditionally overwriting the current name:

```python
values[new_param] = values.get(param)   # never deletes `param`; clobbers new_param
```

So a set could hold both names, and any re-normalisation (`ParameterValues(dict(pv))`, `.update(...)`, serialisation round-trips common in fit harnesses) let the stale deprecated value overwrite a value the user set on the current name.

### Fix

- **`check_parameter_values`**: the deprecated diffusivity name now migrates with the **current name taking precedence** (`setdefault`) and the deprecated key **removed**, so the two can never coexist and clobber each other. The `FuzzyDict` lookup fallback is unchanged, so reads of the old name still resolve.
- **`bpx.py`**: emit the current `particle diffusivity` name (and its activation energy) directly, so `create_from_bpx` yields a clean single-name set with no spurious `DeprecationWarning`.
- Added regression tests for the migration (current name wins; deprecated key dropped) and for the BPX output.

## Type of change

## Verification

- After `ParameterValues({"Negative electrode diffusivity [m2.s-1]": D})`, `keys()` contains only the `particle` name; setting `particle` then round-tripping `ParameterValues(dict(pv))` preserves the set value (no clobber).
- `create_from_bpx` output contains only `particle diffusivity` names and emits no diffusivity deprecation warning.
- Affected test files pass in isolation: `test_bpx.py`, `test_parameter_values.py`, `test_util.py`, `test_parameter_values_serialisation.py`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 ## Bug fixes
 
+- Fixed the deprecated `"<X> electrode diffusivity [m2.s-1]"` alias silently overriding the current `"<X> particle diffusivity [m2.s-1]"`: `ParameterValues` now migrates the deprecated name to the current one and removes the deprecated key (current name wins), so setting or fitting `particle diffusivity` is no longer a silent no-op. `create_from_bpx` now emits only the current `particle diffusivity` name. ([#5643](https://github.com/pybamm-team/PyBaMM/pull/5643))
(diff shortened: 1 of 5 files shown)


BradyPlanden (approved):
Thanks!

PR #5648, opened 2026-07-14
Feature: Add Positive and Negative EIS calculation

Ombrini:
# Description

Adds three-electrode EIS support to `pybamm.EISSimulation`.

When `three_electrodes=True`, `EISSimulation` now inserts a default reference electrode at the separator midpoint if one has not already been inserted. The EIS setup promotes the positive and negative 3E potentials to algebraic probe variables, returns named impedance components in the `EISSolution`, and keeps the default `solution.impedance` as the cell impedance.

`EISSolution.nyquist_plot()` now detects three-electrode impedance components and plots the cell, positive electrode, and negative electrode curves together.

## Type of change

Feature.

# Important checks:


packages/pybamm/src/pybamm/simulation/eis_simulation.py
```diff
@@ -40,17 +44,28 @@ def __init__(
         var_pts=None,
         spatial_methods=None,
         skip_surface_form_check=False,
+        three_electrodes=False,
```
rtimms, line 47: Could we infer the available EIS components from the configured model, rather than add a `three_electrodes` mode flag? The model already owns reference-electrode configuration, including its physical position; EIS can simply return the 3E impedances for whichever 3E potential variables are present. This keeps the normal case unchanged, supports half-cells naturally, and avoids `EISSimulation` modifying the model passed by the user.

packages/pybamm/src/pybamm/simulation/eis_simulation.py
```diff
@@ -40,17 +44,28 @@ def __init__(
...
+        self.three_electrodes = three_electrodes
+        if three_electrodes and (
+            "Positive electrode 3E potential [V]" not in model.variables
+            or "Negative electrode 3E potential [V]" not in model.variables
+        ):
+            reference_position = (
+                parameter_values["Negative electrode thickness [m]"]
+                + parameter_values["Separator thickness [m]"] / 2
+            )
+            model.insert_reference_electrode(reference_position)
```
rtimms, line 63: I think this insertion should remain an explicit model-level action. It lets the user choose the reference position and keeps the EIS simulation observational rather than mutating its input.

The intended public UX would be:

```python
model = pybamm.lithium_ion.SPM(
    options={"surface form": "differential"}
)
model.insert_reference_electrode(position=reference_position)

solution = pybamm.EISSimulation(model).solve(frequencies)
solution.impedance  # Cell impedance
solution["Positive electrode impedance [Ohm]"]
solution["Negative electrode impedance [Ohm]"]
```

If only one 3E potential is present, return only that electrode's named impedance alongside the cell impedance.

packages/pybamm/src/pybamm/simulation/eis_simulation.py
```diff
@@ -233,9 +279,15 @@ def _calculate_impedance(frequency, M, neg_J, b):
         """
         A = 1.0j * 2 * np.pi * frequency * M + neg_J
         x = spsolve(A, b)
-        # Voltage is penultimate, current is last (by construction in
-        # _set_up_model_for_eis)
-        return -x[-2] / x[-1]
+        if three_electrodes:
+            z_cell = -x[-4] / x[-1]
```
rtimms, line 283: Can we resolve these entries from the built model's `y_slices` rather than from their position at the end of `x`? The current order is deterministic for this exact construction, but it is not a general `model.variables` ordering contract; a future model transformation could silently return the wrong component. This should also replace the corresponding `self._cached_b[-1]` assignment, so the current perturbation and all voltage probes are addressed by their named state slices.

rtimms (changes requested):
Thanks for this! This is a welcome feature, but I think the public API can be simpler and the probe extraction more robust (this is actually a pre-existing issue, but we should fix it here).

PR #5663, opened 2026-07-21
Expose per-value parameter serialisation publicly

rtimms:
## Description

PyBaMM can currently serialise a parameter set only as a whole `ParameterValues` object: `convert_parameter_values_to_json` iterates the object's items and dispatches per value (callables are traced to a symbolic expression, everything else is encoded directly). The per-value dispatch and a raw-mapping entry point are not public, so serialising a plain `{name: value}` mapping (or a single value) means either re-implementing the dispatch or constructing a throwaway `ParameterValues` (which has construction side effects and cannot hold arbitrary values).

This PR exposes the dispatch:

- New public `pybamm.serialize_parameter_value(name, value)` and `pybamm.deserialize_parameter_value(value)` — the per-value serialise/deserialise dispatch, usable on a single value or a raw mapping.
- `convert_parameter_values_to_json` now accepts any `Mapping` (not just a `ParameterValues`); it only ever needs `.items()`, so `Mapping` is the honest type.
- `ParameterValues.from_json` now delegates its per-value work to `deserialize_parameter_value`, so the dispatch lives in one place (mirroring how `convert_parameter_values_to_json` delegates to `serialize_parameter_value`).

Behaviour is unchanged for existing callers — the new helpers reproduce the exact branch logic that was previously inlined.

## Type of change

## Tests

Added round-trip tests for the new helpers (scalar/symbol/callable/interpolant), raw-mapping-vs-`ParameterValues` parity for `convert_parameter_values_to_json`, and per-value-vs-whole-mapping agreement. Full serialisation suite (70 tests) passes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 - Reverted the "voltage as a state" default back to "false", undoing the [v26.7.0.0](https://github.com/pybamm-team/PyBaMM/tree/pybamm-v26.7.0.0) change to "true". Promoting voltage to an algebraic state turns SPM/SPMe into DAEs and places voltage under solver error control, which caused IDAKLU error-test failures on solves of discontinuous (pulsed) current profiles at the default tolerances, and lowered voltage-output accuracy. The option remains available (`{"voltage as a state": "true"}`) and is still enabled automatically for the "explicit power" and "explicit resistance" operating modes. ([#5670](https://github.com/pybamm-team/PyBaMM/pull/5670))
 
+## Features
+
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 
 ## Features
 
+- Exposed the per-value parameter serialisation dispatch publicly as `pybamm.serialize_parameter_value`/`pybamm.deserialize_parameter_value`, and `convert_parameter_values_to_json` now accepts any `Mapping` (not just a `ParameterValues`), so a raw `{name: value}` mapping can be serialised without constructing a throwaway `ParameterValues`. ([#5663](https://github.com/pybamm-team/PyBaMM/pull/5663))
```
BradyPlanden: Can you move this up to Unreleased

BradyPlanden (dismissed):
Looks good, thanks!

MarcBerliner (approved):

PR #5670, opened 2026-07-22
Revert "voltage as a state" default to "false"

### LINKED ISSUE #5669: Restore default `voltage as a state` to `false`
I've got a few reports of numerical issues with the `VaaS == true` default that landed in `v26.7.0.0`. The quick switch is to use the escape hatch `"voltage as a state" = "false"`, but after looking at these issues, I think it's best revert this change and release a patch release. Below is what I've confirmed so far for completeness.

- Solver failures can occur for hard discontinuous current interpolants, as we set `suppress_algebraic_error=False` in the IDAKLU config.
- Voltage tolerances have regressed as the default IDAKLU tolerances (atol:1e-4/rtol:1e-6) are defined for O(1) state variables, so a voltage expression resolving to ~3.5V will have ~3.5x more error.

BradyPlanden:
# Description

The v26.7.0.0 default of "true" promotes voltage to an algebraic state, turning SPM/SPMe into DAEs and placing voltage under IDAKLU error control. This causes error-test failures (IDA_ERR_FAIL) on solves of discontinuous (pulsed) current profiles at the default tolerances, and lowers voltage-output accuracy.

Revert the default to "false". The option remains available and is still auto-enabled for the "explicit power" and "explicit resistance" operating modes. Tests that assert DAE-specific behaviour now opt in explicitly.

Fixes #5669 

## Type of change

# Important checks:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # [Unreleased](https://github.com/pybamm-team/PyBaMM/)
 
+## Breaking changes
+
+- Reverted the "voltage as a state" default back to "false", undoing the [v26.7.0.0](https://github.com/pybamm-team/PyBaMM/tree/pybamm-v26.7.0.0) change to "true". Promoting voltage to an algebraic state turns SPM/SPMe into DAEs and places voltage under solver error control, which caused IDAKLU error-test failures on solves of discontinuous (pulsed) current profiles at the default tolerances, and lowered voltage-output accuracy. The option remains available (`{"voltage as a state": "true"}`) and is still enabled automatically for the "explicit power" and "explicit resistance" operating modes. ([#5670](https://github.com/pybamm-team/PyBaMM/pull/5670))
+
 # [v26.7.0.0](https://github.com/pybamm-team/PyBaMM/tree/pybamm-v26.7.0.0) - 2026-07-21
 
 ## Breaking changes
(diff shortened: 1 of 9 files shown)


MarcBerliner (dismissed):

MarcBerliner (approved):

NicolaCourtier:
If you want to change the default `"voltage as a state"` to `"true"`, I found you also need to set the IDAKLU option `"suppress_algebraic_error"` to `True` to avoid the errors you mention - see PyBOP [PR 967](https://github.com/pybop-team/PyBOP/pull/967). If I understand correctly, the accuracy should be very similar, but discontinuities in the output variable(s) will be ignored. I would suggest switching this option on by default anyway, to improve solver stability for existing DAE models such as the DFN.

@BradyPlanden @MarcBerliner @brosaplanella

BradyPlanden:
Thanks! I also somewhat noted it in https://github.com/pybamm-team/PyBaMM/issues/5669. There is still a scalar applied to the voltage state variable error when applying `VaaS == 'true'` due to the magnitude difference between the voltage state and the other variables coupled with the default solver tolerances, which can accumulate in cost functions.

For `suppress_algebraic_error == true` as the default, this is actually discouraged in Sundials ([docs](https://sundials.readthedocs.io/en/v6.1.0/idas/Usage/SIM.html#c.IDASetSuppressAlg:~:text=is%20discouraged%20when%20solving%20DAE%20systems%20of%20index%201)) for DAE's of index-1 as the error term is useful and controlled by the solver. I somewhat suspect that if the voltage state was scaled to O(1), this probably wouldn't be needed and solver would be able to converge, although I haven't tested.

PR #5690, opened 2026-07-31
Add unstructured 2D/3D DFN battery models

aabills:
## Summary
- Add `BasicDFN2DUnstructured` (quad or triangle elements) and `BasicDFN3DUnstructured` (hexahedron or tetrahedron elements), the DFN physics of `BasicDFN2D` discretised with `FiniteVolumeUnstructured` on the built-in `UnstructuredMeshGenerator`. Both take an `element_type` argument; the defaults are the TPFA-orthogonal quad and hexahedron.
- Carry two small expression-tree changes from the full branch: `Max`/`Min` now derive from a `Reduction` base that clears domains and reports a scalar shape, and `domain_size` knows the primary/secondary particle domains. The models solve without them, but the rest of the stack builds on them.

## Stack
1. #5686 — VectorField N-comp (merged)
2. #5687 — Meshing (merged)
3. #5688 — Spatial method + ProcessedVariable (merged)
4. #5689 — VTK plotting
5. **This PR** (#5690) — Unstructured DFN models
6. #5691 — Deprecate `Magnitude`

After this stack merges, #5397 can be closed as superseded.

## Test plan
- [x] Unit: both models build through `Simulation` for every element type, so the default geometry, spatial methods, submesh types and var_pts are exercised under the coverage job; `Max`/`Min` clear domains and have scalar shape.
- [x] Integration: 2D unstructured (quad) voltage matches `BasicDFN2D` to 5 mV over a 1 h discharge; 3D unstructured (hexahedron) matches the 2D unstructured model to 5 mV.
- [x] Lithium-ion model, expression-tree, mesh and simulation unit suites pass locally.

### DIFF
--- a/packages/pybamm/src/pybamm/expression_tree/symbol.py
+++ b/packages/pybamm/src/pybamm/expression_tree/symbol.py
@@ -45,12 +45,20 @@ def domain_size(domain: list[str] | str):
     fixed_domain_sizes = {
         "current collector": 3,
         "negative particle": 5,
+        "negative primary particle": 5,
+        "negative secondary particle": 5,
         "positive particle": 7,
+        "positive primary particle": 7,
+        "positive secondary particle": 7,
         "negative electrode": 11,
         "separator": 13,
         "positive electrode": 17,
         "negative particle size": 19,
+        "negative primary particle size": 19,
+        "negative secondary particle size": 19,
         "positive particle size": 23,
+        "positive primary particle size": 23,
+        "positive secondary particle size": 23,
     }
     if domain in [[], None]:
         size = 1
--- a/packages/pybamm/tests/unit/test_models/test_full_battery_models/test_lithium_ion/test_basic_models.py
+++ b/packages/pybamm/tests/unit/test_models/test_full_battery_models/test_lithium_ion/test_basic_models.py
@@ -25,3 +25,11 @@ class TestBasicModels:
     def test_dfn_2d(self):
         model = pybamm.lithium_ion.BasicDFN2D()
         model.check_well_posedness()
+
+    def test_dfn_2d_unstructured(self):
+        model = pybamm.lithium_ion.BasicDFN2DUnstructured(element_type="quad")
+        model.check_well_posedness()
+
+    def test_dfn_3d_unstructured(self):
+        model = pybamm.lithium_ion.BasicDFN3DUnstructured()
+        model.check_well_posedness()
--- a/packages/pybamm/tests/unit/test_models/test_full_battery_models/test_lithium_ion/test_basic_models.py
+++ b/packages/pybamm/tests/unit/test_models/test_full_battery_models/test_lithium_ion/test_basic_models.py
@@ -47,3 +47,11 @@ class TestBasicModels:
 
         component = solution["Electrolyte current density x [A.m-2]"]
         assert np.all(np.isfinite(component(t=5)))
+
+    def test_dfn_2d_unstructured(self):
+        model = pybamm.lithium_ion.BasicDFN2DUnstructured(element_type="quad")
+        model.check_well_posedness()
+
+    def test_dfn_3d_unstructured(self):
+        model = pybamm.lithium_ion.BasicDFN3DUnstructured()
+        model.check_well_posedness()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ## Features
 
+- Added basic unstructured 2D/3D DFN models (`BasicDFN2DUnstructured`, `BasicDFN3DUnstructured`). ([#5690](https://github.com/pybamm-team/PyBaMM/pull/5690))
 - Added VTK-based plotting (`VTKQuickPlot`) for unstructured mesh solutions, including headless CI OpenGL setup. ([#5689](https://github.com/pybamm-team/PyBaMM/pull/5689))
 - Added `FiniteVolumeUnstructured` spatial method and unstructured processed-variable support for cell-centered data on arbitrary meshes. The TPFA Laplacian carries an implicit non-orthogonal correction (`"non-orthogonal correction"` option: `"over-relaxed"` or `"minimum"`) and gradients use a least-squares reconstruction, so both are exact on linear fields and second-order on skewed triangle and tetrahedral meshes. Diffusion coefficients reach faces through the distance-weighted harmonic mean, as in `FiniteVolume`, so material interfaces carry the exact series flux. ([#5688](https://github.com/pybamm-team/PyBaMM/pull/5688))
 - Added unstructured mesh support (`UnstructuredSubMesh`, generators, and interface coupling) for arbitrary 2D/3D domains. Hexahedra must have planar faces (warped hexes raise a `GeometryError`), and `UserSuppliedUnstructuredMesh` accepts tetrahedral, triangular, and quadrilateral cells only. ([#5687](https://github.com/pybamm-team/PyBaMM/pull/5687))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Features
 
+- Added basic unstructured 2D/3D DFN models (`BasicDFN2DUnstructured`, `BasicDFN3DUnstructured`).
 - Added VTK-based plotting (`VTKQuickPlot`) for unstructured mesh solutions, including headless CI OpenGL setup.
 - Added `FiniteVolumeUnstructured` spatial method and unstructured processed-variable support for cell-centered data on arbitrary meshes.
 - Generalised `VectorField` to N components and added `Component`/`Norm` operators for multi-dimensional vector fields.
--- a/docs/source/api/models/lithium_ion/dfn.rst
+++ b/docs/source/api/models/lithium_ion/dfn.rst
@@ -13,4 +13,10 @@ Doyle-Fuller-Newman (DFN)
 .. autoclass:: pybamm.lithium_ion.BasicDFNHalfCell
     :members:
 
+.. autoclass:: pybamm.lithium_ion.BasicDFN2DUnstructured
+    :members:
+
+.. autoclass:: pybamm.lithium_ion.BasicDFN3DUnstructured
+    :members:
+
 .. footbibliography::
--- a/packages/pybamm/src/pybamm/__init__.py
+++ b/packages/pybamm/src/pybamm/__init__.py
@@ -168,7 +168,6 @@ from .meshes.scikit_fem_submeshes_3d import (
     UserSuppliedSubmesh3D,
 )
 
-
 from .meshes.unstructured_submesh import (
     UnstructuredSubMesh,
     UnstructuredMeshGenerator,
--- a/packages/pybamm/src/pybamm/expression_tree/functions.py
+++ b/packages/pybamm/src/pybamm/expression_tree/functions.py
@@ -719,8 +719,32 @@ def log10(child: pybamm.Symbol):
     return log(child, base=10)
 
 
-class Max(SpecificFunction):
-    """Max function."""
+class Reduction(SpecificFunction):
+    """Base class for reduction operations that collapse a spatial
+    field to a scalar (e.g. max, min).  Automatically clears domains
+    and returns scalar shape."""
+
+    def __init__(self, function: Callable, child: pybamm.Symbol):
+        super().__init__(function, child)
+        self.clear_domains()
+
+    @classmethod
+    def _from_json(cls, snippet: dict):
+        """See :meth:`pybamm.SpecificFunction._from_json()`.
+
+        ``SpecificFunction._from_json`` bypasses ``__init__``, so the domains
+        inherited from the child have to be cleared again here.
+        """
+        instance = super()._from_json(snippet)
+        instance.clear_domains()
+        return instance
+
+    def _evaluate_for_shape(self):
+        return np.nan * np.ones((1, 1))
+
+
+class Max(Reduction):
+    """Max function (reduction to scalar)."""
 
     def __init__(self, child):
         super().__init__(np.max, child)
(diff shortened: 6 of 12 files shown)


packages/pybamm/src/pybamm/expression_tree/symbol.py
```diff
@@ -45,12 +45,20 @@ def domain_size(domain: list[str] | str):
     fixed_domain_sizes = {
         "current collector": 3,
         "negative particle": 5,
+        "negative primary particle": 5,
+        "negative secondary particle": 5,
```
MarcBerliner: I think all the numbers on this list should be different primes

packages/pybamm/src/pybamm/models/full_battery_models/lithium_ion/basic_dfn_unstructured.py
```diff
@@ -0,0 +1,397 @@
...
+)
+
+
+class BasicDFNUnstructured(BaseModel):
+    """Doyle-Fuller-Newman (DFN) model discretised with
+    :class:`~pybamm.FiniteVolumeUnstructured`.
+
+    The ``"dimensionality"`` option is the number of directions resolved besides
+    the through-cell *x*: 1 (default) meshes (x, z) with quads and 2 meshes
+    (x, y, z) with hexahedra. Pass ``submesh_types`` to
```
rtimms, line 18: The docstring advertises tetrahedra, but a 3D model on a tetrahedral mesh does not solve at this model's own default resolution:

```python
m = pybamm.lithium_ion.BasicDFNUnstructured({"dimensionality": 2})
st = m.default_submesh_types
for d in ["negative electrode", "separator", "positive electrode"]:
    st[d] = pybamm.UnstructuredMeshGenerator(element_type="tetrahedron")
pybamm.Simulation(m, submesh_types=st).solve([0, 1800])
```

```
pybamm.SolverError: IDA_CONV_FAIL: Convergence test failures occurred too many
times during one step or with |h| = hmin
[ERROR] At t = 0.639735311493912 and h = 5.63131609204565e-67, the corrector
convergence failed repeatedly or with |h| = hmin.
```

It also fails at `{x_n: 8, x_s: 8, x_p: 8, y: 2, z: 2}` with `IDA_ERR_FAIL`; only very coarse meshes (x <= 3) converge. Quads, hexahedra **and** triangles all solve fine at default resolution, so this looks specific to tets on the highly anisotropic Kuhn-decomposed box (x ~ 1e-4 m vs z ~ 6e-2 m) — the build warns of 90 degrees of non-orthogonality on every domain.

Either fix the tet path, or drop tetrahedra from the docstring and say why (and see the comment on the unit test, which is what lets this through CI).

packages/pybamm/tests/unit/test_models/test_full_battery_models/test_lithium_ion/test_basic_models.py
```diff
@@ -81,3 +81,49 @@ def test_dfn_2d_vector_field_variable(self):
...
+        for domain in ["negative electrode", "separator", "positive electrode"]:
+            submesh_types[domain] = pybamm.UnstructuredMeshGenerator(
+                element_type=element_type
+            )
+        sim = pybamm.Simulation(
+            model,
+            submesh_types=submesh_types,
+            var_pts={k: 3 for k in model.default_var_pts},
+        )
+        sim.build()
```
rtimms, line 105: This stops at `sim.build()`, so the `triangle` and `tetrahedron` parametrisations never exercise a solve — which is why the tetrahedron failure above is green in CI. The test plan says the model "builds in 2D and 3D through `Simulation` for every element type", and that is exactly the gap: building a tet model succeeds, solving it does not.

A short `sim.solve([0, 60])` here would have caught it.

packages/pybamm/src/pybamm/models/full_battery_models/lithium_ion/basic_dfn_unstructured.py
```diff
@@ -0,0 +1,397 @@
...
+            "Positive electrode surface concentration [mol.m-3]": c_s_surf_p,
+            "Positive electrode surface stoichiometry": sto_surf_p,
+            "Positive electrode surface potential difference [V]": delta_phi_p,
+            "Negative electrode surface potential difference [V]": delta_phi_n,
+            "Positive electrode overpotential [V]": eta_p,
+            "Negative electrode overpotential [V]": eta_n,
+            "Positive electrode ocp [V]": self.param.p.prim.U(sto_surf_p, T),
+            "Negative electrode ocp [V]": self.param.n.prim.U(sto_surf_n, T),
+            "Positive electrode current density [A.m-2]": j_p,
+            "Negative electrode current density [A.m-2]": j_n,
```
rtimms, line 312: `"Negative electrode current density [A.m-2]"` (and `"Positive ..."` on the line above) is PyBaMM's standard name for the **solid-phase electronic** current density `i_s` — it is read as such in `base_electrode.py:160` and `base_thermal.py:84` — but here it holds the **interfacial** current density `j`.

A user moving from `pybamm.lithium_ion.DFN` to this model and reading `sol["Negative electrode current density [A.m-2]"]` silently gets a different physical quantity (zero-mean `j`, rather than `i_s` ramping from `i_cell` to 0 across the electrode), with no error. The standard name for `j` is `"Negative electrode interfacial current density [A.m-2]"`.

This is inherited from `basic_dfn_2d.py:370`, so fixing both or neither is a fair call — but it is a new occurrence here. Nice to see `"Sum of volumetric interfacial current densities [A.m-3]"` used correctly for `a_j` on line 299, which `BasicDFN2D` gets wrong.

rtimms (commented):
Reviewed against `ufv-3-plotting`. **The physics checks out** — I ran the new unit and integration tests plus some independent probes:

- Electrolyte and solid lithium inventory at `t=0` match `pybamm.lithium_ion.DFN` to 1e-15 relative, so the `width = L_y` / `width = 1` slice scaling is right in both 2D and 3D.
- Voltage matches the 1D `DFN` to 2.1e-4 V over a 30 min discharge.
- Lithium is conserved to ~4e-15 under **z**-graded active-material loading (the tests only grade in x) in both 2D and 3D, and the 2D/3D `phi_s_p` spreads agree to 1e-10.
- All 35 declared output variables process finitely in both dimensionalities.
- Default `var_pts` solve out of the box in <1 s for both dimensionalities.
- The model unit suite (908 tests) and the 8 new integration tests pass.

Three comments inline. The only substantive one is that the tetrahedron path the docstring advertises does not actually solve.

Nits not worth inline comments:
- `L_scale = L_x**2 * L_z**2` drops `L_y` for `dimensionality: 2`. Conditioning only — it converges fine — but `L_y` is ~24x `L_z` for Chen2020, so including it may be worth a thought.
- `"Voltage expression [V]"` is missing, where every other `Basic*` model defines it. Unreachable today, since basic models never go through `build_model`, so purely a consistency thing.
- The two integration-test models are constructed at collection time and mutated in the test body. Harmless under xdist, but it means they are built for every session, including ones that deselect these tests.
- The blank-line deletion in `packages/pybamm/src/pybamm/__init__.py` is unrelated to this PR.

packages/pybamm/src/pybamm/models/full_battery_models/lithium_ion/basic_dfn_unstructured.py
aabills: Thanks for the reproducer. The cause is `FiniteVolumeUnstructured._COS_THETA_FLOOR = 0.05`: it caps the over-relaxed weight `1/cos(theta)` at 20. The Kuhn tets of this slab have internal faces at cos ~ 1e-3, so the implicit part is under-weighted and the explicit cross term takes over, which makes the electrolyte diffusion block unstable. The fix is #5774, which removes the floor and is about to merge into `main`. I will merge `main` into the stack once it lands, so the docstring stays as it is.

packages/pybamm/tests/unit/test_models/test_full_battery_models/test_lithium_ion/test_basic_models.py
aabills: Fixed in 6a32fe5c1: the test now calls `sim.solve([0, 60])` for all four element types and checks that the solve reaches 60 s with a finite voltage.

packages/pybamm/src/pybamm/models/full_battery_models/lithium_ion/basic_dfn_unstructured.py
aabills: Fixed both in f9849155b. `BasicDFNUnstructured` and `BasicDFN2D` now store `j` under `"Negative/Positive electrode interfacial current density [A.m-2]"`, and the `BasicDFN2D` rename has a CHANGELOG bug-fix entry.

aabills:
Thanks @rtimms. The inline comments are answered in their threads. The nits:

- `L_scale` now includes `L_y**2` for `dimensionality: 2` (020d45c32).
- Added `"Voltage expression [V]"` (129e4b6f7).
- The integration test is parametrized on `dimensionality` and builds its model inside the test, so nothing is built at collection time and no instance is shared (66e9f9055).
- Restored the blank line in `pybamm/__init__.py` (f93c8abc2).

packages/pybamm/src/pybamm/models/full_battery_models/lithium_ion/basic_dfn_unstructured.py
aabills: Correction to my reply above: #5774 is now merged in, but your reproducer still fails (`IDA_ERR_FAIL`). The floor was only the coarse-mesh part of the problem. With the floor removed, the pure `div(grad(c))` operator on this Kuhn-tet slab still has growing modes once the mesh reaches about 10 cells in x and 4 across y and z. At the default resolution the max Re(λ) is +2.2e8, and it gets worse with refinement; hexahedra are stable. The cross term makes the operator non-symmetric (volume-weighted asymmetry 0.4), and dropping the cross term removes the growing modes but loses consistency. So the proper fix is a symmetric, coercive scheme such as SUSHI, which is a follow-up for `FiniteVolumeUnstructured`. For this PR I have removed tetrahedra from the docstring and said why (115846d81).

packages/pybamm/src/pybamm/models/full_battery_models/lithium_ion/basic_dfn_unstructured.py
```diff
@@ -0,0 +1,402 @@
...
+
+
+class BasicDFNUnstructured(BaseModel):
+    """Doyle-Fuller-Newman (DFN) model discretised with
+    :class:`~pybamm.FiniteVolumeUnstructured`.
+
+    The ``"dimensionality"`` option is the number of directions resolved besides
+    the through-cell *x*: 1 (default) meshes (x, z) with quads and 2 meshes
+    (x, y, z) with hexahedra. Pass ``submesh_types`` to
+    :class:`pybamm.Simulation` for triangles. Tetrahedra are not supported:
```
rtimms, line 19: This says tetrahedra are not supported, but `test_dfn_unstructured` still parametrises `(2, "tetrahedron")` and now asserts it solves. Each is defensible alone; together the test locks in behaviour the docstring disowns.

The measured picture after #5774 sits between the two. Tetrahedra now solve at `{x_n: 8, x_s: 8, x_p: 8, y: 2, z: 2}` — `V(1800) = 3.6132`, which failed before that fix — but still blow up at this model's own `default_var_pts` (`x = 10, y = 5, z = 5`):

```
pybamm.SolverError: IDA_ERR_FAIL: Error test failures occurred too many times
during one step or minimum step size was reached
```

So "unstable beyond coarse meshes" is accurate — the unit test just happens to sit inside the coarse window at `var_pts=3`.

Either drop the tetrahedron parametrisation, or soften this to say tetrahedra work only on coarse meshes and let the test stand as the demonstration. A `pybamm.logger.warning` when an `UnstructuredMeshGenerator` with `element_type="tetrahedron"` reaches this model would also save a user a confusing solver failure, since nothing currently stops them from passing one.

rtimms (approved):
Approving — all three findings and the four nits from the last round are addressed. One small inconsistency left inline, fine to fold into a follow-up commit here or a separate PR.

| Previous comment | Fix |
| --- | --- |
| Tetrahedra do not solve | `115846d81` docstring; root cause also fixed upstream in #5774 |
| Simplex cases only `build()` | `6a32fe5c1` — now `sim.solve([0, 60])` with a finite-voltage assertion |
| `"...electrode current density [A.m-2]"` holds `j` | `f9849155b`, with a Bug-fixes CHANGELOG entry |
| `L_scale` drops `L_y` in 3D | `020d45c32` |
| Missing `"Voltage expression [V]"` | `129e4b6f7` |
| Models built at collection time | `66e9f9055` — and the extracted `assert_total_lithium_conserved` is now shared with `test_basic_dfn_2d`, which is nicer than what I asked for |
| Stray blank line in `__init__.py` | `f93c8abc2` |

The `L_y**2` change touches every algebraic residual, so I re-ran the physics checks on this head rather than trusting the earlier round:

- Electrolyte and solid lithium inventory at `t=0` still match `pybamm.lithium_ion.DFN` to 1e-15 relative.
- Lithium conserved to ~3e-15 under **z**-graded active-material loading (the tests only grade in x) in both 2D and 3D, with the 2D and 3D voltages agreeing to 7e-11.
- All declared output variables process finitely in both dimensionalities.
- 1461 model / mesh / spatial-method unit tests and all 16 integration tests in this file pass.

packages/pybamm/src/pybamm/models/full_battery_models/lithium_ion/basic_dfn_unstructured.py
aabills: Addressed in #5789. I softened the docstring to say tetrahedra solve only on coarse meshes, and kept the test as the coarse-mesh demonstration with a comment explaining the mesh size. I did not add a warning: the model never sees `submesh_types`, which go from `Simulation` straight to `Mesh`, so that would need new plumbing. `FiniteVolumeUnstructured.build` already logs a 90° non-orthogonality warning on every domain for these meshes.

PR #5695, opened 2026-08-04
fix: stop ElectrodeSOHSolver leaking an expression tree per energy call

MarcBerliner:
## Description

`ElectrodeSOHSolver.theoretical_energy_integral` built a fresh expression tree on every call, with the 1000 stoichiometry points baked in as literal `Vector`s, and handed it to `ParameterValues.evaluate`. Because `Symbol.__hash__` is the structural id, each of those trees was a new key in the `ParameterSubstitutor` cache — which is unbounded and holds strong references. `evaluate` then discards the processed symbol, so those entries could never be hit again.

The result is a leak of ~0.2 MB per call. `SummaryVariables` calls `esoh_solver.solve` once per cycle, so a long degradation run retains a dead expression tree per cycle.

```python
esoh_solver = pybamm.lithium_ion.ElectrodeSOHSolver(parameter_values)
for i in range(1, 101):
    esoh_solver.solve({"Q_n": Q_n, "Q_p": Q_p, "Q_Li": Q_Li * (1 - 1e-6 * i),
                       "V_min": 2.5, "V_max": 4.2})
```

| after N solves | cache entries | Python heap |
|---|---|---|
| 20 | 523 | 6.1 MB |
| 60 | 1043 | 15.0 MB |
| 100 | 1563 | 23.9 MB |
| 100, then `clear_cache()` | 0 | 2.1 MB |

## Fix

The stoichiometries become `InputParameter`s, so one processed expression is reused across calls, cached per `(points, direction)` in a bounded `lru_cache(maxsize=4)` alongside the existing simulation caches.

That exposed a second bug: `InputParameter` never overrode `set_id`, and `Symbol.set_id` hashes only `(class, name, children, domains)`, so `InputParameter("x", expected_size=100)` and `expected_size=1000` were equal and hash-equal. The parameter cache handed back the wrong-sized processed expression:

```
ValueError: Input parameter 'y' was given an object of size '100' but was expecting an object of size '1000'.
```

`set_id` now includes `expected_size`. This is a latent correctness bug in its own right, not only a memory one.

## Results

Memory is flat — 346 cache entries and 3.0 MB at 100 solves, versus 1563 entries and 23.9 MB before. The energy value is unchanged (`19.0315204761098` vs `19.031520476109808`, rtol 1e-12), and `points=100 → 1000 → 100` no longer aliases.

Timing, interleaved A/B, 3 reps, fresh process each:

| | before | after | |
|---|---|---|---|
| eSOH cold solve | 44.1 / 46.2 / 48.3 ms | 45.2 / 46.1 / 47.9 ms | no change |
| eSOH warm solve (median) | 4.00 ms | 2.53 ms | 1.58x |
| `theoretical_energy_integral` | 0.87 ms | 0.115 ms | 7.6x |
| 10-cycle SPM + SEI run | 0.30 s | 0.29 s | ~3% |
| DFN build / solve (control) | 44.5 / 59.7 ms | 43.2 / 61.6 ms | noise |

Cold is unchanged because the tree is processed exactly once either way; the win is warm, where the old code re-expanded two `FunctionParameter` → `Interpolant` subtrees on every call.

## Tests

Two regression tests, both confirmed failing against unmodified `src` and passing with the fix:

- `TestCalculateTheoreticalEnergy::test_repeated_solves_do_not_grow_parameter_cache`
- `TestInputParameter::test_expected_size_changes_id`

Full unit suite run against both revisions in the same environment: the failure sets are identical apart from those two tests flipping to passing (the remaining failures are pre-existing missing-optional-dependency errors in my local venv — `bpx`, `scikit-fem`, `pybtex`, plotting).

## Notes

The underlying design hazard is untouched: `ParameterValues.evaluate()` still writes every symbol it processes into an unbounded strong-reference cache and then discards the result, so any other caller evaluating value-varying trees leaks the same way. Worth a separate look.

### DIFF
--- a/packages/pybamm/src/pybamm/models/full_battery_models/lithium_ion/electrode_soh.py
+++ b/packages/pybamm/src/pybamm/models/full_battery_models/lithium_ion/electrode_soh.py
@@ -369,9 +369,7 @@ class ElectrodeSOHSolver:
         self._get_electrode_soh_sims_split = lru_cache()(
             self.__get_electrode_soh_sims_split
         )
-        self._get_energy_ocv_function = lru_cache()(
-            self.__get_energy_ocv_function
-        )
+        self._get_energy_ocv_function = lru_cache()(self.__get_energy_ocv_function)
 
     def __getstate__(self):
         """
@@ -394,9 +392,7 @@ class ElectrodeSOHSolver:
         self._get_electrode_soh_sims_split = lru_cache()(
             self.__get_electrode_soh_sims_split
         )
-        self._get_energy_ocv_function = lru_cache()(
-            self.__get_energy_ocv_function
-        )
+        self._get_energy_ocv_function = lru_cache()(self.__get_energy_ocv_function)
 
     def _get_lims_ocp(self, direction):
         parameter_values = self.parameter_values
--- a/packages/pybamm/src/pybamm/models/full_battery_models/lithium_ion/electrode_soh.py
+++ b/packages/pybamm/src/pybamm/models/full_battery_models/lithium_ion/electrode_soh.py
@@ -369,6 +369,9 @@ class ElectrodeSOHSolver:
         self._get_electrode_soh_sims_split = lru_cache()(
             self.__get_electrode_soh_sims_split
         )
+        self._get_energy_ocv_function = lru_cache(maxsize=4)(
+            self.__get_energy_ocv_function
+        )
 
     def __getstate__(self):
         """
@@ -377,6 +380,7 @@ class ElectrodeSOHSolver:
         result = self.__dict__.copy()
         result["_get_electrode_soh_sims_full"] = None  # Exclude LRU cache
         result["_get_electrode_soh_sims_split"] = None  # Exclude LRU cache
+        result["_get_energy_ocv_function"] = None  # Exclude LRU cache
         return result
 
     def __setstate__(self, state):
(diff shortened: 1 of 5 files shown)


packages/pybamm/src/pybamm/models/full_battery_models/lithium_ion/electrode_soh.py
```diff
@@ -369,6 +369,9 @@ def __init__(
         self._get_electrode_soh_sims_split = lru_cache()(
             self.__get_electrode_soh_sims_split
         )
+        self._get_energy_ocv_function = lru_cache(maxsize=4)(
```
BradyPlanden: Q: was the max cache size selected based off a performance investigation or a safety buffer? It's not clear why this cache is capped while the above `_get_electrode_soh_sims_split` is not.

BradyPlanden (dismissed):
Nice find! One question on the cache size limitation, otherwise LGTM

packages/pybamm/src/pybamm/models/full_battery_models/lithium_ion/electrode_soh.py
MarcBerliner: No strong reason, there are few options for the cache's key so I set it to have a max size, but checking the max size causes a performance hit so I removed

BradyPlanden (approved):

PR #5704, opened 2026-08-07
Keep on-boundary 3D points inside the domain (fix corner/edge/face NaN masking)

aabills:
## Description

`UnstructuredSubMesh.contains_points_3d` decides whether 3D query points lie inside the mesh domain, using the generalized winding number (sum of signed solid angles over the boundary triangles), and returns `winding > 2 * np.pi`.

A point sitting **exactly on** the boundary surface gets a *partial* solid angle that lands at or below the halfway threshold, so it is misclassified as outside:

| point location | winding | `> 2π`? |
|---|---|---|
| interior | 4π | ✅ inside |
| flat face | 2π | ❌ (not strictly >) → outside |
| edge | π | ❌ → outside |
| corner | π/2 | ❌ → outside |
| exterior | 0 | ❌ outside |

`contains_points_3d` is called (via `ProcessedVariable._get_boundary_mask` → `_interpolate_spatial`, and via `get_3d_slices`) to blank out-of-domain points to NaN when **interpolating or plotting** a 3D unstructured field. So any query point that lands exactly on a domain face/edge/corner — which happens routinely when a regular plotting/sampling grid touches the domain boundary — is masked to NaN. The solution itself is unaffected; this is post-processing/plotting only.

Verified empirically on a unit cube: winding is exactly π/2 at a corner, π on an edge, 2π on a face, 4π in the interior, and 0 outside — matching the geometry.

## Fix

Lower the threshold below the smallest on-surface value (π/2 at a corner), so interior **and** on-boundary points count as inside while true exterior (winding ≈ 0) stays outside:

```python
return winding > 0.1
```

## Testing

- New test `test_contains_points_3d_boundary_points_are_inside` asserts a corner, edge midpoint, face centre, and interior point are all inside, and an exterior point is not. It **fails before** the change (corner/edge/face classified outside) and passes after — verified by reverting the source line alone.
- `uv run --group dev pytest packages/pybamm/tests/unit/test_meshes` — 163 passed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ## Bug fixes
 
-- 3D unstructured-mesh point-in-domain tests no longer misclassify points lying exactly on a domain face, edge, or corner as outside, so interpolating or plotting a field at the domain boundary no longer blanks it to NaN. ([#XXXX](https://github.com/pybamm-team/PyBaMM/pull/XXXX))
+- 3D unstructured-mesh point-in-domain tests no longer misclassify points lying exactly on a domain face, edge, or corner as outside, so interpolating or plotting a field at the domain boundary no longer blanks it to NaN. ([#5704](https://github.com/pybamm-team/PyBaMM/pull/5704))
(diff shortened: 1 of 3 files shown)


aabills:
Closing: out of scope for #5688. Pre-existing item in already-merged #5687 mesh code, not part of the spatial-method PR.

PR #5705, opened 2026-08-07
Invalidate the tagged-mesh cache when the source file changes

aabills:
## Description

`TaggedSubMeshGenerator` caches parsed `.msh` files so a multi-domain model reading the same file for several regions parses it once. The cache was an unbounded, class-level dict keyed on the raw path object:

```python
_mesh_cache: dict = {}

@classmethod
def _read(cls, path):
    if path not in cls._mesh_cache:
        cls._mesh_cache[path] = meshio.read(str(path))
    return cls._mesh_cache[path]
```

Three problems:

1. **Never invalidated on file change.** The key is just the path, so editing the `.msh` on disk and re-running in the same session silently returns the *old* mesh. This bites anyone iterating on a mesh during development — their edits are ignored until the process restarts.
2. **Unbounded and process-global.** A class attribute shared across all instances, never cleared; every file read stays in memory for the process lifetime.
3. **`str` vs `Path` double-caches.** The key is the raw object, so `"x.msh"` and `Path("x.msh")` create two entries for the same file.

Verified the stale read empirically: after rewriting a file from 9 to 18 points on disk, `_read` returned the same 9-point object.

## Fix

Key on `(os.fspath(path), st_mtime_ns)` via `functools.lru_cache`:
- `os.fspath` collapses `str`/`Path` to one key,
- the modification time in the key makes an edited file a cache miss → re-read,
- `lru_cache(maxsize=8)` bounds retained meshes.

## Testing

- New test `test_tagged_generator_cache_reads_and_invalidates`: same unchanged file returns the same object (cache hit), `str`/`Path` collapse to one entry, and bumping the file's mtime forces a re-read.
- The four existing `test_tagged_generator_*` tests injected fake meshes into `_mesh_cache` directly; they now monkeypatch `_read` (they test region extraction, not caching).
- `uv run --group dev pytest packages/pybamm/tests/unit/test_meshes` — 163 passed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ## Bug fixes
 
-- `TaggedSubMeshGenerator` no longer returns a stale cached mesh when the source `.msh` file changes on disk: its mesh cache is keyed on the file's modification time, collapses `str`/`pathlib.Path` paths to one entry, and is bounded in size. ([#XXXX](https://github.com/pybamm-team/PyBaMM/pull/XXXX))
+- `TaggedSubMeshGenerator` no longer returns a stale cached mesh when the source `.msh` file changes on disk: its mesh cache is keyed on the file's modification time, collapses `str`/`pathlib.Path` paths to one entry, and is bounded in size. ([#5705](https://github.com/pybamm-team/PyBaMM/pull/5705))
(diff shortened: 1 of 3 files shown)


aabills:
Closing: out of scope for #5688. Pre-existing item in already-merged #5687 mesh code, not part of the spatial-method PR.

PR #5717, opened 2026-08-12
fix: stream IDAKLU diagnostics on the GIL thread, buffer them off it

BradyPlanden:
# Description

`SolverLog` called `pybamm.logger.debug` and `PrintStats` called `py::print` from inside the OpenMP region in `IDAKLUSolverGroup::solve`. The thread that entered `solve()` holds the GIL throughout, but the workers are not Python threads and never acquire it, so they raced each other on the logger's reference count and on the output. Diagnostics now route by thread: `set_logger` records the GIL holder, so a solve on that thread streams live, while worker-thread solves buffer and are drained after the parallel region. Solving several input sets still gives one trace and one `print_stats` block each, and a failed solve still emits its trace, partial or thrown.

Three unrelated commits ride along: 
- writing IDAKLU output sensitivities in their final layout (dropping a transpose pass and a duplicate buffer)
-  a comment on why `jacobian_eval` re-copies the sparse index arrays, 
- restoring `sys.modules` after the minimal-install test, which was causing sporadic test failures

Fixes # (issue)

## Type of change

# Important checks:


MarcBerliner (approved):
Thanks!

rtimms (approved):

PR #5724, opened 2026-08-19
Ensure step and cycle lengths match

NicolaCourtier:
# Description

A one-line fix to ensure that the step length is not longer than the cycle length.

## Type of change

# Important checks:

### DIFF
--- a/packages/pybamm/src/pybamm/solvers/solution.py
+++ b/packages/pybamm/src/pybamm/solvers/solution.py
@@ -979,6 +979,9 @@ class Solution(SolutionBase):
                     data_short_names["Step"] = np.concatenate(
                         [data_short_names["Step"], j * np.ones_like(step.t)]
                     )
+                data_short_names["Step"] = data_short_names["Step"][
+                    : len(data_short_names["Cycle"])
+                ]
 
         return data_short_names
 


NicolaCourtier:
Small fix, ready for review (@rtimms ?)

rtimms (changes requested):
Diagnosis is right — `make_cycle_solution` stores the unsliced step solutions on `cycle.steps`, while `cycle.t` comes from `from_sub_solutions`, which drops the duplicated leading sample. Using exact `==` correctly mirrors that.

## 1. Cross-cycle boundaries are still off

`"Cycle"` is built from `np.ones_like(cycle.t)` with no dedup, but the parent solution is *also* `from_sub_solutions(cycle_solutions)`, so it drops the repeat at each cycle boundary too. The `j > 0` guard skips exactly that case. Two cycles × two steps, all boundaries exactly repeated:

```
data len : 197   t len: 197
Cycle len: 198
Step  len: 198
csv FAILED: ValueError All arrays must be of the same length
```

Step now agrees with Cycle, and both still disagree with the data. `save_data(to_format="csv")` is the user-visible failure.

Carrying one boundary across the flattened step list fixes both, and lets `"Cycle"` inherit the dedup instead of needing its own mechanism:

```python
# Save cycle number and step number if the solution has them
if cycles_and_steps and len(self.cycles) > 0:
    cycle_numbers, step_numbers, previous_last_t = [], [], None
    for i, cycle in enumerate(self.cycles):
        for j, step in enumerate(cycle.steps):
            # a step's first point is dropped when it repeats the previous
            # step's last point, so don't number it twice
            repeated = (
                previous_last_t is not None and step.t[0] == previous_last_t
            )
            previous_last_t = step.t[-1]
            length = len(step.t) - repeated
            cycle_numbers.append(np.full(length, i, dtype=float))
            step_numbers.append(np.full(length, j, dtype=float))
    data_short_names["Cycle"] = np.concatenate(cycle_numbers)
    data_short_names["Step"] = np.concatenate(step_numbers)
```

Also one `np.concatenate` at the end rather than one per step, which matters on long cycling runs.

## 2. The test asserts the wrong invariant

`np.zeros(50), np.ones(49)` hardcodes the answer. What matters is `len(Step) == len(Cycle) == len(data["c"])` — assert that, add a two-cycle case, and the gap above gets caught. As written the test passes on a solution that still can't be written to csv.

## 3. Nit

`np.ones_like(step.t[first])` slices an array purely for its shape.

---

Worth noting for severity: normal experiments don't hit this. `BaseSolver.step` nudges each step start with `np.nextafter`, so real `Simulation.solve` runs came out consistent in every configuration I tried (event-terminated, fixed period, multi-cycle). It bites when solutions are assembled directly — `sol1 + sol2`, hand-built cycles, `starting_solution` — which is exactly what the test does.

rtimms:
Thanks @NicolaCourtier ! Just a couple of comments

NicolaCourtier:
Thanks Rob, cc. @SarahRo

Issue #5511, opened 2026-05-12
Model Zoo: structure and support model for contributed models


rtimms:
### Summary

We want a clear, lightweight home for community- and partner-contributed models (the 3D pouch submodel PR being the immediate example) without those models holding up core PyBaMM releases or being silently broken by upstream changes. This issue is to agree the short-term structure and capture the longer-term options.

### Background

At the recent developer meeting we agreed that a dedicated `model_zoo/` (or similarly named) folder inside the PyBaMM repository is the right short-term home for contributed models. The motivation:

- Lowers the barrier to contributing — no separate repo, packaging, or release coordination required.
- Models discoverable next to the core code that they extend.
- Lets us evolve the support model in-place before committing to a heavier solution (separate PyPI package, registry, etc.).

### Proposed short-term structure

- A top-level `model_zoo/` folder, with one subfolder per contributed model.
- Each model folder contains the model code, a short README, and a `tests/` subfolder.
- Each model is assigned a **support tier** at merge time:
  - **Tier 1** — must pass tests on every PR; treated like core code.
  - **Tier 2+** — tests run but failures are allowed/skipped; flagged on the model's page rather than blocking PRs.
- PRs gate on model zoo tests for the tier the changed model belongs to (or above).
- We auto-track which PyBaMM versions a model passes against and surface that history (e.g. a generated compatibility table per model).
- Failure to pass tests in the model zoo does **not** block a core PyBaMM release.

### Open questions / longer-term options (not deciding now)

- Discoverability — do we adopt a pytest-plugin-style convention so external packages can register themselves as model zoo entries?
- Should the model zoo eventually move to a separate PyPI package?
- A registry text file (repo URLs of external models) vs. all-in-tree.
- Per-model Zenodo DOIs for citability when models live in their own repos.
- Ownership/maintainership — how do we surface a contact and avoid orphaned models?
- Relationship to the planned Imperial / MSM extension model repo — we should make sure these don't diverge.

### Suggested next steps

1. Agree the folder layout and tier system in this issue.
2. Land an initial `model_zoo/` skeleton with one example model and the CI wiring.
3. Migrate the 3D pouch submodel PR as the first real entry.
4. Spin out a separate issue for the longer-term discoverability/ownership question once we have a few models in-tree.

cc: @BradyPlanden @agriyakhetarpal @aabills @brosaplanella @martinjrobins (please add anyone I've missed)

martinjrobins:
[...]

PR #5727, opened 2026-08-20
feat: add the model zoo for contributed models

BradyPlanden:
# Description

Adds `packages/pybamm-model-zoo/`, a uv workspace member importable as `pybamm_model_zoo`, as the home for community- and partner-contributed models. One self-contained folder per model holds its code, tests, examples, citation, and a declarative `model.toml` manifest. An SPM linearised about its starting stoichiometry, for GITT fitting is provided as a reference entry at `core` tier, so the gating path is exercised from the start.

Nothing under `packages/pybamm/src/pybamm/` changes; everything outside the zoo is config, CI, or docs.

Addresses #5511

## Using a zoo model

Not on PyPI yet, so `pip install pybamm` does not provide it:

```
pip install "pybamm-model-zoo @ git+https://github.com/pybamm-team/PyBaMM.git#subdirectory=packages/pybamm-model-zoo"
```

```python
import pybamm
import pybamm_model_zoo as zoo

model = zoo.load("LinearisedSPM")()
solution = pybamm.Simulation(model).solve([0, 300])
```

Models come from the registry rather than the `pybamm` namespace: `zoo.list_models()` for what is installed, `zoo.info(name)` for maintainer, tier, and supported PyBaMM range. A `core` model is in PyBaMM's merge gate. A `community` model is tested on every pull request but advisory, so its badge and the compatibility table in the model zoo docs are where to see what it currently passes on. Adding one is `nox -s zoo-new` plus a filled-in `model.toml`.

## Reviewing the implementation

The manifest is the only boilerplate: registry, contract suite, docs, CI routing, and badges all derive from it, and manifests are parsed, never imported (`_registry.py`), so a model with missing dependencies still lists and fails cleanly instead of taking the zoo down.

Weekly, `model_zoo_status.yml` pairs each model with the newest releases its own `pybamm_requires` admits, then opens a reviewable pull request with `status.json`, badges, and the docs table. Contributor and reviewer policy is the package `README.md`, which the docs include verbatim.

## Type of change

# Important checks:


rtimms (commented):
Nice piece of work — the architecture holds up well. Manifest-as-single-source, parse-never-import, tier-derived CI routing, and rendering the template then running the whole contract against the result are all genuinely good calls, and the `--expect` handling so a dead leg can't leave a green badge is a detail most people would miss.

Five things I'd want addressed, roughly in severity order.

### 1. One malformed `model.toml` takes down the whole registry

`Registry._discover` (`_registry.py:304`) lets `ManifestError` from `_entry_from_manifest` escape `Registry.__init__`. Add a community model with a TOML typo, or a missing `[model]` table, and:

`conftest.pytest_ignore_collect` → `_model_folders()` → `zoo.all_entries()` → `Registry()` raises, so `nox -s zoo-gating` errors during collection — *before* `--zoo-tier=core` gets a chance to prune the folder. That's the one thing the README and the `run_zoo_gating` docstring both promise can't happen ("a model that fails to import can only fail its own run… keeps the merge gate independent of the community tier"), and `check_manifest` — the check that exists to catch exactly this — never runs.

Worse for external collections: a third-party package shipping one broken manifest makes `zoo.list_models()` raise for every user, and the `external`-warns branch at `_registry.py:311` is unreachable in that case.

Collecting per-entry failures instead of raising (an "unparseable" entry that fails its own `manifest` check) would restore the stated guarantee.

### 2. An invalid `tier` value silently drops a model out of the merge gate

`conftest.py:95` derives gating from `entry.tier == "core"` with no validation — but the check that validates `tier` (`check_manifest`) is itself gated by that same derived tier. It's circular.

Concretely: change `tier = "core"` to `tier = "Core"` in `linearised_spm/model.toml`. `core_slugs` is empty, so none of its contract cases get the `gating` marker, and `--zoo-tier=core` prunes the folder as well. `nox -s zoo-gating` then collects only the zoo's own machinery tests and goes **green**; the `check_manifest` failure lands only in the advisory job. A core model leaves PyBaMM's merge gate on a one-character manifest edit, and the gate doesn't notice.

Validating the tier eagerly (unknown tier → hard error, not a silent demotion to non-gating) closes it.

### 3. `resolve()` mismatch can silently switch off folder pruning

`_model_folders()` returns `entry.path.resolve()`, but `collection_path` comes from pytest's `absolutepath()`, which does *not* resolve symlinks. Under a symlinked checkout (symlinked home/workspace, `/tmp` → `/private/tmp` on macOS, some CI workspaces) neither `collection_path != folder` nor the `parents` test matches, the loop falls through to `return None`, and `--zoo-tier` / `--zoo-model` stop pruning entirely — silently, which is the bad part, since pruning is the isolation mechanism.

`tests/test_selection.py` hand-builds both sides, so it can't catch this. Resolving (or not resolving) consistently on both sides is the fix.

### 4. `window_for(count=0)` returns every release

`_versions.py:46`: `[-count:]` with `count == 0` is `[-0:]` — the whole list. `matrix.py --releases 0`, a plausible way to say "only `main` this week", fans the weekly job out to every release on PyPI instead of none. `releases[-count:] if count else []`.

### 5. The docs-regen pre-commit hook misses the renderer

The `model-zoo-docs` hook's `files:` pattern watches `scripts/generate.py`, but all the rendering lives in `_docs.py` — `generate.py` is a thin CLI wrapper, as its own docstring says. Edit `_INDEX_HEADER` or `model_page()` and the hook doesn't fire, stale docs get committed, and the failure only shows up later as `TestGeneratedFiles::test_docs_and_badges_are_current` plus the `docs` contract check going red, with no local hook that would have caught it. `badges/*.json` is missing from the pattern too.

---

Checked and happy with: the `_paths` parent-walk depths, the `parse_bibtex` brace scan, the status folding and missing-cell handling, `version_key`'s treatment of `main`, the `ci_gate` / `*_advisory` naming contract (`run_zoo_gating` is in `needs`, the advisory job is correctly suffixed), the OCP submodel override against `base_lithium_ion_model.set_open_circuit_potential_submodel`, the `_compat` shims against the real `pybamm.citations` internals, and that the committed `index.md` and badge JSON are exactly what `_docs` renders with no `status.json` present.

rtimms (dismissed):
Re-reviewed at `2942996a`. All five findings are fixed, each as its own commit with a regression test. Approving — with one small thing below I'd like in before this merges.

Two of them are worth calling out because they go at the cause rather than the symptom:

- **`e40386c5`** — a manifest too broken to key now yields an entry keyed on its folder name with the failure recorded in `ModelEntry.error`, instead of raising past every other model. Pairing that with a `tier` of `""` so `in_tier` keeps it in *every* run is the right move: the entry stays in collection and `check_manifest` reports it from inside the gate. Keeping in-tree duplicates raising is also right — that's a relationship between two manifests, not a malformed field.
- **`09da0d95`** — `in_tier` failing closed (an unrecognised tier belongs to every tier rather than none) is a better fix than validating eagerly, and it's what makes the error entries above reportable in the first place. The two interlock properly.

`e99bf2ab` driving the real `_model_folders` against a real symlink, rather than hand-building both sides, is exactly why the original bug was invisible — good catch in the test as well as the code. `a34c132a` deriving the hook's expected coverage from `_docs.all_files` and the generator's real inputs means the pattern can't go stale the same way twice.

### One thing to fix first

`tests/test_contract.py:42` — introduced by the `e40386c5` fix:

```python
if check.name in entry.tests.skip_contract:      # runs first, ignores check.waivable
    pytest.skip(...)
if entry.error is not None and check.name != "manifest":
    pytest.skip(...)
```

`waivable=False` on `check_manifest` is only ever enforced by `_check_tests_block`, which lives *inside* `check_manifest`. So a manifest with valid TOML, a `[model]` table, `skip_contract = ["manifest"]` and a non-string `slug` sets `error` while leaving `raw` populated: the first branch skips `manifest`, the second then skips everything else, and the model contributes zero contract checks with CI green. The unwaivable check becomes waivable exactly when the manifest is broken.

One line, and it closes the inversion generally rather than just for the error case:

```python
if check.waivable and check.name in entry.tests.skip_contract:
```

Worth a test that a `skip_contract = ["manifest"]` entry still fails, since right now `waivable` is a field nothing enforces at the point of use.

---

Also confirmed the unrelated `.readthedocs.yaml` change in `2942996a` is verified rather than assumed — `docs/readthedocs.org:pybamm` is green on the head, so the native `method: uv` schema resolves. Codacy is the only red check and it isn't in the merge gate.

rtimms (approved):

PR #5730, opened 2026-08-20
Solve the composite electrode SOH by bracketed rootfind

MarcBerliner:
Stacked on #5729, which adds `pybamm.Brent`. Review that first; this diff is against it.

## What changes

Each stoichiometry was a variable in one nine-variable algebraic system handed to a Newton solver, which makes reaching a state a basin-of-attraction problem — worn and lithium-poor cells sat outside it.

Now each stoichiometry is an *expression*, found by inverting the open-circuit potential it sits on with a `pybamm.Brent`. `LithiumIonParameters.U` clips its stoichiometry and adds an asymptote, so every OCP diverges outside `[0, 1]`; every inversion therefore brackets a root and converges from any target — including targets outside the physical range, whose answer is non-physical but exact.

Nothing is left for a solver to iterate, so the stoichiometries are read back through one CasADi function rather than per-variable post-processing. The capacities reach that function as inputs, so only the parameters its graph substitutes can invalidate it: a caller ageing a cell reuses it, a caller swapping an OCP does not.

## Measured against `main`

8000 targets — 1000 states of charge on `[0, 1]`, 1000 voltages on `[2.5, 4.2] V`, three hysteresis directions, four wear states, through the public API:

| | `main` | this branch |
|---|---|---|
| **failures** | **147 / 8000** | **0 / 8000** |
| build | 12.2 ms | 18.5 ms |
| first solve | 58.9 ms | 14.8 ms |

`main`'s failures are silent — every one returns a state violating lithium conservation rather than raising, concentrated at low SOC.

Repeat calls through a reused simulation, reading the stoichiometries back:

| | `main` | this branch |
|---|---|---|
| new target | 18.8 ms | 7.2 ms |
| aged capacities | 15.3 ms | 15.6 ms |
| no reuse (cold) | 75.3 ms | 91.2 ms |

## Known limitations

- **Reading variables through `ProcessedVariable` is slower than `main`** (roughly 4×). It converts and evaluates the whole nested rootfind once per variable; the evaluator bypasses that, but only for `get_initial_stoichiometries_composite`, which is the path `Simulation` uses. A caller driving `ElectrodeSOHComposite` directly pays it.
- The build is ~1.5× slower: the Brent graph is larger symbolically than the algebraic system it replaces. Paid once per model.
- `solve_split` is kept as the fallback and is now actually reachable — the native rootfinder reports failure as `RuntimeError`, which the old ladder did not catch, and the retry re-ran an identical solve. It now returns the split result instead.
- Bracketing guarantees convergence, not uniqueness. A user OCP that is non-monotone *within* `[0, 1]` would still bracket, and Brent would return one of its roots.
- Composite-positive coverage in the sweeps uses parameters aliased from their single-phase equivalents, since `Chen2020_composite` describes a composite negative. That is a degenerate composite, not a real parameter set.

Tests: 4221 unit, 32 sweep tests in `test_electrode_soh_composite_sweeps.py`.

### DIFF
--- a/packages/pybamm/src/pybamm/models/full_battery_models/lithium_ion/electrode_soh_composite.py
+++ b/packages/pybamm/src/pybamm/models/full_battery_models/lithium_ion/electrode_soh_composite.py
@@ -6,6 +6,9 @@ from __future__ import annotations
 import warnings
 from typing import Any
 
+import casadi
+import numpy as np
+
 import pybamm
 
 from .electrode_soh import _ElectrodeSOH, get_esoh_default_solver
@@ -14,6 +17,14 @@ from .util import (
     get_equilibrium_direction,
     get_lithiation_delithiation,
 )
+
+# `LithiumIonParameters.U` clips its stoichiometry and adds an asymptote, so every OCP
+# diverges outside [0, 1] and both brackets straddle a root. The inner one is wider so
+# an inversion can always reach the potential the outer solve asks it for.
+_STOICH_LO, _STOICH_HI = -10.0, 10.0
+_BRACKET_LO, _BRACKET_HI = -5.0, 5.0
+
+_ABSTOL, _MAX_ITER = 1e-14, 200
 
 
 def _get_primary_only_options(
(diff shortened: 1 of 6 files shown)


packages/pybamm/src/pybamm/models/full_battery_models/lithium_ion/electrode_soh_composite.py
```diff
@@ -300,234 +324,195 @@ def __init__(
...
-                ),
+        Q_n = [pybamm.InputParameter("Q_n_1")]
+        Q_p = [pybamm.InputParameter("Q_p_1")]
+        if neg_composite:
+            Q_n.append(pybamm.InputParameter("Q_n_2"))
+        if pos_composite:
+            Q_p.append(pybamm.InputParameter("Q_p_2"))
+
+        T_ref = param.T_ref
+        T_init = param.T_init if initialization_method == "voltage" else T_ref
```
aabills, line 339: is this logic necessary? Kinda strange that we change the value of temperature based on initialization method

aabills (commented):
real review pending #5729

Issue #5667, opened 2026-07-21
Proposal: a native Rust execution backend


BradyPlanden:
CasADi has been the default execution workhorse for years and has served us well. I've been building an alternative backend in Rust (PyO3/maturin) as owning the execution layer lets us target improvements and fixes specifically to PyBaMM's needs, in ways that are hard to do through CasADi or we have previously had challenges upstreaming.

What I'm aiming for:

- Memory safety by construction, avoiding the class of C++ issues we've previously hit in pybammsolvers.
- Faster executions and solves.
- Feature parity with our Casadi current usage
- Clean IDAKLU integration and native Diffsol support

Rollout: this would land as an opt-in backend first so nothing changes for existing users. Once it stabilises, it would be promoted to the default backend. I have a working end-to-end implementation and will open a PR to show what the final solution would look like. If we are happy, I can then split it into reviewable-sized PRs.

---

Some high-level benchmark results (Apple Silicon, `t_eval` as 1000 points, Chen2020, default mesh, 1C discharge):
-  Per call kernel evaluation is 1.4-30x faster than CasADi's VM (eval/JVP 6–30x, Jacobians 1.4-2.3x) at parity (max abs diff ~1e-10). 
- With the IDAKLU solver fixed (atol 1e-6 / rtol 1e-6) and swapping only the evaluation engine, warm solves are 1.3-2.2x faster (SPM 2.2x, SPMe 1.3x, DFN 1.5x) matching CasADi to ~1e-12 on SPM/SPMe and ~1e-5 (DFN).
- Forward-sensitivity solves are ~3-10x faster end-to-end (SPM ~10x, SPMe ~4x, DFN ~3x), driven mainly by native sensitivity observation (CasADi 17-170 ms to materialise sensitivities vs under 5 ms in Rust); integration-only speedup is 1.2-2.6x.
- Warm solves speed match or beat CasADi's AOT-compiled kernels (faster on SPM/DFN, within a few percent on SPMe) with none of AOT's one-time compile cost (~0.8s for SPM / ~17 s for DFN).

I'll include the full benchmark matrix in the PR. Feedback very welcome!

martinjrobins:
sounds fantastic @BradyPlanden, can't wait to see the PR :) Is GPU support and complex numbers support in scope here? These are two areas that have always been impossible with casadi. Also, it would be helpful if you can say a few words on why your implementation is faster than casadi's, are there any up-front costs? i.e. is it slower on cold starts?

[...]

PR #5732, opened 2026-08-20
showcase: rust execution engine

BradyPlanden:
> This is a showcase PR, not intended to be merged.

This PR adds a Rust execution engine as discussed in #5667. This engine compiles a discretised model's expression trees: RHS, algebraic residuals, Jacobians, events, output variables, into a flat instruction tape evaluated entirely in Rust. It updates the IDAKLU solver with an FFI to integrate with this engine without crossing the Python layer, and adds a pure-Rust diffsol solver implementation.

The engine is opt-in through `BaseModel.convert_to_format`:

```python
model.convert_to_format = "rust" # the entire opt-in
sol = pybamm.Simulation(model).solve([0, 3600])
```

This engine lands as an optional backend for PyBaMM, but it offers the structure to replace CasADi as the default once it stabilises. This branch moves the default to `'rust'`, showcasing full test suite coverage.

> GitHub offers an interactive VS Code interface for reviewing PRs, which I would recommend as a good way to poke around (or even better, clone locally). This is enabled by typing a full stop `.` on this PR page.

## Mental model

`convert_to_format` already selected the expression backend (`"casadi"`, `"jax"`, `"python"`); this adds `"rust"`. Four components sit behind it:

|                             | What                                                                                                                                             | Where                                |
|-----------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------|--------------------------------------|
| **Compute core**            | Expression DAG → typed IR → flat tape. Includes forward-tangent + adjoint AD, with DSATUR colouring, and simplifier. Knows nothing about PyBaMM. | `pybamm-rust/pybamm-core`            |
| **Rust-Python bindings**    | Bindings between the pure-Rust core and PyBaMM, with stubs provided for type-hinting.                                                            | `pybamm-rust/pybamm-python`          |
| **IDAKLU swappable backend** | The *same* SUNDIALS IDA + KLU solver, `RustFunctions` substituted for `CasadiFunctions`.                                                        | `pybammsolvers/.../Expressions/Rust` |
| **New solver**              | `pybamm.DiffsolSolver`: pure-Rust BDF (diffsol + faer). No SUNDIALS, no CasADi in the path.                                                      | `solvers/diffsol_solver.py`          |

**Performance has two axes**

- **Integration:** kernel cost in the stepping loop. For IDAKLU, both backends call into preallocated C++ buffers, so it's a pure arithmetic comparison.
- **Observation:** extracting variables and sensitivities. `ProcessedVariable` drives CasADi with a per-timepoint Python loop; the Rust path runs one native tangent sweep *per parameter* across the whole trajectory, GIL released. **Scales with output points.**

## Performance

**The baseline is `origin/main`.** Every CasADi number below was measured in a second checkout of `origin/main` (`abfdff33f`) with its own replicated environment. The Rust numbers come from this branch.

### Top-level
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://github.com/user-attachments/assets/09ae8b30-2503-4ea6-81e8-877bdf287aae">
  Rust IDAKLU speedup against the CasADi default, every lane by model by output-point count
</picture>

`rust_idaklu` beats `main`'s CasADi default in **all 72 cells, from 1.22× to 12.36×**. `rust_diffsol` wins 69 of 72, see *Gradients*.

### 1C discharge
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://github.com/user-attachments/assets/c7ac988f-940f-4014-a5ea-ea1956155ec4">
  Wall-clock against output points, 1C constant-current discharge
</picture>

### Triangle-wave drive cycle
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://github.com/user-attachments/assets/3f018769-af8c-4c35-9890-d2c14bf0bc38">
  Wall-clock against output points, triangle-wave drive cycle
</picture>

### Output variables

Passing `output_variables=["Voltage [V]"]` moves observation inside the solve and skips storing the full state.

<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://github.com/user-attachments/assets/c8de5ba1-aa6f-457a-a5e3-dc163f8fce3e">
  Wall-clock against output points, output_variables solve, 1C discharge
</picture>

### Repeated evaluation

One solve per parameter draw, read through the interpolating call interface, representing the inner loop of a fitting or inference run. With `output_variables=` set the `rust_idaklu` win rises to 1.35–6.98×.

| Model | 1C discharge | drive cycle |
|---|---|---|
| SPM | 1.81–1.96× | 2.59–5.26× |
| SPMe | 2.59–2.91× | 2.18–3.49× |
| DFN | 1.23–1.45× | 1.26–1.52× |

### Gradients

Forward sensitivities. The largest ratios in the set, and the ones needing the most care to read.

| DFN sensitivity vs CasADi on `main` | 10 | 100 | 1,000 | 10,000 |
|---|---:|---:|---:|---:|
| `rust_idaklu`, 1C discharge | 1.39× | 1.69× | 3.87× | **7.87×** |
| `rust_idaklu`, drive cycle | 1.78× | 1.92× | 3.64× | **8.08×** |
| `casadi_idaklu_aot`, 1C discharge | 1.53× | 1.40× | 1.11× | **1.01×** |

AOT-compiling the CasADi kernels is worth ~1.5× when integration dominates and decays to nothing once observation does. The Rust path moves the opposite way and grows to 8.08×.

**Two caveats.** `main`'s IDAKLU never calls `IDASetSensParams`, so it runs at IDAS's `pbar = 1` default; this PR passes `pbar_i = |p_i|`, which moves step selection for both IDAKLU backends. `rust_diffsol` error-controls the scaled sensitivities while neither IDAKLU backend does (`IDASetSensErrCon` is never called), so it integrates strictly more state under error control than the rows it is timed against, which is why it is slower than `main`'s CasADi.

Setting `output_variables=` shrinks the gradient win to 1.13–3.88×, because the observation saving is already taken by the restriction.

### The experiment path

`pybamm.Experiment` with four steps, 10 s period against `main`'s CasADi, as `rust_idaklu` / `rust_diffsol`. Runnable by using `--protocols experiment` in the benchmark suite.

| lane | SPM | SPMe | DFN |
|---|---|---|---|
| solver | 1.49× / 1.49× | 1.81× / 2.04× | 1.47× / 1.67× |
| inference | 1.16× / 1.23× | 1.89× / 2.15× | 1.53× / 1.69× |
| sensitivity | 4.61× / 4.53× | 2.92× / 2.82× | 2.09× / 1.03× |

All 9 cells win. This protocol sits outside the sweep above, as its step period fixes the output grid.

**`output_variables=` plus `calculate_sensitivities` is not supported across a step boundary**, on any backend or either branch: the handoff needs `d(state)/dp` at the boundary and an outputs-only solve does not store it. This PR makes that fail uniformly with a named `SolverError`. On `main` the same combination *silently returns wrong gradients* (off by ~110–120× tolerance), which is how it was found.

### Where the time goes

DFN sensitivities, 1C discharge, at 10,000 output points (ms):

| | integration | observation | end-to-end |
|---|---:|---:|---:|
| CasADi default (`main`) | 127.85 | 1,560.46 | 1,705.20 |
| CasADi AOT-compiled (`main`) | 106.42 | 1,573.71 | 1,695.95 |
| Rust · IDAKLU | **109.15** | **93.11** | **216.69** |

Rust's integration lands level with AOT-compiled CasADi (109 ms against 106 ms) with no compile step in front of it, and the end-to-end 7.87× comes mostly from the other axis: 1,560 ms of observation becomes 93 ms.

### Setup cost

The fair "fast CasADi" comparator is `options={"compile": True}`, which shells out to `gcc -O3 -march=native`. Rust's "compile" is an in-memory IR build:

| Model | CasADi AOT prepare | Rust prepare | Ratio |
|---|---:|---:|---:|
| SPM | 672.3 ms | 5.0 ms | 135× |
| SPMe | 1,691.8 ms | 10.8 ms | 157× |
| DFN | 15,163.3 ms | 5.5 ms | **2,734×** |

*(first-solve setup, 1C discharge at 1,000 points; CasADi measured on `main`, cache isolated per case, each row verified as a fresh compile followed by a fresh-process disk reload. The non-AOT `casadi_idaklu` prepare is 14.6 / 36.2 / 40.3 ms.)*

### Numerical parity and accuracy

Correctness of the Rust backend is established in the test suite, where the kernels (RHS, Jacobians, events, outputs, tangents) are checked against CasADi expression-for-expression at near machine precision (the shared parity fixture asserts rtol `1e-10`, atol `1e-14`). `rust_idaklu` feeds those kernels to the *same* SUNDIALS integrator, so its trajectories are the same computation. The tables below confirm that parity holds end-to-end through the FFI as `rust_idaklu`'s worst errors match `casadi_idaklu`'s **to the digit** in every cell.

We can use the benchmark to measure the solution accuracy of `rust_diffsol` in reference to IDAKLU/CasADi. This is done below, where every row is scored against a **converged reference of `atol = rtol = 1e-10`** (falling back to `1e-9` where that will not converge) with no solution interpolation. Rows outside tolerance, by lane and output-point count:

| lane | 10 | 100 | 1,000 | 10,000 |
|---|---:|---:|---:|---:|
| solver | 0/48 | 0/48 | 1/48 | 1/48 |
| sensitivity | 0/48 | 0/48 | 0/48 | 0/48 |
| inference | 46/48 | 43/48 | 0/48 | 0/48 |

At 1,000 and 10,000 output points **286 of 288 rows pass**. The two misses are a narrow state-accuracy gap for a DFN drive cycle at 1.16–1.17× tolerance.

The inference-lane block at 10 and 100 points is a property of the *output grid*, where that lane reads strictly between stored nodes, and a 10-point grid cannot resolve a discharge between its nodes however the trajectory was computed. `main` misses in the same proportion. Where the interpolants differ the Rust diffsol path is the **most** accurate of the three, because it supplies derivatives for Hermite interpolation where both IDAKLU backends fall back to a linear chord:

| worst normalized error, inference | 10 pts | 100 pts |   1,000 pts | 10,000 pts |
|---|---:|---:|------------:|----------:|
| `casadi_idaklu` (`main`) | 342.9 | 16.4 |        0.87 |       0.02 |
| `rust_idaklu` | 342.9 | 16.4 |        0.87 |      0.02 |
| `rust_diffsol` | **130.6** | **5.2** |   **0.51**  |    0.05  |

The identical `casadi_idaklu` / `rust_idaklu` rows are the end-to-end form of the parity the shared-integrator design was meant to buy.

## Binary size

<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://github.com/user-attachments/assets/a0dd3850-3c53-42c2-98d6-a03ed69aa2b6">
  Shipped binary footprint, Rust core against the CasADi wheel
</picture>

This is mostly because CasADi's wheel carries IPOPT, HiGHS, Cbc/Clp, qpOASES, MUMPS and a bundled gfortran that we never use and can't decline. The Rust improvement is shown in the dependency footprint below where diffsol, faer and the whole CAS/AD stack are statically linked into one object.

```
$ otool -L pybamm/rust/_core.abi3.so
    /usr/lib/libc++.1.dylib
    /usr/lib/libiconv.2.dylib
    /usr/lib/libSystem.B.dylib
```

## Size Overview

+73,938 / −923 against `origin/main`, with the majority coming from tests:

| Bucket | Lines |                                              |
|---|---:|----------------------------------------------|
| Python `src/` | **+4,825 / −580** | **the entire user-visible surface**          |
| C++ (pybammsolvers) | +1,390 / −82 | backend integration, pbar + `IDASetSensParams` |
| Rust `src/` | +20,305 | new, self-contained, behind an FFI boundary  |
| Rust + Python tests | +38,769 / −69 | Rust: 24,886 / Python: 13,883                |
| Benchmarks | +4,678 | the observability suite above                |
| Docs / CI / lockfiles / build glue | +3,971 / −192 | incl. `uv.lock` and the Rust `Cargo.lock`    |

## Reproducing
From the root directory, run the below command. Additional flags are documented in the benchmark suite.

```bash
uv run python benchmarks/run_rust_observability.py \
    --lane solver --models SPM SPMe DFN \
    --protocols cc_discharge drive_cycle \
    --warmup 5 --aot all --output-points 1000 --json solver_1000.json
```

Every JSON payload embeds its own provenance (git revision, dirty flag, platform, package versions).


martinjrobins:
awesome @BradyPlanden :) I'll have a more detailed look at the code, and I'll look into the diffsol slowdowns in comparison with idaklu, some possible sources are:
- as you mentioned, you might be error controlling the sensitivity equations whereas sundials does not do this be default. Note that recently I changed the diffsol defaults to match sundials, so this depends on the version you are using
- the faer cpu backend is slower than the nalgebra backend for small state sizes
- the faer lu solver can be slower than suitesparse's klu, I also provide a ffi interface to klu, so you could try that as well.
- any number of solver configuration differences between sundials and diffsol. I mostly provide similar configuration options now, so you should be able to match the config to what we use for idaklu

packages/pybamm-rust/pybamm-core/src/solver/equations.rs
```diff
@@ -0,0 +1,193 @@
...
+/// trait.
+///
+/// diffsol mints an operator view per callback invocation — `eqn.rhs()` on every
+/// residual call, and again for every `nstates()` — so the views borrow from
+/// here instead of owning: a mint is a few pointer copies, and the compiler,
+/// not a refcount, keeps this alive for as long as a view exists.
+///
+/// `Arc` marks the state shared with the `PreparedSolver` across solves, which
+/// is `Send + Sync`; everything built per solve is owned outright.
+pub struct Equations {
```
martinjrobins, line 37: you already have a ModelEvaluator struct, so this struct seems to be duplicating this
functionality

packages/pybamm-rust/pybamm-core/src/solver/equations.rs
```diff
@@ -0,0 +1,193 @@
...
+    pub(crate) compiled: Arc<CompiledModel>,
+    /// Shared with the caller, which evaluates output tapes against the same
+    /// scratch between steps on the output-variable path.
+    pub(crate) ws: Rc<RefCell<Workspace>>,
```
martinjrobins, line 41: why the Rc? should only need a RefCell, then either share a reference to your
ops, or better yet use a ModelEvaluator struct containing a RefCell and share a referfence
to that.

packages/pybamm-rust/pybamm-core/src/solver/reset.rs
```diff
@@ -0,0 +1,79 @@
+//! Placeholder reset operator.
+//!
+//! `PyBaMM` events terminate the solve rather than triggering a state reset, so
+//! `Equations::reset()` always returns `None` and `ResetOp` is never
+//! constructed. It exists solely to satisfy the `Reset` associated-type bounds
+//! of `OdeEquationsImplicitSens`.
+
```
martinjrobins, line 7: don't need this, use diffsol's UnitCallable

martinjrobins, line 7: if you use pybamm's unified experiment mode, this will give you the reset and you can then
get proper gradients through the reset.

packages/pybamm-rust/pybamm-core/src/solver/rhs.rs
```diff
@@ -0,0 +1,142 @@
...
+use super::{FaerSparsity, dense_faer_sparsity};
+use crate::model::{CompiledModel, Workspace};
+
+/// RHS operator f(t, y; p), borrowed from the [`Equations`](super::equations::Equations)
+/// that owns the solve.
+///
+/// diffsol mints one of these per callback invocation, so it holds nothing of
+/// its own: every field is a reference into the equations, making a mint a
+/// handful of pointer copies.
+pub struct RhsOp<'a> {
```
martinjrobins, line 25: second paragraph of this docstring is redundent

packages/pybamm-rust/pybamm-core/src/solver/solve.rs
```diff
@@ -0,0 +1,3072 @@
...
+///
+/// Holds the shared immutable `CompiledModel`, the tolerances and integrator
+/// options, pre-converted sparsity patterns, and the system sizes. It is *not*
+/// a fully specified problem: `y0`, `t_eval` and the parameter vector all arrive
+/// per call, and each call to [`solve`](Self::solve) builds its own
+/// `Workspace`, its own diffsol `OdeSolverProblem` and its own diffsol solver
+/// from them. Carrying no
+/// per-solve mutable state is what makes this `Send + Sync` and safe to share
+/// across threads.
+pub struct PreparedSolver {
```
martinjrobins, line 268: you are probably slowing down each solve unneccesarily due to creating an OdeSolverProblem
each time, the builder is kinda slow and does lots of allocations, it was only ever meant to be
called once and your OdeEquations struct is then meant to run multiple solves.
I would get rid of PreparedSolve and use your OdeEquations struct instead.

packages/pybamm-rust/pybamm-python/src/evaluator_pool.rs
```diff
@@ -0,0 +1,102 @@
+//! N independent [`ModelEvaluator`]s over one shared compiled tape.
+//!
+//! An `IDAKLUSolverGroup` runs one solver per OpenMP thread, and each solver
+//! evaluates through the FFI with `&mut ModelEvaluator`. The tape is not what
+//! stops that being concurrent — the scratch is, so the pool mints one
+//! `Workspace` per solver against the one immutable `CompiledModel` behind the
+//! `Arc`. N independent `CompiledModel`s would mean N lowerings and N tape
+//! copies, which is the cost the `Arc` exists to avoid.
+
```
martinjrobins, line 9: why not just hold the scratch as a global in thread local storage rather than pass it to C++ via python?

packages/pybamm-rust/pybamm-python/src/evaluator_pool.rs
```diff
@@ -0,0 +1,102 @@
...
+/// N [`ModelEvaluator`]s over one shared compiled tape.
+///
+/// `UnsafeCell` because C++ writes through each address for as long as the
+/// solver group lives, so a `*const` derived from a shared reference and cast
+/// to `*mut` would be the wrong provenance under Stacked/Tree Borrows. The
+/// `Vec` is filled once in [`from_compiled`](Self::from_compiled) and never
+/// resized, so the addresses it hands out stay valid for the pool's life.
+#[pyclass(module = "pybamm.rust")]
+pub struct EvaluatorPool {
+    evaluators: Vec<UnsafeCell<ModelEvaluator>>,
```
martinjrobins, line 36: seems like you are manually implementing a Vec of Mutex's?
Can EvaluatorPool just be Vec<Mutex<ModelEvaluator>>? That would be simpler and safer.

packages/pybamm-rust/pybamm-python/src/expr.rs
```diff
@@ -0,0 +1,962 @@
+// PyO3 bindings require specific argument types that clippy flags incorrectly
+#![allow(clippy::needless_pass_by_value)]
+
+use std::collections::HashMap;
```
martinjrobins, line 4: there is a lot of logic in here for contructing Node's that should be in core

packages/pybamm-rust/pybamm-python/src/expr.rs
```diff
@@ -0,0 +1,962 @@
...
+    ///
+    /// `width` is the number of packed values the parameter occupies (>1 for
+    /// vector-valued inputs). Re-registering an existing name must repeat the
+    /// same width, mismatches would silently corrupt every other parameter's
+    /// offset into the packed `p` array.
+    #[pyo3(signature = (name, width = 1))]
+    fn input_parameter(slf: Py<Self>, py: Python<'_>, name: &str, width: usize) -> PyResult<Expr> {
+        let mut graph = slf.borrow_mut(py);
+        let index = if let Some(&idx) = graph.input_map.get(name) {
+            let existing_width = graph.input_widths[idx];
```
martinjrobins, line 196: seems like input_map / input_widths should be a single struct in core where
you can do this check. Or perhaps alloc could have a branch for InputParameter where
it does the offset calculation internally.

packages/pybamm-rust/pybamm-python/src/expr.rs
```diff
@@ -0,0 +1,962 @@
...
+        expr: &Expr,
+        name: Option<String>,
+        n_states: Option<usize>,
+    ) -> PyResult<crate::function::CompiledFunction> {
+        let node_id = expr.node_id_in(&slf)?;
+        check_supported(slf.borrow(py).arena(), node_id)?;
+        crate::function::CompiledFunction::build(py, slf, node_id, name, n_states)
+    }
+
+    /// Compile a named set of outputs into ONE tape with cross-output
```
martinjrobins, line 784: why have both CompiledFunctionGroup and CompiledFunction? can you just use compile_group with a single
output?

packages/pybamm-rust/pybamm-python/src/expr.rs
```diff
@@ -0,0 +1,962 @@
...
+    /// sharing: synthetic concat root + recorded slice offsets.
+    #[pyo3(signature = (outputs, name = None, n_states = None))]
+    fn compile_group(
+        slf: Py<Self>,
+        py: Python<'_>,
+        outputs: &Bound<'_, pyo3::types::PyDict>,
+        name: Option<String>,
+        n_states: Option<usize>,
+    ) -> PyResult<crate::group::CompiledFunctionGroup> {
+        let mut names = Vec::with_capacity(outputs.len());
```
martinjrobins, line 794: could use unzip to return both names and ids from a map

packages/pybamm-rust/pybamm-python/src/function.rs
```diff
@@ -0,0 +1,903 @@
...
+    pool: ScratchPool<Buffer>,
+}
+
+/// `(graph, root, name, n_states)`, the `__reduce__`/`_rebuild` argument
+/// tuple for `CompiledFunction`'s pickle protocol.
+type RebuildArgs = (Py<ExprGraph>, u32, Option<String>, Option<usize>);
+
+// `module` is required so pickle can locate the class as `pybamm.rust.CompiledFunction`
+// instead of the pyo3 default `builtins.CompiledFunction`, which pickle cannot import.
+#[pyclass(frozen, module = "pybamm.rust")]
```
martinjrobins, line 44: this is a strange comment since you do this everywhere, and is obvious since that is the
purpose of the module arg.

martinjrobins, line 44: missing a docstring

martinjrobins, line 44: filename should be compiled_function.rs

packages/pybamm-rust/pybamm-python/src/function.rs
```diff
@@ -0,0 +1,903 @@
...
+pub struct CompiledFunction {
+    pub(crate) expr: Arc<CompiledExpr>,
+    pub(crate) sig: FunctionSignature,
+    pub(crate) pool: ScratchPool<Buffer>,
+    /// Retained for lazy derivation (jacobian/jvp). Always present:
+    /// bundle views retain the bundle's graph too.
+    pub(crate) graph: Py<ExprGraph>,
+    pub(crate) root: NodeId,
+    /// Memoised JVP tape, wrt y (lazily derived on first jvp call).
```
martinjrobins, line 53: why OnceLock? the docs say this is for a global or static value, not a field in a
struct, why not just Option<TangentEntry>?

packages/pybamm-rust/pybamm-python/src/group.rs
```diff
@@ -0,0 +1,404 @@
...
+impl std::fmt::Debug for CompiledFunctionGroup {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("CompiledFunctionGroup")
+            .field("sig", &self.sig)
+            .field("out_names", &self.out_names)
+            .finish_non_exhaustive()
+    }
+}
+
+impl CompiledFunctionGroup {
```
martinjrobins, line 43: filename should be compiled_function_group.rs

martinjrobins, line 43: having both CompiledFunction and CompiledFunctionGroup is a bit redundant, just have
CompiledFunctionGroup.

packages/pybamm-rust/pybamm-python/src/jacobian.rs
```diff
@@ -0,0 +1,159 @@
...
+use pybamm_core::JacobianData;
+
+use crate::scratch::ScratchPool;
+use crate::signature::FunctionSignature;
+
+#[pyclass(frozen, module = "pybamm.rust")]
+pub struct CompiledJacobian {
+    pub(crate) data: Arc<JacobianData>,
+    pub(crate) sig: FunctionSignature,
+    scratch_pool: ScratchPool<Arc<JacobianData>>,
```
martinjrobins, line 22: why does CompiledJacobian have a scratch pool but CompiledModel has a single scratch field?

martinjrobins, line 22: why is there both data and scratch_pool? it feels like there is a lot of thread pool
related stuff going into pybamm-python that could be eliminated by just making the relevent
parts of pybamm-core thread safe.

packages/pybamm-rust/pybamm-python/src/model.rs
```diff
@@ -0,0 +1,890 @@
...
+/// The Python face of [`core_model::CompiledModel`]: the same immutable
+/// artifact behind the same `Arc`, plus the retained derivation graph and the
+/// bundle accessors (`rhs`, `jacobian`, `outputs`, `events`,
+/// `algebraic_residual`, `algebraic_jacobian`) composed over the same tapes.
+///
+/// Holding the artifact rather than an evaluator is what keeps it shareable;
+/// solvers take their per-solve state from `evaluator_pool`.
+// `module` is required so pickle can locate the class as `pybamm.rust.CompiledModel`
+// instead of the pyo3 default `builtins.CompiledModel`, which pickle cannot import.
+#[pyclass(module = "pybamm.rust")]
```
martinjrobins, line 48: remove this comment

packages/pybamm-rust/pybamm-python/src/pool.rs
```diff
@@ -0,0 +1,79 @@
...
+//!
+//! One solver per experiment step is the common case, so pools are keyed by
+//! count rather than owned per solver: ten `DiffsolSolver`s asking for 8 threads
+//! share one pool of 8, not eighty threads. Idle rayon workers park on a condvar
+//! after a short spin, so a cached pool costs stack address space and no CPU.
+//!
+//! `ThreadPool::install` re-entered from inside the same pool runs inline, so
+//! nested batch calls cannot deadlock.
+
+use std::collections::HashMap;
```
martinjrobins, line 16: calling the filename rayon_pool would match evaluator_pool better.

packages/pybamm-rust/pybamm-python/src/scratch.rs
```diff
@@ -0,0 +1,115 @@
+use std::sync::Mutex;
+
+use pybamm_core::{JacobianData, JacobianScratch};
+
+/// How a [`ScratchPool`] makes a buffer when it has none to hand out.
+///
+/// The recipe rather than a spare buffer: a pool that minted by cloning a
+/// prototype would both retain one extra set of buffers for its lifetime and
+/// memcpy them on every miss, where the shape alone is what a caller needs.
+pub trait Mint {
```
martinjrobins, line 10: why have a Mint trait at all? Why not have a Scratch trait and your pool is then struct
ScratchPool<S: Scratch> { buffers: Mutex<Vec<S>> }, which makes more sense to me.

packages/pybamm-rust/pybamm-python/src/scratch.rs
```diff
@@ -0,0 +1,115 @@
...
+
+impl Mint for std::sync::Arc<JacobianData> {
+    type Item = JacobianScratch;
+
+    fn mint(&self) -> Self::Item {
+        JacobianScratch::new(self)
+    }
+}
+
+/// Reuse pool for the per-call scratch an evaluation needs.
```
martinjrobins, line 38: filename should be scratch_pool.rs.

martinjrobins, line 38: zero information in this docstring, more explanation needed. I don't get how this works
since the scratch field is held by CompiledModel and is of type Workspace...?

packages/pybamm-rust/pybamm-python/src/signature.rs
```diff
@@ -0,0 +1,212 @@
...
+    pub input_names: Vec<String>,
+    /// Packed width per input (>1 for vector-valued inputs), parallel to `input_names`.
+    pub input_widths: Vec<usize>,
+    pub n_states: usize,
+    pub uses_y_dot: bool,
+    pub output_len: usize,
+    pub name: Option<String>,
+}
+
+impl FunctionSignature {
```
martinjrobins, line 18: filename should be function_signature.rs

martinjrobins, line 18: I spent a long time trying to find out where this was used in the ffi, but I don't think it
is. perhaps it should be called FunctionMetadata.

packages/pybamm/src/pybamm/solvers/observation.py
```diff
@@ -0,0 +1,754 @@
...
+import pybamm
+from pybamm.solvers.variable_observer import (
+    CasadiObserver,
+    NativeObserver,
+    check_variable_in_solve,
+    native_sensitivities,
+)
+
+
+class ObservationBackend(ABC):
```
martinjrobins, line 34: Observation sounds like a data measurement, perhaps SolutionBackend?

packages/pybamm/src/pybamm/solvers/observation.py
```diff
@@ -0,0 +1,754 @@
...
+    """How a Solution lowers a variable name into evaluable leaves.
```
martinjrobins, line 35: unclear what a leaf is?
I read this a couple of times but can't figure it out. what is the key
in backend[key]? is it a variable name? a segment index? a slice of segments?
Looking further down, it looks like the backends use interior mutability,
so you might want to say they provide an immutable api.

packages/pybamm/src/pybamm/solvers/observation.py
```diff
@@ -0,0 +1,754 @@
...
+    sub-solutions. ``backend[key]`` restricts them to a slice of that run and
+    :func:`join_observations` concatenates runs, so a derived Solution copies
+    one field instead of a bundle of parallel ones.
+    """
+
+    @abstractmethod
+    def __getitem__(self, key: slice) -> ObservationBackend:
+        """This backend restricted to a slice of the Solution's segments."""
+
+    @abstractmethod
```
martinjrobins, line 47: missing return type on signature

martinjrobins (commented):
Hi @BradyPlanden, I've had a first pass through, I skimmed a lot of it cause it is kinda big :) so I may have misunderstood some things, but these are my initial comments. I'll keep on having a look when I can

BradyPlanden:
Thanks for the initial review @martinjrobins! I'm now on hols now for a few weeks so once I'm back I'll take a deeper look.  From an initial glance, I think you've correctly found some of the duplications / sloppy abstracts that have crept into the implementation over the few months of building. I'm working towards tightening these at the moment and will update this branch with the corrections. That said, I'm not very keen on supporting this branch long-term, so I will start splitting this into smaller PRs pretty quickly :)

PR #5743, opened 2026-09-02
Fix bencher CI

pipliggins:
Fix for currently failing benchmark CI.

* Fixes the format for benchmark_pr.yaml after a _nox.yml change slipped through

Within the repo secrets, I updated the BENCHER key from a user key to a project key (`bencher_run_***`) so the benchmars_main/track workflows can run correctly.


martinjrobins (approved):
Looks good 👍

Issue #2551, opened 2022-12-14
Electrolyte concentration not conserved if surface form is algebraic


tinosulzer:
```python3
import pybamm

pybamm.set_logging_level("INFO")
# load models
models = [
    pybamm.lithium_ion.DFN(name="standard"),
    pybamm.lithium_ion.DFN({"surface form": "algebraic"}, name="algebraic"),
]


sims = []
for model in models:
    sim = pybamm.Simulation(model)
    sim.solve([0, 3600])
    sims.append(sim)

# plot
pybamm.dynamic_plot(
    sims,
    [
        "Total lithium in particles [mol]",
        "Total lithium in electrolyte [mol]",
    ],
)
```
![Figure_1](https://user-images.githubusercontent.com/20817509/207479565-1a78b4d0-15a7-48d3-b195-2bd572330025.png)

This is because of how the current density is concatenatedhttps://github.com/pybamm-team/PyBaMM/blob/develop/pybamm/spatial_methods/finite_volume.py#L1122, and can be fixed by using `t_+ * j` instead of `div(t_+ * i_e)` in the electrolyte diffusion equation

PR #5747, opened 2026-09-07
Keep `t_plus * i_e / F` inside the electrolyte flux in the Basic* models (#5745)

### LINKED ISSUE #5745: [Bug]: BasicDFN electrolyte equation does not conserve lithium when the transference number depends on concentration
### PyBaMM Version

main at `258fdc8` (2026-09-05); also reproduced on v26.4.1 (`d9e5afb`) and the PR #5524 head (`719763b`)

### Python Version

3.12.13

### Describe the bug

Split out from #5700 at @NicolaCourtier's suggestion.

`BasicDFN` writes the electrolyte balance as

```
eps * dc_e/dt = -div(-tor * D_e(c_e,T) * grad(c_e)) + (1 - t_plus(c_e, T)) * a_j / F
```

(`basic_dfn.py`, lines 230-233 at `258fdc8`), with a purely diffusive flux `N_e`. That form is equivalent to the conservative cation balance only when `t_plus` is constant. With `t_plus = t_plus(c_e)` the conservative form is

```
eps * dc_e/dt = -div(-tor * D_e * grad(c_e) + t_plus * i_e / F) + a_j / F
```

and expanding the divergence gives the `BasicDFN` form plus an extra `-i_e * d(t_plus)/dx / F`, which `BasicDFN` omits. Integrating the `BasicDFN` equation together with the particle equations over the cell gives

```
dN/dt = -A * integral( t_plus(c_e) * a_j / F ) dx
```

which is not zero when `t_plus` varies in x. The modular `Full` electrolyte submodel keeps `t_plus * i_e / F` inside the flux (`full_diffusion.py`, lines 72-76 and 99), which is why the full `DFN` conserves lithium with the same parameters. #2666 / #2758 fixed exactly this for the modular submodel in 2023; the `Basic*` models were not part of that change.

**Observed** (ORegan2022, which has a concentration-dependent `t_plus`; negative particle diffusivity 3.3e-14; the protocol from #5700; initial SOC 0.45; 20 cells per x-domain, 20 radial shells; `IDAKLUSolver(rtol=atol=1e-8)`):

| Case | max abs(N - N0) / N0 |
|---|---:|
| `BasicDFN`, variable `t_plus` | 2.44e-2 |
| `BasicDFN`, `t_plus` fixed at its initial value | 1.9e-14 |
| `DFN()`, same parameters | 1.4e-14 |

- The drift is insensitive to tolerance and mesh: max abs(N - N0) is 6.518e-3 mol at 20/20 for rtol=atol=1e-6, 1e-8 and 1e-10, 6.516e-3 mol at 40/40 and 6.5155e-3 mol at 80/80 (1e-10). Time-integration error is not the explanation, and refinement does not remove it.
- The measured change in N matches the time integral of `-A * integral(t_plus * a_j / F) dx` from the solution to 1.2e-6 mol out of 6.5e-3 mol (5 s trapezoid).
- Because the constant-`t_plus` case conserves to round-off with the same node-evaluated source term, this is separate from the finite-volume source-term discretisation question raised in #5700. It is a property of the reduced equation, not of how `a_j` is evaluated on nodes, and a boundary-condition change alone does not remove it (PR #5524 leaves the `BasicDFN` time series byte-identical to main; the model declares no `"Flux"` boundary condition).

The same reduced form appears in `BasicDFN2D` (`basic_dfn_2d.py` line 302), `BasicDFNHalfCell` (`basic_dfn_half_cell.py` line 197) and `BasicDFNComposite` (`basic_dfn_composite.py` line 323) at `258fdc8`. The `BasicDFN2D` drift @SarahRo reported in #5700 is consistent with this. I have only measured the 1D `BasicDFN`.

Two reasonable resolutions, and the current state is neither:

1. Put `t_plus * i_e / F` inside `N_e` in the `Basic*` models, as `full_diffusion.py` does. #2551 records that `div(t_plus * i_e)` once hit a concatenation problem in one configuration; I have not tested whether that applies here.
2. State in the `Basic*` docstrings that they assume a constant `t_plus`, and warn or raise when a function is supplied. The current docstring cites Marquis 2019 and says nothing about `t_plus`, and `ORegan2022` is accepted silently.

### Steps to Reproduce

Run against main `258fdc8`; the two lines under "Relevant log output" are what it printed.

```python
import pybamm

    "Rest for 10 minutes (5 seconds period)",
    "Discharge at 1C for 20 minutes (5 seconds period)",
    "Rest for 10 minutes (5 seconds period)",
    "Charge at 1C for 20 minutes (5 seconds period)",
    "Rest for 20 minutes (5 seconds period)",
])

for constant_t_plus in [False, True]:
    model = pybamm.lithium_ion.BasicDFN()
    pv = pybamm.ParameterValues("ORegan2022")
    pv["Negative particle diffusivity [m2.s-1]"] = 3.3e-14
    if constant_t_plus:
        pv["Cation transference number"] = pv["Cation transference number"](
            pv["Initial concentration in electrolyte [mol.m-3]"], pv["Ambient temperature [K]"]
        )
    # lithium per unit area, mol/m2 (BasicDFN has no built-in "Total lithium [mol]")
    n = 0
    for dom, eps in [("Negative", "Negative electrode porosity"),
                     ("Separator", "Separator porosity"),
                     ("Positive", "Positive electrode porosity")]:
        thick = "Separator thickness [m]" if dom == "Separator" else f"{dom} electrode thickness [m]"
        n += pybamm.x_average(model.variables[f"{dom} electrolyte concentration [mol.m-3]"]
                              * pybamm.Parameter(eps)) * pybamm.Parameter(thick)
    for dom in ["Negative", "Positive"]:
        n += pybamm.x_average(pybamm.Parameter(f"{dom} electrode active material volume fraction")
                              * pybamm.r_average(model.variables[f"{dom} particle concentration [mol.m-3]"])) \
             * pybamm.Parameter(f"{dom} electrode thickness [m]")
    model.variables["Lithium per area [mol.m-2]"] = n
    sim = pybamm.Simulation(model, parameter_values=pv, experiment=experiment,
                            solver=pybamm.IDAKLUSolver(rtol=1e-8, atol=1e-8),
                            var_pts={"x_n": 20, "x_s": 20, "x_p": 20, "r_n": 20, "r_p": 20})
    sol = sim.solve(initial_soc=0.45)
    li = sol["Lithium per area [mol.m-2]"].entries
    print(f"constant t+ = {constant_t_plus}: max |N - N0| / N0 = {abs(li - li[0]).max() / li[0]:.3e}")
```

The refinement table and the comparison with `DFN()` come from a direct finite-volume sum over cell volumes rather than `x_average`; code and per-run outputs are in the package linked from #5700 (https://github.com/repowazdogz-droid/pybamm-5700-lithium-inventory, commit `adaa877`).

### Relevant log output

```
constant t+ = False: max |N - N0| / N0 = 2.442e-02
constant t+ = True: max |N - N0| / N0 = 1.982e-14
```

repowazdogz-droid:
# Final PR — PyBaMM #5745 (rebased 2026-09-07)

**Title**

Keep `t_plus * i_e / F` inside the electrolyte flux in the Basic* models (#5745)

**Body**

Fixes #5745.

The Basic electrolyte models put the `(1 - t_plus) a_j / F` source term in the
concentration equation while omitting the corresponding `t_plus * i_e / F`
contribution from the conservative electrolyte flux. This is harmless for
constant `t_plus`, but violates the conservative balance when `t_plus` varies
spatially (the omitted term is `-i_e . grad(t_plus) / F`).

This change applies the conservative-flux formulation to `BasicDFN`,
`BasicDFN2D`, `BasicDFNComposite`, and `BasicDFNHalfCell`: `N_e` now includes
`t_plus(c_e, T) * i_e / F` and the concentration source is `a_j / F`, matching
the modular `Full` electrolyte submodel. The half-cell ordering is adjusted so
the current block is available before the concentration block; boundary
conditions and the reference-potential convention are unchanged.

Regression evidence (independent finite-volume lithium inventory, #5745
protocol with `ORegan2022`, which has a concentration-dependent transference
number):

- `BasicDFN` variable `t_plus`: relative drift falls from ~2.44e-2 to ~2e-14
  (solver floor) at every mesh (20/20, 40/40, 80/80) and tolerance tested.
- `BasicDFN` constant `t_plus` and modular `DFN` controls: unchanged (~2e-14).
- `BasicDFNComposite`: 2.17e-2 to ~2e-14.
- `BasicDFN2D`: 4.4e-3 to ~3.5e-16.
- `BasicDFNHalfCell`: 3.55e-2 down to a ~1e-6 floor (2.3e-6 at 20 cells,
  scaling ~dx^2). The same floor is present for constant `t_plus` before and
  after the patch, so it is a pre-existing boundary-treatment floor, not exact
  conservation; it is not claimed as zero.

The #2551 concatenation gate passes: the discrete identity telescopes to
round-off (~1e-14) on the patched `BasicDFN`, while still detecting the known
~2.4e-5 residual on `DFN` with `surface form: algebraic`, confirming the #2551
mechanism is structurally absent from the Basic models (their `i_e` is built
once on the concatenated domain; no edge-valued quantity is concatenated).

Tests (all on current `main`, `258fdc8c1`; the rebase was a no-op as `main`
had not moved):

- New `TestElectrolyteConservation` (`test_basic_dfn`,
  `test_basic_dfn_composite`): fail on unpatched models, pass with the fix.
- Basic model unit + integration suites
  (`test_basic_models.py` unit and integration, `test_compare_basic_models.py`):
  15 passed.
- Full `tests/unit/.../test_lithium_ion/` directory: 640 passed, 12 skipped
  (pre-existing skips).
- `ruff@0.16.3` check and format: clean on all touched files.

The half-cell ~1e-6 residual floor is retained and described as above. No
claim is made about `surface form: algebraic`, non-Basic models, or thermal
variants.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Bug fixes
 
+- `BasicDFN`, `BasicDFN2D`, `BasicDFNHalfCell` and `BasicDFNComposite` now keep the migration term `t_plus * i_e / F` inside the electrolyte flux, as the modular `Full` electrolyte submodel does, so the electrolyte balance conserves lithium when the transference number depends on concentration (for example `ORegan2022`). ([#5745](https://github.com/pybamm-team/PyBaMM/issues/5745))
 - The `integration` nox session no longer installs the `pydiffsol` extra on macOS Intel CI runners, where it has no working build. ([#5726](https://github.com/pybamm-team/PyBaMM/pull/5726))
 
 # [v26.8.0.0](https://github.com/pybamm-team/PyBaMM/tree/pybamm-v26.8.0.0) - 2026-08-13
--- a/packages/pybamm/src/pybamm/models/full_battery_models/lithium_ion/basic_dfn.py
+++ b/packages/pybamm/src/pybamm/models/full_battery_models/lithium_ion/basic_dfn.py
@@ -227,10 +227,13 @@ class BasicDFN(BaseModel):
         ######################
         # Electrolyte concentration
         ######################
-        N_e = -tor * self.param.D_e(c_e, T) * pybamm.grad(c_e)
-        self.rhs[c_e] = (1 / eps) * (
-            -pybamm.div(N_e) + (1 - self.param.t_plus(c_e, T)) * a_j / self.param.F
+        # The migration term t_plus * i_e / F is kept inside the flux so that the
+        # balance is conservative when t_plus depends on c_e (see #5745)
+        N_e = (
+            -tor * self.param.D_e(c_e, T) * pybamm.grad(c_e)
+            + self.param.t_plus(c_e, T) * i_e / self.param.F
         )
+        self.rhs[c_e] = (1 / eps) * (-pybamm.div(N_e) + a_j / self.param.F)
         self.boundary_conditions[c_e] = {
             "left": (pybamm.Scalar(0), "Neumann"),
             "right": (pybamm.Scalar(0), "Neumann"),
--- a/packages/pybamm/src/pybamm/models/full_battery_models/lithium_ion/basic_dfn_2d.py
+++ b/packages/pybamm/src/pybamm/models/full_battery_models/lithium_ion/basic_dfn_2d.py
@@ -295,12 +295,16 @@ class BasicDFN2D(BaseModel):
         ######################
         # Electrolyte concentration
         ######################
-        N_e = pybamm.VectorField(
-            -tor * self.param.D_e(c_e, T), -tor * self.param.D_e(c_e, T)
-        ) * pybamm.grad(c_e)
-        self.rhs[c_e] = (1 / eps) * (
-            -pybamm.div(N_e) + (1 - self.param.t_plus(c_e, T)) * a_j / self.param.F
+        # The migration term t_plus * i_e / F is kept inside the flux so that the
+        # balance is conservative when t_plus depends on c_e (see #5745)
+        N_e = (
+            pybamm.VectorField(
+                -tor * self.param.D_e(c_e, T), -tor * self.param.D_e(c_e, T)
+            )
+            * pybamm.grad(c_e)
+            + self.param.t_plus(c_e, T) * i_e / self.param.F
         )
+        self.rhs[c_e] = (1 / eps) * (-pybamm.div(N_e) + a_j / self.param.F)
         self.boundary_conditions[c_e] = {
             "left": (pybamm.Scalar(0), "Neumann"),
             "right": (pybamm.Scalar(0), "Neumann"),
--- a/packages/pybamm/src/pybamm/models/full_battery_models/lithium_ion/basic_dfn_composite.py
+++ b/packages/pybamm/src/pybamm/models/full_battery_models/lithium_ion/basic_dfn_composite.py
@@ -318,10 +318,13 @@ class BasicDFNComposite(BaseModel):
         ######################
         # Electrolyte concentration
         ######################
-        N_e = -tor * self.param.D_e(c_e, T) * pybamm.grad(c_e)
-        self.rhs[c_e] = (1 / eps) * (
-            -pybamm.div(N_e) + (1 - self.param.t_plus(c_e, T)) * a_j / self.param.F
+        # The migration term t_plus * i_e / F is kept inside the flux so that the
+        # balance is conservative when t_plus depends on c_e (see #5745)
+        N_e = (
+            -tor * self.param.D_e(c_e, T) * pybamm.grad(c_e)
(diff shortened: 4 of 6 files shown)


rtimms:
Thanks @repowazdogz-droid LGTM! Can you make the same fix in [sodium_ion/basic_dfn.py:224](https://github.com/pybamm-team/PyBaMM/blob/main/packages/pybamm/src/pybamm/models/full_battery_models/sodium_ion/basic_dfn.py#L224-L227)

repowazdogz-droid:
Sodium-ion BasicDFN now has the analogous conservative-flux fix. Regression coverage was added, and the relevant sodium/lithium tests and style checks pass. The CHANGELOG conflict was resolved while preserving both entries.

rtimms (approved):
Thanks!

PR #5753, opened 2026-09-15
Fix parameter discovery in Variable metadata

rtimms:
## Summary
- include Variable scale, reference, and bounds metadata in BaseModel parameter discovery
- scan equation dictionary keys so discovery matches parameter substitution
- add a regression test for metadata-only parameters

## Testing
- uv run --no-sync --group dev pytest -o addopts='' -n auto --strict-config --strict-markers -m unit packages/pybamm/tests
- uv run --no-sync --group dev pre-commit run --files CHANGELOG.md packages/pybamm/src/pybamm/models/base_model.py packages/pybamm/tests/unit/test_models/test_base_model.py


packages/pybamm/src/pybamm/models/base_model.py
```diff
@@ -1149,8 +1154,24 @@ def _find_symbols_by_submodel(
...
+        variable_metadata = [
+            metadata
+            for variable in variables
+            for metadata in (variable.scale, variable.reference, *variable.bounds)
+        ]
+        return list(
+            pybamm.SymbolUnpacker(typ).unpack_list_of_symbols(
+                all_items + variable_metadata
+            )
+        )
```
MarcBerliner: I think the checks on `scale`, `reference`, and `bounds` should be upstreamed into `SymbolUnpacker`'s `_unpack` method

rtimms: Implemented in c7a23a72e. SymbolUnpacker now traverses a Variable’s scale, reference, and bounds; the BaseModel-specific helper is removed, while BaseModel retains equation-key traversal. I also added direct unpacker regression coverage.

MarcBerliner (approved):
Thanks!

PR #5756, opened 2026-09-17
Add Interpolant support to DiffSLExport

TomTranter:
# Description

Adds `Interpolant` support to `DiffSLExport`, so models whose parameters are lookup tables (OCP, diffusivity, reaction rates, …) can be exported to DiffSL. Previously the exporter had no `Interpolant` branch, so such parameter sets could not be exported at all.

- **1D interpolants** are emitted using DiffSL's native `interp1d` over the table data (declared as constant tensors). This is piecewise-linear with endpoint clamping — the same as `numpy.interp` and pybamm's own linear interpolant — and stays compact for any table size.
- **2D interpolants** use successive 1D interpolation: `interp1d` along the inner axis at each outer breakpoint, then a linear blend along the outer axis (until DiffSL gains an `interp2d`).

This PR also carries one prerequisite bug fix: a `Vector` produced by a primary broadcast dropped its domain during discretisation, leaving downstream operations (including the export) with an empty domain. The fix carries the broadcast's domain onto the `Vector`.

Motivation: exporting to DiffSL lets a discretised PyBaMM model be solved by [diffsol](https://github.com/martinjrobins/diffsol) (Rust). Interpolated parameters are common in real cell models, so `Interpolant` support is needed for those models to export.

Fixes # (no linked issue)

## Type of change

CHANGELOG.md entries added under `# [Unreleased]`:
- `## Features` — `Interpolant` export via `interp1d`.
- `## Bug fixes` — primary-broadcast `Vector` domain fix.

Commits:
1. `fix(discretisation): carry domain on a primary-broadcast Vector`
2. `feat(diffsl): support Interpolant nodes in DiffSLExport`

# Important checks:

- [x] No style issues: `nox -s pre-commit` — ruff check + format clean (ruff v0.16.3, the pinned pre-commit rev).
- [x] Tests added — unit tests assert the export uses `interp1d` for 1D and 2D; an integration test solves the exported model via diffsol and compares the trajectory to an independent pybamm `CasadiSolver` solve of the same model.
- [x] Relevant tests pass — the DiffSL unit suite (34) and integration suite (29) pass locally against a build with `interp1d`/`pydiffsol`; the interpolant solve matches pybamm to 1e-5 for both small and large tables.
- [ ] Full `nox -s tests` run (please run in CI / locally with the full dev env).
- [ ] The documentation builds: `nox -s doctests`.
- [x] Code is commented for hard-to-understand areas (the 2D successive-interpolation blend and the interpolant tensor declaration).


martinjrobins:
related: #5652 adds 1D linear interpolant support for DiffSLExport

martinjrobins:
diffsl has an inbuilt interp1d function for 1d linear interpolation, I'd be happy to add others so you don't have to emit a whole bunch of diffsl code for large n interpolants

packages/pybamm/src/pybamm/spatial_methods/spatial_method.py
```diff
@@ -119,7 +119,13 @@ def broadcast(self, symbol, domains, broadcast_type):
             # Make copies of the child stacked on top of each other
             sub_vector = np.ones((primary_domain_size, 1))
             if symbol.shape_for_testing == ():
-                out = symbol * pybamm.Vector(sub_vector)
+                # Carry `domains` on the broadcast vector itself so binary-op
+                # simplifications can preserve the domain across rewrites.
+                # (Mirrors the "full" branch below; the primary branch was
```
martinjrobins: I think we can simplify this comment to just the first two lines, the rest is not really useful

packages/pybamm/src/pybamm/expression_tree/operations/diffsl.py
```diff
@@ -1136,6 +1177,204 @@ def map_inputs(self, inputs: dict, outputs: list[str] | None = None) -> np.ndarr
         return np.array(values, dtype=float)
 
 
+def _piecewise_linear_1d(x_data, y_data, arg_str, float_precision=20):
```
martinjrobins: can just use diffsl interp1d function (https://martinjrobins.github.io/diffsl/functions.html)

packages/pybamm/tests/unit/test_expression_tree/test_operations/test_diffsl_export.py
```diff
@@ -634,3 +641,144 @@ def test_unified_experiment_step_value_tensor_values(self):
...
+        model.rhs = {a: -interp, b: pybamm.Scalar(0.0)}
+        model.initial_conditions = {a: pybamm.Scalar(0.5), b: pybamm.Scalar(0.5)}
+        model.variables = {"a": a, "b": b}
+        pybamm.Discretisation().process_model(model)
+        export = pybamm.DiffSLExport(model).to_diffeq(outputs=["a"])
+
+        assert "interpolant" not in export.lower()
+
+
+class TestDiffSLInterpolantHelpers:
```
martinjrobins: I'm not a fan of testing private interfaces, it should be possible to change the internals, including private functions, without breaking tests (as long as public interface stays the same)

packages/pybamm/src/pybamm/expression_tree/operations/diffsl.py
```diff
@@ -1136,6 +1177,204 @@ def map_inputs(self, inputs: dict, outputs: list[str] | None = None) -> np.ndarr
...
+        dxi = f"{dx:.{float_precision}g}"
+        # contribution = slope * max(0, min(clamped - x_i, dx_i))
+        terms.append(
+            f"({slope:.{float_precision}g} * max(min(({clamped} - {xi}), {dxi}), 0))"
+        )
+
+    return "(" + " + ".join(terms) + ")"
+
+
+def _chebyshev_approx_1d(x_data, y_data, arg_str, degree=15, float_precision=20):
```
martinjrobins: if you use the diffsl interp1d function I guess you won't need this?

martinjrobins (changes requested):
thanks @TomTranter :) I'd be happy to add an interp2d function to diffsl as well, but this shouldn't hold up this PR if you need it sooner than that

packages/pybamm/src/pybamm/spatial_methods/spatial_method.py
TomTranter: Done — trimmed to the first two lines.

packages/pybamm/src/pybamm/expression_tree/operations/diffsl.py
TomTranter: Good call — switched the 1D path to DiffSL's native `interp1d`. The table data is declared as constant tensors and passed straight to `interp1d`, so the hand-rolled piecewise-linear emitter is gone. Much cleaner, thanks for the pointer.

packages/pybamm/src/pybamm/expression_tree/operations/diffsl.py
TomTranter: Correct — with `interp1d` the Chebyshev approximation is unnecessary (it only existed to keep the emitted source compact for large tables, which `interp1d` handles natively). Removed `_chebyshev_approx_1d` and the threshold entirely.

packages/pybamm/tests/unit/test_expression_tree/test_operations/test_diffsl_export.py
TomTranter: Agreed — dropped the private-function tests. The interpolant paths are now covered through the public `DiffSLExport` interface (asserting the export uses `interp1d`), plus the integration test that solves the exported model and matches an independent pybamm `CasadiSolver` solve.

TomTranter:
Thanks for the review @martinjrobins! For 2D I've kept successive 1D interpolation for now — `interp1d` along the inner axis and a linear blend along the outer. Happy to move it to `interp2d` whenever it lands; agreed it needn't block this.

TomTranter:
Hey @martinjrobins hope I addressed all your concerns. I've been loving using PyBaMM in rust. Thanks for bringing this to the world

packages/pybamm/tests/integration/test_solvers/test_diffsl.py
```diff
@@ -296,3 +296,55 @@ def test_simulations(self, model, inputs, experiment, request, timing_results):
         # unsure why the discrepancy occurs, maybe interpolation near events.
         # Error away from events is below solver tolerances.
         np.testing.assert_allclose(voltage_pybamm, voltage_diffsol, rtol=2e-3)
+
+    @pytest.mark.skipif(not has_pydiffsol, reason="pydiffsol is not installed")
+    @pytest.mark.parametrize(
+        "n_points, tol",
+        # Both sizes export to DiffSL's interp1d (piecewise-linear over the
```
martinjrobins: this comment and parameterisation is irrelevent after the change to using interp1d

packages/pybamm/tests/integration/test_solvers/test_diffsl.py
```diff
@@ -296,3 +296,55 @@ def test_simulations(self, model, inputs, experiment, request, timing_results):
...
+        # table), matching pybamm's own linear interpolation regardless of
+        # table size.
+        [(8, 1e-5), (60, 1e-5)],
+        ids=["small_table", "large_table"],
+    )
+    def test_interpolant_solve_matches_pybamm(self, n_points, tol):
+        # Numerical guard on the Interpolant export paths (the unit tests
```
martinjrobins: I think the test name is sufficient here, don't need the comment

packages/pybamm/tests/unit/test_expression_tree/test_operations/test_diffsl_export.py
```diff
@@ -634,3 +634,76 @@ def test_unified_experiment_step_value_tensor_values(self):
...
+        # A 1D Interpolant becomes a DiffSL interp1d call over declared table
+        # tensors — no raw Interpolant node survives, regardless of table size.
+        for n in (4, 12, 60):
+            x = np.linspace(0.0, 1.0, n)
+            model = self._interpolant_ode(x, 2.0 * x + 1.0, f"ramp{n}")
+            export = pybamm.DiffSLExport(model).to_diffeq(outputs=["y"])
+            assert "interpolant" not in export.lower()
+            assert "interp1d(" in export
+
+    def test_interpolant_large_table_stays_compact(self):
```
martinjrobins: test no longer relevent large/small tables handled exactly the same

packages/pybamm/tests/unit/test_expression_tree/test_operations/test_diffsl_export.py
```diff
@@ -634,3 +634,76 @@ def test_unified_experiment_step_value_tensor_values(self):
...
+        interp = pybamm.Interpolant(x_data, f_data, y, name=name)
+        model.rhs = {y: -interp}
+        model.initial_conditions = {y: pybamm.Scalar(0.5)}
+        model.variables = {"y": y, name: interp}
+        pybamm.Discretisation().process_model(model)
+        return model
+
+    def test_interpolant_1d_exports_interp1d(self):
+        # A 1D Interpolant becomes a DiffSL interp1d call over declared table
+        # tensors — no raw Interpolant node survives, regardless of table size.
```
martinjrobins: comment about table size is no longer relevent

packages/pybamm/tests/unit/test_expression_tree/test_operations/test_diffsl_export.py
```diff
@@ -634,3 +634,76 @@ def test_unified_experiment_step_value_tensor_values(self):
...
+        for n in (4, 12, 60):
```
martinjrobins: don't need to loop over n, size of table no longer relevent

martinjrobins (changes requested):
thanks @TomTranter, looks good, just some tidy up to do after the change to interp1d

packages/pybamm/tests/unit/test_expression_tree/test_operations/test_diffsl_export.py
TomTranter: Removed the loop — it's a single table now, since interp1d handles all sizes the same.

packages/pybamm/tests/unit/test_expression_tree/test_operations/test_diffsl_export.py
TomTranter: Removed the table-size comment.

packages/pybamm/tests/unit/test_expression_tree/test_operations/test_diffsl_export.py
TomTranter: Removed that test — it's redundant now that large and small tables both just emit an `interp1d` call.

packages/pybamm/tests/integration/test_solvers/test_diffsl.py
TomTranter: Dropped the parametrisation — it's a single case now.

packages/pybamm/tests/integration/test_solvers/test_diffsl.py
TomTranter: Removed the comment.

martinjrobins (approved):

PR #5759, opened 2026-09-18
Fix duplicated prefix in minimum particle surface concentration variable name

rtimms:
## Description

`base_particle.py` builds the dimensional minimum particle surface concentration key from an implicit string concatenation whose first line is repeated:

```python
f"Minimum {domain} {phase_name}particle "
f"Minimum {domain} {phase_name}particle "
"surface concentration [mol.m-3]": pybamm.min(c_s_surf),
```

The `Maximum` sibling immediately below has the line once, which is the correct shape. This PR deletes the duplicate.

Two consequences, both fixed here:

1. The variable was exposed under a nonsense name, which is what users see in a variable picker:

   ```
   Minimum negative particle Minimum negative particle surface concentration [mol.m-3]
   Minimum positive particle Minimum positive particle surface concentration [mol.m-3]
   ```

2. **The intended name was absent entirely.** `"Minimum negative particle surface concentration [mol.m-3]"` resolved to nothing, so the dimensional minimum surface concentration was not plottable under any sensible name. The dimensionless `"Minimum negative particle surface concentration"` and the `Maximum` counterparts were unaffected — only this one key. So the fix restores a variable rather than only renaming one.

Before (`SPMe`, v26.8.0.0):

```
'Maximum negative particle surface concentration'
'Maximum negative particle surface concentration [mol.m-3]'
'Maximum positive particle surface concentration'
'Maximum positive particle surface concentration [mol.m-3]'
'Minimum negative particle Minimum negative particle surface concentration [mol.m-3]'
'Minimum negative particle surface concentration'
'Minimum positive particle Minimum positive particle surface concentration [mol.m-3]'
'Minimum positive particle surface concentration'
```

After:

```
'Maximum negative particle surface concentration'
'Maximum negative particle surface concentration [mol.m-3]'
'Maximum positive particle surface concentration'
'Maximum positive particle surface concentration [mol.m-3]'
'Minimum negative particle surface concentration'
'Minimum negative particle surface concentration [mol.m-3]'
'Minimum positive particle surface concentration'
'Minimum positive particle surface concentration [mol.m-3]'
```

Verified on `SPM`, `SPMe` and `DFN`.

## Compatibility

This renames a public variable name, but nothing in the repository reads the doubled key — the only occurrences are stale printed variable lists in notebook *output* cells, which regenerate. Given the old name is self-evidently a typo and the new name was previously unresolvable, a deprecation alias seems like more cost than it is worth; happy to add one if maintainers prefer.

## Type of change

# Key checklist:

- [x] No style issues (`ruff format --check` and `ruff check` pass on the changed files; I did not run the full `nox -s pre-commit`)
- [ ] All tests pass: `nox -s tests` — not run locally; relying on CI. The new test and the before/after variable lists above were verified directly against a patched checkout.
- [ ] The documentation builds: `nox -s doctests` — not run locally; relying on CI.

## Further checks:

- [x] Tests added that prove fix is effective (`tests/unit/test_models/test_submodels/test_particle_variable_names.py`)


packages/pybamm/tests/unit/test_models/test_submodels/test_particle_variable_names.py
valentinsulzer, line 1: this seems a bit unnecessary

valentinsulzer (dismissed):

packages/pybamm/tests/unit/test_models/test_submodels/test_particle_variable_names.py
valentinsulzer: or should be generalized if trying to catch these things

rtimms: fair point, removed

BradyPlanden (approved):

PR #5769, opened 2026-09-21
Sum the heat of mixing over every particle phase

mleot:
# Description

`"heat of mixing": "true"` cannot be used with a composite (multi-phase)
electrode. `BaseThermal._heat_of_mixing` reads the single-phase variable names
and takes its material properties from `param.n.prim` / `param.p.prim` only, so
any model built with `"particle phases": "2"` fails at build time:

```python
import pybamm

pybamm.lithium_ion.DFN(options={
    "particle phases": ("2", "1"),
    "thermal": "lumped",
    "heat of mixing": "true",
})
```

```
ModelError: Missing variable for submodel 'thermal': 'Negative particle radius [m]'.
Check the selected submodels provide all of the required variables.
```

The same failure occurs for SPM, SPMe and DFN, and for a composite positive
(`("1", "2")`) as well as a composite negative.

**Fix.** Richardson's heat of mixing is a per-particle-population quantity, so it
is evaluated per phase and the contributions are summed. The new
`_phase_heat_of_mixing` holds the expression that was previously written out
twice inline; `_electrode_heat_of_mixing` sums it over the electrode's phases,
reusing the `phase_names = ["primary ", "secondary "]` idiom that the reaction,
reversible and hysteresis heat sources in the same class already use.
`_heat_of_mixing` keeps its signature, its `"planar"` negative-electrode guard
and its `"heat of mixing" != "true"` zero-broadcast branch.

**Single-phase behaviour is unchanged.** A single-phase electrode takes the
`num_phases == 1` branch, which builds the same expression from the same
unprefixed variable names as before. Checked numerically: DFN + `Chen2020` +
`"heat of mixing": "true"`, 1800 s at 1C, stock vs patched `base_thermal.py` on
the same `main` tree (`3378e4a32`):

| output | max abs difference |
|---|---|
| `Voltage [V]` | `0.0` |
| `Volume-averaged cell temperature [K]` | `0.0` |
| `Volume-averaged heat of mixing [W.m-3]` | `0.0` |

**Not addressed here.** `_phase_heat_of_mixing` keeps upstream's `U(...).diff(c)`
for `dU/dc`, which requires a differentiable OCP; an OCP supplied as a data
interpolant still fails at parameter processing. That is a pre-existing
limitation that applies equally to single-phase electrodes and is left for a
separate PR. Splitting one material into two identical half-phases does not
recover the single-phase answer exactly (~0.1% on voltage for `Chen2020`/DFN),
but that difference is present with `"heat of mixing": "false"` too, so it is a
property of the composite electrode model, and no test here relies on it.

Fixes # (no linked issue)

## Type of change

Bug fix. CHANGELOG entry added under `## Bug fixes`.

# Important checks:

- [x] No style issues: `nox -s pre-commit` (pre-commit run on the changed files; ruff, comment-slop and repo-review all pass)
- [x] All tests pass: `nox -s tests` (`test_full_battery_models/test_lithium_ion/`: 643 passed, 13 skipped; CI runs the rest)
- [ ] The documentation builds: `nox -s doctests` (no docs changed)
- [x] Tests added that prove fix is effective or that feature works:
  `test_well_posed_particle_phases_heat_of_mixing` in `base_lithium_ion_tests.py`
  covers `"2"`, `("2", "1")` and `("1", "2")` with `"thermal": "lumped"` for every
  model inheriting the base tests. It fails with the `ModelError` above without
  the fix. `NewmanTobias` skips it, matching the existing skips on the other
  `particle phases` tests there (it defaults to `"particle": "uniform profile"`,
  which is rejected with multiple phases regardless of heat of mixing).


packages/pybamm/src/pybamm/models/submodels/thermal/base_thermal.py
```diff
@@ -381,65 +381,65 @@ def _current_collector_heating(self, variables):
...
+            )
+        return sum(
+            self._phase_heat_of_mixing(variables, domain, phase_name, phase_param)
+            for phase_name, phase_param in zip(
+                ["primary ", "secondary "],
+                [electrode_param.prim, electrode_param.sec],
+                strict=True,
+            )
+        )
+
```
BradyPlanden, line 428: suggestion: the below would avoid hard-coding the `['primary', 'secondary']` and `[prim, sec]` while removing the multi-line branching.

```suggestion
        phases = self.options.phases[domain]
        names = [""] if len(phases) == 1 else [f"{p} " for p in phases]
        return sum(
            self._phase_heat_of_mixing(variables, domain, n, electrode_param.phase_params[p])
            for n, p in zip(names, phases, strict=True)
        )

```

BradyPlanden (commented):
Thanks @mleot - this looks good. One suggestion to tighten the phase summing.

